PR #14, opened 2020-01-21
Feature: Inversion of Control

gnarula:
All the instantiation of services and handlers is tightly coupled inside the Agent class at the moment making it difficult for users to use custom implementations or mock them without extending `Agent`.

This PR allows using a DI Container (using InversifyJS) to provide instances of all dependencies throughout the project and as a result, end users can greatly benefit from the decoupling (e.g. in tests).

Note that users should exercise caution while binding classes to the container - particularly for the classes which hold things in memory. All such classes should be declared as singletons so that future use (via `container.get()`) results in fetching the same instance

P.S. I haven't extracted interfaces from a few classes but I reckon this is a good first step.


jakubkoci:
Wow, that's a lot of code, maybe too much 😄 I would need a little bit more time to go through that in detail. 

I agree with the idea of the DI framework. I'm not 100 % sure whether inversify is the best option, I really like DI in NestJS for example. I think it would be awesome to have it.

src/lib/__tests__/agents.test.ts
```diff
@@ -22,24 +35,121 @@ const bobConfig = {
...
+    bobContainer.bind<MessageReceiver>(TYPES.MessageReceiver).to(MessageReceiver);
+    ContainerHelper.registerDefaultHandlers(bobContainer);
+    bobContainer.bind<Dispatcher>(TYPES.Dispatcher).to(Dispatcher);
+
+    bobContainer
+      .bind<Agent>(TYPES.Agent)
+      .to(Agent)
+      .inSingletonScope();
+    bobAgent = bobContainer.get<Agent>(TYPES.Agent);
+  });
```
jakubkoci: I think our goal should be to provide a library ready to use without needing much to configure and the e2e test should demonstrate this. It means to hide library internals, expose just what's necessary and allow a developer to extend or modify functionality via plugins. For example, a developer would be able to define what transport protocol, storage and aries protocols she wants to use, but she shouldn't need to know anything about Dispatcher, MessageSender, MessageReceiver and so on.

gnarula: True. I reckon a decent set of defaults can always be added on to a helper method in `ContainerHelper.ts` and the user can just get them all in by calling that method while setting up the container as I did with the handlers

gnarula:
@jakubkoci @jadhavajay This is ready for review now :) I've added a few helper methods in `ContainerHelper.ts` and updated the tests to use it.

OR13:
This PR has massive conflicts, it should be cleaned up or closed.

gnarula:
Closing it at the moment since it’s not the highest priority.

PR #15, opened 2020-01-27
DIDDoc: Adhere to DIDDoc schema

gnarula:
Modifies DID Doc to follow the specifications laid out in the W3C spec. Adopted a test from ACA-Py. Future work can be to add more test cases.


src/lib/protocols/connections/ConnectionService.ts
```diff
@@ -106,19 +107,9 @@ class ConnectionService {
...
-          id: 'did:example:123456789abcdefghi#did-communication',
-          type: 'did-communication',
-          priority: 0,
-          recipientKeys: [verkey],
-          routingKeys: this.getRoutingKeys(),
-          serviceEndpoint: this.getEndpoint(),
-        },
-      ],
-    };
+    const publicKey = new PublicKey(`${did}#1`, PublicKeyType.ED25519_SIG_2018, did, verkey, true);
```
jakubkoci: What if we use `Authentication` object and pass a `publicKey` to it instead of hiding this information into `PublicKey`. Like this:

```
const publicKey = new PublicKey(`${did}#1`, PublicKeyType.ED25519_SIG_2018, did, verkey)
const authentication = new Authentication(publicKey)
const didDoc = new DidDoc(did, authentication, [publicKey], [service]);
```

This would allow us to get rid of the last flag argument and also align more with didDoc spec, where you have `publicKey` and `authentication` attributes on the same level.

src/lib/protocols/connections/domain/DidDoc.ts
```diff
@@ -1,10 +1,157 @@
...
+  constructor(id: string, publicKey: PublicKey[], service: Service[]) {
+    this['@context'] = 'https://w3id.org/did/v1';
+    this.id = id;
+    this.publicKey = publicKey;
+    this.service = service;
+  }
+
+  toJSON() {
+    const publicKey = this.publicKey.map(element => element.toJSON());
+    const authentication = this.publicKey.map(pk => {
```
jakubkoci: Doesn't this return JSON array with `null` values?
```
[
  null,
  {
    id: '4',
    type: 'RsaVerificationKey2018',
    controller: 'did:sov:LjgpST2rjsoxYegQDRm7EL',
    publicKeyPem: '-----BEGIN PUBLIC 9...',
  },
  null
]
```

jadhavajay:
@gnarula - Can this PR updated to resolve the conflicts?

gnarula:
I was waiting for https://github.com/hyperledger/aries-cloudagent-python/pull/369 to be resolved.

Strictly speaking, we aren't adhering to DID 1.0 specification wrt the "authentication" block because ACA-Py doesn't. .NET ignores that altogether IIRC. Go adheres to the DID spec strictly.

I'd prefer we align with ACA-Py at the moment since it's more widely supported.

jadhavajay:
Yes, it makes sense to align with ACA-PY at this moment.

gnarula:
Resolved conflicts and added the suggestions mentioned above.

PR #16, opened 2020-01-27
ConnectionService: Use Connection Protocol

gnarula:
Switches to the [Aries Connection Protocol](https://github.com/hyperledger/aries-rfcs/tree/04ce2e2b0f956f2b3dcdad8298c6c039b31b57bc/features/0160-connection-protocol).

P.S. I'm aware of message type prefixes [being changed](https://github.com/hyperledger/aries-rfcs/blob/1d676fa8630a4b659ed5d175ea3b0043c6e83016/features/0348-transition-msg-type-to-https/README.md) but that can be accommodated in a future PR. Deadline for that is end of February.

### DIFF
--- a/src/lib/decorators.test.ts
+++ b/src/lib/decorators.test.ts
@@ -36,10 +36,10 @@ describe('decorators', () => {
     '~thread': { thid: 'thread1' },
     'connection~sig': {
       '@type': 'did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/signature/1.0/ed25519Sha512_single',
-      signature: 'FnVvO/NJqmDM9OiIdg3zN4yCZ7dowDjARymMSpO1ngX0f4ehPQzkweNdHwvInm9QfMhNoWgXz4esHpayuhVbDQ==',
+      signature: 'zOSmKNCHKqOJGDJ6OlfUXTPJiirEAXrFn1kPiFDZfvG5hNTBKhsSzqAvlg44apgWBu7O57vGWZsXBF2BWZ5JAw==',
       sig_data:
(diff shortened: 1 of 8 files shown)


src/lib/decorators.test.ts
```diff
@@ -59,7 +59,7 @@ describe('decorators', () => {
     const seed1 = '00000000000000000000000000000My1';
     const verkey = await indy.createKey(wh, { seed: seed1 });
 
-    const result = await sign(wh, message, 'connection', verkey);
+    const result = await sign(wh, message, 'connection', verkey, true);
```
jakubkoci: We should try to avoid flag parameters (`true/false`) see https://martinfowler.com/bliki/FlagArgument.html. In this case, it's also related only to the testing environment and it could confuse regular function consumers. I would suggest two options:

1. Factory function `sign(timestamp)(wh, message, 'connection', verkey)`
2. Extract it into separate module (let's call it `timestamp` until we find better name) and than mock that with `jest.mock('./timestamp');` in tests

I think 2. option is simpler right now and it keeps common API with its counterpart `verify`

src/lib/protocols/helpers.ts
```diff
@@ -9,7 +9,7 @@ export function createOutboundMessage(connection: Connection, payload: Message,
       payload,
       recipientKeys: invitation.recipientKeys,
       routingKeys: invitation.routingKeys,
-      senderVk: null,
+      senderVk: connection.verkey,
```
jakubkoci: 👍 thanks for the fix

src/lib/decorators.test.ts
jakubkoci: I briefly looked at [Jest docs](https://jestjs.io/docs/en/jest-object#jestmockmodulename-factory-options) and something like this should work:
```
jest.mock('./timestamp', () => {
  return jest.fn(() => Uint8Array.of(0, 0, 0, 0, 0, 0, 0, 0));
});
```

src/lib/decorators.ts
```diff
@@ -1,10 +1,17 @@
...
-
-export async function sign(wh: WalletHandle, message: Message, field: string, signer: Verkey): Promise<Message> {
+import timestamp from './timestamp';
+
+export async function sign(
+  wh: WalletHandle,
+  message: Message,
+  field: string,
+  signer: Verkey,
+  isTest: boolean = false
```
jakubkoci: We can remove `isTest`

PR #24, opened 2020-02-01
Add env to run agency with debug logs

jakubkoci:
@gnarula I was confused because the agency didn't log anything and didn't know what was the reason :) So I'm just adding debug logs as default and mentioning the option to turn-off logs in docs.

Signed-off-by: jakubkoci <jakub.koci@gmail.com>


gnarula:
Looks good to me!

PR #25, opened 2020-02-06
decorators: use base64url

gnarula:
Aries RFC 0234 requires signature decorators to encode fields
using base64url instead of base64

Refer https://github.com/hyperledger/aries-rfcs/blob/master/features/0234-signature-decorator/README.md

Signed-off-by: Gaurav Narula <gnarula94@gmail.com>

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
   "dependencies": {
     "@types/express": "^4.17.0",
     "await-poll": "^1.0.2",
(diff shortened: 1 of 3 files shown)


gnarula:
So turns out Connection Protocol expects the signerVk to *not be* base64URL encoded and that's what is going to be a part of AIP 1.0.0. Adding a commit that ensures that

PR #27, opened 2020-02-09
Configuraton: Allow providing wallet configuration/credentials

gnarula:


gnarula:
my bad, strange why the pre-push hook didn't catch it. Anyway, I've updated the commit now

PR #28, opened 2020-02-11
Add initial support for persistence

gnarula:
Default storage implementation uses Indy non_secrets API to persist stuff


src/lib/storage/BaseRecord.ts
```diff
@@ -0,0 +1,37 @@
+import uuid from 'uuid/v4';
+
+export enum RecordTypes {
```
jakubkoci: In TS is naming convention to name enums as singular see https://www.typescriptlang.org/docs/handbook/enums.html. Also, consider using `PascalCase`

src/lib/storage/BaseRecord.ts
```diff
@@ -0,0 +1,37 @@
+import uuid from 'uuid/v4';
```
jakubkoci: It's not used in this file.

src/lib/storage/BaseRecord.ts
```diff
@@ -0,0 +1,35 @@
...
+  tags: { [keys: string]: string };
+
+  // Required because Javascript doesn't allow accessing static types
+  // like instance.static_memeber
+  type: RecordType = RecordType.BaseRecord;
+  static type: RecordType = RecordType.BaseRecord;
+
+  constructor(id: string) {
+    this.id = id;
+    this.createdAt = new Date().getUTCMilliseconds();
```
jakubkoci: Is it correct Date function to use?  

From MDN docs: "The getUTCMilliseconds() method returns the milliseconds portion of the time object's value."

Maybe, the `Date.now()` is better. 
From MDN docs: "The Date.now() method returns the number of milliseconds elapsed since January 1, 1970 00:00:00 UTC."

src/lib/__tests__/agents.test.ts
```diff
@@ -87,10 +87,13 @@ describe('agents', () => {
     await aliceAgent.sendMessageToConnection(aliceConnections[0], message);
 
     const bobMessages = await poll(
-      () => {
-        console.log(`Getting Bob's connection messages...`);
-        const connections = bobAgent.getConnections();
-        return connections[0].messages;
+      async () => {
+        console.log(`Getting Bob's messages from Alice...`);
+        const messages = await bobAgent.basicMessageRepository.findByQuery({
```
jakubkoci: I guess this will be hidden behind some higher level method, like `agent.getMessageBy(...)`, right?

src/lib/storage/ConnectionRecord.ts
```diff
@@ -0,0 +1,38 @@
...
+import { InvitationDetails } from '../protocols/connections/domain/InvitationDetails';
+import { ConnectionState } from '../protocols/connections/domain/ConnectionState';
+import uuid from 'uuid/v4';
+
+export interface ConnectionStorageProps extends ConnectionProps {
+  id?: string;
+  tags: { [keys: string]: string };
+}
+
+export class ConnectionRecord extends BaseRecord implements ConnectionStorageProps {
```
jakubkoci: What if we try to use something like base `Entity` class and `props` attribute: https://github.com/stemmlerjs/white-label/blob/master/src/core/domain/Entity.ts

And use it in this way: https://github.com/stemmlerjs/white-label/blob/master/src/modules/vinyl/domain/album.ts

We would get rid off repeating lists of the same attributes in class and constructor.

I feel there is a lot extending and interfacing and not sure if it's necessary.

src/lib/storage/ConnectionRecord.ts
```diff
@@ -0,0 +1,38 @@
+import { BaseRecord, RecordType } from './BaseRecord';
+import { ConnectionProps } from '../protocols/connections/domain/Connection';
+import { DidDoc } from '../protocols/connections/domain/DidDoc';
+import { InvitationDetails } from '../protocols/connections/domain/InvitationDetails';
+import { ConnectionState } from '../protocols/connections/domain/ConnectionState';
```
jakubkoci: We probably don't want to depend on protocols from storage, do we?

src/lib/storage/repository.test.ts
```diff
@@ -0,0 +1,92 @@
...
+      await cr.find(record.id);
+    }).rejects;
+  });
+
+  test('it is able to query a record', async () => {
+    const record = await insertRecord();
+    const result = await cr.findByQuery({ myTag: 'foobar' });
+    expect(result.length).toBe(1);
+  });
+});
```
jakubkoci: It seems this test actually tests IndyStorage or ConnectionRepository. Maybe we can create some `TestRecrod` extending `BaseRecord` and rename it as `IndyStorageService.test.ts`

src/lib/storage/BaseRecord.ts
```diff
@@ -0,0 +1,35 @@
...
+export abstract class BaseRecord {
+  createdAt: number;
+  updatedAt?: number;
+  id: string;
+  tags: { [keys: string]: string };
+
+  // Required because Javascript doesn't allow accessing static types
+  // like instance.static_memeber
+  type: RecordType = RecordType.BaseRecord;
+  static type: RecordType = RecordType.BaseRecord;
```
jakubkoci: I'm sorry, but I still don't understand why we have both variants here? 🤔 😄

src/lib/storage/StorageService.ts
```diff
@@ -0,0 +1,15 @@
...
+
+  update(record: T): Promise<void>;
+
+  delete(record: T): Promise<void>;
+
+  find<T>(typeClass: { new (...args: any[]): T }, id: string, type: string): Promise<T>;
+
+  findAll<T>(typeClass: { new (...args: any[]): T }, type: string): Promise<T[]>;
+
+  findByQuery<T>(typeClass: { new (...args: any[]): T }, type: string, query: {}): Promise<T[]>;
```
jakubkoci: I've been thinking how to get rid of this `typeClass` and `type` attribute and use just OOP mechanics with static factory method for example, but I couldn't figure anything out 🤷‍♂

src/lib/storage/BaseRecord.ts
gnarula: That's because typescript converts a static member inside a class to be a property of the class.

```typescript
class Foo {
  static bar: string = "FOOBAR";
}
```

ends up being transpiled to:

```javascript
"use strict";
class Foo {
}
Foo.bar = "FOOBAR";
```

Unfortunately, we cannot access `bar` from an instance of `Foo` but only Foo itself.

I used the static member to allow passing a `typeof Class` to the constructor of Repository and allow initializations like `const repo = new Repository<ConnectionRepository>(ConnectionRecord, indyStorageInstance);`

At the same time, I need it to be accessible from an instance since I destructure it in `IndyStorageService::{save,update,delete}`

src/lib/storage/StorageService.ts
gnarula: We can probably get rid of type there because typeClass is actually having a static field of type anyway ;)

`typeClass: { new (...args: any[]): T; type: string }`

The end user however wont deal with it directly anyway since they're going to access everything through the repository and the signature there for find* operations doesn't require any type information. It's supplied while invoking the constructor...

src/lib/storage/repository.test.ts
gnarula: +1, will do

src/lib/__tests__/agents.test.ts
gnarula: uhm, it's just not the messages that the user would want access to. I might want access to the connections that were persisted for instance. Do you suggest we have getXBy for all record types?

src/lib/storage/BaseRecord.ts
gnarula: my bad, I'll fix it

src/lib/storage/ConnectionRecord.ts
gnarula: > We would get rid off repeating lists of the same attributes in class and constructor.

If that's the only objective then `Object.assign(this, props)` should suffice but then we'd have typescript complain since Object.assign isn't type safe... Maybe we can use some suggestion from [here](https://stackoverflow.com/a/55583832)

jakubkoci: I don't like `Object.assign(...)` because that type problem you mentioned, neither the solution in given link 😄 We don't have to address this in this PR ;)

src/lib/storage/BaseRecord.ts
jakubkoci: Ok, I feel like it says us something, but we don't have to address this in this PR ;)

jadhavajay:
@gnarula will need to resolve conflicts for this as well.

gnarula:
Fixed conflicts. Should be good to merge now

PR #41, opened 2020-05-07
Add connections persistence

jakubkoci:
I'm adding persistence to connections. I don't see it as a final solution for 2 reasons, but let's put it there for the reference at least. These reasons are:

1. It still preserves the in-memory connections array. The reason is that we use connection object as emitter and when retrieving a connection from the database we're creating a new object and then a new instance of an event emitter, therefore there could be two instances of same connection with different listeners. So, I load all connections into an array (memory) and use only in-memory instances, but updating the database for every change. 

2. I feel there is to much duplication between `Connection`, `ConnectionRecord`,  `ConnectionProps` and `ConnectionStorageProps` objects. We won't probably have only one class, interface, but it would be good to minimize it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "prettier": "prettier --ignore-path .gitignore '**/*.+(js|json|ts)'",
     "format": "yarn prettier --write",
     "check-format": "yarn prettier --list-different",
-    "test": "jest",
+    "test": "jest --verbose",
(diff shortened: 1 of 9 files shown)


jakubkoci:
@TimoGlastra FYI ;)

jadhavajay:
Looks like the DCO is missing.

jakubkoci:
> Looks like the DCO is missing.

Ah, forget as always. Fixed now.

Issue #40, opened 2020-05-02
Mediation for Edge Agents in Aries Framework JavaScript


TimoGlastra:
Current thoughts on implementing mediation for Edge Agents into Aries Framework JavaScript. I think in the beginning especially the Edge Agent role would be nice to implement, but eventually implementing the Mediator role would also be nice. This approach allows for interoperable mediaton with other agents that support the protocols mentioned in this document. Still some unresolved questions at the end. Would appreciate some other thoughts on this.

## Creating A Connection with the Mediator

The Mediator creates a [`Connection Invitation`](https://github.com/hyperledger/aries-rfcs/blob/master/features/0160-connection-protocol/README.md#0-invitation-to-connect) and sends it to the Edge Agent. If the Edge Agent wishes to connect to the Mediator it should respond with a [`Connection Request`](https://github.com/hyperledger/aries-rfcs/blob/master/features/0160-connection-protocol/README.md#1-connection-request). Because the Edge Agent has no endpoint to include in the `Connection Request`, an endpoint URI of `didcomm:transport/queue` will indicate to the Mediator that the Edge Agent does not have a direct endpoint and messages should be queued.

The `Connection Request` sent by the Edge Agent should include the [Transport Return Route Decorator](https://github.com/hyperledger/aries-rfcs/blob/master/features/0092-transport-return-route/README.md) to allow the Mediator to directly respond with a `Connection Response` message.

The Edge Agent and Mediator now have a connection. The Mediator will queue all messages it wants to send to the Edge Agent due to the `didcomm:transport/queue` endpoint.

## Requesting Mediation

The Mediator and Edge Agent now have a connection, but it is not clear yet whether the Mediator would like to mediate for the Edge Agent.

Using the [Mediator Coordination Protocol](https://github.com/hyperledger/aries-rfcs/tree/master/features/0211-route-coordination) the Edge Agent can send a [`Mediation Request`](https://github.com/hyperledger/aries-rfcs/tree/master/features/0211-route-coordination#mediation-request) to the Mediator. If the Mediator wishes to accept it will respond with a [`Mediation Grant`](https://github.com/hyperledger/aries-rfcs/tree/master/features/0211-route-coordination#mediation-request) message. The `Mediation Grant` will include the `endpoint` and `routing_keys` properties. The Edge Agent should store these values as they will be used when creating new connections with other agents.

## Creating A Connection with Other Agents

[...]

PR #45, opened 2020-05-15
Add connection with mediator

jakubkoci:
Implementation based on notes in #40 without Requesting Mediation part of the protocol
- Add provision method inspired by https://github.com/hyperledger/aries-framework-dotnet/blob/302d70bfb2/src/Hyperledger.Aries.Routing.Edge/EdgeProvisioningService.cs
- Add pickup protocol


src/samples/__tests__/e2e.test.ts
```diff
@@ -145,27 +145,20 @@ class PollingInboundTransporter implements InboundTransporter {
...
-    console.log('agentConnectionAtAgency\n', agentConnectionAtAgency);
-
     const { verkey: agencyVerkey } = JSON.parse(await get(`${agencyUrl}/`));
-    agent.establishInbound(agencyVerkey, agentConnectionAtAgency);
+    await agent.provision({ verkey: agencyVerkey, invitationUrl: agencyInvitationUrl });
+    this.pollDownloadMessages(agent);
   }
 
-  pollMessages(agent: Agent, agencyUrl: string, verkey: Verkey) {
+  pollDownloadMessages(agent: Agent) {
```
matt-raffel-kiva, line 153: I am going to add this comment realizing I may not be fully understanding what is intended (since Im just getting back into aries).  Would it be better to keep pollMessages and create a new method pollDownloadMessages?

jakubkoci: Thanks for your comment @matt-raffel-kiva It made me think twice, but the previous implementation of `pollMessages` was my custom way how to achieve the required behavior unrelated to any Aries RFCS, so throwing it away is the right choice here. This newer implementation does the same but relies more on Aries RFCS mentioned in #40

PR #46, opened 2020-05-16
add supported message types to handlers

TimoGlastra:
This PR adds a `supportedMessageTypes` to the `Handler` interface that every handler must implement. Inspired by the .NET implementation.

Now the handler defines which message types it supports instead of declaring this in the agent. I think it fits better in the handler as the handler is coupled to which message types it support.

This also makes the agent class a bit smaller, and will help in the future with pluggable protocol handlers.

### DIFF
--- a/src/lib/agent/Agent.ts
+++ b/src/lib/agent/Agent.ts
@@ -1,14 +1,9 @@
 import logger from '../logger';
 import { InitConfig } from '../types';
-import { encodeInvitationToUrl, decodeInvitationFromUrl } from '../helpers';
+import { decodeInvitationFromUrl } from '../helpers';
 import { IndyWallet } from '../wallet/IndyWallet';
 import { Connection } from '../protocols/connections/domain/Connection';
 import { ConnectionService } from '../protocols/connections/ConnectionService';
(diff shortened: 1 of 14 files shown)


TimoGlastra:
Yeah we should think about it. I think it would be best to see why this would happen and implement according to this reason. 

I can image two scenarios:

1. Two of the same message types sometimes handled by one handler, sometimes by another. That would need custom logic before passing the message to the handler.
1. Override handler for already registered message type. In the handler (as in the example below) we could change the `supportedMessageTypes` to `handledMessageTypes` and if another handler is registered that supports a message that is already supported by another handler we could (during runtime) remove the message from the first instance.

Let's see how it works out. We should at least reverse the array for now so adding a new handler that supports a message type that is already supported gets picked over the first registered one. Now (see example) the default will be used over the custom. We could add some custom logic to allow the framework user to define if a certain message should always be handled by a certain handler. 

```ts
const handler = [
    DefaultBasicMessageHandler(),
    CustomBasicMessageHandler()
]
```

TimoGlastra:
The main reason I made this change is to allow more type safe handlers as types could be automatically inferred from the `supportedMessageTypes`. I made a little demo of what I mean that I will show you this friday.

PR #48, opened 2020-05-19
Creates docker images

matt-raffel-kiva:
This creates alice and bob agents in separate docker images


Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+ARG libindy_ver=1.14.0
+# Recommended way to get setup with libindy: https://github.com/hyperledger/indy-sdk#ubuntu-based-distributions-ubuntu-1604
+ARG indy_stream=stable
+RUN apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 68DB5E88 && \
+    add-apt-repository "deb https://repo.sovrin.org/sdk/deb xenial $indy_stream" && \
+    apt-get update && \
+    apt-get install -y libindy=${libindy_ver}
+
+# Setup nodejs
+RUN curl -sL https://deb.nodesource.com/setup_8.x | bash && \
```
jakubkoci, line 23: What if we use a newer version of Node, 12 at least?

Dockerfile
```diff
@@ -0,0 +1,37 @@
+# Note that the indy-sdk requires ubuntu 16 and some custom dependencies so we can't use node:carbon-apline like the others
```
jakubkoci, line 1: Is Ubuntu 16 really still required version? I spot Docker file in indy-sdk using Ubuntu 18 https://github.com/hyperledger/indy-sdk/blob/master/libindy/ci/ubuntu18.dockerfile

Dockerfile
matt-raffel-kiva: I wasn't able to upgrade to new version of node: 
```
Error: The module '/www/node_modules/indy-sdk/build/Release/indynodejs.node'
was compiled against a different Node.js version using
NODE_MODULE_VERSION 57. This version of Node.js requires
NODE_MODULE_VERSION 72. Please try re-compiling or re-installing
the module (for instance, using `npm rebuild` or `npm install`).
```
So this is why I kept it at v8.

Dockerfile
matt-raffel-kiva: updated to 18.04

Dockerfile
jakubkoci: Oh, I see

PR #52, opened 2020-05-30
Replace Connection object with Connection record

jakubkoci:
When I added persistence for connections I kept both in-memory and persisted versions of connections collection, but it didn't work correctly, so I decide to the in-memory collection of `Connection` objects, remove the object and use only `ConnectionRecord` and `connectionRepository`.

`Connection` object was also `EventEmitter`, so I move this behavior to the service layer, into the `ConnectionService`.

I also added `closeAndDeleteWallet ` method to the `Agent` class for cleaning up after tests, but I think it could be useful also for other purposes.

This PR contains also some tests for `ConnectionService`. Take it as a starting point and demonstration of a possible testing approach.


src/lib/storage/ConnectionRecord.ts
```diff
@@ -34,4 +50,23 @@ export class ConnectionRecord extends BaseRecord implements ConnectionStoragePro
...
+    }
+    return this.didDoc.service[0].recipientKeys[0];
+  }
+
+  get theirKey() {
+    if (!this.theirDidDoc) {
+      return null;
+    }
+    return this.theirDidDoc.service[0].recipientKeys[0];
+  }
```
TimoGlastra, line 66: Currently all connections use the first key with the first service so it should work fine, but i think it would be better to look for the `IndyAgent` service.

src/lib/agent/Agent.ts
```diff
@@ -193,6 +206,27 @@ export class Agent {
...
   }
 
+  async returnWhenIsConnected(verkey: Verkey) {
+    const connectionRecord = await poll(
+      () => this.findConnectionByVerkey(verkey),
+      (c: ConnectionRecord) => c.state !== ConnectionState.COMPLETE,
+      100
+    );
+    return connectionRecord;
+  }
```
TimoGlastra, line 216: Maybe we could use the event emitter to subscribe for this event? I think that would be a clear approach

src/lib/protocols/basicmessage/BasicMessageService.ts
```diff
@@ -36,7 +37,7 @@ class BasicMessageService {
       tags: { from: connection.theirDid || '', to: connection.did || '' },
     });
     await this.basicMessageRepository.save(basicMessageRecord);
-    connection.emit('basicMessageReceived', message);
+    this.emit('newMessage', { verkey: connection.verkey, message });
```
TimoGlastra, line 40: I think it would be better to make this a constant or enum, so subscribing is not prone to typos. This way it also becomes easier to add typing to events (i.e. subscribe to 'newMessage' type always return a verkey and message when called)

jakubkoci: Good point. I thought about it. I wasn't sure if it's necessary to solve it now. There is a question of a naming convention of such event types:
- Tense: `stateChange`, `stateUpdate` vs. `stateChanged`, `stateUpdated`, I picked the latter, according to the CQRS naming convention, but I see a lot of examples of the first option in the JS world.
- Case: `stateChanged`, `state-changed`, `StateChanged`, `statechange` ordered by my preferences :)

src/lib/agent/Agent.ts
jakubkoci: I started with this approach, but it occurred to be more complicated than I expected.
- We would need to remove a listener afterward, even if it's there for the short life of tests, it could direct someone to wrong solution if we wouldn't do that.
- It mixes async and callbacks together.
- I realized that consuming API via event emitter should be only optional and it's good to have tests demonstrating that. I'm open to adding other tests demonstrating the use of event emitter, but I would also consider showing that in an example instead.

src/lib/storage/ConnectionRecord.ts
jakubkoci: I agree it's really not a robust solution :) It's implemented in the simplest possible way. I suggest to do it in other PR because it's not part of this change so much. Contribution welcome here ;)

I'm not sure if such logic should be part of the `ConnectionRecord`, maybe it'll be more complicated at the end. To be honest, I don't know what's the 100 % correct implementation.

TimoGlastra: Not sure either. Maybe we can create a diddoc class as implemented here: https://github.com/decentralized-identity/did-common-typescript/blob/master/lib/DidDocument.ts

Getting my key vs their key would then be something like. 
```ts
const myKey = connection.didDoc.getServiceByType('IndyAgent').recipientKeys[0]
const theirKey = connection.theirDidDoc.getServiceByType('IndyAgent').recipientKeys[0]
```

But i agree this should be addressed in a separate PR.

src/lib/protocols/basicmessage/BasicMessageService.ts
TimoGlastra: Nice. I would prefer `stateChanged` and `stateUpdated` over `stateChange` and `stateUpdate`. And i would prefer camelCase for naming (`stateChanged`). Not based on any valuable arguments, just preference :)

jakubkoci: Then we're on the same page :)

src/lib/storage/ConnectionRecord.ts
jakubkoci: Cool, I like that idea 👍 I didn't know about that repo, thanks for pointing that out. We can take it as an inspiration, or maybe, add the whole package as a dependency. Although, I'm not sure about the work needed for the persistence of such an object with our current storage mechanism.

jakubkoci: Created #53

PR #61, opened 2020-07-15
add extra troubleshooting help

TimoGlastra:
Extracted the troubleshooting to a separate file and added the steps to resolve the issues both Karim and I faced while installing AFJ / libindy on macOS.


TimoGlastra:
@karimStekelenburg

PR #71, opened 2020-07-25
replace inbound message with message context

TimoGlastra:
adds a message context class that is populated after a message is received. If found, it attaches the connection record associated with the message to the context.

I started to work on replacing the verkey with connection_id (which we apparently already have?, but is not used everywhere) but felt this could help avoid repetitive patterns of each handler retrieving the connection associated with the message.

### DIFF
--- a/src/lib/agent/MessageReceiver.ts
+++ b/src/lib/agent/MessageReceiver.ts
@@ -41,15 +41,16 @@ class MessageReceiver {
     logger.logJson('inboundMessage', unpackedMessage);
 
     const message = await this.transformMessage(unpackedMessage);
-    let senderKey = unpackedMessage.sender_verkey;
+    const senderKey = unpackedMessage.sender_verkey;
     let connection = undefined;
     if (senderKey && unpackedMessage.recipient_verkey) {
+      // TODO: only attach if theirKey is present. Otherwise a connection that may not be complete, validated or correct will
+      // be attached to the message context. See #76
       connection = (await this.connectionService.findByVerkey(unpackedMessage.recipient_verkey)) || undefined;
 
-      // we validate the sender key if it is the one we have the connection with
-      // otherwise everyone could send message for our key, and we would just accept
-      // it as if it was send by the connection.
-      // TODO: does this correctly work for the connection process? Keys can be swapped during the protocol
+      // We check whether the sender key is the same as the key we have stored in the connection
+      // otherwise everyone could send messages to our key and we would just accept
+      // it as if it was send by the key of the connection.
       if (connection && connection.theirKey != null && connection.theirKey != senderKey) {
         throw new Error(
           `Inbound message 'sender_key' ${senderKey} is different from connection.theirKey ${connection.theirKey}`
@@ -57,7 +58,7 @@ class MessageReceiver {
       }
     }
 
-    const messageContext = new MessageContext(message, {
+    const messageContext = new InboundMessageContext(message, {
       connection,
       senderVerkey: senderKey,
       recipientVerkey: unpackedMessage.recipient_verkey,
--- a/src/lib/agent/MessageReceiver.ts
+++ b/src/lib/agent/MessageReceiver.ts
@@ -4,7 +4,7 @@ import { Dispatcher } from './Dispatcher';
 import { EnvelopeService } from './EnvelopeService';
 import { UnpackedMessage } from '../types';
 import { MessageType } from '../protocols/routing/messages';
-import { MessageContext } from './models/MessageContext';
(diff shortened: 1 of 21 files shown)


src/lib/protocols/connections/ConnectionService.ts
```diff
@@ -128,18 +129,15 @@ class ConnectionService extends EventEmitter {
...
+  async acceptAck(messageContext: MessageContext<AckMessage>) {
+    const connection = messageContext.connection;
 
-    if (!connectionRecord) {
-      throw new Error(`Connection for verkey ${recipient_verkey} not found!`);
+    if (!connection) {
+      throw new Error(`Connection for ${messageContext.recipientVerkey} not found!`);
     }
 
-    validateSenderKey(connectionRecord, sender_verkey);
```
TimoGlastra, line 139: The LGTM warning is from the `validateSenderKey` that isn't used anymore. I now do this check at the start after a message is received.

I didn't remove the function however because it contains comments with questions that may still need to be resolved?

src/lib/agent/MessageReceiver.ts
```diff
@@ -4,37 +4,90 @@ import { Dispatcher } from './Dispatcher';
...
+    let senderKey = unpackedMessage.sender_verkey;
+    let connection = undefined;
+    if (senderKey && unpackedMessage.recipient_verkey) {
+      connection = (await this.connectionService.findByVerkey(unpackedMessage.recipient_verkey)) || undefined;
+
+      // we validate the sender key if it is the one we have the connection with
+      // otherwise everyone could send message for our key, and we would just accept
+      // it as if it was send by the connection.
+      // TODO: does this correctly work for the connection process? Keys can be swapped during the protocol
+      if (connection && connection.theirKey != null && connection.theirKey != senderKey) {
```
jakubkoci, line 53: "Keys can be swapped during the protocol" - Do you mean during the connection protocol, after the Accept Invitation step?

src/lib/protocols/connections/ConnectionService.ts
jakubkoci: You can remove it. I still have these questions unanswered, but we the problem has been moved to `MessageReceiver` to line with `if (connection && connection.theirKey != null && connection.theirKey != senderKey) {`. We can discuss it and eventually write some comment there.

src/lib/agent/MessageReceiver.ts
jakubkoci: If so, we should allow the change just for that case, otherwise, it shouldn't be allowed.

When making a connection, I create some `did`, `verkey` and send it in invitation to other party. I get `theirDid`, `theirVerkey` in connection request and respond with `myDid` and `myVerkey` in connection response.

Then, when I get a message. The `unpack` operation tries to find  `recipient_key` between stored keys in my wallet and decrypt the message with it. But I still need to check if `sender_verkey`, if it’s authCrypt communication, is equal to `connection.theirVerkey` I stored in DIDExchange process. It `sender_verkey` is different, someone who knows `myVerkey` in this DID connection could send me a message signed with his key.

src/lib/protocols/connections/ConnectionService.ts
TimoGlastra: Allright, removed in next commit.

I added it to the agenda for tomorrow

src/lib/agent/models/MessageContext.ts
```diff
@@ -0,0 +1,29 @@
...
+    if (context.connection) {
+      this.connection = context.connection;
+      // TODO: which senderkey should we prioritize
+      // Or should we throw an error when they don't match?
+      this.senderVerkey = context.connection.theirKey || context.senderVerkey || undefined;
+    } else if (context.senderVerkey) {
+      this.senderVerkey = context.senderVerkey;
+    }
+  }
+}
```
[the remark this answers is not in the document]
TimoGlastra: Yeah that is a good idea. It can serve as a general purpose class to add extra information to a message.

For the notation: let's discuss on the call

src/lib/agent/MessageReceiver.ts
TimoGlastra: Yes I meant during the connection protocol. Will update the comments to make it a bit more clear.

The question on my side if this will work with the current checks:

```
alice: invitation with public did
bob: request with pairwise did
alice: response with pairwise did
```

when bob receives the response with the pairwise did the sender key is different than the one from the invitation. It is then linked together because the attached did document is signed with the same key as the invitation. After verifying this we should update theirkey.

Will look into it this weekend.

PR #73, opened 2020-07-25
Replace verkey with connection id

### LINKED ISSUE #70: Replace verkey as connection identifier

TimoGlastra:
Seems the connection already had an id. Verkey wasn't used that much, so pretty small change after all.

Fixes #70

### DIFF
--- a/src/lib/protocols/connections/ConnectionService.ts
+++ b/src/lib/protocols/connections/ConnectionService.ts
@@ -177,6 +177,15 @@ class ConnectionService extends EventEmitter {
     return this.connectionRepository.findAll();
   }
 
+  async getById(connectionId: string): Promise<ConnectionRecord | null> {
+    try {
+      return this.connectionRepository.find(connectionId);
+    } catch {
+      // connection not found.
+      return null;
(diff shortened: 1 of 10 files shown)


src/lib/protocols/connections/ConnectionService.ts
```diff
@@ -182,6 +182,15 @@ class ConnectionService extends EventEmitter {
     return this.connectionRepository.findAll();
   }
 
+  async getById(connectionId: string): Promise<ConnectionRecord | null> {
```
jakubkoci, line 185: What if we have a convention to name every such method that can return null `findByXXX`? So, it will be probably just `getAll()` and similar `getConnections()`, `getCredentials()` and `findBy` for other. I can imagine to have just `find` for the `findById` version, because you can read this from context (method params would have an `id` parameter)

TimoGlastra:
I changed `getById` to `find` for now and created a corresponding issue to revamp the handling when records are not found (throwing errors vs returning null): https://github.com/hyperledger/aries-framework-javascript/issues/78

PR #82, opened 2020-08-13
Add credential schema and definition registration to ledger

jakubkoci:


src/lib/agent/LedgerService.ts
```diff
@@ -0,0 +1,181 @@
...
+  authorAgreement?: AuthorAgreement;
+
+  constructor(wallet: Wallet, indy: Indy) {
+    this.wallet = wallet;
+    this.indy = indy;
+  }
+
+  async connect(poolName: string, poolConfig: PoolConfig) {
+    try {
+      logger.log(__dirname);
```
TimoGlastra, line 17: seems unnecessary?

src/lib/agent/LedgerService.ts
```diff
@@ -0,0 +1,181 @@
...
+
+      logger.log('Creating pool config');
```
TimoGlastra, line 19: Maybe we can add some more details to log messages?  In this case we could add `poolName`.

src/lib/agent/LedgerService.ts
```diff
@@ -0,0 +1,181 @@
...
+      await this.indy.createPoolLedgerConfig(poolName, poolConfig);
+    } catch (e) {
+      logger.log('PoolLedgerConfigAlreadyExists');
```
TimoGlastra, line 22: This should be below the line that checks whether it is indeed a `PoolLedgerConfigAlreadyExistsError`

src/lib/agent/LedgerService.ts
```diff
@@ -0,0 +1,181 @@
...
+    const [schemaId, schema] = await this.indy.issuerCreateSchema(myDid, name, version, attributes);
+    logger.log('schemaId, schema', schemaId, schema);
+
+    const request = await this.indy.buildSchemaRequest(myDid, schema);
+    logger.log('request', request);
+
+    const requestWithTaa = await this.appendTaa(myDid, request);
+
+    const signedRequest = await this.wallet.signRequest(myDid, requestWithTaa);
+    logger.log('signedRequest', signedRequest);
```
TimoGlastra, line 65: Here we could add more detail to the log messages.

Just an example, would not use this directly:

```suggestion
    logger.log(`registering schema with schemaId: ${schemaId}`, schema);

    const request = await this.indy.buildSchemaRequest(myDid, schema);
    logger.log('created create schema request', request);

    const requestWithTaa = await this.appendTaa(myDid, request);

    const signedRequest = await this.wallet.signRequest(myDid, requestWithTaa);
    logger.log('signed create schema request', signedRequest);
```

src/lib/agent/LedgerService.ts
```diff
@@ -0,0 +1,181 @@
...
+    logger.log('response', response);
+
+    return [schemaId, schema];
+  }
+
+  async getSchema(myDid: Did, schemaId: SchemaId) {
+    if (!this.poolHandle) {
+      throw new Error('Pool has not been initialized.');
+    }
+    const request = await this.indy.buildGetSchemaRequest(myDid, schemaId);
```
TimoGlastra, line 77: Looking at .NET code we could pass `null` for the `myDid` variable. You don't need a public DID to read from the ledger. This way all read operations don't need a did passed in

src/lib/agent/LedgerService.ts
```diff
@@ -0,0 +1,181 @@
...
+    const response = await this.indy.submitRequest(this.poolHandle, request);
+    logger.log('response', response);
+
+    const result = await this.indy.parseGetCredDefResponse(response);
+    logger.log('Credential Definition from ledger: ', result);
+
+    return result;
+  }
+
+  private async appendTaa(myDid: Did, request: LedgerRequest) {
```
TimoGlastra, line 126: Maybe I don't understand it correctly, but do we need to do this for every request? I thought it was an accept once, then make requests as you like kinda agreement.

src/lib/agent/LedgerService.ts
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    logger.log('Setting protocol version');
+    await this.indy.setProtocolVersion(2);
+
+    logger.log('Opening pool');
+    this.poolHandle = await this.indy.openPoolLedger(poolName);
+  }
+
+  async getPublicDid(myDid: Did) {
```
TimoGlastra, line 35: As this is inside the ledger service I think we should just call it `getDid`. Every did we retrieve from the ledger is public.

TimoGlastra: We should at least change `myDid` to `did` as I can retrieve DIDs that are not mine.

src/lib/wallet/IndyWallet.ts
```diff
@@ -30,34 +31,28 @@ export class IndyWallet implements Wallet {
...
-        const [_, verkey] = await this.createDid({ did, seed });
-        didInfo = {
-          did,
-          verkey,
-        };
-      } else {
-        throw error;
-      }
-    }
+  async initAgentDid(did: Did, seed: string) {
```
TimoGlastra, line 34: A DID is generated from a seed right? so passing DID shouldn't be necessary

TimoGlastra, line 34: From what I understand from the previous call the agent did is used for the mediator? 

I think this doesn't necessarily needs a separate function / special handling. We create a pairwise DID for the connection with the mediator just like any connection, so I don't see the benefit of using this over the already present function.

Maybe I'm missing something?

src/lib/wallet/IndyWallet.ts
```diff
@@ -68,6 +63,20 @@ export class IndyWallet implements Wallet {
     return this.indy.createAndStoreMyDid(this.wh, didConfig || {});
   }
 
+  async createCredDef(
```
TimoGlastra, line 66: This may be really nit-picky but I think we should either name it `createCredentialDef()` as indy does with `issuerCreateAndStoreCredentialDef` or use the full name `createCredentialDefinition`

src/lib/agent/LedgerService.ts
```diff
@@ -0,0 +1,181 @@
...
+    const response = await this.indy.submitRequest(this.poolHandle, request);
+    logger.log('response', response);
+
+    const result = await this.indy.parseGetSchemaResponse(response);
+    logger.log('Credential Schema from ledger: ', result);
+
+    return result;
+  }
+
+  async registerDefinition(myDid: Did, credentialDefinitionTemplate: CredDefTemplate): Promise<[CredDefId, CredDef]> {
```
TimoGlastra, line 89: `registerCredentialDefinition`?

src/lib/wallet/IndyWallet.ts
TimoGlastra, line 66: Also, why is this included in the wallet? I think we could do it in the schema service. Seems like it doesn't belong here

src/lib/wallet/IndyWallet.ts
jakubkoci: Yes, that's a good point, but I think it could cause changes outside the scope of this PR so I would rather do it in a different one.

src/lib/wallet/IndyWallet.ts
jakubkoci: I have the same problem with it, but Indy function inside has this signature `indy.issuerCreateAndStoreCredentialDef(this.wh, issuerDid, schema, tag, signatureType, config)` which requires wallet handle (`wh`) and I didn't want to expose it directly. Maybe, what we could do is to add `getWalletHandle` method.

src/lib/agent/LedgerService.ts
jakubkoci: I would still go with `getPublicDid`, there is a lot of Dids around when you're working with Indy/Aries. I agree with renaming `myDid` to `did` 👍

src/lib/agent/LedgerService.ts
jakubkoci: Yes, it seems we need to add it to every request. VCX also mentions it here:
https://github.com/AbsaOSS/libvcx/blob/c3de77cb12da134f23b3434c534aebeb43119ea9/libvcx/src/api/vcx.rs#L419

src/lib/agent/LedgerService.ts
TimoGlastra: I'm still in favour of naming it `getDid` as we already have `getPublicDid` on the agent, which returns our public did. 

For example this part from the tests where we call `getPublicDid` two times. One time on agent itself, one time in ledger service. Both do different things. I find it a bit confusing.

```
  test('get public DID from ledger', async () => {
    const { did } = faberAgent.getPublicDid();

    if (!did) {
      throw new Error('Agent does not have publid did.');
    }

    const result = await faberAgent.ledger.getPublicDid(did);
    expect(result).toEqual({
      did: faberAgentPublicDid.did,
      verkey: faberAgentPublicDid.verkey,
      role: '101',
    });
  });
```

But having mentioned my preference, I rest my case. If this is what we're discussing about we're definitely on the right track ;)

src/lib/agent/Agent.ts
```diff
@@ -98,12 +103,20 @@ export class Agent {
     const { publicDid, publicDidSeed } = this.agentConfig;
     if (publicDid && publicDidSeed) {
       // If an agent has publicDid it will be used as routing key.
-      this.wallet.initPublicDid(publicDid, publicDidSeed);
+      this.wallet.initAgentDid(publicDid, publicDidSeed);
     }
```
TimoGlastra, line 107: I just noticed that `initAgentDid` is using `publicDid` from the config. Why is it using `publicDid` and not `agentDid` or something?

src/lib/agent/LedgerService.ts
jakubkoci: Removed

src/lib/agent/LedgerService.ts
jakubkoci: Updated

src/lib/agent/LedgerService.ts
jakubkoci: Updated

src/lib/agent/LedgerService.ts
jakubkoci: Replaced with `null`

jakubkoci:
I've updated some issues. These are still remaining:
- initAgentDid 
- issuerCreateAndStoreCredentialDef
- Setup CI variable

src/lib/wallet/IndyWallet.ts
jakubkoci: I would keep it as it is right now. I don't like either of the solutions. Maybe, when we'll be implementing integration with other ledgers we would know more how to approach it correctly

src/lib/agent/Agent.ts
jakubkoci: I gave it another look and you're right we can use just `publicDid` 👍

src/lib/wallet/IndyWallet.ts
jakubkoci: I got rid of `Wallet`'s `initAgentDid` and `Agent`'s `initPublicDid` method. Now, the `initPublicDid` is called inside `Agent.init()` as it was before, I just refactored method signature to be more benevolent and take `DidInfo`.

jakubkoci: Updated.

jakubkoci:
I think we're ready for merge. If it's not possible to merge with failing hyperledger.aries-framework-javascript Azure build, I will also remove it.

TimoGlastra (approved):

TimoGlastra:
Looks like the azure pipeline is required to pass. However this is a setting. Do you have permission to turn the requirement for azure off, and github actions on?

jakubkoci:
> Looks like the azure pipeline is required to pass. However this is a setting. Do you have permission to turn the requirement for azure off, and github actions on?

I don't have permission.

@ryjones Could you please remove Azure Pipeline check for the project as we moved to Github Actions? All tests already run successfully there.

PR #85, opened 2020-08-14
Use github actions instead of azure

TimoGlastra:
@jakubkoci The commit history is a bit weird. The changes I made require the 'docker improvements' commit, which you didn't have in your 'credentials-schema' branch yet. Basically it's the last two commits that matter. 

Maybe you can add them to your branch after rebasing current master.


TimoGlastra:
And it seems like actions aren't automatically enabled

jakubkoci:
What if we add GA into the current master first, make it work with current tests, and then I would rebase PR #82? And after that we would remove current CI setup? 

It would be in smaller and safer steps.

TimoGlastra:
Closing in favour of #86

PR #89, opened 2020-08-23
Local indy ledger

TimoGlastra:
First CI ledger setup used VON-Network. This works great and is easy to setup, but can't be used outside of docker. For CI this is okay, however when developing locally it can become a bit tedious to build the framework into a docker image every time. 

This setup works both inside and outside of docker. CI tests are still run inside docker to make the CI setup easier.

Also updated readme with ledger setup and some general improvements.


src/lib/__tests__/ledger.test.ts
```diff
@@ -62,11 +59,30 @@ describe('ledger', () => {
...
+    //
+    // The verkey is 32 bytes, and by default in Indy the DID is chosen as the first 16 bytes of that key, before base58 encoding.
+    // The abbreviated verkey just replaces the first 16 bytes with ~ when it matches the DID.
+    //
+    // When full verkey is used to register on ledger, this is stored on ledger and also returned as full verkey
+    // same applies to abbreviated verkey. If it is used it will be stored abbreviated and returned abbreviated.
+    let verkey = faberAgentPublicDid.verkey as string;
+    if (isFullVerkey(verkey) && isAbbreviatedVerkey(result.verkey)) {
+      verkey = await indy.abbreviateVerkey(faberAgentPublicDid.did as string, verkey);
+    }
```
jakubkoci, line 77: What if we declared verkey type in `isFullVerkey` and `isAbbreviatedVerkey` function signature as `Verkey | string` would it help to get rid off these "overtyping" `as string`?

src/lib/__tests__/ledger.test.ts
```diff
@@ -62,11 +59,30 @@ describe('ledger', () => {
...
+    //
+    // .NET also abbreviates verkey before sending to ledger:
+    // https://github.com/hyperledger/aries-framework-dotnet/blob/f90eaf9db8548f6fc831abea917e906201755763/src/Hyperledger.Aries/Ledger/DefaultLedgerService.cs#L139-L147
+    //
+    //
+    // The verkey is 32 bytes, and by default in Indy the DID is chosen as the first 16 bytes of that key, before base58 encoding.
+    // The abbreviated verkey just replaces the first 16 bytes with ~ when it matches the DID.
+    //
+    // When full verkey is used to register on ledger, this is stored on ledger and also returned as full verkey
+    // same applies to abbreviated verkey. If it is used it will be stored abbreviated and returned abbreviated.
```
jakubkoci, line 73: I would maybe add something like this (with some correction of my English if needed, of course 😄 ) and move these two paragprahs above somewhere else, perhaps to `LedgerService` or `utils/did` 🤔 
```
// Agent’s public did stored locally in Indy wallet and created from public did seed during 
// its initialization always returns full verkey. Therefore we need to align that here.
```

src/lib/__tests__/ledger.test.ts
```diff
@@ -120,7 +136,9 @@ describe('ledger', () => {
         }),
       })
     );
-  });
+    // Submitting new credential definition can take a while
+    // and somtimes exceeds the default 15000ms
+  }, 30000);
```
jakubkoci, line 141: Ouch, that's weird 🤔 I would understand that if there is communication with the outside world, but if it's running locally? But ok 🤷‍♂️

jakubkoci (changes requested):
This PR has been here for a while so even if I've added some suggestions I'm approving this

src/lib/__tests__/ledger.test.ts
TimoGlastra: I'm not sure why this is. But this is the same when using ACA-Py or framework .NET. So I think the operation requires some crypto that takes a while on the ledger or in the indy sdk.

src/lib/__tests__/ledger.test.ts
TimoGlastra: Yeah sure, i'll move it. This was more as a note to myself while debugging the problem.

src/lib/__tests__/ledger.test.ts
TimoGlastra: The `as string` is because the `faberAgentPublicDid.did` can be `undefined` or `string` but as we set it explicitly some lines above we know it won't be `undefined`. 

But with the type improvements in the indy sdk I'm doing I think we can resolve this. The whole `faberAgentPublicDid` object will then be undefined and if it is set the `did` property will always be type `string`. Does that make sense?

TimoGlastra: Typescript knows `Verkey` type is the same as `string` so it won't complain about the differnce

src/lib/utils/did.test.ts
```diff
@@ -0,0 +1,153 @@
...
+  '8kyt-fzzq-qpqq-ljsc-5l',
+  '9noxi4nL4SiJAsFcMLp2U4',
+  'QdAJFDpbVoHYrUpNAMe3An',
+  'B9Y3e8PUKrM1ShumWU36xW',
+  '0xf3beac30c498d9e26865f34fcaa57dbb935b0d74',
+];
+
+describe('Utils | Did', () => {
+  describe('isAbbreviatedVerkey()', () => {
+    it('should return true when passed in verkey is an abbreviated verkey', () => {
```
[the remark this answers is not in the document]
TimoGlastra: Sure, this is just my personal convention for writing tests. I'll change it. But I'll like to make it `returns` instead of `return` then. Is that okay?

```
 PASS  src/lib/utils/did.ts (17.982 s)
  isAbbreviatedVerkey()
    ✓ returns true when passed in verkey is an abbreviated verkey (3332 ms)
```

When running tests, it will read like a story when you combine the `describe` and `test`:

```
isAbbreviatedVerkey() returns true when passed in verkey is an abbreviated verkey
```

That's why I normally use `should`:

```
isAbbreviatedVerkey() should return true when passed in verkey is an abbreviated verkey
```

src/lib/agent/LedgerService.ts
```diff
@@ -137,7 +137,9 @@ export class LedgerService {
       authorAgreement.version,
       authorAgreement.digest,
       this.getFirstAcceptanceMechanism(authorAgreement),
-      authorAgreement.ratification_ts
+      // Current time since epoch
+      // We can't use ratification_ts, as it must be greater than 1499906902
+      Math.floor(new Date().getTime() / 1000)
```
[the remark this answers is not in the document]
TimoGlastra: I have no idea why. Maybe this is something hardcoded in the test ledger. It's indeed `Thu Jul 13 2017 02:48:22 GMT+0200 (Central European Summer Time)`

src/lib/utils/did.test.ts
```diff
@@ -0,0 +1,153 @@
...
+      for (const verkey of validAbbreviatedVerkeys) {
+        expect(isAbbreviatedVerkey(verkey)).toBe(true);
+      }
```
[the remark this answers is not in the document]
TimoGlastra: Any preference for these two approaches? 

One test, combined testcases array:

```
    const testCases = [
      ...validAbbreviatedVerkeys.map((v) => [true, v] as const),
      ...invalidAbbreviatedVerkeys.map((v) => [false, v] as const),
    ];

    test.each(testCases)('returns %s when abbreviated verkey "%s" is passed in', (expected, verkey) => {
      expect(isAbbreviatedVerkey(verkey)).toBe(expected);
    });
```
two tests:

```
    test.each(validAbbreviatedVerkeys)('returns true when valid abbreviated verkey "%s" is passed in', (verkey) => {
      expect(isAbbreviatedVerkey(verkey)).toBe(true);
    });

    test.each(invalidAbbreviatedVerkeys)(
      'returns false when invalid abbreviated verkey "%s" is passed in',
      (verkey) => {
        expect(isAbbreviatedVerkey(verkey)).toBe(false);
      }
```

TimoGlastra:
@jakubkoci I made some changes based on your feedback. Let me know if you want any other changes

src/lib/utils/did.test.ts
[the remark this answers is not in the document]
jakubkoci: Not necessarily. Maybe, the second one. It does not introduce any special logic/code.

src/lib/utils/did.test.ts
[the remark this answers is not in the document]
jakubkoci: Yeah, `returns` is good.

src/lib/__tests__/ledger.test.ts
jakubkoci: Ah, ok, that's good to know that ACA-Py and .NET have the same issue 👍

src/lib/__tests__/ledger.test.ts
jakubkoci: > Typescript knows `Verkey` type is the same as `string` so it won't complain about the differnce

Good, that's what I would expect, but I thought maybe TS is not so great 😄 

I'm thinking if the `if` check and throwing error would not be better for understanding. Of course, it's not better from the boilerplating perspective, but I was confused by that `as string` a little bit and if someone decides to move `faberAgentPublicDid = faberAgent.getPublicDid();`  to `beforeAll` and remove the test we're in trouble :)

Not saying this is bad and we'll solve it by your work with types anyway, just sharing my perspective for such situations in general ;)

jakubkoci (approved):

PR #90, opened 2020-08-27
Fix formatting and add lint step to CI

jakubkoci:
I've found formatting error when pushing to my fork. I fix formatting and also add `lint` job to CI.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -33,6 +33,9 @@ jobs:
       - name: Build framework docker image
         run: docker build -t aries-framework-javascript .
 
+      - name: Run lint and format validation
+        run: docker run aries-framework-javascript yarn validate
+
       - name: Start mediator agents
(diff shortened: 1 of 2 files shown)


jakubkoci:
It didn't work with a specific `lint` job because of dependency on `libindy`. So I just added step into the `integration-test` job and run the validation with docker. We can eventually split this job into smaller jobs like `prepare`, `lint`, `integration-tests` or whatever is a good practice :)

TimoGlastra:
Fine for now, but it should really be a separate job.

Valid options:
- Skip the `rebuild` step of `indy-sdk`. We don't need it for linting / compiling
- Create a local `setup-indy` action (like `actions/setup-node`) that we use in every job to setup the indy dependencies. This way we don't need to build a framework docker image anymore
- As you mentioned, separate `prepare` job to build the framework image

TimoGlastra (approved):

Issue #64, opened 2020-07-17
Issue credential protocol


TimoGlastra:
Issue Credential protocol is currently at v1.1. This adds some fields in comparison to v1.0, but is completely backwards compatible with the version in AIP 1.0. 

So we should implement the current version (1.1) from the master branch.

Checklist
- [x] Credential offer exchange
  - Issuer sends credential offer which holder accepts. 
  - A credential record is created and persisted on both sides in a given state: offer-sent, offer-received.
  - Emit events about the state change.
- [x] Credential exchange
  - Issuer sends credential which holder accepts.
  - The credential state is updated accordingly: request-sent, request-received, credential-issued, credential-received.
  - Emit events about the state change.
- [x] Others 
  - Acks: We have implemented just the "Happy Path" part of the [Aries RFC 0317: Please ACK Decorator](https://github.com/hyperledger/aries-rfcs/blob/master/features/0317-please-ack/README.md), but I think it's good enough for now. The spec is still in a PROPOSED state anyway.
  - Threads: We have implemented basic threading mechanism. Other parts of the protocol will be addressed in #128.
  - ~~Encode and decode attachments with a decorator.~~ This doesn't seem as necessary to me right now.
- [x] ~~Alternative beginnings~~ 
  - ~~Start with Credential proposal~~. I've created a new task #129 for this.

Resources
- https://github.com/hyperledger/aries-rfcs/blob/master/features/0036-issue-credential/README.md

Branch with work in-progress
- https://github.com/jakubkoci/aries-framework-javascript/tree/feature/accept-credential-offer

jakubkoci:
This issue depends on #62 which is partially implemented but waiting for some infra preparation to run tests against Indy pool. However, I would like to move forward with this so I start the work and I'll try to fix #62 in parallel.

jakubkoci:
I've started with implementation. Now, I have a better understanding of the task so I've updated the issue with more specific tasks. I would like to create standalone PR for every task to keep our PRs small. I also added a branch where you can see the work in-progress.

I should have the first task pretty soon, perhaps tomorrow.

biligunb:
@jakubkoci @TimoGlastra Good luck guys.
We are using Aries Framework JS for our project currently. And eagerly waiting for the VC and Present Proof protocols.
If I may, can I ask what is the timeline for these 2 to be merged into the master branch?
[...]

Issue #60, opened 2020-07-10
Add option to disable auto-accept of connection


jakubkoci:
As a developer, I want to have an option to create a connection by accepting an invitation but wait with connection establishment until some form of confirmation happens to preview connection for approval by a user for example.

Currently, calling `agent.acceptInvitation(invitation)` creates connection and automatically starts DID exchange process responding with `ConnectionRequest` and accepting consequently incoming `ConnectionResponse` message until the connection is established. 

- AgentConfig will get option to auto-accept connections. This can be set during startup, and modified during runtime. Should we persist this? Or is that the task of the framework consumer?
- Each connection will have an option (`autoAcceptConnection`) that states whether the connection should automatically be accepted. This can be passed as option when creating / receiving invitation, but also modified whenever you want.

In the future this will work the same for issue-cred / present-proof

Specific connection config overrides global config. If not set on connection, global will be used.

Checklist

- [x] add option to enable / disable auto-accept of connections
    - [x] in agent config
    - [x] to connection record
    - [x] to create invitation function
    - [x] to accept invitation function

PR #92, opened 2020-08-28
Send and accept credential offer

jakubkoci:
This PR contains the first task "Credential offer exchange" from the #64

### DIFF
--- a/src/lib/__tests__/credentials.test.ts
+++ b/src/lib/__tests__/credentials.test.ts
@@ -32,6 +32,7 @@ describe('credentials', () => {
   let faberAgent: Agent;
   let aliceAgent: Agent;
   let schemaId: SchemaId;
+  let credDefId: CredDefId;
   let faberAgentPublicDid: DidInfo | Record<string, undefined>;
 
   beforeAll(async () => {
@@ -70,7 +71,7 @@ describe('credentials', () => {
       config: { support_revocation: false },
     };
 
-    const [credDefId] = await faberAgent.ledger.registerCredentialDefinition(credentialDefinitionTemplate);
+    [credDefId] = await faberAgent.ledger.registerCredentialDefinition(credentialDefinitionTemplate);
     const [ledgerCredDefId, ledgerCredDef] = await faberAgent.ledger.getCredentialDefinition(credDefId);
     console.log('ledgerCredDefId, ledgerCredDef', ledgerCredDefId, ledgerCredDef);
 
@@ -82,11 +83,17 @@ describe('credentials', () => {
     await aliceAgent.closeAndDeleteWallet();
   });
 
-  test(`initialization of agent's public DID`, async () => {
+  test(`when faber issues credential then alice gets credential offer`, async () => {
     faberAgentPublicDid = await faberAgent.ledger.getPublicDid('Th7MpTaRZVRYnPiabds81Y');
     console.log('faberAgentPublicDid', faberAgentPublicDid);
 
-    expect({}).toEqual(null);
+    const [firstConnection] = await faberAgent.connections.getAll();
+    console.log(firstConnection);
+
+    await faberAgent.credentials.issueCredential(firstConnection, { credDefId });
+    const [firstCredential] = await aliceAgent.credentials.getCredentials();
+
+    expect(firstCredential).toEqual({});
   });
 });
 
--- a/src/lib/__tests__/credentials.test.ts
+++ b/src/lib/__tests__/credentials.test.ts
@@ -26,7 +26,7 @@ const aliceConfig = {
   walletCredentials: { key: '00000000000000000000000000000Test01' },
 };
 
-const poolName = 'test-pool';
+const poolName = 'test-pool-credentials';
 const poolConfig = {
   genesis_txn: process.env.GENESIS_TXN_PATH
     ? path.resolve(process.env.GENESIS_TXN_PATH)
--- a/src/lib/__tests__/credentials.test.ts
+++ b/src/lib/__tests__/credentials.test.ts
@@ -1,13 +1,13 @@
 /* eslint-disable no-console */
 // @ts-ignore
 import { poll } from 'await-poll';
-import { Agent } from '..';
 import { Subject } from 'rxjs';
 import path from 'path';
 import indy from 'indy-sdk';
-import { DidInfo } from '../wallet/Wallet';
+import { Agent } from '..';
 import { SubjectInboundTransporter, SubjectOutboundTransporter } from './helpers';
 import { CredentialRecord } from '../storage/CredentialRecord';
+import { SchemaTemplate, CredDefTemplate } from '../agent/LedgerService';
 
 jest.setTimeout(15000);
 
@@ -34,9 +34,7 @@ const poolConfig = {
 describe('credentials', () => {
   let faberAgent: Agent;
   let aliceAgent: Agent;
-  let schemaId: SchemaId;
   let credDefId: CredDefId;
-  let faberAgentPublicDid: DidInfo | Record<string, undefined>;
 
   beforeAll(async () => {
     const faberMessages = new Subject();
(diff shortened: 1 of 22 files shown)


src/lib/protocols/credentials/__tests__/CredentialService.test.ts
```diff
@@ -0,0 +1,184 @@
...
+        credDefId: 'Th7MpTaRZVRYnPiabds81Y:3:CL:17:TAG',
+        comment: 'some comment',
+        preview,
+      });
+
+      expect(eventListenerMock).toHaveBeenCalledWith(
+        expect.objectContaining({
+          newState: CredentialState.OfferSent,
+          credentialId: expect.any(String),
+        })
```
jakubkoci, line 128: I'm still not sure what should be a convention for content returned by an event emitter. I'm returning `newState` and `credentialId` here, but maybe it would be better to return `credential` record itself and something what would highlight state change `{ from: offer-received, to: request-sent }`

src/lib/protocols/credentials/__tests__/CredentialService.test.ts
```diff
@@ -0,0 +1,184 @@
...
+      credentialService.on(EventType.StateChanged, eventListenerMock);
+
+      const credentialOffer = await credentialService.createCredentialOffer({
+        credDefId: 'Th7MpTaRZVRYnPiabds81Y:3:CL:17:TAG',
+        comment: 'some comment',
+        preview,
+      });
+      const messageContext = new InboundMessageContext(credentialOffer);
+
+      await credentialService.processCredentialOffer(messageContext);
```
jakubkoci, line 174: I started with `acceptCredentialOffer` but then I realized I would be implementing more than it's necessary to pass the current test. We could also solve #60 `auto-accept` for connection in a similar way. We would have two methods `processInvitation` just to create a connection record and `acceptInvitation` to create `ConnectionRequestMessage`. The logic to decide if we want to `auto-accept` should be part of `Handler`.

src/lib/protocols/credentials/messages/CredentialOfferMessage.ts
```diff
@@ -0,0 +1,99 @@
...
+  public id: string;
+
+  @Expose({ name: 'mime-type' })
+  public mimeType: string;
+
+  @Expose({ name: 'data' })
+  public data: {
+    base64: string;
+  };
+}
```
jakubkoci, line 99: These classes will be probably moved into separate files. I just want to keep it here to simplify the work with them while it's still in progress and could be changing quite often.

src/lib/__tests__/credentials.test.ts
```diff
@@ -0,0 +1,195 @@
...
+
+    const schemaTemplate = {
+      name: `test-schema-${Date.now()}`,
+      attributes: ['name', 'age'],
+      version: '1.0',
+    };
+    const [, ledgerSchema] = await registerSchema(faberAgent, schemaTemplate);
+
+    const definitionTemplate = {
+      schema: ledgerSchema,
```
TimoGlastra, line 81: I'm curious. Is there a reason the definitionTemplate includes the schema instead of the schema id? I see indy requires the schema, but as the schema should already be registered an id seems sufficient.

src/lib/__tests__/credentials.test.ts
```diff
@@ -0,0 +1,195 @@
...
+    console.log(`Ensure test DID ${publicDid} is written to ledger: Success`, agentPublicDid);
+  } catch (error) {
+    // Unfortunately, this won't prevent from the test suite running because of Jest runner runs all tests
+    // regardless thorwn errors. We're more explicit about the problem with this error handling.
+    throw new Error(`Test DID ${publicDid} is not written on ledger or ledger is not available.`);
+  }
+}
+
+async function makeConnection(agentA: Agent, agentB: Agent) {
+  const { connection: aliceConnectionAtAliceBob, invitation } = await agentA.connections.createConnection();
```
TimoGlastra, line 178: Not related to this PR, but maybe we should make the return type of `createConnection` an array just like the `createSchema` and `createCredentialDefinition` types. As the return types are almost always destructured, an array would allow to use a custom name for the return variable name.

Typescript 4.0 now also supports labeled tuple elements: https://devblogs.microsoft.com/typescript/announcing-typescript-4-0/#labeled-tuple-elements

src/lib/agent/Agent.ts
```diff
@@ -37,6 +37,10 @@ import { ConnectionsModule } from '../modules/ConnectionsModule';
 import { RoutingModule } from '../modules/RoutingModule';
 import { BasicMessagesModule } from '../modules/BasicMessagesModule';
 import { LedgerModule } from '../modules/LedgerModule';
+import { CredentialsModule } from '../modules/CredentialsModule';
+import { CredentialService } from '../protocols/credentials/CredentialService';
```
TimoGlastra, line 41: Any reason for `CredentialService` vs `CredentialsModule` (singular vs plural)? I see connection has the same difference. It makes sense in a way.

src/lib/protocols/credentials/CredentialService.ts
```diff
@@ -0,0 +1,66 @@
...
+  private credentialRepository: Repository<CredentialRecord>;
+
+  public constructor(wallet: Wallet, credentialRepository: Repository<CredentialRecord>) {
+    super();
+    this.wallet = wallet;
+    this.credentialRepository = credentialRepository;
+  }
+
+  public async createCredentialOffer({
+    credDefId,
```
TimoGlastra, line 25:
```suggestion
    credentialDefinitionId,
```

My understanding was that we will use `credDefId` only for typing indy-sdk and use the full version everywhere else in the framework

src/lib/protocols/credentials/__tests__/CredentialService.test.ts
```diff
@@ -0,0 +1,184 @@
+/* eslint-disable no-console */
```
TimoGlastra, line 1:
```suggestion
```

This is not needed in test files anymore

src/lib/protocols/credentials/__tests__/CredentialService.test.ts
TimoGlastra: I think it would be useful to return the whole credential record. I also think it would be useful to get the previous state.

We could also use `newState` and `prevState`, but I kinda like `from` and `to`

src/lib/protocols/credentials/__tests__/CredentialService.test.ts
TimoGlastra: Agree!

src/lib/protocols/credentials/messages/CredentialOfferMessage.ts
```diff
@@ -0,0 +1,95 @@
...
+import { MessageType } from './MessageType';
+
+interface CredentialPreviewOptions {
+  attributes: CredentialPreviewAttribute[];
+}
+
+export class CredentialPreview {
+  public constructor(options: CredentialPreviewOptions) {
+    this.attributes = options.attributes;
+  }
```
TimoGlastra, line 14: You should check if options is defined or provide a default value (` = {}`) because class transformer calls the constructor.

src/lib/protocols/credentials/messages/CredentialOfferMessage.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  @Expose({ name: 'offers~attach' })
+  public offersAttachments!: Attachment[];
+}
+
+export class Attachment {
+  public constructor(options: Attachment) {
+    this.id = options.id;
+    this.mimeType = options.mimeType;
+    this.data = options.data;
```
TimoGlastra, line 82: Same here with options check

src/lib/wallet/IndyWallet.ts
```diff
@@ -69,6 +69,13 @@ export class IndyWallet implements Wallet {
     return this.indy.issuerCreateAndStoreCredentialDef(this.wh, issuerDid, schema, tag, signatureType, config);
   }
 
+  public async createCredentialOffer(credDefId: CredDefId) {
+    if (!this.wh) {
+      throw Error('Wallet has not been initialized yet');
+    }
```
TimoGlastra, line 75: I'm not a fan of this function also added to the wallet. Basically the same problem as with create credential definition....

TimoGlastra (approved):
Great work Jakub.

Left some small comments, but nothing that is a blocker for me

src/lib/__tests__/credentials.test.ts
jakubkoci: Good point. I don't know 🤷‍♂️  Do you suggest passing just `schemaId` into `definitionTemplate` and then get `schema` by calling `ledgerService.getCredentialSchema(schemaId)`? That would nice simplification, but not sure about doing this step right now. I would rather see the framework in real-world usage first and get some feedback.

src/lib/__tests__/credentials.test.ts
jakubkoci: That's an interesting idea. I would be worried about confusion in comparison with `createSchema` and `createCredentialDefinition` where you get `id` as the first and `object` as the second. As I'm thinking about that I've never liked this `createConnection` API. We should make `invitation` part of connection and add `role` = `Inviter/Invitee` as attribute.

src/lib/agent/Agent.ts
jakubkoci: Hmm, you got me :) `CredentialService` similarly like `CredentialRepository` is about the work with one entity. Plural naming in `CredentialsModule` is about the work with various parts related to credentials, but it's perhaps driven more by outside API.

src/lib/protocols/credentials/__tests__/CredentialService.test.ts
jakubkoci: Yeah. I'll remove it.

src/lib/protocols/credentials/CredentialService.ts
jakubkoci: Ok 👍  I probably missed that or was lazy to replace it :)

src/lib/wallet/IndyWallet.ts
jakubkoci: Yeah, unfortunately. I got an idea to have more objects according to their responsibility, each with reference to the same `wh` which could be passed during the creation of such objects by a builder.

src/lib/protocols/credentials/__tests__/CredentialService.test.ts
TimoGlastra: I think previous can still be useful. What if i receive a credential proposal and don't have the credential in memory at the moment. The proposal-received state could come from either a new proposal, or an already existing credential exchange that needs renegotiation.

jakubkoci: Hmm, good point 🤔

PR #98, opened 2020-09-07
auto accept connection config (and connection overhaul)

### LINKED ISSUE #60: Add option to disable auto-accept of connection

### LINKED ISSUE #74: Only register key with mediator if mediator is used
Currently when a connection invitation is received in the `InvitationHandler` the key of the did used for that connection is always registered with the mediator. This should only happen in the case a mediator is used. 

This is already correctly handled when creating a connection invitation (instead of receiving one): https://github.com/hyperledger/aries-framework-javascript/blob/master/src/lib/modules/ConnectionsModule.ts#L36-L40

But this is not handled correctly here:
https://github.com/hyperledger/aries-framework-javascript/blob/master/src/lib/handlers/connections/InvitationHandler.ts#L22-L23

TimoGlastra:
The intent of this PR was to add a configuration to auto accept connections, but I got carried away and did a connection protocol overhaul. All messages now have a `createXX` and `processXXX` in the connection service. The module combines these functions to get the right flow.

I like this approach because there is a clear distinction between the modules and the services. 

ConnectionService now has 100% test coverage (yay!). To not get the PR any bigger than it already is, ConnectionModule tests will be addressed in a separate PR.

- Removed `ConnectionInvitationHandler`. I think it is better to just handle it via the module and service directly, as it is always send out of band and there has no reason to pass trough the handler.
- added global `autoAcceptConnections` parameter to `AgentConfig`
- added `autoAcceptConnection` config to `ConnectionRecord` and related functions
- `ConnectionModule` and connection handlers only invoke next step in protocol if auto accept of connection is turned on
- add some extra checks to improve connection handling
- `createConnectionWithInvitation` now just returns a `ConnectionRecord` that contains the `ConnectionInvitationMessage`
    - I think it would be handy to also move the records to `class-transformer`. this way it would be really easy to store and transform agent messages inside records. Now we have to do some custom transforming before and after persisting.

Fixes #74 
Fixes #60


src/lib/protocols/connections/ConnectionService.test.ts
```diff
@@ -1,74 +1,820 @@
...
+        recipientVerkey: 'test-verkey',
+        connection,
+      });
+
+      const updatedConnection = await connectionService.processAck(messageContext);
+
+      expect(updatedConnection.state).toBe(ConnectionState.Complete);
+    });
+
+    it('does not update the state when te state is not Responded or the role is not Inviter', async () => {
```
jakubkoci, line 620: Typo `when te state`?

src/lib/protocols/connections/ConnectionService.ts
```diff
@@ -96,58 +187,127 @@ class ConnectionService extends EventEmitter {
...
 
-  public async acceptResponse(messageContext: InboundMessageContext<ConnectionResponseMessage>) {
+  /**
+   * Process a received connection response message. This will not accept the connection request
+   * or send a connection acknowledgement message. It will only update the existing connection record
+   * with all the new information from the connection response message. Use {@link ConnectionService#createTrustPing}
+   * after calling this function to create a trust ping message.
+   *
+   * @param messageContext the message context containing a connetion response message
+   */
```
jakubkoci, line 205: I don't see `@returns` clause in docs for `processXxx` methods, is it on purpose?

jakubkoci (changes requested):
Awesome work @TimoGlastra 👍  My concern is just about `listener` on events as I described in the comment.

src/lib/modules/ConnectionsModule.ts
```diff
@@ -1,69 +1,148 @@
...
+
+    return new Promise(resolve => {
+      const listener = ({ connection }: ConnectionStateChangedEvent) => {
+        if (isConnected(connection)) {
+          this.events().off(ConnectionEventType.StateChanged, listener);
+          resolve(connection);
+        }
+      };
+
+      this.events().on(ConnectionEventType.StateChanged, listener);
```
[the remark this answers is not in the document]
TimoGlastra: As far as I know this doesn't cause unnecessary memory allocation.

1. Get connection record from connection repository. If connection state is already `Complete`, return before adding event listener.
1. Start listener to listen for all connection events. Every time the state of a connection changed, an event is emitted.
1. If the state of the connection we were looking for is updated to `Complete` the `this.events.off()` is called before resolving. This way the listener won't be called anymore.

We could optimise this however, if we only create one listener and keep a list of connection ids to look for. However this seems like premature optimisation at the moment.

src/lib/protocols/connections/ConnectionService.ts
TimoGlastra: No not on purpose. This is because I first implemented all `processXXX` to not return anything but I later changed them all to return the updated connection record.

src/lib/modules/ConnectionsModule.ts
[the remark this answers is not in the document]
jakubkoci: Ah, I see, I missed that `off(..., listener)`

jakubkoci (approved):

PR #100, opened 2020-10-01
auto init ledger pool if genesisPath is configured

### LINKED ISSUE #88: Allow to pass genesis path (or URL) to agent config
Currently the framework connects to the indy pool using the `agent.ledger.connect(poolName, poolConfig)` function outside of the framework.

I think it would be convenient to allow to set the genesis path using the agent config. This is also present in framework .NET and would make the setup of the framework easier.

```javascript
const agentConfig = {
  // ... rest of agent config
  genesisPath: process.env.GENESIS_TXN_PATH
};
```

If the `genesisPath` is present in the agent config, it will be used to automatically connect to the ledger when needed. If not set, the framework consumer can get more control by using the already present `agent.ledger.connect(poolName, poolConfig)`. This will make it easier to use the framework, while still allowing to control over ledger setup.

---

We could also allow to pass `genesisUrl` to automatically download the genesis.

TimoGlastra:
This is a small change that automatically connect to the ledger pool (instead of manually needing to do it) if `genesisPath` configuration is set. It is backwards compatible (by manually calling `ledger.connect`). 

In the future I'll like to optimise this to only connect when the pool is actually needed, but this way the configuration and API is already available. The optimisation can be an internal change. Maybe we can take a similar approach as Aries Framework .NET that uses an awaitable pool that initialises when you retrieve it.

Fixes #88

### DIFF
--- a/src/lib/__tests__/credentials.test.ts
+++ b/src/lib/__tests__/credentials.test.ts
@@ -12,12 +12,18 @@ import { CredentialPreview } from '../protocols/credentials/messages/CredentialO
 import { CredentialState } from '../protocols/credentials/CredentialState';
 import { InitConfig } from '../types';
 
+const genesisPath = process.env.GENESIS_TXN_PATH
+  ? path.resolve(process.env.GENESIS_TXN_PATH)
+  : path.join(__dirname, '../../../network/genesis/local-genesis.txn');
+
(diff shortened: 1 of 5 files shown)


jakubkoci (approved):
Good idea 👍

PR #108, opened 2020-10-20
add diddoc class transformer

### LINKED ISSUE #53: Add better lookup for DIDDoc keys
As @TimoGlastra mentioned in #53 "Currently all connections use the first key with the first service so it should work fine, but i think it would be better to look for the IndyAgent service." And also suggested this link https://github.com/decentralized-identity/did-common-typescript/blob/master/lib/DidDocument.ts as one of the possible solutions.

We can take it as an inspiration, or maybe, add the whole package as a dependency. Although, I'm not sure about the work needed for the persistence of such an object with our current storage mechanism

TimoGlastra:
for now a draft PR as I'm using an unreleased version of class-transformer from the master branch that contains features not in the latest version. Hope they'll release it soon.

`ConnectionRecord` is a bit of a mess now because of all the custom transforms. This should be resolved soon once the connection classes are also transformed using class transformer.

100% test coverage for the new DidDoc related classes.

Fixes #53


TimoGlastra:
@jakubkoci this is up to date with master

src/lib/protocols/connections/domain/did/__tests__/Service.test.ts
```diff
@@ -0,0 +1,148 @@
...
+      };
+      const serviceWrapper = plainToClass(ServiceTransformerTest, serviceWrapperJson);
+
+      expect(serviceWrapper.service.length).toBe(serviceArray.length);
+
+      for (let i = 0; i < serviceArray.length; i++) {
+        const serviceJson = serviceArray[i];
+        const service = serviceWrapper.service[i];
+
+        expect(service).toBeInstanceOf(serviceTypes[serviceJson.type]);
```
jakubkoci, line 144: I usually try to avoid classic `for` loops if possible. Using `forEach` for example seems more readable to me. 

What do you think?
```
serviceArray.forEach((serviceJson, i) => { 
  const service = serviceWrapper.service[i];  
  expect(service).toBeInstanceOf(serviceTypes[serviceJson.type])
})
```

src/lib/protocols/connections/domain/did/__tests__/Service.test.ts
```diff
@@ -0,0 +1,148 @@
...
+        serviceEndpoint: 'https://example.com',
+      };
+      const service = plainToClass(IndyAgentService, json);
+
+      expect(service.id).toBe(json.id);
+      expect(service.type).toBe(json.type);
+      expect(service.recipientKeys).toEqual(json.recipientKeys);
+      expect(service.routingKeys).toEqual(json.routingKeys);
+      expect(service.priority).toBe(json.priority);
+      expect(service.serviceEndpoint).toBe(json.serviceEndpoint);
```
jakubkoci, line 52: I guess we can't use `toMatchObject` because we try to test class attributes against JSON properties right?

jakubkoci (commented):
It seems good. I added just a few points I spot. These transformations and use of classes and inheritance look too heavyweight for my taste, maybe I just need to get used to it :)

package.json
```diff
@@ -25,7 +25,7 @@
     "await-poll": "^1.0.2",
     "bn.js": "^5.1.3",
     "buffer": "^6.0.2",
-    "class-transformer": "^0.3.1",
+    "class-transformer": "^0.3.2",
```
jakubkoci, line 28: I figured out that the problem I had after the framework update was caused by a different version of `class-transformer`. Could you please change this to strict version number `0.3.2`?

jest.config.js
```diff
@@ -11,5 +11,5 @@ module.exports = {
   transform: {
     '^.+\\.(ts|tsx)?$': 'ts-jest',
   },
-  testTimeout: 30000,
+  testTimeout: 60000,
```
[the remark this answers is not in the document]
jakubkoci: What about lowering it to 45 at least? I don't have a strong argument, just feeling :)

jakubkoci (approved):

PR #109, opened 2020-10-20
only return object for get schema and cred def

### LINKED ISSUE #105: getCredentialDefinition, getSchema should only return the object, not an array of [id, object]
Currently `getCredentialDefinition` and `getSchema` from `LedgerService` return an array with `[id, object]` (e.g. `[schemaId, schema]`) however as you need to pass the id of the cred def / schema it doesn't make sense to also return the id.

https://github.com/hyperledger/aries-framework-javascript/blob/b0b5a64e9592e4aeaa2a4d5b92ac6f9f0a57a131/src/lib/agent/LedgerService.ts#L70

https://github.com/hyperledger/aries-framework-javascript/blob/b0b5a64e9592e4aeaa2a4d5b92ac6f9f0a57a131/src/lib/agent/LedgerService.ts#L110

TimoGlastra:
Only return the objects for `getCredentialSchema` and `getCredentialDefinition` instead of both object and id

Fixes #105


jakubkoci (approved):
Nice improvement 👍

PR #111, opened 2020-10-21
add JsonTransformer for generic json transform

### LINKED ISSUE #107: add JsonTransformer class
We already have a `MessageTransformer` class that transforms `AgentMessages` (DIDComm messages) from and to JSON. However now that we're also (going to) using class-transformer for other purposes (records, diddocs, ...) we should have a general `JsonTransformer` class that transforms from and to JSON.

This way we can avoid direct usage of `classToPlain` and `plainToClass` in the codebase. Especially now that we're going to use some custom settings when calling these functions.

TimoGlastra:
I wasn't exactly sure about the location of `JsonTransformer`. @jakubkoci I see you include d the `JsonEncoder` inside the credentials protocol directory. Maybe we can move them both to a fitting location?

Fixes #107


src/lib/protocols/credentials/CredentialService.ts
```diff
@@ -103,7 +103,11 @@ export class CredentialService extends EventEmitter {
     credDef: CredDef
   ): Promise<CredentialRequestMessage> {
     const proverDid = connection.did;
-    const offer = MessageTransformer.toMessageInstance(credential.offer, CredentialOfferMessage);
+
+    // FIXME: credential.offer is already CredentialOfferMessage type
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    const offer = JsonTransformer.fromJSON(credential.offer, CredentialOfferMessage);
```
jakubkoci, line 110: I don't understand why credential.offer is already CredentialOfferMessage type but we still calling the transformation?

jakubkoci (approved):
Looks good. Just if you could elaborate a little bit on the offer transformation?

src/lib/protocols/credentials/CredentialService.ts
TimoGlastra, line 110: Should've made this more clear. TS thinks it is already `CredentialOfferMessage`, however it still needs to be transformed. This can be confusing as you think it is already transformed, but it is not. See the suggested change

```suggestion
    // FIXME: TypeScript thinks the type of credential.offer is already CredentialOfferMessage, but it still needs to be transformed
    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
    // @ts-ignore
    const offer = JsonTransformer.fromJSON(credential.offer, CredentialOfferMessage);
```

jakubkoci: I see. Yes, that's better 👍

PR #119, opened 2020-10-24
fix: Correctly persist createdAt attribute

jakubkoci:
I needed to move `createAt` attribute initialization up in the inheritance chain to every particular record. I'm not sure if this is a good approach in the long-term but it at least fixes the issue for now. I also aligned `id` attribute initialization across all records, so even if this is a wrong approach we have it wrong in the same way everywhere :)

### DIFF
--- a/src/lib/storage/BasicMessageRecord.ts
+++ b/src/lib/storage/BasicMessageRecord.ts
@@ -18,7 +18,7 @@ export class BasicMessageRecord extends BaseRecord implements BasicMessageStorag
   public readonly type = BasicMessageRecord.type;
 
   public constructor(props: BasicMessageStorageProps) {
-    super(props.id || uuid(), props.createdAt || Date.now());
+    super(props.id ?? uuid(), props.createdAt ?? Date.now());
     this.content = props.content;
     this.sentTime = props.sentTime;
     this.tags = props.tags;
--- a/src/lib/storage/BasicMessageRecord.ts
+++ b/src/lib/storage/BasicMessageRecord.ts
@@ -3,6 +3,7 @@ import { BaseRecord, RecordType, Tags } from './BaseRecord';
 
 export interface BasicMessageStorageProps {
   id?: string;
+  createdAt?: number;
   tags: Tags;
 
   content: string;
@@ -17,7 +18,7 @@ export class BasicMessageRecord extends BaseRecord implements BasicMessageStorag
   public readonly type = BasicMessageRecord.type;
 
   public constructor(props: BasicMessageStorageProps) {
-    super(props.id || uuid());
+    super(props.id || uuid(), props.createdAt || Date.now());
     this.content = props.content;
     this.sentTime = props.sentTime;
     this.tags = props.tags;
--- a/src/lib/storage/ConnectionRecord.ts
+++ b/src/lib/storage/ConnectionRecord.ts
@@ -51,7 +51,7 @@ export class ConnectionRecord extends BaseRecord implements ConnectionStoragePro
   public readonly type = ConnectionRecord.type;
 
   public constructor(props: ConnectionStorageProps) {
-    super(props.id || uuid(), props.createdAt || Date.now());
+    super(props.id ?? uuid(), props.createdAt ?? Date.now());
     this.did = props.did;
     this.didDoc = props.didDoc;
     this.verkey = props.verkey;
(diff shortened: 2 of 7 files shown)


src/lib/storage/BasicMessageRecord.ts
```diff
@@ -17,7 +18,7 @@ export class BasicMessageRecord extends BaseRecord implements BasicMessageStorag
   public readonly type = BasicMessageRecord.type;
 
   public constructor(props: BasicMessageStorageProps) {
-    super(props.id || uuid());
+    super(props.id || uuid(), props.createdAt || Date.now());
```
TimoGlastra, line 21: TypeScript 4.0 introduced some new [short circuit assignment operators](https://devblogs.microsoft.com/typescript/announcing-typescript-4-0/#short-circuiting-assignment-operators). Which are perfect for this usage.

```suggestion
    super(props.id ?? uuid(), props.createdAt ?? Date.now());
```

src/lib/storage/ConnectionRecord.ts
```diff
@@ -49,7 +51,7 @@ export class ConnectionRecord extends BaseRecord implements ConnectionStoragePro
   public readonly type = ConnectionRecord.type;
 
   public constructor(props: ConnectionStorageProps) {
-    super(props.id);
+    super(props.id || uuid(), props.createdAt || Date.now());
```
TimoGlastra, line 54: Same

src/lib/storage/CredentialRecord.ts
```diff
@@ -27,7 +28,7 @@ export class CredentialRecord extends BaseRecord implements CredentialStoragePro
   public state: CredentialState;
 
   public constructor(props: CredentialStorageProps) {
-    super(props.id ? props.id : uuid());
+    super(props.id || uuid(), props.createdAt || Date.now());
```
TimoGlastra, line 31: Same

src/lib/storage/IndyStorageService.test.ts
```diff
@@ -18,20 +19,20 @@ class TestRecord extends BaseRecord {
   public readonly type = TestRecord.type;
 
   public constructor(props: TestRecordProps) {
-    super(props.id || uuid());
+    super(props.id || uuid(), props.createdAt || Date.now());
```
TimoGlastra, line 22: Same

src/lib/storage/ProvisioningRecord.ts
```diff
@@ -15,7 +17,7 @@ export class ProvisioningRecord extends BaseRecord {
   public readonly type = ProvisioningRecord.type;
 
   public constructor(props: ProvisioningRecordProps) {
-    super(props.id);
+    super(props.id || uuid(), props.createdAt || Date.now());
```
TimoGlastra, line 20: Same

TimoGlastra (approved):
Looks good, just the suggestion to use the new short circuit operator

jakubkoci:
> Looks good, just the suggestion to use the new short circuit operator

Hmm, that interesting, but how does it actually work? I'm looking around but can't find anything about that 😄

TimoGlastra:
> > Looks good, just the suggestion to use the new short circuit operator
> 
> Hmm, that interesting, but how does it actually work? I'm looking around but can't find anything about that 😄

I linked to the wrong thing. What I actually mean is ["nullish coalescing"](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing)

Basically when you use `||` it will look if the left hand side results in a truthy or falsy value. This means if I provide `createdAt`. value of 0 it will equal to falsy and use `Date.now`. With `??` it will check whether the variable has a value (so not `null` or `undefined`). This means that a value of 0 will not be seen as falsy and will be used.

src/lib/storage/BasicMessageRecord.ts
jakubkoci: Ah, I see. That's nice. I can't even count how often I solve issues with `if (number) ...` 😄

jakubkoci: Hmm, as I'm thinking about it. Are they really perfect for this usage? Maybe values `''` and `0` are not quite valid as `id` or `createAt`

jakubkoci: But that's perhaps the responsibility of some eventual validation. Ok, I'll change it to `??`

TimoGlastra: Well yeah, 0 is probably not a commonly used value. However it is unexpected behaviour IMO to change the value without notic. When I pass a value of `0` and it is automatically changed to `Date.now` without warning or error that is weird I think. Same with empty string.

PR #126, opened 2020-10-30
refactor: Credential service method signatures

### LINKED ISSUE #117: process methods in CredentialService don't return anything while ConnectionService methods return the connection
For the sake of consistency, the same as with #115 

As a user of the framework I expect the API's to be consistent. The `processXXX` methods in ConnectionService all return the connection afterwards. However the `processXXX` methods in CredentialService return nothing.

This is inconsistent and I think we should pick either one.

My preference would go to returning the credential object, as we already have it. But open to either.

jakubkoci:
Small changes in credential service method signatures, renaming params, add typing. This PR together with #122 fixes #117.

### DIFF
--- a/src/lib/modules/CredentialsModule.ts
+++ b/src/lib/modules/CredentialsModule.ts
@@ -54,8 +54,7 @@ export class CredentialsModule {
     const credentialRequestMessage = await this.credentialService.createCredentialRequest(
       connection,
       credential,
-      credentialDefinition,
-      {}
+      credentialDefinition
     );
 
     const outboundMessage = createOutboundMessage(connection, credentialRequestMessage);
--- a/src/lib/modules/CredentialsModule.ts
+++ b/src/lib/modules/CredentialsModule.ts
@@ -54,7 +54,8 @@ export class CredentialsModule {
     const credentialRequestMessage = await this.credentialService.createCredentialRequest(
       connection,
       credential,
-      credentialDefinition
+      credentialDefinition,
+      {}
     );
 
     const outboundMessage = createOutboundMessage(connection, credentialRequestMessage);
--- a/src/lib/protocols/credentials/CredentialService.ts
+++ b/src/lib/protocols/credentials/CredentialService.ts
@@ -106,7 +106,7 @@ export class CredentialService extends EventEmitter {
     connection: ConnectionRecord,
     credential: CredentialRecord,
     credentialDefinition: CredDef,
-    options: CredentialRequestOptions
+    options: CredentialRequestOptions = {}
   ): Promise<CredentialRequestMessage> {
     const proverDid = connection.did;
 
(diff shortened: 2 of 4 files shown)


src/lib/protocols/credentials/CredentialService.ts
```diff
@@ -100,12 +100,13 @@ export class CredentialService extends EventEmitter {
...
    * @param credential
-   * @param credDef
+   * @param credentialDefinition
    */
   public async createCredentialRequest(
     connection: ConnectionRecord,
     credential: CredentialRecord,
-    credDef: CredDef
+    credentialDefinition: CredDef,
+    options: CredentialRequestOptions
```
TimoGlastra, line 109: I think `options` should be optional, as all parameters are optional

src/lib/modules/CredentialsModule.ts
```diff
@@ -54,7 +54,8 @@ export class CredentialsModule {
     const credentialRequestMessage = await this.credentialService.createCredentialRequest(
       connection,
       credential,
-      credentialDefinition
+      credentialDefinition,
+      {}
```
TimoGlastra, line 58: Then we don't have to pass in empty objects

src/lib/protocols/credentials/CredentialService.ts
```diff
@@ -167,7 +167,10 @@ export class CredentialService extends EventEmitter {
    * @param credentialId Credential record ID
    * @param credentialResponseOptions
    */
-  public async createCredentialResponse(credentialId: string, { comment }: CredentialResponseOptions) {
+  public async createCredentialResponse(
+    credentialId: string,
+    options: CredentialResponseOptions
```
TimoGlastra, line 172: Same

src/lib/protocols/credentials/__tests__/CredentialService.test.ts
```diff
@@ -286,15 +286,22 @@ describe('CredentialService', () => {
...
 
     test('returns credential request message base on existing credential offer message', async () => {
-      const credentialRequest = await credentialService.createCredentialRequest(connection, credentialRecord, credDef);
+      // given
+      const comment = 'credential request comment';
+
+      // when
+      const credentialRequest = await credentialService.createCredentialRequest(connection, credentialRecord, credDef, {
+        comment,
+      });
```
TimoGlastra, line 295: Also 4 parameters is a lot I think, wouldn't it be better to make it a single object? This is not for this PR, but just a thought.

```suggestion
      const credentialRequest = await credentialService.createCredentialRequest({
          connection, 
          credentialRecord, 
          credentialDefinition: credDef,
          comment,
      });
```

TimoGlastra (changes requested):

TimoGlastra (approved):

PR #130, opened 2020-10-31
feat: Validate credential record state transitions

### LINKED ISSUE #123: Before updating credential state to Done we should check if the state is CredentialIssued
Before updating the state we should check if the state is `CredentialIssued`. Otherwise we could go directly from `OfferSent` to `Done` without actually issuing a credential

_Originally posted by @TimoGlastra in https://github.com/hyperledger/aries-framework-javascript/pull/122#discussion_r513807673_

jakubkoci:
I added validation to every credential service method, not sure if it is necessary. I use different parametrized tests than `jest.each`. I'm aware that's not so simple to read in this way, but I wanted to have just one test for every invalid state. This should fix #123.

I also added two unrelated tests for credential record attributes.

### DIFF
--- a/src/lib/protocols/credentials/CredentialService.ts
+++ b/src/lib/protocols/credentials/CredentialService.ts
@@ -106,7 +106,7 @@ export class CredentialService extends EventEmitter {
     credentialDefinition: CredDef,
     options: CredentialRequestOptions = {}
   ): Promise<CredentialRequestMessage> {
-    this.validateState(credential.state, CredentialState.OfferReceived);
+    this.assertState(credential.state, CredentialState.OfferReceived);
 
     const proverDid = connection.did;
 
@@ -154,7 +154,7 @@ export class CredentialService extends EventEmitter {
       threadId: messageContext.message.thread?.threadId,
     });
 
-    this.validateState(credential.state, CredentialState.OfferSent);
+    this.assertState(credential.state, CredentialState.OfferSent);
 
     logger.log('Credential record found when processing credential request', credential);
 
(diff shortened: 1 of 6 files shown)


src/lib/protocols/credentials/CredentialService.ts
```diff
@@ -266,6 +280,12 @@ export class CredentialService extends EventEmitter {
     return this.credentialRepository.find(id);
   }
 
+  private validateState(current: CredentialState, valid: CredentialState) {
```
TimoGlastra, line 283: Maybe call it `assertState` instead? To me validate state suggests there is some custom validation done in the function, while it basically checks if the first state passed is equal to the second state passed.

```suggestion
  private assertState(current: CredentialState, expected: CredentialState) {
```

src/lib/protocols/credentials/CredentialService.ts
```diff
@@ -167,24 +169,25 @@ export class CredentialService extends EventEmitter {
    * @param credentialId Credential record ID
    * @param credentialResponseOptions
    */
-  public async createCredentialResponse(
+  public async createResponse(
```
TimoGlastra, line 172: I'm not sure `createResponse` is the best name? I'm thinking of `createCredential`. Then we could also have `processCredential`

TimoGlastra (approved):

src/lib/protocols/credentials/CredentialService.ts
jakubkoci: That's good point, let's change it. Updated.

src/lib/protocols/credentials/CredentialService.ts
jakubkoci: Hmm, I understand. I also had a problem with this naming. It could be confused with an HTTP request/response. However, now we have consistency at least. Otherwise, we would need to change also other methods, I think. If you look at it from the protocol perspective, this naming is not actually that bad (`createRequest`, `createResponse`). Maybe, this message creation and processing stuff could evolve into something called `CredentialProtocol` calling service to do other tasks (saving credential record, creating protocol attachments, ...)

src/lib/protocols/credentials/CredentialService.ts
jakubkoci: I'm thinking that we could create another object, something like `CredentialStateMachine` which would be responsible for such assertions and state transitions. It could simplify tests. I haven't thought through all the details yet.

PR #133, opened 2020-11-03
create master secret per wallet

### LINKED ISSUE #131: create master_secret once
Currently the `master_secret` used for credentials is created for every credentials. This makes credentials unlinkable. 

We could take the same approach as ACA-Py by creating the master secret when opening (and possibly creating) the wallet. Where we could catch if the secret already exists (just like WalletAlreadyExistsError). This way each wallet will have an unique master secret.

TimoGlastra:
Taken from the [ACA-Py implementation](https://github.com/hyperledger/aries-cloudagent-python/blob/ec9871f84d1faae696284c1d399ff8e4eb1f86e7/aries_cloudagent/wallet/indy.py#L278-L289). They create a master secret on wallet open that uses the same id as the wallet.

The current method created a master secret for each credential, with the same id. Which wouldn't allow to request more than one credential because the second time `proverCreateMasterSecret` would be called `AnoncredsMasterSecretDuplicateNameError` error would be thrown.

Fixes #131

### DIFF
--- a/src/lib/wallet/IndyWallet.ts
+++ b/src/lib/wallet/IndyWallet.ts
@@ -4,6 +4,7 @@ import { Wallet, DidInfo } from './Wallet';
 
 export class IndyWallet implements Wallet {
   private wh?: number;
+  private masterSecretId?: string;
   private walletConfig: WalletConfig;
   private walletCredentials: WalletCredentials;
   private publicDidInfo: DidInfo | undefined;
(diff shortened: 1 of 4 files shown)


src/lib/wallet/IndyWallet.ts
```diff
@@ -32,6 +33,22 @@ export class IndyWallet implements Wallet {
...
+      logger.log('error', error);
+      if (error.indyName && error.indyName === 'AnoncredsMasterSecretDuplicateNameError') {
+        // master secret id is the same as the master secret id passed in the create function
+        // so if it already exists we can just assign it.
+        this.masterSecretId = this.walletConfig.id;
+        logger.log(`master secret with id ${this.masterSecretId} already exists`, error.indyName);
+      } else {
+        throw error;
+      }
+    }
```
TimoGlastra, line 50: I'd like to make a generic approach for calling indy functions that are allowed to fail. As this snippet is exactly the. same as the snippet above that tries to create a wallet that possibly already exists. But that's something for a future PR I think

jakubkoci: That's right and we don't have to hurry with this. I decided to leave it and wait for more cases like that to create better abstraction afterward.

jakubkoci (approved):
It seems like a simple, good enough solution 👍

PR #145, opened 2020-12-10
Added default mediator alias + support to specify mediator alias

TheArhaam:
**Concern:** Unable to set `alias` for Mediator

**Cause:** `processInvitation` does have support to set `alias` however, no `alias` was being passed from `RoutingModule.provision()`

**Correction:**
- Added default `alias` within `RoutingModule.provision()`
- Added `alias` to `MediatorConfiguration`

(Updated tests as well)

### DIFF
--- a/src/lib/protocols/connections/ConnectionService.test.ts
+++ b/src/lib/protocols/connections/ConnectionService.test.ts
@@ -162,7 +162,7 @@ describe('ConnectionService', () => {
 
   describe('processInvitation', () => {
     it('returns a connection record containing the information from the connection invitation', async () => {
-      expect.assertions(7);
+      expect.assertions(9);
 
       const recipientKey = 'key-1';
       const invitation = new ConnectionInvitationMessage({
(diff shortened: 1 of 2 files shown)


src/lib/protocols/connections/ConnectionService.test.ts
```diff
@@ -185,6 +186,8 @@ describe('ConnectionService', () => {
         })
       );
       expect(connection.invitation).toMatchObject(invitation);
+      expect(connection.alias).toBeUndefined();
+      expect(connectionAlias.alias).toBe('test-alias');
```
TimoGlastra, line 190: You already added these tests to the previous PR right?

TimoGlastra (approved):

src/lib/protocols/connections/ConnectionService.test.ts
TheArhaam: @TimoGlastra No I missed out on the `processInvitation` tests in the previous PR

PR #149, opened 2020-12-13
make routing keys optional in connection invitation message

TimoGlastra:

### DIFF
--- a/src/lib/protocols/connections/__tests__/ConnectionInvitationMessage.test.ts
+++ b/src/lib/protocols/connections/__tests__/ConnectionInvitationMessage.test.ts
@@ -1,0 +1,28 @@
+import { validateOrReject } from 'class-validator';
+
+import { ConnectionInvitationMessage } from '../ConnectionInvitationMessage';
+import { JsonTransformer } from '../../../utils/JsonTransformer';
+
+describe('ConnectionInvitationMessage', () => {
+  it('should allow routingKeys to be left out of inline invitation', async () => {
+    const json = {
+      '@type': ConnectionInvitationMessage.type,
+      '@id': '04a2c382-999e-4de9-a1d2-9dec0b2fa5e4',
+      recipientKeys: ['recipientKeyOne', 'recipientKeyTwo'],
+      serviceEndpoint: 'https://example.com',
+      label: 'test',
+    };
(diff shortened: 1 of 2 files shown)


src/lib/protocols/connections/__tests__/ConnectionInvitationMessage.test.ts
```diff
@@ -0,0 +1,28 @@
...
+  it('should allow routingKeys to be left out of inline invitation', async () => {
+    const json = {
+      '@type': ConnectionInvitationMessage.type,
+      '@id': '04a2c382-999e-4de9-a1d2-9dec0b2fa5e4',
+      recipientKeys: ['recipientKeyOne', 'recipientKeyTwo'],
+      serviceEndpoint: 'https://example.com',
+      label: 'test',
+    };
+    const invitation = JsonTransformer.fromJSON(json, ConnectionInvitationMessage);
+    await expect(validateOrReject(invitation)).resolves.toBeUndefined();
```
jakubkoci, line 16: This `toBeUndefined` seems kind of weird here. Do I understand it correctly that we just need to know the promise `resolves`?

TimoGlastra: Yes. We don't care about the value.

I've found [this stackoverflow answer](https://stackoverflow.com/a/54548606) which mentions:

> If you know that foo() always resolves to a value then you can use .toBeDefined():
> ```js
> it('resolves', async () => {
>   await expect(foo()).resolves.toBeDefined();
> });
> ```
> This doesn't work if foo() completes using return undefined or return (without a value) or it doesn't use return at all.
> 
> In this case you cannot use .resolves but you can simply return the Promise returned by foo() and Jest takes care of the rest:
> ```js
> it('resolves', () => {
>   return foo();
> });
> ```

So we could also return the promise:

```js
...
const invitation = JsonTransformer.fromJSON(json, ConnectionInvitationMessage);
return validateOrReject(invitation)
```

But that doesn't necessarily seem like a better option to me.

jakubkoci: I see. Ok, no problem, that's rather a minor thing, so we can merge it anyway.

jakubkoci (approved):

PR #150, opened 2020-12-14
add general purpose attachment decorator

TimoGlastra:
This adds a general purpose attachment decorator.

This way both the issue credential and present proof protocol can reuse the same attachment format.

It does not support signed attachments yet, we can add those when needed (probably for didexchange protocol)


TimoGlastra:
I'll add tests soon, but I want to have this in master as multiple people are working on present proof

jakubkoci (approved):

PR #158, opened 2020-12-30
add ssi-agent-wire type to mediator

TimoGlastra:
ACA-Py sends outbound messages with the `application/ssi-agent-wire` Content-Type header, which the mediator didn't recognise.


jakubkoci (approved):
Nice catch @TimoGlastra

PR #160, opened 2021-01-11
add method to check encoding, improve encode

TimoGlastra:
I'm working on the present proof protocol, and part of it is verifying the raw against the encoded values. As the PR is already quite big I thought this could be easily added as a separate PR.

The encoding method didn't properly encode all value types. I used the example test cases from the RFC as tests and updated the encode method


jakubkoci (approved):
Looks good 👍

PR #161, opened 2021-01-13
update wallet error handling

TimoGlastra:
Update the wallet error handling to be a bit less repetitive

﻿Signed-off-by: Timo Glastra <timo@animo.id>

### DIFF
--- a/src/lib/protocols/credentials/__tests__/StubWallet.ts
+++ b/src/lib/protocols/credentials/__tests__/StubWallet.ts
@@ -3,13 +3,6 @@ import { Wallet, DidInfo } from '../../../wallet/Wallet';
 import { UnpackedMessageContext } from '../../../types';
 
 export class StubWallet implements Wallet {
-  private wh?: number | undefined;
-
-  public get walletHandle() {
-    if (!this.wh) throw new Error('Wallet not initialized');
-    return this.wh;
-  }
-
   public init(): Promise<void> {
     return Promise.resolve();
   }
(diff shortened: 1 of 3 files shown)


src/lib/protocols/credentials/__tests__/StubWallet.ts
```diff
@@ -4,6 +4,12 @@ import { UnpackedMessageContext } from '../../../types';
 
 export class StubWallet implements Wallet {
   private wh?: number | undefined;
+
+  public get walletHandle() {
```
jakubkoci, line 8: Could we keep it private?

src/lib/protocols/credentials/__tests__/StubWallet.ts
```diff
@@ -4,6 +4,12 @@ import { UnpackedMessageContext } from '../../../types';
 
 export class StubWallet implements Wallet {
   private wh?: number | undefined;
```
jakubkoci, line 6: Maybe we can also rename it to `_walletHandle` to keep consistency with other getters. But I don't have a strong opinion here 🤷‍♂️

jakubkoci (commented):
Nice refactor 👍  Just some details...

src/lib/protocols/credentials/__tests__/StubWallet.ts
TimoGlastra: I see the `SignatureDecoratorUtils.test.ts` uses it:

```ts
const verkey = await indy.createKey(wallet.walletHandle as number, { seed: seed1 });
```

The test compares the result from the wallet to a native indy call. So this woud need a refactor. It was already public. Or do you mean just the `StubWallet`?

TimoGlastra: I left `walletHandle` in `IndyWalet` public for now. but applied the rest of your feedback

src/lib/protocols/credentials/__tests__/StubWallet.ts
TimoGlastra: Updated

jakubkoci (approved):

PR #164, opened 2021-01-13
make events property instead of method

TimoGlastra:
Same as with other property changes. More natural usage when it is a property IMO

﻿Signed-off-by: Timo Glastra <timo@animo.id>

### DIFF
--- a/src/lib/modules/CredentialsModule.ts
+++ b/src/lib/modules/CredentialsModule.ts
@@ -29,6 +29,16 @@ export class CredentialsModule {
     this.messageSender = messageSender;
   }
 
+  /**
+   * Get the event emitter for the credential service. Will emit state changed events
+   * when the state of credential records changes.
+   *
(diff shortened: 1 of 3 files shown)


jakubkoci (approved):

src/lib/modules/CredentialsModule.ts
```diff
@@ -29,6 +29,16 @@ export class CredentialsModule {
     this.messageSender = messageSender;
   }
 
+  /**
+   * Get the event emitter for the credential service. Will emit state changed events
+   * when the state of credential records changes.
+   *
+   * @returns event emitter for proof related actions
```
jakubkoci, line 36: Just spot that you have `proof` in here, it should be `credential` I think.

jakubkoci: Shouldn't it be `proof related state changes`? 🤔

PR #165, opened 2021-01-13
refactor/credential changes

TimoGlastra:


src/lib/protocols/credentials/CredentialService.ts
```diff
@@ -309,7 +314,13 @@ export class CredentialService extends EventEmitter {
     const prevState = credential.state;
     credential.state = newState;
     await this.credentialRepository.update(credential);
-    this.emit(EventType.StateChanged, { credential, prevState });
+
+    const event: CredentialStateChangedEvent = {
+      credential: credential,
```
jakubkoci, line 319: Couldn't we have just `credential` instead of `credential: credential`? 🤔 

```
const event: CredentialStateChangedEvent = {
      credential,
      prevState,
}
```

jakubkoci (commented):
Just one nitpick :)

src/lib/protocols/credentials/CredentialService.ts
TimoGlastra:
```suggestion
      credentialRecord: credential,
```

TimoGlastra:
Now that we changed the service API to:

```ts
const { message, record } = this.xxxService.createXXX()
```

Maybe we should also change the event to be the same. So `record` instead of `credentialRecord`:

```ts
export interface CredentialStateChangedEvent {
  record: CredentialRecord;
  prevState: CredentialState;
}
```

You can merge this for now, but I'm curious to hear what you think

jakubkoci (approved):

jakubkoci:
> Now that we changed the service API to:
> 
> ```ts
> const { message, record } = this.xxxService.createXXX()
> ```
> 
> Maybe we should also change the event to be the same. So `record` instead of `credentialRecord`:
> 
> ```ts
> export interface CredentialStateChangedEvent {
>   record: CredentialRecord;
>   prevState: CredentialState;
> }
> 
> export interface XXXStateChangedEvent {
>   record: XXXRecord;
>   prevState: XXXState;
> }
> ```
> 
> You can merge this for now, but I'm curious to hear what you think

I agree 👍

PR #168, opened 2021-01-15
add codecov github action

TimoGlastra:


TimoGlastra:
`89.47%`

Not bad :smile:

jakubkoci (approved):
Vau. How did we get to that coverage percentage? I don't believe, or maybe I don't understand how it can be so high 😄

jakubkoci:
When I click on that Codecov report link I see this:

image

Is it ok?

TimoGlastra:
> Vau. How did we get to that coverage percentage?

Now that you mention it, what I think may be happening is that it only takes the files into account that are being tested. 

We should add: [`collectCoverageFrom`](https://jestjs.io/docs/en/configuration#collectcoveragefrom-array). I'll make an PR that fixes this

> Is it expected?

I think it is for this PR. Next PR will have the codecov in master and can use that as base. It has nothing to compare to yet

PR #170, opened 2021-01-29
present proof protocol

TimoGlastra:


TimoGlastra:
@jakubkoci this is ready for review now. Basic flow is tested (starting from proposal + starting from request) some advanced negotiation cases + unit tests are not yet but I think it would be good to get it in main.

TimoGlastra:
Tested for interop with both ACA-Py and Aries Framework .NET: https://github.com/hyperledger/aries-agent-test-harness/pull/158.

The interop testing revealed some issues that are addressed in #181, but after that we can issue, present and verify with both ACA-Py and .NET :)

src/lib/protocols/present-proof/ProofService.ts
```diff
@@ -0,0 +1,914 @@
...
+            `Actual '${attribute.encoded}'`
+        );
+      }
+    }
+
+    // TODO: pre verify proof json
+    // I'm not 100% sure how much indy does. Also if it checks whether the proof requests matches the proof
+    // @see https://github.com/hyperledger/aries-cloudagent-python/blob/master/aries_cloudagent/indy/sdk/verifier.py#L79-L164
+
+    const schemas = await this.buildSchemas(new Set(proof.identifiers.map(i => i.schemaId)));
```
jakubkoci, line 761: I would say this is about "getting" schemas rather than "building" 🤔  

To my mind, LedgerService could be extracted away from this service, and schemas passed as a method parameter. What do you think?

src/lib/protocols/present-proof/messages/PresentationMessage.ts
```diff
@@ -0,0 +1,68 @@
...
+   */
+  @Expose({ name: 'request-presentations~attach' })
+  @Type(() => Attachment)
+  @IsArray()
+  @ValidateNested({
+    each: true,
+  })
+  public attachments!: Attachment[];
+
+  public get indyProof(): IndyProof | null {
```
jakubkoci, line 56: This method introduces a dependency on Indy.  Not sure how to work around that, just saying 😄

src/lib/protocols/present-proof/messages/RequestPresentationMessage.ts
```diff
@@ -0,0 +1,71 @@
...
+   */
+  @Expose({ name: 'request-presentations~attach' })
+  @Type(() => Attachment)
+  @IsArray()
+  @ValidateNested({
+    each: true,
+  })
+  public attachments!: Attachment[];
+
+  public get indyProofRequest(): ProofRequest | null {
```
jakubkoci, line 57: I have the same concern I mentioned in a comment for `PresentationMessage`.

src/lib/protocols/present-proof/models/AttributeFilter.ts
```diff
@@ -0,0 +1,136 @@
...
+
+      case TransformationType.PLAIN_TO_CLASS:
+        const regex = new RegExp('^attr::([^:]+)::(value)$');
+
+        for (const [key, value] of Object.entries(attributeFilter)) {
+          const match = regex.exec(key);
+
+          if (match) {
+            const attributeValue = new AttributeValue({
+              name: match[1],
```
jakubkoci, line 119: Nitpick/Tip: If possible and make sense, I try to use array destructuring and name the element like `[,firstMatch]` instead of

jakubkoci (approved):
There is a lot of changes so I went through that quite briefly and just added few notes. We can address them in upcoming PRs, I don't want to block it :)  Very good job guys 👍

src/lib/__tests__/proofs.test.ts
```diff
@@ -0,0 +1,288 @@
...
+    // Faber accepts presentation
+    await faberAgent.proof.acceptPresentation(faberProofRecord.id);
+
+    // Alice waits till it receives presentation ack
+    aliceProofRecord = await waitForRecord(aliceAgent, {
+      threadId: aliceProofRecord.tags.threadId,
+      state: ProofState.Done,
+    });
+  });
+});
```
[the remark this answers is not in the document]
TimoGlastra: I think it is better to keep them separate. We're testing different things and are probably going to add a lot more tests in the future. I think we could create some helper methods that will set up a connection + create schema / cred def and issue a credential

src/lib/modules/ProofsModule.ts
```diff
@@ -0,0 +1,229 @@
...
+  private proofService: ProofService;
+  private connectionService: ConnectionService;
+  private messageSender: MessageSender;
+  private indy: Indy;
+
+  public constructor(
+    proofService: ProofService,
+    connectionService: ConnectionService,
+    messageSender: MessageSender,
+    indy: Indy
```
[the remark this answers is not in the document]
TimoGlastra: Agree. Will update in a future PR. I don't think it should all be added to wallet as some indy operations don't belong in the wallet. But maybe we can make an indy service or something that does the indy specific stuff.

For example ACA-Py has `IndyIssuer`, `IndyVerifier`, `IndyHolder` classes.

src/lib/__tests__/proofs.test.ts
```diff
@@ -0,0 +1,288 @@
...
+
+    logger.log('Alice waits for presentation request from Faber');
+    aliceProofRecord = await waitForRecord(aliceAgent, {
+      threadId: aliceProofRecord.tags.threadId,
+      state: ProofState.RequestReceived,
+    });
+
+    logger.log('Alice accepts presentation request from Faber');
+    const indyProofRequest = aliceProofRecord.requestMessage?.indyProofRequest;
+    const requestedCredentials = await aliceAgent.proof.getRequestedCredentialsForProofRequest(
```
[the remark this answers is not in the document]
TimoGlastra: No. Or maybe as a separate method. For auto accepting this is great, but let's say I have a mobile wallet with 10 credentials that could fulfil this request. I want the user / UI to handle which credentials to present in the proof.

src/lib/__tests__/proofs.test.ts
```diff
@@ -0,0 +1,288 @@
...
+        name: 'name',
+        restrictions: [
+          new AttributeFilter({
+            credentialDefinitionId: credDefId,
+          }),
+        ],
+      })
+    );
+
+    const predicates = new Map();
```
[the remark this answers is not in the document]
TimoGlastra: No I already regret doing that. It made the transformation easier, but makes usage a lot harder. I'll remove in an upcoming PR

src/lib/modules/ProofsModule.ts
```diff
@@ -0,0 +1,229 @@
...
+  }
+
+  /**
+   * Initiate a new presentation exchange by sending a presentation proposal message
+   * to the connection with the specified connection id.
+   *
+   * @param connectionId The connection to send the proof proposal to
+   * @param presentationProposal The presentation proposal to include in the message
+   * @param config Additional configuration to use for the proposal
+   * @returns Proof record associated with the sent proposal message
```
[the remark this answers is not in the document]
TimoGlastra: You mean like if this is an Holder / Verifier action? I agree!

src/lib/protocols/present-proof/ProofService.ts
TimoGlastra: Took this from .NET. Fine by me. Same as the others, will resolve in next PR. Too much changes already

src/lib/protocols/present-proof/messages/PresentationMessage.ts
TimoGlastra: Not necessarily. It creates a dependency on indy types. But with my new PR (#183) the compiled JS will have no reference to Indy at all. It's only for type context.

And as the v1 protocols are very indy specific I don't see the problem in adding indy specific utility methods. In v2 we'll do it different of course

src/lib/protocols/present-proof/ProofService.ts
TimoGlastra: I will rename from build to get, but leaving the abstraction out for now. That would mean both ProofsModule and PresentationHandler would need to implement that method

PR #171, opened 2021-01-31
refactor: Hide wallet handle inside a Wallet class

jakubkoci:
I recently complained in some PR that we have `walletHandle` attribute publicly accessible. I got an idea to use `createDid` wallet method in the test. What do you think?


TimoGlastra (approved):
Makes sense 👍

---

Unit tests should succeed ofcourse

PR #180, opened 2021-02-14
update record states to match rfc

### LINKED ISSUE #176: Use record state values that match RFC
Currently the state values that we store in records are capitalized and separated by underscores. For interop purposes we should just use the values as defined in the RFCs.

e.g. `PROPOSAL_RECEIVED` becomes `proposal-received`

Record to update (at least, may be more):
- [`ConnectionState.ts`](https://github.com/hyperledger/aries-framework-javascript/blob/d9fe683ab577b73fbf6333f3acc428bd5085a3ee/src/lib/protocols/connections/domain/ConnectionState.ts)
- [`CredentialState.ts.`](https://github.com/hyperledger/aries-framework-javascript/blob/d9fe683ab577b73fbf6333f3acc428bd5085a3ee/src/lib/protocols/credentials/CredentialState.ts)
- test files

TimoGlastra:
Fixes #176 

#170 already uses the rfc states so doesn't require any additional changes

### DIFF
--- a/src/lib/protocols/present-proof/__tests__/ProofState.test.ts
+++ b/src/lib/protocols/present-proof/__tests__/ProofState.test.ts
@@ -1,0 +1,13 @@
+import { ProofState } from '../ProofState';
+
+describe('ProofState', () => {
+  test('state matches Present Proof 1.0 (RFC 0037) state value', () => {
+    expect(ProofState.ProposalSent).toBe('proposal-sent');
+    expect(ProofState.ProposalReceived).toBe('proposal-received');
+    expect(ProofState.RequestSent).toBe('request-sent');
+    expect(ProofState.RequestReceived).toBe('request-received');
+    expect(ProofState.PresentationSent).toBe('presentation-sent');
+    expect(ProofState.PresentationReceived).toBe('presentation-received');
+    expect(ProofState.Done).toBe('done');
+  });
+});
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -19,7 +19,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Get docker cache
-        uses: satackey/action-docker-layer-caching@v0.0.5
+        uses: satackey/action-docker-layer-caching@v0.0.11
 
       - name: Start indy pool
         run: |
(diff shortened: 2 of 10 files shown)


src/lib/protocols/connections/ConnectionService.test.ts
```diff
@@ -275,17 +275,20 @@ describe('ConnectionService', () => {
       ConnectionState.Responded,
       ConnectionState.Complete,
     ];
-    test.each(invalidConnectionStates)('throws an error when connection state is %s and not INVITED', state => {
-      expect.assertions(1);
+    test.each(invalidConnectionStates)(
+      `throws an error when connection state is %s and not ${ConnectionState.Invited}`,
```
jakubkoci, line 279: I'm still not sure about using constants in the test message. I like it because it does reuse the enum value, on the other side if you have that value incorrect, the test doesn't reveal that.

For example, if I have `ConnectionState.Invited = 'requested'` then it could be confusing

jakubkoci (approved):

src/lib/protocols/connections/ConnectionService.test.ts
TimoGlastra: I agree! I'll make an additional PR to fix that

TimoGlastra: After thinking about this using the string value everywhere also doesn't seem right. I'll make a test per state that tests the string value. Then we can use the enum value everywhere else.

src/lib/protocols/present-proof/__tests__/ProofState.test.ts
```diff
@@ -0,0 +1,13 @@
...
+  test('state matches Present Proof 1.0 (RFC 0037) state value', () => {
+    expect(ProofState.ProposalSent).toBe('proposal-sent');
+    expect(ProofState.ProposalReceived).toBe('proposal-received');
+    expect(ProofState.RequestSent).toBe('request-sent');
+    expect(ProofState.RequestReceived).toBe('request-received');
+    expect(ProofState.PresentationSent).toBe('presentation-sent');
+    expect(ProofState.PresentationReceived).toBe('presentation-received');
+    expect(ProofState.Done).toBe('done');
+  });
+});
```
TimoGlastra, line 13: @jakubkoci I added tests for each of the state object to test for the string value. The other tests now use the enum value. This is to keep the tests separated

jakubkoci: Good 👍

PR #181, opened 2021-02-14
Improve issue credential consistency

TimoGlastra:
This PR tries to introduce some consistency between the protocol implementations. Connections + Issue Credential + Present Proof now all use the same API and flow.

Note: This is dependant on the present proof protocol and is therefore opened against the `feature/present-proof` branch. Either this one should be merged into feature/present-proof first and then both merged into main. Or first we get the present proof merged, then I'll rebase this one


TimoGlastra:
Edit: this has now become more of an random mix of fixes/improvements etc...

It also includes changes required for AATH interop (found quite some bugs from it!). So let's first merge the present proof features and then I'll make some additional PRs if you think this is too much

PR #186, opened 2021-02-18
add note about usage in react native

TimoGlastra:


jakubkoci (commented):
What about the trick with types you mentioned to me a few times. Shouldn't we add it here? Or it's a different topic?

jakubkoci (approved):

PR #189, opened 2021-02-20
changes based on pr feedback

TimoGlastra:
- remove indy from proofs module
- add role names to module methods
- rename `buildSchemas`, `buildCredentialDefinitions` to `getSchemas`, `getCredentialDefinitions` (not abstracted yet. That's a bigger change and will get a separate PR)
- Use records (json objects) instead of Maps to make API usage easier


jakubkoci (approved):
Nice improvements :)

PR #191, opened 2021-03-03
Colocate handlers and protocol implementations into modules

jakubkoci:
This is just the first and the largest step in task #192. It should be just about moving files almost without any code changes. I can find the description of the following steps I want to do in that issue.


jakubkoci:
@TimoGlastra If you don't have any critical showstopper, I would suggest merging this one and continue with eventual tweaks and updates in upcoming PRs.

TimoGlastra (approved):
A lot of changes so can't check it all, but the overall change looks good.

I liked the clear naming of protocol directories (`issue-credentials` vs `credentials`) but thinking of modules I agree it's better to group by functionality instead of the protocol names

jakubkoci:
I just created an issue and move the description of this PR there which I should've done it first.

PR #193, opened 2021-03-05
fix: return valid schema in create schema method

### LINKED ISSUE #146: Can't use schema and credential definition returned from register function
As the title suggests, we can't use the schema/cred def returned from the register functions directly for issuing. This means the following won't work:

```js
const [schemaId, ledgerSchema] = await agent1.ledger.registerCredentialSchema(schemaTemplate);
const [credDefId, ledgerCredDef] = await agent1.ledger.registerCredentialDefinition(definitionTemplate);
```

But this does:

```js
const [schemaId] = await agent.ledger.registerCredentialSchema(schemaTemplate);
const ledgerSchema = await agent.ledger.getSchema(schemaId);
const [credDefId] = await agent.ledger.registerCredentialDefinition(definitionTemplate);
const ledgerCredDef = await agent.ledger.getCredentialDefinition(credDefId);
```

We need to figure out why this is happening.

TimoGlastra:
- Add `seqNo` to returned schema after creating. This makes it usable out of the box, without making another ledger request. Taken from ACA-Py impl
- `getPublicDid()` -> `publicDid` (as a getter method)
- `createCredentialSchema` -> `createSchema` (more consistent with rest of codebase)
- `support_revocation` -> `supportRevocation`

Fixes #146


jakubkoci (approved):
Nice update, thanks @TimoGlastra 👍

PR #201, opened 2021-03-11
add logging interface with console and test logger

TimoGlastra:
- Add `ILogger` interface with multiple log levels
- Add simple `ConsoleLogger` logger
- Add `TestLogger` that is used for testing and samples using `tslog`
- Make logger a config parameter
- Make indy a config parameter (not really related to this change, but I already did the change ...)


TimoGlastra:
@icc-romeu -- could you please review and see if this matches your idea with a logger interface?

src/lib/logger/Logger.ts
```diff
@@ -0,0 +1,24 @@
...
+  logLevel: LogLevel;
+
+  test(message: string, data?: Record<string, any>): void;
+  trace(message: string, data?: Record<string, any>): void;
+  debug(message: string, data?: Record<string, any>): void;
+  info(message: string, data?: Record<string, any>): void;
+  warn(message: string, data?: Record<string, any>): void;
+  error(message: string, data?: Record<string, any>): void;
+  fatal(message: string, data?: Record<string, any>): void;
+}
```
TimoGlastra, line 24: Mainly the interface of this file

icc-romeu: Good work Timo! It's almost that! I would recommend also to have a "per section" logger. This can help "activate/deactivate" loggers per section. Say:

```typescript

/**
 * We define the logger sections
 */
export enum LogSection {
  default = 'DEFAULT',
  transports = 'TRANSPORTS',
  whatever = 'WHATEVER',
}

export interface ILogger {
  logLevel: LogLevel;
  /**
   * All log instances will have a "section" value. If the default logger will fallback to "default"
   */
  readonly section: LogSection;

 // Other stuff

  /**
   * This method returns an instance of the logger, with the "section" set to the parameter.
   * If the section already exists, the same logger should be returned.
   *
   * Section loggers are disabled by default
   *
   * @param section the section name
   */
  logger(section: LogSection): ILogger;

  /**
   * Enable this section logger
   *
   * @param section the section name
   */
  enableLogger(section: LogSection): void;

  /**
   * Disable the section logger
   *
   * @param section the section name
   */
  disableLogger(section: LogSection): void;
}

// Then when you use the logger...

const aLogger: ILogger = ...;

const sectionLogger = aLogger.logger(LogSection.transports);
aLogger.enableLogger(LogSection.transports);
sectionLogger.debug('Whatever'); // This prints something like "SECTION: Whatever"
aLogger.disableLogger(LogSection.transports);
sectionLogger.debug('Whatever'); // This does not print anything

```

This can help you enable/disable the logs for some sections of the framework.

On the implementation, I would use transports as winston does. This can help enable file, NewRelic, JSON, Database or console transports with the same instance of the logger. Or even disable a transport for a level or section. But this is far away from "console" basic transport so there is no need to implement this. Anyone can do it if needed with this interface. I think the basic transport will do👌

src/lib/logger/Logger.ts
```diff
@@ -0,0 +1,24 @@
...
+  trace = 1,
+  debug = 2,
+  info = 3,
+  warn = 4,
+  error = 5,
+  fatal = 6,
+  off = 7,
+}
+
+export interface ILogger {
```
jakubkoci, line 14: We could name it just as `Logger` without the `I`.

jakubkoci (changes requested):
Just a very small naming suggestion :) Otherwise looking forward to using it 👍

src/lib/logger/Logger.ts
TimoGlastra: Thanks for the feedback @icc-romeu!

> On the implementation, I would use transports as winston does. This can help enable file, NewRelic, JSON, Database or console transports with the same instance of the logger. Or even disable a transport for a level or section. But this is far away from "console" basic transport so there is no need to implement this. Anyone can do it if needed with this interface. I think the basic transport will do👌



I started out with all these things but then realized I was creating a new logger library. I wanted to keep it as simple as possible to start with and then expand over time.

> Good work Timo! It's almost that! I would recommend also to have a "per section" logger. This can help "activate/deactivate" loggers per section. Say:

My idea was to allow to create scoped sub loggers with a name. The logger implementation can then handle all the specifics.

TimoGlastra:
@jakubkoci I renamed the class.

@icc-romeu made some good points. I'd like to merge this version for now and extend the logger in the future to support multiple scopes/sections

src/lib/logger/ConsoleLogger.ts
```diff
@@ -0,0 +1,50 @@
...
+      data ? console.error('ERROR: ' + message, this.formatData(data)) : console.error('ERROR: ' + message);
+    }
+  }
+
+  public fatal(message: string, data?: Record<string, unknown>): void {
+    if (this.isEnabled(LogLevel.fatal)) {
+      data ? console.error('FATAL: ' + message, this.formatData(data)) : console.error('FATAL: ' + message);
+    }
+  }
+}
```
jakubkoci, line 50: Just some food for thought :) I like how the `TestLogger` moved the condition into one method. It would be nice to achieve something like that here

jakubkoci:
I just added another note, but it's definitely not bloker for merging this. However, tests are failing for some reason 🤷‍♂️ I've tried to re-run them which usually helps with this `Ledger not found` error

TimoGlastra:
Yeah I'm digging into that error now, would like to have it fixed

src/lib/logger/ConsoleLogger.ts
TimoGlastra: Agree 👍

TimoGlastra:
@jakubkoci updated `ConsoleLogger`. A bit more magic, with a bit less duplication

jakubkoci (approved):

TimoGlastra:
I updated the tsconfig target to `ES2019`. Node V12 supports this 100%, React Native environments also transpile. This allows for logging with very detailed tracing, which I think is very cool. This is the default logging for the mediator sapmles / the tests now:

image

jakubkoci:
> I updated the tsconfig target to `ES2019`. Node V12 supports this 100%, React Native environments also transpile. This allows for logging with very detailed tracing, which I think is very cool. This is the default logging for the mediator sapmles / the tests now:
> 

Vau, that looks really nice. What about Electron? :) Hopefully, it should not be a problem.

TimoGlastra:
> > I updated the tsconfig target to `ES2019`. Node V12 supports this 100%, React Native environments also transpile. This allows for logging with very detailed tracing, which I think is very cool. This is the default logging for the mediator sapmles / the tests now:
> 
> Vau, that looks really nice. What about Electron? :) Hopefully, it should not be a problem.

What version of electron are you using? Seems like it is supported according to [this SO answer](https://stackoverflow.com/a/57074772/10552895):
 
> Electron 5.0.7 uses Node v12.0.0 and Chromium 73 so its ES10/ES2019 compatible. Based on these tables.

jakubkoci:
I have `"electron": "^6.0.12",` so no problem.

PR #208, opened 2021-03-11
update dependencies, ledger improvements

TimoGlastra:
- remove await poll
- update dependencies
- improve ledger response handling, catch errors

I created 3 separate PRs but it became a hassle with trying to fix the `LedgerNotFound` during the tests.


TimoGlastra:
Allright with the recent fix where we can directly use the schema/cred def returned from the register method I removed all the places where schemas/cred def were directly created and then retrieved and now just used the returned object from register.

The only place where a schema/cred def is retrieved from the ledger after creation is the ledger tests and that's also the only place (it seems) that's randomly failing now. So I think it has to do with retrieving a schema /cred def from the ledger too quickly after creating it. I've added a 2second sleep and would like to see how that holds up for now. 

If it resolves the issue with failing tests, we can see to dive deeper and maybe properly fix it. But for now the sleep will do I think.

src/lib/modules/ledger/services/LedgerService.ts
```diff
@@ -182,6 +185,29 @@ export class LedgerService {
     }
   }
 
+  private async submitWriteRequest(request: LedgerRequest, signDid: string): Promise<LedgerWriteReplyResponse> {
+    request = await this.appendTaa(request);
+    request = await this.wallet.signRequest(signDid, request);
```
jakubkoci, line 190: Doesn't this mutate the `request` param? What about naming that like `requestWithTaa` and `signedRequestWithTaa`? Or do you think it's too much and chatty? :)

src/samples/__tests__/e2e.test.ts
```diff
@@ -139,19 +136,21 @@ class PollingInboundTransporter implements InboundTransporter {
   }
 
   private pollDownloadMessages(agent: Agent) {
-    poll(
-      async () => {
+    const sleep = (ms: number) => new Promise(res => setTimeout(res, ms));
```
jakubkoci, line 139: I see the same function in `src/lib/__tests__/helpers.ts` maybe we could re-use it

.eslintrc.js
```diff
@@ -2,7 +2,6 @@ module.exports = {
   parser: '@typescript-eslint/parser',
   extends: [
     'plugin:@typescript-eslint/recommended', // Uses the recommended rules from the @typescript-eslint/eslint-plugin
-    'prettier/@typescript-eslint', // Uses eslint-config-prettier to disable ESLint rules from @typescript-eslint/eslint-plugin that would conflict with prettier
```
jakubkoci, line 5: Isn't it good to have just Prettier caring about formatting and not eslint? I don't remember why I exactly added this line, but I'm pretty sure it had some reasons 😄 

Maybe it's outdated settings. I found just this https://robertcooper.me/post/using-eslint-and-prettier-in-a-typescript-project from April 25, 2020 also using the same settings

jakubkoci (changes requested):
I like the ledger update and refactor 👍   Just one question, one change request and one nitpick :)

.eslintrc.js
TimoGlastra: I changed the settings because of updated dependencies, but we need the prettier/eslint integration because otherwise we're getting conflicts. We don't want eslint to do certain formatting that is then changed by prettier. This solves that problem.

src/samples/__tests__/e2e.test.ts
TimoGlastra: I don't think it's a problem that samples depends on lib tests. Updated

TimoGlastra:
@jakubkoci updated the sleep part. I'm fine with the current prettier / eslint config, but if we want to tackle it, I want to do that in a separate PR.

src/lib/modules/ledger/services/LedgerService.ts
TimoGlastra: It reassigns. So yeah it overwrites the request param, but it doesn't change the value outside of the method

TimoGlastra:
Missed your taa comment, updated

jakubkoci (approved):

PR #213, opened 2021-03-24
Default to sending https didcomm message prefix

### LINKED ISSUE #121: Update the message type URIs from the old did:sov format to https format
We should update the message types according to this [commit](https://github.com/hyperledger/aries-rfcs/commit/6509b84abaf5760a8ba1744c8078d513f28456db) in Aries RFCS.

I would like also to raise the question here on how to version the messages in the future. I haven't thought about that yet, so I don't have any opinion right now.

Specs:
* [Aries RFC 0348: Transition Message Type to HTTPs](https://github.com/hyperledger/aries-rfcs/blob/master/features/0348-transition-msg-type-to-https/README.md)

TimoGlastra:
Still supports receiving the old "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec" prefix

https://github.com/hyperledger/aries-rfcs/blob/master/features/0348-transition-msg-type-to-https/README.md

Fixes #121

### DIFF
--- a/src/lib/__tests__/credentials.test.ts
+++ b/src/lib/__tests__/credentials.test.ts
@@ -145,10 +145,10 @@ describe('credentials', () => {
       id: expect.any(String),
       offerMessage: {
         '@id': expect.any(String),
-        '@type': 'did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/issue-credential/1.0/offer-credential',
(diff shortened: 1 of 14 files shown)


jakubkoci (approved):
That's great. Very elegant way how to support both versions :)

PR #215, opened 2021-03-25
fix: Add samples folder as test root

jakubkoci:
I'm sorry, I forgot to add `samples` as test root the tests there has been ignored.


TimoGlastra (approved):
Hmm it seems weird to test our samples and include that in our test coverage.  But that is probably easier addressed when we internalize transports

PR #216, opened 2021-03-26
fix: legacy did:sov prefix on invitation

TimoGlastra:
Invitations do not go trough the message receiver which means the prefix was not replaced and validation failed because it expected the https://didcomm.org prefix.

This add an `@Transform` to the type of the invitation message so it is replaced with the didcomm prefix. As it is the only out-of-band message, this seemed like the best approach


jakubkoci (approved):
👍

PR #218, opened 2021-03-30
Websockets

jakubkoci:
First working solution with WebSockets. There is definitely still some space for improvement. Take it as minimal mergeable PR :)

Each edge agent communicates with its own mediator via WebSockets. The communication from Alice to Bob goes through Bob's mediator via HTTP. Then bob gets the message from the mediator via WebSockets and vice versa.

I basically copy-pasted `e2e.test.ts` and `mediator.ts` files to `e2e.ws.test.ts` respective `mediator.ws.ts`. We could have just one mediator, maybe even one e2e test file, after we add multiple inbound transporters which I would like to do as the next step.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,14 +14,16 @@
     "format": "yarn prettier --write",
     "check-format": "yarn prettier --list-different",
     "test": "jest --verbose",
-    "dev": "ts-node-dev --respawn --transpile-only ./src/samples/mediator.ts",
-    "prod:start": "node ./build/samples/mediator.js",
+    "dev": "ts-node-dev --respawn --transpile-only ./samples/mediator-ws.ts",
+    "prod:start": "node ./build/samples/mediator-ws.js",
     "prod:build": "rm -rf build && yarn compile",
     "validate": "npm-run-all --parallel lint compile",
     "prepack": "rm -rf build && yarn compile",
     "prepare": "husky install"
   },
   "dependencies": {
+    "@types/socket.io": "^2.1.13",
+    "@types/socket.io-client": "^1.4.36",
     "bn.js": "^5.2.0",
     "buffer": "^6.0.3",
     "class-transformer": "^0.4.0",
@@ -29,6 +31,8 @@
     "events": "^3.3.0",
     "js-sha256": "^0.9.0",
     "reflect-metadata": "^0.1.13",
+    "socket.io": "2.4.1",
+    "socket.io-client": "2.4.0",
     "uuid": "^8.3.0"
   },
   "devDependencies": {
--- a/src/agent/Agent.ts
+++ b/src/agent/Agent.ts
@@ -30,6 +30,7 @@ import { ProofsModule } from '../modules/proofs/ProofsModule';
 import { RoutingModule } from '../modules/routing/RoutingModule';
 import { BasicMessagesModule } from '../modules/basic-messages/BasicMessagesModule';
 import { LedgerModule } from '../modules/ledger/LedgerModule';
+import { Transport, TransportService } from './TransportService';
 
 export class Agent {
   protected logger: Logger;
@@ -38,6 +39,7 @@ export class Agent {
   protected messageReceiver: MessageReceiver;
   protected dispatcher: Dispatcher;
   protected messageSender: MessageSender;
+  protected transportService: TransportService;
   protected connectionService: ConnectionService;
   protected proofService: ProofService;
   protected basicMessageService: BasicMessageService;
@@ -81,8 +83,9 @@ export class Agent {
     });
     this.wallet = new IndyWallet(this.agentConfig);
     const envelopeService = new EnvelopeService(this.wallet, this.agentConfig);
+    this.transportService = new TransportService();
 
-    this.messageSender = new MessageSender(envelopeService, outboundTransporter);
+    this.messageSender = new MessageSender(envelopeService, this.transportService, outboundTransporter);
     this.dispatcher = new Dispatcher(this.messageSender);
     this.inboundTransporter = inboundTransporter;
 
@@ -92,6 +95,7 @@ export class Agent {
     this.provisioningRepository = new Repository<ProvisioningRecord>(ProvisioningRecord, storageService);
     this.credentialRepository = new Repository<CredentialRecord>(CredentialRecord, storageService);
     this.proofRepository = new Repository<ProofRecord>(ProofRecord, storageService);
+
     this.provisioningService = new ProvisioningService(this.provisioningRepository, this.agentConfig);
     this.connectionService = new ConnectionService(this.wallet, this.agentConfig, this.connectionRepository);
     this.basicMessageService = new BasicMessageService(this.basicMessageRepository);
@@ -112,6 +116,7 @@ export class Agent {
     this.messageReceiver = new MessageReceiver(
       this.agentConfig,
       envelopeService,
+      this.transportService,
       this.connectionService,
       this.dispatcher
     );
@@ -147,8 +152,8 @@ export class Agent {
     return this.agentConfig.mediatorUrl;
   }
 
-  public async receiveMessage(inboundPackedMessage: unknown) {
-    return await this.messageReceiver.receiveMessage(inboundPackedMessage);
+  public async receiveMessage(inboundPackedMessage: unknown, transport?: Transport) {
+    return await this.messageReceiver.receiveMessage(inboundPackedMessage, transport);
   }
 
   public async closeAndDeleteWallet() {
@@ -178,6 +183,7 @@ export class Agent {
     this.routing = new RoutingModule(
       this.dispatcher,
       this.agentConfig,
+      this.transportService,
       this.providerRoutingService,
       this.provisioningService,
       this.messagePickupService,
--- a/src/agent/MessageReceiver.ts
+++ b/src/agent/MessageReceiver.ts
@@ -9,10 +9,12 @@ import { AgentMessage } from './AgentMessage';
 import { JsonTransformer } from '../utils/JsonTransformer';
 import { Logger } from '../logger';
 import { replaceLegacyDidSovPrefixOnMessage } from '../utils/messageType';
+import { Transport, TransportService } from './TransportService';
 
 class MessageReceiver {
   private config: AgentConfig;
   private envelopeService: EnvelopeService;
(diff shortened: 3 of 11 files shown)


package.json
```diff
@@ -14,21 +14,25 @@
...
+    "dev": "ts-node-dev --respawn --transpile-only ./samples/mediator-ws.ts",
+    "prod:start": "node ./build/samples/mediator-ws.js",
     "prod:build": "rm -rf build && yarn compile",
     "validate": "npm-run-all --parallel lint compile",
     "prepack": "rm -rf build && yarn compile",
     "prepare": "husky install"
   },
   "dependencies": {
+    "@types/socket.io": "^2.1.13",
+    "@types/socket.io-client": "^1.4.36",
```
TimoGlastra, line 26: Should be in `devDependencies`

src/agent/Agent.ts
```diff
@@ -38,6 +39,7 @@ export class Agent {
   protected messageReceiver: MessageReceiver;
   protected dispatcher: Dispatcher;
   protected messageSender: MessageSender;
+  protected transportService: TransportService;
```
TimoGlastra, line 42: I like the transport service

src/agent/MessageReceiver.ts
```diff
@@ -60,6 +64,12 @@ class MessageReceiver {
           `Inbound message 'sender_key' ${senderKey} is different from connection.theirKey ${connection.theirKey}`
         );
       }
+
+      if (connection) {
+        if (transport) {
+          this.transportService.saveTransport(connection.id, transport);
+        }
+      }
```
TimoGlastra, line 72:
```suggestion
      if (connection && transport) 
      	this.transportService.saveTransport(connection.id, transport);
      }
```

src/agent/MessageSender.ts
```diff
@@ -22,6 +29,12 @@ class MessageSender {
 
   public async sendMessage(outboundMessage: OutboundMessage): Promise<void> {
     const outboundPackage = await this.envelopeService.packMessage(outboundMessage);
+    const transport = this.transportService.getTransport(outboundMessage.connection.id);
+    if (transport) {
+      outboundPackage.transport = transport;
+    } else {
+      outboundPackage.transport = new HttpTransport(outboundMessage.endpoint);
```
TimoGlastra, line 36: This should look at the scheme of the endpoint instead of creating a hardcoded transport. Ideally the `outboundMessage.endpoint` is not set here yet. As we could then dynamically select the transport to use based on available endpoints of the the connection and available transports.

src/agent/MessageSender.ts
```diff
@@ -32,6 +45,12 @@ class MessageSender {
     outboundMessage.payload.setReturnRouting(ReturnRouteTypes.all);
 
     const outboundPackage = await this.envelopeService.packMessage(outboundMessage);
+    const transport = this.transportService.getTransport(outboundMessage.connection.id);
+    if (transport) {
+      outboundPackage.transport = transport;
+    } else {
+      outboundPackage.transport = new HttpTransport(outboundMessage.endpoint);
```
TimoGlastra, line 52: Same

src/modules/routing/RoutingModule.ts
```diff
@@ -120,4 +131,5 @@ interface MediatorConfiguration {
   verkey: Verkey;
   invitationUrl: string;
   alias?: string;
+  transport?: Transport;
```
TimoGlastra, line 134: I don't think we should save a transport with the mediator. We're moving more towards mediator coordination protocol in which case the flow probably looks something like this:

- Mediator creates invitation
- Mediatee receives invitation, sends request (probably HTTP)
- Mediatior sends response with DIDDoc containing both ws and http endpoint.
- We make a preference to use WS, but could also use HTTP. So it should be more dynamic I think.

src/agent/TransportService.ts
```diff
@@ -0,0 +1,41 @@
...
+
+interface TransportTable {
+  [connectionRecordId: string]: Transport;
+}
+
+type TransportType = 'ws' | 'http';
+
+export interface Transport {
+  type: TransportType;
+}
```
TimoGlastra, line 23: I think it would be nice to add a `canReply` method to this interface

src/agent/TransportService.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export class WebSocketTransport implements Transport {
+  public type: TransportType = 'ws';
+  public socket: Socket;
+
+  public constructor(socket: Socket) {
+    this.socket = socket;
+  }
+}
```
TimoGlastra, line 32: When the socket (http or ws) will be closed `canReply` can be set to false.

Also when we already replied with a message on the http channel we can also set `canReply` to false

src/agent/Agent.ts
```diff
@@ -147,8 +152,8 @@ export class Agent {
     return this.agentConfig.mediatorUrl;
   }
 
-  public async receiveMessage(inboundPackedMessage: unknown) {
-    return await this.messageReceiver.receiveMessage(inboundPackedMessage);
+  public async receiveMessage(inboundPackedMessage: unknown, transport?: Transport) {
+    return await this.messageReceiver.receiveMessage(inboundPackedMessage, transport);
```
TimoGlastra, line 156: I think it would be nice to add transport to the `InboundMessageContext`

src/agent/TransportService.ts
```diff
@@ -0,0 +1,41 @@
+import { Socket } from 'socket.io';
```
TimoGlastra, line 1: I've mentioned this before, but AFAIK socket.io uses non-standard communication protocols. This means this won't work with ACA-Py as a mediator.

I have objections to merging this with socket.io instead of a general WS server. If you can make it such that socket.io can communicate with non-socket.io agents that's also fine

TimoGlastra (changes requested):
Thanks for this Jakub! I think this is a long awaited PR :)

Most comments are thoughts and don't have to be fixed right now. I think a lot those would be resolved when multiple transports are added.

However the usage of socket.io is a blocker for me, and I don't think I can be convinced to approve this before we use a generic WS library.

src/agent/TransportService.ts
icc-romeu: As @TimoGlastra suggests, Javascript WebSocket works with ACA-py. I have tested this in my own code and works. https://developer.mozilla.org/en-US/docs/Web/API/WebSocket

jakubkoci:
Thanks @TimoGlastra and @JamesKEbert for the helpful discussion and explanation of the mediation protocol in our last call. I went through the code and specs again and I understand it a little bit more. I still have some questions but also some ideas about the implementation. 

First of all, I read the [Aries RFC 0025: DIDComm Transports](https://github.com/hyperledger/aries-rfcs/tree/master/features/0025-didcomm-transports) and it says:
* “POST requests are considered transmit only by default. No agent messages will be returned in the response. This behavior may be modified with additional signaling.”
* “Websockets are considered transmit only by default. Messages will only flow from the agent that opened the socket. This behavior may be modified with additional signaling.”

Do I understand correctly that “This behavior may be modified with additional signaling.” means [Aries RFC 0092: Transports Return Route](https://github.com/hyperledger/aries-rfcs/tree/master/features/0092-transport-return-route)?

Regarding the implementation, we should probably solve the problem with `sendAndReceiveMessage`  and `return_route` before we do the change from socket.io to WebSocket implementation.

I implemented the current solution in a way to achieve some sort of synchronicity to establish a connection with a mediator. I realized that we can achieve the same with the same or similar concept like `returnWhenIsConnected`. That would mean just create and send a message and wait for the rest to be handled by the framework.
```ts
// Simplified code
const connReq = await this.connectionService.createRequest(connection.id);
await this.messageSender.sendMessage(connReq);
await returnWhenIsConnected(conn)
```

This change will require some changes in `HttpOutboundTransporter` and `downloadMessages` implementation. Before that, I just want to be sure I understand the DIDComm Transports and `return_route` correctly.

If `return_route` is set, then a request message sender wants to receive a response message via the same transport channel as it sent the message. It means that the mediator will send an outbound message via HTTP response for HTTP request and via the same socket ID for WebSocket connection. If the return route is not set, then an outbound message should be send according to the information from a connection DIDDoc. An agent should decide outbound transport channel based on information from the DIDDoc (http endpoint or indication to store the message in case of `didcomm:transport/queue` endpoint) together with current transport state in transport service (connected socket ID in case of `ws` transport). 

An edge agent connected to a mediator can set `return_route` and get message response via HTTP response or as WebSocket message. If the edge agent doesn’t set `return_route`  the mediator just stores messages and the edge agent doesn’t get them until it sends `batchPickup` message with `return_route`.

Is my understanding correct? :)

src/modules/routing/RoutingModule.ts
jakubkoci: I see. The transport here is passed mainly to say what transport should be used for this step:
* Mediatee receives invitation, **sends request** (probably HTTP)

I assumed that it should be possible to do it either via HTTP or WS.

Maybe, we should select the transport channel based on the service endpoint in the invitation? I see that [0. Invitation to Connect](https://github.com/hyperledger/aries-rfcs/blob/master/features/0160-connection-protocol/README.md#0-invitation-to-connect) defines just one serviceEndpoint but [Aries RFC 0434: Out-of-Band Protocol 1.1](https://github.com/hyperledger/aries-rfcs/blob/master/features/0434-outofband/README.md#invitation-httpsdidcommorgout-of-bandverinvitation) shows an example with services as an array.

TimoGlastra: > Maybe, we should select the transport channel based on the service endpoint in the invitation?

I think for creating the connection with the mediator it should use the one from the invitation, just like we do for any other connection. After that it should use the transport(s) as defined in the received did doc

TimoGlastra:
> Do I understand correctly that “This behavior may be modified with additional signaling.” means [Aries RFC 0092: Transports Return Route](https://github.com/hyperledger/aries-rfcs/tree/master/features/0092-transport-return-route?rgh-link-date=2021-04-05T07%3A34%3A12Z)?

Yep!

> Is my understanding correct? :)

Yeah seems about right

TimoGlastra:
@jakubkoci can this be closed now that we have #256?

jakubkoci:
Replaced by #250

PR #224, opened 2021-04-04
style: use minimal formatting rules

TimoGlastra:
To keep some sanity I wanted to make the prettier / linting config for the mobile agent and AFJ the same. I made some changes to remove semicolons for example.

Accompanying PR for the mobile agent: https://github.com/hyperledger/aries-mobile-agent-react-native/pull/34


samples/__tests__/e2e.test.ts
```diff
@@ -25,150 +25,153 @@ const bobConfig: InitConfig = {
...
-  let aliceAtAliceBobId: string;
+  let aliceAgent: Agent
+  let bobAgent: Agent
+  let aliceAtAliceBobId: string
 
   afterAll(async () => {
-    (aliceAgent.inboundTransporter as PollingInboundTransporter).stop = true;
-    (bobAgent.inboundTransporter as PollingInboundTransporter).stop = true;
+    ;(aliceAgent.inboundTransporter as PollingInboundTransporter).stop = true
+    ;(bobAgent.inboundTransporter as PollingInboundTransporter).stop = true
```
jakubkoci, line 37: I don't understand why semicolons are at the start of the line here? 🤔

jakubkoci (changes requested):
It's a good idea to align it and use more default options. However, I would rather keep semicolons as it's also a default option and it could prevent weird behavior with TypeScript (see my other comment).

samples/__tests__/e2e.test.ts
TimoGlastra: That's because we use `()`. Bit weird...

TimoGlastra:
I personally like to not use semicolons as it adds characters that are not required. That said if semicolons are the default I'm fine with using them. My main goal is to share the same configuration between the projects.

As this also concerns the mobile agent, maybe @JamesKEbert @karimStekelenburg @DavidClawson can weigh in?

JamesKEbert:
I also personally lean towards not using semicolons as it increases readability from my point of view. However, I am not set on this as a requirement--I more fully second @TimoGlastra's goal to have the configurations be shared between projects.

karimStekelenburg:
My personal preference also is not to use semicolons. I agree with @JamesKEbert that it improves readability. Furthermore I don't see the benefits in using them, apart from the weird behaviour mentioned. That being said, I won't lose sleep over it and fully agree with the goal of a unified configuration between the projects :-)

TimoGlastra:
Ok seems like the overall preference is leaning towards no semicolons. Are you fine with this @jakubkoci?

jakubkoci (approved):
Yeah, I see. Personally, I don't have any preference from the readability point of view. I've worked with both without any problem, so if it's about that it doesn't make sense to waste our time by arguing about it here.

TimoGlastra:
> However, we should use [eslint.org/docs/rules/no-unexpected-multiline](https://eslint.org/docs/rules/no-unexpected-multiline) at least, if we're not using it already :)

I added `eslint:recommended` which includes this rule. It required only minimal change.

I see we use `"plugin:@typescript-eslint/recommended",` but don't use `"plugin:@typescript-eslint/recommended-requiring-type-checking"` which could add some valuable eslint checking. I will amend in an additional PR. E.g. `@typescript-eslint/await-thenable`: "Disallows awaiting a value that is not a Thenable" seems valuable

Issue #225, opened 2021-04-05
React Native record storage non-iterable async function generator issues


JamesKEbert:
Hey @TimoGlastra @jakubkoci @karimStekelenburg, we've been working on compiling AFJ for use in Aries Bifold. We've been running into an issue with the generator functions utilized in searching for records within a React Native context. In IndyStorageService.ts we get the following error in find(), findAll(), or findByQuery()  (which makes sense since they call the same generator):

```
[TypeError: Invalid attempt to iterate non-iterable instance.
In order to be iterable, non-array objects must have a [Symbol.iterator]() method.]
```

I've even tested creating an entirely new, basic generator (to just yield a number and log the number), and the same behavior is present. The previously generated aries-framework-javascript.tgz file we were using worked fine, and AFAIK the generator capabilities likely exist in this build too (although since it's not tied to a version and I didn't generate it I can't guarantee it).

If this is something you've run into before or have recommendations to pursue, it'd be appreciated! Thanks!

TimoGlastra:
I guess I'm to blame for this. I'm not 100% sure, but I recently changed the tsconfig target to `ES2019`. I read that it works in NodeJS and Electron, and I assumed it would work in RN, because it uses babel with babel env. Guess not....

Could you verify if changing the target back to `ES2015` works?

https://github.com/hyperledger/aries-framework-javascript/blame/b6111fbe97e99c5479e5d2ad101e62eaff79f7d8/tsconfig.json#L5

I made the change because higher target means better tracing with the new logger... But if it doesn't work in RN that's a problem of course

TimoGlastra:
Thinking of it now I remember that RN does transpile your code to the correct es version, just not your node modules

JamesKEbert:
Awesome, I agree with the thought process, and us targeting the newest version as is effective makes sense to me. I will change this locally and see if that fixes this issue. 

Yeah that sounds correct to me on your second comment.

JamesKEbert:
Yes, this fixed it! Thank you, definitely was confusing me there for a bit! I tested es2015 - es2019, and es2017 worked. Would you like me to make a PR in AFJ for this?

TimoGlastra:
Nice!

A PR would be great. Thanks

TimoGlastra:
I wish every issue was solved this quickly :)

PR #226, opened 2021-04-05
fix: revert to ES2017 to fix function generator issues in react native

JamesKEbert:
Fixes React Native record storage non-iterable async function generator issues due to TS configurations: #225 

ES2017 was the highest target specifiable to not have not-iterable errors in React Native.

Signed-off-by: James Ebert <jamesebert.k@gmail.com>


TimoGlastra (approved):
Thanks @JamesKEbert :)

PR #231, opened 2021-04-10
refactor: make a connection with mediator asynchronously

jakubkoci:
The previous implementation assumed that a connection with a mediator is made “synchronously “ via HTTP transport. By “synchronously” I mean that the provision method handled every step of making a connection and waiting for the response message assuming HTTP request-response communication. Now it should be independent on the transport layer. It just accepts the invitation and waits until the connection is made regardless of how the messages are exchanged.

I needed to do the following changes:
* Add `return_route` parameter to outbound when there is no inbound connection (there is no service with endpoint other than `didcomm:transport/queue`)
* Add `keylist-update-response` message and handler, just with naive implementation.
* Extract `outboundTransporter` as direct dependency in Agent’s `constructor`.

Next steps on the way to support WebSockets transport I would like to do:
* Extract also `inboundTransporter` from Agent’s `constructor`. (I can perhaps do it as part of this PR, but the changes started growing and I wanted to share what I have now.)
* Take outbound endpoint from `DidDoc`.
* Add WebSocket transporters.
* Add TransportService to keep info about current transport channels.
* Allow to set more inbound and outbound transporters.

Next steps I see we will need to do someday in near future but not necessarily before or right after adding the WebSockets:
* Update download messages functionality to process `batch` message and to remove `sendAndReceiveMessage` that is still implemented in the old way relaying on synchronicity of HTTP request-response.

BREAKING CHANGE: extracts outbound transporter from `Agent`'s constructor.


samples/__tests__/e2e.test.ts
```diff
@@ -159,22 +159,42 @@ class PollingInboundTransporter implements InboundTransporter {
...
-    testLogger.test(`Sending outbound message to connection ${outboundPackage.connection.id}`, outboundPackage.payload)
+    logger.test(`Sending outbound message to connection ${outboundPackage.connection.id}`, outboundPackage.payload)
 
     if (receiveReply) {
       const response = await post(`${endpoint}`, JSON.stringify(payload))
+      logger.test(`received response:\n ${response}`)
       const wireMessage = JSON.parse(response)
-      testLogger.test('received response', wireMessage)
-      return wireMessage
+      this.agent.receiveMessage(wireMessage)
```
TimoGlastra, line 180: We had this wrong before you made the changes, but as you're already updating this code. receiveReply means we MAY receive a reply. We're currently just parsing it without checking if we received a reply.

src/agent/MessageSender.ts
```diff
@@ -21,18 +24,26 @@ class MessageSender {
...
   ): Promise<InboundMessageContext<T>> {
+    if (!this.outboundTransporter) {
+      throw new Error('Agent has no outbound transporter!')
+    }
+
     outboundMessage.payload.setReturnRouting(ReturnRouteTypes.all)
 
     const outboundPackage = await this.envelopeService.packMessage(outboundMessage)
-    const inboundPackedMessage = await this.outboundTransporter.sendMessage(outboundPackage, true)
+    const inboundPackedMessage = await this.outboundTransporter.sendAndReceiveMessage(outboundPackage)
```
TimoGlastra, line 46: I'm not fully understanding. Why did you add a `sendAndReceiveMessage` to the outbound transporter?

src/modules/routing/RoutingModule.ts
```diff
@@ -49,18 +46,14 @@ export class RoutingModule {
...
-        ConnectionResponseMessage
-      )
-      await this.connectionService.processResponse(connectionResponse)
-      const { message: trustPing } = await this.connectionService.createTrustPing(connectionRecord.id)
-      await this.messageSender.sendMessage(createOutboundMessage(connectionRecord, trustPing))
+      const connectionRecord = await this.connectionService.processInvitation(mediatorInvitation, { alias })
+      const { message: connectionRequest } = await this.connectionService.createRequest(connectionRecord.id)
+
+      const outboundMessage = createOutboundMessage(connectionRecord, connectionRequest, connectionRecord.invitation)
+      outboundMessage.payload.setReturnRouting(ReturnRouteTypes.all)
```
TimoGlastra, line 53: You're already setting return routing if the connection has no inbound endpoint. this shouldn't be needed right?

src/modules/routing/handlers/KeylistUpdateResponseHandler.ts
```diff
@@ -0,0 +1,10 @@
+import { Handler, HandlerInboundMessage } from '../../../agent/Handler'
+import { KeylistUpdateResponseMessage } from '../messages'
+
+export class KeylistUpdateResponseHandler implements Handler {
+  public supportedMessages = [KeylistUpdateResponseMessage]
+
+  public async handle(messageContext: HandlerInboundMessage<KeylistUpdateResponseHandler>) {
+    // TODO It should handle the response when agent calls `await this.consumerRoutingService.createRoute(connectionRecord.verkey)` and notify about the result.
```
TimoGlastra, line 8: Why aren't we doing anything with the response? Are you going to address this in a separate PR?

src/modules/routing/messages/KeylistUpdateResponseMessage.ts
```diff
@@ -0,0 +1,62 @@
...
+  @IsArray()
+  @ValidateNested()
+  public updated!: KeylistUpdated[]
+}
+
+export enum KeylistUpdateResult {
+  client_error = 'client_error',
+  server_error = 'server_error',
+  no_change = 'no_change',
+  success = 'success',
```
TimoGlastra, line 43: I would prefer to use PascalCase for enum names

src/modules/routing/messages/KeylistUpdateResponseMessage.ts
```diff
@@ -0,0 +1,62 @@
...
+export class KeylistUpdated {
+  public constructor(options: { recipientKey: Verkey; action: KeylistUpdateAction }) {
+    if (options) {
+      this.recipientKey = options.recipientKey
+      this.action = options.action
+    }
+  }
+
+  @IsString()
+  public recipientKey!: Verkey
```
TimoGlastra, line 55: You're exposing the wrong key here. The RFC uses snake_case

```suggestion
  @Expose({ "name": "recipient_key" })
  public recipientKey!: Verkey
```

src/modules/routing/messages/RoutingMessageType.ts
```diff
@@ -1,6 +1,7 @@
 export enum RoutingMessageType {
   // TODO: add other messages from mediator coordination protocol
   KeylistUpdate = 'https://didcomm.org/coordinatemediation/1.0/keylist_update',
```
TimoGlastra, line 3: This is still snake_case. the PR was recently updated to use kebab-case

```suggestion
  KeylistUpdate = 'https://didcomm.org/coordinatemediation/1.0/keylist-update',
```

src/modules/routing/services/ProviderRoutingService.ts
```diff
@@ -12,11 +18,15 @@ export interface RoutingTable {
...
-  /**
-   * @todo use connection from message context
-   */
-  public updateRoutes(messageContext: InboundMessageContext<KeylistUpdateMessage>, connection: ConnectionRecord) {
-    const { message } = messageContext
+  public updateRoutes(messageContext: InboundMessageContext<KeylistUpdateMessage>): KeylistUpdateResponseMessage {
+    const { connection, message } = messageContext
+
+    if (!connection) {
+      // TODO We could eventually remove this check if we do it at some higher level where we create messageContext that must have a connection.
```
TimoGlastra, line 25: Yup would be nice to be able to declare if a connection is required for a handler or something...

TimoGlastra (changes requested):
I like these changes @jakubkoci! :)

The only thing I'm not sure about is the addition of `sendAndReceiveMessage` to the `OutboundTransporter` interface. Maybe you can clarify a bit on that

src/modules/routing/handlers/KeylistUpdateResponseHandler.ts
jakubkoci: We think should do it in separate PR which would also add the step to wait for the response after calling `createRoute`. I didn't plan to add this at first, but I had to when I set `return_route` and update HTTP inbound transporter. I wasn't 100 % sure what to do here and it seemed kind of out of the scope of this PR.

I created #234 to address this.

src/modules/routing/messages/KeylistUpdateResponseMessage.ts
jakubkoci: Sure :)

src/modules/routing/messages/KeylistUpdateResponseMessage.ts
jakubkoci: Good catch 👍

src/modules/routing/messages/RoutingMessageType.ts
jakubkoci: Updated.

samples/__tests__/e2e.test.ts
jakubkoci: Yeah, good point. Updated.

src/agent/MessageSender.ts
jakubkoci: Good question. I wanted to preserve the current behavior of `await agent.routing.downloadMessages()`. 

First, I thought that using `sendMessage(outbaoundPackage, true)` would be good enough. The problem is that in HTTP outbound we now pass the response immediately to the agent calling `this.agent.receiveMessage(wireMessage)`. It means that the `batch` response is not processed inside `downloadMessages` anymore. 

This is related to the problem we discussed in our last AFJ call about dispatching messages to the agent from a message handler. I listed as last point in the description:
"Update download messages functionality to process batch message and to remove sendAndReceiveMessage that is still implemented in the old way relaying on synchronicity of HTTP request-response."

I'm thinking if we could implement the solution as part of this PR. I usually tend to work in smaller batches. Maybe prioritize it and do it before WebSockets. Maybe it's not so big change 🤔

src/modules/routing/RoutingModule.ts
jakubkoci: I had to move it into `Dispatcher`, where is not called with this manual invitation acceptance process, because when `return_route` is set, we don't call `sendMessage` where I had it previously in #218

jakubkoci:
Thanks, @TimoGlastra for the good suggestions. I see we have 3 different areas to address (ordered by priority from my perspective):
1. `sendAndReceiveMessage`
2. calling `outboundMessage.payload.setReturnRouting(ReturnRouteTypes.all)` at two places
3. missing implementation of `KeylistUpdateResponseHandler`

I don't think we have to solve them all here. I'm quite curious about the solution to the 1. issue so I'm going to look at it briefly at least, considering prioritizing that.

jakubkoci:
I added an event emitter to Agent and Batch message handler to emit incoming messages to Agent. That allowed to remove `sendAndReceiveMessage`.

I'm not saying it's perfect, but I'm quite happy with this for now. What do you think @TimoGlastra?

jakubkoci:
I updated the PR name according to conventional commits. I guess this also brings some breaking change in agent API which I expressed in the description of the PR. That could go into the commit message when merging.

src/modules/routing/handlers/BatchHandler.ts
```diff
@@ -0,0 +1,26 @@
...
+    if (!messageContext.connection) {
+      throw new Error(`Connection for verkey ${messageContext.recipientVerkey} not found!`)
+    }
+
+    const { message } = messageContext
+    const forwardedMessages = message.messages
+
+    forwardedMessages.forEach((message) => {
+      this.eventEmitter.emit('agentMessage', message.message)
+    })
```
TimoGlastra, line 24: Do you think it will cause problems that all messages are emitted at the same time? E.g. If we receive 10 messages, they all get processed at the same time, meaning message 10 could be processed before message 5. It could also mean we fetch a record, another received message updated the record in the meantime and then we overwrite the new data.


I'm not sure if we want to take such things into account. A message queue could solve this, but  being able to handle message in parallel would be nice. ACA-Py handles this by fetching a record 'for-update' which means it'll lock other sessions from writing until you are done. Reading is always possible.

TimoGlastra (approved):
Nice! I like it.

With the addition of the new event emitter I think it'll be good to look at a global event emitter rather sooner than later

src/modules/routing/handlers/BatchHandler.ts
jakubkoci: Yeah. I'm afraid that the problem can occur. My first idea was also a message queue. You're right that it could be convenient to handle messages in parallel in some cases. I'm surprised that ACA-Py decided to go with locking of objects which sounds quite sophisticated to me, but maybe it's the only option 🤷‍♂️

jakubkoci:
> With the addition of the new event emitter I think it'll be good to look at a global event emitter rather sooner than later

Agree. It could solve even more problems like:
* dispatching incoming messages (as we're already using it for here)
* independent communication between modules
* notification about `createRoute` result #234

PR #233, opened 2021-04-10
support newer did-communication service type

TimoGlastra:
This adds support for the newer `did-communication` service type in did docs according to [RFC 0076](https://github.com/hyperledger/aries-rfcs/blob/master/features/0067-didcomm-diddoc-conventions/README.md#service-conventions)

IndyAgent services are still supported and we now send out both. This means it is backwards compatible in both ways (receiving and sending diddocs)


src/modules/connections/models/did/DidDoc.ts
```diff
@@ -65,4 +65,18 @@ export class DidDoc {
   public getServicesByClassType<S extends Service = Service>(classType: new (...args: never[]) => S): S[] {
     return this.service.filter((service) => service instanceof classType) as S[]
   }
+
+  /**
+   * Get all DIDComm services ordered by priority descending. This means the highest
+   * priority will be the first entry.
+   */
+  public get didCommServices(): Array<IndyAgentService | DIDCommService> {
+    const didCommServiceClasses = [IndyAgentService.type, DIDCommService.type]
```
jakubkoci, line 74: I would rather name it `didCommServiceTypes` or something like that. When I see classes I expect the class type itself and not its `type` attribute, something like when you call `instanceof SomeClass`.

jakubkoci: I found `serviceTypes` exported from `src/modules/connections/models/did/service/index.ts` Shouldn't we reuse it? 🤔

jakubkoci (changes requested):
Happy to see we will be up to date 👍  We should just align naming conventions.

src/agent/__tests__/AgentConfig.test.ts
```diff
@@ -1,6 +1,6 @@
 import type Indy from 'indy-sdk'
 import { getMockConnection } from '../../modules/connections/__tests__/ConnectionService.test'
-import { DidDoc, IndyAgentService } from '../../modules/connections'
+import { DIDCommService, DidDoc } from '../../modules/connections'
```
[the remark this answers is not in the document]
TimoGlastra: I agree!

src/modules/connections/models/did/DidDoc.ts
TimoGlastra: Ah that's still an old name. It were classes before, but I changed it to types. 

`serviceTypes` is all service types, not just the didcomm ones. I.e. if we add identity hub support it will be a new service type, but not a new didcomm service type

src/agent/__tests__/AgentConfig.test.ts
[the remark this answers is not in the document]
TimoGlastra: After thinking about this a bit more. `DIDComm` makes sense in this case, as it's the official name for `DIDComm`

TimoGlastra:
@jakubkoci I updated the PR according to your feedback

jakubkoci (approved):
👍

PR #236, opened 2021-04-11
chore: add release-it setup

TimoGlastra:
This adds automatic releasing with release-it. It doesn't yet use conventional commits, I'd like to do this at a later stage.

- On every push to main a new `unstable` version will be released. This is for ease of use with developing now that the API is still unstable
- A separate workflow dispatch is added to manually release a new version. This requires an input of `patch`, `minor`, `major` to specify the version.

There were some problems with the format of the first release, so I reverted that. We can let release-it make the first release

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -5,6 +5,12 @@ on:
     branches: [main]
   push:
     branches: [main]
+  workflow_dispatch:
+    inputs:
+      releaseType:
+        description: The type of release. Should be one of "major", "minor", "patch"
+        required: true
+        default: 'patch'
 
 env:
   TEST_AGENT_PUBLIC_DID_SEED: 000000000000000000000000Trustee9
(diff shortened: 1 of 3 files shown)


TimoGlastra:
OK -- This can be merged now @jakubkoci. The NPM token is set. If we merge it should automagically release to NPM.

TimoGlastra:
@jakubkoci could you merge this?

I removed the package from NPM so it's up for grabs again, would like to publish so it can't be taken by anyone else. In addition, I may have made some errors in the actions file, so then I can sort that out

jakubkoci (approved):

.github/workflows/continuous-integration.yml
```diff
@@ -69,3 +75,65 @@ jobs:
...
+          node-version: '12.x'
+
+      - name: Install dependencies
+        run: yarn install --ignore-optional
+
+      - name: Release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: yarn release ${{ github.event.inputs.releaseType }}
```
jakubkoci, line 139: It would nice to not have so much duplication in stable-release and unstable-release, but that's probably how it's done in GH Actions, right? :)

TimoGlastra: It's already on my todo list!

I just discovered you can do `if` statement per step, not per job. So that'd mean only the last step would be different

TimoGlastra:
Darn it! I messed it up. New PR incoming....

PR #245, opened 2021-04-15
refactor: extract inbound transporter from agent constructor

jakubkoci:
It aligns API with the setting of the outbound transporter. It's also better for the cases when a developer wants to pass messages into the agent by themself.

BREAKING CHANGE

### DIFF
--- a/src/agent/Agent.ts
+++ b/src/agent/Agent.ts
@@ -40,7 +40,7 @@ export class Agent {
   protected messageReceiver: MessageReceiver
   protected dispatcher: Dispatcher
   protected messageSender: MessageSender
-  public inboundTransporter?: InboundTransporter
+  public readonly inboundTransporter?: InboundTransporter
 
   protected connectionService: ConnectionService
   protected proofService: ProofService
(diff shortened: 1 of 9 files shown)


src/agent/Agent.ts
```diff
@@ -40,6 +40,8 @@ export class Agent {
   protected messageReceiver: MessageReceiver
   protected dispatcher: Dispatcher
   protected messageSender: MessageSender
+  public inboundTransporter?: InboundTransporter
```
TimoGlastra, line 43:
```suggestion
  public readonly inboundTransporter?: InboundTransporter
```

src/agent/Agent.ts
```diff
@@ -147,7 +146,9 @@ export class Agent {
       })
     }
 
-    return this.inboundTransporter.start(this)
+    if (this.inboundTransporter) {
+      await this.inboundTransporter.start(this)
+    }
```
TimoGlastra, line 151: Maybe we should log a warning for both the inbound/outbound transporter if they are not set at initialization yet. Just for information

TimoGlastra (approved):

TimoGlastra:
Please merge my other PR #244 first. Otherwise I need to do quite some manual work again

src/agent/Agent.ts
jakubkoci: Hmm, actually I can imagine it's not a problem not having an inbound transported set. But at least the info could be good. For example, I run AFJ as a mediator with Nest.js framework and it's quite convenient I don't have to set inbound transport but I handle it by myself.

Issue #187, opened 2021-02-19
Credentials have no easy way to access its attributes


TimoGlastra:
Currently when you have a credential record it is not easy to access its attributes. You need to either:

- parse the preview from the offer message
- parse the raw values from the credential message
- retrieve the credential from indy storage

Aries Framework .NET just stores the attributes in the record. I think we should do the same

jakubkoci:
I'm still curious if it's possible to utilize `credentialPreview` for that

TimoGlastra:
I think that's doable. We can add a connivence method to the credential record that returns the credential preview (or something else based on the credential preview)

jakubkoci:
My concern is security. I'm not sure if the user can rely on data in `credentialPreview`, because it could differ from data in actual `credentialOffer/credenital`.

TimoGlastra:
Well it should be trustable. If the issuer makes an offer, the holder requests it and then issues a different credential it should throw an error and not acknowledge it.

If we don't reject that at the moment, we should add a check for it.

jakubkoci:
Hmm. I assumed that libindy takes care of consistency between credOffer, credReq and cred, therefore we wouldn't need to validate if the received credential matches the offer. Then we just would need to validate `credentialPreview` against `credentialOffer` and stores the `credentialPreview` as attribute in `credentialRecord`

TimoGlastra:
How to format credentials: 

```jsonc
{
	// metadata contains credentials format specific stuff
	"metadata": {
		"credentialDefinitionId": "creddefid"
	}
	"claims": {
		"name": "name"
	}
}
```

I'll add a few utility methods such as returning the attributes in a properly cased or returning them alphabetically

@JamesKEbert thoughts?

JamesKEbert:
I think as you've proposed is great. 👍

PR #253, opened 2021-04-29
feat: automatic transformation of record classes

### LINKED ISSUE #106: Record classes should be transformed using class-transformer
All records now use custom serialize functions that don't integrate well with classes decorated using `class-transformer`. We should also transform record classes using class transformer to make working with didcomm messages inside records easier

TimoGlastra:
Transform record classes and its sub objects using class transformer.

Fixes #106

Was sort of a blocker for some of the work I wanted to do for #187, and it was long overdue this was fixed.


jakubkoci (approved):
This is great we finally have it align. So pleasure to review this and see how we're getting rid of those ugly conditional transformations 😄

PR #254, opened 2021-05-01
feat: add credential info to access attributes

### LINKED ISSUE #187: Credentials have no easy way to access its attributes

### LINKED ISSUE #190: Validate received credential against offer
When receiving a credential we should check whether the data inside the credential matches the data inside the credential offer.

Not doing this means we can accept an offer, while the credential being issued contains different attributes from the credential preview in the credential offer.

Related to #187

TimoGlastra:
Fixes #190
Fixes #187

Example:

```ts
// credentialRecord.getCredentialInfo()
{
  claims: {
    name: 'Timo',
    date_of_birth: '1998-07-29',
    'country-of-residence': 'The Netherlands',
    'street name': 'Test street',
    age: '22',
  },
  metadata: {
    credentialDefinitionId: 'Th7MpTaRZVRYnPiabds81Y:3:CL:17:TAG',
    schemaId: 'TL1EaPFCZ8Si5aUrqScBDt:2:test-schema-1599055118161:1.0',
  },
}
```

```ts
// credentialInfo.getFormattedClaims()
{
  Age: '22',
  'Country Of Residence': 'The Netherlands',
  'Date Of Birth': '1998-07-29',
  'Street Name': 'Test street',
  Name: 'Timo',
}
```

```ts
const credentialInfo = credentialRecord.getCredentialInfo()
const claims = credentialInfo.claims
const formattedClaims = credentialInfo.getFormattedClaims()
```


TimoGlastra:
This is dependant on #253, so once that's merge I'll rebase

jakubkoci:
I like the addition of credential preview and metadata but to be honest I don't think it's necessary to do the transformation from the array or the formatting operations. That should be the responsibility of the UI layer and not the framework.

From my experience from UI, you do the iteration through attributes over an array anyway:

```
Object.keys(attributes).map(key => <Item value={attributes[key]} />
```

jakubkoci (commented):
I like the addition of credential preview and metadata but to be honest I don't think it's necessary to do the transformation from the array or the formatting operations. That should be the responsibility of the UI layer and not the framework. I realize that you put some effort into the implementation of these functions but maybe we could rather move them to Aries Bifold. What do you think?

From my experience from UI, you do the iteration through attributes over an array anyway:

```js
Object.keys(attributes).map(key => <Item label={key} value={attributes[key]} />
```

Then it's more convenient to write:
```js
attributes.map(attr => <Item label={attr.key} value={attr.value} />
```

TimoGlastra:
I'm fine with removing the formatting stuff, not so making it an array.

An array works well for one level of nesting, not so for multiple levels (IMO). Besides the natural structure of a credential is dict like, not an array. Making it an array would tailor the structure to the UI (which is the responsibility of the UI layer as you said)

Are you okay with that?

jakubkoci:
Yeah, of course, my point wasn't about making it array because of convenience for UI. But it makes sense to have the claims as a dict because it aligns more with the W3C spec 👍  (if I understand it correctly 🤷‍♂️  😄 )

JamesKEbert:
Thanks for your efforts here @TimoGlastra! I agree with you that keeping the existing dictionary structure allows for more flexibility with claim formats.
I also initially agree with @jakubkoci on not performing formatting at the framework layer. 
Also, is there overlap we want to consider in terms of internationalization for future work here? Some relevant conversation in the [Aries RFCs](https://github.com/hyperledger/aries-rfcs/issues/630). Likely a separate issue, but I figure potentially worth mentioning.

TimoGlastra:
OK. Thanks for the feedback both. Updated and ready to merge now

src/modules/credentials/services/CredentialService.ts
```diff
@@ -543,11 +571,23 @@ export class CredentialService extends EventEmitter {
       )
     }
 
+    // Assert the values in the received credential match the values
+    // that were negotiated in the credential exchange
+    // TODO: Ideally we don't throw here, but instead store that it's not equal.
+    // the credential may still have value, and we could just respond with an ack
+    // status of fail
+    if (credentialRecord.credentialAttributesValues) {
```
jakubkoci, line 579: It seems we could shorten `credentialAttributesValues` just to `credentialAttributes`. Does it contain a list of `CredentialPreviewAttribute` right? I thought that it contains just values from credential preview attributes but the call `convertAttributesToValues(attributesValues)` confused me a little bit. But maybe I'm missing something 🤷‍♂️

jakubkoci (approved):
Sorry I wanted to go through that yet again after rebase. Overall good, just one non-blocking note about the naming.

src/modules/credentials/services/CredentialService.ts
TimoGlastra: If we only store the values we lose the mime-types. These are not stored by indy. So you'd need to extract them from one of the exchanged messages.

I think it's a burden the mime-types are not encoded in the values itself like `data:image/png;base64,xxxxxxxx`, but until that's the case (which is the pattern I've seen for W3C credentials) we need to store the preview attributes.

TimoGlastra: But `credentialAttributes` sounds fine. Will update

PR #255, opened 2021-05-01
feat: add internal http outbound transporter

TimoGlastra:
This is a first attempt at internalizing the http transport. Only the outbound for now, as it was the easiest. Made two http post implementation so it will work in both RN and Node.JS without having to add another dependency.

You can import it and works out of the box:

```ts
import { Agent, HttpOutboundTransporter } from 'aries-framework'

const agent = new Agent(config)
const httpOutboundTransport = new HttpOutboundTransporter(agent)
agent.setOutboundTransporter(httpOutboundTransport)
```

I made a start at `supportedSchemes`. It is not checked yet, but this allows us to determine the transport to use for a specific message when we add support for multiple transport. That should be a fairly small addition once this is merged. 

Also added an option to specify the didcomm mime type to use. This differs for AIP1.0/AIP2.0.

What do you think?


JamesKEbert:
Are we against adding another dependency? Such as Axios? It could potentially reduce complexity? I'm not in strong favor of this, just would be interested in hearing more of your thoughts!

TimoGlastra:
Not necessarily against it. Just don't want to introduce dependencies for things that are not that hard to implement ourselves.

I did think about adding axios, but noticed it wasn't a lot of work to do myself. I can be convinced otherwise however :)

TimoGlastra:
use node-fetch for nodejs environments

TimoGlastra:
updated to use `node-fetch`. Should be good to go now

jakubkoci (approved):
👍

Issue #250, opened 2021-04-23
Add support for WebSockets


jakubkoci:
I needed to do the following changes (solved by #231):
* ~~Add `return_route` parameter to outbound when there is no inbound connection (there is no service with endpoint other than `didcomm:transport/queue`)~~
* ~~Add `keylist-update-response` message and handler, just with naive implementation.~~
* ~~Extract `outboundTransporter` as direct dependency in Agent’s `constructor`.~~

Next steps on the way to support WebSockets transport I would like to do:
* ~~Extract also `inboundTransporter` from Agent’s `constructor`. (I can perhaps do it as part of this PR, but the changes started growing and I wanted to share what I have now.)~~ (solved by #245)
* ~~Take outbound endpoint from `DidDoc`.~~
* ~~Add WebSocket transporters.~~
* ~~Add TransportService to keep info about current transport channels.~~
* Allow to set more inbound and outbound transporters. (see #268)

Next steps I see we will need to do someday in near future but not necessarily before or right after adding the WebSockets (solved by #231):
* ~~Update download messages functionality to process `batch` message and to remove `sendAndReceiveMessage` that is still implemented in the old way relaying on synchronicity of HTTP request-response.~~

jakubkoci:
@TimoGlastra @JamesKEbert Is there any specification about how a mediation recipient should make a connection with a mediator? I'm asking mainly about the part about how the recipient is supposed to get the connection invitation, the rest is connection protocol of course.

TimoGlastra:
Not an exact specification. Couple of routes we could take:

- Add mediator invitation as startup parameter. In ACA-Py you can specify `--mediator-invitation` to connect and request mediation afterwards
- Use a public did as implicit invitation. E.g. using a `did:sov` or `did:web:example.com` to resolve the service endpoint and send a request afterwards.
	- either as startup config `{ "mediatorDid": "did:sov:xxxxx" }`
	- or dynamically after startup
		- create connection with public did
		- request mediation from connection created with public did

I think it would be cool to have the mediator expose a `did:web` containing the service endpoint (sort of like trinsic: https://trinsic.id/.well-known/did.json, but they don't use it for their mediation endpoint). 

[...]

PR #256, opened 2021-05-02
feat: Add support for WebSocket transports

jakubkoci:
This is how it works now:
* A mediator takes endpoint for invitation or DidDoc from `config`.
* An edge agent takes endpoint for invitation or DidDoc from `inboundConnection` if it has any.
* The edge agent gets out of band invitation via http request to mediator (http://mediator1.com/invitation). The invitation contains `ws` endpoint because we want to make a connection via WebSocket.
* The edge agent takes an endpoint for outbound communication from invitation or DidDoc. For a connection with another edge agent, it takes an endpoint from `indbound` connection. Therefore it’s the same `ws` endpoint as it is contained in the mediator invitation.

To integrate ws and http servers together and support multiple transports we would need to enable following:
* A mediator sends a different endpoint for mediation to edge agent.
* The edge agent sets the mediation endpoint somewhere else and do not take it from `inbound` connection invitation or DidDoc.

Other notes:
* I set ws endpoint directly in mediator server by editing `config.endpoint` to value `ws://localhost:${PORT}`
* I skip the ws tests. To run them it's needed to change the mediator server to `mediator-ws.ts` (or integrate ws and HTTP together as I mentioned)

Related to #250

### DIFF
--- a/src/agent/Agent.ts
+++ b/src/agent/Agent.ts
@@ -104,6 +104,10 @@ export class Agent {
     this.messageSender.setOutboundTransporter(outboundTransporter)
   }
 
+  public getOutboundTransporter() {
+    return this.messageSender.getOutboundTransporter()
+  }
+
   public async init() {
     await this.wallet.init()
 
--- a/src/agent/Agent.ts
+++ b/src/agent/Agent.ts
@@ -20,6 +20,7 @@ import { BasicMessagesModule } from '../modules/basic-messages/BasicMessagesModu
 import { LedgerModule } from '../modules/ledger/LedgerModule'
 import { InMemoryMessageRepository } from '../storage/InMemoryMessageRepository'
 import { Symbols } from '../symbols'
+import { Transport } from './TransportService'
 
 export class Agent {
   protected agentConfig: AgentConfig
--- a/src/agent/Agent.ts
+++ b/src/agent/Agent.ts
@@ -31,6 +31,7 @@ export class Agent {
   protected messageReceiver: MessageReceiver
   protected messageSender: MessageSender
   public inboundTransporter?: InboundTransporter
+  public outboundTransporter?: OutboundTransporter
 
   public readonly connections!: ConnectionsModule
   public readonly proofs!: ProofsModule
--- a/src/agent/Agent.ts
+++ b/src/agent/Agent.ts
@@ -132,8 +132,8 @@ export class Agent {
     return this.agentConfig.mediatorUrl
   }
 
-  public async receiveMessage(inboundPackedMessage: unknown) {
-    return await this.messageReceiver.receiveMessage(inboundPackedMessage)
+  public async receiveMessage(inboundPackedMessage: unknown, transport?: Transport) {
+    return await this.messageReceiver.receiveMessage(inboundPackedMessage, transport)
   }
 
   public async closeAndDeleteWallet() {
--- a/src/agent/TransportService.ts
+++ b/src/agent/TransportService.ts
@@ -72,7 +72,7 @@ interface TransportTable {
   [connectionRecordId: string]: Transport
 }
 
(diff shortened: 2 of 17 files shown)


src/agent/TransportService.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  public constructor(logger: Logger) {
+    this.logger = logger
+  }
+
+  public saveTransport(connectionId: string, transport: Transport) {
+    this.transportTable[connectionId] = transport
+  }
+
+  public resolveTransport(connection: ConnectionRecord): Transport {
```
TimoGlastra, line 18: Would be nice if we can make this more dynamic. this means we need to support the transport in core. Ideally adding an outbound/inbound transport class is enough

src/agent/TransportService.ts
```diff
@@ -0,0 +1,94 @@
...
+
+type TransportType = 'ws' | 'http' | 'queue'
+
+export interface Transport {
+  type: TransportType
+  endpoint: string
+}
+
+export class WebSocketTransport implements Transport {
+  public type: TransportType = 'ws'
```
TimoGlastra, line 72: makes for strict type of `ws`:
```suggestion
  public readonly type = "ws"
```

and then (haven't checked syntax, but something like this should be possible):

```
type TransportType = WebSocketTransport['type'] | ....
```

TimoGlastra: Also is `ws` bound to the scheme. In that case, how about `wss`?

src/agent/TransportService.ts
jakubkoci: Agree. I think we would need to move this logic to outbound transporters and just provide some utility functions to get the endpoint from DidDoc instead of having it in the core. That would also solve your other suggestions like "update this method to pass in supported outbound transport /schemes".

src/agent/TransportService.ts
jakubkoci: I mean `ws` just as WebSocket in general.

jakubkoci: I updated the attribute using `readonly`

src/agent/Agent.ts
```diff
@@ -159,8 +162,8 @@ export class Agent {
     return this.agentConfig.mediatorUrl
   }
 
-  public async receiveMessage(inboundPackedMessage: unknown) {
-    return await this.messageReceiver.receiveMessage(inboundPackedMessage)
+  public async receiveMessage(inboundPackedMessage: unknown, transport?: Transport) {
```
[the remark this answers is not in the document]
jakubkoci: What about creating the `InboundMessageContext` here and then enhanced it with other attributes inside `MessageReciever`? 

I would rather not introduce yet another type because we already have perhaps too much (XxxMessage, XxxMessageContext, XxxPackage, ...) Then we could do the same for the outbound messages as `OutboundMessageContext`.

src/agent/TransportService.ts
TimoGlastra: In that case, maybe `WebSocket` or `WebSocketTransport`? Seems more in line with the other transport type strings we use (e.g. for records)

src/agent/Agent.ts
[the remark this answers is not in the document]
TimoGlastra: Ah we already have `InboundMessageContext` of course. Yes that sounds perfect!

jakubkoci:
@TimoGlastra Updated the PR according to your suggestions. 

I think we can update and refactor the transport resolve mechanism when we add multi-transport. It will be more obvious how to structure the code then.

TimoGlastra (approved):
LGTM. 

This introduces some technical debt with how transports are managed that we can hopefully fix in the near future, but I think it is good to get this merged now.

TimoGlastra:
The release pipeline is failing. I'll take a look at it today

PR #257, opened 2021-05-02
feat: add dependency injection

### LINKED ISSUE #211: Improve dependency injection
As a developer, I want to dynamically inject and use various framework modules and their classes, so that I can preserve the open-closed principle when I want to enhance or change the agent's behavior based on a custom implementation.

Currently, we have dependencies hardcoded in Agent class and wired together manually. If someone wants to update or change the behavior of the framework they must change the Agent class directly.

We consider the use of some existing JS dependency injection solutions:
- [Nest.js DI](https://docs.nestjs.com/), [Minimal implementation example](https://github.com/jakubkoci/nest-di)
- [Awilix](https://github.com/jeffijoe/awilix)
- [Inversify](https://github.com/inversify/InversifyJS)
- [TSyringe](https://github.com/microsoft/tsyringe)
- [TypeDI](https://github.com/typestack/typedi)

Other resources:
- [JavaScript dependency injection in Node.js – friends or foes?](https://tsh.io/blog/dependency-injection-in-node-js/)

Acceptance criteria:
- It works in React Native and Electron

TimoGlastra:
This was considerably easier than I thought it would be.

Just DI for now. No extension API yet.

- Used TSyringe. Very lightweight (1 already used dependency)
- Used symbols for interface based injectables. Maybe we want to use base classes later on to make it easier, but I wanted to keep most of the current working intact. Just replace it with DI
- Use `@scoped(Lifecycle.ContainerScoped)` for injection. This is basically a singleton, however only for the current DI container. This means a child container won't share the same instance. Two reasons:
	- The tests were not happy with a singleton (could have bypassed this however)
	- This allows to run two agents side by side and don't have problems with the singletons
- Instead of using `Repository<RecordClass>` we're now using `RecordRepository`. It was very hard to use this pattern. Maybe we can simplify it later on. For now it was easiest to just create a separate repository per record type.

Let me know what you think. I think this will make extension a lot easier moving forward.

Fixes #211


TimoGlastra:
Errors fixed, tests added

JamesKEbert (commented):
This looks great! Definitely moves us towards having extensibility capabilities.
To your singleton comments, this would provide the modules and such as singletons to only that Agent per say since we're using the scoped container, right? So it essentially means modules are singletons, except that agents have their own singletons (which I think make sense IMO).

TimoGlastra:
> So it essentially means modules are singletons, except that agents have their own singletons (which I think make sense IMO).

Exactly!

jakubkoci (approved):
Awesome! Although it would be nice to avoid interface injection via Symbols, I would rather keep interfaces than creating classes where it shouldn't be a class (but maybe I'm too strict here 😄 )

I don't see much of a problem in changing `Repository<RecordClass>` to `RecordRepository`

jakubkoci:
> * This allows to run two agents side by side and don't have problems with the singletons

Yes, this is important to preserve 👍

jakubkoci:
I didn't merge it just for the case you want to merge it in some order with other PRs :)

jakubkoci:
One more thing. Perhaps it's not necessary but have you tested if it works with React Native? I know it should but TSyringe docs say something about Babel transform setup. It could be worthy to be sure it works and eventually describe what to do in our docs here. What do you think, @TimoGlastra ?

TimoGlastra:
Good point @jakubkoci. We need a way to easily test if the framework works in RN (not sure if bifold is already the easiest to to do that)

TimoGlastra:
> TSyringe docs say something about Babel transform setup

This is only when using babel to do TS compilation. We use the native TS compiler and publish that to NPM so it should already include the TS metadata. But would be good to check

JamesKEbert:
Definitely a good point @jakubkoci! I _think_ Bifold would be the easiest to do that--and I'd like to get it updated as necessary anyways. So, I could test this out later today if that'd be helpful.

TimoGlastra:
> Definitely a good point @jakubkoci! I _think_ Bifold would be the easiest to do that--and I'd like to get it updated as necessary anyways. So, I could test this out later today if that'd be helpful.

That would definitely be helpful @JamesKEbert! Thanks. It still needs to be updated to use the NPM package right? If we make that update it will be easier to quickly test new releases

JamesKEbert (approved):
This is great--and I have tested it and it is working as expected in a React Native context in Aries Bifold.

PR #260, opened 2021-05-07
ci: sperate linting and validation tasks

### LINKED ISSUE #91: Make linting and compiling a separate CI job
#90 added linting to the CI test job. However this should be a separate job.

However the jobs need some common setup for libindy. Valid options: 

- Skip the rebuild step of indy-sdk. We don't need it for linting / compiling
- Create a local setup-indy action (like actions/setup-node) that we use in every job to setup the indy dependencies. This way we don't need to build a framework docker image anymore
- separate prepare job to build the framework image

TimoGlastra:
image

Because we run the linting jobs inside a docker container we are not benefiting from all of the github features. See image above. This separates the linting /validation from the other jobs.

Fixes #91

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -17,6 +17,30 @@ env:
   GENESIS_TXN_PATH: network/genesis/local-genesis.txn
 
 jobs:
+  validate:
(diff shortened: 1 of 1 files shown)


JamesKEbert (approved):
Looks good to me!

PR #262, opened 2021-05-10
refactor: extract credential logic from wallet

TimoGlastra:
This became a bigger PR than I anticipated, but I think the changes are quite nice.

- Extract all indy credential logic from wallet to Issuer/Holder/Verifier services. this makes the wallet less indy specific, which will be nice if we add other wallet types in the future
	- This is mainly done as a refactor for revocation so we don't overload the wallet again with indy specific revocation methods.
- Update ledger create schema/cred def method signatures to not return array of [id, object] as the object already contains the id. So just return the object for both
-  Add `FileSystem` interface with implementation for both NodeJS and React Native
	- We need file access for revocation
	- This will allow us to add a `genesisUrl` and `genesisTransactions` config to the framework so we can handle the complexity of storing the genesis file for the consumer (will address in a separate PR)
- Created `indy` module directory, but I don't see the need for an `IndyModule` at the moment. Can we create 'modules' without a module class? (IMO yes, but would like your opinion)

Please let me know what you think. I'd like to test this in RN first, so do not merge yet.


TimoGlastra:
Also, maybe we should look at creating a monorepo sooner than later. I'd really like to make everything work out of the box, but that means adding dependencies for both RN + NodeJS to core.... If we make it a monorepo we wouldn't have to do that as we have an easy way to publish multiple packages

TimoGlastra:
Dependant on https://github.com/DefinitelyTyped/DefinitelyTyped/pull/52885. The types is already ready to merge

src/storage/fs/index.ts
```diff
@@ -0,0 +1,15 @@
+import { isNodeJS } from '../../utils/environment'
+import { Constructor } from '../../utils/mixins'
+import { FileSystem } from './FileSystem'
+
+let FileSystemClass: Constructor<FileSystem>
+
+if (isNodeJS()) {
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  FileSystemClass = require('./NodeFileSystem').NodeFileSystem
```
jakubkoci, line 9: Not sure if this is gonna work in RN. I think that the RN metro bundler tries to load everything which is required somewhere in the code. This condition is then called only in runtime

src/storage/fs/FileSystem.ts
```diff
@@ -0,0 +1,7 @@
+export interface FileSystem {
+  readonly basePath: string
+
+  exists(path: string): Promise<boolean>
+  write(path: string, data: string): Promise<void>
+  read(path: string): Promise<string>
```
jakubkoci, line 6: I wanted to suggest preserving `fs` API and not creating our own, which would be better, but I see there are tiny differences between them.

jakubkoci (commented):
It seems quite good. I like the improvement of expressivness by adding issuer, holder, verifier to the names.

Not sure about the `indy` module. Other modules represent more general concepts while `indy` is more technology-based name. Most of the code is related only to credentials, which could suggest putting it inside `credentials` module. On the other hand, there are a few methods used by `LedgerService` 🤷‍♂️

jakubkoci:
> Also, maybe we should look at creating a monorepo sooner than later. I'd really like to make everything work out of the box, but that means adding dependencies for both RN + NodeJS to core.... If we make it a monorepo we wouldn't have to do that as we have an easy way to publish multiple packages

I agree, I don't like having those dependencies inside the framework.

src/storage/fs/FileSystem.ts
TimoGlastra: Yeah there's some differences between the NodeJS and RN API so I created a very simple interface that works for both systems. NodeJS does not have `exists` which I think is nice to have

src/storage/fs/index.ts
TimoGlastra: Hmm didn't think about it. Seems complex: https://reactnative.dev/docs/ram-bundles-inline-requires#inline-requires

If we already want to move to multiple packages the best way for now I think would be to export both classes and require it to be passed as a config:

We could do the same for the `fetch` method so we don't need to depend on `node-fetch`.

```ts
new Agent({
	fileSystem: new ReactNativeFileSystem()
})
```

TimoGlastra:
> I agree, I don't like having those dependencies inside the framework.

I'll do some research on this

TimoGlastra:
> Not sure about the `indy` module. Other modules represent more general concepts while `indy` is more technology-based name. Most of the code is related only to credentials, which could suggest putting it inside `credentials` module. On the other hand, there are a few methods used by `LedgerService` 🤷‍♂️

Hmm credentials also doesn't seem like the best place. Indy also contains code for ledger and proofs. Are you ok for keeping it in indy for now so we can refactor later on? I think the extraction was the hard part, moving it to another directory will be a fairly easy task to do later. Thinking of it from a monorepo perspective this would probably be hosted in an `indy` package that all those modules depend on.

TimoGlastra:
Tested to work in aries-mobile-agent-react-native

jakubkoci (approved):
Okey let's leave it for now, maybe it has its place there :)

I see a monorepo mainly as a way to get rid of mixing platform-specific dependencies (node and RN) together.

PR #263, opened 2021-05-11
feat: add support for Multibase, Multihash and Hashlinks

blu3beri:
Implements Multihash, Multibase and Hashlinks. 

Adds decoding, encoding and validating for multibase, multihash and hashlinks.

Hashlinks have been implemented accoring to [draft-sporny-hashlink-07](https://datatracker.ietf.org/doc/html/draft-sporny-hashlink)

The main reason for this implementation is to add multibase for the upcoming did exchange protocol and hashlink for linking credentials to attachments.


src/utils/MultibaseEncoder.ts
```diff
@@ -0,0 +1,43 @@
...
+    return { data: multibase.decode(data), baseName }
+  }
+
+  /**
+   *
+   * Validates if it is a valid multibase encoded value
+   *
+   * @param {Uint8Array} data the multibase that needs to be validated
+   *
+   * @returns {boolean} bool wether the multibase value is encoded
```
TimoGlastra, line 38:
```suggestion
   * @returns {boolean} bool whether the multibase value is encoded
```

src/utils/MultihashEncoder.ts
```diff
@@ -0,0 +1,41 @@
...
+  public static decode(data: Uint8Array): { data: Uint8Array; hashName: string } {
+    const decodedHash = multihash.decode(data)
+    return { data: decodedHash.digest, hashName: decodedHash.name }
+  }
+
+  /**
+   *
+   * Validates if it is a valid mulithash
+   *
+   * @param {Uint8Array} data the multihash that needs to be validated
```
TimoGlastra, line 31:
```suggestion
   * @param {Uint8Array} data the multihash that needs to be validated
   * @returns {boolean} whether the multihash is valid
```

src/utils/__tests__/MultibaseEncoder.test.ts
```diff
@@ -0,0 +1,36 @@
+import { MultibaseEncoder } from '../MultibaseEncoder'
+import { Buffer } from 'buffer'
+import { BufferEncoder } from '../BufferEncoder'
+
+const validData = Buffer.from('Hello World!')
+const validMultibase = 'zKWfinQuRQ3ekD1danFHqvKRg9koFp8vpokUeREEgjSyHwweeKDFaxVHi'
+const invalidMultibase = 'gKWfinQuRQ3ekD1danFHqvKRg9koFp8vpokUeREEgjSyHwweeKDFaxVHi'
+
+describe('multibase', () => {
```
TimoGlastra, line 9: We often try to match the describes with the class and method names:

```suggestion
describe('MultiBaseEncoder', () => {
	describe('encode()', () => {
	})
})
```

src/utils/hashlink.ts
```diff
@@ -0,0 +1,118 @@
+import cbor from 'cbor'
```
TimoGlastra, line 1: `HashLinkEncoder.ts` and `class HashLinkEncoder` to be consistent with other similar files/classes

src/utils/hashlink.ts
```diff
@@ -0,0 +1,118 @@
...
+
+    return BufferEncoder.toUtf8String(multibaseMetadata)
+  }
+
+  private static decodeMetadata(mb: string): Metadata {
+    const obj = { urls: [], contentType: '' }
+    const { data } = MultibaseEncoder.decode(mb)
+    try {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      const cborData: Map<number, any> = cbor.decode(data)
```
TimoGlastra, line 100:
```suggestion
      const cborData: Map<number, unknown> = cbor.decode(data)
```

TimoGlastra (approved):
Some small suggestions

TimoGlastra:
Although conventional commits allows to use upper case types, I prefer `feat:` over `Feat:` for consistency?

src/utils/hashlink.ts
blu3beri: Unknown does not work in this situation, because of:

`Type 'unknown' is not assignable to type 'string[]'.ts(2322)`

When assigning the url list to the url list in the object.

src/utils/hashlinkEncoder.ts
```diff
@@ -0,0 +1,124 @@
...
+        case 'urls':
+          metadataMap.set(URLS, metadata.urls)
+          break
+        case 'contentType':
+          metadataMap.set(CONTENTTYPE, metadata.contentType)
+          break
+        default:
+          throw new Error(`Invalid metadata: ${metadata}`)
+      }
+    }
```
jakubkoci, line 92: I'm thinking about how to get rid of the switch statement and something like this could work:
```js

hexTable = {
  urls: 0x0f, 
  contentType: 0x0e
}

for (const key of Object.keys(metadata)) {
  if (hexTable[key]) { 
    metadataMap.set(hexTable[key], metadata[key])
  } else {
    throw new Error(...)
  }
}
```

jakubkoci: Similarily for the `decodeMetadata`. But maybe I'm missing something which would complicate the code anyway 🤷‍♂️

jakubkoci (changes requested):
Good job. Code is clean, readable, and well-structured 👍  I could resist adding a few suggestions but nothing critical :)

We should rename the files according to a class naming convention at least.

package.json
```diff
@@ -23,12 +23,16 @@
...
+    "bignumber.js": "^9.0.1",
     "bn.js": "^5.2.0",
     "buffer": "^6.0.3",
+    "cbor": "^7.0.5",
     "class-transformer": "^0.4.0",
     "class-validator": "^0.13.1",
     "events": "^3.3.0",
     "js-sha256": "^0.9.0",
+    "multibase": "^4.0.4",
+    "multihashes": "^4.0.2",
```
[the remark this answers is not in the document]
blu3beri: Just reverted back to multibase@4.0.2, since 4.0.3+ does not work anymore in React Native according to [js-multibase#98](https://github.com/multiformats/js-multibase/issues/98).

package.json
```diff
@@ -23,14 +23,18 @@
     "prepare": "husky install"
   },
   "dependencies": {
+    "bignumber.js": "^9.0.1",
     "bn.js": "^5.2.0",
     "buffer": "^6.0.3",
+    "cbor": "^7.0.5",
```
jakubkoci, line 29: I'm sorry I'm not sure about this one. It seems it's dependent on Node.js `stream` I tried to install `cbor` in RN app but getting the following error:
image

TimoGlastra: The library is called node-cbor so good chance it won't work in RN

blu3beri: Ah sorry, forgot about cbor. I will find an alternative.

blu3beri: Changed the dependency to `borc`. Tested it in React Native and it works for me.

jakubkoci: Great 😄

jakubkoci (approved):

PR #264, opened 2021-05-12
ci: do not run concurrent release jobs

TimoGlastra:
This should fix the issues with the release. Github added a `concurrency` key a few weeks ago that allows us to prevent concurrent jobs from running. See the comment in the action file for more context

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idconcurrency


jakubkoci (approved):
Thanks, Timo for the quick investigation and fix 👍

PR #266, opened 2021-05-12
feat: Added attachment extension

blu3beri:
Added the extension to the attachment decorator and renamed the attachment property on the issue and verify messages to avoid confliction.

Has been added for the support of linking credentials to attachments.

### DIFF
--- a/src/decorators/attachment/AttachmentExtension.ts
+++ b/src/decorators/attachment/AttachmentExtension.ts
@@ -1,5 +1,5 @@
 import { Expose, Type } from 'class-transformer'
-import { IsArray, ValidateNested } from 'class-validator'
+import { IsArray, IsOptional, ValidateNested } from 'class-validator'
 
 import { BaseMessageConstructor } from '../../agent/BaseMessage'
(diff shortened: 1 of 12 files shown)


src/decorators/attachment/AttachmentExtension.ts
```diff
@@ -0,0 +1,33 @@
...
+     */
+    @Expose({ name: '~attach' })
+    @Type(() => Attachment)
+    @ValidateNested()
+    @IsArray()
+    @IsOptional()
+    public attachments?: Attachment[]
+
+    public getAttachmentById(id: string): Attachment | undefined {
+      return this.attachments?.find((attachment) => attachment.id === id.substring(0, 64))
```
TimoGlastra, line 20: We shouldn't truncate the `id` property here, it should just match the two string values.

Truncating can be done on a higher level

TimoGlastra (changes requested):
Thanks @blu3beri, one small suggestion

TimoGlastra (approved):

jakubkoci (approved):
Nice 👍

PR #267, opened 2021-05-12
feat: add internal ws outbound transporter

TimoGlastra:
add internal ws outbound transporter. Same as with HTTP. Next step is multi outbound transport support.

### DIFF
--- a/samples/mediator-ws.ts
+++ b/samples/mediator-ws.ts
@@ -49,8 +49,16 @@ class WsInboundTransporter implements InboundTransporter {
   }
 }
 
+// TODO: use WsOutboundTransporter from the agent
 class WsOutboundTransporter implements OutboundTransporter {
   public supportedSchemes = ['ws', 'wss']
+
+  public async start(): Promise<void> {
+    // Nothing required to start WS
+  }
+  public async stop(): Promise<void> {
+    // Nothing required to stop WS
+  }
 
   public async sendMessage(outboundPackage: OutboundPackage) {
     const { connection, payload, transport } = outboundPackage
--- a/samples/mediator-ws.ts
+++ b/samples/mediator-ws.ts
@@ -50,7 +50,7 @@ class WsInboundTransporter implements InboundTransporter {
 }
 
 class WsOutboundTransporter implements OutboundTransporter {
-  public supportedSchemes = ['ws']
+  public supportedSchemes = ['ws', 'wss']
 
   public async sendMessage(outboundPackage: OutboundPackage) {
     const { connection, payload, transport } = outboundPackage
--- a/src/agent/MessageSender.ts
+++ b/src/agent/MessageSender.ts
@@ -9,7 +9,7 @@ import { TransportService } from './TransportService'
 export class MessageSender {
   private envelopeService: EnvelopeService
   private transportService: TransportService
-  private outboundTransporter?: OutboundTransporter
+  private _outboundTransporter?: OutboundTransporter
 
   public constructor(envelopeService: EnvelopeService, transportService: TransportService) {
     this.envelopeService = envelopeService
@@ -17,11 +17,11 @@ export class MessageSender {
   }
 
   public setOutboundTransporter(outboundTransporter: OutboundTransporter) {
-    this.outboundTransporter = outboundTransporter
+    this._outboundTransporter = outboundTransporter
   }
 
-  public getOutboundTransporter() {
-    return this.outboundTransporter
+  public get outboundTransporter() {
+    return this._outboundTransporter
   }
 
   public async packMessage(outboundMessage: OutboundMessage): Promise<OutboundPackage> {
--- a/src/transport/WsOutboundTransporter.ts
+++ b/src/transport/WsOutboundTransporter.ts
@@ -18,6 +18,7 @@ export class WsOutboundTransporter implements OutboundTransporter {
     this.agent = agent
     this.logger = agent.injectionContainer.resolve(Symbols.Logger)
   }
+
   public async start(): Promise<void> {
     // Nothing required to start WS
   }
@@ -68,7 +69,9 @@ export class WsOutboundTransporter implements OutboundTransporter {
     return socket
   }
 
-  private handleMessageEvent(event: any) {
+  // NOTE: Because this method is passed to the event handler this must be a lambda method
+  // so 'this' is scoped to the 'WsOutboundTransporter' class instance
+  private handleMessageEvent = (event: any) => {
     this.logger.debug('WebSocket message event received.', { url: event.target.url, data: event.data })
     this.agent.receiveMessage(JSON.parse(event.data))
   }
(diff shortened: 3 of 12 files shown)


JamesKEbert (approved):
LGTM, I have not tested as of yet.

src/transport/WsOutboundTransporter.ts
```diff
@@ -0,0 +1,98 @@
+import { OutboundTransporter } from './OutboundTransporter'
+import { Agent } from '../agent/Agent'
+import { WebSocketTransport } from '../agent/TransportService'
+import { Logger } from '../logger'
+import { ConnectionRecord } from '../modules/connections'
+import { OutboundPackage } from '../types'
+import { Symbols } from '../symbols'
+import { WebSocket } from '../utils/ws'
+
+export class WsOutboundTransporter implements OutboundTransporter {
```
jakubkoci, line 10: This is only the WebSocket outbound transporter from the client-side perspective. It could be good to mention it somewhere. Folder structure `src/transport/client`, name of the class `WsClientOutboundTransporter`? Or `edge`/`holder` I'm not sure 🤷‍♂️  😄

src/utils/ws.ts
```diff
@@ -0,0 +1,14 @@
+import { isNodeJS } from './environment'
+
+// RN exposes global WebSocket
+let WebSocket = global.WebSocket
+
+// NodeJS doesn't have WebSocket by default
+if (!WebSocket && isNodeJS()) {
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const nodeWebSocket = require('ws')
```
jakubkoci, line 9: I think this `ws` package will be part of the React Native bundle anyway. Not saying it's bad, because we would like to address this by monorepo. Just thought it's worth mentioning.

src/transport/WsOutboundTransporter.ts
```diff
@@ -0,0 +1,98 @@
...
+    this.logger = agent.injectionContainer.resolve(Symbols.Logger)
+  }
+  public async start(): Promise<void> {
+    // Nothing required to start WS
+  }
+
+  public async stop() {
+    this.transportTable.forEach((socket) => {
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
```
jakubkoci, line 28: Why do we have `@ts-ignore` here? Is it because `WebSocket` from `ws` interface differs from ES standard and doesn't have `removeAllListeners`?

samples/mediator-ws.ts
```diff
@@ -49,8 +49,16 @@ class WsInboundTransporter implements InboundTransporter {
   }
 }
 
+// TODO: use WsOutboundTransporter from the agent
 class WsOutboundTransporter implements OutboundTransporter {
-  public supportedSchemes = ['ws']
+  public supportedSchemes = ['ws', 'wss']
```
jakubkoci, line 54: I haven't tested `wss` and I don't know much details about that protocol so I'm not sure if it will work out of the box.

jakubkoci (commented):
Nothing serious, mainly just questions. Overall seems good 👍

src/utils/ws.ts
TimoGlastra: React Native has a built-in WebSocket client implementation. Why would we need the `ws` library?

samples/mediator-ws.ts
TimoGlastra: Both `ws` and the React Native WebSocket client supports `wss`.

src/transport/WsOutboundTransporter.ts
TimoGlastra: Maybe I'm not understanding correctly.

But there is not necessarily a difference between a client / edge / holder outbound transport and another outbound transport. That should be the concern of the higher level. Outbound transport should just send (and optionally receive) messages to an endpoint. Same with HTTP outbound transport.

src/agent/MessageSender.ts
```diff
@@ -9,19 +9,19 @@ import { TransportService } from './TransportService'
...
   private transportService: TransportService
-  private outboundTransporter?: OutboundTransporter
+  private _outboundTransporter?: OutboundTransporter
 
   public constructor(envelopeService: EnvelopeService, transportService: TransportService) {
     this.envelopeService = envelopeService
     this.transportService = transportService
   }
 
   public setOutboundTransporter(outboundTransporter: OutboundTransporter) {
```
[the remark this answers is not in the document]
TimoGlastra: Generally I'm in favour of `get XXX` over `getXXX` if there are no side effects.

I didn't change the `setXXX` as It may feel weird to set the transport by doing `agent.outboundTransporter = XXX` over `agent.setOutboundTransporter(XXX)`

src/transport/WsOutboundTransporter.ts
TimoGlastra: To clarify further. On the server side it would be the `WsInboundTransporter` that hosts the WebSocket

TimoGlastra: The naming of inbound and outbound is confusing, but is from the standpoint of transport being simplex by default in Aries

TimoGlastra: So outbound transport is you initiate the interaction (so for HTTP and WS that means you're the client, server can't initiate the interaction)

For inbound transport it means the other agent initiates the interaction (so for HTTP and WS that means you host the server/endpoint, you can't initiate the interaction)

src/transport/WsOutboundTransporter.ts
TimoGlastra: Did some research. React Native / Browser websocket does not have remove all I believe. Updated

src/utils/ws.ts
jakubkoci: Oh, sorry, I meant that this logic will happen only in runtime, so when you build RN app then the output JS bundle may contain also the `ws` package which is not needed as you said.

src/agent/MessageSender.ts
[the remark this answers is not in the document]
jakubkoci: I see. Yeah, that was exactly the reason I kept getXxx and setXxx for both 😄 . But I'm good with it. API will probably change with multi-transports anyway

src/transport/WsOutboundTransporter.ts
jakubkoci: Ah, sorry. I realized I was too focused on the concrete implementation for edge agent and mediation where the logic differs, but the one doesn't exclude the other 👍

src/utils/ws.ts
jakubkoci: But do you see where the problem is here, right? :)

TimoGlastra: Yeah I see the problem. I'm just not that set on fixing this before we move to a monorepo :)

I already have the structure, the most important thing is determining how to separate the packages.

jakubkoci (approved):

src/utils/ws.ts
jakubkoci: Good good, of course, I didn't mean it as a blocker 👍

TimoGlastra: See #270 for discussion on monorepo

PR #269, opened 2021-05-13
build: make indy-sdk dev dependency

TimoGlastra:
Currently when installing AFJ from npm it tries to install indy-sdk. Moving it back to dev dependencies so this isn't the case anymore. This requires a bit more setup as yarn install now requires indy to be installed.

I created a local action setup-libindy that handles this and is reused by all jobs.

related to discussion over here https://github.com/hyperledger/aries-mobile-agent-react-native/issues/41


jakubkoci (approved):
Good job with jobs 👍 :)

PR #272, opened 2021-05-14
Dev/mediation config

burdettadam:
@DavidClawson work on configuration for meditation recipient. 
Introduces an initialize method to the recipient module(src/modules/routing/RecipientModule.ts).
- initialize is called from agent.init() after wallet, ledger, and transport are set up. 
- recipient.init() checks for provided configurations to automatically set up mediation at the start.


DavidClawson (approved):
Looks good!

TimoGlastra:
@burdettadam @DavidClawson I updated `feature/mediation-rebased` with the most recent changes: https://github.com/hyperledger/aries-framework-javascript/tree/feature/mediation-rebased

If you can do more smaller PRs I can make sure the branch stays up to date with changes in main. It'd rather not wait until the end to update it as this was already way too much merge conflicts. Does that sound reasonable?


So if we can use this branch from now on that would really help:

https://github.com/hyperledger/aries-framework-javascript/tree/feature/mediation-rebased

PR #273, opened 2021-05-14
feat: only connect to ledger when needed

### LINKED ISSUE #219: Pass a Genesis string to agent config instead of a file path
Currently the agent receives a genesisPath to a file. This is not convenient for dynamic configurations. It should be able to receive a string.

If a genesis file is included it is trivial to read it and pass the resulting string.

### LINKED ISSUE #265: Improve ledger pool management
Currently when the `genesisPath` property is passed into the config the `Agent.init()` method will automatically connect to the ledger and stay connected indefinitely. There are two main downsides to this approach (probably more):

- This has a significant impact on startup time as `init()` will not finish until it is connected to the ledger
- It is unnecessary to connect the ledger before doing any actual ledger operations

This is related to multi-ledger support

We probably need to add something to ledger service that before doing any operation it will check if the pool is already open, if not open it and then close it after not being used for e.g. two minutes

TimoGlastra:
- only connect to ledger when needed
- allow to pass `genesisTransactions` so you don't have to deal with storing the genesis yourself

Fixes #219 
Fixes #265 

The ledger management can be improved, but I think this is a simple first step. Only open the ledger connection if a ledger call is made

---

Tested in both NodeJS and React Native

### DIFF
--- a/src/modules/ledger/LedgerModule.ts
+++ b/src/modules/ledger/LedgerModule.ts
@@ -1,7 +1,7 @@
 import type { CredDefId, Did, SchemaId } from 'indy-sdk'
 import { inject, scoped, Lifecycle } from 'tsyringe'
 
-import { LedgerService, SchemaTemplate, CredentialDefinitionTemplate, LedgerConnectOptions } from './services'
+import { LedgerService, SchemaTemplate, CredentialDefinitionTemplate } from './services'
 import { Wallet } from '../../wallet/Wallet'
 import { Symbols } from '../../symbols'
 import { AriesFrameworkError } from '../../error'
@@ -14,10 +14,6 @@ export class LedgerModule {
   public constructor(@inject(Symbols.Wallet) wallet: Wallet, ledgerService: LedgerService) {
     this.ledgerService = ledgerService
     this.wallet = wallet
-  }
-
-  public async connect(poolName: string, poolConfig: LedgerConnectOptions) {
-    return this.ledgerService.connect(poolName, poolConfig)
   }
 
   public async registerPublicDid() {
--- a/src/modules/ledger/services/LedgerService.ts
+++ b/src/modules/ledger/services/LedgerService.ts
@@ -12,16 +12,12 @@ import type {
   LedgerWriteReplyResponse,
 } from 'indy-sdk'
 import { AgentConfig } from '../../../agent/AgentConfig'
-import { AriesFrameworkError } from '../../../error'
 import { Logger } from '../../../logger'
 import { isIndyError } from '../../../utils/indyError'
 import { Wallet } from '../../../wallet/Wallet'
 import { Symbols } from '../../../symbols'
 import { IndyIssuerService } from '../../indy'
-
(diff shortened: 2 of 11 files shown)


src/modules/ledger/services/LedgerService.ts
```diff
@@ -30,27 +27,43 @@ export class LedgerService {
...
-      throw new Error('Pool has not been initialized yet.')
+      return this.connect()
     }
 
     return this._poolHandle
   }
 
-  public async connect(poolName: string, poolConfig: LedgerConnectOptions) {
-    this.logger.debug(`Connecting to ledger pool '${poolName}'`, poolConfig)
+  private async connect() {
```
jakubkoci, line 55: I think a developer should be able to make the connection if they decide. In our use case, we found some problems when connecting to the staging from the mobile app. Sometimes it takes tens of seconds. Therefore it's convenient to separate the connection to the ledger from the wallet initialization but start making the connection right after and don't wait until the connection is needed.

jakubkoci (changes requested):

src/modules/ledger/services/LedgerService.ts
TimoGlastra: I agree. Updated

jakubkoci:
There is this mention "if not open it and then close it after not being used for e.g. two minutes" in #265. Is it also implemented in this PR? Are we able to somehow detect that the connection to the ledger has been lost?

src/modules/ledger/LedgerModule.ts
```diff
@@ -16,10 +16,6 @@ export class LedgerModule {
     this.wallet = wallet
   }
 
-  public async connect(poolName: string, poolConfig: LedgerConnectOptions) {
-    return this.ledgerService.connect(poolName, poolConfig)
-  }
```
jakubkoci, line 21: Can we make this one also public?

jakubkoci (approved):
We should just make the `connect` method in the module public. My other comment is not a blocker for this PR.

src/modules/ledger/LedgerModule.ts
TimoGlastra: oops forgot about that one

TimoGlastra:
> There is this mention "if not open it and then close it after not being used for e.g. two minutes" in #265. Is it also implemented in this PR? Are we able to somehow detect that the connection to the ledger has been lost?

No that's not implemented as part of this PR, I wanted to address that in a separate PR. 

Next step is allowing to close the pool and reconnect. But to keep PRs smaller I'm taking baby steps :)

jakubkoci:
> Next step is allowing to close the pool and reconnect. But to keep PRs smaller I'm taking baby steps :)

That's good 👍  of course.. I just wanted to mention it because the issue is closed by this PR now, but it could be good to have some opened for this work ;)

TimoGlastra:
I created a new issue #277.

Could you make the CI jobs required for merge? That's easier with auto-merge. I don't have the permission to do so

PR #275, opened 2021-05-15
refactor: replace all errors with framework errors

### LINKED ISSUE #78: Consistent handling when records are not found
When a record is not found sometimes this will throw an error, sometimes this will return null. We should avoid this:

- If we do throw an error, make it specific to that error. Now we get an indy sdk error. This should be either a `FrameworkError` or a `RecordNotFoundError`.
- use consistent naming. `findXXX` can return value or null. `getXXX` can only return value. If value can't be returned error should be thrown.

TimoGlastra:
This PR adds a base `AriesFrameworkError` that is now thrown instead of the generic `Error`. It also adds `RecordNotFoundError` and `RecordDuplicateError` that is now used when storing/retrieving records. It could be hard to find out why an indy operation was failing. This gives a lot more context to the framework consumer.

Over time we can add more specific error classes for other parts of the codebase, but I thought this was already a great start.

Fixes #78


src/error/AriesFrameworkError.ts
```diff
@@ -0,0 +1,19 @@
...
+   * @param options.cause the error that caused this error to be created
+   * @param options.constructorOpt a function above which all frames, including constructorOpt, will be omitted from the generated stack trace.
+   *
+   * @see "https://nodejs.org/api/errors.html#errors_error_capturestacktrace_targetobject_constructoropt"
+   */
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  public constructor(message: string, { cause, constructorOpt }: { cause?: Error; constructorOpt?: Function } = {}) {
+    super(message)
+    this.cause = cause
+    Error.captureStackTrace(this, constructorOpt ?? this.constructor)
```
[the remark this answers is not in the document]
jakubkoci: I usually use [make-error-cause](https://www.npmjs.com/package/make-error-cause) for custom errors and leave the heavy weight on that :)

src/error/RecordDuplicateError.ts
```diff
@@ -0,0 +1,7 @@
+import { AriesFrameworkError } from './AriesFrameworkError'
+
+export class RecordDuplicateError extends AriesFrameworkError {
+  public constructor(message: string, { recordType, cause }: { recordType: string; cause?: Error }) {
```
jakubkoci, line 4: I'm not sure about mixing domain info with infra info into one param `{ recordType, cause }`. On the other hand, I understand that if we have AriesFrameworkError(message, cause, details/metadata) or vice versa `AriesFrameworkError(message, details/metadata, cause)` we could end up with empty param for one case or another 🤔  🤷‍♂️

src/error/RecordNotFoundError.ts
```diff
@@ -0,0 +1,7 @@
+import { AriesFrameworkError } from './AriesFrameworkError'
```
jakubkoci, line 1: Off-topic: I like the import `from './AriesFrameworkError'` and not `from '.'` as it's at some other places. It would be good to have some convention and keep with that. What do you think?

The convention would be: If you import from the same module you should import by the whole path. If you import from a different module you can import from `index.ts`

jakubkoci (approved):
Just a few notes, but overall good 👍

src/storage/IndyStorageService.ts
```diff
@@ -1,16 +1,15 @@
...
 import { JsonTransformer } from '../utils/JsonTransformer'
-import { Constructor } from '../utils/mixins'
 import { Symbols } from '../symbols'
 
-export interface BaseRecordConstructor<T> extends Constructor<T> {
-  type: string
-}
+import { handleIndyError, IndyError, isIndyError } from '../utils/indyError'
+import { AriesFrameworkError, RecordNotFoundError } from '../error'
+import { RecordDuplicateError } from '../error/RecordDuplicateError'
```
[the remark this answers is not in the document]
TimoGlastra: Auto-import does some weird things sometimes. I've been fiddling a bit with eslint-plugin-import a bit so we can enforce (and auto-fix!) such things.

However they don't support the `import type` from TS very well so it will break your code

src/error/RecordDuplicateError.ts
TimoGlastra: Not sure how to act on this. Do you want me to change it?

My thinking was you have a message and all other stuff goes into the object. This way you don't have that problem with order of parameters

src/error/AriesFrameworkError.ts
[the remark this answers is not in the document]
TimoGlastra: You're right. It thought that was also a browser/RN thing :shrug:

src/error/RecordNotFoundError.ts
TimoGlastra: I agree with this convention. I'll try to maintain this convention going forward

TimoGlastra:
@jakubkoci updated based on your comments. Could you give it another look (especially the usage of make-error-cause? Haven't worked with that before)

src/error/RecordDuplicateError.ts
jakubkoci: Okay, I see. No, we don't have to change it. We can go this way and we'll see how it suits us.

jakubkoci (approved):
Yeah, I think it's good 👍

JamesKEbert:
This is great--thank you for your efforts here. Question, is there a way in which we should be able to access errors through the Logger? Like could errors get recorded there, for in a case in which we record logs from the Logger vs just everything getting printed to the console? If that makes any sense.

icc-romeu:
@JamesKEbert I think you can simply create your own implementation of the logger that stores the messages. Hence, just storing the "yourLogger.error(whatever)" calls you can access the errors from the framework.

PR #280, opened 2021-05-19
refactor: Move a transport protocol-related logic from the framework core

jakubkoci:
I did some refactorings to help me add support for multiple transports in the following PR:
* Add test to `MessageSender`
* Move up a logic from `EnvelopeService` to `MessageSender`
* Remove transport-related details from the `TransportService` and from the core of the framework.
* Rename `transport` to `session` which hopefully express the meaning better.

As a next step I'm considering:
* Rename transporters to transports
* Move up more logic from `MessageSender.packMessage` and pass `service` object to `packMessage`
* Then we could call `packMessage` and `sendMessage` for every available service. Eventually, we could remove the logic from `createOutboundMessage`.
* Improve validation of `outboundMessage` inside `OutboundTransporter` and error handling
* Add multiple transports

If you think it doesn't make sense to merge this without full functionality I can continue with adding changes. But I realized that it contains a good amount of changes already and it could be worth merging sooner.

### DIFF
--- a/src/agent/MessageSender.ts
+++ b/src/agent/MessageSender.ts
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
 import { inject, Lifecycle, scoped } from 'tsyringe'
 
 import { OutboundMessage, OutboundPackage } from '../types'
@@ -34,10 +35,11 @@ export class MessageSender {
   }
 
   public async packMessage(outboundMessage: OutboundMessage): Promise<OutboundPackage> {
-    const { connection, payload, endpoint } = outboundMessage
+    const { connection, payload } = outboundMessage
     const { verkey, theirKey } = connection
+    const endpoint = this.transportService.findEndpoint(connection)
     const message = payload.toJSON()
-    this.logger.info('outboundMessage', { verkey, theirKey, endpoint, message })
+    this.logger.info('outboundMessage', { verkey, theirKey, message })
     const responseRequested = outboundMessage.payload.hasReturnRouting()
     const wireMessage = await this.envelopeService.packMessage(outboundMessage)
     return { connection, payload: wireMessage, endpoint, responseRequested }
@@ -48,7 +50,7 @@ export class MessageSender {
       throw new AriesFrameworkError('Agent has no outbound transporter!')
     }
     const outboundPackage = await this.packMessage(outboundMessage)
-    const transport = this.transportService.resolveTransport(outboundMessage.connection)
+    const transport = this.transportService.findTransport(outboundMessage.connection.id)
     outboundPackage.transport = transport
     await this.outboundTransporter.sendMessage(outboundPackage)
   }
--- a/src/agent/MessageSender.ts
+++ b/src/agent/MessageSender.ts
@@ -1,4 +1,3 @@
-/* eslint-disable no-console */
 import { inject, Lifecycle, scoped } from 'tsyringe'
 
 import { OutboundMessage, OutboundPackage } from '../types'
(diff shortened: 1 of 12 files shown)


src/agent/__tests__/MessageSender.test.ts
```diff
@@ -0,0 +1,144 @@
...
+import { OutboundPackage } from '../..'
+import { OutboundMessage } from '../../types'
+import { ConnectionRecord } from '../../modules/connections'
+import { ReturnRouteTypes } from '../../decorators/transport/TransportDecorator'
+import { getMockConnection } from '../../__tests__/helpers'
+
+jest.mock('../TransportService')
+jest.mock('../EnvelopeService')
+
+const logger = testLogger
```
TimoGlastra, line 17: maybe just `import logger from ../../__tests__/logger`?

src/agent/__tests__/MessageSender.test.ts
```diff
@@ -0,0 +1,144 @@
...
+    y: '4Etl6SRW2YiLUrN5vfvVHuhp7x8PxltmWWlbbM4IFyM',
+    use: 'enc',
+    kid: '1',
+  }
+
+  const enveloperService = new EnvelopeService()
+  const envelopeServicePackMessageMock = enveloperService.packMessage as jest.Mock<
+    Promise<JsonWebKey>,
+    [OutboundMessage]
+  >
```
TimoGlastra, line 53:
```suggestion
  const envelopeServicePackMessageMock = mockFunction(enveloperService.packMessage)
```

src/agent/__tests__/MessageSender.test.ts
```diff
@@ -0,0 +1,144 @@
...
+
+/**
+ * Returns mock of function with correct type annotations according to original function `fn`.
+ * It can be used also for class methods.
+ *
+ * @param fn function you want to mock
+ * @returns mock function with type annotations
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function mockFunction<T extends (...args: any[]) => any>(fn: T): jest.MockedFunction<T> {
```
TimoGlastra, line 142: Can't we import this from the helpers file?

src/transport/WsOutboundTransporter.ts
```diff
@@ -1,12 +1,21 @@
...
 
+export class WebSocketTransportSession implements TransportSession {
+  public readonly type = 'websocket'
+  public socket?: WebSocket
+
+  public constructor(socket?: WebSocket) {
+    this.socket = socket
+  }
+}
+
```
TimoGlastra, line 18: Is the transport session for the inbound transport, outbound transport or both?

TimoGlastra:
My last PR broke the test and got into main 😬. I'm on it

src/transport/WsOutboundTransporter.ts
jakubkoci: Hmm. That's a good question. I would say it's for both. A developer sets the session when calling `agent.recieveMessage(message, session)` passing inbound message. Then it can be used for outbound messages.

src/agent/__tests__/MessageSender.test.ts
jakubkoci: I like `const logger = testLogger` more because you can simply change it to `const logger = new ConsoleLogger(...)` when want to log into the console during the development.

src/agent/MessageSender.ts
```diff
@@ -26,16 +35,22 @@ export class MessageSender {
   }
 
   public async packMessage(outboundMessage: OutboundMessage): Promise<OutboundPackage> {
-    return this.envelopeService.packMessage(outboundMessage)
+    const { connection, payload } = outboundMessage
+    const { verkey, theirKey } = connection
+    const endpoint = this.transportService.findEndpoint(connection)
+    const message = payload.toJSON()
+    this.logger.info('outboundMessage', { verkey, theirKey, message })
```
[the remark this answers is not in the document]
jakubkoci: Good point. I agree. I don't like it either. 

For now, I change this one to `debug` because it logs an unencrypted message. It seems outbound encrypted messages are logged in outbound transporters now. That we could change to `trace`. 

This logic will likely change, but I'll think about that more during the implementation. We can revise that in the next PR.

src/agent/__tests__/MessageSender.test.ts
jakubkoci: Ahh, I didn't know about that. I updated the JS doc at least :)

jakubkoci:
Thanks, @TimoGlastra for the review and suggestions 👍   PR updated.

TimoGlastra (approved):

PR #283, opened 2021-05-19
fix: convert from buffer now also accepts uint8Array

blu3beri:
Also decodes from uint8Array now instead of only a Buffer.

Necessary for anything that is not in NodeJs.

Signed-off-by: blu3beri <berend@animo.id>


TimoGlastra (approved):
Nice

TimoGlastra:
@jakubkoci we ran into an issue with buffer being transformed to uint8Array over the electron bridge

TimoGlastra:
Failing tests should be fixed by #282

PR #284, opened 2021-05-20
refactor: Rename samples to test

jakubkoci:
The initial goal with `samples` was to represent API and how to use the framework. However, the content of the folder is more about e2e tests so we can rename it. It doesn't change anything on the fact that it still could (and should) be a good place where the user can see examples of how to use it.


README.md
```diff
@@ -87,7 +87,7 @@ yarn install rn-indy-sdk
 
 ### Using the framework
 
-While the framework is still in early development the best way to know what API the framework exposes is by looking at the [tests](src/__tests__), the [source code](src) or the [samples](samples). As the framework reaches a more mature state, documentation on the usage of the framework will be added.
+While the framework is still in early development the best way to know what API the framework exposes is by looking at the [tests](src/__tests__), the [source code](src) or the [test](test). As the framework reaches a more mature state, documentation on the usage of the framework will be added.
```
TimoGlastra, line 90:
```suggestion
While the framework is still in early development the best way to know what API the framework exposes is by looking at the [unit tests](src/__tests__), the [source code](src) or the [e2e tests](test). As the framework reaches a more mature state, documentation on the usage of the framework will be added.
```

package.json
```diff
@@ -14,9 +14,9 @@
     "format": "yarn prettier --write",
     "check-format": "yarn prettier --list-different",
     "test": "jest --verbose",
-    "dev": "ts-node-dev --respawn --transpile-only ./samples/mediator.ts",
-    "prod:start": "node ./build/samples/mediator.js",
-    "prod:start-ws": "node ./build/samples/mediator-ws.js",
+    "dev": "ts-node-dev --respawn --transpile-only ./test/mediator.ts",
```
TimoGlastra, line 17: Maybe we can use the same format as the rest of the codebase?

So `__tests__`?

jakubkoci: Hmm, not sure about that. To me, using `__tests__` has a kind of different purpose. In that case you're collocating tests with your source code files, then it's good to have this convention to express the difference. But at the highest level of hierarchy, you expect various types of folders. Maybe, what I was considering as an alternative is `e2e` or `e2e-tests` 🤷‍♂️

README.md
jakubkoci: Good point I agree with `or the [e2e tests](test)` but I wouldn't call all tests in `src/__tests__` as unit tests.

package.json
TimoGlastra: I get your point. Ok. Maybe `tests` (plural) in that case? More in line with the rest of the codebase

README.md
TimoGlastra: Sure. Just having `test` and `tests` is confusing.

package.json
jakubkoci: Agree 👍  changed

TimoGlastra (approved):

Issue #281, opened 2021-05-19
Custom errors doesn't not work in React Native


jakubkoci:
I found out that `make-error-cause` we're using for custom framework errors doesn't work in React Native. It has dependency on `inspect` function from `util` Node.js module.

I'm sorry, that's my fault. I recommended to use the library because I thought we're using it in another React Native project. However, I found out we're using just `make-error`. 

I'm thinking about how to fix it. One way could be to use `make-error` and just store `cause` error as a property of custom error. Then we wouldn't lose original `error`. The problem could be a missing stack trace.

TimoGlastra:
Ouch. I also didn't take the extra effort to validate it works in RN...

Does `make-error` still provide benefit? The nice thing of `make-error-cause` is that it handled the `cause` part. But if we need to do that ourselves, we can just use the native `Error` rigth?

jakubkoci:
It's a question how we would implement that. It seems that `make-error` provides these [features](https://www.npmjs.com/package/make-error#features). 

Maybe TypeScript helps with some of them already. I found this https://stackoverflow.com/questions/41102060/typescript-extending-error-class.

I'm not sure about preserving stack trace. The `Error.captureStackTrace();` is not available in RN. 

I've made it compile with my fork of `make-error-cause` using [util-inspect](https://github.com/Automattic/util-inspect). But haven't tested yet, I have some dependency injection error now :)

jakubkoci:
So this is the output log with my updates to `make-error-cause`. It seems it works:
```
 LOG  ====================
 LOG  inner error
 ERROR  [SyntaxError: JSON Parse error: Unexpected EOF]
 LOG  outer error
 ERROR  [AriesFrameworkError: Custom Aries Framework message]
 ERROR  [SyntaxError: JSON Parse error: Unexpected EOF]
 LOG  captureStackTrace@http://localhost:8081/index.bundle?platform=android&dev=true&minify=false&app=com.ariesmobileagent&modulesOnly=false&runModule=true:134672:32
BaseError@http://localhost:8081/index.bundle?platform=android&dev=true&minify=false&app=com.ariesmobileagent&modulesOnly=false&runModule=true:134714:22
BaseError@http://localhost:8081/index.bundle?platform=android&dev=true&minify=false&app=com.ariesmobileagent&modulesOnly=false&runModule=true:133288:30
construct@[native code]
[...]

PR #285, opened 2021-05-20
fix: Use custom make-error with cause that works in RN

jakubkoci:
I forked [make-error-cause](https://github.com/blakeembrey/make-error-cause) and replaced Node.js `inspect` dependency with [util-inspect](https://github.com/Automattic/util-inspect). I'm not saying this is the final solution but it should make it work at least for now (I tested it with RN).

The issue describing the problem is #281.


TimoGlastra:
The library is one file. Can't we just copy it into the framework?

TimoGlastra:
It's apache. So if we add a copyright notice of the original author we're fine

jakubkoci:
> `util-inspect` doesn't have a license which means we legally can't use it.
> 
> https://choosealicense.com/no-permission/
> https://github.com/Automattic/util-inspect

🤦‍♂️ 😄 

I found a different module [object-inspect](https://www.npmjs.com/package/object-inspect), works in React Native, MIT license.

I realize that the `inspect` method is actually used only when you specifically call `error.inspect()` and it doesn't provide any benefits in common scenarios.

TimoGlastra (approved):
Nice! A lot less dependencies also

PR #288, opened 2021-05-20
Created default mediator tag & wallet retrieval

DavidClawson:
Signed-off-by: David Clawson <david.clawson@gmail.com>

### DIFF
--- a/src/modules/routing/services/RecipientService.ts
+++ b/src/modules/routing/services/RecipientService.ts
@@ -39,6 +39,15 @@ export class RecipientService {
     this.eventEmitter = eventEmitter
   }
 
+  public async init() {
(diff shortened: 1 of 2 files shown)


src/modules/routing/services/RecipientService.ts
```diff
@@ -39,6 +39,15 @@ export class RecipientService {
     this.eventEmitter = eventEmitter
   }
 
+  public async init() {
```
burdettadam, line 42: do you call this anywhere?

burdettadam (changes requested):

src/modules/routing/services/RecipientService.ts
DavidClawson: Just fixed it. Thanks!

burdettadam (approved):

PR #293, opened 2021-05-25
feat: add isInitialized agent property

TimoGlastra:


JamesKEbert (approved):
I like this addition--thanks

PR #294, opened 2021-05-26
feat: support node v12+

TimoGlastra:
I updated the indy-sdk to support node v12+: https://github.com/hyperledger/indy-sdk/pull/2389

This PR loosens the node engine requirement to support 12+ and also adds a test martrix to test for node 12, 14 and 16.

Prettier had some updates to their defaults I think. that's where the other changes come from


package.json
```diff
@@ -58,7 +58,7 @@
     "eslint-plugin-prettier": "^3.3.1",
     "express": "^4.17.1",
     "husky": "^5.1.3",
-    "indy-sdk": "^1.16.0",
+    "indy-sdk": "^1.16.0-dev-1633",
```
JamesKEbert, line 61: Can I ask what the impact of this upgrade is?

TimoGlastra: Sure. The impact should be minimal. This new version of indy-sdk supports Node 12+ and drops support for NodeJS 8. Node 8 is already end of life for a while and we have never supported it. So in the context of AFJ it just means we support newer NodeJS versions.

However as developers install the indy-sdk themselves they are not required to update to this new version. If they want to use an older version of indy-sdk they can do so. But I don't think there should be any reason not to update. There's nothing functionally different in this indy-sdk version

I think it is nice we can use the Current and LTS NodeJS versions again.

JamesKEbert: Awesome, fully agree with the thought processes here. Thanks for the clarification.

JamesKEbert (approved):

Issue #268, opened 2021-05-13
Add support for multiple transports


jakubkoci:
As a developer, I want to add more inbound/outbound transporters, so that an agent can receive/send messages from/to various transport protocols

I would just implement it in a way that we change properties inboundTransporter, outboundTransporter. We call `start` method for each inbound transporter. 

Similarly, for outboundTransporter:
a) we call `sendMessage` for each of them, but first, we should perhaps check if a transport `canReply` according to information in `outboundMessageContext`.
b) we call `sendMessage` for each of them and process any error to find out if it `canReply`

By doing this in `MessageSender` we can remove the majority of the logic from `resolveTransport`. `TransportService` will just return if there is any inbound transport available that could be used for outbound communication.

TimoGlastra:
Duplicate of #101

TimoGlastra:
> a) we call sendMessage for each of them, but first, we should perhaps check if a transport canReply according to information in

I'm not sure If I understand. Do you want to call `sendMessage` on all transporters and check if it returns true (or something similar?) so we know the message has been delivered?

jakubkoci:
> Duplicate of #101

Ouch, I'm sorry I overlooked that.

jakubkoci:
I thought we could add `canReply(outboundMessageContext)` method, call it before `sendMessage` like a filtering mechanism.

Just an idea. The problem with this solution is that the `sendMessage` called later can fail even after `canReply` returning `true`.  But then that failure should be caused by some other error.

For example, for WebSocket transport:
```
canReply(outboundCtx) {
  if (endpoint.startsWith('ws')) {
    return true
  }
}
```

There are also other questions that I haven't thought through yet. Should we try to send a message via all available transports? Or only via the first one which succeeds?

TimoGlastra:
> I thought we could add canReply(outboundMessageContext) method, call it before sendMessage like a filtering mechanism.

I think that's a good approach. I'm not sure if `canReply` is the best term. Maybe we're not replying but just sending a messing.

> Just an idea. The problem with this solution is that the sendMessage called later can fail even after canReply returning true. But then that failure should be caused by some other error.

Yeah so it won't guarantee delivery, it just checks that it is theoretically possible to send the message. E.g. WS transport cannot deliver to HTTP endpoint.

> Should we try to send a message via all available transports? Or only via the first one which succeeds?

Only the first one, sorted by priority of available services in the diddoc. Otherwise we'll send the message two (or more) times. 

[...]

PR #304, opened 2021-06-03
feat: Pack and send message based on DidDoc services

jakubkoci:
Related to #268

### DIFF
--- a/src/agent/Dispatcher.ts
+++ b/src/agent/Dispatcher.ts
@@ -45,13 +45,12 @@ class Dispatcher {
 
       // Check for return routing, with thread id
       if (message.hasReturnRouting(threadId)) {
-        // Find service with highest priority and keys to pack message
-        const [service] = this.transportService.findDidCommServices(outboundMessage.connection)
-        if (!service) {
-          throw new AriesFrameworkError(`Connection with id ${outboundMessage.connection.id} has no service!`)
+        const keys = {
+          recipientKeys: messageContext.senderVerkey ? [messageContext.senderVerkey] : [],
+          routingKeys: [],
+          senderKey: messageContext.connection?.verkey || null,
         }
-        // TODO Should we validate inbound transport session scheme and service endpoint scheme?
-        return await this.messageSender.packMessage(outboundMessage, service)
+        return await this.messageSender.packMessage(outboundMessage, keys)
       }
 
       await this.messageSender.sendMessage(outboundMessage)
--- a/src/agent/Dispatcher.ts
+++ b/src/agent/Dispatcher.ts
@@ -46,7 +46,7 @@ class Dispatcher {
       // Check for return routing, with thread id
       if (message.hasReturnRouting(threadId)) {
         // Find service with highest priority and keys to pack message
-        const [service] = this.transportService.findServices(outboundMessage.connection)
+        const [service] = this.transportService.findDidCommServices(outboundMessage.connection)
         if (!service) {
           throw new AriesFrameworkError(`Connection with id ${outboundMessage.connection.id} has no service!`)
         }
--- a/src/agent/Dispatcher.ts
+++ b/src/agent/Dispatcher.ts
@@ -43,9 +43,15 @@ class Dispatcher {
         outboundMessage.payload.setReturnRouting(ReturnRouteTypes.all)
       }
 
-      // check for return routing, with thread id
+      // Check for return routing, with thread id
       if (message.hasReturnRouting(threadId)) {
-        return await this.messageSender.packMessage(outboundMessage)
+        // Find service with highest priority and keys to pack message
+        const [service] = this.transportService.findServices(outboundMessage.connection)
+        if (!service) {
+          throw new AriesFrameworkError(`Connection with id ${outboundMessage.connection.id} has no service!`)
+        }
+        // TODO Should we validate inbound transport session scheme and service endpoint scheme?
+        return await this.messageSender.packMessage(outboundMessage, service)
       }
 
       await this.messageSender.sendMessage(outboundMessage)
--- a/src/agent/EnvelopeService.ts
+++ b/src/agent/EnvelopeService.ts
@@ -1,5 +1,7 @@
 import type { Logger } from '../logger'
-import type { OutboundMessage, UnpackedMessageContext } from '../types'
+import type { UnpackedMessageContext } from '../types'
+import type { AgentMessage } from './AgentMessage'
+import type { Verkey } from 'indy-sdk'
 
 import { inject, scoped, Lifecycle } from 'tsyringe'
 
@@ -8,6 +10,12 @@ import { ForwardMessage } from '../modules/routing/messages'
 import { Wallet } from '../wallet/Wallet'
 
 import { AgentConfig } from './AgentConfig'
+
+interface Keys {
+  recipientKeys: Verkey[]
+  routingKeys: Verkey[]
+  senderVk: Verkey | null
+}
 
 @scoped(Lifecycle.ContainerScoped)
 class EnvelopeService {
@@ -19,8 +27,8 @@ class EnvelopeService {
     this.logger = agentConfig.logger
   }
 
-  public async packMessage(outboundMessage: OutboundMessage): Promise<JsonWebKey> {
-    const { routingKeys, recipientKeys, senderVk, payload } = outboundMessage
+  public async packMessage(keys: Keys, payload: AgentMessage): Promise<JsonWebKey> {
+    const { routingKeys, recipientKeys, senderVk } = keys
     const message = payload.toJSON()
 
     let wireMessage = await this.wallet.pack(message, recipientKeys, senderVk)
--- a/src/agent/EnvelopeService.ts
+++ b/src/agent/EnvelopeService.ts
@@ -11,7 +11,7 @@ import { Wallet } from '../wallet/Wallet'
 
 import { AgentConfig } from './AgentConfig'
 
-interface Keys {
+export interface EnvelopeKeys {
   recipientKeys: Verkey[]
   routingKeys: Verkey[]
   senderKey: Verkey | null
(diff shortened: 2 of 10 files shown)


src/agent/TransportService.ts
```diff
@@ -29,23 +28,24 @@ export class TransportService {
     return this.transportSessionTable[connectionId]
   }
 
-  public findEndpoint(connection: ConnectionRecord) {
+  public findServices(connection: ConnectionRecord): DidCommService[] {
```
TimoGlastra, line 31: I'd like to be more verbose:

```suggestion
  public findDidCommServices(connection: ConnectionRecord): DidCommService[] {
```

src/agent/__tests__/MessageSender.test.ts
```diff
@@ -47,50 +47,93 @@ describe('MessageSender', () => {
...
+      expect(sendMessageCallArg).toEqual({
         connection,
         payload: wireMessage,
-        endpoint,
+        endpoint: firstDidCommService.serviceEndpoint,
         responseRequested: false,
         session,
       })
+      expect(sendMessageSpy.mock.calls.length).toEqual(1)
+    })
```
TimoGlastra, line 117: I prefer to use:

`expect(sendMessageSpy).toHaveBeenCalledNthWith()` (not sure if the syntax is 100% correct.

src/agent/helpers.ts
```diff
@@ -1,39 +1,13 @@
 import { ConnectionRecord } from '../modules/connections'
 import { AgentMessage } from './AgentMessage'
 import { OutboundMessage } from '../types'
-import { ConnectionInvitationMessage } from '../modules/connections'
-import { AriesFrameworkError } from '../error'
 
 export function createOutboundMessage<T extends AgentMessage = AgentMessage>(
```
TimoGlastra, line 5: This method seems pretty useless now. It takes in two arguments and returns it as an object.

I'm generally not a fan of this method. Rather have OutboundMessage be a class instance in that case

src/agent/Dispatcher.ts
```diff
@@ -40,9 +40,15 @@ class Dispatcher {
...
+      // Check for return routing, with thread id
       if (message.hasReturnRouting(threadId)) {
-        return await this.messageSender.packMessage(outboundMessage)
+        // Find service with highest priority and keys to pack message
+        const [service] = this.transportService.findServices(outboundMessage.connection)
+        if (!service) {
+          throw new AriesFrameworkError(`Connection with id ${outboundMessage.connection.id} has no service!`)
+        }
+        // TODO Should we validate inbound transport session scheme and service endpoint scheme?
+        return await this.messageSender.packMessage(outboundMessage, service)
```
TimoGlastra, line 51: This doesn't make sense to me. Why are we using the first service entry for return routing? We should use the same keys that were used for receiving the message.

TimoGlastra: Maybe packMessage shouldn't take a service as input, as you can pack messages without a service. You just need the keys

src/agent/EnvelopeService.ts
```diff
@@ -1,10 +1,19 @@
...
+
+import { UnpackedMessageContext } from '../types'
 import { Wallet } from '../wallet/Wallet'
 import { ForwardMessage } from '../modules/routing/messages'
 import { AgentConfig } from './AgentConfig'
 import { Logger } from '../logger'
 import { Symbols } from '../symbols'
+import { AgentMessage } from './AgentMessage'
+
+interface Keys {
```
TimoGlastra, line 12: I'd prefer a different name that Keys. Maybe something more verbose? 

`RecipientTarget`?

src/agent/EnvelopeService.ts
```diff
@@ -1,10 +1,19 @@
...
+  recipientKeys: Verkey[]
+  routingKeys: Verkey[]
+  senderVk: Verkey | null
```
TimoGlastra, line 15: Consistency:

```suggestion
  senderVerkey: Verkey | null
```
OR

```suggestion
  senderKey: Verkey | null
```

src/agent/MessageSender.ts
```diff
@@ -33,23 +34,50 @@ export class MessageSender {
...
+        } catch (error) {
+          this.logger.debug(
+            `Sending outbound message to service with id ${service.id} failed with the following error:`,
+            {
+              error,
+            }
+          )
+        }
+      }
+    }
```
TimoGlastra, line 81:
```suggestion
    for (const service of services) {
      this.logger.debug(`Trying to send outbound message to service:`, { service })
        try {
          const outboundPackage = await this.packMessage(outboundMessage, service)
          outboundPackage.session = this.transportService.findSession(outboundMessage.connection.id)
          await this.outboundTransporter.sendMessage(outboundPackage)
          
          break
          // Or return
        } catch (error) {
          this.logger.debug(
            `Sending outbound message to service with id ${service.id} failed with the following error:`,
            {
              error,
            }
          )
      }
    }
```

TimoGlastra: Not a fan of the `success` property

src/agent/MessageSender.ts
```diff
@@ -33,23 +34,50 @@ export class MessageSender {
...
+          const outboundPackage = await this.packMessage(outboundMessage, service)
+          outboundPackage.session = this.transportService.findSession(outboundMessage.connection.id)
+          await this.outboundTransporter.sendMessage(outboundPackage)
+          success = true
+        } catch (error) {
+          this.logger.debug(
+            `Sending outbound message to service with id ${service.id} failed with the following error:`,
+            {
+              error,
+            }
```
TimoGlastra, line 77: This can result in cryptic errors, also adding the message can help
```suggestion
            {
              error,
              errorMessage: error.message
            }
```

TimoGlastra (changes requested):

src/agent/EnvelopeService.ts
TimoGlastra: Also I think it should be exported if it is going to be used by external services

src/agent/TransportService.ts
jakubkoci: Good point 👍

src/agent/helpers.ts
jakubkoci: Yes, we can eventually get rid of this function. Perhaps by having something like `OutboundMessageContext`.

src/agent/__tests__/MessageSender.test.ts
jakubkoci: Ah. That's right, it's much better. I don't know why I'm not using that 🤷‍♂️

src/agent/EnvelopeService.ts
jakubkoci: Hmm, I don't see `RecipientTarget` as more expressive 🤔  The interface represents a set of keys for packing a message. You can derive from context (it's used in `EnvelopeService` and `packMessage` method) the purpose of this object.

Right now, it seems it doesn't have to be exported. I don't like exporting stuff upfront. Or, can you give me an example of how you see the interface can be used externally?

TimoGlastra: > Hmm, I don't see RecipientTarget as more expressive

I'm fine with another name, however keys is just so generic. It's not just keys. It's public keys used for didcomm messaging

TimoGlastra: > Right now, it seems it doesn't have to be exported. I don't like exporting stuff upfront. Or, can you give me an example of how you see the interface can be used externally?

IMO all types used in public methods should be exported. As `packMessage` is public and uses the `Keys` type I think it should be exported. I find it very annoying when libraries have beautifully defined interfaces for public methods that I can't use (without a lot of typescript quirks) because they are exported. As the framework is extensible, I think it should be exported.

Maybe this is something we can solve with the monorepo:
- `@aries-framework/agent` -> all the normal public stuff. Does not export more than minimally required
- `@aries-framework/core` -> does export more than minimally needed, allows you to import types needed for extension

src/agent/MessageSender.ts
jakubkoci: Ok, no problem. What do you mean by "cryptic error", do you have an example? I would assume that you can read the message from the error object 🤔

src/agent/MessageSender.ts
jakubkoci: I also don't feel good about the `success` variable, but I don't consider the `break` statement as a much better solution. It doesn't tell you the what's the intent of breaking the loop. However, I see a benefit in removing the condition and variable so I can make the change.

src/agent/EnvelopeService.ts
jakubkoci: Good catch, updated.

src/agent/Dispatcher.ts
jakubkoci: Yes, I'm gradually refactoring this method. Now it's just about encryption of payload with keys. I just kept interface/signature of the method until I replace `createOutboundMessage` with OutboundMessageContext and update interface of `Agent` class.

jakubkoci: Usage of service has been removed.

jakubkoci:
@TimoGlastra I think I updated almost everything. As a next step, I'm going to add `sendMessage` to `TransportSession` interface allowing us to send a message and do not return the outbound message to inbound transport. I can perhaps replace `createOutboundMessage` with `OutboundMessageContext`

src/agent/MessageSender.ts
TimoGlastra: I think this is the case in Electron and React Native. In electron because we transfer the error over the bridge, meaning it is transformed from an IndyError to a general Error. In React Native (I think) because we're not actually creating an error but rather an object that looks like an error (https://github.com/AbsaOSS/rn-indy-sdk/blob/master/src/index.js#L873-L908, we should fix this)

TimoGlastra: Maybe the things I mentioned are totally unrelated. But I've had trouble with the error message appearing until I explicitly used error.message

TimoGlastra (approved):
Awesome. Great work @jakubkoci :)

> I can perhaps replace createOutboundMessage with OutboundMessageContext

Agree

src/agent/Dispatcher.ts
```diff
@@ -43,9 +43,14 @@ class Dispatcher {
         outboundMessage.payload.setReturnRouting(ReturnRouteTypes.all)
       }
 
-      // check for return routing, with thread id
+      // Check for return routing, with thread id
       if (message.hasReturnRouting(threadId)) {
-        return await this.messageSender.packMessage(outboundMessage)
+        const keys = {
+          recipientKeys: messageContext.senderVerkey ? [messageContext.senderVerkey] : [],
```
TimoGlastra, line 49: BTW -- what happens when there are no recipient keys

PR #315, opened 2021-06-09
fix: export module classes from framework root

TimoGlastra:
Exports the module classes from the framework root. We've noticed that you oftentimes need the classes, with either dependency injection, or as TS type. 

I know you've pushed back on this before @jakubkoci, but currently we need to import like this:

```
import { ConnectionsModule } from 'aries-framework/build/src/modules/connections/ConnectionsModule.ts`
```


jakubkoci (approved):
I'm good with this 👍  It makes sense because it's part of the framework's public API. I'm just against exporting everything from everything :) 

I'm also not sure if it's good to import one `XxxModule` class from another. But if it's called from outside of the framework, definitely all good.

TimoGlastra:
Yeah I agree, I'll see if I can maybe go over all exports an do more specific exports instead of `* from`.

PR #316, opened 2021-06-09
style: forbid cyclic import, enforce type imports

TimoGlastra:
It seems the CI of #313 is failing due to a cyclic import. We've had an issue with this before, so time to fix it properly

This PR fixes all cyclic imports and adds an eslint rule that prohibits cyclic imports.

Main reason we had so much cyclic imports was because of type only imports. TS added something exactly for this in TS 3.8 using `import type {} from ''`. I also added an eslint rule to enfore type imports. I.e. if an import is only used as type it must be imported using `import type` syntax. This is autofixable by eslint.

Thank god for eslint auto fixer


TimoGlastra:
Lol, that's why our test coverage was so high, from 88 to 56...

jakubkoci (approved):
Ouch. Good job Timo with fixing this. 

The coverage is not so bad imho but a nice example of how coverage can be misleading sometimes

PR #317, opened 2021-06-10
ci: cancel old job, require commits for release 

TimoGlastra:
This fixes a require cycle that got into main.

It also adds some extra guards to hopefully prevent race conditions with parallel merges. I discovered what the problem is, just now sure how to fix it yet: https://github.com/release-it/release-it/issues/773


JamesKEbert (approved):
LGTM

PR #320, opened 2021-06-11
feat: adds support for linked attachments

blu3beri:
Implements [0641](https://github.com/hyperledger/aries-rfcs/pull/641).

It allows for attachments outside of the credential to be linked to a credential. This means that it will be way less resource intensive as it does not have to sign the entire attachment, but just the link (a sub 100 char string).

For now a default threshold has been set at 5 megabytes inside the mediator as with bigger images it would just get too cumbersome and as stated in the [attachment RFC](https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0017-attachments/README.md#size-considerations), messages should remain small. Attachments above 5 megabytes are possible, however they have to be hosted and the link(s) have to be supplied in the linked attachment object. 

this will help a lot of mobile agents as after some testing with larger attachments (1mb+) in the [Animo react native wallet](https://github.com/animo/aries-mobile-agent-react-native), it crashed.

Setting of the threshold might be better to do somewhere else, but since the mediator is external it might be not something worth implementing.


src/__tests__/proofs.test.ts
```diff
@@ -120,6 +126,22 @@ describe('Present Proof', () => {
         credentialDefinitionId: credDefId,
         comment: 'some comment about credential',
         preview: credentialPreview,
+        linkedAttachments: [
+          new LinkedAttachment({
+            name: 'image_0',
```
TimoGlastra, line 131: maybe `attributeName` would be more descriptive over `name`?

src/modules/credentials/CredentialUtils.ts
```diff
@@ -1,10 +1,40 @@
...
+   * Adds a credential to the preview where it links to the attachment
+   *
+   * @param attachments a list of the attachments that need to be linked to a credential
+   * @param preview the credential previews where the new linked credential has to be appended to
+   *
+   * @returns a modified version of the credential preview with the linked credentials
+   * */
+  public static createAndLinkCredentialToAttachment(attachments: LinkedAttachment[], preview: CredentialPreview) {
+    const credentialPreview = new CredentialPreview({ attributes: [...preview.attributes] })
+    attachments?.forEach((linkedAttachment) => {
```
TimoGlastra, line 23: remove ?
```suggestion
    attachments.forEach((linkedAttachment) => {
```

src/modules/credentials/CredentialUtils.ts
```diff
@@ -1,10 +1,40 @@
...
+      if (credentialPreview.attributes.map((attribute) => attribute.name).includes(linkedAttachment.name)) {
```
TimoGlastra, line 24: You're mapping the attributes in every loop. Maybe move outside of the for loop?

src/modules/credentials/CredentialUtils.ts
```diff
@@ -1,10 +1,40 @@
...
 export class CredentialUtils {
+  /**
+   * Adds a credential to the preview where it links to the attachment
+   *
+   * @param attachments a list of the attachments that need to be linked to a credential
+   * @param preview the credential previews where the new linked credential has to be appended to
+   *
+   * @returns a modified version of the credential preview with the linked credentials
+   * */
+  public static createAndLinkCredentialToAttachment(attachments: LinkedAttachment[], preview: CredentialPreview) {
```
TimoGlastra, line 21: Maybe?

```suggestion
  public static createAndLinkAttachmentsToPreview(attachments: LinkedAttachment[], preview: CredentialPreview) {
```


Also the meaning of the following sentence is not fully clear to me. I wouldn't use credentials here, but rather attachments and preview

```
Adds a credential to the preview where it links to the attachment
```

src/modules/credentials/CredentialsModule.ts
```diff
@@ -85,6 +84,8 @@ export class CredentialsModule {
 
     const credentialDefinitionId = config?.credentialDefinitionId ?? credentialProposalMessage.credentialDefinitionId
 
+    credentialRecord.attachments = credentialProposalMessage.attachments
```
TimoGlastra, line 87: I think we only want to save the linked attachments

```suggestion
    credentialRecord.linkedAttachments = credentialProposalMessage.attachments.filter(....)
```

src/modules/credentials/repository/CredentialRecord.ts
```diff
@@ -98,6 +104,7 @@ export class CredentialRecord extends BaseRecord implements CredentialStoragePro
 
     return new CredentialInfo({
       claims,
+      attachments: this.attachments ? this.attachments : undefined,
```
TimoGlastra, line 107: I think we can just use this.attachments. (as it will be undefined if it has not value)
```suggestion
      attachments: this.attachments,
```

Otherwise

```suggestion
      attachments: this.attachments ?? undefined,
```

src/modules/credentials/services/CredentialService.ts
```diff
@@ -73,11 +75,20 @@ export class CredentialService {
...
+    config?: CredentialProposeOptions
   ): Promise<CredentialProtocolMsgReturnType<ProposeCredentialMessage>> {
     // Assert
     connectionRecord.assertReady()
 
+    // Add the linked attachments to the credentialProposal
+    if (config?.linkedAttachments) {
+      config.credentialProposal = CredentialUtils.createAndLinkCredentialToAttachment(
+        config.linkedAttachments,
+        config.credentialProposal ? config.credentialProposal : new CredentialPreview({ attributes: [] })
```
TimoGlastra, line 87:
```suggestion
        config.credentialProposal ?? new CredentialPreview({ attributes: [] })
```

src/modules/credentials/services/CredentialService.ts
```diff
@@ -73,11 +75,20 @@ export class CredentialService {
...
     connectionRecord: ConnectionRecord,
-    config?: Omit<ProposeCredentialMessageOptions, 'id'>
+    config?: CredentialProposeOptions
   ): Promise<CredentialProtocolMsgReturnType<ProposeCredentialMessage>> {
     // Assert
     connectionRecord.assertReady()
 
+    // Add the linked attachments to the credentialProposal
+    if (config?.linkedAttachments) {
+      config.credentialProposal = CredentialUtils.createAndLinkCredentialToAttachment(
```
TimoGlastra, line 85: I'd use separate variables instead of modifying the input `config` object. This can result in unexpected behaviour for the framework user

src/modules/credentials/services/CredentialService.ts
```diff
@@ -86,6 +97,7 @@ export class CredentialService {
       connectionId: connectionRecord.id,
       state: CredentialState.ProposalSent,
       proposalMessage,
+      attachments: config?.linkedAttachments?.map((linkedAttachment) => linkedAttachment.attachment),
```
TimoGlastra, line 100: I think attachments should be called linkedAttachments. Maybe also store the whole `LinkedAttachment` object?

```suggestion
      linkedAttachments: config?.linkedAttachments?.map((linkedAttachment) => linkedAttachment.attachment),
```

src/modules/credentials/services/CredentialService.ts
```diff
@@ -205,20 +217,23 @@ export class CredentialService {
     credentialRecord.assertState(CredentialState.ProposalReceived)
 
     // Create message
-    const { credentialDefinitionId, comment, preview } = credentialTemplate
+    const { credentialDefinitionId, comment, preview, attachments } = credentialTemplate
     const credOffer = await this.indyIssuerService.createCredentialOffer(credentialDefinitionId)
-    const attachment = new Attachment({
+    const offerAttachments = new Attachment({
```
TimoGlastra, line 222:
```suggestion
    const offerAttachment = new Attachment({
```

src/modules/credentials/services/CredentialService.ts
```diff
@@ -249,26 +264,35 @@ export class CredentialService {
...
+      : preview
+
+    // Construct offer message
     const credentialOfferMessage = new OfferCredentialMessage({
       comment,
-      offerAttachments: [attachment],
-      credentialPreview: preview,
+      offerAttachments: [offerAttachment],
+      credentialPreview,
+      attachments: linkedAttachments?.map((linkedAttachment) => linkedAttachment.attachment),
```
TimoGlastra, line 287: Just a thought: Instead of attachments, use `appendedAttachments`? This is how it is called in the Attachments RFC. https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0017-attachments/README.md#appending

However I'm not sure about this one. What do you think?

src/modules/credentials/services/CredentialService.ts
```diff
@@ -403,12 +429,15 @@ export class CredentialService {
     const { comment } = options
     const credentialRequest = new RequestCredentialMessage({
       comment,
-      requestAttachments: [attachment],
+      requestAttachments: [requestAttachment],
+      attachments: credentialRecord.offerMessage?.attachments,
```
TimoGlastra, line 433: This seems wrong. the attachments array can contain any attachment, not just linked attachments. We should only send the linked attachments

src/modules/credentials/services/CredentialService.ts
```diff
@@ -403,12 +429,15 @@ export class CredentialService {
...
     })
+
     credentialRequest.setThread({ threadId: credentialRecord.tags.threadId })
 
     credentialRecord.metadata.requestMetadata = credReqMetadata
     credentialRecord.requestMessage = credentialRequest
+    credentialRecord.attachments = credentialRecord.offerMessage?.attachments
```
TimoGlastra, line 440: You're not filtering the attachments here

src/utils/HashlinkEncoder.ts
```diff
@@ -37,17 +40,16 @@ export class HashlinkEncoder {
    */
   public static encode(
     buffer: Buffer | Uint8Array,
-    hashAlgorithm: 'sha2-256',
-    baseEncoding: BaseName = 'base58btc',
-    metadata?: Metadata
+    metadata?: Metadata,
+    hashAlgorithm: 'sha2-256' = 'sha2-256',
```
TimoGlastra, line 44: I'd like to not make `hashAlgorithm` optional. This makes it easier in the future when we support multiple hashing algorithms

src/utils/HashlinkEncoder.ts
```diff
@@ -37,17 +40,16 @@ export class HashlinkEncoder {
...
+    baseEncoding: BaseName = 'base58btc'
   ) {
-    const checksum = this.encodeMultihashEncoder(buffer, hashAlgorithm, baseEncoding)
-    const mbMetadata = metadata ? this.encodeMetadata(metadata, baseEncoding) : null
+    const checksum = this.encodeMultiHashEncoder(buffer, hashAlgorithm, baseEncoding)
```
TimoGlastra, line 47: This name seems weird. We don't want to encode the multiHashEncoder right?

```suggestion
    const checksum = this.encodeMultiHash(buffer, hashAlgorithm, baseEncoding)
```

src/utils/HashlinkEncoder.ts
```diff
@@ -68,35 +67,49 @@ export class HashlinkEncoder {
...
+  ) {
+    if (attachment.data.sha256) {
+      return `hl:${attachment.data.sha256}`
+    } else if (attachment.data.base64) {
+      return this.encode(BufferEncoder.fromBase64(attachment.data.base64), {}, hashAlgorithm, baseName)
+    } else if (attachment.data.json) {
+      return this.encode(JsonEncoder.toBuffer(attachment.data.json), {}, hashAlgorithm, baseName)
+    } else {
+      throw new AriesFrameworkError(`Attachment: (${attachment.id}) has no data to create a link with`)
+    }
```
TimoGlastra, line 101: This doesn't seem like the right place. Would rather see this in an attachment util or something

src/utils/HashlinkEncoder.ts
```diff
@@ -68,35 +67,49 @@ export class HashlinkEncoder {
...
   }
 
-  private static encodeMultihashEncoder(
+  private static encodeMultiHashEncoder(
     buffer: Buffer | Uint8Array,
-    hashName: 'sha2-256',
-    baseEncoding: BaseName
+    hashName: 'sha2-256' = 'sha2-256',
+    baseEncoding: BaseName = 'base58btc'
```
TimoGlastra, line 107: Same, would rather keep it explicit

src/utils/MultiHashEncoder.ts
```diff
@@ -10,7 +10,7 @@ export class MultihashEncoder {
    *
    * @returns a multihash
    */
-  public static encode(buffer: Uint8Array, hashName: 'sha2-256'): Uint8Array {
+  public static encode(buffer: Uint8Array, hashName: 'sha2-256' = 'sha2-256'): Uint8Array {
```
TimoGlastra, line 13: Same

TimoGlastra (changes requested):
Overall looks good, just some suggestions :)

src/modules/credentials/CredentialsModule.ts
blu3beri: The attachments in the `credentialProposalMessage` are just regular `attachments` and not `linkedAttachments`. I think storing them as regular attachments is a better solution as it can also help with other use cases.

src/modules/credentials/services/CredentialService.ts
blu3beri: I chose to store the attachments only and not the linkedAttachment as I felt like it is redundant and does not improve usability. And if they are not `linkedAttachments` then I would just like to call them attachments to stay consistent.

src/modules/credentials/services/CredentialService.ts
blu3beri: Not the biggest fan. The method as discribed in the RFC is indeed appending, however every element in a list is appended in a way so the list already infers that it is using the appending method from the RFC.

src/modules/credentials/services/CredentialService.ts
blu3beri: I am not quite sure what you mean by this. If you mean that `credentialRecord.offerMessage?.attachments` can contain attachments that are not linked, then I think it should be fixed when the attachments are appended to the `offerMessage`. However, why not any attachment?

src/modules/credentials/services/CredentialService.ts
blu3beri: The only way to get an `attachment` in the offerMessage is if it were a `linkedAttachment` before. So it seems a bit redundant.

src/modules/credentials/services/CredentialService.ts
TimoGlastra: It is not standard practice to include all attachments in all messages. Different use cases require different flows. One of them could be adding an attachment to the offer message only. I would find it weird as a framework user if it is then automatically sent in the credential request. That should be a manual step.

I want to keep the logic of handling linked attachments (which have the process defined) over general attachments (which are general purpose, and don't have the process defined)

src/modules/credentials/services/CredentialService.ts
TimoGlastra: That's not true. You could have received a message with any attachments.

TimoGlastra: Or if that's true, then it should be called `linkedAttachments`, not `attachments`

TimoGlastra: Maybe we're having a misunderstanding here. I'm not saying it should be `LinkedAttachment` class instance in this case. I'm saying that `Attachment` could be meaning a linked attachment, or a non-linked attachment

src/modules/credentials/services/CredentialService.ts
blu3beri: Okay yeah that sounds good.

src/modules/credentials/services/CredentialService.ts
blu3beri: So, the attribute in the credentialRecord should be called `linkedAttachments` and contains instances of the `Attachment` class? I am fine either way, I just chose this name to stay consistent with the classname.

TimoGlastra: I think so. I agree it's consistent with the class name, but it is leaving an important part of the context out. Namely that the `Attachments` in that array are only attachments used for linked attachments, not general attachments

src/modules/proofs/services/ProofService.ts
```diff
@@ -395,6 +398,12 @@ export class ProofService {
       )
     }
 
+    // Get the matching attachments to the requested credentials
+    const attachments = await this.getRequestedAttachmentsForRequestedCredentials(
+      indyProofRequest,
+      requestedCredentials
+    )
```
TimoGlastra, line 405: Would it be possible to check beforehand whether this concerns linked attachments? There are not used that commonly, but this method always retrieves records for proofs. Would be nice to know beforehand if the retrieval of records for attachments is necessary

src/utils/HashlinkEncoder.ts
TimoGlastra: This doesn't seem resolved @blu3beri

src/utils/MultiHashEncoder.ts
TimoGlastra: Same

src/modules/proofs/services/ProofService.ts
blu3beri: Unfortunately, it can only be known from the `credential attribute value`. It does indeed feel quite inefficient, but the only alternative I could think is manually sending an attachment object when you call `createPresentation`.

TimoGlastra (approved):

src/utils/attachment.ts
```diff
@@ -0,0 +1,42 @@
...
+  hashAlgorithm: 'sha2-256' = 'sha2-256',
+  baseName: BaseName = 'base58btc'
+) {
+  if (attachment.data.sha256) {
+    return `hl:${attachment.data.sha256}`
+  } else if (attachment.data.base64) {
+    return HashlinkEncoder.encode(BufferEncoder.fromBase64(attachment.data.base64), hashAlgorithm, baseName)
+  } else if (attachment.data.json) {
+    return HashlinkEncoder.encode(JsonEncoder.toBuffer(attachment.data.json), hashAlgorithm, baseName)
+  } else {
```
TimoGlastra, line 29: Hashing JSON requires a canonicalization algorithm to consistently go from JSON to string/buffer. The `jws` property also doesn't support the `data.json` field at the moment. From the attachments RFC:

> Signatures over inlined JSON attachments are not currently defined as this depends upon a canonical serialization for the data.

We could use JSON Canonicalization Scheme (JCS) but that should also be added to the attachments RFC in that case.

I think for now we can just not support json for linked attachments?

blu3beri: Yeah that sounds good. I did not think of that, I will remove it right away.

PR #322, opened 2021-06-13
build: move to monrepo

TimoGlastra:
Basically ready to be merged. 

Task list:
- [x] Split into `@aries-framework/core`, `@aries-framework/node` & `@aries-framework/react-native`
- [x] Tests working with new structure
- [x] Linting, prettier with monorepo
- [x] Automatic release script with lerna instead of release-it
- [x] update NPM token (to aries-framework org instead of hyperledger)


TimoGlastra:
Ready for review! I've coordinated with Ry to get everything set up for automatic releases so if we merge it we should automatically release under the new branch.

This PR doesn't introduce a lot of changes in how things works, except for the extraction of platform specific dependencies. Those are now the second parameter to agent.

```
import { Agent } from '@aries-framework/core'
import { agentDependencies } from '@aries-framework/node'

const config = {}
const agent = new Agent(config, agentDependencies)
```

packages/react-native/tsconfig.build.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": "../../tsconfig.build.json",
+
+  "compilerOptions": {
+    "outDir": "./build",
+    // FIXME https://github.com/hyperledger/aries-framework-javascript/pull/327
```
JamesKEbert, line 6: Available for fixing?

tests/e2e.test.ts
```diff
@@ -14,15 +14,16 @@ import {
   PredicateType,
   CredentialState,
   ProofState,
-} from '../src'
+} from '../packages/core/src'
```
JamesKEbert, line 17: Could change to from `from '@aries-framework/core'`?

JamesKEbert (approved):
This looks good to me--there's a lot to look through on this, I left a couple of comments, but I may be missing some context behind the thoughts.
Thanks Timo for all of the effort here.

docs/setup-react-native.md
```diff
@@ -77,20 +77,18 @@ module.exports = {
 Initializing the Agent also requires some React Native specific setup, mainly for the Indy SDK and File System. Below is a sample config, see the [README](../README.md#getting-started) for an overview of getting started guides. If you want to jump right in, check the [Getting Started: Agent](./getting-started/0-agent.md) guide.
 
 ```ts
-import { Agent } from 'aries-framework'
-
-// Import rn-indy-sdk and File System for React Native
-import indy from 'rn-indy-sdk'
-import { ReactNativeFileSystem } from 'aries-framework/build/storage/fs/ReactNativeFileSystem'
+import { Agent } from 'aries-framework/core'
+import { agentDependencies } from '@aries-framework/react-native'
```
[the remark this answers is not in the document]
TimoGlastra: Not yet! But that's a good idea! I'll add them to the react-native package

docs/setup-nodejs.md
```diff
@@ -9,33 +9,29 @@ To start using Aries Framework JavaScript in NodeJS some platform specific depen
...
    - [Windows](../docs/libindy/windows.md)
-3. Add `indy-sdk` and `aries-framework` to your project.
+3. Add `@aries-framework/core` and `@aries-framework/node` to your project.
 
 ```bash
-yarn add aries-framework indy-sdk   # npm install aries-framework indy-sdk
+yarn add @aries-framework/core @aries-framework/node
 ```
 
-> ⚠️ If making use of NodeJS > 12, make sure you're also using indy-sdk >= 1.16.0-dev-1633
```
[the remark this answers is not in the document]
TimoGlastra: I think we can remove it? We specify indy-sdk in the dependencies of the node package, so we could just enforce that indy-sdk >= 1.16.0-dev-1633

packages/core/src/constants.ts
```diff
@@ -1,6 +1,5 @@
 export const InjectionSymbols = {
   Wallet: Symbol('Wallet'),
-  Indy: Symbol('Indy'),
```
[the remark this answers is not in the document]
TimoGlastra: That one's a bit more tricky because we initiate it inside the agent and want to have it available across the agent. But thinking of it now I'll add it the agentConfig

packages/react-native/tsconfig.build.json
TimoGlastra: Yes and no. It's really a hassle to not depend on any global node/react-native/browser types. We want types for `fetch` and `WebSocket`. We can either do that by depending on types for specific platform, or we should add them ourselves...

tests/e2e.test.ts
TimoGlastra: Yes!

docs/setup-nodejs.md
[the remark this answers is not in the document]
JamesKEbert: Ahh yeah you're right on this front--it's a part of the whole benefit here. Previously it was only defined as a dev dependency, which allowed you to continue using the wrong Indy version, which could run into problems. Now we're specifying it in the node package.json. 👍

packages/react-native/tsconfig.build.json
JamesKEbert: Hmmm.. can we rely on global Node types, and then specify/override those in the respective environments? (my knowledge is a little light here so)

.eslintrc.js
```diff
@@ -4,25 +4,31 @@ module.exports = {
     'eslint:recommended',
     'plugin:import/recommended',
     'plugin:import/typescript',
-    'plugin:@typescript-eslint/recommended', // Uses the recommended rules from the @typescript-eslint/eslint-plugin
-    'plugin:prettier/recommended', // Enables eslint-plugin-prettier and displays prettier errors as ESLint errors. Make sure this is always the last configuration in the extends array.
```
jakubkoci, line 8: I think it's good to have such comments

packages/core/package.json
```diff
@@ -0,0 +1,53 @@
...
+    "reflect-metadata": "^0.1.13",
+    "rxjs": "^7.1.0",
+    "tsyringe": "^4.5.0",
+    "uuid": "^8.3.2"
+  },
+  "devDependencies": {
+    "@types/bn.js": "^5.1.0",
+    "@types/events": "^3.0.0",
+    "@types/indy-sdk": "^1.16.5",
+    "@types/node-fetch": "^2.5.10",
```
jakubkoci, line 45: I guess we will eventually remove this from the core, right?

packages/core/src/agent/Agent.ts
```diff
@@ -70,19 +71,17 @@ export class Agent {
...
     this.container.registerInstance(InjectionSymbols.Logger, this.logger)
-    this.container.registerInstance(InjectionSymbols.Indy, this.agentConfig.indy)
     this.container.register(InjectionSymbols.Wallet, { useToken: IndyWallet })
     this.container.registerSingleton(InjectionSymbols.StorageService, IndyStorageService)
     this.container.registerSingleton(InjectionSymbols.MessageRepository, InMemoryMessageRepository)
 
-    // File system differs based on NodeJS / React Native
-    this.container.registerInstance(InjectionSymbols.FileSystem, this.agentConfig.fileSystem)
+    // Platform specific dependencies
+    this.container.registerInstance(InjectionSymbols.FileSystem, new dependencies.FileSystem())
```
jakubkoci, line 79: Nitpick: What about creating an instance outside? Then you will have more flexibility to pass params into the constructor or set the instance as you need before passing it into the agent

packages/core/src/agent/AgentDependencies.ts
```diff
@@ -0,0 +1,15 @@
+import type { FileSystem } from '../storage/FileSystem'
+import type { EventEmitter as NativeEventEmitter } from 'events'
+import type * as Indy from 'indy-sdk'
+import type fetch from 'node-fetch'
+import type WebSocket from 'ws'
```
jakubkoci, line 5: Maybe you already mention something about thsese two deps. Would it be possible to remove it later? In other words. Just mentioning it in case we forgot it here :)

packages/core/src/transport/WsOutboundTransporter.ts
```diff
@@ -3,22 +3,27 @@ import type { Logger } from '../logger'
...
 
   public async start(agent: Agent): Promise<void> {
     this.agent = agent
-    this.logger = agent.injectionContainer.resolve(InjectionSymbols.Logger)
+
+    const agentConfig = agent.injectionContainer.resolve(AgentConfig)
+
+    this.logger = agentConfig.logger
     this.logger.debug('Starting WS outbound transport')
+    this.WebSocket = agentConfig.agentDependencies.WebSocket
```
jakubkoci, line 26: The `this.WebSocket` could be hard to understand what it means 🤔  like what's actually behind that property. Just saying, don't have any idea for improvement right now. 😄

packages/react-native/tsconfig.build.json
jakubkoci: I would rather avoid writing our own types, but I understand it's a tricky problem 🙁

jakubkoci (approved):
That's Awesome 🎉 Great job Timo. 

I added just a few insignificant notes :)

.eslintrc.js
TimoGlastra: I've added the prettier one back in. The other one is not adding anything IMO

packages/core/package.json
TimoGlastra: Yes once we resolve the typing issues as described above (declare them ourselves or use some trick to make it work)

packages/core/src/agent/Agent.ts
TimoGlastra: I'll look at that in a separate PR. That would mean extra steps to get started, even though you won't need in 99% of the cases

We can make it an optional config property?

packages/core/src/agent/AgentDependencies.ts
TimoGlastra: Yes. But we only depend on the @types packages. Not the actual packages themselves

packages/core/src/transport/WsOutboundTransporter.ts
TimoGlastra: I've renamed to `WebSocketClass` for now. Let me know if you want something else

PR #327, opened 2021-06-17
fix: remove dependency on global types

TimoGlastra:
We hadn't restricted the types we could used in the tsconfig, meaning we were using both NodeJS globals and DOM globals.

I've restricted the usage of both. This means some `@ts-ignore` for now, but that will be fixed when #322 is merged.


TimoGlastra:
waiting for https://github.com/microsoft/DefinitelyTyped-tools/pull/281

package.json
```diff
@@ -2,32 +2,34 @@
...
-    "validate": "npm-run-all --parallel lint compile",
+    "mediator:start": "ts-node ./tests/mediator.ts",
+    "mediator:start-ws": "ts-node ./tests/mediator-ws.ts",
+    "validate": "npm-run-all --parallel lint check-types",
     "prepack": "rm -rf build && yarn compile",
     "release": "release-it",
     "prepare": "husky install"
   },
   "dependencies": {
+    "@types/ws": "^7.4.5",
```
jakubkoci, line 32: Is it in regular `dependencies` with a purpose?

jakubkoci (approved):

package.json
TimoGlastra: It was. But I removed it again. We were directly importing the type,  but not the package and therefor it couldn't find the type. However the dependency on this type package also introduces a dependency on `@types/node` which clashes with `@types/react-native`. Very annoying

PR #329, opened 2021-06-17
feat: method to retrieve credentials for proof request

TheTreek:
Added new object that contains the arrays of potential credentials for proof requests
getRequestedCredentialsForProofRequest returns the new object
Added autoSelectCredentialsForProofRequest to have similar functionality as before
Indy credential info is now stored in RequestedPredicate and RequestedAttribute objects


package-lock.json
```diff
@@ -0,0 +1,9571 @@
+{
```
TimoGlastra, line 1: We use yarn instead of NPM. Could you remove the package-lock in favour of the yarn.lock?

src/modules/proofs/ProofsModule.ts
```diff
@@ -193,24 +193,36 @@ export class ProofsModule {
...
    * use credentials in the wallet to build indy requested credentials object for input to proof creation.
    * If restrictions allow, self attested attributes will be used.
    *
-   * Use the return value of this method as input to {@link ProofService.createPresentation} to automatically
-   * accept a received presentation request.
    *
    * @param proofRequest The proof request to build the requested credentials object from
    * @param presentationProposal Optional presentation proposal to improve credential selection algorithm
-   * @returns Requested credentials object for use in proof creation
+   * @returns RetrievedCredential object
```
TimoGlastra, line 203:
```suggestion
   * @returns RetrievedCredentials object
```

src/modules/proofs/services/ProofService.ts
```diff
@@ -611,110 +612,104 @@ export class ProofService {
...
    * use credentials in the wallet to build indy requested credentials object for input to proof creation.
    * If restrictions allow, self attested attributes will be used.
    *
-   * Use the return value of this method as input to {@link ProofService.createPresentation} to automatically
-   * accept a received presentation request.
    *
    * @param proofRequest The proof request to build the requested credentials object from
    * @param presentationProposal Optional presentation proposal to improve credential selection algorithm
-   * @returns Requested credentials object for use in proof creation
+   * @returns RetrievedCredential object
```
TimoGlastra, line 622:
```suggestion
   * @returns RetrievedCredentials object
```

src/storage/fs/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './NodeFileSystem'
+export * from './FileSystem'
+export * from './ReactNativeFileSystem'
```
TimoGlastra, line 3: We can't export this from the framework root as that means you automatically import `fs` if you import anything from the framework. `fs` is not available in React Native

src/modules/proofs/models/RequestedCredentials.ts
```diff
@@ -21,12 +21,10 @@ interface RequestedCredentialsOptions {
...
-  public constructor(options: RequestedCredentialsOptions) {
-    if (options) {
-      this.requestedAttributes = options.requestedAttributes ?? {}
-      this.requestedPredicates = options.requestedPredicates ?? {}
-      this.selfAttestedAttributes = options.selfAttestedAttributes ?? {}
-    }
+  public constructor(options: RequestedCredentialsOptions = {}) {
+    this.requestedAttributes = options.requestedAttributes ?? {}
+    this.requestedPredicates = options.requestedPredicates ?? {}
+    this.selfAttestedAttributes = options.selfAttestedAttributes ?? {}
```
TimoGlastra, line 27: I think we should keep the `if (options)` here. Otherwise you'd get weir behaviour when transforming from json to class

TimoGlastra (changes requested):
This is looking great! thanks @TheThreek.

I have a few small suggestions

TimoGlastra (approved):
LGTM

JamesKEbert (approved):
Looks good to me @TheTreek, thanks for your efforts here!

PR #331, opened 2021-06-24
feat: allow for lazy wallet initialization

TimoGlastra:
- remove agent `closeAndDeleteWallet` as it was only used for tests. They now inject the wallet to close and delete
- make `walletConfig` and `walletCredentials` optional in `InitConfig`
- user can now manually initiale the wallet
- If agent is initialized and wallet is not yet initialized and no `walletConfig` and `walletCredentials` are available an error will be thrown

### DIFF
--- a/src/wallet/IndyWallet.ts
+++ b/src/wallet/IndyWallet.ts
@@ -22,21 +22,30 @@ import { AriesFrameworkError } from '../error'
 import { JsonEncoder } from '../utils/JsonEncoder'
 import { isIndyError } from '../utils/indyError'
 
+import { WalletDuplicateError, WalletNotFoundError, WalletError } from './error'
+
+export interface IndyOpenWallet {
+  walletHandle: number
+  masterSecretId: string
+  walletConfig: WalletConfig
+  walletCredentials: WalletCredentials
+}
+
 @scoped(Lifecycle.ContainerScoped)
 export class IndyWallet implements Wallet {
-  private _walletHandle?: number
-  private _masterSecretId?: string
-  private walletConfig: WalletConfig
-  private walletCredentials: WalletCredentials
+  private openWalletInfo?: IndyOpenWallet
+
   private logger: Logger
   private publicDidInfo: DidInfo | undefined
   private indy: typeof Indy
 
   public constructor(agentConfig: AgentConfig) {
-    this.walletConfig = agentConfig.walletConfig
-    this.walletCredentials = agentConfig.walletCredentials
     this.logger = agentConfig.logger
     this.indy = agentConfig.indy
+  }
+
+  public get isInitialized() {
+    return this.openWalletInfo !== undefined
   }
 
   public get publicDid() {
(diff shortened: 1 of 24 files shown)


src/wallet/IndyWallet.ts
```diff
@@ -22,86 +22,267 @@ import { AriesFrameworkError } from '../error'
...
+    return this.openWalletInfo !== undefined
+  }
+
   public get publicDid() {
     return this.publicDidInfo
   }
 
   public get walletHandle() {
-    if (!this._walletHandle) {
+    if (!this.openWalletInfo) {
```
jakubkoci, line 56: What if we re-use `isInitialized` here? 🤔

src/wallet/IndyWallet.ts
```diff
@@ -22,86 +22,267 @@ import { AriesFrameworkError } from '../error'
...
+   *
+   * @param walletConfig the wallet config
+   * @param walletCredentials the wallet credentials
+   * @throws {WalletNotFoundError} if the wallet does not exist
+   * @throws {WalletError} if another error occurs
+   */
+  public async deleteWallet(): Promise<void> {
+    const walletInfo = this.openWalletInfo
+
+    if (!walletInfo) {
```
jakubkoci, line 183: Same here. What if we re-use `isInitialized`?

src/wallet/IndyWallet.ts
```diff
@@ -22,86 +22,267 @@ import { AriesFrameworkError } from '../error'
...
 import { isIndyError } from '../utils/indyError'
 
+import { WalletDuplicateError, WalletNotFoundError, WalletError } from './error'
+
+export interface IndyOpenWallet {
+  walletHandle: number
+  masterSecretId: string
+  walletConfig: WalletConfig
+  walletCredentials: WalletCredentials
+}
```
jakubkoci, line 32: I'm note sure if this type has right cohesion. It seems these attributes are kind of unrealated, at least from the first point of view 🤷‍♂️

jakubkoci (commented):
I think it's good to have some "init" kind of method simplifying a lot of stuff for a developer. On the other hand, it would be good to provide some sort of flexibility out of the box (which we don't have to address in this PR of course).

src/wallet/IndyWallet.ts
```diff
@@ -22,86 +22,267 @@ import { AriesFrameworkError } from '../error'
...
 
-    return this._walletHandle
+    return this.openWalletInfo.walletHandle
   }
 
   public get masterSecretId() {
-    // In theory this is not possible if the wallet handle is available
-    if (!this._masterSecretId) {
-      throw new AriesFrameworkError('Master secret has not been initialized yet')
+    if (!this.openWalletInfo) {
```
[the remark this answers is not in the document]
TimoGlastra: Yeah that works. Problem is with type guarding. openWalletInfo may be undefined and isInitialized guards for that, but TS doesn't know that. Updated

src/wallet/IndyWallet.ts
TimoGlastra: It contains related to an open wallet. I agree with your point, but would like to solve that in a separate PR. 

I think we can make the wallet a bit more straightforward by splitsing the functionality in two parts:

- A "controller" or something that does the following
	- create
	- open
	- delete
	- close
- A wallet that does the actual wallet stuff. You need an open wallet to use it.

This will make it easier to handle multiple wallets I think and separates the lifecycle of a wallet from the usage of the wallet

I'd like to do that in a future PR

TimoGlastra:
@jakubkoci made updates based on your feedback

src/wallet/IndyWallet.ts
jakubkoci: Cool, good idea 👍

jakubkoci (approved):

PR #336, opened 2021-06-29
feat: negotiation and auto accept credentials

blu3beri:
This allows for negotiation between the holder and the issuer via the `negotiateOffer` and `negotiateProposal` functions on `agent.credentials`. 

It also allows for auto accept of credentials. the `autoAcceptCredentials` can be set on the `agent` itself or on a specific `credentialRecord` where the `credentialRecord` has priority.

The `autoAcceptCredential` has three states: `always`, `contentApproved` and `never` where `never` is the default.

`always`: This means that any credential, no matter from who is automatically accepted (even if the attributes or `credentialDefinitionId` changed).
`contentApproved`: This means that when a credential is received it has to be accepted once and the attributes and `credentialDefinitionId` did not change. If something did change it will not automate the flow.
`never`: This simply means that no credentials are automatically accepted.


src/modules/credentials/handlers/IssueCredentialHandler.ts
```diff
@@ -1,17 +1,58 @@
...
 
-  public constructor(credentialService: CredentialService) {
+  public constructor(credentialService: CredentialService, agentConfig: AgentConfig) {
     this.credentialService = credentialService
+    this.agentConfig = agentConfig
   }
 
   public async handle(messageContext: HandlerInboundMessage<IssueCredentialHandler>) {
-    await this.credentialService.processCredential(messageContext)
+    const credentialRecord = await this.credentialService.processCredential(messageContext)
```
TimoGlastra, line 23: Can you add some logging methods that explain what is happening with regards to auto accepting. (e.g. "auto sending ack with "always" policy" etc..)

src/modules/credentials/handlers/IssueCredentialHandler.ts
```diff
@@ -1,17 +1,58 @@
...
+      }
+    }
+  }
+
+  private async nextStep(
+    credentialRecord: CredentialRecord,
+    messageContext: HandlerInboundMessage<IssueCredentialHandler>
+  ) {
+    const { message } = await this.credentialService.createAck(credentialRecord)
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
TimoGlastra, line 55: We shouldn't use `!` here

src/modules/credentials/handlers/OfferCredentialHandler.ts
```diff
@@ -1,17 +1,64 @@
...
+      credentialRecord.autoAcceptCredential,
+      this.agentConfig.autoAcceptCredentials
+    )
+
+    if (autoAccept === AutoAcceptCredential.always) {
+      return await this.nextStep(credentialRecord, messageContext)
+    } else if (autoAccept === AutoAcceptCredential.contentApproved) {
+      if (credentialRecord.proposalMessage && credentialRecord.offerMessage) {
+        const proposalValues = CredentialUtils.convertAttributesToValues(
+          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
TimoGlastra, line 34: shouldn't use `!`. We can assume we have the message, but if we change something in the future this can break

src/modules/credentials/handlers/OfferCredentialHandler.ts
```diff
@@ -1,17 +1,64 @@
...
 
-  public constructor(credentialService: CredentialService) {
+  public constructor(credentialService: CredentialService, agentConfig: AgentConfig) {
     this.credentialService = credentialService
+    this.agentConfig = agentConfig
   }
 
   public async handle(messageContext: HandlerInboundMessage<OfferCredentialHandler>) {
-    await this.credentialService.processOffer(messageContext)
+    const credentialRecord = await this.credentialService.processOffer(messageContext)
```
TimoGlastra, line 22: Same feedback as IssueCredentialHandler (and for all other handlers)

src/modules/credentials/handlers/ProposeCredentialHandler.ts
```diff
@@ -1,17 +1,76 @@
...
   public async handle(messageContext: HandlerInboundMessage<ProposeCredentialHandler>) {
-    await this.credentialService.processProposal(messageContext)
+    const credentialRecord = await this.credentialService.processProposal(messageContext)
+
+    const autoAccept = CredentialUtils.composeAutoAccept(
+      credentialRecord.autoAcceptCredential,
+      this.agentConfig.autoAcceptCredentials
+    )
+
+    if (autoAccept === AutoAcceptCredential.always) {
```
TimoGlastra, line 30: I think it would be nice if we can extract all this logic into an auto respond handler. So something like `xxxx.shouldAutoRespondToProposal()`

src/modules/credentials/handlers/RequestCredentialHandler.ts
```diff
@@ -1,17 +1,34 @@
...
   public async handle(messageContext: HandlerInboundMessage<RequestCredentialHandler>) {
-    await this.credentialService.processRequest(messageContext)
+    const credentialRecord = await this.credentialService.processRequest(messageContext)
+
+    const autoAccept = CredentialUtils.composeAutoAccept(
+      credentialRecord.autoAcceptCredential,
+      this.agentConfig.autoAcceptCredentials
+    )
+
+    if (autoAccept === AutoAcceptCredential.always || autoAccept === AutoAcceptCredential.contentApproved) {
```
TimoGlastra, line 28: We should still check the credential definition id here right?

src/modules/credentials/services/CredentialService.ts
```diff
@@ -88,6 +89,8 @@ export class CredentialService {
       state: CredentialState.ProposalSent,
       proposalMessage,
       credentialAttributes: proposalMessage.credentialProposal?.attributes,
+      tags: { threadId: proposalMessage.threadId, connectionId: connectionRecord.id },
```
TimoGlastra, line 92: Tags are automatically computed  now (#313)

src/modules/credentials/services/CredentialService.ts
```diff
@@ -275,6 +281,8 @@ export class CredentialService {
         schemaId: credOffer.schema_id,
       },
       state: CredentialState.OfferSent,
+      tags: { threadId: credentialOfferMessage.id, connectionId: connectionRecord.id },
```
TimoGlastra, line 284: #313

src/modules/credentials/handlers/IssueCredentialHandler.ts
```diff
@@ -1,17 +1,58 @@
...
+    )
+
+    if (autoAccept === AutoAcceptCredential.always) {
+      return await this.nextStep(credentialRecord, messageContext)
+    } else if (autoAccept === AutoAcceptCredential.contentApproved) {
+      if (credentialRecord.credentialAttributes && credentialRecord.credentialMessage) {
+        const indyCredential = credentialRecord.credentialMessage.indyCredential
+
+        if (!indyCredential) {
+          throw new AriesFrameworkError(`Missing required base64 encoded attachment data for credential`)
```
TimoGlastra, line 37: I don't think we should throw an error here, but rather do an error log (this.logger.error). As we can't do anything with the error on a higher level (it is not invoked by the user, but by another agent.)

src/modules/credentials/handlers/IssueCredentialHandler.ts
```diff
@@ -1,17 +1,58 @@
...
+    const credentialRecord = await this.credentialService.processCredential(messageContext)
+
+    const autoAccept = CredentialUtils.composeAutoAccept(
+      credentialRecord.autoAcceptCredential,
+      this.agentConfig.autoAcceptCredentials
+    )
+
+    if (autoAccept === AutoAcceptCredential.always) {
+      return await this.nextStep(credentialRecord, messageContext)
+    } else if (autoAccept === AutoAcceptCredential.contentApproved) {
```
TimoGlastra, line 32: You haven't removed this from the `processCredential` method:

```
    // Assert the values in the received credential match the values
    // that were negotiated in the credential exchange
    // TODO: Ideally we don't throw here, but instead store that it's not equal.
    // the credential may still have value, and we could just respond with an ack
    // status of fail
    if (credentialRecord.credentialAttributes) {
      CredentialUtils.assertValuesMatch(
        indyCredential.values,
        CredentialUtils.convertAttributesToValues(credentialRecord.credentialAttributes)
      )
    }
```

src/types.ts
```diff
@@ -64,3 +65,14 @@ export interface InboundConnection {
...
+export enum AutoAcceptCredential {
+  // Always auto accepts the credential no matter if it changed in subsequent steps
+  'always' = 'always',
+
+  // Needs one acceptation and the rest will be automated if nothing changes
+  'contentApproved' = 'contentApproved',
+
+  // DEFAULT: Never auto accept a credential
+  'never' = 'never',
+}
```
TimoGlastra, line 78:
```suggestion
export enum AutoAcceptCredentialType {
  // Always auto accepts the credential no matter if it changed in subsequent steps
  Always = 'Always',

  // Needs one acceptation and the rest will be automated if nothing changes
  ContentApproved = 'ContentApproved',

  // DEFAULT: Never auto accept a credential
  Never = 'Never',
}
```

TimoGlastra: Also, I think we should move this to the credential module

src/__tests__/credentials-autoAccept.test.ts
```diff
@@ -0,0 +1,874 @@
...
+        offerMessage: expect.any(Object),
+        requestMessage: expect.any(Object),
+        state: CredentialState.Done,
+      })
+    })
+  })
+
+  describe('Auto accept on `contentApproved`', () => {
+    beforeAll(async () => {
+      ;({ faberAgent, aliceAgent, credDefId, schemaId, faberConnection, aliceConnection } = await setupCredentialTests(
```
TimoGlastra, line 501: weird formatting

src/__tests__/credentials-autoAccept.test.ts
```diff
@@ -0,0 +1,874 @@
+import type { Agent } from '../agent/Agent'
```
TimoGlastra, line 1: I think it would be nice if we can remove some of the bloat from this file. `credentials.test.ts` does a lot of checks. I think with auto-accept we can remove a lot of those checks and only test what we're really testing: auto-acceptance. This makes it easier to maintain this test over time

TimoGlastra (changes requested):
This is really great @blu3beri! Please see my suggestions.

One thing I didn't mention in the comments yet, is that . We said that `CredentialRecord.credentialAttributes` would be the source of truth on our interpretation of what we expect to issue or get issued. Currently e.g. `processCredential` overwrites this value. This should only happen if we agreed to this proposal. Could you go over all methods in the module/service/handlers and make sure credentialAttributes is always what we see as the source of truth? I'm open to other approaches.

src/modules/credentials/handlers/IssueCredentialHandler.ts
```diff
@@ -1,17 +1,58 @@
...
+        const credentialRecordValues = CredentialUtils.convertAttributesToValues(credentialRecord.credentialAttributes)
+
+        if (CredentialUtils.checkValuesMatch(credentialMessageValues, credentialRecordValues)) {
+          return await this.nextStep(credentialRecord, messageContext)
+        }
+      }
+    }
+  }
+
+  private async nextStep(
```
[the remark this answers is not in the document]
TimoGlastra: or `createAck` or something. But `nextStep` is vague

src/modules/credentials/handlers/ProposeCredentialHandler.ts
TimoGlastra: This way we can reuse some of the logic and have a single place to handle this.

src/modules/credentials/handlers/IssueCredentialHandler.ts
blu3beri: I did not change anything here. I could replace the assertion here with my check, however it states `but instead store that it's not equal`. I am not sure how to handle that specifically.

src/__tests__/credentials-autoAccept.test.ts
blu3beri: When trying to return multiple variables from a function and putting them into a `let` this is the only way how. I can't say I am a fan.

TimoGlastra: But we don't have to return variables here? I'm talking about the `;({`.

```suggestion
      { faberAgent, aliceAgent, credDefId, schemaId, faberConnection, aliceConnection } = await setupCredentialTests(
```

TimoGlastra: It's because the method call is wrapped inside `()`

src/modules/credentials/handlers/IssueCredentialHandler.ts
TimoGlastra: I think we can leave the `but instead store that it's not equal` out and not do the check in here at all. So the record should be updated with the message and the state (As is already happening). Then we have the following possibilities:
- user calls `acceptCredential`. This means they accept the contents of the credential.
- user has `AutoAcceptCredential.always` -> auto accept
- user has `AutoAcceptCredential.contentApproved` -> auto accept (send ack) if credential values match

src/__tests__/credentials-autoAccept.test.ts
blu3beri: the brackets are necessary for using the return values in the object.

`setupCredentialTests` returns multiple variables inside an object which I have to bind into a higher level `let`. This is only possible with wrapping it inside the `()`.

Without these brackets the following error is thrown: `Parsing error: Declaration or statement expected. This '=' follows a block of statements, so if you intended to write a destructuring assignment, you might need to wrap the the whole assignment in parentheses.`.

solution was found here: [SO](https://stackoverflow.com/questions/35576307/declaration-or-statement-expected-javascript-typescript)

TimoGlastra: Ah I didn't knew. Sorry

Weird...

src/modules/credentials/handlers/IssueCredentialHandler.ts
blu3beri: So just remove the if statement and add nothing else if I understand it correctly.

TimoGlastra: Yes

blu3beri:
> This is really great @blu3beri! Please see my suggestions.
> 
> One thing I didn't mention in the comments yet, is that . We said that `CredentialRecord.credentialAttributes` would be the source of truth on our interpretation of what we expect to issue or get issued. Currently e.g. `processCredential` overwrites this value. This should only happen if we agreed to this proposal. Could you go over all methods in the module/service/handlers and make sure credentialAttributes is always what we see as the source of truth? I'm open to other approaches.

I understand what you mean, but `processCredential` does not overwrite the credentialAttributes. Am I missing something or did you mean another function?

TimoGlastra:
I see. Forget that method. I should look before writing... 

But the point stands: only overwrite the credentialAttributes if that's our interpretation of the credential data (so not in processCredential, but do in createAck. not in processOffer, but do in createRequest)

src/modules/credentials/handlers/AutoResponseHandler.ts
```diff
@@ -0,0 +1,179 @@
+import type { Logger } from '../../../logger/Logger'
```
TimoGlastra, line 1: Few last points:
- I think we shouldn't call this a handler as is doesn't handle the auto response
- `handlers/` directory is for message handlers
- If you inject the agent config and logger in the constructor (using DI) you don't need to pass these to the methods. I think this is a bit cleaner

TimoGlastra: Also, please update the `issueCredential` method in `src/__tests__/helpers.ts` to use auto issuance.

src/__tests__/helpers.ts
```diff
@@ -281,27 +281,11 @@ export async function issueCredential({
...
-    threadId: holderCredentialRecord.threadId,
-    state: CredentialState.RequestReceived,
-  })
-
-  issuerCredentialRecord = await issuerAgent.credentials.acceptRequest(issuerCredentialRecord.id)
-
-  holderCredentialRecord = await waitForCredentialRecord(holderAgent, {
+  const holderCredentialRecord = await waitForCredentialRecord(holderAgent, {
     threadId: issuerCredentialRecord.threadId,
     state: CredentialState.CredentialReceived,
```
TimoGlastra, line 286:
```suggestion
    state: CredentialState.Done,
```

src/modules/credentials/CredentialResponseCoordinator.ts
```diff
@@ -0,0 +1,188 @@
...
+    }
+    return false
+  }
+
+  /**
+   * Checks whether it should automatically respond to an offer
+   *
+   * @param credentialRecord The credentialRecord that contains the message(s) to respond to
+   * @param agentAutoAccept The configuration on the agent whether to auto accept
+   * @returns a message that will be send to the other agent
```
TimoGlastra, line 87: Good point recently made by @jakubkoci is that the `@param` and `@return` comments often don't add much value besides in the public API (which this isn't) and can be easily forgotten to be updated (as is the case here).

src/types.ts
```diff
@@ -64,3 +65,14 @@ export interface InboundConnection {
   verkey: Verkey
   connection: ConnectionRecord
 }
+
+export enum AutoAcceptCredential {
```
TimoGlastra, line 69: I'd move this to the credentials module

TimoGlastra (approved):
Few minor nits

TimoGlastra (approved):
Few minor nits

TimoGlastra:
@jakubkoci could you take a look?

jakubkoci:
@TimoGlastra Yes, I'll do it, first thing tomorrow morning :)

TimoGlastra:
@blu3beri can you resolve the conflicts?

TimoGlastra:
> @TimoGlastra Yes, I'll do it, first thing tomorrow morning :)

Awesome! :)

blu3beri:
Tests are currenlty failing due to merging. Will resolve asap.

src/modules/credentials/CredentialResponseCoordinator.ts
```diff
@@ -0,0 +1,168 @@
...
+
+    if (autoAccept === AutoAcceptCredential.Always) {
+      return true
+    } else if (autoAccept === AutoAcceptCredential.ContentApproved) {
+      if (
+        credentialRecord.proposalMessage &&
+        credentialRecord.offerMessage &&
+        credentialRecord.proposalMessage.credentialProposal &&
+        credentialRecord.credentialAttributes
+      ) {
```
jakubkoci, line 52: What if we use destructuring and remove repeating `credentialRecord.`:
```js
const { proposalMessage, offerMessage, credentialAttributes } = credentialRecord
if (proposalMessage && offerMessage && proposalMessage.credentialProposal && credentialAttributes) {
  const offerValues = CredentialUtils.convertAttributesToValues(offerMessage.credentialPreview.attributes)
  const defaultValues = CredentialUtils.convertAttributesToValues(credentialAttributes)
  const proposalCredentialDefinitionId = proposalMessage.credentialDefinitionId
  const offerCredentialDefinitionId = offerMessage.indyCredentialOffer?.cred_def_id

  if (
    CredentialUtils.checkValuesMatch(defaultValues, offerValues) &&
    proposalCredentialDefinitionId === offerCredentialDefinitionId
  ) {
    return true
  }
}
```

src/modules/credentials/CredentialResponseCoordinator.ts
```diff
@@ -0,0 +1,168 @@
...
+          this.agentConfig.logger.error(`Missing required base64 encoded attachment data for credential`)
+          return false
+        }
+
+        const credentialMessageValues = indyCredential.values
+        const defaultValues = CredentialUtils.convertAttributesToValues(credentialRecord.credentialAttributes)
+
+        if (CredentialUtils.checkValuesMatch(credentialMessageValues, defaultValues)) {
+          return true
+        }
```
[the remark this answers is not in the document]
jakubkoci: I've been playing around with that a bit and got this idea:
```js
public async shouldAutoRespondToOffer(credentialRecord: CredentialRecord) {
  const autoAccept = CredentialResponseCoordinator.composeAutoAccept(
    credentialRecord.autoAcceptCredential,
    this.agentConfig.autoAcceptCredentials
  )

  if (autoAccept === AutoAcceptCredential.Always) {
    return true
  } else if (autoAccept === AutoAcceptCredential.ContentApproved) {
    if (this.areOfferValuesValid(credentialRecord) && this.isProposalCredentialDefinitionIdValid(credentialRecord)) {
      return true
    }
  }
  return false
}
```

The similarity for others:
```
isProposalCredentialDefinitionIdValid
isOfferDefinitionIdValid
isRequestDefinitionIdValid

areProposalValuesValid()
areOfferValuesValid()
areCredentialValuesValid()
```

jakubkoci (changes requested):
I'm not sure if I understand `ContentApproved` correctly but otherwise, this seems like a nice feature.

I was struggling a bit with understanding the validation part inside `CredentialResponseCoordinator` :)

jakubkoci:
Ah, sorry, I missed this "it has to be accepted once" in the explanation of `ContentApproved`. But does it mean accepting a credential offer?

src/modules/credentials/handlers/OfferCredentialHandler.ts
```diff
@@ -1,17 +1,54 @@
...
+
+    if (await this.credentialReponseCoordinator.shouldAutoRespondToOffer(credentialRecord)) {
+      return await this.sendRequest(credentialRecord, messageContext)
+    }
+  }
+
+  /**
+   * Sends a request to the other agent
+   */
+  private async sendRequest(
```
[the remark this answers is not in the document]
blu3beri: With `createRequest` I would assume it returns a request. Sends here specifies that a message is going out with a `request`.`createAndSendRequest` might be the most accurate, but I am not sure if that is a nice name.

src/modules/credentials/CredentialResponseCoordinator.ts
```diff
@@ -0,0 +1,168 @@
...
+   * Checks whether it should automatically respond to the issuance of a credential
+   */
+  public async shouldAutoRespondToIssue(credentialRecord: CredentialRecord) {
+    const autoAccept = CredentialResponseCoordinator.composeAutoAccept(
+      credentialRecord.autoAcceptCredential,
+      this.agentConfig.autoAcceptCredentials
+    )
+
+    if (autoAccept === AutoAcceptCredential.Always) {
+      return true
```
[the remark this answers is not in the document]
blu3beri: Yeah, so the `contentApproved` was to check two things:
1. The user has accepted the credential once (offer or proposal)
2. The content of the credential did not change (if it changed the automation will be cancelled)

I am completely open to better naming or comprehensive description at the enum.

src/modules/credentials/handlers/OfferCredentialHandler.ts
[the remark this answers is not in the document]
TimoGlastra: I agree with @jakubkoci, it doesn't send the request, it only creates it

blu3beri:
> Ah, sorry, I missed this "it has to be accepted once" in the explanation of `ContentApproved`. But does it mean accepting a credential offer?

It means that once in the whole issuance flow the credential has been accepted. Whether it is a propose or offer does not matter. Also with `contentApproved` the attribute values and credentialDefinitionId are not allowed to change.

src/modules/credentials/CredentialResponseCoordinator.ts
```diff
@@ -0,0 +1,168 @@
...
+    recordConfig: AutoAcceptCredential | undefined,
+    agentConfig: AutoAcceptCredential | undefined
+  ) {
+    return recordConfig ?? agentConfig ?? AutoAcceptCredential.Never
+  }
+
+  /**
+   * Checks whether it should automatically respond to a proposal
+   */
+  public shoudlAutoRespondToProposal(credentialRecord: CredentialRecord) {
```
JamesKEbert, line 38: Should be `shouldAutoRespondToProposal()`

src/modules/credentials/CredentialResponseCoordinator.ts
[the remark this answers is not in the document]
jakubkoci: Yeah, I see. That's clear to me now, also not able to think of better naming 😄

jakubkoci:
> > Ah, sorry, I missed this "it has to be accepted once" in the explanation of `ContentApproved`. But does it mean accepting a credential offer?
> 
> It means that once in the whole issuance flow the credential has been accepted. Whether it is a propose or offer does not matter. Also with `contentApproved` the attribute values and credentialDefinitionId are not allowed to change.

“Whether it is a propose or offer does not matter”

But we need to distinguish between the roles of issuer and holder, right?. As I understand it:
* an issuer can accept the proposal, and then the rest of the issuance is automated on the issuer side
* a holder can accept the offer, and then the rest of the issuance is automated on the holder side

Is that correct?

I’m still unsure if it’s good that there is no validation when a developer handles the process manually. Is it a valid use case when credential definition ID is changed during the process?

Maybe we can discuss it in the AFJ call. It's not definitely a blocker for merging.

jakubkoci (approved):
I still have some questions, but they're rather about explaining the behavior. That can be updated in documentation or comments later if needed.

blu3beri:
> But we need to distinguish between the roles of issuer and holder, right?. As I understand it:
> 
>     * an issuer can accept the proposal, and then the rest of the issuance is automated on the issuer side
> 
>     * a holder can accept the offer, and then the rest of the issuance is automated on the holder side
> 
> 
> Is that correct?


Yes, you only automate the flow for your agent. When negotiation starts, `contentApproved` sees that and will cancel the automation. This credential changed detection might be nice to expose to the public API so a framework consumer could easily detect changes and alert a user.

> Is it a valid use case when credential definition ID is changed during the process?

I think so yeah, this could happen with negotiation.


I do think that the documentation is a bit lacking and unclear.

PR #340, opened 2021-06-30
build: docker containers are compatible with m1

blu3beri:
Adds support for Apple silicon M1. Just specifies the platforms for which the containers were build and with Rosetta it should be possible to start the test ledger and mediators now.


TimoGlastra (approved):
🙌

PR #342, opened 2021-07-02
fix: add error message to log

TimoGlastra:
Adds error message to error log.

@jakubkoci this is what you get when you don't include the error message specifically. I think we need to fix this in rn-indy-sdk. I'll try to look at it in the near future. For now this fix helps

image

### DIFF
--- a/src/transport/HttpOutboundTransporter.ts
+++ b/src/transport/HttpOutboundTransporter.ts
@@ -60,8 +60,9 @@ export class HttpOutboundTransporter implements OutboundTransporter {
         this.agent.receiveMessage(wireMessage)
       }
     } catch (error) {
-      this.logger.error(`Error sending message to ${endpoint}`, {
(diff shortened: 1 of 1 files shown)


TimoGlastra:
CI is failing because of error with indy-sdk publishing:

> For anyone receiving an error like the following. We are aware of the issue and are working toward a resolution. I'll update the community when the issue has been resolved. Thank you for your patience.

jakubkoci:
Vau 😮  So it means that it prints empty `error` object like `{}` but if you call `error.message` it prints out an error message? 🤔 🤯

TimoGlastra:
> Vau 😮 So it means that it prints empty `error` object like `{}` but if you call `error.message` it prints out an error message? 🤔 🤯

Exactly. It doesn't happen however in Node.JS

blu3beri (approved):

PR #359, opened 2021-07-03
feat(core): connection-less issuance and verification

### LINKED ISSUE #346: Add support for RFC 0056: Service Decorator
https://github.com/hyperledger/aries-rfcs/tree/527849ec3aa2a8fd47a7bb6c57f918ff8bcb5e8c/features/0056-service-decorator

### LINKED ISSUE #347: Add support for connection-less issuance and verification
- [x] https://github.com/hyperledger/aries-framework-javascript/issues/346
- [ ] Connection-less issuing
- [ ] Connection-less proving

TimoGlastra:
Adds connection-less issuance and verification.

- `Agent.credentials.createOutOfBandOffer` - to create credential offer not bound to connection
- `Agent.proofs.createOutOfBandRequest` - to create presentation request not bound to connection
- Add `~service` decorator

Required more changes than initially thought because I had to remove the required `connection` parameter everywhere.

Fixes #347 
Fixes #346


src/__tests__/helpers.ts
```diff
@@ -313,6 +325,55 @@ export async function issueCredential({
   }
 }
 
+export async function issueConnectionLessCredential({
```
blu3beri, line 328:
```suggestion
export async function issueConnectionlessCredential({
```

blu3beri (commented):
I think `connectionless` instead of `connection-less` might be better.

TimoGlastra:
> I think `connectionless` instead of `connection-less` might be better.

I used the notation as used in the Aries RFCs: https://github.com/hyperledger/aries-rfcs/search?q=connection-less

TimoGlastra:
Ah but seems google like connectionless better

blu3beri:
> > I think `connectionless` instead of `connection-less` might be better.
> 
> I used the notation as used in the Aries RFCs: https://github.com/hyperledger/aries-rfcs/search?q=connection-less

Ah yeah that makes sense then.

JamesKEbert:
This is awesome, I have not given this a thoroooough review yet, but I wanted to gauge/express concern over the naming on createOutOfBandOffer, since it could cause confusion/collision with the Out Of Band (OOB) invitations? What are your thoughts here @TimoGlastra?

TimoGlastra:
> This is awesome, I have not given this a thoroooough review yet, but I wanted to gauge/express concern over the naming on createOutOfBandOffer, since it could cause confusion/collision with the Out Of Band (OOB) invitations? What are your thoughts here @TimoGlastra?

I actually think it matches perfectly with the OOB protocol. Out-Of-Band is a protocol to facilitate out-of-band exchange of messages. ACA-Py has it as a separate out-of-band topic but IMO it makes more sense to house them under the different modules. If we adopt OOB and want to issue a connectionless credentials, we put it inside the OOB message.

I think we can abstract the OOB protocol mostly away unless you have very specific requirements besides connection invitations, connectionless issuance/verification. Does my thinking make sense?

JamesKEbert:
I tend to agree with you here actually. I think having a separate module/service here for Out-Of-Band messages would be helpful, since it can do the bulk of the formatting and any needed storage (for non-connectionless). However I think other modules being able to use Out-Of-Band messages as you've outlined makes sense (essentially initiating the offer, but then collaborating with the OOB module to create the message). 
So, in short I agree with your thought process here @TimoGlastra.

JamesKEbert (approved):
This looks good to me--I would like to take another look after your updates here, specifically given some transport and mediation changes that I think are likely relevant.

TimoGlastra:
> This looks good to me--I would like to take another look after your updates here, specifically given some transport and mediation changes that I think are likely relevant.

@JamesKEbert Updated

TimoGlastra:
It was quite a hassle to merge. I think some pieces could be handled better, but I'd like to handle that in separate PRs and have some discussions about it first. Mainly:

- With auto accept merged, I had to duplicate quite some code between handler and module because they do a lot of the same stuff.
- Message sender is becoming quite complex. I think it's unavoidable, but would be good to streamline it as much as possible. We can now send unpacked messages, packed messages, and also connection-less messages. All have overlap but all have differences.
- Connection less requires quite some code in the module / handlers. I would like to look at a way to reduce this and just let it be handled by the router / messaging / packing logic. However I think we can do this best in combination with out-of-band.

So I'd like to merge in this state and make additional improvements in the future

packages/core/src/modules/proofs/ProofsModule.ts
```diff
@@ -145,14 +154,59 @@ export class ProofsModule {
...
+  }> {
+    const nonce = proofRequestOptions.nonce ?? (await this.proofService.generateProofRequestNonce())
+
+    const proofRequest = new ProofRequest({
+      name: proofRequestOptions.name ?? 'proof-request',
+      version: proofRequestOptions.name ?? '1.0',
+      nonce,
+      requestedAttributes: proofRequestOptions.requestedAttributes,
+      requestedPredicates: proofRequestOptions.requestedPredicates,
+    })
```
jakubkoci, line 188: Shouldn't this be part of `proofService.createRequest` service? It seems like duplication from another method calling `proofService.createRequest`

packages/core/src/modules/proofs/ProofsModule.ts
```diff
@@ -145,14 +154,59 @@ export class ProofsModule {
...
+    // Save ~service decorator to record (to remember our verkey)
+    proofRecord.requestMessage = message
+    await this.proofRepository.update(proofRecord)
+
+    return { proofRecord, requestMessage: message }
+  }
+
   /**
    * Accept a presentation request as prover (by sending a presentation message) to the connection
    * associated with the proof record.
```
jakubkoci, line 212: I built JS doc blindness so I usually don't read it so much :) but I'm not sure if it's updated according to the changes.

packages/core/tests/helpers.ts
```diff
@@ -338,6 +376,49 @@ export async function issueCredential({
...
+  // eslint-disable-next-line prefer-const
+  let { credentialRecord: issuerCredentialRecord, offerMessage } = await issuerAgent.credentials.createOutOfBandOffer({
+    ...credentialTemplate,
+    autoAcceptCredential: AutoAcceptCredential.ContentApproved,
+  })
+
+  const credentialRecordPromise = waitForCredentialRecord(holderAgent, {
+    threadId: issuerCredentialRecord.threadId,
+    state: CredentialState.OfferReceived,
+  })
```
jakubkoci, line 397: Is there any benefit of making a Promise variable rather than `awat`ing it later at line 400?

jakubkoci:
I added a few notes, but it seems good overall 👍  I believe we're able to improve MessageSender. There is a lot of small duplication all over the class, and it was hard for me to get oriented.

packages/core/tests/helpers.ts
```diff
@@ -72,27 +73,31 @@ export async function waitForProofRecord(
...
-    agent.events.on<ProofStateChangedEvent>(ProofEventTypes.ProofStateChanged, listener)
-  })
+) {
+  const observable = agent.events.observable<ProofStateChangedEvent>(ProofEventTypes.ProofStateChanged).pipe(
+    filter((e) => previousState === undefined || e.payload.previousState === previousState),
+    filter((e) => threadId === undefined || e.payload.proofRecord.threadId === threadId),
+    filter((e) => state === undefined || e.payload.proofRecord.state === state),
+    timeout(timeoutMs),
+    catchError(() => {
+      throw new Error(`ProofStateChanged event not emitted within specified timeout: {
```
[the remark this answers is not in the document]
TimoGlastra: Error doesn't take additional arguments

packages/core/src/agent/__tests__/MessageSender.test.ts
```diff
@@ -114,12 +112,12 @@ describe('MessageSender', () => {
...
     })
 
-    test('throw error when there is no service', async () => {
+    test('throw error when there is no service or queue', async () => {
       messageSender.setOutboundTransporter(outboundTransporter)
       transportServiceFindServicesMock.mockReturnValue([])
 
       await expect(messageSender.sendMessage(outboundMessage)).rejects.toThrow(
-        `Connection with id test-123 has no service!`
+        `Message is undeliverable to connection test-123 (Test 123)`
```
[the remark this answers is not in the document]
TimoGlastra: I don't think that could cause any harm? thread id is also external. You mean like `eval()` harm?

packages/core/src/modules/credentials/CredentialsModule.ts
```diff
@@ -23,32 +23,36 @@ import {
   ProposeCredentialHandler,
   RequestCredentialHandler,
 } from './handlers'
+import { CredentialRepository } from './repository/CredentialRepository'
 import { CredentialService } from './services'
 
 @scoped(Lifecycle.ContainerScoped)
 export class CredentialsModule {
   private connectionService: ConnectionService
+  private credentialRepository: CredentialRepository
```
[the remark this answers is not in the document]
TimoGlastra: Agree!

packages/core/src/modules/credentials/CredentialsModule.ts
```diff
@@ -287,12 +403,33 @@ export class CredentialsModule {
...
-    const outboundMessage = createOutboundMessage(connection, message)
-    await this.messageSender.sendMessage(outboundMessage)
+    if (credentialRecord.connectionId) {
+      const connection = await this.connectionService.getById(credentialRecord.connectionId)
+      const outboundMessage = createOutboundMessage(connection, message)
+
+      await this.messageSender.sendMessage(outboundMessage)
+    }
+    // Use ~service decorator otherwise
+    else if (credentialRecord.credentialMessage?.service && credentialRecord.requestMessage?.service) {
```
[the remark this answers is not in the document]
TimoGlastra: I'm not sure I fully understand. I've never heard that else if is an anti-pattern in JS.

However I'm that a big fan of this code. The if/else shouldn't be there in the first place. But until out-of-band I'd like to keep it there

packages/core/src/modules/proofs/ProofsModule.ts
TimoGlastra: I didn't want to make opinions about default values in the service. I'm not even sure we should set default values...

TimoGlastra: So for now I'd like to keep it in the module (and have a discussion about the duplication during the AFJ call)

packages/core/tests/helpers.ts
TimoGlastra: I'm in the progress of refactoring the method to work with observables. We need to start listening before calling the method that will emit the event. Otherwise we start listening potentially after the event has already been emitted. So we need to start the listener, make the function call and then await. 

I've partly refactored some other call to use ReplaySubject so we can replay all events that came in.

TimoGlastra:
> There is a lot of small duplication all over the class. It was hard for me to get oriented.

Definitely would like to improve the duplication in message sender

packages/core/src/modules/credentials/CredentialsModule.ts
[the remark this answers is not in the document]
jakubkoci: I mean this
```
    }
    // Use ~service decorator otherwise
    else if 
```
instead of 
```
    } else if {
      // Use ~service decorator otherwise
    }
```

packages/core/src/modules/proofs/ProofsModule.ts
jakubkoci: ok

JamesKEbert:
I tend to agree on the complexity of the message sender, but I think I'm comfortable improving on that in subsequent PRs.

> Connection less requires quite some code in the module / handlers. I would like to look at a way to reduce this and just let it be handled by the router / messaging / packing logic. However I think we can do this best in combination with out-of-band.

I think this is an important item/consideration ^

packages/core/src/agent/__tests__/MessageSender.test.ts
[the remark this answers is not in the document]
jakubkoci: Ok, sorry I was probably confused remembering some call where there was a discussion about the security of invitations.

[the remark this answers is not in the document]
jakubkoci: Wouldn't it be better to use `connection.label` in such cases instead? I'm also a little bit confused what's the case for `label` and `theirLabel` 🤔  😄

jakubkoci (approved):
It would be nice to merge it before adding support for mulitransports. It seems there is quite some logic in the same area.

PR #362, opened 2021-07-04
feat: Use session to send outbound message

jakubkoci:
This PR contains the following changes:
* A session object is saved only if there is a `return_route` param in an inbound message. 
* An outbound message is sent via session object, and the dispatcher doesn't return the message to the caller. It works for both HTTP and WebSocket.
* The session object also contains the inbound message. The reason for this is that we can decide if we want to use the session to send a message later in MessageSender. We don't try to send a message via session if there is no `return_route` set for a given session.

### DIFF
--- a/src/agent/Agent.ts
+++ b/src/agent/Agent.ts
@@ -184,7 +184,7 @@ export class Agent {
     await this.wallet.delete()
   }
 
-  public closeSession(session: TransportSession) {
+  public removeSession(session: TransportSession) {
     this.transportService.removeSession(session)
   }
 
--- a/src/agent/Agent.ts
+++ b/src/agent/Agent.ts
@@ -30,6 +30,7 @@ import { EventEmitter } from './EventEmitter'
 import { AgentEventTypes } from './Events'
 import { MessageReceiver } from './MessageReceiver'
 import { MessageSender } from './MessageSender'
+import { TransportService } from './TransportService'
 
 export class Agent {
   protected agentConfig: AgentConfig
@@ -38,6 +39,7 @@ export class Agent {
   protected eventEmitter: EventEmitter
   protected wallet: Wallet
   protected messageReceiver: MessageReceiver
+  protected transportService: TransportService
   protected messageSender: MessageSender
   public inboundTransporter?: InboundTransporter
   private _isInitialized = false
@@ -96,6 +98,7 @@ export class Agent {
     this.eventEmitter = this.container.resolve(EventEmitter)
     this.messageSender = this.container.resolve(MessageSender)
     this.messageReceiver = this.container.resolve(MessageReceiver)
+    this.transportService = this.container.resolve(TransportService)
     this.wallet = this.container.resolve(InjectionSymbols.Wallet)
 
     // We set the modules in the constructor because that allows to set them as read-only
@@ -174,6 +177,15 @@ export class Agent {
 
   public async receiveMessage(inboundPackedMessage: unknown, session?: TransportSession) {
     return await this.messageReceiver.receiveMessage(inboundPackedMessage, session)
+  }
+
+  public async closeAndDeleteWallet() {
+    await this.wallet.close()
+    await this.wallet.delete()
+  }
+
+  public closeSession(session: TransportSession) {
+    this.transportService.removeSession(session)
   }
 
   public get injectionContainer() {
--- a/src/agent/MessageSender.ts
+++ b/src/agent/MessageSender.ts
@@ -55,21 +55,18 @@ export class MessageSender {
       connection: { id, verkey, theirKey },
     })
 
-    const session = this.transportService.findSession(connection.id)
-    if (
-      session?.inboundMessage?.hasReturnRouting() ||
-      session?.inboundMessage?.hasReturnRouting(outboundMessage.payload.threadId)
-    ) {
+    const session = this.transportService.findSessionByConnectionId(connection.id)
+    if (session?.inboundMessage?.hasReturnRouting(outboundMessage.payload.threadId)) {
       this.logger.debug(`Existing ${session.type} transport session has been found.`)
-      if (!session.keys) {
-        throw new AriesFrameworkError(`There are no keys for the given ${session.type} transport session.`)
-      }
-      const outboundPackage = await this.packMessage(outboundMessage, session.keys)
       try {
+        if (!session.keys) {
+          throw new AriesFrameworkError(`There are no keys for the given ${session.type} transport session.`)
+        }
+        const outboundPackage = await this.packMessage(outboundMessage, session.keys)
         await session.send(outboundPackage)
         return
       } catch (error) {
-        this.logger.info('The transport session has been closed or failed to send the outbound message.', error)
+        this.logger.info(`Sending an outbound message via session failed with error: ${error.message}.`, error)
       }
     }
 
--- a/src/agent/MessageSender.ts
+++ b/src/agent/MessageSender.ts
@@ -60,6 +60,21 @@ export class MessageSender {
       throw new AriesFrameworkError(`Connection with id ${connection.id} has no service!`)
     }
 
+    const session = this.transportService.findSession(connection.id)
+    if (session) {
+      this.logger.debug(`Existing ${session.type} transport session has been found.`)
+      if (!session.keys) {
+        throw new AriesFrameworkError(`There are no keys for the given ${session.type} transport session.`)
+      }
+      const outboundPackage = await this.packMessage(outboundMessage, session.keys)
+      try {
+        await session.send(outboundPackage)
+        return
+      } catch (error) {
+        this.logger.info('The transport session has been closed or failed to send the outbound message.', error)
+      }
+    }
+
     for await (const service of services) {
       this.logger.debug(`Sending outbound message to service:`, { messageId: message.id, service })
       try {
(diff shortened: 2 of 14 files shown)


src/agent/MessageSender.ts
```diff
@@ -55,6 +55,24 @@ export class MessageSender {
...
 
+    const session = this.transportService.findSession(connection.id)
+    if (
+      session?.inboundMessage?.hasReturnRouting() ||
+      session?.inboundMessage?.hasReturnRouting(outboundMessage.payload.threadId)
+    ) {
+      this.logger.debug(`Existing ${session.type} transport session has been found.`)
+      if (!session.keys) {
+        throw new AriesFrameworkError(`There are no keys for the given ${session.type} transport session.`)
+      }
```
TimoGlastra, line 66: Do we want to throw an error here? Maybe just log that we couldn't find them and then move on to trying the outbound transports?

src/agent/TransportService.ts
```diff
@@ -54,4 +57,7 @@ interface TransportSessionTable {
 
 export interface TransportSession {
   type: string
+  keys?: EnvelopeKeys
+  inboundMessage?: AgentMessage
```
TimoGlastra, line 61: Why is `inboundMessage` optional? Because we also use it for outbound transport?

src/agent/__tests__/MessageSender.test.ts
```diff
@@ -35,7 +36,12 @@ class DummyOutboundTransporter implements OutboundTransporter {
 }
 
 class DummyTransportSession implements TransportSession {
-  public readonly type = 'dummy'
+  public readonly type = 'http'
+  public keys: EnvelopeKeys | undefined
```
TimoGlastra, line 40: wouldn't this be possible?

```suggestion
  public keys?: EnvelopeKeys
```

src/agent/TransportService.ts
TimoGlastra: Same for `keys`

TimoGlastra: I think it is nice to make as much required as possible

src/decorators/transport/TransportDecoratorExtension.ts
```diff
@@ -31,6 +31,11 @@ export function TransportDecorated<T extends BaseMessageConstructor>(Base: T) {
       // transport is thread but threadId is either missing or doesn't match. Return false
       return false
     }
+
+    public hasAnyReturnRoute() {
+      const returnRoute = this.transport?.returnRoute
```
TimoGlastra, line 36: Not really a fan of this method in addition to the already exiting method. But I see the limitation of the first method

src/transport/WsOutboundTransporter.ts
```diff
@@ -10,11 +10,19 @@ import { WebSocket } from '../utils/ws'
...
-  public socket?: WebSocket
+  public socket: WebSocket
 
-  public constructor(socket?: WebSocket) {
+  public constructor(socket: WebSocket) {
     this.socket = socket
   }
+
+  public async send(outboundMessage: OutboundPackage): Promise<void> {
+    // logger.debug(`Sending outbound message via ${this.type} transport session`)
```
TimoGlastra, line 20:
```suggestion
    logger.debug(`Sending outbound message via ${this.type} transport session`)
```

src/transport/WsOutboundTransporter.ts
```diff
@@ -10,11 +10,19 @@ import { WebSocket } from '../utils/ws'
...
+    if (this.socket.readyState !== WebSocket.OPEN) {
+      throw new Error(`${this.type} transport session has been closed.`)
```
TimoGlastra, line 22:
```suggestion
      throw new AriesFrameworkError(`${this.type} transport session has been closed.`)
```

tests/__tests__/e2e.test.ts
```diff
@@ -1,8 +1,9 @@
 import { Agent, HttpOutboundTransporter, PollingInboundTransporter } from '../../src'
 import { getBaseConfig, waitForBasicMessage } from '../../src/__tests__/helpers'
-import logger from '../../src/__tests__/logger'
+import testLogger from '../../src/__tests__/logger'
 import { get } from '../http'
 
+const logger = testLogger
```
TimoGlastra, line 6: Can you explain the reason for doing this one more time? I forgot :)

tests/mediator.ts
```diff
@@ -13,6 +14,29 @@ import { DidCommMimeType } from '../src/types'
...
+  public constructor(req: Request, res: Response) {
+    this.req = req
+    this.res = res
+  }
+
+  public async send(outboundMessage: OutboundPackage): Promise<void> {
+    logger.debug(`Sending outbound message via ${this.type} transport session`)
+
+    if (this.res.headersSent) {
+      throw new Error(`${this.type} transport session has been closed.`)
```
TimoGlastra, line 33:
```suggestion
      throw new AriesFrameworkError(`${this.type} transport session has been closed.`)
```

tests/mediator.ts
```diff
@@ -13,6 +14,29 @@ import { DidCommMimeType } from '../src/types'
 
 import config from './config'
 
+const logger = testLogger
+
+class HttpTransportSession implements TransportSession {
```
TimoGlastra, line 19: Why is `WebSocketTransportSession` part of the framework, but `HttpTransportSession` not?

tests/mediator.ts
```diff
@@ -26,13 +50,15 @@ class HttpInboundTransporter implements InboundTransporter {
...
+        const session = new HttpTransportSession(req, res)
+        await agent.receiveMessage(packedMessage, session)
+
+        // If agent did not use session when processing message we need to send response here.
+        if (!res.headersSent) {
           res.status(200).end()
         }
-      } catch {
+      } catch (error) {
+        logger.error(error)
```
TimoGlastra, line 61:
```suggestion
        logger.error(`Error processing message in mediator: ${error.message}`, error)
```

TimoGlastra (changes requested):
Nice improvements.

How are transport sessions removed? I see the send message on a session now throws an error if the session is already closed. This means as you keep receiving messages using HTTP the list will grow and grow. Is this something you will work on next?

src/agent/MessageSender.ts
jakubkoci: Ok, I moved it inside the try-catch. What do you think?

src/agent/__tests__/MessageSender.test.ts
jakubkoci: Yes, that works 👍

src/decorators/transport/TransportDecoratorExtension.ts
jakubkoci: Hmm, I'm not a fan of that either. We should perhaps refactor both to better communicate what's going on there 🤔  I made a mistake when using the other method inside `MessageSender` as you found.

src/transport/WsOutboundTransporter.ts
jakubkoci: Updated

tests/__tests__/e2e.test.ts
jakubkoci: Yeah. It's handy to be able to set the logger in one test to `ConsoleLogger` in this way.

tests/mediator.ts
jakubkoci: Updated.

tests/mediator.ts
jakubkoci: Ah, good point. I thought it's gonna be used inside `WsOutboundTransporter`. I moved `WebSocketTransportSession` to `mediator-ws` and we have it in the same place now. Eventually, it can go inside framework together with those inbound transports. (I can move it there when I'll be renaming transporters to transport)

tests/mediator.ts
jakubkoci: Updated.

jakubkoci:
Regarding the session removal... I planned to do this as one of the next steps. However, I realized it could be a lot of data stored in memory when we use framework inside mediator with polling, so I added a solution into this PR. It's possible to remove a session via `agent.closeSession` now. I needed to do some workarounds by adding a connection to the session object. I would say "not great not terrible" :)

tests/mediator.ts
```diff
@@ -26,13 +53,16 @@ class HttpInboundTransporter implements InboundTransporter {
...
+
+        // If agent did not use session when processing message we need to send response here.
+        if (!res.headersSent) {
           res.status(200).end()
         }
-      } catch {
+        agent.closeSession(session)
+      } catch (error) {
+        logger.error(`Error processing message in mediator: ${error.message}`, error)
         res.status(500).send('Error processing message')
```
TimoGlastra, line 66: catch should also close the session

src/agent/Agent.ts
```diff
@@ -156,6 +159,10 @@ export class Agent {
     await this.wallet.delete()
   }
 
+  public closeSession(session: TransportSession) {
```
TimoGlastra, line 162: `closeSession` is misleading as it won't close the session. `removeSession` on transport service makes more sense.

Also, I don't think we should expose the `removeSession` on the agent. It doesn't feel like public API to me, so would rather see the classes that need it inject it.

TimoGlastra (approved):
Approving, but would like to see the `removeSession` method removed from the agent class

src/agent/Agent.ts
jakubkoci: Agree. Do you have any particular idea, could you give me an example of that "classes that need it inject it."?

src/agent/TransportService.ts
jakubkoci: Sorry, I forgot to respond to this. The reason is that session is constructed outside the framework where `keys`, `connection`, or `inboundMessage` are not available. If these attributes are gathered at the same place, we can maybe introduce another data structure similar to `context` in `InboundMessageContext` and set it via a setter function.

PR #364, opened 2021-07-05
fix: Remove apostrophe from connection request message type

jakubkoci:


TimoGlastra (approved):
Ouch... good catch

PR #367, opened 2021-07-06
feat: auto accept proofs

blu3beri:
This adds automatically accepting of proof requests. This functions the same as #336 to stay as consistent as possible, yet they work completely seperate to eachother.

It allows for an agent- and recordwide configuration where the record configuration has priority.

Three states are allowed to be set:
- `always`: accept any proof request/proposal no matter what the content might be.
- `contentApproved`: accept any proof request/proposal as long as none of the content changes and it has to be accepted manually at least once.
- `never` (DEFAULT) Does not automate anything and will just behave the same as before these changes.

Note: I am not sure whether we want to mention this as a feature in the README, so if there is any problem with this I will remove it.


blu3beri:
Tests on Github are failing while local tests work. Will resolve this asap.

TimoGlastra (approved):
LGTM. Nice work @blu3beri!

JamesKEbert (commented):
If I recall correctly, we're waiting for additional followup to move the relevant validation changes into the service regardless of the automation configuration? Or am I remembering incorrectly here?

blu3beri:
> If I recall correctly, we're waiting for additional followup to move the relevant validation changes into the service regardless of the automation configuration? Or am I remembering incorrectly here?

My understanding was first to merge this. And in a seprate PR abstract the functionality, add documentation stating the types of autoAccept.  I think for now this will suffice, but I would not mind to change that stuff and merge it later.

packages/core/tests/proofs-auto-accept.test.ts
```diff
@@ -0,0 +1,206 @@
...
+  ProofAttributeInfo,
+  AttributeFilter,
+  ProofPredicateInfo,
+  PredicateType,
+} from '../src'
+
+import { setupProofsTest, waitForProofRecord } from './helpers'
+import testLogger from './logger'
+
+describe('Auto accept present proof', () => {
```
jakubkoci, line 16: Shouldn't we have tests also for accept never?

jakubkoci: It seems we're testing just a happy path.

packages/core/src/modules/proofs/ProofResponseCoordinator.ts
```diff
@@ -0,0 +1,86 @@
+import type { ProofRecord } from './repository'
+
+import { scoped, Lifecycle } from 'tsyringe'
+
+import { AgentConfig } from '../../agent/AgentConfig'
+
+import { AutoAcceptProof } from './ProofAutoAcceptType'
```
jakubkoci, line 7: Maybe we should have two types, like `AutoAcceptProof` for the holder and `AutoAcceptProofRequest` request for the issuer. Similarly with credentials. But that's of course more of a future improvement idea.

jakubkoci (approved):

packages/core/tests/proofs-auto-accept.test.ts
blu3beri: Since I assumed the default is `Never`, the `proofs.test.ts will error if it does any automatic response. I could add the `never` clause as a test, but it would feel a bit redundant.

packages/core/src/modules/proofs/handlers/RequestPresentationHandler.ts
```diff
@@ -1,17 +1,64 @@
...
+    if (this.proofResponseCoordinator.shouldAutoRespondToRequest(proofRecord)) {
+      return await this.createPresentation(proofRecord, messageContext)
+    }
+  }
+
+  private async createPresentation(
+    proofRecord: ProofRecord,
+    messageContext: HandlerInboundMessage<RequestPresentationHandler>
+  ) {
+    const indyProofRequest = proofRecord.requestMessage?.indyProofRequest
```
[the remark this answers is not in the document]
blu3beri: Will fix this when we support multiple credential types.

packages/core/src/modules/proofs/ProofResponseCoordinator.ts
blu3beri: Any particular reason for this? Maybe it allows for some use cases that I don't know of.

jakubkoci: I see 2 reasons:
The first is expressiveness. `AutoAcceptProof` means auto acceptance proof from the verifier point of view but also auto acceptance of the proof request from the holder's perspective. Even now, I'm not sure if I wrote it correctly.

The second is a potential use case when someone uses an agent for more than one role. Let's say I don't want to I want to accept proofs as a verifier but don't want to accept proof requests as a holder.

TimoGlastra:
Merging. Additional improvements can be made in follow-up PRs

PR #368, opened 2021-07-07
refactor: inject transport service in transports

TimoGlastra:
@jakubkoci as per your comment in #322 here's how you can achieve it using dependency injection

I think we should use injection for the transports in general instead of manually passing the agent in. But for now I think this is a fine solution.

### DIFF
--- a/src/agent/Agent.ts
+++ b/src/agent/Agent.ts
@@ -184,10 +184,6 @@ export class Agent {
     await this.wallet.delete()
   }
 
-  public removeSession(session: TransportSession) {
-    this.transportService.removeSession(session)
-  }
-
   public get injectionContainer() {
     return this.container
   }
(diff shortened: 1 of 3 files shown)


jakubkoci (approved):
Ah, interesting 🤔  We must be careful with this because it can expose framework internals depending on `TransportService` dependencies. It complicates transport implementation a bit. I still consider transport implementation as something that happens outside of the framework, but maybe I should give up that idea 🤷‍♂️   The problem is that HTTP inbound transport is now depending on Express. I wasn't able to make it work with Nest.js. A lot of people are using other Node.js HTTP frameworks.

TimoGlastra:
> I still consider transport implementation as something that happens outside of the framework

I see it a something separate from the core of the framework, but I do think it is essential for the ease-of-use of the framework to offer ready-to-use transports. Currently you have to copy the inbound transport, that's just not developer friendly and leaves one of the most complex parts (routing) to the user of the framework.

My idea is to not necessarily have a single HttpInboundTransport, but rather a default ExpressHttpInboundTransport, and later maybe a NestHttpInboundTransport. (either under `@aries-framework/node`, or as `@aries-framework/transport-http-inbound-express`)

Can you share the code with me from your experiments of Nest.JS. Maybe I can take a look? If I know where the problem is, I can take this into account for future development. I don't want to exclude a library.

PR #369, opened 2021-07-08
fix: adds theirLabel to connectionRecord

blu3beri:
Simply adds `theirLabel` to the connectionRecord as retreiving it from the `connectionRecord.invitation.label` is too cumbersome and not inline with the other `their...` properties.

### DIFF
--- a/src/__tests__/helpers.ts
+++ b/src/__tests__/helpers.ts
@@ -248,6 +248,7 @@ export function getMockConnection({
     tags,
     verkey,
     invitation,
+    theirLabel: invitation.label,
   })
 }
 
--- a/src/__tests__/helpers.ts
+++ b/src/__tests__/helpers.ts
@@ -248,7 +248,6 @@ export function getMockConnection({
     tags,
     verkey,
     invitation,
-    theirLabel: invitation.label,
   })
 }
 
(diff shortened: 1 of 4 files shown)


TimoGlastra (changes requested):
I'd make this a computed property. So add the following to connection record. This way we don't have to deal with setting the label property manually

```
get theirLabel() {
	return this.invitation?.label
}
```

TimoGlastra (approved):

JamesKEbert:
I've been meaning to make a PR to alter this functionality. Worth noting is that this only works when AFJ is the invitee. If AFJ is the inviter, then this will return the wrong label.

TimoGlastra:
Ah interesting. Maybe we can take the label from the request if we're the inviter

JamesKEbert:
Exactly, so I had added it to the connection service `processRequest()`:
`connectionRecord.theirLabel = message.label`

This means we can't use this as a computed property (since IIRC we're not storing the request long term), but that shouldn't be too problematic.
If it's helpful I can open a PR pretty quickly with what I had created. Let me know. 👍

blu3beri:
If that would not be too much effort @JamesKEbert then it would be awesome! Will close this when yours is up.

PR #388, opened 2021-07-16
feat(redux-store): move from mobile agent repo

TimoGlastra:
Moves the `@aries-framework/redux-store` package from the mobile agent to here for easier management.

This is just the state as it was in the mobile agent repo. @jakubkoci You'll probably have opinions on this, and I'm interested to hear it. 

I'd like to merge first so we can use the package without manually packing every time, and then discuss maybe at the AFJ WG call.

Fixes hyperledger/aries-mobile-agent-react-native#64

### DIFF
--- a/packages/redux-store/jest.config.ts
+++ b/packages/redux-store/jest.config.ts
@@ -1,0 +1,13 @@
+import type { Config } from '@jest/types'
+
+import base from '../../jest.config.base'
+
+import packageJson from './package.json'
+
+const config: Config.InitialOptions = {
+  ...base,
+  name: packageJson.name,
+  displayName: packageJson.name,
+}
+
+export default config
--- a/packages/redux-store/package.json
+++ b/packages/redux-store/package.json
@@ -1,0 +1,34 @@
+{
+  "name": "@aries-framework/redux-store",
+  "main": "build/index",
+  "types": "build/index",
+  "version": "0.0.0",
+  "files": [
+    "build"
+  ],
+  "publishConfig": {
+    "access": "public"
+  },
+  "homepage": "https://github.com/hyperledger/aries-framework-javascript/tree/main/packages/redux-store",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/hyperledger/aries-framework-javascript",
+    "directory": "packages/redux-store"
+  },
+  "scripts": {
+    "build": "yarn run clean && yarn run compile",
+    "clean": "rimraf -rf ./build",
+    "compile": "tsc -p tsconfig.build.json",
+    "prepublishOnly": "yarn run build",
+    "test": "jest"
+  },
+  "dependencies": {
+    "@aries-framework/core": "*",
+    "@reduxjs/toolkit": "^1.6.0",
+    "react-redux": "^7.2.4"
+  },
+  "devDependencies": {
+    "rimraf": "~3.0.2",
+    "typescript": "~4.3.0"
+  }
+}
--- a/packages/redux-store/src/index.ts
+++ b/packages/redux-store/src/index.ts
@@ -1,0 +1,23 @@
+export { initializeStore } from './store'
+
+export { createAsyncAgentThunk, AgentThunkApiConfig } from './utils'
+
+export {
+  agentSlice,
+  AgentThunks,
+  // Connections
+  connectionsSlice,
+  ConnectionThunks,
+  startConnectionsListener,
+  ConnectionsSelectors,
+  // Credentials
+  credentialsSlice,
+  CredentialsThunks,
+  startCredentialsListener,
+  CredentialsSelectors,
+  // Proofs
+  proofsSlice,
+  ProofsThunks,
+  startProofsListener,
+  ProofsSelectors,
+} from './slices'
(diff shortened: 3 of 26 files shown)


karimStekelenburg (approved):

JamesKEbert (approved):
Again, good work with this @karimStekelenburg & @TimoGlastra

jakubkoci (commented):
Great work 👏  I think it can be beneficial for developers using Redux. I didn't go through the code but I'm interested in how it looks like to use it inside an app. Is it possible to find it in Aries Mobile agent or in any other place?

However, I'm not sure it should be part of the framework. It doesn't fit into our current package structure and it's highly related to a specific UI library 🤷‍♂️  Why not have it inside Aries Mobile Agent? I'm worried that changes inside the framework would force updates in this package and complicate the development.

TimoGlastra:
I get your reasoning, but I have a few reasons to do include it in here (at least for now).

1. AFJ has a lot of breaking changes currently, updating this package is really not a lot of work. Managing it inside another repo creates another maintenance burden
2. AFJ has the CI/CD pipeline set up for this. Another package is just another directory and it will be released automatically to NPM. The mobile agent doesn't have this.
3. We use a same version strategy. This makes it a lot easier for use, because we can require all packages to have the same version. I'd like this package to be included in this versioning strategy.

Re it doesn't fit our current package structure, My intention was always to add more packages over time. I'm working on a wallet implementation that works in Node.JS, Browser and React Native. I would like to host that in this repo, under the `@aries-framework` org, but without adding all the dependencies to `core`.

Re complicating the development, it does indeed complicate the development a bit. However, the package doesn't do that much, and you don't have to care about updating this package, as long as it keeps working. No need to add new features in core to the redux-store, I'll take care of that. 

Let's discuss in more detail during the AFJ call Thursday, but I have strong arguments to include this (and probably a lot of other packages) in this repo. I don't want to create an ecosystem of outdated packages because we want to keep the core lean. I would rather complicate the development a bit and make sure that when you NPM install the libraries it just works™

TimoGlastra:
I'm interested to hear more opinions on this

karimStekelenburg:
I'm in favour of moving the code to AFJ as well.

My main reason for this is that it's closely coupled to AFJ and thus I agree with @TimoGlastra's argument regarding versioning. It makes it a lot less complicated.

Besides that, Redux isn't strictly related to React/React-Native. Redux is pure JS and can therefore be used in NodeJS and the browser as well. I think this package can benefit many developers, regardless of their platform/framework of choice.

That being said, there currently is a dependency to React, because the usage of hooks. We could separate that out at a later point to make it truly platform independent.

genaris:
Thinking in the long-term, it could be adequate to keep it in aries-mobile-agent-react-native, because as far as I understood, the goal of that project is to serve as a base for every React Native app willing to integrate Aries (i.e. RN app developers will not care about AFJ but just import components). 

However, it is not the case right now and maybe there would be apps that integrate directly AFJ with their own components/structure (it's my case actually) and can take profit from this store. Also, @karimStekelenburg has a good point about Redux usage in different environments, so I think it's appropriate to move here (if @TimoGlastra swears that it's easy to maintain :-P).

karimStekelenburg:
I think this is the point of moving it here @genaris! `aries-mobile-agent-react-native` is meant to be a white-label app that developers who need a wallet application for their solution can use and customise, instead of building one from scratch. As you said, developers who **do** want to build their own mobile agent, can benefit from this package without having to adopt any of the RN components inside `aries-mobile-agent-react-native`.

In my view, one of the primary goals of AFJ is to lower the barrier for developers to integrate SSI into their solutions. Any (platform independent) package that helps this cause is a good candidate for this repository in my opinion.

@jakubkoci, AFAIK the Redux store hasn't been adopted by aries-mobile-agent-react-native yet. We have been using it in a client app, but sadly we can't show that code. I'll try to build a small example project for the call next week so you have an idea of how it would work :)

JamesKEbert:
I do somewhat agree with @genaris, but I am flexible in either direction--really the Aries Bifold group is intended to create a community to drive mobile development forward, so if the associated packages end up in AFJ then I am on board with that. I think if we can make the package environment agnostic then it does make the most sense to place within AFJ.

And as far as aries-mobile-agent-react-native implementing the redux store, that is slated for work this week by @NeilSMyers.

jakubkoci:
It could eventually be part of some extensions packages, but there should be a clear distinction between what's part of the framework and what's not. Packages like `core`, `transports,`, `wallet` probably make sense. Then packages like `node` and `react-native` are more about platform-specific dependencies. Hard to say where to put them long-term. Their current placement is not a problem because that's what we exactly wanted to solve by monorepo right now.

jakubkoci:
If I look at packages in the following structure:
- core
- node
- react-native
- redux-store

I would be really confused about what's going on there. That's my point.

That being said, I fully understand your troubles with maintaining the code if it's outside of the framework and I definitely don't want it to cause you sleepless nights :)

TimoGlastra:
@jakubkoci Merging as I think we settled the discussion in the WG call. If you have objections I'll extract it as soon as I get the time to set up another CI/CD pipeline in another repo

PR #390, opened 2021-07-16
feat(core): add discover features protocol

TimoGlastra:
Add discover features protocol.

Currently working on aries-toolbox integration where this is required for.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -49,5 +49,8 @@
     "rimraf": "~3.0.2",
     "tslog": "^3.2.0",
     "typescript": "~4.3.0"
+  },
+  "resolutions": {
+    "@types/node": "^15.14.1"
   }
 }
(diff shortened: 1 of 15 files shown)


packages/core/package.json
```diff
@@ -49,5 +49,8 @@
     "rimraf": "~3.0.2",
     "tslog": "^3.2.0",
     "typescript": "~4.3.0"
+  },
+  "resolutions": {
+    "@types/node": "^15.14.1"
```
TimoGlastra, line 54: @types/node v16 was released recently which causes errors with some the types. I'm not sure if it is an error in the types (the error seems weird), so for now we can stick to types of node 15

packages/core/src/agent/Dispatcher.ts
```diff
@@ -34,6 +36,15 @@ class Dispatcher {
 
     const outboundMessage = await handler.handle(messageContext)
 
+    // Emit event that allows to hook into received messages
+    this.eventEmitter.emit<AgentMessageProcessedEvent>({
```
JamesKEbert, line 40: Really like this addition

JamesKEbert (approved):
LGTM--I have not tested

jakubkoci (approved):

PR #392, opened 2021-07-19
feat(node): add http and ws inbound transport

TimoGlastra:
Moves http and ws inbound transports to the node package. I know there's still work to be done on the transports, but this makes it 1000x easier to set up an agent. They're in the node package, so not strictly tied to core

### DIFF
--- a/packages/node/src/transport/WsInboundTransport.ts
+++ b/packages/node/src/transport/WsInboundTransport.ts
@@ -1,22 +1,17 @@
-import type { Agent, InboundTransporter, Logger, OutboundPackage } from '../../packages/core/src'
-import type { TransportSession } from '../../packages/core/src/agent/TransportService'
+import type { Agent, InboundTransporter, Logger, OutboundPackage, TransportSession } from '@aries-framework/core'
 
-import WebSocket from 'ws'
+import { AriesFrameworkError, AgentConfig, TransportService, utils } from '@aries-framework/core'
+import WebSocket, { Server } from 'ws'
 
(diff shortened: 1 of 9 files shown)


TimoGlastra:
@karimStekelenburg Once this is merged you can directly use the inbound transports

packages/node/src/transport/WsInboundTransport.ts
```diff
@@ -1,22 +1,17 @@
...
+export class WsInboundTransport implements InboundTransporter {
+  private socketServer: Server
   private logger!: Logger
 
   // We're using a `socketId` just for the prevention of calling the connection handler twice.
   private socketIds: Record<string, unknown> = {}
 
-  public constructor(socketServer: WebSocket.Server) {
-    this.socketServer = socketServer
+  public constructor({ server, port }: { server: Server; port?: undefined } | { server?: undefined; port: number }) {
```
jakubkoci, line 13: Would it work with the following types?
```suggestion
  public constructor({ server, port }: { server: Server; port?: string } | { server?: Server; port: number }) {
```

jakubkoci (approved):
Yes, this is a good step forward 👍

packages/node/src/transport/WsInboundTransport.ts
TimoGlastra: I want it to be an XOR statement. So either port or server should be provided but not both. If you create the server yourself, you set the port on the server. If you don't create the server yourself, you need to pass the port.

Could probably improve, but this was the easiest way I could think of with explicit typing that you shouldn't provide both

PR #396, opened 2021-07-20
feat: break out indy wallet, better indy handling

### LINKED ISSUE #330: Extract some method from the wallet interface into other interfaces
- Move to `IndyStorageService`
	- `addWalletRecord`
	- `updateWalletRecordValue`
	- `updateWalletRecordTags`
	- `getWalletRecord`
	- `deleteWalletRecord`
	- `search` 
- Move to `IndyLedgerService`
	- `signRequest`

TimoGlastra:
Extracts part of the code in #360. 

Main reasons for this are:
- remove dependency on indy-sdk where it is not strictly required (mainly wallet interface is less indy-specific now)
- make indy stuff more indy specific
- other points below

I want to make a browser compatible wallet (already have a working one locally) so we can use the agent in the browser

### Break out indy wallet

Fixes #330

It didn't make a lot of sense to me that the indy wallet was handling indy storage and ledger stuff. I think it belongs more in the ledger and storage services. This is also more in line with where we're headed with the shared components libraries

- Rename `LedgerService` to `IndyLedgerService`. It's very indy focussed now.
- move storage related methods to `IndyStorageService`
- move ledger related methods to `IndyLedgerService`

### Better indy message handling

Sometimes you would get very cryptic errors such as `CommonInvalidParam` without a stack trace or anything that would help you get a sense of what's erroring out. I've wrapped all indy calls with a try catch and added a new `IndySdkError`. This will take the indy error as input, but with very nice stack tracing so you know exactly where things went wrong, without losing information from the indy error


packages/core/src/modules/routing/services/MessagePickupService.ts
```diff
@@ -16,14 +17,6 @@ export class MessagePickupService {
     this.messageRepository = messageRepository
   }
 
-  public async batchPickup(inboundConnection: InboundConnection) {
```
burdettadam, line 19: Was the batchPickup message removed intentionally?

packages/core/src/storage/IndyStorageService.ts
```diff
@@ -88,115 +91,147 @@ export class IndyStorageService<T extends BaseRecord> implements StorageService<
...
-      throw error
+      throw isIndyError(error) ? new IndySdkError(error) : error
     }
   }
 
-  /** @inheritDoc {StorageService#getAll} */
+  /** @inheritDoc */
   public async getAll(recordClass: BaseRecordConstructor<T>): Promise<T[]> {
-    const recordIterator = await this.wallet.search(recordClass.type, {}, IndyStorageService.DEFAULT_QUERY_OPTIONS)
+    const recordIterator = this.search(recordClass.type, {}, IndyStorageService.DEFAULT_QUERY_OPTIONS)
```
burdettadam, line 175: Is the search not async any more? Are we missing an await?
```suggestion
    const recordIterator = await this.search(recordClass.type, {}, IndyStorageService.DEFAULT_QUERY_OPTIONS)
```

burdettadam (approved):

packages/core/src/modules/routing/services/MessagePickupService.ts
TimoGlastra: It wasn't used because the `RecipientModule` directly creates a batch pickup message. I choose to remove this method, but it's probably better to use the service method instead.

packages/core/src/storage/IndyStorageService.ts
TimoGlastra: `this.search` returns a generator, which we can use as an async iterator. So the `this.search` call itself is not async

packages/core/src/agent/Agent.ts
```diff
@@ -77,7 +77,7 @@ export class Agent {
       logger: initialConfig.logger != undefined,
     })
 
-    if (!this.agentConfig.walletConfig || !this.agentConfig.walletCredentials) {
+    if (!this.agentConfig.walletConfig) {
       this.logger.warn(
         'Wallet config and/or credentials have not been set on the agent config. ' +
```
blu3beri, line 82: "...and/or credentials..." not needed anymore.

packages/core/src/agent/Agent.ts
```diff
@@ -94,6 +94,7 @@ export class Agent {
 
     // We set the modules in the constructor because that allows to set them as read-only
     this.connections = this.container.resolve(ConnectionsModule)
+
```
blu3beri, line 97: Any reason for this space?

packages/core/src/agent/EnvelopeService.ts
```diff
@@ -28,12 +27,12 @@ class EnvelopeService {
   }
 
   public async packMessage(payload: AgentMessage, keys: EnvelopeKeys): Promise<PackedMessage> {
-    const { routingKeys, recipientKeys, senderKey: senderVk } = keys
+    const { routingKeys, recipientKeys, senderKey: senderVerkey } = keys
```
blu3beri, line 30: Probably a rename symbol mistake?

packages/core/src/agent/__tests__/Agent.test.ts
```diff
@@ -84,7 +84,7 @@ describe('Agent', () => {
       expect(wallet.isInitialized).toBe(false)
 
       // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-      await wallet.initialize(walletConfig!, walletCredentials!)
+      await wallet.initialize(walletConfig!)
```
blu3beri, line 87: Not quite sure what the flow exactly is here, but do we want a non-null-assertion or could we check it before?

packages/core/src/modules/connections/__tests__/ConnectionService.test.ts
```diff
@@ -39,7 +39,7 @@ describe('ConnectionService', () => {
   beforeAll(async () => {
     wallet = new IndyWallet(config)
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-    await wallet.initialize(config.walletConfig!, config.walletCredentials!)
+    await wallet.initialize(config.walletConfig!)
```
blu3beri, line 42: Same here for the assertion.

packages/core/src/modules/connections/__tests__/ConnectionService.test.ts
```diff
@@ -431,8 +431,8 @@ describe('ConnectionService', () => {
     it('returns a connection record containing the information from the connection response', async () => {
       expect.assertions(3)
 
-      const [did, verkey] = await wallet.createDid()
-      const [theirDid, theirVerkey] = await wallet.createDid()
+      const { did, verkey } = await wallet.createDid()
+      const { did: theirDid, verkey: theirVerkey } = await wallet.createDid()
```
blu3beri, line 435: Is this intentional or symbol renaming?

packages/core/src/modules/connections/__tests__/ConnectionService.test.ts
```diff
@@ -505,8 +505,8 @@ describe('ConnectionService', () => {
     it('throws an error when the connection sig is not signed with the same key as the recipient key from the invitation', async () => {
       expect.assertions(1)
 
-      const [did, verkey] = await wallet.createDid()
-      const [theirDid, theirVerkey] = await wallet.createDid()
+      const { did, verkey } = await wallet.createDid()
+      const { did: theirDid, verkey: theirVerkey } = await wallet.createDid()
```
blu3beri, line 509: Same.

packages/core/src/modules/ledger/services/LedgerService.ts
```diff
@@ -97,17 +93,23 @@ export class LedgerService {
         this.logger.debug(`Pool '${poolName}' already exists`, {
           indyError: 'PoolLedgerConfigAlreadyExistsError',
         })
+      } else if (isIndyError(error)) {
+        throw new IndySdkError(error)
```
blu3beri, line 97:
```suggestion
        throw isIndyError(error) ? new IndySdkError(error) : error
```

Could be a ternary expression

packages/core/src/storage/IndyStorageService.ts
```diff
@@ -88,115 +91,147 @@ export class IndyStorageService<T extends BaseRecord> implements StorageService<
...
     query: Partial<ReturnType<T['getTags']>>
   ): Promise<T[]> {
     const indyQuery = this.transformFromRecordTagValues(query as unknown as TagsBase)
 
-    const recordIterator = await this.wallet.search(
-      recordClass.type,
-      indyQuery,
-      IndyStorageService.DEFAULT_QUERY_OPTIONS
-    )
+    const recordIterator = await this.search(recordClass.type, indyQuery, IndyStorageService.DEFAULT_QUERY_OPTIONS)
```
blu3beri, line 190: It is not async anymore right?

packages/core/src/storage/IndyStorageService.ts
```diff
@@ -88,115 +91,147 @@ export class IndyStorageService<T extends BaseRecord> implements StorageService<
...
+    query: WalletQuery,
+    { limit = Infinity, ...options }: WalletSearchOptions & { limit?: number }
+  ) {
+    try {
+      const searchHandle = await this.indy.openWalletSearch(this.wallet.walletHandle, type, query, options)
+
+      let records: Indy.WalletRecord[] = []
+
+      // Allow max of 256 per fetch operation
+      const chunk = limit ? Math.min(256, limit) : 256
```
blu3beri, line 209: This chunk finder seems to be reused somewhere else. Might it be possible to extract it? If the max changes it could get weird.

packages/core/src/storage/__tests__/IndyStorageService.test.ts
```diff
@@ -45,7 +48,7 @@ describe('IndyStorageService', () => {
         },
       })
 
-      const got = await wallet.getWalletRecord(record.type, record.id, {
+      const got = await indy.getWalletRecord(wallet.walletHandle, record.type, record.id, {
```
blu3beri, line 51: I am not sure if `got` is a verbose enough name. Maybe `walletRecord`?

packages/core/src/wallet/IndyWallet.ts
```diff
@@ -80,43 +68,40 @@ export class IndyWallet implements Wallet {
...
+      await this.open(walletConfig)
     } catch (error) {
       // If the wallet does not exist yet, create it and try to open again
       if (error instanceof WalletNotFoundError) {
-        await this.create(walletConfig, walletCredentials)
-        await this.open(walletConfig, walletCredentials)
+        await this.create(walletConfig)
+        await this.open(walletConfig)
       } else {
         throw error
```
blu3beri, line 78: Should this not be an `AriesFrameworkError` or `IndySdkError`?

packages/core/src/wallet/IndyWallet.ts
```diff
@@ -279,95 +264,85 @@ export class IndyWallet implements Wallet {
...
+      throw new WalletError('Error creating Did', { cause: error })
+    }
   }
 
   public async pack(
     payload: Record<string, unknown>,
-    recipientKeys: Verkey[],
-    senderVk: Verkey
+    recipientKeys: string[],
+    senderVerkey?: string | null
```
blu3beri, line 287: Do we want to check for `string`, `undefined` and `null`?

packages/core/src/wallet/IndyWallet.ts
```diff
@@ -279,95 +264,85 @@ export class IndyWallet implements Wallet {
...
-  public async updateWalletRecordTags(type: string, id: string, tags: Record<string, string>) {
-    return this.indy.addWalletRecordTags(this.walletHandle, type, id, tags)
-  }
-
-  public async deleteWalletRecord(type: string, id: string) {
-    return this.indy.deleteWalletRecord(this.walletHandle, type, id)
-  }
+  public async sign(data: Buffer, verkey: string): Promise<Buffer> {
+    try {
+      const signatureBuffer = await this.indy.cryptoSign(this.walletHandle, verkey, data)
```
blu3beri, line 322: Very nitpicky, but I prefer to just instantly return the value instead of saving it in a variable.

packages/core/src/wallet/IndyWallet.ts
```diff
@@ -279,95 +264,85 @@ export class IndyWallet implements Wallet {
...
-    return generator(this.indy, this.walletHandle)
   }
 
-  public getWalletRecord(type: string, id: string, options: WalletRecordOptions): Promise<WalletRecord> {
-    return this.indy.getWalletRecord(this.walletHandle, type, id, options)
-  }
+  public async verify(signerVerkey: string, data: Buffer, signature: Buffer): Promise<boolean> {
+    try {
+      // check signature
+      const isValid = await this.indy.cryptoVerify(signerVerkey, data, signature)
```
blu3beri, line 333: Same here.

packages/core/src/decorators/signature/SignatureDecoratorUtils.test.ts
```diff
@@ -52,7 +52,7 @@ describe('Decorators | Signature | SignatureDecoratorUtils', () => {
 
   test('signData signs json object and returns SignatureDecorator', async () => {
     const seed1 = '00000000000000000000000000000My1'
```
blu3beri, line 54: No need for this variable, right?

blu3beri (changes requested):
Some minor things. Nothing about the functionality but more I think some overlooked things. 

Great changes! Love the errors :-)

packages/core/src/agent/__tests__/Agent.test.ts
TimoGlastra: We know its defined, TypeScript doesn't. I don't think it's necessary to do for the test

packages/core/src/modules/connections/__tests__/ConnectionService.test.ts
TimoGlastra: This is intentional. Using an object made more sense to me

packages/core/src/storage/IndyStorageService.ts
TimoGlastra: I've thought about it, however implementations differ a bit so choose to not do it for now

packages/core/src/wallet/IndyWallet.ts
TimoGlastra: `this.open` already handles this, so we'll always get an aries framework error here

packages/core/src/wallet/IndyWallet.ts
TimoGlastra: I think in this case `isValid` adds some useful context to what `cryptoVerify` (and `verify()`) returns

packages/core/src/decorators/signature/SignatureDecoratorUtils.test.ts
TimoGlastra: It's used in the line below this line

blu3beri (approved):

TimoGlastra:
@jakubkoci could you please take a look at this PR? The slow review process is really slowing me down

packages/core/src/types.ts
```diff
@@ -4,7 +4,11 @@ import type { ConnectionRecord, DidCommService } from './modules/connections'
 import type { AutoAcceptCredential } from './modules/credentials/CredentialAutoAcceptType'
 import type { AutoAcceptProof } from './modules/proofs'
 import type { MediatorPickupStrategy } from './modules/routing'
-import type { Verkey, WalletConfig, WalletCredentials } from 'indy-sdk'
+
+export interface WalletConfig {
+  walletId: string
+  walletKey: string
+}
```
jakubkoci, line 11: Will we put here also other `config` and `credential` properties eventually? 

I think we can get rid of `wallet` prefix:

```suggestion
export interface WalletConfig {
  id: string
  key: string
}
```

packages/core/src/wallet/IndyWallet.ts
```diff
@@ -279,92 +264,83 @@ export class IndyWallet implements Wallet {
...
-  public signRequest(myDid: Did, request: LedgerRequest) {
-    return this.indy.signRequest(this.walletHandle, myDid, request)
+  public async verify(signerVerkey: string, data: Buffer, signature: Buffer): Promise<boolean> {
+    try {
+      // check signature
+      const isValid = await this.indy.cryptoVerify(signerVerkey, data, signature)
+
+      return isValid
+    } catch (error) {
+      throw new WalletError(`Error signing data with verkey ${signerVerkey}`, { cause: error })
```
jakubkoci, line 335:
```suggestion
      throw new WalletError(`Error verifying signature of data signed with verkey ${signerVerkey}`, { cause: error })
```

packages/core/src/modules/indy/services/IndyHolderService.ts
```diff
@@ -22,15 +24,19 @@ export class IndyHolderService {
...
-      proofRequest,
-      requestedCredentials,
-      this.indyWallet.masterSecretId,
-      schemas,
-      credentialDefinitions,
-      revocationStates
-    )
+    try {
+      return await this.indy.proverCreateProof(
+        this.indyWallet.walletHandle,
```
jakubkoci, line 29: `IndyLedgerService` is also using `IndyWallet`, but we name it just as `this.wallet`. I'm quite good with having just `this.wallet`, but if we want to use `this.indyWallet`, we should use it consistently.

jakubkoci: Maybe even just `this.wallet.handle`? 🤔

packages/core/src/error/IndySdkError.ts
```diff
@@ -0,0 +1,11 @@
+import type { IndyError } from '../utils/indyError'
+
+import { AriesFrameworkError } from './AriesFrameworkError'
+
+export class IndySdkError extends AriesFrameworkError {
```
jakubkoci, line 5: This is kind of interesting. I would assume that `IndySdk` is lower-level detail than framework level error. Therefore, I'm not sure about `IndySdkError` extending `AriesFrameworkError`

packages/core/src/types.ts
TimoGlastra: > Will we put here also other `config` and `credential` properties eventually?

Yes that was the idea. So whenever we support other storage types / wallet types we can create an extended interface for that specific type

packages/core/src/error/IndySdkError.ts
TimoGlastra: The normal `IndyError` is lower-level detail than the framework level error. However, because the normal error is not very nice in stacktracing/information we've wrapped it with a framework specific error.

To me `AriesFrameworkError` is the basis for all framework created errors (which is the case if we wrap the original error). I think it can be beneficial to have a single base error that allows to catch all framework errors. Maybe AriesFrameworkError is not the best name as the base error

jakubkoci (approved):

PR #398, opened 2021-07-20
feat(core): ledger module registerPublicDid implementation

genaris:
Add missing implementation for registerPublicDid in ledger module.

Apart from ledger source files, there were some updates in ledger setup scripts in order to properly run the tests, as it is needed to have some permissions to write public DIDs.

### DIFF
--- a/packages/core/src/modules/ledger/services/LedgerService.ts
+++ b/packages/core/src/modules/ledger/services/LedgerService.ts
@@ -11,6 +11,7 @@ import type {
   SchemaId,
   LedgerReadReplyResponse,
   LedgerWriteReplyResponse,
+  NymRole,
 } from 'indy-sdk'
 
 import { inject, scoped, Lifecycle } from 'tsyringe'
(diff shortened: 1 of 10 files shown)


packages/core/tests/ledger.test.ts
```diff
@@ -57,11 +55,26 @@ describe('ledger', () => {
...
+    if (!faberAgent.publicDid) {
+      throw new Error('Agent does not have public did.')
+    }
+
+    const targetDid = 'PNQm3CwyXbN5e39Rw3dXYx'
+    const targetVerkey = '~AHtGeRXtGjVfXALtXP9WiX'
+
+    const result = await faberAgent.ledger.registerPublicDid(targetDid, targetVerkey, 'alias', 'TRUST_ANCHOR')
+
+    await sleep(2000)
```
TimoGlastra, line 73: Is this sleep necessary? Is the result object mutated after the value is returned?

TimoGlastra (commented):
Nice work @genaris! I have one question

packages/core/tests/ledger.test.ts
genaris: True! Not necessary, forgot to remove it when copy-pasting tests :-P

packages/core/src/modules/ledger/services/LedgerService.ts
```diff
@@ -110,6 +111,34 @@ export class LedgerService {
...
+
+      const request = await this.indy.buildNymRequest(submitterDid, targetDid, verkey, alias, role || null)
+
+      const response = await this.submitWriteRequest(request, submitterDid)
+
+      this.logger.debug(`Registered public did '${targetDid}' on ledger`, {
+        response,
+      })
+
+      return targetDid
```
TimoGlastra, line 126: One final nitpick: Maybe we shouldn't return the targetDid? It's already an input parameter, so it doesn't provide extra context to the method caller to return it. WDYT?

TimoGlastra (approved):

packages/core/src/modules/ledger/services/LedgerService.ts
genaris: Yes, you are right about that. Actually I left it for consistency with other methods from ledger service such as `registerSchema` and `registerCredentialDefinition`. Do you think it'd better to return void?

In that way, in the registerPublicDid test I can do a getPublicDid right after the DID registration (something I don't want to do because that that method is tested in a separate test, but probably should be the unique 'good' way to test it).

TimoGlastra: I think it is fine for now. We can tackle this in a separate PR and overhaul all methods

PR #403, opened 2021-07-22
priority sorted transports

burdettadam:
Signed-off-by: Adam Burdett <burdettadam@gmail.com>

### DIFF
--- a/packages/core/src/agent/MessageSender.ts
+++ b/packages/core/src/agent/MessageSender.ts
@@ -36,8 +36,8 @@ export class MessageSender {
   }
 
   public registerOutboundTransporter(outboundTransporter: OutboundTransporter, priority?: number) {
-    if (priority && priority >= 0) {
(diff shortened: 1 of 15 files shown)


packages/core/src/agent/MessageSender.ts
```diff
@@ -32,14 +32,38 @@ export class MessageSender {
     this.transportService = transportService
     this.messageRepository = messageRepository
     this.logger = logger
+    this.outboundTransports = []
   }
 
-  public setOutboundTransporter(outboundTransporter: OutboundTransporter) {
-    this._outboundTransporter = outboundTransporter
+  public registerOutboundTransporter(outboundTransporter: OutboundTransporter, priority: number) {
```
TheTreek, line 38: We should have a default priority

TheTreek (changes requested):

PR #404, opened 2021-07-22
chore: update to indy-sdk-react-native

TimoGlastra:

Use indy-sdk-react-native instead of rn-indy-sdk. Best to wait until https://github.com/hyperledger/indy-sdk-react-native/pull/14 is merged


packages/react-native/package.json
```diff
@@ -29,17 +29,17 @@
...
   "devDependencies": {
     "@types/react-native": "^0.64.10",
-    "@types/rn-indy-sdk": "npm:@types/indy-sdk@^1.16.5",
+    "@types/indy-sdk-react-native": "npm:@types/indy-sdk@^1.16.6",
     "react": "17.0.1",
     "react-native": "0.64.2",
     "rimraf": "~3.0.2",
     "typescript": "~4.3.0",
-    "rn-indy-sdk": "^0.1.11",
+    "indy-sdk-react-native": ">= 0.1.13",
```
blu3beri, line 37: Would it not be better if we used `^0.1.13`? This way we don't update to a new major release accidentally.

packages/react-native/package.json
```diff
@@ -29,17 +29,17 @@
...
     "react-native-get-random-values": "^1.7.0",
     "react-native-fs": "^2.18.0"
   },
   "peerDependencies": {
-    "rn-indy-sdk": ">= 0.1.11",
+    "indy-sdk-react-native": ">= 0.1.13",
```
blu3beri, line 42: Same.

packages/react-native/package.json
TimoGlastra: Agree.

TimoGlastra:
Ready for review. Before merging i'd like to test it out in one of our mobile apps to see if everything works as expected

yarn.lock
```diff
@@ -2154,13 +2154,6 @@
   dependencies:
     "@types/node" "*"
```
blu3beri, line 2155: Should that not be a dev dependency?

TimoGlastra: yarn.lock doesn't discriminate between dev dependencies and non dev dependencies. Package.json is correct i think

blu3beri (approved):

jakubkoci (approved):

JamesKEbert (approved):
All looks good to me! I haven't had the opportunity to test installation via the package as of yet though. Should be fairly straightforward though, so I can do so and report back here if desired.

TimoGlastra:
I've tested the package directly (without AFJ) and everything seemed to work fine. Could create wallets, open pools
Etc...

PR #408, opened 2021-07-23
docs: mention monorepo

blu3beri:
This mainly adds to the README that this is currenly a monorepo. It shows all the packages and their respective versioning. Since we use the same versionig strategy for every package, it might be a bit redundant. Open for suggestions, but IMO this is a relatively nice way to do it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,8 +48,6 @@ Aries Framework JavaScript is a framework written in TypeScript for building **S
 
 ## Packages
 
-Since this is a monorepo, it contains multiple packages.
-
 <table>
   <tr>
     <th><b>Package</b></th>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -51,6 +46,47 @@
 
 Aries Framework JavaScript is a framework written in TypeScript for building **SSI Agents and DIDComm services** that aims to be **compliant and interoperable** with the standards defined in the [Aries RFCs](https://github.com/hyperledger/aries-rfcs).
 
+## Packages
```
TimoGlastra, line 49: I'd move this under the `## Features` section

README.md
```diff
@@ -51,6 +46,47 @@
...
+
+<table>
+  <tr>
+    <th><b>Package</b></th>
+    <th><b>Version</b></th>
+  </tr>
+  <tr>
+    <td>Core</td>
```
TimoGlastra, line 57: Maybe we can be more explicit:

`@aries-framework/core` instead of Core, etc..?

TimoGlastra (approved):

PR #410, opened 2021-07-26
build: fix start scripts for mediators

seriousManual:
the start scripts used in the docker-compose setup do not match with the actual name of the start scripts that are defined in the package.json


blu3beri (approved):
👍

PR #411, opened 2021-07-27
Fix bug/mediator restarts

burdettadam:


packages/core/src/agent/Agent.ts
```diff
@@ -171,14 +172,45 @@ export class Agent {
...
-      await this.mediationRecipient.setDefaultMediator(mediationRecord)
+      const invitation = await ConnectionInvitationMessage.fromUrl(mediatorConnectionsInvite)
+      // Check if invitation has been used already
+      const connections = await this.connections.getAll()
+      let defaultMediatorBootstrapped = false
+      for (const connection of connections) {
+        if (connection.invitation === invitation) {
+          this.logger.warn(
+            `Mediator Invitation in configuration has already been used to ${
+              connection.isReady ? 'make' : 'initialized'
```
JamesKEbert, line 183: 'initialize' would be better here

packages/core/src/agent/Agent.ts
```diff
@@ -171,14 +172,45 @@ export class Agent {
...
-      // TODO: add timeout to returnWhenIsConnected
-      connectionRecord = await this.connections.returnWhenIsConnected(connectionRecord.id)
-      const mediationRecord = await this.mediationRecipient.requestAndAwaitGrant(connectionRecord, 60000) // TODO: put timeout as a config parameter
-      await this.mediationRecipient.setDefaultMediator(mediationRecord)
+      const invitation = await ConnectionInvitationMessage.fromUrl(mediatorConnectionsInvite)
+      // Check if invitation has been used already
+      const connections = await this.connections.getAll()
+      let defaultMediatorBootstrapped = false
+      for (const connection of connections) {
+        if (connection.invitation === invitation) {
```
JamesKEbert, line 180: Might be smart to compare the connection ids vs the invitation itself, in case there's comparison issues or something?

packages/core/src/agent/Agent.ts
```diff
@@ -171,14 +172,45 @@ export class Agent {
...
-        autoAcceptConnection: true,
-      })
-
-      // TODO: add timeout to returnWhenIsConnected
-      connectionRecord = await this.connections.returnWhenIsConnected(connectionRecord.id)
-      const mediationRecord = await this.mediationRecipient.requestAndAwaitGrant(connectionRecord, 60000) // TODO: put timeout as a config parameter
-      await this.mediationRecipient.setDefaultMediator(mediationRecord)
+      const invitation = await ConnectionInvitationMessage.fromUrl(mediatorConnectionsInvite)
+      // Check if invitation has been used already
+      const connections = await this.connections.getAll()
```
TimoGlastra, line 177: Retrieving all connections is not a very scalable solution. We already store the invitation key, so I think we should be able to do the following:

```ts
const invitationKey = invitation.recipientKeys[0]
const connection = await connectionRepository.findSingleByQuery(invitationKey)

if (connection) {
  // ...
}
```

I think this should be a method exposed from the mediator or connection module, so we don't have to interact with connection repository directly

packages/core/src/agent/Agent.ts
```diff
@@ -171,14 +172,45 @@ export class Agent {
...
+      let defaultMediatorBootstrapped = false
+      for (const connection of connections) {
+        if (connection.invitation === invitation) {
+          this.logger.warn(
+            `Mediator Invitation in configuration has already been used to ${
+              connection.isReady ? 'make' : 'initialized'
+            } a connection`
+          )
+          if (connection.isReady) {
+            const mediators = await this.mediationRecipient.getMediators()
```
TimoGlastra, line 187: This way we don't have to retrieve all mediators:

```suggestion
            const mediator = await this.mediationRecipient.findByConnectionId(connection.id)
```

TimoGlastra (commented):
I see you're still in draft, but leaving some initial comments

packages/core/src/agent/Agent.ts
```diff
@@ -171,14 +172,45 @@ export class Agent {
...
+      const connections = await this.connections.getAll()
+      let defaultMediatorBootstrapped = false
+      for (const connection of connections) {
+        if (connection.invitation === invitation) {
+          this.logger.warn(
+            `Mediator Invitation in configuration has already been used to ${
+              connection.isReady ? 'make' : 'initialized'
+            } a connection`
+          )
+          if (connection.isReady) {
```
JamesKEbert, line 186: I don't know if we need to actually check if it's ready? Thoughts?

packages/core/src/agent/Agent.ts
burdettadam: I see. I will incorporate this.

packages/core/src/agent/Agent.ts
burdettadam: if the connection is not ready, we have not requested mediation over it. I think it's a needed check.

PR #413, opened 2021-07-28
fix(redux-store): add reducers to initializeStore

karimStekelenburg:


packages/redux-store/src/store.ts
```diff
@@ -6,10 +6,13 @@ import { useDispatch, useSelector } from 'react-redux'
 
 import { agentSlice } from './slices/agent'
 import { connectionsSlice } from './slices/connections/connectionsSlice'
+import { credentialsSlice, proofsSlice } from './slices'
```
TimoGlastra, line 9: Inconsistent import

karimStekelenburg: Good point, fixed

TimoGlastra (approved):

PR #414, opened 2021-08-01
docs: add overview

MosCD3:
I found this very useful to remove the confusion many developers face when first trying the framework and thinking everything works like normal API call.


swcurran:
I like the intention, but I don't think the drawing of the communicating Agents is accurate. It looks like from the image that the two agents communicate through an instance of AFJ.   Suggest rethinking the image.

MosCD3:
> I like the intention, but I don't think the drawing of the communicating Agents is accurate. It looks like from the image that the two agents communicate through an instance of AFJ. Suggest rethinking the image.

Good point

how about that

Screen Shot 2021-08-01 at 7 47 31 PM

swcurran:
Looks better to me.  Although you do have to fix -- "JavaScript" :-)

Nice work -- much appreciated.

docs/getting-started/overview.md
```diff
@@ -0,0 +1,42 @@
...
+
+# Http communication protocol
+
+In most applications, the client communicates with the server using http protocol where a request is made to the server and the client awaits the response from the server with a specific timeout after which a timeout exception is raised if we didn’t get any response from server.
+
+![http request](../images/dcd1.png)
+
+# Agent to agent communication
+
+On the other hand, agents communicate using [didcomm](https://github.com/hyperledger/aries-rfcs/tree/master/concepts/0005-didcomm) communication protocol, while this protocols have much more to talk about but the most important concern here is how the communication flow goes. For the sake of demonstration, lets assume 2 agents want to communicate Agent Alice and Agent Bob.
```
TimoGlastra, line 13:
```suggestion
On the other hand, agents communicate using [DIDComm](https://github.com/hyperledger/aries-rfcs/tree/master/concepts/0005-didcomm) communication protocols. While protocols have much more to talk about the most important concern here is how the communication flow goes. For the sake of demonstration, lets assume 2 agents want to communicate, Agent Alice and Agent Bob.
```

docs/getting-started/overview.md
```diff
@@ -0,0 +1,42 @@
...
+
+1- Agent Alice will send a connection request to agent Bob either directly or through a mediator (another routing agent) using outbound transporter
```
TimoGlastra, line 15:
```suggestion
1- Agent Alice will send a connection request to Agent Bob either directly or through a mediator (another routing agent) using outbound transporter
```

docs/getting-started/overview.md
```diff
@@ -0,0 +1,42 @@
...
+2- Agent Bob receives the message (through inbound transporter) and process the message 3- Agent Bob sends the response in a new request (using outbound TP) sent back to agent Alice
```
TimoGlastra, line 16:
```suggestion
2- Agent Bob receives the message (through inbound transporter) and process the message
3- Agent Bob sends the response in a new request (using outbound TP) sent back to agent Alice
```

docs/getting-started/overview.md
```diff
@@ -0,0 +1,42 @@
...
+4- Agent Alice receives the message through the in-bound TP
+5- Agent Alice process the message (under the hood through Aries) and raises an event with attached data relevant to communication context
+6- The controller (your app) receives the data using a subscribed callback to the corresponding event types and act upon it
```
TimoGlastra, line 19: This doesn't render very nicely. I'd suggest to use markdown numbered lists

1. ...
2. addd

docs/getting-started/overview.md
```diff
@@ -0,0 +1,42 @@
...
+
+![http request](../images/dcd1.png)
+
+# Agent to agent communication
+
+On the other hand, agents communicate using [didcomm](https://github.com/hyperledger/aries-rfcs/tree/master/concepts/0005-didcomm) communication protocol, while this protocols have much more to talk about but the most important concern here is how the communication flow goes. For the sake of demonstration, lets assume 2 agents want to communicate Agent Alice and Agent Bob.
+
+1- Agent Alice will send a connection request to agent Bob either directly or through a mediator (another routing agent) using outbound transporter
+2- Agent Bob receives the message (through inbound transporter) and process the message 3- Agent Bob sends the response in a new request (using outbound TP) sent back to agent Alice
+4- Agent Alice receives the message through the in-bound TP
```
TimoGlastra, line 17:
```suggestion
4- Agent Alice receives the message through the inbound TP
```

TimoGlastra: It wasn't directly clear to me that TP stands for Transport

docs/getting-started/overview.md
```diff
@@ -0,0 +1,42 @@
...
+5- Agent Alice process the message (under the hood through Aries) and raises an event with attached data relevant to communication context
+6- The controller (your app) receives the data using a subscribed callback to the corresponding event types and act upon it
+
+![agents communication](../images/dcd2.png)
+
+So to wrap things up, as a developer for a controller app that will use the Aries framework you will not worry about most of that scenarios and negotiations going under the hood, the only thing that you will do as a controller is initiate the connection and wait for an event to be raised that has the parameters you are waiting for to see if the connection succeeded or the a specific credential have been received.
```
TimoGlastra, line 23:
```suggestion
So to wrap things up, as a developer for a controller app that will use the Aries framework you will not worry about most of these scenarios and negotiations going on under the hood. The only thing that you will do as a controller is initiate the connection and wait for an event with specific parameters to be raised to see if the connection succeeded or the specific credential has been received.
```

docs/getting-started/overview.md
```diff
@@ -0,0 +1,42 @@
...
+
+From the previous diagrams the normal controller(app) structure would consist of
```
TimoGlastra, line 25:
```suggestion
From the previous diagrams the normal controller (app) structure would consist of
```

docs/getting-started/overview.md
```diff
@@ -0,0 +1,42 @@
...
+
+## [Agent](0-agent.md)
+
+The main object that holds the core functionalities of the Aries framework.
+
+## Event handlers
+
+A call back methods passed to the agent event handler to be called on different events such as
```
TimoGlastra, line 33:
```suggestion
A callback method passed to the agent event handler to be called on different events such as 
```

docs/getting-started/overview.md
```diff
@@ -0,0 +1,42 @@
...
+## Event handlers
+
+A call back methods passed to the agent event handler to be called on different events such as 
+- Message received from other agent
+- A recent connection with other agent has changed state
+- A credential received or has a state changed
+
+## [Transporters](transports.md)
+
+Services that will handle the outbound and in-bound transports. Remember we mentioned that unlike http request which happens on one channel, the connection here has two ways one outgoing using the out-bound transporter and one incoming using the inbound transporter
```
TimoGlastra, line 40:
```suggestion
Services that will handle the outbound and inbound transports. Remember we mentioned that unlike http request which happens on one channel, the connection here has two ways one outgoing using the outbound transporter and one incoming using the inbound transporter
```

TimoGlastra (commented):
Thanks for the overview @MosCD3, we really need more general docs in the repo. I have a few suggestions on wording

TimoGlastra:
The formatting is failing. You can run `yarn format` in the root of the repo to automatically fix it

docs/getting-started/overview.md
MosCD3: Yes @TimoGlastra I am working on more documentation and guides
And yes the list is formatted

MosCD3:
> The formatting is failing. You can run `yarn format` in the root of the repo to automatically fix it

Done

TimoGlastra (approved):

PR #416, opened 2021-08-02
feat: added decline credential offer method

seajensen:
# Summary of Changes

Added "decline offer" method to credential modules. 

# Related Issues

N/A

# Pull Request Checklist

Signed-off-by: seajensen <seajensen@gmail.com>


TimoGlastra:
Thanks @seajensen,

I can't seem to find a `declineOffer` method on the `CredentialService`. Am I missing something?

https://github.com/Indicio-tech/aries-framework-javascript/blob/feature/declineCredential/packages/core/src/modules/credentials/services/CredentialService.ts

seajensen:
> Thanks @seajensen,
> 
> I can't seem to find a `declineOffer` method on the `CredentialService`. Am I missing something?
> 
> https://github.com/Indicio-tech/aries-framework-javascript/blob/feature/declineCredential/packages/core/src/modules/credentials/services/CredentialService.ts

Apologies. The previous commit was missing some important pieces, but everything should be good now. 

Thank you.

packages/core/src/modules/credentials/CredentialState.ts
```diff
@@ -8,6 +8,8 @@ export enum CredentialState {
   ProposalReceived = 'proposal-received',
   OfferSent = 'offer-sent',
   OfferReceived = 'offer-received',
+  OfferDeclinedReceived = 'offer-declined-received',
+  OfferDeclinedSent = 'offer-declined-sent',
```
JamesKEbert, line 12: Given issue-credential v1, this would be an internal state as we won't send anything to the issuer when declining an offer (such as a problem report), so I think we want one state here--I'd recommend `OfferDeclined = 'offer-declined'`.

packages/core/src/modules/credentials/__tests__/CredentialState.test.ts
```diff
@@ -6,6 +6,8 @@ describe('CredentialState', () => {
     expect(CredentialState.ProposalReceived).toBe('proposal-received')
     expect(CredentialState.OfferSent).toBe('offer-sent')
     expect(CredentialState.OfferReceived).toBe('offer-received')
+    expect(CredentialState.OfferDeclinedReceived).toBe('offer-declined-received')
+    expect(CredentialState.OfferDeclinedSent).toBe('offer-declined-sent')
```
JamesKEbert, line 10: These should change based off of the state changes suggested above ^

JamesKEbert (changes requested):
Looks good--minor state changes requested

packages/core/src/modules/credentials/__tests__/CredentialService.test.ts
```diff
@@ -955,4 +955,64 @@ describe('CredentialService', () => {
...
+
+      // when
+      await credentialService.declineOffer(credential)
+
+      // then
+      expect(repositoryUpdateSpy).toHaveBeenCalledTimes(1)
+      const [[updatedCredentialRecord]] = repositoryUpdateSpy.mock.calls
+      expect(updatedCredentialRecord).toMatchObject({
+        state: CredentialState.OfferDeclinedSent,
+      })
```
TimoGlastra, line 982: I prefer to have this syntax:

```suggestion
	  expect(repositoryUpdateSpy).toHaveBeenNthCalledWith(1, {
        state: CredentialState.OfferDeclinedSent,
      })
```

packages/core/src/modules/credentials/CredentialState.ts
TimoGlastra: I'm thinking whether we can go even more basic, and just store state `declined`? This way it can also be used for the issuer side, and other states. We can then also make `declineOffer` more generic, as it works for other states than `offer-received`.

Just a thought, I'm also fine with using `offer-declined` state.

JamesKEbert: I'm in favor of that--so you're thinking no state validation for the `declineOffer` method since the credential can be in any state?

TimoGlastra (approved):
LGTM. Nice!

packages/core/src/modules/credentials/CredentialsModule.ts
```diff
@@ -305,6 +305,17 @@ export class CredentialsModule {
...
   }
 
+  /**
+   * Declines an offer as holder
+   * @param credentialRecordId the id of the credential to be declined
+   * @returns credential record that was declined
+   */
+  public async declineOffer(credentialRecordId: string) {
+    const credentialRecord = await this.credentialService.getById(credentialRecordId)
+    await this.credentialService.declineOffer(credentialRecord)
```
blu3beri, line 315: I think returning the return value of this function would be cleaner. However since in `this.credentialService.declineOffer(...)` we modify the parameter we receive it does not matter as much.

JamesKEbert (approved):
LGTM!

blu3beri:
> I think good to go. I wanted to merge but I see @blu3beri added a note and I'm not sure if it's just a non-blocking suggestion or not and I don't want to step on his toes :)

It is 100% non-blocking for me. So I am good with merging it.

PR #418, opened 2021-08-04
updated initializeAgent to take an agent param instead of initializeS…

NeilSMyers:
…tore

Signed-off-by: NeilSMyers <mmyersneil@gmail.com>

### DIFF
--- a/packages/redux-store/package.json
+++ b/packages/redux-store/package.json
@@ -2,7 +2,8 @@
   "name": "@aries-framework/redux-store",
   "main": "build/index",
   "types": "build/index",
-  "version": "0.0.0",
+  "version": "1.0.0",
+  "license": "Apache-2.0",
   "files": [
     "build"
   ],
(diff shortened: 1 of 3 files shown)


TimoGlastra (commented):
Thanks @NeilSMyers

Could you give some context on how this flow works? The reason we need the agent in initializeStore is to set it as the extra parameter. This makes the agent available in all async thunks. Without this the other thunks (such as getAllConnections) won't work.

NeilSMyers:
oh! that makes sense. I didn't realize that. Ok! nevermind, I thought I may have been missing something obvious

karimStekelenburg (changes requested):
As @TimoGlastra mentioned, the agent needs to be passed into the extraArguments object within the `configureStore` function call. If you want to call the agent from within a thunk, you can access it by calling `thunkApi.extra.agent` :)

PR #419, opened 2021-08-05
fix: mediator transports

burdettadam:
This PR has code that corrects issues related to mediation.
Corrections
- multiple restarts with the same mediation invitation 
- second sorting of connection transport services based on inbound services
- multiple outbound transport support
  - add transport is now register transport, which allows using multiple types of transport.

### DIFF
--- a/packages/core/src/agent/Agent.ts
+++ b/packages/core/src/agent/Agent.ts
@@ -118,12 +118,12 @@ export class Agent {
     this.inboundTransporter = inboundTransporter
   }
 
-  public setOutboundTransporter(outboundTransporter: OutboundTransporter) {
-    this.messageSender.setOutboundTransporter(outboundTransporter)
+  public registerOutboundTransporter(outboundTransporter: OutboundTransporter) {
+    this.messageSender.registerOutboundTransporter(outboundTransporter)
   }
 
-  public get outboundTransporter() {
-    return this.messageSender.outboundTransporter
+  public get outboundTransporters() {
+    return this.messageSender.outboundTransporters
   }
 
   public get events() {
@@ -162,23 +162,15 @@ export class Agent {
       await this.inboundTransporter.start(this)
     }
 
-    if (this.outboundTransporter) {
-      await this.outboundTransporter.start(this)
+    for (const transport of this.messageSender.outboundTransporters) {
+      transport.start(this)
     }
 
     // Connect to mediator through provided invitation if provided in config
     // Also requests mediation ans sets as default mediator
     // Because this requires the connections module, we do this in the agent constructor
     if (mediatorConnectionsInvite) {
-      // Assumption: processInvitation is a URL-encoded invitation
-      let connectionRecord = await this.connections.receiveInvitationFromUrl(mediatorConnectionsInvite, {
-        autoAcceptConnection: true,
-      })
-
-      // TODO: add timeout to returnWhenIsConnected
-      connectionRecord = await this.connections.returnWhenIsConnected(connectionRecord.id)
-      const mediationRecord = await this.mediationRecipient.requestAndAwaitGrant(connectionRecord, 60000) // TODO: put timeout as a config parameter
-      await this.mediationRecipient.setDefaultMediator(mediationRecord)
+      await this.mediationRecipient.provision(mediatorConnectionsInvite)
     }
 
     await this.mediationRecipient.initialize()
@@ -192,7 +184,9 @@ export class Agent {
     this.agentConfig.stop$.next(true)
 
     // Stop transports
-    await this.outboundTransporter?.stop()
+    for (const transport of this.messageSender.outboundTransporters) {
+      transport.stop()
+    }
     await this.inboundTransporter?.stop()
 
     // close/delete wallet if still initialized
--- a/packages/core/src/agent/MessageSender.ts
+++ b/packages/core/src/agent/MessageSender.ts
@@ -266,10 +266,10 @@ export class MessageSender {
     transportPriority?: TransportPriorityOptions
   ) {
     this.logger.debug(
-      `Retrieving services for connection '${connection.id}'${' (' && connection.theirLabel && ')'}${
-        (transportPriority?.restrictive && ' restrictively ',
-        transportPriority && ' by priority of the follow schemes: ' + transportPriority.schemes)
-      }`
+      `Retrieving services for connection '${connection.id}' (${connection.theirLabel})`,
+      {
+      	transportPriority
+      }
     )
     // Retrieve DIDComm services
     const allServices = this.transportService.findDidCommServices(connection)
--- a/packages/core/src/agent/MessageSender.ts
+++ b/packages/core/src/agent/MessageSender.ts
@@ -1,4 +1,4 @@
-import type { DidCommService, ConnectionRecord } from '../modules/connections'
+import type { DidCommService, ConnectionRecord, Connection } from '../modules/connections'
 import type { OutboundTransporter } from '../transport/OutboundTransporter'
 import type { OutboundMessage, OutboundPackage, WireMessage } from '../types'
 import type { AgentMessage } from './AgentMessage'
@@ -15,6 +15,11 @@ import { MessageRepository } from '../storage/MessageRepository'
 
 import { EnvelopeService } from './EnvelopeService'
 import { TransportService } from './TransportService'
+
+export interface TransportPriorityOptions {
+  schemes: string[]
+  restrictive?: boolean
+}
 
 @scoped(Lifecycle.ContainerScoped)
 export class MessageSender {
@@ -39,18 +44,6 @@ export class MessageSender {
 
   public registerOutboundTransporter(outboundTransporter: OutboundTransporter) {
     this.outboundTransports.push(outboundTransporter)
-  }
-
-  public get supportedTransportSchemes() {
-    // map all supported schema into a new list
-    // reduce the list of listed schemas into a single list
-    // remove duplicates by creating a new set and spreading it into a list
-    const schemeTuples = this.outboundTransporters.map((transport) => transport.supportedSchemes)
-
-    const allSchemes = new Array<string>().concat(...schemeTuples)
-    const uniqueSchemes = [...new Set(allSchemes)]
-
-    return uniqueSchemes
   }
 
   public get outboundTransporters() {
@@ -94,7 +87,7 @@ export class MessageSender {
   }: {
     connection: ConnectionRecord
     packedMessage: WireMessage
-    options?: { preferredTransport: string }
+    options?: { transportPriority?: TransportPriorityOptions }
   }) {
     // Try to send to already open session
     const session = this.transportService.findSessionByConnectionId(connection.id)
@@ -108,29 +101,14 @@ export class MessageSender {
     }
 
     // Retrieve DIDComm services
-    const allServices = this.transportService.findDidCommServices(connection)
-    let reachableServices = allServices.filter((s) => !isDidCommTransportQueue(s.serviceEndpoint))
-    if (options && options.preferredTransport) {
-      reachableServices = [
-        ...new Set(
-          reachableServices
-            .filter((s) => s.serviceEndpoint.split(':')[0] === options.preferredTransport)
-            .concat(reachableServices)
-        ),
-      ]
-    }
-    const queueService = allServices.find((s) => isDidCommTransportQueue(s.serviceEndpoint))
-
-    this.logger.debug(
-      `Found ${allServices.length} services for message to connection '${connection.id}' (${connection.theirLabel})`
-    )
+    const { services, queueService } = await this.retrieveServicesByConnection(connection, options?.transportPriority)
 
     if (this.outboundTransporters.length === 0 && !queueService) {
       throw new AriesFrameworkError('Agent has no outbound transporter!')
     }
 
     // Loop trough all available services and try to send the message
-    for await (const service of reachableServices) {
+    for await (const service of services) {
       this.logger.debug(`Sending outbound message to service:`, { service })
       try {
         const protocol = service.serviceEndpoint.split(':')[0]
@@ -171,7 +149,12 @@ export class MessageSender {
     throw new AriesFrameworkError(`Message is undeliverable to connection ${connection.id} (${connection.theirLabel})`)
   }
 
-  public async sendMessage(outboundMessage: OutboundMessage, options?: { preferredTransport: string }) {
+  public async sendMessage(
+    outboundMessage: OutboundMessage,
+    options?: {
+      transportPriority?: TransportPriorityOptions
+    }
+  ) {
     const { connection, payload } = outboundMessage
 
     this.logger.debug('Send outbound message', {
@@ -191,21 +174,10 @@ export class MessageSender {
     }
 
     // Retrieve DIDComm services
-    const allServices = this.transportService.findDidCommServices(connection)
-    const reachableServices = allServices.filter((s) => !isDidCommTransportQueue(s.serviceEndpoint))
-    if (options?.preferredTransport) {
-      reachableServices.sort((f, s) => {
-        return f.serviceEndpoint.split(':')[0] === options.preferredTransport ? 1 : 0
-      })
-    }
-    const queueService = allServices.find((s) => isDidCommTransportQueue(s.serviceEndpoint))
-
-    this.logger.debug(
-      `Found ${allServices.length} services for message to connection '${connection.id}' (${connection.theirLabel})`
-    )
+    const { services, queueService } = await this.retrieveServicesByConnection(connection, options?.transportPriority)
 
     // Loop trough all available services and try to send the message
-    for await (const service of reachableServices) {
+    for await (const service of services) {
       try {
         // Enable return routing if the
         const shouldUseReturnRoute = !this.transportService.hasInboundEndpoint(connection.didDoc)
@@ -289,6 +261,48 @@ export class MessageSender {
         break
       }
     }
+  }
+
+  private async retrieveServicesByConnection(
+    connection: ConnectionRecord,
+    transportPriority?: TransportPriorityOptions
+  ) {
+    this.logger.debug(
+      `Retrieving services for connection '${connection.id}'${' (' && connection.theirLabel && ')'}${
+        (transportPriority?.restrictive && ' restrictively ',
+        transportPriority && ' by priority of the follow schemes: ' + transportPriority.schemes)
+      }`
+    )
+    // Retrieve DIDComm services
+    const allServices = this.transportService.findDidCommServices(connection)
+
+    //Separate queue service out
+    const services = allServices.filter((s) => !isDidCommTransportQueue(s.serviceEndpoint))
+    const queueService = allServices.find((s) => isDidCommTransportQueue(s.serviceEndpoint))
+
+    //If restrictive will remove services not listed in schemes list
+    if (transportPriority?.restrictive) {
+      services.filter((service) => {
+        const serviceSchema = service.serviceEndpoint.split(':')[0]
+        return transportPriority.schemes.includes(serviceSchema)
+      })
+    }
+
+    //If transport priority is set we will sort services by our priority
+    if (transportPriority?.schemes) {
+      services.sort(function (a, b) {
+        const aScheme = a.serviceEndpoint.split(':')[0]
+        const bScheme = b.serviceEndpoint.split(':')[0]
+        return transportPriority?.schemes.indexOf(aScheme) - transportPriority?.schemes.indexOf(bScheme)
+      })
+    }
+
+    this.logger.debug(
+      `Retrieved ${services.length} services for message to connection '${connection.id}'${
+        ' (' && connection.theirLabel && ')'
+      }`
+    )
+    return { services, queueService }
   }
 }
 
--- a/packages/core/src/agent/MessageSender.ts
+++ b/packages/core/src/agent/MessageSender.ts
@@ -265,12 +265,9 @@ export class MessageSender {
     connection: ConnectionRecord,
     transportPriority?: TransportPriorityOptions
   ) {
-    this.logger.debug(
-      `Retrieving services for connection '${connection.id}' (${connection.theirLabel})`,
-      {
-      	transportPriority
-      }
-    )
+    this.logger.debug(`Retrieving services for connection '${connection.id}' (${connection.theirLabel})`, {
+      transportPriority,
+    })
     // Retrieve DIDComm services
     const allServices = this.transportService.findDidCommServices(connection)
 
@@ -296,9 +293,7 @@ export class MessageSender {
     }
 
     this.logger.debug(
-      `Retrieved ${services.length} services for message to connection '${connection.id}'${
-        ' (' && connection.theirLabel && ')'
-      }`
+      `Retrieved ${services.length} services for message to connection '${connection.id}'(${connection.theirLabel})'`
     )
     return { services, queueService }
   }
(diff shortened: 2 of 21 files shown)


packages/core/src/agent/MessageSender.ts
```diff
@@ -93,26 +115,31 @@ export class MessageSender {
     }
 
     // Retrieve DIDComm services
-    const allServices = this.transportService.findDidCommServices(connection)
+    const allServices = this.transportService.findDidCommServices(connection, this.supportedTransports)
```
TimoGlastra, line 118: As I understand the code now, it seems like `didcomm:transport/queue` is not taken into account here. Maybe I'm missing something, but that would mean all `didcomm:tranport/queue` services are ignored.

As I'm thinking of it right now, maybe we can do something like the following:

1. `findDidCommServices` returns a diddoc based priority sorted list of all services (already did that)
2. We loop through the didoc services and for each service we check if we can send it using one of the registered transports (`transport.suportedSchemes`)
3. If none of the inbound or outbound transports can deliver the message, we check if a queue service is included. If this is the case, we queue the message


Addition could be made between 1. and 2. where we sort the return services from `findDidCommServices`, based on a preferred protocol scheme for that message. This would be useful in the mediation recipient class where we do message pickup from the mediator.

packages/core/src/agent/MessageSender.ts
```diff
@@ -250,7 +275,14 @@ export class MessageSender {
     }
 
     const outboundPackage = await this.packMessage({ message, keys, endpoint: service.serviceEndpoint })
-    await this.outboundTransporter.sendMessage(outboundPackage)
+    outboundPackage.endpoint = service.serviceEndpoint
+    const protocol = outboundPackage.endpoint.split(':')[0]
```
TimoGlastra, line 279: Maybe we can add `protocolScheme` getter to the service class?

so we can do 
```
transport.supportedSchemes.includes(service.protocolScheme)
```

TimoGlastra (changes requested):
Thanks @burdettadam!

I think this PR packs some good changes with some other changes that need some ironing out. I think it would be ideal to split up the PR, as the fixes for the mediator should be resolved ASAP I think

packages/core/src/agent/Agent.ts
```diff
@@ -162,23 +162,51 @@ export class Agent {
...
-        autoAcceptConnection: true,
-      })
-
-      // TODO: add timeout to returnWhenIsConnected
-      connectionRecord = await this.connections.returnWhenIsConnected(connectionRecord.id)
-      const mediationRecord = await this.mediationRecipient.requestAndAwaitGrant(connectionRecord, 60000) // TODO: put timeout as a config parameter
-      await this.mediationRecipient.setDefaultMediator(mediationRecord)
+      const invitation = await ConnectionInvitationMessage.fromUrl(mediatorConnectionsInvite)
+      // Check if invitation has been used already
+      const connections = await this.connections.getAll()
```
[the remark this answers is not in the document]
JamesKEbert: Great point! +1

packages/core/src/agent/Agent.ts
```diff
@@ -162,23 +162,51 @@ export class Agent {
...
+      let defaultMediatorBootstrapped = false
+      for (const connection of connections) {
+        if (connection.invitation?.id === invitation.id) {
+          this.logger.warn(
+            `Mediator Invitation in configuration has already been used to ${
+              connection.isReady ? 'make' : 'initialize'
```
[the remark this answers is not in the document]
JamesKEbert: I tend to agree--I'd potentially rather have two logs that call out whether the connection has been established **and** mediation granted.

packages/core/src/agent/MessageSender.ts
```diff
@@ -34,14 +34,34 @@ export class MessageSender {
     this.transportService = transportService
     this.messageRepository = messageRepository
     this.logger = logger
+    this.outboundTransports = []
   }
 
-  public setOutboundTransporter(outboundTransporter: OutboundTransporter) {
-    this._outboundTransporter = outboundTransporter
+  public registerOutboundTransporter(outboundTransporter: OutboundTransporter, priority?: number) {
+    if (typeof priority != 'undefined' && priority >= 0 && priority < this.outboundTransports.length) {
```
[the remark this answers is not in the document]
JamesKEbert: I _think_ that could make fairly good sense. How would it work if we sent two messages to the same agent, and one we sent with our preferredTransport set and the other not, would we essentially open two separate transports to that agent? Or do we value open transports first? And then following with their DIDDoc preference? 
Could it also make sense to set our internal preference for if they don't designate a preference (such as with ACA-Py IIRC doesn't set transport preferences by default).
An additional worry I have here is that we'd ultimately prefer to be able to send all messages as soon as viable to the mediator over WS to save on consumed resources. This means that the trust ping sent as the ack as a part of the connections protocol would open a WS connection.

JamesKEbert:
Awesome work here @burdettadam! I'd agree here with @TimoGlastra that splitting this into multiple PRs is likely helpful to work through the various aspects. 
Also @TimoGlastra happy to talk with you guys on a call or at the AFJ call on Thursday if helpful.

tests/transport/SubjectOutboundTransport.ts
```diff
@@ -11,7 +11,7 @@ export class SubjectOutboundTransporter implements OutboundTransporter {
   private ourSubject: Subject<SubjectMessage>
   private subjectMap: { [key: string]: Subject<SubjectMessage> | undefined }
 
-  public supportedSchemes = []
+  public supportedSchemes = ['subject', 'http', 'https', 'rxjs']
```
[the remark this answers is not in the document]
burdettadam: I was going to ask that. I will update this to use `rxjs`.

packages/core/src/agent/MessageSender.ts
burdettadam: if I am not mistaken, sending a message on a connection with their did_doc containing only services with `didcomm:transport/queue` would error without explicit details, no queueing supported. The code ignores them. This will need to be updated with the check for a queue service described in suggestion 3.

Moving the preferred protocol scheme sort into the mediation recipient class would work for message pickup from mediators, but would not catch any other message sending. A nice feature of having the sort in the message sender is speed. While establishing a connection between an edge agent and an agent that supports WebSockets, the edge agent may initialize the connection protocol using HTTP, once the edge agent receives the other agent's did_doc the edge agent will finish establishing the connection using WebSockets. Why have multiple services in a did_doc if we are restricted to use the service a connection invitation is created with?

We could try and build a preferred protocol scheme by using `findDidCommServices` on senders did_doc. This will return the sender's inbound services, not the senders outbound services. I assumed we would need to update the did_doc services to have services that define sending services alongside the receiving services. To avoid changing/learning the definition of a service in a did_doc and supported code changes we opted to do list sorting and filtering based on a list of supported schemas returned from a list of registered transporters.
If you could provide an example of a did_doc with services that represent supported outbound transports we can update the message sender to use this. Another issue arises if we use the sender's did_doc services to build a preferred protocol list. connections and the did_docs associated are persisted, while transporters are not. We would need to check the did_doc services at the start of AFJ and update them along with the agent on the other end of the connection.

packages/core/src/agent/TransportService.ts
```diff
@@ -33,9 +33,31 @@ export class TransportService {
...
+        serviceB: { serviceEndpoint: string; priority: number }
+      ) {
+        const protocolA = serviceA.serviceEndpoint.split(':')[0] || ''
+        const protocolB = serviceB.serviceEndpoint.split(':')[0] || ''
+        const preferred =
+          (supportedProtocolsIndexTable.get(protocolA) || 0) - (supportedProtocolsIndexTable.get(protocolB) || 0)
+        const priority = serviceA.priority - serviceB.priority
+        return preferred || priority
+      })
+      return filteredServices
```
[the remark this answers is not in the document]
jakubkoci: I agree with Timo here. We should always prefer didComm service priorities and do not filter them according currently supported transports. Filtering can happen after we have more services without any specific priority.

jakubkoci (commented):
Just added a few comments for the case you discussed that during the call today.

packages/core/src/agent/Agent.ts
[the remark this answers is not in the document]
burdettadam: invitation key is the recipient key of the invitation? what if there is more than one recipient key?

packages/core/src/agent/Agent.ts
```diff
@@ -162,23 +162,51 @@ export class Agent {
...
+            } a connection`
+          )
+          if (connection.isReady) {
+            const mediator = await this.mediationRecipient.findByConnectionId(connection.id)
```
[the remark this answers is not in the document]
burdettadam: this code is to report information about mediator connection invitations that have already been used. To recover from a connection from a mediator invitation that is not ready the consumer of AFJ would have to remove the invitation from the config, restart AFJ, and remove the connection or get it in a ready state. I will throw an error.

packages/core/src/agent/MessageSender.ts
burdettadam: I have updated the code. `didcomm:transport/queue` is being handled correctly, I didn't notice that after the merge, no need to fix it. 
 The proposed flow is the intended behavior of this code except for priority sort. instead of moving priority sorting into mediation messages I have added an optional parameter priority with a default in send message and send package. `sendMessage(outboundMessage: OutboundMessage, priority = false)`. 
 If priority is true we pass in a secondary list to findDidCommServices, otherwise an empty list. 
 ```
     // Retrieve DIDComm services
    let allServices: Array<DidCommService | IndyAgentService>
    if (priority) {
      allServices = this.transportService.findDidCommServices(connection, this.supportedTransportSchemes)
    } else {
      allServices = this.transportService.findDidCommServices(connection, [])
    }
 ```
 Now any messaging were a consumer of AFJ wishes to prioritize supported transports can be sent by setting the priority to true. Enabling desired features for mediated messages while keeping the current functionality of the message sender.

packages/core/src/agent/TransportService.ts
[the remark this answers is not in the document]
burdettadam: I have a compromise. 
```
    const allServices = this.transportService.findDidCommServices(
      connection,
      priority ? this.supportedTransportSchemes : []
    )
```
with priority default to false. I think this is a flexible solution that provides my intended feature while maintaining current behavior.

packages/core/src/agent/MessageSender.ts
```diff
@@ -34,14 +34,34 @@ export class MessageSender {
...
+    // remove duplicates by creating a new set and spreading it into a list
+    return [
+      ...new Set(
+        (this.outboundTransporters || [])
+          .map((transportTuple) => transportTuple.supportedSchemes)
+          .reduce((prev, cur) => {
+            return prev.concat(cur)
+          })
+      ),
+    ]
```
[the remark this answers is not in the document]
burdettadam: I tried this and was getting a syntax error. I will revisit this.

packages/core/src/agent/MessageSender.ts
burdettadam: I would use a `protocolScheme` getter if the service class had it. We should add it.

packages/core/src/modules/routing/RecipientModule.ts
```diff
@@ -178,6 +179,50 @@ export class RecipientModule {
...
+          const mediator = await this.findByConnectionId(connection.id)
+          if (mediator) {
+            this.agentConfig.logger.warn(
+              `Mediator Invitation in configuration has already been ${
+                mediator.isReady ? 'granted' : 'requested'
+              } mediation`
+            )
+            if (mediator.isReady) {
+              defaultMediatorBootstrapped = true
+            }
```
burdettadam, line 209: if the mediator is not ready, throw new Errror('')

packages/core/src/agent/MessageSender.ts
```diff
@@ -94,25 +112,40 @@ export class MessageSender {
...
+        const protocol = service.serviceEndpoint.split(':')[0]
+        for (const transport of this.outboundTransporters) {
+          if (transport.supportedSchemes.includes(protocol)) {
+            await transport.sendMessage({
+              payload: packedMessage,
+              endpoint: service.serviceEndpoint,
+            })
+            break
+          }
+        }
```
TimoGlastra, line 148: I think we should extract this to a separate method as it is duplicated a few times:

```suggestion
        const transports = this.findTransportsForService(service)
        for (const transport of transports) {
            await transport.sendMessage({
              payload: packedMessage,
              endpoint: service.serviceEndpoint,
            })
            break
        }
```

packages/core/src/agent/MessageSender.ts
TimoGlastra: Awesome! Could you add it as part of this PR?

packages/core/src/agent/MessageSender.ts
```diff
@@ -167,6 +196,11 @@ export class MessageSender {
     // Retrieve DIDComm services
     const allServices = this.transportService.findDidCommServices(connection)
     const reachableServices = allServices.filter((s) => !isDidCommTransportQueue(s.serviceEndpoint))
+    if (options && options.preferredTransport) {
```
TimoGlastra, line 199: I see you use sort here, but filter in another method. I think sort is the way to go. Could you extract the sorting logic into a separate method. Also I see you use the protocol scheme as the preferred transport indicator. Maybe we should rename `preferredTransport` to `preferredProtocolScheme` to better indicate what the value of `preferredTransport` is?

TimoGlastra:
```suggestion
    if (options?.preferredTransport) {
```

packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -567,6 +580,10 @@ export class ConnectionService {
       return connection.id === connectionId && connection.state === ConnectionState.Complete
     }
 
+    // Check if already done
+    const connection = await this.connectionRepository.findById(connectionId)
+    if (connection && isConnected(connection)) return connection //TODO: check if this leaves trailing listeners behind?
```
TimoGlastra, line 585: I think this could introduce a realllllly small gap where we retrieve the record, and before the listener is started, the connection state is updated to done.

By starting the listener before doing findById, this could never happen. Is there a specific reason you changed the order?

TimoGlastra: I think we could clean this up a bit in the future, but for now I think it works fine

packages/core/src/modules/routing/RecipientModule.ts
```diff
@@ -178,6 +183,66 @@ export class RecipientModule {
     return event.payload.mediationRecord
   }
 
+  public async provision(mediatorConnInvite: string) {
+    this.logger.debug('Provision Mediation with invitation', { invite: mediatorConnInvite })
+    if (!mediatorConnInvite) {
+      throw new AriesFrameworkError('Mediation provision requires an invitation to be passed')
+    }
```
TimoGlastra, line 190: Theoretically impossible with `mediatorConnInvite: string`

packages/core/src/modules/routing/RecipientModule.ts
```diff
@@ -178,6 +183,66 @@ export class RecipientModule {
...
+    // Connect to mediator through provided invitation if provided in config
+    // Also requests mediation ans sets as default mediator
+    // Because this requires the connections module, we do this in the agent constructor
```
TimoGlastra, line 193: Outdated comment

packages/core/src/modules/routing/RecipientModule.ts
```diff
@@ -178,6 +183,66 @@ export class RecipientModule {
...
+
+    // Assumption: processInvitation is a URL-encoded invitation
+    const invitation = await ConnectionInvitationMessage.fromUrl(mediatorConnInvite)
+    // Check if invitation has been used already
+    if (!invitation || !invitation.recipientKeys || !invitation.recipientKeys[0]) {
+      throw new Error(`Invalid mediation invitation`)
```
TimoGlastra, line 199:
```suggestion
      throw new AriesFrameworkError(`Invalid mediation invitation. Invitation must have at least one recipient key.`)
```

packages/core/src/modules/routing/RecipientModule.ts
```diff
@@ -178,6 +183,66 @@ export class RecipientModule {
...
+    }
+    const connection = await this.connectionService.findByInvitationKey(invitation.recipientKeys[0])
+    if (!connection) {
+      this.logger.debug('Mediation Connection does not exist, creating connection')
+      const routing = await this.mediationRecipientService.getRouting()
+
+      const invitationConnectionRecord = await this.connectionService.processInvitation(invitation, {
+        autoAcceptConnection: true,
+        routing: routing,
```
TimoGlastra, line 208:
```suggestion
        routing,
```

packages/core/src/modules/routing/RecipientModule.ts
```diff
@@ -178,6 +183,66 @@ export class RecipientModule {
...
+      })
+      this.logger.debug('Processed invitation')
+      const { message, connectionRecord: connectionRecord } = await this.connectionService.createRequest(
```
TimoGlastra, line 211:
```suggestion
      const { message, connectionRecord } = await this.connectionService.createRequest(
```

packages/core/src/modules/routing/RecipientModule.ts
```diff
@@ -178,6 +183,66 @@ export class RecipientModule {
...
+    const connection = await this.connectionService.findByInvitationKey(invitation.recipientKeys[0])
+    if (!connection) {
+      this.logger.debug('Mediation Connection does not exist, creating connection')
+      const routing = await this.mediationRecipientService.getRouting()
+
+      const invitationConnectionRecord = await this.connectionService.processInvitation(invitation, {
+        autoAcceptConnection: true,
+        routing: routing,
+      })
+      this.logger.debug('Processed invitation')
```
TimoGlastra, line 210:
```suggestion
      this.logger.debug('Processed mediation invitation', {
      	connectionId: invitationConnectionRecord
      })
```

packages/core/src/modules/routing/RecipientModule.ts
```diff
@@ -178,6 +183,66 @@ export class RecipientModule {
...
+      const { message, connectionRecord: connectionRecord } = await this.connectionService.createRequest(
+        invitationConnectionRecord.id
+      )
+      const outbound = createOutboundMessage(connectionRecord, message)
```
TimoGlastra, line 214: We should find a way around this as we're exactly duplicating the connection module code here (not for this PR, just something to think about)

packages/core/src/modules/routing/RecipientModule.ts
```diff
@@ -178,6 +183,66 @@ export class RecipientModule {
...
+      // TODO: add timeout to returnWhenIsConnected
+      const completedConnectionRecord = await this.connectionService.returnWhenIsConnected(connectionRecord.id)
+      this.logger.debug('Connection completed, requesting mediation')
+      const mediationRecord = await this.requestAndAwaitGrant(completedConnectionRecord, 60000) // TODO: put timeout as a config parameter
+      this.logger.debug('Mediation Granted, setting as default mediator')
+      await this.setDefaultMediator(mediationRecord)
+      this.logger.debug('Default mediator set')
+      return
+    } else if (connection && !connection.isReady) {
+      const connectionRecord = await this.connectionService.returnWhenIsConnected(connection.id)
```
TimoGlastra, line 226: This will probably never resolve. If we need to wait for an action from the other party we're fine. However if we shutdown the agent mid connection and never sent the message, we must manually trigger the next step in the protocol. Not sure what the best approach is here.

TimoGlastra (commented):
I left another round of comments. I think we're almost there. Lots of great improvements! :)

packages/core/src/agent/MessageSender.ts
```diff
@@ -250,7 +252,55 @@ export class MessageSender {
...
+  private async retrieveServicesByConnection(
+    connection: ConnectionRecord,
+    transportPriority?: TransportPriorityOptions
+  ) {
+    this.logger.debug(
+      `Retrieving services for connection '${connection.id}'${' (' && connection.theirLabel && ')'}${
+        (transportPriority?.restrictive && ' restrictively ',
+        transportPriority && ' by priority of the follow schemes: ' + transportPriority.schemes)
+      }`
+    )
```
TimoGlastra, line 273: There is a lot happening in this debug statement, I'd prefer to pass it as extra argument to the second parameter

```suggestion
    this.logger.debug(
      `Retrieving services for connection '${connection.id}' (${connection.theirLabel})`,
      {
      	transportPriority
      }
    )
```

packages/core/src/agent/MessageSender.ts
```diff
@@ -250,7 +252,55 @@ export class MessageSender {
...
+      services.sort(function (a, b) {
+        const aScheme = a.protocolScheme
+        const bScheme = b.protocolScheme
+        return transportPriority?.schemes.indexOf(aScheme) - transportPriority?.schemes.indexOf(bScheme)
+      })
+    }
+
+    this.logger.debug(
+      `Retrieved ${services.length} services for message to connection '${connection.id}'${
+        ' (' && connection.theirLabel && ')'
```
TimoGlastra, line 300: I think this will still just render the value of theirLabel so `(undefined)`. So we can just do this I think:
```
'(${connection.theirLabel})'
```

TimoGlastra (approved):
Still some improvements to be made, but I think it is good enough to merge now. 

One thing that would be a good addition is having an e2e test that creates a connection with mediator using mediator connections invite, shuts down the agent and then restarts the agent and creates a new connection with another agent. But this can also be added in a separate PR.

TimoGlastra:
@jakubkoci @JamesKEbert @burdettadam Can I go ahead and merge this PR?

packages/core/src/modules/routing/RecipientModule.ts
JamesKEbert: Do we want to be able to import the module into a module? I don't think right now the modules are set up to be used via dependency injection IIRC.

JamesKEbert (approved):
The changes look good to me--we aren't actually making use of the priority options for message sending from the RecipientModule for implicit mediation, and I've also been working on a revamp for the message sender. However, I'd prefer to merge this (in the better state it's in), and do followup PR / work, especially since it fixes the agent mediation restart issues. I think this is a big topic (especially around transports), so I think there's more to do, but I'd rather merge this now.

packages/core/src/transport/WsOutboundTransporter.ts
```diff
@@ -109,9 +112,29 @@ export class WsOutboundTransporter implements OutboundTransporter {
...
+            }
+            setTimeout(
+              () => {
+                this.agent.connections.acceptResponse(socketId)
+              },
+              interval < 1000 ? interval : 1000
+            )
+            this.recursiveBackOff[socketId] = interval * 2
+          }
+        }
```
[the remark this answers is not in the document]
jakubkoci: Shouldn't be it `setInterval` instead of `setTimout` then? But it's highly probable I didn't understand the logic here 😄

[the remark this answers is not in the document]
jakubkoci: Do I understand it correctly that `socketId` == `connectionId` here?

jakubkoci (approved):
I'm good with merging.

packages/core/src/modules/routing/RecipientModule.ts
TimoGlastra: No as we've set it up, modules don't use modules. But it has introduced some limitations...

Issue #301, opened 2021-06-02
ISO Date Parsing Failure in React Native


JamesKEbert:
I uncovered an issue just a little bit ago with ISO date parsing in React Native. When receiving basic messages, we perform `new Date(sentTime).toISOString()`. Unfortunately the date parsing involved here was not universally standardized effectively in JS, and so therefore this operation is incompatible in React Native with timestamps in ISO 8601 UTC ("2019-01-15 18:42:01Z"), unless it contains a `t` instead of a space.

See this stackoverflow for additional context: https://stackoverflow.com/questions/13363673/javascript-date-is-invalid-on-ios/13363791

I am not initially sure the best approach to resolving this issue, but I can dig in more. I would prefer to not introduce an entirely separate package to accomplish just this, and wonder if we can do some kind of polyfill or compilation change.

TimoGlastra:
```ts
"2019-01-15 18:42:01Z".replace(" ", "T") // :-)
```

No we should see to properly fix this. Maybe a polyfill that we only use on React Native? As that seems to be the evil-doer

PR #426, opened 2021-08-17
fix: date parsing

seajensen:
# Summary of Changes

Added transform function for parsing dates from multiple formats. Added transform function to basic messaging.

# Related Issues

N/A

# Pull Request Checklist

_PR template adapted from the Python attrs project._

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -30,6 +30,7 @@
     "class-transformer": "^0.4.0",
     "class-validator": "^0.13.1",
     "js-sha256": "^0.9.0",
+    "luxon": "^1.27.0",
     "make-error": "^1.3.6",
     "multibase": "^4.0.4",
     "multihashes": "^4.0.2",
@@ -42,6 +43,7 @@
   "devDependencies": {
     "@types/bn.js": "^5.1.0",
     "@types/events": "^3.0.0",
+    "@types/luxon": "^1.27.0",
     "@types/node-fetch": "^2.5.10",
     "@types/object-inspect": "^1.8.0",
     "@types/uuid": "^8.3.0",
--- a/packages/core/src/modules/basic-messages/messages/BasicMessage.ts
+++ b/packages/core/src/modules/basic-messages/messages/BasicMessage.ts
@@ -1,7 +1,8 @@
-import { Expose, Type } from 'class-transformer'
+import { Expose, Transform, Type } from 'class-transformer'
 import { Equals, IsDate, IsString } from 'class-validator'
 
 import { AgentMessage } from '../../../agent/AgentMessage'
+import { DateParser } from '../../../utils/transformers'
 
 export class BasicMessage extends AgentMessage {
   /**
@@ -25,7 +26,7 @@ export class BasicMessage extends AgentMessage {
   public static readonly type = 'https://didcomm.org/basicmessage/1.0/message'
 
   @Expose({ name: 'sent_time' })
-  @Type(() => Date)
+  @Transform(({ value }) => DateParser(value))
   @IsDate()
   public sentTime!: Date
 
--- a/packages/core/src/modules/basic-messages/messages/BasicMessage.ts
+++ b/packages/core/src/modules/basic-messages/messages/BasicMessage.ts
@@ -1,4 +1,4 @@
-import { Expose, Transform, Type } from 'class-transformer'
+import { Expose, Transform } from 'class-transformer'
 import { Equals, IsDate, IsString } from 'class-validator'
 
 import { AgentMessage } from '../../../agent/AgentMessage'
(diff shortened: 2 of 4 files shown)


TimoGlastra:
@seajensen What format is not working with the JS date? Basic message uses ISO8601, which I think JS supports

seajensen:
> @seajensen What format is not working with the JS date? Basic message uses ISO8601, which I think JS supports

Apologies, I forgot to link the issue. This is in reference to https://github.com/hyperledger/aries-framework-javascript/issues/301. @JamesKEbert might be able to fill in some additional details.

JamesKEbert:
Yes! I had meant to follow up on the issue. So, from my findings, ISO 8601 dates do not allow spaces in the name, but [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339#section-5.6) that is an extension to ISO 8601 (as I currently understand it) would allow a space. You are correct that we can parse ISO 8601 strings in both React Native and Node (and I assume Electron), however what is being received from ACA-Py isn't strictly ISO 8601, which maybe is something I should raise on that side. 
Regardless, I was unable to find any polyfill for this unfortunately, and date issues are not exactly a rare occurrence. Using Luxon gives additional flexibility and assurance of parsing across different environments. Happy to hear your thoughts though!

blu3beri:
ISO 3339 is indeed an extension of ISO 8601. Both ISO's specify that you are allowed to use another delimiter if any interacting party knows this ( quite odd IMO ). The Basic messaging RFC gives an example for an ISO 8601 date with a space as delimiter between date and time. 

My suggestion is for ACA-Py to change it and change it also in the RFC. However, since both RFCs technically allow it, I am not opposed to this fix. 

We could also write a parser that splits on `space` and joins on `T`, but I do not know enough about the ISO specifications to know if that will never fail.

JamesKEbert (changes requested):

JamesKEbert (approved):
LGTM

TimoGlastra (approved):

TimoGlastra:
> Yes! I had meant to follow up on the issue. So, from my findings, ISO 8601 dates do not allow spaces in the name, but [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339#section-5.6) that is an extension to ISO 8601 (as I currently understand it) would allow a space. You are correct that we can parse ISO 8601 strings in both React Native and Node (and I assume Electron), however what is being received from ACA-Py isn't strictly ISO 8601, which maybe is something I should raise on that side.
> Regardless, I was unable to find any polyfill for this unfortunately, and date issues are not exactly a rare occurrence. Using Luxon gives additional flexibility and assurance of parsing across different environments. Happy to hear your thoughts though!

@JamesKEbert As I understand it from this thread, ACA-Py _does_ send ISO 8601.

https://chat.hyperledger.org/group/os2iqWFM5k2usnrEP?msg=RR5mMYJdxFcYFBLMQ

And after doing a bit more research, it seems that JS `Date` only supports a simplified version of ISO8601. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse#parameters

So ACA-Py is not necessarily the issue here, which makes this PR more an appropriate fix imo. 

Is basic message the only date that should be in ISO8601 format? No other properties/messages?

swcurran:
Not quite, Timo.  ACA-Py does send ISO860, but is not sending timezone because the default call in Python doesn't send it.   So, yes, we are going to fix ACA-Py to include the timezone.  However, all recipients of W3C proofs have to handle cases with and without timezones.

JamesKEbert:
@TimoGlastra it's worth noting @blu3beri's thoughts above:

> ISO 3339 is indeed an extension of ISO 8601. Both ISO's specify that you are allowed to use another delimiter if ALL interacting parties know this ( quite odd IMO ).
> 
> The Basic messaging RFC also gives an example for an ISO 8601 date with a space as delimiter between date and time.
> 
> My suggestion is for ACA-Py to change it and change it also in the RFC. However, since both RFCs technically allow it, I am not opposed to this fix.

I'd recommend the Aries RFC changing to not include the space for additional clarity--however, you are correct that this is technically feasible under ISO 8601 according to Berend's findings, since it's defined in the Aries basic messaging RFC. I am not actually immediately sure on other date fields in Aries (I hadn't gotten to it), but I can look into it.

JamesKEbert:
Ahh I just identified/realized in [RFC 74 DIDComm Best Practices](https://github.com/hyperledger/aries-rfcs/blob/08653f21a489bf4717b54e4d7fd2d0bdfe6b4d1a/concepts/0074-didcomm-best-practices/README.md#_time) that the use of a space instead of "T" is allowed and called out as supported behavior.

PR #428, opened 2021-08-19
feat: add from record method to cred preview

TimoGlastra:
Two api improvements
1. register credential definition doesn't require `signatureType` parameter anymore (just add bloat to the API)
2. add `fromRecord` method to credential preview to be able to create a preview from json object. this makes it a lot simpler:

Before 

```ts
const credentialPreview = new CredentialPreview({
  attributes: [
    new CredentialPreviewAttribute({
      name: 'name',
      mimeType: 'text/plain',
      value: 'John',
    }),
    new CredentialPreviewAttribute({
      name: 'age',
      mimeType: 'text/plain',
      value: '99',
    }),
  ],
})
```

After

```ts
const credentialPreview = CredentialPreview.fromRecord({
  name: 'John',
  age: '99',
})
```

### DIFF
--- a/tests/e2e-test.ts
+++ b/tests/e2e-test.ts
@@ -1,14 +1,9 @@
 import type { Agent } from '@aries-framework/core'
 
-import {
-  issueCredential,
-  makeConnection,
-  prepareForIssuance,
-  presentProof,
-  previewFromAttributes,
-} from '../packages/core/tests/helpers'
+import { issueCredential, makeConnection, prepareForIssuance, presentProof } from '../packages/core/tests/helpers'
 
 import {
+  CredentialPreview,
   AttributeFilter,
   CredentialState,
   MediationState,
@@ -53,7 +48,7 @@ export async function e2eTest({
     issuerConnectionId: senderRecipientConnection.id,
     credentialTemplate: {
       credentialDefinitionId: definition.id,
-      preview: previewFromAttributes({
+      preview: CredentialPreview.fromRecord({
         name: 'John',
         age: '25',
         // year month day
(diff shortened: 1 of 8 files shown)


blu3beri:
Once test passes, I will merge.

tests/e2e-test.ts
```diff
@@ -53,7 +48,7 @@ export async function e2eTest({
     issuerConnectionId: senderRecipientConnection.id,
     credentialTemplate: {
       credentialDefinitionId: definition.id,
-      preview: previewFromAttributes({
+      preview: CredentialPreview.fromRecord({
```
jakubkoci, line 51: I like this improvement. I would consider calling it `fromJSON` to prevent confusion with other framework's records, that's what I would expect as a consumer at least. 

To me, the best API would be just passing JSON objects, but this is definitely step forward 👍

TimoGlastra: The problem with `fromJSON` is that it clashes with `toJSON`.

Currently, `toJSON` transforms the class instance to the JSON representation. If we have a `fromJSON` method, I'd expect it to be the exact opposite of `toJSON`. Which it is not.

The reason I didn't change the constructor is because the `fromRecord` method removes the option to set a mime-type. Most time we just use text/plain, but it should be possible to set it

blu3beri (approved):

tests/e2e-test.ts
TimoGlastra: I'm not set on `fromRecord`, so if you have suggestions I'm happy to create a new PR

PR #433, opened 2021-08-20
feat: add multiple inbound transports

TimoGlastra:
- add support for multiple inbound transports
- rename all occurrences of transporter to transport


packages/node/src/transport/WsInboundTransport.ts
```diff
@@ -20,8 +20,9 @@ export class WsInboundTransport implements InboundTransporter {
 
     this.logger = config.logger
 
-    this.logger.debug(`Starting HTTP inbound transporter`, {
-      endpoint: config.getEndpoint(),
+    const httpEndpoint = config.endpoints.find((e) => e.startsWith('http'))
```
blu3beri, line 23: I am not the most knowledgeable on this subject, but this will also match https. Which might be intended but the log message below is incorrect then.

blu3beri (approved):

burdettadam (approved):

JamesKEbert (approved):
LGTM

jakubkoci (approved):

PR #434, opened 2021-08-21
fix: alter mediation recipient websocket transport priority

JamesKEbert:
Alters the mediationRecipient to properly indicate the priority for using Websockets when doing implicit mediation. 
Updated `@aries-framework/react-native` to include buffer as a peer dependency to allow for processing and sending of messages via websockets in React Native, updated docs to reflect this.

@burdettadam assisted on this as well :)


burdettadam (approved):

packages/react-native/package.json
```diff
@@ -28,19 +28,20 @@
...
-    "typescript": "~4.3.0",
-    "indy-sdk-react-native": "^0.1.13",
+    "react-native-fs": "^2.18.0",
     "react-native-get-random-values": "^1.7.0",
-    "react-native-fs": "^2.18.0"
+    "rimraf": "~3.0.2",
+    "typescript": "~4.3.0"
   },
   "peerDependencies": {
+    "buffer": "^6.0.3",
```
TimoGlastra, line 42: `buffer` is already a dependency of core and we use it in other places. In `WsOutboundTransport` you can do: 

```
import { Buffer } from '../utils/buffer'
```

packages/core/src/modules/routing/RecipientModule.ts
```diff
@@ -173,15 +181,15 @@ export class RecipientModule {
   public async requestMediation(connection: ConnectionRecord): Promise<MediationRecord> {
     const { mediationRecord, message } = await this.mediationRecipientService.createRequest(connection)
     const outboundMessage = createOutboundMessage(connection, message)
-    await this.messageSender.sendMessage(outboundMessage)
+
+    await this.sendMessage(outboundMessage)
```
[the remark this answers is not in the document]
JamesKEbert: I agree, I think that we'll want to adjust this in the future, especially with pickup-protocol v2.

JamesKEbert:
I've fixed a thing or two as requested, as well as fixed the failing tests. I added a try catch in the mediation websocket opener to catch failed attempts at opening the WS (motivated by iOS network errors). 
Let me know if there's anything else needed to change here @TimoGlastra or others.

packages/core/src/modules/routing/RecipientModule.ts
```diff
@@ -85,17 +103,21 @@ export class RecipientModule {
...
+          // TODO: add keepAlive: true to enforce through the public api
+          // we need to keep the socket alive. It already works this way, but would
+          // be good to make more explicit from the public facing API.
+          // This would also make it easier to change the internal API later on.
+          // keepAlive: true,
+        },
+      })
+    } catch (error) {
+      this.logger.warn('Unable to open websocket connection to mediator', { error })
+    }
```
TimoGlastra, line 120: I'm not sure we want to swallow this error?

TimoGlastra (approved):
One comment, but overall LGTM

packages/core/src/modules/routing/RecipientModule.ts
JamesKEbert: The issue that I ran into is that it will throw an error if you don't have internet access, etc. Which means that if you throw it, it won't be caught and will not be able to reconnect properly. As far as I saw at least. Happy to go about this a different way in the future if desired.

PR #444, opened 2021-08-30
fix: legacy did:sov prefix on invitation

TimoGlastra:
Addition to #442 where we forgot to also transform the invitation `@type`. 

This is required to interop with agents that don't support the https://didcomm.org message type prefix yet (e.g. Trinsic wallet)

### DIFF
--- a/packages/core/src/modules/connections/messages/ConnectionInvitationMessage.ts
+++ b/packages/core/src/modules/connections/messages/ConnectionInvitationMessage.ts
@@ -5,7 +5,7 @@ import { AgentMessage } from '../../../agent/AgentMessage'
 import { AriesFrameworkError } from '../../../error'
 import { JsonEncoder } from '../../../utils/JsonEncoder'
 import { JsonTransformer } from '../../../utils/JsonTransformer'
-import { replaceLegacyDidSovPrefix } from '../../../utils/messageType'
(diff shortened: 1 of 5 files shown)


packages/core/src/modules/connections/messages/ConnectionInvitationMessage.ts
```diff
@@ -91,8 +91,19 @@ export class ConnectionInvitationMessage extends AgentMessage {
    * @param domain domain name to use for invitation url
    * @returns invitation url with base64 encoded invitation
    */
-  public toUrl(domain = 'https://example.com/ssi') {
+  public toUrl({
+    domain = 'https://example.com/ssi',
+    useLegacyDidSovPrefix = false,
+  }: {
+    domain: string
```
blu3beri, line 98: Not a 100% sure, but with a default parameter value, should it not be `domain?: string`?

TimoGlastra: Yes. My intention was to remove the default parameter as 'https://example.com/ssi' seems weird. But obviously forgot to remove it....

TimoGlastra: Updated

blu3beri (approved):

PR #452, opened 2021-09-02
fix(core): handle undefined requested attributes and requested predicates

blu3beri:
I had some issues with accepting a proof request that did not contain requested attributes (but only predicates or self attested attributes).

This should resolve this issue.

### DIFF
--- a/packages/core/src/modules/proofs/services/ProofService.ts
+++ b/packages/core/src/modules/proofs/services/ProofService.ts
@@ -911,7 +911,13 @@ export class ProofService {
       ...Object.values(requestedCredentials.requestedAttributes),
       ...Object.values(requestedCredentials.requestedPredicates),
     ]
-      .filter((c) => c !== undefined && c !== null)
+      .filter((c) => {
+        const isDefined = c !== undefined && c !== null
+        if (!isDefined) {
+          this.logger.warn('Received an undefined attribute or predicate')
+        }
+        return isDefined
+      })
       .map((c) => c.credentialInfo)
 
     const schemas = await this.getSchemas(new Set(credentialObjects.map((c) => c.schemaId)))
--- a/packages/core/src/modules/proofs/services/ProofService.ts
+++ b/packages/core/src/modules/proofs/services/ProofService.ts
@@ -659,9 +659,11 @@ export class ProofService {
       requestedAttributesNames.push(...(requestedAttributes.names ?? [requestedAttributes.name]))
 
       // Find the attributes that have a hashlink as a value
-      for (const attribute of Object.values(requestedAttribute.credentialInfo.attributes)) {
-        if (attribute.toLowerCase().startsWith('hl:')) {
-          credentialIds.add(requestedAttribute.credentialId)
+      if (requestedAttribute) {
+        for (const attribute of Object.values(requestedAttribute.credentialInfo.attributes)) {
+          if (attribute.toLowerCase().startsWith('hl:')) {
+            credentialIds.add(requestedAttribute.credentialId)
+          }
         }
       }
     }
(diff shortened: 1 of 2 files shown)


packages/core/src/modules/proofs/services/ProofService.ts
```diff
@@ -908,7 +910,9 @@ export class ProofService {
     const credentialObjects = [
       ...Object.values(requestedCredentials.requestedAttributes),
       ...Object.values(requestedCredentials.requestedPredicates),
-    ].map((c) => c.credentialInfo)
+    ]
+      .filter(Boolean)
```
jakubkoci, line 914: It's a bit harder to understand what we want to achieve here with the `Boolean`. Does it check whether the value is not empty?

blu3beri: Yes. It uses the boolean constructor to check for truthy/falsy

```js
[].filter((c) => c !== undefined)
```
Would be more verbose though.

jakubkoci: I see. You're right that it's more verbose. But it better expresses the intent.

I think the `null` could also be the case, right? Then it should be
```
[].filter((c) => c !== undefined && c !== null)
```

Then I wouldn't be afraid of creating new function `isNil` or `isNotNil` and extract this logic.

TimoGlastra:
@blu3beri Could you add some tests for this before we merge?

blu3beri:
> @blu3beri Could you add some tests for this before we merge?

Yes, will do!

JamesKEbert:
Are we properly throwing an error or warning when these fields are not provided? If not that's a potential error that'll come up in the future with other agents than ACA-Py that may not be handled very well. I'd rather us throw an error or warning, to better indicate it's the other party's fault. 
I also think that better presentation tests would be good, which you've opened an issue for already: https://github.com/hyperledger/aries-framework-javascript/issues/481

blu3beri:
> Are we properly throwing an error or warning when these fields are not provided? If not that's a potential error that'll come up in the future with other agents than ACA-Py that may not be handled very well. I'd rather us throw an error or warning, to better indicate it's the other party's fault. 

ATM we are not handling this case. This pr "handles'" it, e.g. ignoring undefined/null cases. 

But then a new question comes to mind, do we want to throw an error/warning when this happens or just ignore the undefined/null cases (as a holder)? 

The flow can continue as expected without the undefined/null cases so maybe just ignoring them. 

In AFJ when creating a proof request it is possible to not pass `requestedAttributes`, but it will fallback to an empty object. So AFJ as a verifier does not need any error handling.

JamesKEbert:
I would lean towards logging a warning **and** then continuing with the flow as expected. That way we highlight to the user that there is a problem, even if it's not completely stopping everything?

blu3beri:
> I would lean towards logging a warning **and** then continuing with the flow as expected. That way we highlight to the user that there is a problem, even if it's not completely stopping everything?

Sounds good to me, I will add it.

packages/core/src/modules/proofs/services/ProofService.ts
```diff
@@ -908,7 +910,15 @@ export class ProofService {
     const credentialObjects = [
       ...Object.values(requestedCredentials.requestedAttributes),
       ...Object.values(requestedCredentials.requestedPredicates),
-    ].map((c) => c.credentialInfo)
+    ]
+      .filter((c) => {
+        const isDefined = c !== undefined && c !== null
+        if (!isDefined) {
```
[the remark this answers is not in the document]
JamesKEbert: @blu3beri ^ ?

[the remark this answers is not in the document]
blu3beri: Wow I completely forgot about this PR. Will resolve ASAP.

blu3beri:
After some discussion with @TimoGlastra we concluded that this will be fixed by the new proof validation in PR #525. If this issue, for some reason, occurs again, I will pick it up.

PR #456, opened 2021-09-07
feat(core): d_m invitation parameter and invitation image

blu3beri:
- Adds support for the `d_m` invitation parameter in the url
- Improved url parsing (with polyfill for React Native)
- Able to add an image to the invitation

### DIFF
--- a/packages/core/src/modules/connections/__tests__/ConnectionInvitationMessage.test.ts
+++ b/packages/core/src/modules/connections/__tests__/ConnectionInvitationMessage.test.ts
@@ -103,10 +103,10 @@ describe('ConnectionInvitationMessage', () => {
       await expect(validateOrReject(invitation)).resolves.toBeUndefined()
     })
 
-    it('should throw error if url does not contain `c_i` or `d_m`', () => {
+    it('should throw error if url does not contain `c_i` or `d_m`', async () => {
       const invitationUrl = 'https://example.com?param=123'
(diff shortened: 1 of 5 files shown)


genaris:
Nice to have these invitation improvements, and actually I find very useful to have that `imageUrl` field. Is it part of any Aries RFC or currently a Trinsic-specific feature?  

I'm asking this because maybe there are other implementations that use other metadata to make invitations 'user-friendlier' and other fields would be needed to use the framework with them.

blu3beri:
> Is it part of any Aries RFC or currently a Trinsic-specific feature? 

I found it from decoding the trinsic invitation, but I can take a look if it is specified in any RFC. I will take a look tomorrow and get back to you!

blu3beri:
@genaris I have checked around in the rfcs, but I could not find any other metadata for the invitation. If you have other data we can always discuss that during a working group call. However, we should really keep the invitation as small as possible. A large invitation means a large QRcode which is always annoying :-)

TimoGlastra:
This is not native to the Aries RFCs but more an non-standard extension that a lot of wallets use (e.g. Lissi, Connect.Me, Trinsic)

TimoGlastra (approved):
Nice!

I think we may get away with using a smaller lib instead of the URL polyfill (e.g. `query-string` package). But this looks fine to me for now

PR #458, opened 2021-09-18
Feature/offline message retrieval

JamesKEbert:
Adds access to sending a trust ping via the connectionsModule and allows passing of trust ping message arguments (to be able to set response requested, for instance).
Fixes mediation websocket reconnect exponential backoff incrementing behavior. 
Fixes mediation to allow queued messages to be picked up from an ACA-Py Mediator. ACA-Py as a mediator only will deliver messages from the undelivered queue if a message does not trigger a response, such as a trust ping with response requested: false.


packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -295,12 +296,16 @@ export class ConnectionService {
...
   /**
-   * Create a trust ping message for the connection with the specified connection id.
+   * Create an ack using a trust ping message for the connection with the specified connection id.
    *
    * @param connectionId the id of the connection for which to create a trust ping message
+   * @param options optional trust ping options
    * @returns outbound message containing trust ping message
    */
-  public async createTrustPing(connectionId: string): Promise<ConnectionProtocolMsgReturnType<TrustPingMessage>> {
+  public async createAck(
```
TimoGlastra, line 305: What is the difference between createAck and createTrustPing?

TimoGlastra (commented):
Overall LGTM. I'm not so sure on the duplication of the createTrustPing and createAck and would rather call createTrustPing in createAck.

This also better describes what is happening, as createAck gives me the impression we're sending an Ack

packages/core/src/modules/connections/services/ConnectionService.ts
JamesKEbert: No that's completely fair! I've had my fair share of pedantic things--and I see your point too  :) 
I'll go ahead and update this to reflect your thought process here

TimoGlastra:
@JamesKEbert bump :)

TimoGlastra:
@JamesKEbert the required fix for mediation with ACA-Py has already been made in another pr recently. I'm going ahead with the 0.1.0 release for now and we can make a follow up release with these once merged

https://github.com/hyperledger/aries-framework-javascript/blob/4d7d3c1502d5eafa2b884a4a84934e072fe70ea6/packages/core/src/modules/routing/RecipientModule.ts#L95-L97

JamesKEbert:
Closing as this is superseded by #631, as this PR is quite out of date with main & most changes were agreed to not be required.

PR #463, opened 2021-09-21
feat(core): support image url in invitations

TimoGlastra:
non-standard approach for showing images with connections. Almost all wallets and implementations support this (ACA-Py, AF.NET, Connect.Me, Lissi, Trinsic, etc...)

### DIFF
--- a/packages/core/src/agent/AgentConfig.ts
+++ b/packages/core/src/agent/AgentConfig.ts
@@ -115,4 +115,8 @@ export class AgentConfig {
   public get useLegacyDidSovPrefix() {
     return this.initConfig.useLegacyDidSovPrefix ?? false
   }
+
+  public get connectionImageUrl() {
+    return this.initConfig.connectionImageUrl
+  }
 }
--- a/packages/core/src/modules/connections/__tests__/ConnectionService.test.ts
+++ b/packages/core/src/modules/connections/__tests__/ConnectionService.test.ts
@@ -25,9 +25,12 @@ import { ConnectionService } from '../services/ConnectionService'
 jest.mock('../repository/ConnectionRepository')
 const ConnectionRepositoryMock = ConnectionRepository as jest.Mock<ConnectionRepository>
 
+const connectionImageUrl = 'https://example.com/image.png'
+
 describe('ConnectionService', () => {
   const config = getAgentConfig('ConnectionServiceTest', {
     endpoints: ['http://agent.com:8080'],
+    connectionImageUrl,
   })
 
   let wallet: Wallet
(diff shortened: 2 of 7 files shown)


TimoGlastra:
@janrtvld ^^

TimoGlastra:
I understand the concern. There was a brief discussion about this at the Aries WG. It's not in the protocol I think because of privacy reasons (so that should maybe also be a reason to not include it in the mobile wallet)

Re #456, d_m is the oob approach but it seems Trinsic uses this for their invitations. I'd vote for having interop at the moment and keep good track of the non-standard features we introduce. But maybe we should have a discussion about it at the WG call.

E.g. if you create an invitation in trinsic studio it shows a shortened url which resolved to: https://trinsic.studio/link/?d_m=eyJsYWJlbCxxxxx

JamesKEbert:
Thanks for the feedback!

> It's not in the protocol I think because of privacy reasons (so that should maybe also be a reason to not include it in the mobile wallet)
I must have missed that discussion, but I'd be interested in exploring what we can do to either include in the protocol, or identify a better methodology for doing it given that everyone is doing it but not having it documented in the protocol. 

> Re #456, d_m is the oob approach but it seems Trinsic uses this for their invitations.
Perhaps I'm missing something but shouldn't the query be for OOB: `?oob=`? I think we need to add URL shortening checking, but after Trinsic's is resolved I don't think it's compliant with OOB or connections v1? 

> I'd vote for having interop at the moment and keep good track of the non-standard features we introduce.
I am okay for including for now, but I think it'd be wise to add a MD file or list in the repo of deviations made, IMO. Happy to talk about it more in the WG call.

TimoGlastra:
> Perhaps I'm missing something but shouldn't the query be for OOB: `?oob=`? I think we need to add URL shortening checking, but after Trinsic's is resolved I don't think it's compliant with OOB or connections v1?

No you're right, I think it is often used for connnectionless exchanges. So if I want to issue you a credential without connection, pre-oob d_m (from didcomm_message) is used. Just found a reference to it in RFC 0268: https://github.com/hyperledger/aries-rfcs/blob/08653f21a489bf4717b54e4d7fd2d0bdfe6b4d1a/concepts/0268-unified-didcomm-agent-deeplinking/README.md#message-requirements

TimoGlastra:
> I must have missed that discussion, but I'd be interested in exploring what we can do to either include in the protocol, or identify a better methodology for doing it given that everyone is doing it but not having it documented in the protocol.

Let's see if we can discuss it at the next Aries WG

JamesKEbert:
> So if I want to issue you a credential without connection, pre-oob d_m (from didcomm_message) is used. Just found a reference to it in RFC 0268

Ah yeah--maybe followup dumb question, but given that it's in a state of proposed this isn't something broadly accepted yet and would warrant consideration before adding to the framework? But it does make me a little bit more comfortable with the previous PR #456 though. Hopefully we'll be able to better align around OOB going forward.


So yeah, I'm not opposed here then, but I think a followup discussion in the Aries WG call would be good. Thanks for the discussion!

TelegramSam:
Will discuss in Aries call.

The primary problem with this is the ease of impersonation and phishing. I believe the desire to include more information in an invitation is the result of a bad mental model (likely contributed to by me) that a user must approve the creation of a connection and exchange of verified information.

PR #487, opened 2021-10-13
docs: improved electron setup 

blu3beri:
- Improved the setup guide for electron
- Added some information to the macos part of the libindy setup 
  - M1 still needs some work...

### DIFF
--- a/docs/libindy/macos.md
+++ b/docs/libindy/macos.md
@@ -1,23 +1,14 @@
 # Setup Libindy for MacOS
 
-> NOTE: If these steps do not work, or you want to do a manual install, refer to [here](https://github.com/hyperledger/indy-sdk#ios)
+> NOTE: If these steps do not work, or you want to do a manual install, refer to [here](https://github.com/hyperledger/indy-sdk#macos)
 
-To see if Libindy is already installed, execute the following command:
+To see if Libindy is correctly installed for javascript usage, execute the following command:
 
-> A better test would be to clone the [indy-sdk](https://github.com/hyperledger/indy-sdk/) and run:
->
-> ```
-> cd indy-cli
-> cargo build --release
-> cd ./target/release/
-> ./indy-cli
-> wallet create 123 key
-> ```
+```ts
+npx -p @aries-framework/node is-indy-installed
 
-```bash
-ls /usr/local/lib/libindy.dylib
-# ✅ /usr/local/lib/libindy.dylib
-# ❌ ls: /usr/local/lib/libindy.dylib : No such file or directory
+# output
+# Libindy was installed correctly
 ```
 
 ## MacOS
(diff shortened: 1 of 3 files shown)


docs/setup-electron.md
```diff
@@ -2,84 +2,153 @@
...
-function wrapIndyCallWithErrorHandling(func: any) {
+const wrapIndyCallWithErrorHandling = (func: any) => {
   return async (...args: any[]) => {
     try {
       return await func(...args)
     } catch (e) {
-      e.name = 'IndyError'
-      e.indyName = e.message
-      throw e
+      if (e instanceof Error || e instanceof AriesFrameworkError || e instanceof IndySdkError) {
```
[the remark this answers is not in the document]
JamesKEbert: Definitely interested in this alteration as well! I really need to take a look at the associated Toolbox PR still

docs/libindy/macos.md
```diff
@@ -4,6 +4,16 @@
...
 
+> A better test would be to clone the [indy-sdk](https://github.com/hyperledger/indy-sdk/) and run:
+>
+> ```
+> cd indy-cli
+> cargo build --release
+> cd ./target/release/
+> ./indy-cli
+> wallet create 123 key
+> ```
```
JamesKEbert, line 15: Should we just replace the old method then?

JamesKEbert (commented):
Great work @blu3beri :)

docs/libindy/macos.md
blu3beri: This method would require cloning the indy-sdk repository and building via Cargo. The other check might be a bit less accurate, but it only requires the `ls` command. 

The issue with this test is that libindy can be installed correctly for the indy cli, but not for node. 

I will see if there is a better way, e.g. clone a small repo that just runs `indy.createWallet({id:'12'},{key:'abc'});` to test.

TimoGlastra: A thought. What if we add a binary to the `@aries-framework/node` package. This way you can just use `npx` to check if the installation was succeful

With [this PR](https://github.com/hyperledger/aries-framework-javascript/pull/510) merged we will be able to (@blu3beri I copied your example from https://github.com/animo/test-libindy-installation-for-node)

```
npx -p @aries-framework/node is-indy-installed

# output
# Libindy was installed correctly
```

What do you think?

blu3beri: That's awesome! Way better than what I had in mind.

TimoGlastra (approved):
LGTM @blu3beri. I'm going to merge this now as it already is a huge improvement on the current docs. We can then fix the last point with the indy error in a follow up PR :)

JamesKEbert:
Good work here @blu3beri, thank you ❤️

PR #504, opened 2021-10-26
fix: export indy pool config

TimoGlastra:
it is used in public config so should be exported


JamesKEbert (approved):
LGTM

PR #505, opened 2021-10-26
feat(core)!: metadata on records

blu3beri:
BREAKING CHANGE: `credentialRecord.credentialMetadata` has been replaced by `credentialRecord.metadata`.

- Adds metadata to the baserecord
- Does not do any persistency and should be done manually after updating the metadata no the record.
- Can be used by framework developers and consumers.

Maintainers: Before merging and squashing make sure to copy the `BREAKING CHANGE` entry to the commit message field, this is needed for automatic tracking of breaking changes


packages/core/src/storage/Metadata.ts
```diff
@@ -0,0 +1,79 @@
...
+}
+
+/**
+ * Metadata access class to get, set (create and update) and delete
+ * metadata on any record.
+ *
+ * set will override the previous value if it already exists
+ *
+ * note: To add in-memory persistence to these records, you have to
+ * update the record in the correct repository
```
JamesKEbert, line 12: Dumb confirmation here.. So you essentially _do need_ to store the altered record in the associated record repository.. but that shouldn't require changing any types associated with that record, right?

packages/core/src/modules/credentials/repository/CredentialRecord.ts
```diff
@@ -97,6 +95,22 @@ export class CredentialRecord extends BaseRecord<DefaultCredentialTags, CustomCr
...
+          this.metadata.add('indyCredentialMetadata', {
+            schemaId: props.metadata.schemaId,
+          })
+        }
+        if (props.metadata.credentialDefinitionId) {
+          this.metadata.add('indyCredentialMetadata', {
+            credentialDefinitionId: props.metadata.credentialDefinitionId,
+          })
+        }
+      }
```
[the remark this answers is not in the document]
JamesKEbert: It may be wise to increment our version in some manner to indicate a breaking change here, since this could cause fairly significant issues for existing systems? Is there harm in utilizing @blu3beri's original mechanism until a later date?

packages/core/src/storage/Metadata.ts
blu3beri: Yes! 

'''ts
connectionRecordExample.metadata.set('test', { foo : 'bar' } );

connectionRepository.update(connectionRecordExample);
'''

I was unable to automatically store the record, so I think that this is the best solution.

packages/core/src/modules/credentials/repository/CredentialRecord.ts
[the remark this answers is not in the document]
TimoGlastra: I think @blu3beri's approach wouldn't work out of the box. The constructor is not called with properties when a record is retrieved from storage. We could add a custom transformer, but I'd rather only take these things into account between stable releases.

Maybe we can look at writing transform scripts. I think we should be very cautious with adding these if statements to the codebase. I don't want to be stuck with this piece of code for eternity. We can never assume when all records are transformed. We're going to have a lot more breaking changes going forward that I'll rather solve by writing update scripts. I'd day that is a reasonable for pre-1.0 releases

Maybe we can keep it in here till 0.1.0. Then for 0.1.0 we provide scripts, and for 0.2.0 we again provide a script for the transformation.

Thoughts?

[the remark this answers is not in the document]
blu3beri: Maybe I am misunderstanding your point, but right now we set this data in the constructor already and the `credentialService`, line 612 and 627, relay on these fields. It will even error out when we do not have these fields in the metadata. 

I am okay with leaving this out, but how do you propose we handle that error case?

packages/core/src/modules/credentials/services/CredentialService.ts
```diff
@@ -604,7 +609,9 @@ export class CredentialService {
       previousSentMessage: credentialRecord.requestMessage,
     })
 
-    if (!credentialRecord.metadata.requestMetadata) {
+    const credentialRequestMetadata = credentialRecord.metadata.get<CredReqMetadata>('requestMetadata')
```
TimoGlastra, line 612:
```suggestion
    const credentialRequestMetadata = credentialRecord.metadata.get<CredReqMetadata>('indyRequestMetadata')
```

packages/core/src/storage/BaseRecord.ts
```diff
@@ -1,16 +1,33 @@
...
 export type TagsBase = {
   [key: string]: TagValue
   [key: number]: never
 }
 
 export type Tags<DefaultTags extends TagsBase, CustomTags extends TagsBase> = CustomTags & DefaultTags
+
 export type RecordTags<Record extends BaseRecord> = ReturnType<Record['getTags']>
 
+export function MetadataTransformer() {
```
TimoGlastra, line 17: Maybe we can host this in the file that contains the metadata class?

packages/core/src/storage/Metadata.ts
```diff
@@ -0,0 +1,79 @@
...
+  [key: string]: Record<string, unknown>
+}
+
+/**
+ * Metadata access class to get, set (create and update) and delete
+ * metadata on any record.
+ *
+ * set will override the previous value if it already exists
+ *
+ * note: To add in-memory persistence to these records, you have to
```
TimoGlastra, line 11:
```suggestion
 * note: To persist the metadata on these records, you have to
```

packages/core/src/modules/credentials/repository/CredentialRecord.ts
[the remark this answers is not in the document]
TimoGlastra: Ah okay we were talking about two different things here. For your point @blu3beri, I suggest we remove the metadata from the constructor and require it to be set from using the `record.metadata.set` afterwards.

What I was mainly thinking about here is the persistence. The metadata on the current credential records has a different structure than the metadata we're adding in this PR. My thinking was that we add something like this migration script called `0.1.0-alpha.291-0.1.0-alpha.292.tsx


```ts
export async function migrate291To292(agent: Agent) {
  const credentialRecords = await agent.credentials.getAll();
  const credentialRepository = await agent.injectionContainer.resolve(
    CredentialRepository
  );

  for (const credentialRecord of credentialRecords) {
    const data = credentialRecord.metadata.data;
    const metadata = MetaData();

    if (data.requestMetadata) {
      metadata.set("indyRequestMetadata", data.requestMetadata);
    }
    if (data.schemaId) {
      this.metadata.add("indyCredentialMetadata", {
        schemaId: data.schemaId,
      });
    }
    if (data.credentialDefinitionId) {
      this.metadata.add("indyCredentialMetadata", {
        credentialDefinitionId: data.credentialDefinitionId,
      });
    }

    credentialRecord.metadata = metadata;
    await credentialRepository.update(credentialRecord);
  }
}
```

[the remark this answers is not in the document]
blu3beri: Ahh okay. I am in favour of the migration scripts, but for the 0.1.0 and so on. 👍

TimoGlastra (approved):
Nice!

packages/core/src/modules/credentials/repository/CredentialRecord.ts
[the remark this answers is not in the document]
JamesKEbert: Yeah I'd be in favor of migration scripts, as that'd be very nice for breaking changes. We could either tackle that now or at a 0.1.0 release, as you mentioned. :) 
I'll admit I'm not entirely sure how that should be handled from a mobile app perspective, but that can be figured out. :)

JamesKEbert:
So, just to confirm, this would have a breaking change at this point still, right? I have some cases that would need a way to migrate--so this would be a spot that would need a migration script, right?

blu3beri:
> So, just to confirm, this would have a breaking change at this point still, right? I have some cases that would need a way to migrate--so this would be a spot that would need a migration script, right?

Yes, a migration script from all the previous versions to this one is required. @TimoGlastra gave one in the comments above that should, or maybe with some minor tweaking.

TimoGlastra:
Decision made on WG call: Use `@Transformer` for now to transform the record when it is constructed. When moving to 0.1.0 we will add a migration script.

@blu3beri could you make the required changes?

packages/core/src/utils/transformers.ts
```diff
@@ -39,6 +41,29 @@ export function RecordTransformer<T>(Class: { new (...args: any[]): T }) {
...
+    switch (type) {
+      case TransformationType.CLASS_TO_PLAIN:
+        return { ...value.data }
+
+      case TransformationType.PLAIN_TO_CLASS:
+        // TODO: Remove conversion in 0.1.0 via a migration script
+        JsonTransformer.renameKey('requestMetadata', 'indyRequestMetadata', value)
+        JsonTransformer.renameKey('credentialMetadata', 'indyCredentialMetadata', value)
+
+        return new Metadata(value)
```
TimoGlastra, line 60: The previous structure of the metadata object in `CredentialRecord` was as follows:

```typescript
{
    requestMetadata?: Record<string, unknown>
    credentialDefinitionId?: string
    schemaId?: string
}
```

While it now should look something like the following:

```typescript
{
   indyRequestMetadata?: Record<string, unknown>
   indyCredentialMetadata?: {
      schemaId?: string
      credentialDefinitionId?: string
  }
]
```

It is fine to rename `requestMetadata` to `indyRequestMetadata`, but renaming `credentialMetadata` to `indyCredentialMetadata` is not going to do the trick.

I suggest doing something like the following:

```suggestion
      case TransformationType.PLAIN_TO_CLASS:
      
	  if (value.schemaId || value.credentialDefinitionId || value.requestMetadata) {
		  const metadata = new Metadata({})
		  
		  if (value.schemaId) metadata.add('indyCredentialMetadata', { schemaId: value.schemaId })
		  if (value.credentialDefinitionId) metadata.add('indyCredentialMetadata', { schemaId: value.credentialDefinitionId })
		  if (value.requestMetadata) metadata.set('indyRequestMetadata', value.requestMetadata)
	  
		  return metadata
	  }
	  
	return new Metadata(value)
```

TimoGlastra: Also would be good to add a test that checks if the transformation was successful

packages/core/tests/credentials-auto-accept.test.ts
```diff
@@ -278,7 +302,15 @@ describe('auto accept credentials', () => {
...
-        metadata: { requestMetadata: expect.any(Object) },
+        metadata: {
+          data: {
+            indyRequestMetadata: expect.any(Object),
+            indyCredentialMetadata: {
+              schemaId,
+              credentialDefinitionId: credDefId,
+            },
+          },
+        },
```
jakubkoci, line 313: It's a bit confusing that we have `metadata.data.indyRequestMetadata`? Wouldn't be more straightforward to have `metadata.indyRequest`?

jakubkoci:
What if a developer creates some metadata:

```ts
credentialRecord.metadata.set('something', { a: 1 })
credentialRecord.metadata.set('something else', { b: 2 })
```

And then decides to remove all of them:
```ts
Object.keys(credentialRecord.metadata.data).forEach(key => credentialRecord.metadata.delete(key))
```

Is it possible to do so? Would that delete also internal metadata?

blu3beri:
> Is it possible to do so? Would that delete also internal metadata?

Yes that is possible and it would also delete the internal metadata.

packages/core/tests/credentials-auto-accept.test.ts
blu3beri: The data object is there just for structure. My design was that the user could not easily call `metadata.indyRequestMetdata` and do it through the `get`,`set`, etc. Now, the `data` does not remove this, but I do think it promotes the usage of the functions to get values.

renaming `indyRequestMetdata` to `indyRequest` is okay with me. It does seem a bit too verbose now that I think about it.

jakubkoci: I see. I think it would be more concise to use just `indyRequest` instead of `indyRequestMetdata` at least.

jakubkoci:
If I understand correctly, deleting the internal metadata would break the agent, right?

It would be great if we're able to prevent it somehow.

TimoGlastra:
> If I understand correctly, deleting the internal metadata would break the agent, right?
> 
> 
> 
> It would be great if we're able to prevent it somehow.

I'd say we prefix it with `internal/` if you then delete it that would be weird. Nothing prevents the user from overwriting the state of a record and saving that

blu3beri:
> If I understand correctly, deleting the internal metadata would break the agent, right?
> 
> It would be great if we're able to prevent it somehow.

I agree with @TimoGlastra here. It will always be possible for a framework consumer to modify the data in a record.

If we would, for instance, throw an error when we try to delete a key-value via `record.metadata.delete(key)` we could still call the underlying functionality which would not error. e.g. `delete metadata.data[key]`.

I completely understand your point, but next to maybe prefixing it, which I am not even sure we should do, but I don't think we should/can do anything about this.

jakubkoci:
> I agree with @TimoGlastra here. It will always be possible for a framework consumer to modify the data in a record.

That's not so easy to do. And I realized it's therefore not so easy to manipulate with a metadata object.
But if the purpose of `metadata` is (or will be) to offer the users some space where they can store (create, update, delete) their metadata for a credential record, it's dangerous to mix those "custom" metadata and "framework" metadata into the same object.

jakubkoci (approved):
We can merge it, but I still have concerns that mixing custom and framework metadata can cause trouble in the future.

blu3beri:
> We can merge it, but I still have concerns that mixing custom and framework metadata can cause trouble in the future.

If you could create an issue for it I will think about the options that are available.

TimoGlastra:
> We can merge it, but I still have concerns that mixing custom and framework metadata can cause trouble in the future.

@jakubkoci What would your ideal solution be for this? Add two types of metadata to each record (framework and consumer metadata)?

TimoGlastra:
@blu3beri I think it would be good to prefix internal metadata to avoid collisions. We can then say that you are never allowed to use this prefix yourself. I've seen this pattern used in other libraries. Could you make a follow-up PR @blu3beri (rather quickly cause it's breaking again) to add the prefix? I'd say we prefix it with something like `_internal/` (other suggestions also welcome)

blu3beri:
> @blu3beri I think it would be good to prefix internal metadata to avoid collisions. We can then say that you are never allowed to use this prefix yourself. I've seen this pattern used in other libraries. Could you make a follow-up PR @blu3beri (rather quickly cause it's breaking again) to add the prefix? I'd say we prefix it with something like `_internal/` (other suggestions also welcome)

What would you say about an `internal` key? like

```ts
metadata.set('internal', {foo: 'bar'...})
```

This way we group ALL of the internal stuff in one object. (This might be considered a downside, that's why I am asking) :)

blu3beri:
> Yes that could work, but thinking about v2 protocols supporting multiple credential types I think it may be good to e.g. split it metadata per credential type. But that could of course also be nested under the same key. It would however make updating harder (you can't use .add()`)

Okay I 10000% agree with the fact that it is harder to update. That was the main reason I did not propose this solution earlier.  I will create a new pr with the prefix. I think that `_internal` is good.

PR #516, opened 2021-11-02
feat(core): update agent label and imageUrl plus per connection label and imageUrl

janrtvld:
#506 

This pr contains:

- Support for updating agent label
- Support for custom label on connection/invitation

I would also like to add support for a custom imageUrl on connection/label, but before that i would like some feedback on this draft.

### DIFF
--- a/packages/core/src/modules/connections/services/ConnectionService.ts
+++ b/packages/core/src/modules/connections/services/ConnectionService.ts
@@ -70,6 +70,7 @@ export class ConnectionService {
     alias?: string
     multiUseInvitation?: boolean
     myLabel?: string
+    myImageUrl?: string
   }): Promise<ConnectionProtocolMsgReturnType<ConnectionInvitationMessage>> {
     // TODO: public did
     const connectionRecord = await this.createConnection({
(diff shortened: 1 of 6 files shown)


blu3beri:
Nice! Great addition. I was really kissing this feature.

packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -154,14 +155,17 @@ export class ConnectionService {
    * @param connectionId the id of the connection for which to create a connection request
    * @returns outbound message containing connection request
    */
-  public async createRequest(connectionId: string): Promise<ConnectionProtocolMsgReturnType<ConnectionRequestMessage>> {
+  public async createRequest(
+    connectionId: string,
+    myLabel?: string
```
TimoGlastra, line 160: I'd add the myLabel in a config like the other ones. This will make it easier to extend with more properties later on (e.g. the `imageUrl`)

TimoGlastra (commented):
Looking good @janrtvld!

TimoGlastra (approved):
LGTM

PR #527, opened 2021-11-07
fix(core)!: do not request ping res for connection

### LINKED ISSUE #515: Flakiness in tests
Recently the tests have become flaky again and will randomly fail. Last time this was caused by not waiting for some promises to resolve in test execution, it's probably something like that again.

TimoGlastra:
BREAKING CHANGE: a trust ping response will not be requested anymore after completing a connection. This is not required, and also non-standard behaviour. It was also causing some tests to be flaky as response messages were stil being sent after one of the agents had already shut down.

Maintainers: Before merging and squashing make sure to copy the `BREAKING CHANGE` entry to the commit message field, this is needed for automatic tracking of breaking changes

Fixes #515

### DIFF
--- a/packages/core/src/modules/connections/ConnectionsModule.ts
+++ b/packages/core/src/modules/connections/ConnectionsModule.ts
@@ -162,7 +162,9 @@ export class ConnectionsModule {
    * @returns connection record
    */
   public async acceptResponse(connectionId: string): Promise<ConnectionRecord> {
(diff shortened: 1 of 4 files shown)


TimoGlastra:
@JamesKEbert this has some overlap with the changes in #458 but is only adding the responseRequested property to not request a trust ping response after creating a connection. There is a flaky tests that errors often because a trust ping response is sent after the tests has already been finished. This seemed like the easiest (and correct way) to fix it.

blu3beri (approved):

PR #528, opened 2021-11-07
fix(core): log errors if message is undeliverable

TimoGlastra:
The error message being logged just indicated that a message is underliverable to a certain connection. It swallowed the underlying errors. This adds an `errors` property to the error log so that it is clear what the underlying errors are when a message is underliverable.

### DIFF
--- a/packages/core/src/agent/MessageSender.ts
+++ b/packages/core/src/agent/MessageSender.ts
@@ -85,6 +85,8 @@ export class MessageSender {
     packedMessage: WireMessage
     options?: { transportPriority?: TransportPriorityOptions }
   }) {
+    const errors: Error[] = []
+
     // Try to send to already open session
     const session = this.transportService.findSessionByConnectionId(connection.id)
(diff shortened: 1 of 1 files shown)


packages/core/src/agent/MessageSender.ts
```diff
@@ -140,9 +143,10 @@ export class MessageSender {
     // Message is undeliverable
     this.logger.error(`Message is undeliverable to connection ${connection.id} (${connection.theirLabel})`, {
       message: packedMessage,
+      errors,
       connection,
     })
-    throw new AriesFrameworkError(`Message is undeliverable to connection ${connection.id} (${connection.theirLabel})`)
+    throw new AriesFrameworkError(`Message is undeliverable to connection ${connection.id} (${connection.theirLabel})}`)
```
blu3beri, line 149:
```suggestion
    throw new AriesFrameworkError(`Message is undeliverable to connection ${connection.id} (${connection.theirLabel})`)
```

blu3beri (approved):

Issue #344, opened 2021-07-03
Add support for RFC 0434: Out-of-Band


TimoGlastra:
Spec: https://github.com/hyperledger/aries-rfcs/blob/b3a3942ef052039e73cd23d847f42947f8287da2/features/0434-outofband/README.md

jakubkoci:
I wrote down a few notes I would like to share with you here.

## Inbound Message processing rules

1. Unpack if it’s packed message (`JweLikeMessage`)
	- Packed message can by
		- anoncrypt (unpack returns recipient key)
		- authcrypt (unpack returns recipient key and sender key) 
2. Transform `JSON` message to `AgentMessage` instance
3. Validate message
4. Validate if an agent is capable to respond to the message
	- agent has a connection identified by both recipient and sender key, we can’t identify connection just by recipient key because we cannot be sure that the sender is the same, therefore sender key must be the same as `connection.theirKey`
	- message has a `~service` decorator -> deprecated when a new OOB is in use
	- it’s an old OOB message with `did` or service-related attributes
	- it’s a new OOB message and has a `service` attribute (maybe this can or should be validated in OOB module 🤷‍♂️ or `receiveInvitation` from OOB module moves to `MessageReceiver` and keeps just `createInvitation`)
5. What if we don’t have information on how to respond at the moment of receiving a message?
	- Throw an error immediately and stop message processing.
	- Throw an error immediately and store the message for later. Yes but for how long?
	- Dispatch the message and leave the problem to `MessageSender`.  `MessageSender` can store the messages which wasn't able to process for later… Hmm that doesn’t make much sense
6. Dispatch message inside `InboundMessageContext` instance

## Implementation Ideas
I started with an idea to create a separate OOB module which would create and process OOB messages. The problem is that I needed to copy-paste a lot of code from `ConnectionModule` and `MessageReceiver`. There is a similarity with ConnectionModule because we need to create connections and with MessageReceiver because we need to check if we have a connection and dispatch messages. Also, OOB module `oob.receiveInvitation` method is quite similar to `messageSender.receiveMessage`:

MessageReceiver
- isPacked -> unpack
- connectionExists -> reuse connection
- transformMessage
- validateMessage
- saveSession
- dispatch one message

OOB
- connectionExists -> create or reuse connection
- transformMessage
- validateMessage
- dispatch one or more messages

[...]

PR #531, opened 2021-11-08
feat: add support for new out-of-band protocol

jakubkoci:
This is the first step in the implementation of #344. 

It corresponds approximately to Step 1 defined in [Aries RFC 0496: Transition to the Out of Band and DID Exchange Protocols](https://github.com/hyperledger/aries-rfcs/blob/main/features/0496-transition-to-oob-and-did-exchange/README.md).

As you can see I decided to create an Out-of-band module and keep the implementation there. I realized that putting it into the Connection module could cause more issues than benefits right now. It has only two drawbacks:
* API users would need to decide what method to call for now. But it allows more gradual adoption of new out-of-band by users. When we have a full implementation, users will ideally just call one method for OOB messages.
* I copy-pasted methods for connection creation. That can be solved by calling the whole Connections module from the OOB module.

### DIFF
--- a/packages/core/src/modules/oob/OutOfBandMessage.ts
+++ b/packages/core/src/modules/oob/OutOfBandMessage.ts
@@ -1,4 +1,4 @@
-import type { UnpackedMessage, UnpackedMessageContext } from '../../types'
+import type { UnpackedMessage } from '../../types'
 import type { DidCommService } from '../connections/models/did/service/DidCommService'
 
 import { Expose, Type } from 'class-transformer'
@@ -10,7 +10,7 @@ import { JsonEncoder } from '../../utils/JsonEncoder'
 
 const VERSION = '1.1'
 
-interface OutOfBandMessageOptions {
+export interface OutOfBandMessageOptions {
   id?: string
   label?: string
   goalCode?: string
--- a/packages/core/src/modules/oob/OutOfBandMessage.ts
+++ b/packages/core/src/modules/oob/OutOfBandMessage.ts
@@ -28,7 +28,13 @@ export class OutOfBandMessage extends AgentMessage {
     }
   }
 
+  public addHandshakeProtocol(protocol: string) {
+    if (!this.handshakeProtocols) this.handshakeProtocols = []
+    this.handshakeProtocols.push(protocol)
+  }
+
   public addRequest(message: AgentMessage) {
+    if (!this.requests) this.requests = []
     const requestAttachment = new Attachment({
       id: 'request-0',
       mimeType: 'application/json',
@@ -39,14 +45,18 @@ export class OutOfBandMessage extends AgentMessage {
     this.requests.push(requestAttachment)
   }
 
-  public getRequests(): UnpackedMessage[] {
-    return this.requests.map((request) => {
+  public getRequests(): UnpackedMessage[] | undefined {
+    return this.requests?.map((request) => {
       if (request.data.base64) {
         return JsonEncoder.fromBase64(request.data.base64)
       } else {
         throw new Error('There is no base64 encoded data in attachment')
       }
     })
+  }
+
+  public addService(service: DidCommService) {
+    this.services.push(service)
   }
 
   @Equals(OutOfBandMessage.type)
@@ -65,9 +75,7 @@ export class OutOfBandMessage extends AgentMessage {
 
   // TODO what type is it, should we create an enum
   @Expose({ name: 'handshake_protocols' })
-  public readonly handshakeProtocols: string[] = []
-
-  public readonly services: DidCommService[] = []
+  public handshakeProtocols?: string[]
 
   @Expose({ name: 'requests~attach' })
   @Type(() => Attachment)
@@ -76,5 +84,7 @@ export class OutOfBandMessage extends AgentMessage {
     each: true,
   })
   @IsInstance(Attachment, { each: true })
-  private readonly requests: Attachment[] = []
+  private requests?: Attachment[]
+
+  public readonly services: DidCommService[] = []
 }
--- a/packages/core/src/modules/oob/OutOfBandMessage.ts
+++ b/packages/core/src/modules/oob/OutOfBandMessage.ts
@@ -8,8 +8,6 @@ import { AgentMessage } from '../../agent/AgentMessage'
 import { Attachment, AttachmentData } from '../../decorators/attachment/Attachment'
 import { JsonEncoder } from '../../utils/JsonEncoder'
 
-const VERSION = '1.1'
-
 export interface OutOfBandMessageOptions {
   id?: string
   label?: string
@@ -36,7 +34,7 @@ export class OutOfBandMessage extends AgentMessage {
   public addRequest(message: AgentMessage) {
     if (!this.requests) this.requests = []
     const requestAttachment = new Attachment({
-      id: 'request-0',
+      id: this.generateId(),
       mimeType: 'application/json',
       data: new AttachmentData({
         base64: JsonEncoder.toBase64(message.toJSON()),
@@ -61,7 +59,7 @@ export class OutOfBandMessage extends AgentMessage {
 
   @Equals(OutOfBandMessage.type)
   public readonly type = OutOfBandMessage.type
-  public static readonly type = `https://didcomm.org/out-of-band/${VERSION}/invitation`
+  public static readonly type = `https://didcomm.org/out-of-band/1.1/invitation`
 
   public readonly label?: string
 
--- a/packages/core/src/modules/oob/OutOfBandMessage.ts
+++ b/packages/core/src/modules/oob/OutOfBandMessage.ts
@@ -8,11 +8,14 @@ import { AgentMessage } from '../../agent/AgentMessage'
 import { Attachment, AttachmentData } from '../../decorators/attachment/Attachment'
 import { JsonEncoder } from '../../utils/JsonEncoder'
 
-export interface OutOfBandMessageOptions {
+interface OutOfBandMessageOptions {
   id?: string
   label?: string
   goalCode?: string
   goal?: string
+  accept?: string[]
+  handshakeProtocols?: string[]
+  services: DidCommService[]
 }
 
 export class OutOfBandMessage extends AgentMessage {
@@ -23,12 +26,10 @@ export class OutOfBandMessage extends AgentMessage {
       this.label = options.label
       this.goalCode = options.goalCode
       this.goal = options.goal
+      this.accept = options.accept
+      this.handshakeProtocols = options.handshakeProtocols
+      this.services = options.services
     }
-  }
-
-  public addHandshakeProtocol(protocol: string) {
-    if (!this.handshakeProtocols) this.handshakeProtocols = []
-    this.handshakeProtocols.push(protocol)
   }
 
   public addRequest(message: AgentMessage) {
@@ -53,10 +54,6 @@ export class OutOfBandMessage extends AgentMessage {
     })
   }
 
-  public addService(service: DidCommService) {
-    this.services.push(service)
-  }
-
   @Equals(OutOfBandMessage.type)
   public readonly type = OutOfBandMessage.type
   public static readonly type = `https://didcomm.org/out-of-band/1.1/invitation`
@@ -69,7 +66,7 @@ export class OutOfBandMessage extends AgentMessage {
   public readonly goal?: string
 
   // TODO what type is it, is there any enum or should we create a new one
-  public readonly accept: string[] = []
+  public readonly accept?: string[] = []
 
   // TODO what type is it, should we create an enum
   @Expose({ name: 'handshake_protocols' })
--- a/packages/core/src/modules/oob/OutOfBandMessage.ts
+++ b/packages/core/src/modules/oob/OutOfBandMessage.ts
@@ -1,0 +1,80 @@
+import type { UnpackedMessage, UnpackedMessageContext } from '../../types'
+import type { DidCommService } from '../connections/models/did/service/DidCommService'
+
+import { Expose, Type } from 'class-transformer'
+import { Equals, IsArray, IsInstance, ValidateNested } from 'class-validator'
+
+import { AgentMessage } from '../../agent/AgentMessage'
+import { Attachment, AttachmentData } from '../../decorators/attachment/Attachment'
+import { JsonEncoder } from '../../utils/JsonEncoder'
+
+const VERSION = '1.1'
+
+interface OutOfBandMessageOptions {
+  id?: string
+  label?: string
+  goalCode?: string
+  goal?: string
+}
+
+export class OutOfBandMessage extends AgentMessage {
+  public constructor(options: OutOfBandMessageOptions) {
+    super()
+    if (options) {
+      this.id = options.id ?? this.generateId()
+      this.label = options.label
+      this.goalCode = options.goalCode
+      this.goal = options.goal
+    }
+  }
+
+  public addRequest(message: AgentMessage) {
+    const requestAttachment = new Attachment({
+      id: 'request-0',
+      mimeType: 'application/json',
+      data: new AttachmentData({
+        base64: JsonEncoder.toBase64(message.toJSON()),
+      }),
+    })
+    this.requests.push(requestAttachment)
+  }
+
+  public getRequests(): UnpackedMessage[] {
+    return this.requests.map((request) => {
+      if (request.data.base64) {
+        return JsonEncoder.fromBase64(request.data.base64)
+      } else {
+        throw new Error('There is no base64 encoded data in attachment')
+      }
+    })
+  }
+
+  @Equals(OutOfBandMessage.type)
+  public readonly type = OutOfBandMessage.type
+  public static readonly type = `https://didcomm.org/out-of-band/${VERSION}/invitation`
+
+  public readonly label?: string
+
+  @Expose({ name: 'goal_code' })
+  public readonly goalCode?: string
+
+  public readonly goal?: string
+
+  // TODO what type is it, is there any enum or should we create a new one
+  public readonly accept: string[] = []
+
+  // TODO what type is it, should we create an enum
+  @Expose({ name: 'handshake_protocols' })
+  public readonly handshakeProtocols: string[] = []
+
+  public readonly services: DidCommService[] = []
+
+  @Expose({ name: 'requests~attach' })
+  @Type(() => Attachment)
+  @IsArray()
+  @ValidateNested({
+    each: true,
+  })
+  @IsInstance(Attachment, { each: true })
+  private readonly requests: Attachment[] = []
+}
--- a/packages/core/src/modules/oob/OutOfBandModule.ts
+++ b/packages/core/src/modules/oob/OutOfBandModule.ts
@@ -1,4 +1,5 @@
 import type { AgentMessage } from '../../agent/AgentMessage'
+import type { AgentMessageReceivedEvent } from '../../agent/Events'
 import type { ConnectionRecord } from '../../modules/connections'
 import type { UnpackedMessageContext } from '../../types'
 
@@ -6,6 +7,7 @@ import { Lifecycle, scoped } from 'tsyringe'
 
 import { Dispatcher } from '../../agent/Dispatcher'
 import { EventEmitter } from '../../agent/EventEmitter'
+import { AgentEventTypes } from '../../agent/Events'
 import { MessageSender } from '../../agent/MessageSender'
 import { createOutboundMessage } from '../../agent/helpers'
 import { InboundMessageContext } from '../../agent/models/InboundMessageContext'
@@ -66,7 +68,7 @@ export class OutOfBandModule {
     // Either way, we need to get routing
     const mediationRecord = await this.mediationRecipientService.discoverMediation()
     const routing = await this.mediationRecipientService.getRouting(mediationRecord)
-    const { connectionRecord: connectionRecord, message: invitation } = await this.connectionService.createInvitation({
+    const { connectionRecord: connectionRecord } = await this.connectionService.createInvitation({
       routing,
     })
 
@@ -85,6 +87,56 @@ export class OutOfBandModule {
     return { outOfBandMessage, connectionRecord }
   }
 
+  public async createOobMessage(message: AgentMessage) {
+    if (!message.service) {
+      throw new AriesFrameworkError(
+        `Out of band message with id ${message.id} and type ${message.type} does not have a ~service decorator`
+      )
+    }
+
+    const outOfBandMessage = new OutOfBandMessage({
+      goal: 'To issue a Faber College Graduate credential',
+      goalCode: 'issue-vc',
+      label: 'Faber College',
+    })
+
+    outOfBandMessage.accept.push('didcomm/aip2;env=rfc587')
+    outOfBandMessage.accept.push('didcomm/aip2;env=rfc19')
+
+    // Protocol compatibilty!
+    // To support newer OOB messages we need to add service from message and then remove `~service` attribute from message
+    outOfBandMessage.services.push(message.service.toDidCommService())
+    message.service = undefined
+    outOfBandMessage.addRequest(message)
+
+    return outOfBandMessage
+  }
+
+  public async receiveOobMessage(outOfBandMessage: OutOfBandMessage) {
+    if (outOfBandMessage.handshakeProtocols.length > 0) {
+      throw new AriesFrameworkError('OOB message contains unsupported `handshake_protocols` attribute.')
+    }
+    const messages = outOfBandMessage.getRequests()
+    messages.forEach(async (unpackedMessage) => {
+      try {
+        // Protocol compatibilty!
+        // To support older OOB messages we need to decorate message with `~service` attribute
+        const message = await this.transformMessage({ message: unpackedMessage })
+        message.setService(outOfBandMessage.services[0])
+
+        this.eventEmitter.emit<AgentMessageReceivedEvent>({
+          type: AgentEventTypes.AgentMessageReceived,
+          payload: {
+            message: message.toJSON(),
+          },
+        })
+      } catch (error) {
+        // eslint-disable-next-line no-console
+        console.error('error', error)
+      }
+    })
+  }
+
   public async receiveInvitation(outOfBandMessage: OutOfBandMessage, config: { autoAccept: boolean }) {
     const mediationRecord = await this.mediationRecipientService.discoverMediation()
     const routing = await this.mediationRecipientService.getRouting(mediationRecord)
@@ -95,39 +147,9 @@ export class OutOfBandModule {
       connectionRecord = await this.acceptInvitation(connectionRecord.id)
     }
 
-    this.connectionService
-      .returnWhenIsConnected(connectionRecord.id)
-      .then((connection) => {
-        const messages = outOfBandMessage.getRequests()
-        messages.forEach(async (unpackedMessage) => {
-          // We can't use event emitter because it will pass unpacked message into MessageReceiver without keys therefore we won't be able to find the connection created in previous step
-          // this.eventEmitter.emit<AgentMessageReceivedEvent>({
-          //   type: AgentEventTypes.AgentMessageReceived,
-          //   payload: {
-          //     message,
-          //   },
-          // })
-          try {
-            const message = await this.transformMessage({ message: unpackedMessage })
-            const messageContext = new InboundMessageContext(message, {
-              // Only make the connection available in message context if the connection is ready
-              // To prevent unwanted usage of unready connections. Connections can still be retrieved from
-              // Storage if the specific protocol allows an unready connection to be used.
-              connection: connection?.isReady ? connection : undefined,
-              // senderVerkey: senderKey,
-              // recipientVerkey: recipientKey,
-            })
-            await this.dispatcher.dispatch(messageContext)
-          } catch (error: any) {
-            // eslint-disable-next-line no-console
-            console.error('message dispatch error', error, error.message)
-          }
-        })
-      })
-      .catch((error) => {
-        // eslint-disable-next-line no-console
-        console.error('error', error, error.message)
-      })
+    if (outOfBandMessage.getRequests().length > 0) {
+      throw new AriesFrameworkError('OOB invitation contains unsupported `request~attach` attribute.')
+    }
 
     return connectionRecord
   }
--- a/packages/core/src/modules/oob/OutOfBandModule.ts
+++ b/packages/core/src/modules/oob/OutOfBandModule.ts
@@ -337,21 +337,10 @@ export class OutOfBandModule {
       const [service] = services
 
       if (typeof service === 'string') {
-        const did = service
-        const didResolutionResult = await this.dids.resolve(did)
-        const [didCommService] = didResolutionResult.didDocument?.didCommServices || []
-        if (!didCommService) {
-          throw new AriesFrameworkError('Resolved did does not have any service.')
-        }
-        const { recipientKeys, routingKeys, serviceEndpoint } = didCommService
-        plaintextMessage['~service'] = {
-          recipientKeys,
-          routingKeys,
-          serviceEndpoint,
-        }
-      } else {
-        plaintextMessage['~service'] = service
+        throw new AriesFrameworkError('Dids are not currently supported in out-of-band message services attribute.')
       }
+
+      plaintextMessage['~service'] = service
 
       this.eventEmitter.emit<AgentMessageReceivedEvent>({
         type: AgentEventTypes.AgentMessageReceived,
--- a/packages/core/src/modules/oob/OutOfBandModule.ts
+++ b/packages/core/src/modules/oob/OutOfBandModule.ts
@@ -248,9 +248,13 @@ export class OutOfBandModule {
   }
 
   private async findExistingConnection(services: DidCommService[]) {
-    const [service] = services
-    const existingConnection = await this.connectionsModule.findByTheirKey(service.recipientKeys[0])
-    return existingConnection
+    for (const service of services) {
+      for (const recipientKey of service.recipientKeys) {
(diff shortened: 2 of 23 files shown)


packages/core/src/modules/oob/OutOfBandMessage.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  public addHandshakeProtocol(protocol: string) {
+    if (!this.handshakeProtocols) this.handshakeProtocols = []
+    this.handshakeProtocols.push(protocol)
+  }
+
+  public addRequest(message: AgentMessage) {
+    if (!this.requests) this.requests = []
+    const requestAttachment = new Attachment({
+      id: 'request-0',
```
TimoGlastra, line 39:
```suggestion
      id: this.generateId(),
```

packages/core/src/modules/oob/OutOfBandMessage.ts
```diff
@@ -0,0 +1,90 @@
...
+        base64: JsonEncoder.toBase64(message.toJSON()),
+      }),
+    })
+    this.requests.push(requestAttachment)
+  }
+
+  public getRequests(): UnpackedMessage[] | undefined {
+    return this.requests?.map((request) => {
+      if (request.data.base64) {
+        return JsonEncoder.fromBase64(request.data.base64)
```
TimoGlastra, line 51: We should add something to handle attachments in a generic way (ACA-Py has it). Data could be base64 or json (or others but the first two are most important)

```ts
const data = request.data.get() // could be either json or base64, but it doesn't matter
```

packages/core/src/modules/oob/OutOfBandMessage.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  public readonly label?: string
+
+  @Expose({ name: 'goal_code' })
+  public readonly goalCode?: string
+
+  public readonly goal?: string
+
+  // TODO what type is it, is there any enum or should we create a new one
+  public readonly accept: string[] = []
```
TimoGlastra, line 74: No enum yet, need to create a new one

packages/core/src/modules/oob/OutOfBandMessage.ts
```diff
@@ -0,0 +1,90 @@
...
+  @Expose({ name: 'requests~attach' })
+  @Type(() => Attachment)
+  @IsArray()
+  @ValidateNested({
+    each: true,
+  })
+  @IsInstance(Attachment, { each: true })
+  private requests?: Attachment[]
+
+  public readonly services: DidCommService[] = []
```
TimoGlastra, line 89: Can also be plain dids (strings)

packages/core/src/modules/oob/OutOfBandModule.ts
```diff
@@ -0,0 +1,181 @@
...
+      query: `*`,
+    })
+    const featuresMessage = await this.disoverFeaturesService.createDisclose(queryMessage)
+    const { protocols } = featuresMessage
+
+    const handshakeProtocols = ['https://didcomm.org/didexchange', 'https://didcomm.org/connections']
+
+    const supportedHandshakeProtocols = protocols
+      .map((p) => p.protocolId.slice(0, -1))
+      .filter((pId) => handshakeProtocols.find((hp) => pId.startsWith(hp)))
```
TimoGlastra, line 62: I think it would be nice to extract this to the discover features service and call it like this:

``` ts
const supported = await this.disoverFeaturesService.getSupportedProtocols(['https://didcomm.org/didexchange', 'https://didcomm.org/connections'])
```

packages/core/src/modules/oob/OutOfBandModule.ts
```diff
@@ -0,0 +1,181 @@
...
+      throw new AriesFrameworkError('There is no handshake protocol supported. Agent can not create a connection.')
+    }
+
+    // Create connection.
+    // Eventually, we can create just an OutOfBand record here.
+    // The OOB record can be also used for connection-less communication in general.
+    // Either way, we need to get routing
+    const mediationRecord = await this.mediationRecipientService.discoverMediation()
+    const routing = await this.mediationRecipientService.getRouting(mediationRecord)
+    const { connectionRecord: connectionRecord } = await this.connectionService.createInvitation({
```
TimoGlastra, line 74:
```suggestion
    const { connectionRecord } = await this.connectionService.createInvitation({
```

packages/core/src/modules/oob/OutOfBandModule.ts
```diff
@@ -0,0 +1,181 @@
...
+      routing,
+    })
+
+    const outOfBandMessage = new OutOfBandMessage(options)
+    outOfBandMessage.accept.push('didcomm/aip2;env=rfc587')
+    outOfBandMessage.accept.push('didcomm/aip2;env=rfc19')
```
TimoGlastra, line 80: See https://github.com/hyperledger/aries-rfcs/blob/main/features/0044-didcomm-file-and-mime-types/README.md#defined-profiles

We don't support `didcomm/aip2;env=rfc587`, and I think also not `didcomm/aip2;env=rfc19` but not sure

packages/core/src/modules/oob/OutOfBandModule.ts
```diff
@@ -0,0 +1,181 @@
...
+    supportedHandshakeProtocols.forEach((p) => outOfBandMessage.addHandshakeProtocol(p))
+    connectionRecord.didDoc.didCommServices.forEach((s) => outOfBandMessage.addService(s))
```
TimoGlastra, line 82: I'd be in favour of adding most of it in the constructor of the OOB message, feel quite verbose this all the maps and for eaches.

packages/core/src/modules/oob/OutOfBandMessage.ts
```diff
@@ -0,0 +1,90 @@
...
+    })
+    this.requests.push(requestAttachment)
+  }
+
+  public getRequests(): UnpackedMessage[] | undefined {
+    return this.requests?.map((request) => {
+      if (request.data.base64) {
+        return JsonEncoder.fromBase64(request.data.base64)
+      } else {
+        throw new Error('There is no base64 encoded data in attachment')
```
TimoGlastra, line 53: AriesFrameworkError

packages/core/src/modules/oob/OutOfBandModule.ts
```diff
@@ -0,0 +1,181 @@
...
+    if (!outOfBandMessage.handshakeProtocols || outOfBandMessage.handshakeProtocols.length === 0) {
+      throw new AriesFrameworkError('Missing required `handshake_protocols` attribute in OOB message.')
+    }
+
+    const mediationRecord = await this.mediationRecipientService.discoverMediation()
+    const routing = await this.mediationRecipientService.getRouting(mediationRecord)
+    const invitation = new ConnectionInvitationMessage({ label: 'connection label', ...outOfBandMessage.services[0] })
+
+    let connectionRecord = await this.connectionService.processInvitation(invitation, { routing })
+    if (config.autoAccept) {
```
TimoGlastra, line 107: Do we want this to take into account agent config autoAcceptConnection?

packages/core/src/modules/oob/OutOfBandModule.ts
```diff
@@ -0,0 +1,181 @@
...
+      )
+    }
+
+    const outOfBandMessage = new OutOfBandMessage(options)
+    outOfBandMessage.accept.push('didcomm/aip2;env=rfc587')
+    outOfBandMessage.accept.push('didcomm/aip2;env=rfc19')
+
+    // To support newer OOB messages we need to add service from message and then remove `~service` attribute from message
+    outOfBandMessage.services.push(message.service.toDidCommService())
+    message.service = undefined
```
TimoGlastra, line 138: I think the advantage of using OOB is that the `AgentMessage` itself doesn't need the `~service` anymore.

packages/core/src/modules/oob/OutOfBandModule.ts
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  /**
+   * Takes all messages from `requests~attach` attribute and pass them to the agent via event emitter.
+   *
+   * @param outOfBandMessage
+   */
+  public async receiveOobMessage(outOfBandMessage: OutOfBandMessage): Promise<void> {
+    if (outOfBandMessage.handshakeProtocols) {
+      throw new AriesFrameworkError('OOB message contains unsupported `handshake_protocols` attribute.')
```
TimoGlastra, line 151: This feels a bit misleading, `handshake_protocols` is definitely supported, just not in this method.

Also, does this mean `handshake_protocols` and `requests~attach` can't be combined in AFJ?

TimoGlastra (changes requested):
Nice work @jakubkoci! Glad to see the OOB protocol finally being added to AFJ.

I'm not so sure on splitting the `createInvitation` from `createOobMessage`. Both create an `OOB` invitation. And I think we're missing some of the nice combinations here. `handshake_protocols` and `requests~attach` are not exclusive, so I think it may be easier to just create a single `createInvitation` and a single `receiveInvitation` method that allows to specify what should be done with handshake protocols / attached messages.

Second point is that I'd rather not merge this PR as is. It feels like we are leaving a lot of TODOs out there, while I would rather just add OOB as a whole, where we have really thought about all implementation details. There are still a lot of unknowns with connection reuse, connectionless, did-exchange and an OOB record that I would rather keep it in a separate OOB branch for now. What do you think?

packages/core/src/modules/oob/OutOfBandMessage.ts
jakubkoci: Thanks, updated.

packages/core/src/modules/oob/OutOfBandModule.ts
jakubkoci: That's a good point I also felt like it should be extracted somewhere. I moved it to `disoverFeaturesService`. Eventually, it can be part of the dispatcher.

packages/core/src/modules/oob/OutOfBandModule.ts
jakubkoci: Fixed

packages/core/src/modules/oob/OutOfBandModule.ts
jakubkoci: This is just a workaround for now. It allows us to pass the `AgentMessage` into `receiveMessage` and still be able to process it with the current implementation.

packages/core/src/modules/oob/OutOfBandMessage.ts
```diff
@@ -0,0 +1,90 @@
...
+    })
+  }
+
+  public addService(service: DidCommService) {
+    this.services.push(service)
+  }
+
+  @Equals(OutOfBandMessage.type)
+  public readonly type = OutOfBandMessage.type
+  public static readonly type = `https://didcomm.org/out-of-band/${VERSION}/invitation`
```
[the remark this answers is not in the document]
jakubkoci: Updated.

packages/core/src/modules/oob/OutOfBandModule.ts
jakubkoci: I'll definitely look at the unification of those methods as you suggested in another comment.

But the combination of `handshake_protocols` and `requests~attach` is not allowed according to Step 1 in [Aries RFC 0496: Transition to the Out of Band and DID Exchange Protocols](https://github.com/hyperledger/aries-rfcs/blob/main/features/0496-transition-to-oob-and-did-exchange/README.md)

packages/core/src/modules/oob/OutOfBandMessage.ts
jakubkoci: That could be good, at least for `base64` and `json`. However, if I look at [Aries RFC 0017: Attachments](https://github.com/hyperledger/aries-rfcs/blob/main/concepts/0017-attachments/README.md) it can also contain `links`. Should we call some HTTP request in that case or just return the link?

jakubkoci:
Hi @TimoGlastra. Thanks for the suggestions and feedback. I unified it into just two methods `createMessage` and `receiveMessage`. It make sense for `receiveMessage` I don't have doubt there, but `createMessage` looks much worse now. I'm gonna look at oob record and other missing parts you mentioned, how they fit in and maybe it improves implementation by itself :) 

I'm good with not merging it right now. I wasn't sure if we want to do it according to that transition spec or not. If we don't we can wait and merge as a whole, I guess.

packages/core/src/modules/oob/OutOfBandMessage.ts
jakubkoci: This perhaps should not be implemented as a part of this PR, right?

jakubkoci: I added also if statement for `json` and created issue #548

packages/core/src/modules/oob/OutOfBandMessage.ts
jakubkoci: Not sure if the enum is the best solution. It could prevent the framework from being dynamically enhanced.

packages/core/src/modules/oob/OutOfBandModule.ts
jakubkoci: Updated.

packages/core/src/modules/oob/OutOfBandModule.ts
jakubkoci: Updated.

packages/core/src/modules/oob/OutOfBandMessage.ts
jakubkoci: Fixed.

packages/core/src/modules/oob/OutOfBandModule.ts
jakubkoci: Updated. It supports both `handshake_protocols` and `requests~attach`.

jakubkoci:
It's still not polished as I wished, but it's working according to Step 1 at least. It actually does more than Step 1. I wrote a brief overview of what's done and what's missing:

### Done
* create oob with connection invitation
* create oob with connection invitation with message requests
* create oob only with message requests
* create a connection based on oob invitation
* create a connection based on oob invitation and process message requests
* process connection-less message requests
* support for both old and new encoded URL

### Todo
* Handle handshake reuse message
* Replace old `ConnectionInvitationMessage` with the new `OutOfBandMessage`.
* Use DIDs. Transform to/from Peer DID, DID key, DID peer (#554 helps a lot with it)
* Allow creating oob for an existing connection.
* Process only the first supported message
* Validate services

packages/core/src/modules/oob/OutOfBandModule.ts
```diff
@@ -0,0 +1,297 @@
...
+
+    if ((!handshakeProtocols || handshakeProtocols.length === 0) && (!messages || messages?.length === 0)) {
+      throw new AriesFrameworkError(
+        'One or both of handshake_protocols and requests~attach MUST be included in the message.'
+      )
+    }
+
+    if (handshakeProtocols) {
+      if (!this.areHandshakeProtocolsSupported(handshakeProtocols)) {
+        throw new AriesFrameworkError('Handshake protocols are not supported.')
```
TimoGlastra, line 177: Maybe would be good to include in the error which handshakeprotocols were requestsed vs which are supported

packages/core/src/modules/oob/OutOfBandModule.ts
```diff
@@ -0,0 +1,297 @@
...
+                  `Agent unsubscribed before connection got into ${ConnectionState.Complete} state`,
+                  error
+                )
+              } else {
+                this.logger.error('Promise waiting for the connection to be complete failed.', error)
+              }
+            })
+        } else {
+          const connectionServices = connectionRecord.theirDidDoc?.didCommServices
+          this.emitMessages(connectionServices, messages)
```
TimoGlastra, line 222: I know you put this in the TODO, but we shouldn't emit all messages. I think it is important to fix this as it is not an incomplete way to handle it, but an incorrect way to handle it.

packages/core/src/modules/oob/OutOfBandModule.ts
```diff
@@ -0,0 +1,297 @@
...
+            .catch((error) => {
+              if (error instanceof EmptyError) {
+                this.logger.warn(
+                  `Agent unsubscribed before connection got into ${ConnectionState.Complete} state`,
+                  error
+                )
+              } else {
+                this.logger.error('Promise waiting for the connection to be complete failed.', error)
+              }
+            })
```
TimoGlastra, line 219: If the agent shuts down mid creating a connection, the messages are lost and won't be processed when starting the agent again. If we were to do this with a middleware of some kind (or listening to events) this wouldn't be dependant on the agent being alive during the whole process.

Maybe that is a fair tradeoff to make for now. If we create a connection, we could store the invitation in the connection record and resume later on. If we don't create a connection, we see it as more ephemeral and the agent must be online during the complete exchange.

TimoGlastra: I think we should start storing the oob invitation in a didcomm message record (as we talked about during wg calls). This way we don't need a didexchange record and we keep a lot more flexibility going forward

packages/core/src/modules/oob/OutOfBandModule.ts
```diff
@@ -0,0 +1,297 @@
...
+        const existingConnection = await this.connectionsModule.findByTheirKey(recipientKey)
+        return existingConnection
+      }
+    }
+  }
+
+  private async createConnection(outOfBandMessage: OutOfBandMessage, autoAccept: boolean) {
+    const { services, label } = outOfBandMessage
+    const invitation = new ConnectionInvitationMessage({
+      label: label || '',
```
TimoGlastra, line 260: To my understanding `label` is not optional, either way we shouldn't use `''`

packages/core/src/modules/oob/OutOfBandModule.ts
```diff
@@ -0,0 +1,297 @@
...
+    for (const service of services) {
+      for (const recipientKey of service.recipientKeys) {
+        // TODO Encode the key and endpoint of the service block in a Peer DID numalgo 2 and using that DID instead of a service block
+        const existingConnection = await this.connectionsModule.findByTheirKey(recipientKey)
+        return existingConnection
+      }
+    }
+  }
+
+  private async createConnection(outOfBandMessage: OutOfBandMessage, autoAccept: boolean) {
```
TimoGlastra, line 257: I'd make this an object. We have had this multiple times in the framework that we had to change the API because a config parameter was added. High chance we'll be adding something like an `alias` later on. Making it an object now means we only need to add a parameter to the config

```suggestion
  private async createConnection(outOfBandMessage: OutOfBandMessage, { autoAccept }: { autoAccept: boolean }) {
```

packages/core/src/modules/oob/OutOfBandModule.ts
```diff
@@ -0,0 +1,297 @@
...
+      label: label || '',
+      ...services[0],
+    })
+    const connectionRecord = await this.connectionsModule.receiveInvitation(invitation, {
+      autoAcceptConnection: autoAccept,
+    })
+    return connectionRecord
+  }
+
+  private emitMessages(services: DidCommService[] | undefined, messages: PlaintextMessage[]) {
```
TimoGlastra, line 269: Having a possibly undefined parameter before an always defined parameter seems weird to me. Maybe we can switch them?

Also, in what case is `services` undefined? Shouldn't we handle it there.

packages/core/src/modules/oob/handlers/HandshakeReuseHandler.ts
```diff
@@ -0,0 +1,12 @@
+import type { Handler, HandlerInboundMessage } from '../../../agent/Handler'
+
+import { HandshakeReuseMessage } from '../messages/HandshakeReuseMessage'
+
+export class HandshakeReuseHandler implements Handler {
+  public supportedMessages = [HandshakeReuseMessage]
+
+  public async handle(inboundMessage: HandlerInboundMessage<HandshakeReuseHandler>) {
+    // const connection = inboundMessage.assertReadyConnection()
+    // return createOutboundMessage(connection, new AgentMessage())
```
TimoGlastra, line 10: commented code 🤢

TimoGlastra: Maybe log an error message (`this.logger.error('Out of band reuse not implemented yet')`)?

packages/core/src/modules/oob/messages/OutOfBandMessage.ts
```diff
@@ -0,0 +1,123 @@
...
+import { AgentMessage } from '../../../agent/AgentMessage'
+import { Attachment, AttachmentData } from '../../../decorators/attachment/Attachment'
+import { AriesFrameworkError } from '../../../error'
+import { JsonEncoder } from '../../../utils/JsonEncoder'
+import { JsonTransformer } from '../../../utils/JsonTransformer'
+import { MessageValidator } from '../../../utils/MessageValidator'
+
+interface OutOfBandMessageOptions {
+  id?: string
+  label?: string
```
TimoGlastra, line 17: Not optional to my understanding

packages/core/src/modules/oob/messages/OutOfBandMessage.ts
```diff
@@ -0,0 +1,123 @@
...
+      if (request.data.base64) {
+        return JsonEncoder.fromBase64(request.data.base64)
+      }
+
+      if (request.data.json) {
+        return request.data.json
+      }
+
+      throw new AriesFrameworkError('There is no supported data attribute in attachment')
+    })
```
TimoGlastra, line 62:
```suggestion
    return this.requests?.map((request) => request.data.getDataAsJson())
```

packages/core/src/modules/oob/messages/OutOfBandMessage.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  public static async fromJson(json: Record<string, unknown>) {
+    const invitation = JsonTransformer.fromJSON(json, OutOfBandMessage)
+    await MessageValidator.validate(invitation)
+    return invitation
+  }
+
+  @Equals(OutOfBandMessage.type)
+  public readonly type = OutOfBandMessage.type
+  public static readonly type = `https://didcomm.org/out-of-band/1.1/invitation`
```
TimoGlastra, line 96:
```suggestion
  public static readonly type = `https://didcomm.org/out-of-band/1.0/invitation`
```

packages/core/src/modules/oob/__tests__/OutOfBandMessage.test.ts
```diff
@@ -0,0 +1,87 @@
...
+import { JsonEncoder } from '../../../utils/JsonEncoder'
+import { JsonTransformer } from '../../../utils/JsonTransformer'
+import { OutOfBandMessage } from '../messages/OutOfBandMessage'
+
+describe('OutOfBandMessage', () => {
+  describe('toUrl', () => {
+    test('encode the message into the URL containg the base64 encoded invitation as the oob query parameter', async () => {
+      const domain = 'https://example.com/ssi'
+      const json = {
+        '@type': 'https://didcomm.org/out-of-band/1.1/invitation',
```
TimoGlastra, line 12:
```suggestion
        '@type': 'https://didcomm.org/out-of-band/1.0/invitation',
```

packages/core/src/modules/oob/__tests__/OutOfBandMessage.test.ts
```diff
@@ -0,0 +1,87 @@
...
+  describe('fromUrl', () => {
+    test('decode the URL containing the base64 encoded invitation as the oob parameter into an `OutOfBandMessage`', async () => {
+      const invitationUrl =
+        'http://example.com/ssi?oob=eyJAdHlwZSI6Imh0dHBzOi8vZGlkY29tbS5vcmcvb3V0LW9mLWJhbmQvMS4xL2ludml0YXRpb24iLCJAaWQiOiI2OTIxMmEzYS1kMDY4LTRmOWQtYTJkZC00NzQxYmNhODlhZjMiLCJsYWJlbCI6IkZhYmVyIENvbGxlZ2UiLCJnb2FsX2NvZGUiOiJpc3N1ZS12YyIsImdvYWwiOiJUbyBpc3N1ZSBhIEZhYmVyIENvbGxlZ2UgR3JhZHVhdGUgY3JlZGVudGlhbCIsImhhbmRzaGFrZV9wcm90b2NvbHMiOlsiaHR0cHM6Ly9kaWRjb21tLm9yZy9kaWRleGNoYW5nZS8xLjAiLCJodHRwczovL2RpZGNvbW0ub3JnL2Nvbm5lY3Rpb25zLzEuMCJdLCJzZXJ2aWNlcyI6WyJkaWQ6c292OkxqZ3BTVDJyanNveFllZ1FEUm03RUwiXX0K'
+
+      const invitation = await OutOfBandMessage.fromUrl(invitationUrl)
+      const json = JsonTransformer.toJSON(invitation)
+
+      expect(json).toEqual({
+        '@type': 'https://didcomm.org/out-of-band/1.1/invitation',
```
TimoGlastra, line 38:
```suggestion
        '@type': 'https://didcomm.org/out-of-band/1.0/invitation',
```

packages/core/src/modules/oob/__tests__/OutOfBandMessage.test.ts
```diff
@@ -0,0 +1,87 @@
...
+        handshake_protocols: ['https://didcomm.org/didexchange/1.0', 'https://didcomm.org/connections/1.0'],
+        services: ['did:sov:LjgpST2rjsoxYegQDRm7EL'],
+      })
+    })
+  })
+
+  describe('fromJson', () => {
+    test('create an instance of `OutOfBandMessage` from JSON object', async () => {
+      const json = {
+        '@type': 'https://didcomm.org/out-of-band/1.1/invitation',
```
TimoGlastra, line 52:
```suggestion
        '@type': 'https://didcomm.org/out-of-band/1.0/invitation',
```

packages/core/src/modules/oob/__tests__/OutOfBandMessage.test.ts
```diff
@@ -0,0 +1,87 @@
...
+
+      const invitation = await OutOfBandMessage.fromJson(json)
+
+      expect(invitation).toBeDefined()
+      expect(invitation).toBeInstanceOf(OutOfBandMessage)
+    })
+
+    test('throw validation error when services attribute is empty', async () => {
+      const json = {
+        '@type': 'https://didcomm.org/out-of-band/1.1/invitation',
```
TimoGlastra, line 69:
```suggestion
        '@type': 'https://didcomm.org/out-of-band/1.0/invitation',
```

packages/core/src/modules/oob/messages/OutOfBandMessage.ts
```diff
@@ -0,0 +1,123 @@
...
+  @ValidateNested({
+    each: true,
+  })
+  @IsInstance(Attachment, { each: true })
+  @IsOptional()
+  private requests?: Attachment[]
+
+  @IsArray()
+  @ArrayNotEmpty()
+  public services: DidCommService[] = []
```
TimoGlastra, line 122: Can also be string values. 


```suggestion
  public services: Array<DidCommService | string> = []
```

packages/core/src/modules/oob/OutOfBandModule.ts
```diff
@@ -0,0 +1,297 @@
...
+interface CreateOutOfBandMessageConfig {
+  label?: string
+  goalCode?: string
+  goal?: string
+  handshake: boolean
+  multiUseInvitation: boolean
+}
+
+interface ReceiveOutOfBandMessageConfig {
+  autoAccept: boolean
```
TimoGlastra, line 33: I was thinking whether we should also call this `autoAcceptConnection`? If we don't set `autoAccept`: true we currently still process the messages. Whether they are auto accepted depends on the agent config (`autoAcceptCredentials`, `autoAcceptProofs`, etc...). `autoAccept` makes me think  the `autoAccept` property has influence on the accepance of the `requests~attach` messages

TimoGlastra (commented):
Overall structure LGTM 👍 , but have left a lot of quibbles

packages/core/src/modules/oob/__tests__/OutOfBandMessage.test.ts
jakubkoci: Hmm, I'm confused why the header in Aries RFC says version 1.1? 🤔 

Aries RFC 0434: Out-of-Band Protocol 1.1

TimoGlastra: Ah my bad. You're right it should be 1.1. But then we should use 1.1 everywhere. Other messages are using 1.0, also the handshake_protocols array

jakubkoci: You're right that `HandshakeReuseMessage` is using `1.0` and should use `1.1`, I'll fix that. I'm not sure what you mean by version in other messages and `handshake_protocols` array. The `handshake_protocols` array contains just what's supported by the framework and `did-exchange` and `connections` RFCs define version `1.0`.

packages/core/src/modules/oob/OutOfBandModule.ts
jakubkoci: [OOB spec](https://github.com/hyperledger/aries-rfcs/tree/bed4989dd6517f7a9de3696800e57e4c6ef49231/features/0434-outofband#invitation-httpsdidcommorgout-of-bandverinvitation) defines `label` as optional.

However 0160: Connection Protocol does not define it as optional.

What do you suggest instead of `''`? I don't think we should put any predefined string there.

packages/core/src/modules/oob/messages/OutOfBandMessage.ts
jakubkoci: According to [OOB spec](https://github.com/hyperledger/aries-rfcs/tree/bed4989dd6517f7a9de3696800e57e4c6ef49231/features/0434-outofband#invitation-httpsdidcommorgout-of-bandverinvitation):  

> `label` - [optional] a self-attested string that the receiver may want to display to the user, likely about who sent the out-of-band message.

I think it makes sense. There is no reason for that to be required. It's not trusted attributed anyway. It's unnecessary because the receiver can decide what `label` the connection should have.

packages/core/src/modules/oob/OutOfBandModule.ts
jakubkoci: Yeah, I renamed it to `autoAcceptConnection`. Would it make sense to eventually have a similar `autoAccept` config for `requests~attach` messages?

packages/core/src/modules/oob/OutOfBandModule.ts
TimoGlastra: Make `label` optional in the connection record. It feels weird to store an empty string as the label. This means from now on when using AFJ in a mobile application I would need to check all labels to make sure they don't have the value `''`.

packages/core/src/modules/oob/messages/OutOfBandMessage.ts
TimoGlastra: Ah sorry my bad 👍

packages/core/src/modules/oob/OutOfBandModule.ts
jakubkoci: I agree that it should be optional in a connection record. But that doesn't solve the problem here. We're trying to build upon existing code and reuse `ConnectionInvitationMessage` where it's not optional. 

Maybe, we should not use `ConnectionInvitationMessage` then.

packages/core/src/modules/oob/OutOfBandModule.ts
jakubkoci: Good point. I made it mandatory and it throws an error when there are no services for a connection record.

packages/core/src/modules/oob/messages/OutOfBandMessage.ts
jakubkoci: That's great 👍  

FYI I was surprised a bit when I tried to call it as `request.getDataAsJson()`. Wouldn't be better to move the method up to `Attchment` class?

TimoGlastra: Yeah sure, I think we can move it one level up!

packages/core/src/modules/oob/handlers/HandshakeReuseHandler.ts
jakubkoci: 👍  Updated.

packages/core/src/modules/oob/messages/OutOfBandMessage.ts
jakubkoci: This change required resolving the did if the services contain a string, so I added it.

packages/core/src/modules/oob/OutOfBandModule.ts
jakubkoci: Updated. It takes only the first supported message and throws an error if there is no message supported.

I haven't found a precise definition in the spec if we should allow only the "request" type of message from each protocol. What do you think?

packages/core/src/modules/oob/messages/OutOfBandMessage.ts
TimoGlastra: > resolving the did if the services contain a string

I'm not sure if we're quite ready for that. We don't support public dids yet to create connections, and the did resolver returns the new diddoc format while the rest of AFJ is still built on the old format.

We should write some transformers for the connection record diddoc, and update it in the connection reocrd and also integrate did based message sending in the message sender (so we don't store the did document in the connection record, but rather resolve it (or get from cache) when sending a message)

jakubkoci: Should we throw an error when there is a string in that array? Then, maybe, this change doesn't make sense.

jakubkoci:
I consider the code in here approved. I moved it to the `feat/out-of-band` branch, where I'll continue with adding did exchange protocol.

PR #535, opened 2021-11-15
fix(core)!: prefixed internal metadata with _internal/

blu3beri:
BREAKING CHANGE: internal metadata is now prefixed with `_internal` to avoid clashing and accidental overwriting of internal data.

- All the internal metadata MUST be prefixed with `_internal` to indicate to the framework consumer that this data is internal metadata and is required by the framework.

### DIFF
--- a/packages/core/src/modules/credentials/__tests__/CredentialRecord.test.ts
+++ b/packages/core/src/modules/credentials/__tests__/CredentialRecord.test.ts
@@ -17,7 +17,7 @@ describe('CredentialRecord', () => {
         ],
       })
 
-      credentialRecord.metadata.set('indyCredential', {
+      credentialRecord.metadata.set('_internal/indyCredential', {
         credentialDefinitionId: 'Th7MpTaRZVRYnPiabds81Y:3:CL:17:TAG',
         schemaId: 'TL1EaPFCZ8Si5aUrqScBDt:2:test-schema-1599055118161:1.0',
       })
@@ -29,7 +29,7 @@ describe('CredentialRecord', () => {
           age: '25',
         },
         metadata: {
-          indyCredential: {
+          '_internal/indyCredential': {
             credentialDefinitionId: 'Th7MpTaRZVRYnPiabds81Y:3:CL:17:TAG',
             schemaId: 'TL1EaPFCZ8Si5aUrqScBDt:2:test-schema-1599055118161:1.0',
           },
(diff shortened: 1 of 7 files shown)


TimoGlastra (approved):
Nice, I'm quite happy with the prefix solution.

TimoGlastra:
@jakubkoci you also happy (enough) with this approach?

If not let's discuss other solutions during the working group call.

jakubkoci (approved):
I'm good with that. 

It still seems a bit weird from the framework developer perspective. How should I know why is that and what does the prefix mean? I feel like I'm quite annoying already. Just giving feedback, sorry 😄   

Maybe, we should have two different `metadata` attributes. Another question I'm asking is if those attributes `indyRequest`, `indyCredential` should be part of the record, not credential record but `IndyCredentialRecord` 🤷‍♂️  Just thinking aloud.

On the other hand, I can imagine that `metadata` can be good for plugins and extensions. That would need to be part of publicly accessible `metadata` and we will have the same problem when users can delete or change the metadata belonging to an extension or plugin. Maybe I'm nitpicking too much and I want to achieve something impossible, or only possible with a crazy amount of effort. Overall, don't take me wrong, although I see some issues I don't see this as a critical problem.

blu3beri:
> I'm good with that.
> 
> It still seems a bit weird from the framework developer perspective. How should I know why is that and what does the prefix mean? I feel like I'm quite annoying already. Just giving feedback, sorry 😄
> 
> Maybe, we should have two different `metadata` attributes. Another question I'm asking is if those attributes `indyRequest`, `indyCredential` should be part of the record, not credential record but `IndyCredentialRecord` 🤷‍♂️ Just thinking aloud.
> 
> On the other hand, I can imagine that `metadata` can be good for plugins and extensions. That would need to be part of publicly accessible `metadata` and we will have the same problem when users can delete or change the metadata belonging to an extension or plugin. Maybe I'm nitpicking too much and I want to achieve something impossible, or only possible with a crazy amount of effort. Overall, don't take me wrong, although I see some issues I don't see this as a critical problem.

I agree with your points here and before 0.1.0 we probably should revisit the design of this. I think the "easiest" solution is to only have public metadata. This does mean that for all of our potential internal metadata we have to find another solution. I will create an issue for this and if you have any suggestion for a better design feel free to leave a comment there. I will think about this myself as well.

PR #537, opened 2021-11-15
refactor(core): allow map or record on proof request

janrtvld:
Signed-off-by: janrtvld <jan@animo.id>

### DIFF
--- a/packages/core/src/modules/proofs/models/ProofRequest.ts
+++ b/packages/core/src/modules/proofs/models/ProofRequest.ts
@@ -16,8 +16,8 @@ export interface ProofRequestOptions {
   nonce: string
   nonRevoked?: RevocationInterval
   ver?: '1.0' | '2.0'
-  requestedAttributes?: Record<string, ProofAttributeInfo>
-  requestedPredicates?: Record<string, ProofPredicateInfo>
+  requestedAttributes?: Record<string, ProofAttributeInfo> | Map<string, ProofAttributeInfo>
+  requestedPredicates?: Record<string, ProofPredicateInfo> | Map<string, ProofPredicateInfo>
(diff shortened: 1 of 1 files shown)


TimoGlastra (approved):

JamesKEbert:
Can I ask what the motivations here were? I'm running into a couple of issues related to presentations/revocation with this recent change and just want to understand this :)
Thanks!

jakubkoci:
I'm also curious. If I understand the change correctly it's part of public API. As a developer using some library I really like to use just plain JS objects. If we see huge benefits in using `Map` then we should maybe convert all such attributes into a map. Otherwise, it can be confusing for users what data structure they're actually calling.

jakubkoci:
In what cases does it contain `Record<string, ProofAttributeInfo>` instead of `Map<string, ProofAttributeInfo>`?

TimoGlastra:
Are you talking specifically about allowing both map and record, or the change of using a map again.

For allowing both a map and record is for ease of use. The change of using a map again is that the validators weren't working correctly. I recently added tests and checks because I noticed that whatever I used for the proof request, AFJ wasn't complaining. class-transformer doesn't support records, but does support maps.

> As a developer using some library I really like to use just plain JS objects.

I agree with this @jakubkoci, but at the moment that is not the case. You should actually pass in class instances in the public API. It works because in TS if an interface matches that of a class it is valid. Allowing a map `Record<string, ProofAttributeInfo` in addition to a `Map` is exactly for making it easier for the user

>  I'm running into a couple of issues related to presentations/revocation with this recent change and just want to understand this :)

What are the issues you're running into. Don't want this cause any issues, it should make the framework more robust :)

PR #541, opened 2021-11-22
fix(node): added node v12 support for libindy check

blu3beri:


TimoGlastra (approved):

TimoGlastra (changes requested):
-

PR #543, opened 2021-11-23
refactor: move methods to get supported protocols to Dispatcher

jakubkoci:
It's mainly about refactoring. I will use the "Add filter of protocols by message families" (86f336c) in #531. 

However, I'm not sure about the "Remove slash from protocol ID to be aligned with spec" (ee5a3a4). It seems like spec https://github.com/hyperledger/aries-rfcs/blob/main/concepts/0003-protocols/README.md#piuri defines protocol URI without the last `/` character, so I removed that. Let me know If you think it's dangerous breaking change or incorrect change :)


TimoGlastra (approved):
LGTM!

If the spec says we should not use the trailing slash, that's probably right :)

PR #556, opened 2021-12-01
fix(core): expose record metadata types

blu3beri:

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -79,10 +79,10 @@ export interface OutboundPackage {
 }
 
 // Metadata type for `_internal/indyCredential`
-export interface indyCredential {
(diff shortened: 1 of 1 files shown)


packages/core/src/types.ts
```diff
@@ -76,3 +77,12 @@ export interface OutboundPackage {
   endpoint?: string
   connectionId?: string
 }
+
+// Metadata type for `_internal/indyCredential`
+export interface indyCredential {
```
TimoGlastra, line 82: I think we should make interfaces always uppercase.

Maybe also make them a bit more elaborate?

```ts
export interface IndyCredentialMetadata

// and

export interface IndyCredentialRequestMetadata
```

TimoGlastra (approved):

PR #557, opened 2021-12-01
feat: add question answer protocol

seajensen:
# Summary of Changes

Added a question answer module (https://github.com/hyperledger/aries-rfcs/blob/main/features/0113-question-answer/README.md) to send and receive question messages and to send and receive answer messages with a valid response. 

# Pull Request Checklist


packages/core/src/modules/question-answer/QuestionAnswerModule.ts
```diff
@@ -0,0 +1,99 @@
...
+   * Create a question message with possible valid responses, then send message to the
+   * holder
+   *
+   * @param connectionId connection to send the question message to
+   * @param question question text included in message
+   * @param validResponses array of possible responses to question
+   * @param detail optional details for question
+   * @returns QuestionAnswer record
+   */
+  public async sendQuestion(connectionId: string, question: string, validResponses: ValidResponse[], detail?: string) {
```
TimoGlastra, line 43: With this many parameters I'd lean to an object. In most modules we take the connection/record id as the first parameter and all other properties go into a second config parameter

packages/core/src/modules/question-answer/QuestionAnswerModule.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  /**
+   * Create an answer message as the holder and send it in response to a question message
+   *
+   * @param connectionId connection to send the answer message to
+   * @param threadId thread id for the QuestionAnswer record
+   * @param response response included in the answer message
+   * @returns QuestionAnswer record
+   */
+  public async sendAnswer(connectionId: string, threadId: string, response: string) {
```
TimoGlastra, line 66: When a record already exists we usually use the record id, maybe we can also use that here?

packages/core/src/modules/question-answer/handlers/QuestionAnswerHandler.ts
```diff
@@ -0,0 +1,30 @@
+import type { Handler, HandlerInboundMessage } from '../../../agent/Handler'
+import type { QuestionAnswerService } from '../services'
+
+import { QuestionMessage, AnswerMessage } from '../messages'
+
+export class QuestionMessageHandler implements Handler {
```
TimoGlastra, line 6: Usually we only place a single class per file

packages/core/src/modules/question-answer/messages/AnswerMessage.ts
```diff
@@ -0,0 +1,31 @@
+import { Expose } from 'class-transformer'
+import { Equals, IsString } from 'class-validator'
+
+import { AgentMessage } from '../../../agent/AgentMessage'
+
+export class AnswerMessage extends AgentMessage {
+  /**
+   * Create new AnswerMessage instance.
+   * sentTime will be assigned to new Date if not passed, id will be assigned to uuid/v4 if not passed
```
TimoGlastra, line 9: Outdated comment

packages/core/src/modules/question-answer/messages/AnswerMessage.ts
```diff
@@ -0,0 +1,31 @@
...
+      this.setThread({ threadId: options.threadId })
+      this.response = options.response
+    }
+  }
+
+  @Equals(AnswerMessage.type)
+  public readonly type = AnswerMessage.type
+  public static readonly type = 'https://didcomm.org/questionanswer/1.0/answer'
+
+  //   public validResponse: ValidResponse[string]
```
TimoGlastra, line 26: ?

packages/core/src/modules/question-answer/services/QuestionAnswerService.ts
```diff
@@ -0,0 +1,250 @@
...
+    return { questionMessage, questionAnswerRecord }
+  }
+
+  /**
+   * receive question message and create record for responder role
+   *
+   * @param messageContext the message context containing a question message
+   * @returns QuestionAnswer record
+   */
+  public async receiveQuestion(messageContext: InboundMessageContext<QuestionMessage>): Promise<QuestionAnswerRecord> {
```
TimoGlastra, line 82: I think we usually call methods used by the handlers `processXXX`

packages/core/src/modules/question-answer/services/QuestionAnswerService.ts
```diff
@@ -0,0 +1,250 @@
...
+
+    this.logger.debug(`Receiving question message with id ${questionMessage.id}`)
+
+    if (!connection) {
+      throw new AriesFrameworkError(`Connection for verkey ${messageContext.recipientVerkey} not found!`)
+    }
+
+    if (!connection.theirKey) {
+      throw new AriesFrameworkError(`Connection with verkey ${connection.verkey} has no recipient keys.`)
+    }
```
TimoGlastra, line 93:
```suggestion
    // Assert ready connection
    const connection = messageContext.assertReadyConnection()

```

packages/core/src/modules/question-answer/services/QuestionAnswerService.ts
```diff
@@ -0,0 +1,250 @@
...
+   */
+  public async createAnswer(questionAnswerRecord: QuestionAnswerRecord, response: string) {
+    const answerMessage = new AnswerMessage({ response: response, threadId: questionAnswerRecord.threadId })
+
+    questionAnswerRecord.response = response
+
+    if (questionAnswerRecord.validResponses.some((e) => e.text === response)) {
+      await this.updateState(questionAnswerRecord, QuestionAnswerState.AnswerSent)
+    } else {
+      this.logger.error(`Response does not match valid responses`)
```
TimoGlastra, line 131: Maybe we can throw an error here? It is not clear when calling this that the record won't get updated.

packages/core/src/modules/question-answer/QuestionAnswerModule.ts
```diff
@@ -0,0 +1,99 @@
...
+      questionRecord,
+      response
+    )
+
+    if (questionAnswerRecord.state === QuestionAnswerState.AnswerSent) {
+      const outboundMessage = createOutboundMessage(connection, answerMessage)
+      await this.messageSender.sendMessage(outboundMessage)
+    } else {
+      throw new AriesFrameworkError(`Unable to send message without valid response`)
+    }
```
TimoGlastra, line 81: We could then also ditch this check

packages/core/src/modules/question-answer/services/QuestionAnswerService.ts
```diff
@@ -0,0 +1,250 @@
...
+
+      await this.updateState(questionAnswerRecord, QuestionAnswerState.AnswerReceived)
+
+      return questionAnswerRecord
+    } catch (error) {
+      this.logger.error(
+        `Unable to get question answer record by threadId "${answerMessage.threadId}" and connection id "${connection?.id}" and unable to update state`,
+        { error }
+      )
+      throw error
```
TimoGlastra, line 169: Why the try/catch here? I don't see it in other places. Maybe the base errors are sufficient here?

packages/core/src/modules/question-answer/services/QuestionAnswerService.ts
```diff
@@ -0,0 +1,250 @@
...
+      type: QuestionAnswerEventTypes.QuestionAnswerStateChanged,
+      payload: {
+        previousState,
+        questionAnswerRecord: questionAnswerRecord,
+      },
+    })
+  }
+
+  /**
+   * @todo use connection from message context
```
TimoGlastra, line 196: Old comment?

TimoGlastra (changes requested):
Nice work @seajensen.

Only thing I'm thinking is whether we want to put this in core or ext? E.g push notifications lives in ext, QA is also not part of AIP so should it also live in ext? Would be good to have some consistency in what goed where, but I do see the potential of this getting added to the core package.

genaris:
> Nice work @seajensen.
> 
> Only thing I'm thinking is whether we want to put this in core or ext? E.g push notifications lives in ext, QA is also not part of AIP so should it also live in ext? Would be good to have some consistency in what goed where, but I do see the potential of this getting added to the core package.

First of all, good work @seajensen! It's great to see how easy it becomes to create modules implementing new protocols in AFJ.

This is a good point @TimoGlastra. I think the choice could depend on the current RFC lifecycle status, which is eventually related to AIPs (i.e. most ACCEPTED RFCs become part of an AIP version). Push Notification Native RFC is currently PROPOSED, so it is not likely to be widely used right now and it's fine in ext repository. In this particular case, Question/Answer is DEMONSTRATED, which is not so different. 

If we choose to put it in aries-javascript-framework-ext repository, I think there are a few classes/methods to be exposed (for instance `BaseRecord` and `createOutboundMessage`). This will be useful also for anyone who wants to use custom protocols in AFJ. I can create a PR including these exports shortly.

TimoGlastra:
We had a discussion about this during the WG call, but I forgot the outcome. Do we add this to the core of the framework, or make it an extension?

@JamesKEbert @seajensen do you know what the outcome was?

JamesKEbert:
I unfortunately I was out sick during that WG call, but I recall @seajensen communicating to me that we were going to merge it in.

TimoGlastra (approved):

TimoGlastra:
@seajensen if you can resolve the conflicts we can get this merged!

PR #558, opened 2021-12-01
ci: create tag on alpha release

TimoGlastra:

@JamesKEbert I think this should do it. It will create a tag names `alpha.<number>` on each release. I think lerna search for the default `vX.X.X` prefix so should not cause any issues. 

Let's try!


JamesKEbert (approved):
Awesome, thanks for working on this here!

TimoGlastra:
CI failed, but that is unrelated to this change. Need to merge a new PR for it to actually run

Issue #58, opened 2020-07-03
Add support for RFC 0035: Problem Report


jakubkoci:
It would be also nice to define some structure or rules for error handling answering questions: 
- When and where we want to throw/handle errors?
- Problem report adopted into other protocols
- initiating problem report as standalone protocol

nbAmit:
@jakubkoci @TimoGlastra  I Would like to work on this

TimoGlastra:
Awesome!

jakubkoci:
Shouldn't we reopen this issue? I think we have still a few things we should address before closing.

TimoGlastra:
I think it would be better to open specific issues for the outstanding tasks.

Are you mainly talking about the outstanding tasks from the PR or the tasks form the list in this issue?

PR #560, opened 2021-12-02
feat: add problem report protocol

nbAmit:
### Summary of Changes
Implemented the problem report [RFC-0035](https://github.com/hyperledger/aries-rfcs/blob/main/features/0035-report-problem/README.md)

Problem Report RFC effective for reporting of errors and warnings to other agents. Once this PR is merged agent will be able to report problems occurred in `Connection`, `Issue Credential`  & `Present Proof`

### Related Issue 
#58


packages/core/src/modules/connections/handlers/ConnectionResponseHandler.ts
```diff
@@ -16,14 +17,29 @@ export class ConnectionResponseHandler implements Handler {
...
-      const { message } = await this.connectionService.createTrustPing(connection.id, { responseRequested: false })
-      return createOutboundMessage(connection, message)
+      // TODO: should we only send ping message in case of autoAcceptConnection or always?
+      // In AATH we have a separate step to send the ping. So for now we'll only do it
+      // if auto accept is enable
+      if (connection.autoAcceptConnection ?? this.agentConfig.autoAcceptConnections) {
+        const { message } = await this.connectionService.createTrustPing(connection.id, { responseRequested: false })
+        return createOutboundMessage(connection, message)
+      }
+    } catch (error) {
```
TimoGlastra, line 30: Same here, we then wouldn't need a try/catch in every handler class

packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -213,14 +213,18 @@ export class ConnectionService {
...
     }
 
     // Create new connection if using a multi use invitation
     if (connectionRecord.multiUseInvitation) {
       if (!routing) {
-        throw new AriesFrameworkError(
-          'Cannot process request for multi-use invitation without routing object. Make sure to call processRequest with the routing parameter provided.'
+        throw new ConnectionProblemReportError(
+          'Cannot process request for multi-use invitation without routing object. Make sure to call processRequest with the routing parameter provided.',
+          ConnectionProblemReportReason.RequestNotAccepted
```
TimoGlastra, line 227: This is a framework error so I think we  shouldn't send a problem report for it. Or at least not with this message and this reason

packages/core/src/modules/credentials/services/CredentialService.ts
```diff
@@ -713,6 +722,32 @@ export class CredentialService {
...
+   * @param messageContext The message context containing a credential problem report message
+   * @returns credential record associated with the credential problem report message
+   *
+   */
+  public async processProblemReport(
+    messageContext: InboundMessageContext<CredentialProblemReportMessage>
+  ): Promise<CredentialRecord> {
+    const { message: credentialProblemReportMessage, connection } = messageContext
+
+    this.logger.debug(`Processing credential ack with id ${credentialProblemReportMessage.id}`)
```
TimoGlastra, line 737: ack should be problem report

packages/core/src/modules/credentials/services/CredentialService.ts
```diff
@@ -713,6 +722,32 @@ export class CredentialService {
...
+
+    const credentialRecord = await this.getByThreadAndConnectionId(
+      credentialProblemReportMessage.threadId,
+      connection?.id
+    )
+
+    // Update record
+    credentialRecord.errorMsg = `${credentialProblemReportMessage.description.code} : ${credentialProblemReportMessage.description.en}`
```
TimoGlastra, line 745:
```suggestion
    credentialRecord.errorMsg = `${credentialProblemReportMessage.description.code}: ${credentialProblemReportMessage.description.en}`
```

packages/core/src/modules/problem-reports/errors/ProblemReportError.ts
```diff
@@ -0,0 +1,20 @@
...
+export class ProblemReportError extends AriesFrameworkError {
+  public problemReport: ProblemReportMessage
+
+  public constructor(message: string, { problemCode }: ProblemReportErrorOptions) {
+    super(message, undefined)
+    this.problemReport = new ProblemReportMessage({
+      description: {
+        en: message,
+        code: problemCode,
+      },
```
TimoGlastra, line 17: Ah I see you used the problem report message in error approach here. I think we can also leverage that in the connection / credential / proofs messages

packages/core/src/modules/problem-reports/messages/ProblemReportMessage.ts
```diff
@@ -0,0 +1,115 @@
+// Create a base ProblemReportMessage message class and add it to the messages directory
+import { Equals, IsEnum, IsOptional, IsString } from 'class-validator'
+
+import { AgentMessage } from '../../../agent/AgentMessage'
+import { CommonMessageType } from '../../common/messages/CommonMessageType'
+
+export enum WhoRetriesStatus {
+  YOU = 'YOU',
```
TimoGlastra, line 8: We usually write enums as 

`PascalCase = '<rfc value>'`

packages/core/src/modules/problem-reports/messages/ProblemReportMessage.ts
```diff
@@ -0,0 +1,115 @@
...
+      this.id = options.id || this.generateId()
+      this.description = options.description
+      this.problem_items = options.problem_items
+      this.who_retries = options.who_retries
+      this.fix_hint = options.fix_hint
+      this.impact = options.impact
+      this.where = options.where
+      this.noticed_time = options.noticed_time
+      this.tracking_uri = options.tracking_uri
+      this.escalation_uri = options.escalation_uri
```
TimoGlastra, line 76: we use camelCame for the property names. You can use the `@Expose({ name: 'the_snake_case_name' })` to map it to another name in the json. See some of the other message classes for an example.

packages/core/src/modules/problem-reports/messages/ProblemReportMessage.ts
```diff
@@ -0,0 +1,115 @@
...
+    }
+  }
+
+  @Equals(ProblemReportMessage.type)
+  public readonly type: string = ProblemReportMessage.type
+  public static readonly type: string = CommonMessageType.ProblemReport
+
+  public description!: DescriptionOptions
+
+  @IsOptional()
```
TimoGlastra, line 86:
```suggestion
  @IsOptional()
  @IsArray()
  @IsString({ each: true })
```

packages/core/src/modules/proofs/services/ProofService.ts
```diff
@@ -558,6 +564,29 @@ export class ProofService {
...
+   * @param messageContext The message context containing a presentation problem report message
+   * @returns proof record associated with the presentation acknowledgement message
+   *
+   */
+  public async processProblemReport(
+    messageContext: InboundMessageContext<PresentationProblemReportMessage>
+  ): Promise<ProofRecord> {
+    const { message: presentationProblemReportMessage, connection } = messageContext
+
+    this.logger.debug(`Processing presentation ack with id ${presentationProblemReportMessage.id}`)
```
TimoGlastra, line 579: should be problem report instead of ack

packages/core/src/modules/proofs/services/ProofService.ts
```diff
@@ -558,6 +564,29 @@ export class ProofService {
...
+
+    const proofRecord = await this.getByThreadAndConnectionId(presentationProblemReportMessage.threadId, connection?.id)
+
+    // Update record
+    await this.updateState(proofRecord, ProofState.None)
+    proofRecord.errorMsg = `${presentationProblemReportMessage.description.code} : ${presentationProblemReportMessage.description.en}`
```
TimoGlastra, line 585:
```suggestion
    proofRecord.errorMsg = `${presentationProblemReportMessage.description.code}: ${presentationProblemReportMessage.description.en}`
```

TimoGlastra, line 585: If you add the errorMsg assignment to before the updateState call you don't need have to call `update` again. (the updateState will also store the record)

packages/core/src/modules/credentials/services/CredentialService.ts
```diff
@@ -713,6 +722,32 @@ export class CredentialService {
...
+    this.update(credentialRecord)
```
TimoGlastra, line 746: update not needed if you call updateState

packages/core/src/modules/proofs/ProofState.ts
```diff
@@ -12,4 +12,5 @@ export enum ProofState {
   PresentationReceived = 'presentation-received',
   Declined = 'declined',
   Done = 'done',
+  None = 'none',
```
TimoGlastra, line 15: I'm not sure if state `none` is a good representation of the state. Shouldn't this be more something like `error`?

@jakubkoci @JamesKEbert any insights?

packages/core/src/modules/connections/messages/ConnectionProblemReportMessage.ts
```diff
@@ -0,0 +1,24 @@
...
+
+import { Equals } from 'class-validator'
+
+import { ProblemReportMessage } from '../../problem-reports/messages/ProblemReportMessage'
+
+export type ConnectionProblemReportMessageOptions = ProblemReportMessageOptions
+
+/**
+ * @see https://github.com/hyperledger/aries-rfcs/blob/main/features/0035-report-problem/README.md
+ */
```
TimoGlastra, line 11: I can't find the connection problem report handler.

TimoGlastra (commented):
Hi @nbAmit this is really great!

I left some comments, but the main overarching points are:
- Try to avoid non-null assertions, as it bypasses the typescript checks
- If we try include the problem report message also in the error for connection/credential/proofs we can handle it generically in the dispatcher instead of the separate handler classes

Also I have some questions:
- Do you think we need to expose this as a seperate module / do we need to expose this on the issue / proofs / connections modules (like aca-py)?
- Have you tested this with connectionless?
- Maybe it would be good to also add some e2e tests using problem report messages

TimoGlastra (changes requested):
.

nbAmit:
Thanks @TimoGlastra , I will check and update the code base on your suggestions and also check the e2e testing part as well.

packages/core/src/agent/Dispatcher.ts
```diff
@@ -45,15 +48,32 @@ class Dispatcher {
...
-
-      throw error
+      const problemReportMessage = error.problemReport
+
+      if (
+        (problemReportMessage instanceof ConnectionProblemReportMessage ||
+          problemReportMessage instanceof CredentialProblemReportMessage ||
+          problemReportMessage instanceof PresentationProblemReportMessage) &&
+        messageContext.connection
+      ) {
```
TimoGlastra, line 58: If we check if the error is an instance of `ProblemReportError` we can generically handle it for all errors that extend from the `ProblemReportError`. this way we don't have to check for specific message type in the dispatcher

```suggestion
      if (problemReportMessage instanceof ProblemReportError && messageContext.connection) {
          const problemReportMessage = error.problemReport
```

packages/core/src/agent/MessageReceiver.ts
```diff
@@ -84,15 +93,18 @@ export class MessageReceiver {
       unpackedMessage.message
     )
 
-    const message = await this.transformMessage(unpackedMessage)
+    let message: AgentMessage | null = null
     try {
-      await MessageValidator.validate(message)
+      message = await this.transformMessage(unpackedMessage)
+      this.validateMessage(message)
```
TimoGlastra, line 99:
```suggestion
      await this.validateMessage(message)
```

packages/core/src/agent/MessageReceiver.ts
```diff
@@ -84,15 +93,18 @@ export class MessageReceiver {
...
-      this.logger.error(`Error validating message ${message.type}`, {
-        errors: error,
-        message: message.toJSON(),
-      })
-
+      if (message) {
+        this.logger.error(`Error validating message ${message.type}`, {
+          errors: error,
+          message: message.toJSON(),
+        })
```
TimoGlastra, line 105: The validate method also logs an error, I think logging it once is good enough

packages/core/src/agent/MessageReceiver.ts
```diff
@@ -174,12 +186,65 @@ export class MessageReceiver {
...
+   */
+  private async validateMessage(message: AgentMessage) {
+    try {
+      await MessageValidator.validate(message)
+    } catch (error) {
+      this.logger.error(`Error validating message ${message.type}`, {
+        errors: error,
+        message: message.toJSON(),
+      })
+      throw new ProblemReportError(`No message class found for message type "${message.type}"`, {
```
TimoGlastra, line 212: The error messages is not updated I think?

packages/core/src/agent/MessageReceiver.ts
```diff
@@ -174,12 +186,65 @@ export class MessageReceiver {
...
+      threadId: unpackedMessage.message['@id'],
+    })
+    const outboundMessage = createOutboundMessage(connection, problemReportMessage)
+    if (outboundMessage && isOutboundServiceMessage(outboundMessage)) {
+      await this.messageSender.sendMessageToService({
+        message: outboundMessage.payload,
+        service: outboundMessage.service,
+        senderKey: outboundMessage.senderKey,
+        returnRoute: true,
+      })
```
TimoGlastra, line 245: If connection is provided (which is always the case in this method) this will never be true, as it used for connectionless exchanges.

you can call `messageSender.sendMessage` directly

packages/core/src/modules/connections/errors/ConnectionProblemReportError.ts
```diff
@@ -0,0 +1,22 @@
+import type { ConnectionProblemReportReason } from '.'
+import type { ProblemReportErrorOptions } from '../../problem-reports'
+
+import { AriesFrameworkError } from '../../../error/AriesFrameworkError'
+import { ConnectionProblemReportMessage } from '../messages'
+
+interface ConnectionProblemReportErrorOptions extends ProblemReportErrorOptions {
+  problemCode: ConnectionProblemReportReason
+}
+export class ConnectionProblemReportError extends AriesFrameworkError {
```
TimoGlastra, line 10: I think we should create a generic `ProblemReportError` where the `ConnectionProblemReportError` can extend from. This way we can check just for `instanceof ProblemReportError` in the dispatcher. As the `ConectionProblemReportMessage` extends the `ProblemReportMessage` it is also possible to use the connection specific problem report message for the `problemReport` property

packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -403,6 +407,31 @@ export class ConnectionService {
...
+   *
+   */
+  public async processProblemReport(
+    messageContext: InboundMessageContext<ConnectionProblemReportMessage>
+  ): Promise<ConnectionRecord> {
+    const { message: connectionProblemReportMessage, connection, recipientVerkey } = messageContext
+
+    this.logger.debug(`Processing connection problem report for verkey ${recipientVerkey}`)
+
+    if (!connection) {
```
TimoGlastra, line 424: There is a good chance the `connection` won't be availalbe on the `messageContext` yet as the connection is not complete yet. If you look at the other methods in the conenction service you can see they often retrieve it from the repository. This is only the case in the connection implementation as before the protocol is finished there is no complete connection yet.

packages/core/src/agent/MessageReceiver.ts
```diff
@@ -174,12 +186,65 @@ export class MessageReceiver {
...
+    }
+  }
+
+  /**
+   * Send the problem report message (https://didcomm.org/notification/1.0/problem-report) to the recipient.
+   * @param message error message to send
+   * @param connection connection to send the message to
+   * @param unpackedMessage received unpackedMessage
+   */
+  private async sendProblemReportMessage(
```
TimoGlastra, line 224: I think we should add an check here whether the received message is of `@type` == `https://didcomm.org/notification/1.0/problem-report`and otherwise throw an error (not a problem report error) that we are not sending a problem report as response to a problem.  to prevent accidential problem report loops from happening. We should probably never respond with a problem report to a problem report.

packages/core/src/modules/credentials/errors/CredentialProblemReportError.ts
```diff
@@ -0,0 +1,22 @@
+import type { ProblemReportErrorOptions } from '../../problem-reports'
+import type { CredentialProblemReportReason } from './CredentialProblemReportReason'
+
+import { AriesFrameworkError } from '../../../error/AriesFrameworkError'
+import { CredentialProblemReportMessage } from '../messages'
+
+interface CredentialProblemReportErrorOptions extends ProblemReportErrorOptions {
+  problemCode: CredentialProblemReportReason
+}
+export class CredentialProblemReportError extends AriesFrameworkError {
```
TimoGlastra, line 10: Same, should extend `ProblemReportError`

packages/core/src/modules/problem-reports/messages/ProblemReportMessage.ts
```diff
@@ -0,0 +1,123 @@
...
+      this.escalationUri = options.escalationUri
+    }
+  }
+
+  @Equals(ProblemReportMessage.type)
+  public readonly type: string = ProblemReportMessage.type
+  public static readonly type: string = CommonMessageType.ProblemReport
+
+  @IsOptional()
+  public description!: DescriptionOptions
```
TimoGlastra, line 86: I think description is not optional

TimoGlastra (commented):
Great improvements @nbAmit!

I have a few additional comments, but once those are resolved this is good to go I think

packages/core/src/modules/connections/errors/ConnectionProblemReportError.ts
TimoGlastra, line 10: This doesn't extend problem report error yet

TimoGlastra (commented):
Great work @nbAmit! One comment left

TimoGlastra (approved):
LGTM

packages/core/src/agent/MessageReceiver.ts
```diff
@@ -9,21 +9,29 @@ import { Lifecycle, scoped } from 'tsyringe'
...
 import { AgentConfig } from './AgentConfig'
 import { Dispatcher } from './Dispatcher'
 import { EnvelopeService } from './EnvelopeService'
+import { MessageSender } from './MessageSender'
 import { TransportService } from './TransportService'
+import { createOutboundMessage } from './helpers'
 import { InboundMessageContext } from './models/InboundMessageContext'
 
+export enum ProblemReportReason {
+  MessageParseFailure = 'message-parse-failure',
```
jakubkoci, line 27: Just from my curiosity... Is `message-parse-failure` part of some RFC?

Maybe, we can move it to `modules/problem-reports`. What do you think?

jakubkoci (commented):
Good job. Just added a few notes. Nothing serious.

TimoGlastra:
Sorry @jakubkoci, just merged.....

TimoGlastra:
@nbAmit could you take a look at Jakubs comments and address those in a follow up PR?

TimoGlastra:
Also I think we should only allow problem reports to be received for exchanges with connections for now, as with the `processProblemReport` method now in every service, they are not tied to a specific connection and we also don't check whether this is a valid connectionless exchange (hard to do right now). This is quite a large security flaw, as everyone who nows a thread id can now send a problem report message to it.

Could you add a `messageContext.assertReadyConnection` to the credential and proof problem report message handlers?  For the connection handler we will need to add checks based on where in the protocol we are

TimoGlastra:
@nbAmit are you going to pick these up? The comments from @jakubkoci have some breaking changes to records so we should try to fix those ASAP before people update to this version. If you don't have the time let me know and I'll pick it up.

nbAmit:
Hi @TimoGlastra This week I am on vacation so is it okay if I start working on Monday?

packages/core/src/agent/MessageReceiver.ts
nbAmit: I have referred to the Aries cloud agent python codebase and they are using the same enum https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/core/dispatcher.py#L150

nbAmit: Yes, We can move it to modules/problem-reports

packages/core/src/decorators/signature/SignatureDecoratorUtils.ts
```diff
@@ -29,7 +29,9 @@ export async function unpackAndVerifySignatureDecorator(
   const isValid = await wallet.verify(signerVerkey, signedData, signature)
 
   if (!isValid) {
-    throw new AriesFrameworkError('Signature is not valid!')
+    throw new ConnectionProblemReportError('Signature is not valid', {
+      problemCode: ConnectionProblemReportReason.RequestProcessingError,
+    })
```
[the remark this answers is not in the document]
nbAmit:
```ts
    let connectionJson = null
    try {
      connectionJson = await unpackAndVerifySignatureDecorator(message.connectionSig, this.wallet)
    } catch (error) {
      if(error instanceof AriesFrameworkError) {
        throw new ConnectionProblemReportError(error.message, {
          problemCode: ConnectionProblemReportReason.RequestProcessingError,
        })
      }
    }
```
Is it the right way to handle it in connection service? or any other suggestion @jakubkoci @TimoGlastra

PR #563, opened 2021-12-04
style(eslint): enforce strict type checking rules

### LINKED ISSUE #376: Enable typescript eslint rules that require type checking
add: "plugin:@typescript-eslint/recommended-requiring-type-checking"

TimoGlastra:
Enforces strict type checking rules

@blu3beri we were getting an enormous amount of errors due to how typescript was configured, removing that it reduced it to about ~900 errors. I then disabled some rules for test files as they're quite strict which is not needed for testing which reduced to about ~300 errors.

I have already fixes some of them, when you have the time maybe you can try to fix some more? Would be good to have this added.

Fixes #376


blu3beri:
Yeah will get at it next week somewhere. Good to add this ASAP :)

TimoGlastra:
Closing for now as too many merge conflicts. If we want to pick this up we can best do it in batches of rules

blu3beri:
> Closing for now as too many merge conflicts. If we want to pick this up we can best do it in batches of rules

I would like to get this in ASAP. I was thinking after 0.3.0 is merged in we take some time from big features and add some rules in. Doing both at the same time will be annoying for both and if rather do this sooner than later.

PR #566, opened 2021-12-07
feat: expose wallet API

jakubkoci:
As a developer, I want to have more granular control of the wallet so that I can decide what flow to run in my business logic according to the state of the wallet. For example, I want to run a different branch (UI screens) of my code if the wallet does not exist without creating the wallet immediately.

I moved the creation of master secret into a create method. It should happen only once. It was unnecessarily called during every open wallet operation.


packages/core/src/wallet/IndyWallet.ts
```diff
@@ -96,6 +94,20 @@ export class IndyWallet implements Wallet {
...
+      }
+
+      // We usually want to create master secret only once, therefore, we can to do so when creating a wallet.
+      await this.open(walletConfig)
+
+      // We need to open wallet before creating master secret because we need wallet handle here.
+      await this.createMasterSecret(this.handle, walletConfig.id)
+
+      // We opened wallet just to create master secret, we can close it now.
+      await this.close()
```
jakubkoci, line 110: We need to open the wallet to create a master secret. I think it's a good trade-off to do so in here. The create wallet method should be called only once (or at least fewer times) than the open wallet method. It's a question, if we should close the wallet to avoid user's confusion or assume that the developer will want to call `open` operation after `create` and keep it open to save some computer time.

packages/core/src/agent/Agent.ts
```diff
@@ -194,13 +194,13 @@ export class Agent {
...
-        await this.wallet.delete()
-      } else {
-        await this.wallet.close()
-      }
+      await this.wallet.close()
+    }
+    // delete wallet if it has been created
+    if (this.wallet.isCreated && deleteWallet) {
+      await this.wallet.delete()
     }
```
jakubkoci, line 204: There was a problem when I created a wallet but didn't call open then the shutdown method didn't delete the wallet.

packages/core/src/wallet/IndyWallet.ts
```diff
@@ -34,32 +28,36 @@ export class IndyWallet implements Wallet {
     this.indy = agentConfig.agentDependencies.indy
   }
 
+  public get isCreated() {
```
TimoGlastra, line 31: isCreated feels a bit misleading. I would assume this means whether the wallet was created in the current session, but if you call open it will also be set. 

Maybe we can call it something like `isProvisioned`. According to google the term provision means:

> the action of providing or supplying something for use.

Which is exactly what having the wallet config means: it has the needed info to use the wallet

packages/core/src/agent/Agent.ts
```diff
@@ -194,13 +194,13 @@ export class Agent {
...
     if (this.wallet.isInitialized) {
-      if (deleteWallet) {
-        await this.wallet.delete()
-      } else {
-        await this.wallet.close()
-      }
+      await this.wallet.close()
+    }
+    // delete wallet if it has been created
+    if (this.wallet.isCreated && deleteWallet) {
```
TimoGlastra, line 202: I think we should throw an error here, or update the config parameter to include that it will only delete if the wallet is provisioned.

Or maybe remove the option? It is mostly used for tests, but now that the wallet api is public we can just use that right?

TimoGlastra (approved):
Nice! I do think exposing the wallet api may expose some lower level methods that we don't necessarily want to make public, but I'm good with this change 👍

packages/core/src/wallet/IndyWallet.ts
```diff
@@ -171,31 +179,31 @@ export class IndyWallet implements Wallet {
...
-        'Can not delete wallet that is not initialized. Make sure to call initialize before deleting the wallet'
+        'Can not delete wallet that does not have wallet config set. Make sure to call create wallet before deleting the wallet'
       )
     }
 
-    this.logger.info(`Deleting wallet '${walletInfo.walletConfig.id}'`)
+    this.logger.info(`Deleting wallet '${this.walletConfig.id}'`)
 
-    await this.close()
+    if (this.walletHandle) {
```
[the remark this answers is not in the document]
jakubkoci: Good point. But this one is tricky.

As I'm thinking about it now, I don't think it's a good idea to just silently do nothing if the developer calls `close` and `walletHandle` is not set, because it can lead to situations when the developer accidentally sets `agent.wallet.walletHandle = undefined` but wallet will be still opened. It should actually throw an error when you try to call `close` directly and `walletHandle` is not set. TypeScript doesn't allow me to call only `await this.indy.closeWallet(this.walletHandle)`, it says:

```
Argument of type 'number | undefined' is not assignable to parameter of type 'number'.
  Type 'undefined' is not assignable to type 'number'.ts(2345)
```

Therefore, I should validate `this.walletHandle` and eventually throw an error before calling `await this.indy.closeWallet(this.walletHandle)`.

However, if I'm deleting a wallet that has not been opened yet, I don't want to call `close` on it because that could throw an error I described above. So, although I don't like this double-check (each time for a different reason), I think it's a more robust solution.

packages/core/src/wallet/IndyWallet.ts
jakubkoci: Agree. I also didn't like it so much.

packages/core/src/agent/Agent.ts
jakubkoci: I kind of like that `shutdown` method, also deleting a wallet if you need it. But you're right. It does not make sense in the "real" world outside the test. We need to delete the wallet manually, which is likely worth the effort.

TimoGlastra:
BTW, this is a breaking change now, so make sure to add the `BREAKING CHANGE` line to the squashed commit message

PR #573, opened 2021-12-11
build: add arm test ledger setup

TimoGlastra:
Ledger setup didn't fully work for ARM baed macs. This PR adds an extra dockerfile that can be used to set up the test ledger on ARM macs.


jakubkoci (approved):
Now, I can finally buy a new Macbook :) Good job 👍

PR #577, opened 2021-12-14
fix: support mediation for connectionless exchange

TimoGlastra:

### DIFF
--- a/packages/core/src/modules/routing/RecipientModule.ts
+++ b/packages/core/src/modules/routing/RecipientModule.ts
@@ -280,8 +280,8 @@ export class RecipientModule {
     const connection = await this.connectionService.findByInvitationKey(invitation.recipientKeys[0])
     if (!connection) {
       this.logger.debug('Mediation Connection does not exist, creating connection')
-      // null means we don't want to use the current mediator when connecting to another mediator
-      const routing = await this.mediationRecipientService.getRouting(null)
+      // We don't want to use the current default mediator when connecting to another mediator
+      const routing = await this.mediationRecipientService.getRouting({ useDefaultMediator: false })
 
       const invitationConnectionRecord = await this.connectionService.processInvitation(invitation, {
         autoAcceptConnection: true,
--- a/packages/core/src/modules/routing/RecipientModule.ts
+++ b/packages/core/src/modules/routing/RecipientModule.ts
@@ -280,7 +280,8 @@ export class RecipientModule {
     const connection = await this.connectionService.findByInvitationKey(invitation.recipientKeys[0])
     if (!connection) {
       this.logger.debug('Mediation Connection does not exist, creating connection')
-      const routing = await this.mediationRecipientService.getRouting()
+      // null means we don't want to use the current mediator when connecting to another mediator
+      const routing = await this.mediationRecipientService.getRouting(null)
 
       const invitationConnectionRecord = await this.connectionService.processInvitation(invitation, {
         autoAcceptConnection: true,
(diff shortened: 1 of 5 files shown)


blu3beri (approved):

packages/core/tests/connectionless-proofs.test.ts
```diff
@@ -1,15 +1,36 @@
...
+  getBaseConfig,
+  issueCredential,
+  makeConnection,
+  prepareForIssuance,
+  setupProofsTest,
+  waitForProofRecordSubject,
+} from './helpers'
 import testLogger from './logger'
 
+import { CredentialPreview } from '@aries-framework/core'
```
jakubkoci, line 32: Maybe we should stay consistent and import `CredentialPreview` with relative path?

packages/core/src/modules/routing/RecipientModule.ts
```diff
@@ -280,7 +280,8 @@ export class RecipientModule {
     const connection = await this.connectionService.findByInvitationKey(invitation.recipientKeys[0])
     if (!connection) {
       this.logger.debug('Mediation Connection does not exist, creating connection')
-      const routing = await this.mediationRecipientService.getRouting()
+      // null means we don't want to use the current mediator when connecting to another mediator
+      const routing = await this.mediationRecipientService.getRouting(null)
```
jakubkoci, line 284: Would your fix work if we preserved the previous API? I'm asking because I don't like passing null and having different behavior based on the difference between `null` ,`undefined` or value.

Although, I would welcome if we would not need to call `getDefaultMediator` before every `getRouting`, I don't think this is a good approach.

jakubkoci: It seems like your goal is not to get routing but just to create new keys. Maybe, we need to have 2 separate methods. 

1. creating did, keys, and endpoint not involving a mediator
2. creating did, keys, and endpoint involving a mediator 

The name `getRouting` is kind of misleading because it's actually not only about routing.

TimoGlastra: I agree on the `null` vs `undefined` point. That's a nasty fix from me.

I disagree however that the naming of `getRouting` is misleading as IMO mediation _is_ routing. The protocol used for forward messages is called the `routing` protocol (`https://didcomm.org/routing/1.0/forward`) which is what is used when you use a mediator that adds extra keys to the `routingKeys` array.

what if I add an optional parameter to the `getRouting`  method to specify whether the default mediator should be taken into account. It will be true by default, but can be set to false?

```ts
const routing = await this.mediationRecipientService.getRouting({
   mediationRecord: xxx,
   useDefaultMediator: true
})
```

jakubkoci: That would be better 👍 

As you wrote, mediation is routing. Then I would expect that routing returns things only related to routing/mediation. I was surprised by the following line in `createConnection` when I was looking at how we're creating a did and verkey for a new connection:

`const { endpoints, did, verkey, routingKeys, mediatorId } = options.routing`

I thought that `endpoints, did, verkey` are somehow related just to mediation. But we use it to create a connection even if there is no mediation in use.

Perhaps, it's not a big deal. We don't have to address that. I'm familiar with how it works now and can live with that :)

TimoGlastra: > I thought that endpoints, did, verkey are somehow related just to mediation. But we use it to create a connection even if there is no mediation in use.

It's a bit weird it lives in the mediation service, even thought it handled non-mediation stuff. I think the idea was to eventually move it to a routing service that handles everything related to routing (mediators, relays, other possible weird routing stuff). The connection service would then call the routing service whenever it needs keys for a connections.

TimoGlastra:
Could you give a final look @jakubkoci?

jakubkoci (approved):
Good to go 👍

Issue #548, opened 2021-11-24
Handle attachments in a generic way


jakubkoci:
We should add something to handle attachments in a generic way (ACA-Py has it). Data could be base64 or json (or others but the first two are most important)

Currently we assume base64 everywhere and use the `data.base64`, but it could also be in json form.

```ts
// new
const data = request.data.get() // could be either json or base64, but it doesn't matter

// old
const base64 = request.data.base64 // could be null
const data = base64decode(base64)

const data = request.data.json
```

Should be updated in `packages/core/src/decorators/attachment/Attachment.ts`

_Originally posted by @TimoGlastra in https://github.com/hyperledger/aries-framework-javascript/pull/531#discussion_r744883442_

jakubkoci:
That could be good, at least for base64 and json. However, if I look at Aries RFC 0017: Attachments it can also contain links. Should we call some HTTP request in that case or just return the link?

TimoGlastra:
@morrieinmaas and @annelein will be working on this

TimoGlastra:
Fixed

PR #578, opened 2021-12-14
feat: generic attachment handler

morrieinmaas:
Adds method getDataAsJson to Attachment data and therefore also to Attachment.data
This return the attachmnt data as JSON if .json or .base64 are present in attachmentdata.
Otherwise it throws an error.

Co-authored-by: annelein <anneleinvanreijen@gmail.com>

Signed-off-by: morrieinmaas <moritz@animo.id>


morrieinmaas:
closes #548

packages/core/src/decorators/attachment/Attachment.ts
```diff
@@ -45,6 +47,16 @@ export class AttachmentData {
     }
   }
 
+  public getDataAsJson() {
```
blu3beri, line 50: Would it be a good idea to add a generic here? This would allows us to get the typing from the return type in a nice way.

morrieinmaas: probably 😅 will have to have a quick read into how that works in TS as I'm a noob.

morrieinmaas: But not sure as there is no args (input)

packages/core/src/decorators/attachment/Attachment.ts
```diff
@@ -45,6 +47,16 @@ export class AttachmentData {
...
+    if (this.base64 !== null && typeof this.base64 === 'string') {
```
TimoGlastra, line 51: I would use just `this.base64` to also capture `base64` being undefined

```suggestion
if (this.base64 && typeof this.base64 === 'string') {
```

packages/core/src/decorators/attachment/Attachment.ts
```diff
@@ -45,6 +47,16 @@ export class AttachmentData {
...
+      return JsonEncoder.fromBase64(this.base64)
+    } else if (this.json !== null) {
```
TimoGlastra, line 53: same here

```suggestion
} else if (this.json) {
```

packages/core/src/decorators/attachment/Attachment.ts
blu3beri: not tested but this should work

```ts
public getDataAsJson<T>(): T {
return json as T
}
```

packages/core/src/decorators/attachment/Attachment.ts
```diff
@@ -45,6 +47,16 @@ export class AttachmentData {
...
+      return this.json
+    } else {
+      throw new AriesFrameworkError('No attachment data available.')
```
TimoGlastra, line 56: Maybve we can be a bit more explicit in the error message, as it could be that data is available, just not in the base64 or json fields

`No attachment data found in 'json' or 'base64' data fields.`

packages/core/src/modules/credentials/messages/IssueCredentialMessage.ts
```diff
@@ -48,13 +47,8 @@ export class IssueCredentialMessage extends AgentMessage {
...
     const attachment = this.credentialAttachments.find((attachment) => attachment.id === INDY_CREDENTIAL_ATTACHMENT_ID)
 
-    // Return null if attachment is not found
-    if (!attachment?.data?.base64) {
-      return null
-    }
-
     // Extract credential from attachment
-    const credentialJson = JsonEncoder.fromBase64(attachment.data.base64)
+    const credentialJson = attachment?.data ? attachment.data.getDataAsJson() : null
```
TimoGlastra, line 51: Wouldn't this work? (also in the other places)

```suggestion
const credentialJson = attachment?.data?.getDataAsJson()
```

TimoGlastra (changes requested):
Nice work @morrieinmaas and @annelein, have a few comments

packages/core/src/modules/credentials/messages/IssueCredentialMessage.ts
morrieinmaas: I'll check that

packages/core/src/decorators/attachment/Attachment.ts
```diff
@@ -45,6 +47,16 @@ export class AttachmentData {
     }
   }
 
+  public getDataAsJson<T>(): T {
+    if (this.base64 && typeof this.base64 === 'string') {
+      return JsonEncoder.fromBase64(this.base64)
```
blu3beri, line 52: This must also be casted to T`

packages/core/src/modules/credentials/messages/IssueCredentialMessage.ts
```diff
@@ -48,14 +47,9 @@ export class IssueCredentialMessage extends AgentMessage {
...
-    if (!attachment?.data?.base64) {
-      return null
-    }
-
     // Extract credential from attachment
-    const credentialJson = JsonEncoder.fromBase64(attachment.data.base64)
+    const credentialJson = attachment?.data?.getDataAsJson()
 
-    return credentialJson
+    return credentialJson as Cred
```
blu3beri, line 53: This casting removes the `null`. Is this intentional? If so, could you remove it from the function signature aswell?

packages/core/src/modules/credentials/messages/OfferCredentialMessage.ts
```diff
@@ -63,14 +62,9 @@ export class OfferCredentialMessage extends AgentMessage {
...
-    if (!attachment?.data?.base64) {
-      return null
-    }
-
     // Extract credential offer from attachment
-    const credentialOfferJson = JsonEncoder.fromBase64(attachment.data.base64)
+    const credentialOfferJson = attachment?.data?.getDataAsJson()
 
-    return credentialOfferJson
+    return credentialOfferJson as CredOffer
```
blu3beri, line 68: Same here. Seems to happen in a couple of places. If something changes, could you change it everywhere?

packages/core/src/modules/credentials/messages/IssueCredentialMessage.ts
morrieinmaas: hmm... yeah intention is that according the linter type of credentialJson is unknown if I leave out `as Cred` which 'not assignable to Cred'

packages/core/src/modules/credentials/messages/OfferCredentialMessage.ts
morrieinmaas: as above - added this to match the type. Maybe we can remove null from the fct signature  as getDataAsJson actually throws an error now if it can't provide data. Again, I'm pretty new to typescript. Thought this would result either in an error or undefined but not null the way it is now.  Any help appreciated

packages/core/src/modules/credentials/messages/IssueCredentialMessage.ts
morrieinmaas: basically that's why i had sth along the lines of  
```typescript
const credentialJson = attachment?.data ? attachment.data.getDataAsJson() : null 
```
in the first commit... Again greatful for any help

blu3beri: If we use optional chaining, `.?` we short circuit to return an `undefined`. This means that the value we want to return from the function could be undefined. 

This could happen if `attachment` or `attachment.data` does not exist. 

So if we want a 100% to return an object and not `null` from this function we must check if these two fields exist:

```ts
If (attachment?.data) {
  // some case handling
}
```

After this we 100% have `attachment.data ` to call `getDataAsJson()` on. 

If we do not want to return the object OR `null`, as the api is right now, we could go back to your ternary expression.

packages/core/src/modules/credentials/messages/OfferCredentialMessage.ts
morrieinmaas: or we use try catch ?

packages/core/src/modules/credentials/messages/IssueCredentialMessage.ts
morrieinmaas: or we use try...catch?

blu3beri (approved):

morrieinmaas:
@TimoGlastra do you have anymore feedback? Or can we merge this?

TimoGlastra (approved):
LGTM! :)

morrieinmaas:
@blu3beri @TimoGlastra merge please

PR #579, opened 2021-12-14
feat: support revocation notification messages

TheTreek:
Revocation notification support:
- Added revocation notification model for credential records (optional property)
- Added revocation notification message handling
- Added revocation notification received event
- Added support for upcoming revocation notification v2 messages

Revocation notification process:
- Listens for revocation notification message to be received
- Fetches matching credential record
- Adds a new property to record called revocationNotification
- Updates record and triggers revocation notification received event that has the credential record as the payload

`revocationRegistryId` and `credentialRevocationId` need to be saved to the credential record. We are storing them as properties of CredentialRecord, I know that storing them in metadata may also be an option. If we wanted to store them as metadata then some guidance will be needed.


packages/core/src/modules/credentials/__tests__/CredentialService.test.ts
```diff
@@ -1,3 +1,5 @@
+import type { RevocationNotificationReceivedEvent } from '..'
```
TimoGlastra, line 1: We usually try to avoid importing from only dots

packages/core/src/modules/credentials/__tests__/CredentialService.test.ts
```diff
@@ -22,6 +24,8 @@ import { CredentialState } from '../CredentialState'
 import { CredentialUtils } from '../CredentialUtils'
 import { CredentialProblemReportReason } from '../errors/CredentialProblemReportReason'
 import {
+  RevocationNotificationMessageV2,
```
TimoGlastra, line 27: Nit: with the issue cred v1/v2 work we prepend the version instead of append

packages/core/src/modules/credentials/repository/CredentialRecord.ts
```diff
@@ -26,6 +27,8 @@ export interface CredentialRecordProps {
   threadId: string
 
   credentialId?: string
+  revocationRegistryId?: string
```
TimoGlastra, line 30: We're in the process of making the credential record non indy specific. Maybe you can extend the metadata type for the indy credential and store it alongside the schemaId and credentialDefinitionId. This will save some migrations soon

packages/core/src/modules/credentials/services/CredentialService.ts
```diff
@@ -672,6 +672,15 @@ export class CredentialService {
       credentialDefinition,
       revocationRegistryDefinition,
     })
+
+    // If we have the rev_reg_id, then we also want to set the cred_rev_id, which only can be fetched by retrieving credential from indy
+    if (indyCredential.rev_reg_id) {
+      const credential = await this.indyHolderService.getCredential(credentialId)
+
+      credentialRecord.credentialRevocationId = credential.cred_rev_id?.toString()
```
TimoGlastra, line 680: Is cred_rev_id not tupe string already?

packages/core/src/modules/credentials/services/RevocationService.ts
```diff
@@ -0,0 +1,105 @@
...
+    const threadRegex = /(indy)::(.+)::(\d+)$/
+    const threadId = messageContext.message.issueThread
+    try {
+      const threadIdGroups = threadId.match(threadRegex)
+      if (threadIdGroups) {
+        const [, , revocationRegistryId, credentialRevocationId] = threadIdGroups
+        const comment = messageContext.message.comment
+        await this.processRevocationNotification(revocationRegistryId, credentialRevocationId, comment)
+      } else {
+        throw new AriesFrameworkError(
```
TimoGlastra, line 68: Should we throw a problem report error here to notify the other party?

packages/core/src/modules/credentials/services/RevocationService.ts
```diff
@@ -0,0 +1,105 @@
...
+   * {@link RevocationNotification} and store it in the corresponding {@link ConnectionRecord}
+   *
+   * @param messageContext message context of RevocationNotificationMessageV2
+   */
+  public async processRevocationNotificationV2(
+    messageContext: InboundMessageContext<RevocationNotificationMessageV2>
+  ): Promise<void> {
+    this.logger.info('Processing revocation notification v2', { message: messageContext.message })
+    // CredentialId = <revocation_registry_id>::<credential_revocation_id>
+    const credentialIdRegex = /(.*)::(\d+)$/
```
TimoGlastra, line 88: I see v1 included an indy: prefix while v2 doesn't. Why is this removed, it would make sense to me to keep it in to be able to detect the type of credential that is revoked?

packages/core/src/modules/credentials/repository/CredentialRecord.ts
```diff
@@ -91,6 +100,7 @@ export class CredentialRecord extends BaseRecord<DefaultCredentialTags, CustomCr
       this.credentialAttributes = props.credentialAttributes
       this.autoAcceptCredential = props.autoAcceptCredential
       this.linkedAttachments = props.linkedAttachments
+      this.revocationNotification = props.revocationNotification
```
TimoGlastra, line 103: Maybe we can make this object a bit more generic and not specifically about the revocation notification. In the future we could then just store revocation data in the object. E.g we fetch the revocation data ourselves from the ledger and store the result in the credentialrecord.

For the framework user this gives an easy place to find the revocation state whether it was fetched ourselves or informed by the issuer.

```json
{
    "date": "the date",
    "method": "notification" // or "fetched" or something (maybe "ledger", but that wouldn't work for revocation methods that are not anchored on a ledger),
    "comment": "comment only present for notification method"
}
```

What do you think?

TimoGlastra (changes requested):
This is really great @TheTreek :)

Left a few suggestions

packages/core/src/modules/credentials/services/RevocationService.ts
TheTreek: A v1 message looked like
```JSON
{
  "@type": "https://didcomm.org/revocation_notification/1.0/revoke",
  "@id": "<uuid-revocation-notification>",
  "~please_ack": ["RECEIPT","OUTCOME"],
  "thread_id": "<revocation_format>::<revocation_registry_id>::<credential_revocation_id>",
  "comment": "Some comment"
}
```
A v2 message will look like this
```JSON
{
  "@type": "https://didcomm.org/revocation_notification/2.0/revoke",
  "@id": "<uuid-revocation-notification>",
  "~please_ack": ["RECEIPT","OUTCOME"],
  "revocation_format": "<revocation_format>",
  "credential_id": "<revocation_registry_id>::<credential_revocation_id>",
  "comment": "Some comment"
}
```
So we will still be able to detect it from the revocation_format property.

packages/core/src/modules/credentials/services/CredentialService.ts
TheTreek: According to the [@types/indy-sdk](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/74b19cf32a1b2f10958f5729f798245afb1125f5/types/indy-sdk/index.d.ts#L455) it is a number

JamesKEbert: We probably need to test this--as the type definition may be incorrect. I know for sure that the revocation registry delta contains the cred_rev_id's in the revoked field are numbers, but I _think_ the cred_rev_id is a string when doing getCredential. 
@TheTreek if you could that'd be great :)

packages/core/src/modules/credentials/services/RevocationService.ts
JamesKEbert: If we don't recognize the format, I'd say yes. If we can't find the credential, then no, I think that's an internal AFJ issue & potentially shouldn't be exposed to the other agent.

packages/core/src/modules/credentials/services/RevocationService.ts
JamesKEbert: The format was moved in v2 from the thread_id field in v2 to a separate field to help with parsing/making it more explicit/clear.

packages/core/src/modules/credentials/repository/CredentialRecord.ts
JamesKEbert: > revocationRegistryId and credentialRevocationId need to be saved to the credential record. We are storing them as properties of CredentialRecord, I know that storing them in metadata may also be an option. If we wanted to store them as metadata then some guidance will be needed.

We were thinking the same thing ^ , but had some difficulty understanding the black magic happening here 😆 

Additionally there's a grossness in that when we can only get the cred_rev_id out of the credential in getCredential() and not when we're storing the credential, which makes it more challenging.

packages/core/src/modules/credentials/services/CredentialService.ts
TheTreek: Testing confirmed that it is a string

TheTreek: Created pr for it here: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/59079

packages/core/src/modules/credentials/repository/CredentialRecord.ts
TimoGlastra: Ah let me guide you. In `credentialMetadataTypes.ts` extend the  value of the `CredentialMetadataKeys.IndyCredential` key with the `revocationRegistryId` and the `credentialRevocationId`. 

You can now use it like this:

```ts
const indyCredential = credentialRecord.metadata.get(CredentialMetadataKeys.IndyCredential)

indyCredential?.revocationRegistryId

// .add adds to the existing metadata object, creating it if it doesn't exist yet
credentialRecord.metadata.add(CredentialMetadataKeys.IndyCredential, {
   revocationRegistryId: 'a rev id',
   credentialRevocationId: 'the value'
})
```

You should make sure to save the record afterwards

TimoGlastra: On the `cred_rev_id` only be present when retrieving the credential. It seems that ACA-Py extract the `cred_rev_id` from the received credential attachment: https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/protocols/issue_credential/v2_0/formats/indy/handler.py#L517

Could it be that our typing is not reflecting the actual object? Otherwise I don't see a problem with storing + fetching it directly afterwards. I think receiving a credential will be done a lot less than showing it for proof, so don't see the problem in one extra DB query. However indy has to extract it from somewhere, so it should be in the received `Cred` type right?

packages/core/src/modules/credentials/services/RevocationService.ts
TimoGlastra: I should have read the RFC before making a comment. Thanks for explaining!

packages/core/src/modules/credentials/repository/CredentialRecord.ts
TimoGlastra: Hmm looking at practical use cases what is often used for non-revocation intervals? Isn't the most often used question now:now? I think according to RFC 0441 we can only ask for the `from` being the same as `to` (which gives us an exact timestamp) or we get a single to which means basically the same as an exact timestamp right?

So we would know for sure that at a specific time the credential was revoked/not-revoked. I agree it may not be 100% applicable to every use case, but I think it could be really useful to show the revocation status of a credential next to it, and e.g. maybe showing a refresh button which will fetch the latest revocation state.

TimoGlastra:
Any updates on this @JamesKEbert? What are the blockers on this PR? Would be great to get this merged!

JamesKEbert:
> Any updates on this @JamesKEbert? What are the blockers on this PR? Would be great to get this merged!

This has been primarily waiting on me. The three remaining items/blockers here:
* The implemented version of the v1 implementation is not compliant with the RFC, but instead is compatible with ACA-Py. So my initial proposal would be to merge and note the difference from the RFC in the [Divergence from RFCs section](https://github.com/hyperledger/aries-framework-javascript#divergence-from-aries-rfcs) as well as in the RFC's implementations section. 
* I have a minor addition/tweak to make to the revocation-notification-v2 RFC to add a requirement to ensure the holder is verifying the validity of the issuer. 
* We may want to update this to use the metadata vs as a property on the credential.

I expect to be able to have this PR ready to merge within the next week.

packages/core/src/modules/credentials/repository/CredentialRecord.ts
JamesKEbert: Awesome, thank you for the insight on the metadata bit here--we'll let you know if we run into any other struggles there. 

Also, thank you for the insight into ACA-Py's implementation, we'll re-review if we can access the cred_rev_id in the credential attachment. 👍

packages/core/src/modules/credentials/repository/CredentialRecord.ts
JamesKEbert: You're correct in that what we're supporting right now based off of RFC 0441 is only supplying the same from and to, such as: now:now, 30 seconds ago/30 second ago, etc. 

Right now the ability to query the ledger for a rev reg delta (to get revocation status) is exposed on the LedgerModule, which would enable what you're describing, although it's maybe not as obvious to use for this case.

packages/core/src/modules/credentials/repository/CredentialRecord.ts
TheTreek: Hey @TimoGlastra, I was looking at adding the revocationRegistryId and credentialRevocationId to the credential metadata instead of the credential record when I ran into an issue. We need to be able to search for a credential based off of those two properties, I am not seeing a clear way to do that if they are only in the metadata. Any insight here would be appreciated.

TimoGlastra: Interesting. Maybe you could access the metadata from inside the `getTags` method?

This is an example with the (already existing) credentialDefinitionId. I think you could do the same for `credentialRevocationId`:

```ts
  public getTags() {
    const metadata = this.metadata.get(CredentialMetadataKeys.IndyCredential)

    return {
      ...this._tags,
      threadId: this.threadId,
      connectionId: this.connectionId,
      state: this.state,
      credentialId: this.credentialId,
      credentialDefinitionId: metadata?.credentialDefinitionId,
    }
  }
```

JamesKEbert:
Hey @TimoGlastra, everything's been addressed _I think_, besides the need to re-fetch the indy credential in order to get the credential revocation id, which we're looking at with our Python team. However, I think I'd be in favor of completing this and having that as a followup if it's a possibility.

TimoGlastra:
Sounds good @JamesKEbert, thanks! I'm fine with doing it in a follow up. However wouldn't that cause any issues with the tags and the credential not queryable?

Also CodeQL is complaining about the regexes. Maybe we can make them a bit more specific by looking at the ACA-Py regexes: https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/messaging/valid.py#L403-L436?

packages/core/src/modules/credentials/services/RevocationService.ts
```diff
@@ -0,0 +1,114 @@
...
+    this.logger.info('Processing revocation notification v1', { message: messageContext.message })
+    // ThreadID = indy::<revocation_registry_id>::<credential_revocation_id>
+    const threadRegex = /(indy)::(.+)::(\d+)$/
+    const threadId = messageContext.message.issueThread
+    try {
+      const threadIdGroups = threadId.match(threadRegex)
+      if (threadIdGroups) {
+        const [, , revocationRegistryId, credentialRevocationId] = threadIdGroups
+        const comment = messageContext.message.comment
+        const connection = messageContext.connection
```
TimoGlastra, line 73: If we do `const connection = messageContext.assertReadyConnection` we don't have to make the connection parameter optional in the `processRevocationNotification` method

packages/core/src/modules/credentials/services/RevocationService.ts
```diff
@@ -0,0 +1,114 @@
...
+    this.eventEmitter.emit<RevocationNotificationReceivedEvent>({
+      type: CredentialEventTypes.RevocationNotificationReceived,
+      payload: {
+        credentialRecord,
+      },
+    })
+  }
+
+  /**
+   * Process a recieved {@link RevocationNotificationMessagev1}. This will create a
```
TimoGlastra, line 56:
```suggestion
   * Process a recieved {@link V1RevocationNotificationMessage}. This will create a
```

packages/core/src/modules/credentials/services/RevocationService.ts
```diff
@@ -0,0 +1,114 @@
...
+   * {@link RevocationNotification} and store it in the corresponding {@link ConnectionRecord}
```
TimoGlastra, line 57:
```suggestion
   * {@link RevocationNotification} and store it in the corresponding {@link CredentialRecord}
```

packages/core/src/modules/credentials/services/CredentialService.ts
```diff
@@ -672,6 +672,20 @@ export class CredentialService {
       credentialDefinition,
       revocationRegistryDefinition,
     })
+
+    // If we have the rev_reg_id, then we also want to set the cred_rev_id, which only can be fetched by retrieving credential from indy
+    if (indyCredential.rev_reg_id) {
```
TimoGlastra, line 677: It seems you're re-fetching the credential here? I'm not sure what you mean with the comment on re-fetching in that case

TimoGlastra (approved):

packages/core/src/modules/credentials/repository/CredentialRecord.ts
```diff
@@ -43,6 +47,8 @@ export type DefaultCredentialTags = {
   connectionId?: string
   state: CredentialState
   credentialId?: string
+  revocationRegistryId?: string
+  credentialRevocationId?: string
```
TimoGlastra, line 51: Oh maybe we can call this `indyRevocationRegistryId` and `indyCredentialRevocationId` to not make it harder to use a generic name for non-indy revocation formats in the future? Or do you think this way we'll be able to reuse the name for other formats? In that case also fine, but not sure how applicable these parameters are to other revocation methods.

JamesKEbert: @TheTreek are these needed in the tags here since we've got this information in the metadata now?

packages/core/src/modules/credentials/services/CredentialService.ts
JamesKEbert: The point of the comment was to communicate why we were refetching the credential when we had just received it. It appears that the comment didn't quite communicate that 😆

packages/core/src/modules/credentials/repository/CredentialRecord.ts
TheTreek: >@TheTreek are these needed in the tags here since we've got this information in the metadata now?

These are needed in the tags so we can search for the credential by them. We don't have that ability if it is solely in metadata.

JamesKEbert:
> However wouldn't that cause any issues with the tags and the credential not queryable?

Since we're still re-fetching the credential it should be fine querying it as I understand it, so a followup would improve the performance as we wouldn't need to re-fetch it, but it wouldn't change any of the API or interactions. 

So far we've been stumped on finding a way to retrieve the cred_rev_id (and maybe we're just missing something obvious).

This is the log of a revocable credential offer that's been decoded, no sign of the cred_rev_id:
```
{
  "schema_id": "schema_id_here",
  "cred_def_id": "cred_def_id_here",
  "rev_reg_id": "rev_reg_id_here",
  "values": {
    "First Name": {
      "raw": "Patrick",
      "encoded": "19326420889721584150907756434744763515517200385607639812518004696795377894526"
    },
    "Age": {
      "raw": "27",
      "encoded": "27"
    },
    "Last Name": {
      "raw": "Kenyon",
      "encoded": "43826990635406025838022124681260409630854046085697036781736711625850312985001"
    }
  },
  "signature": {
    "p_credential": {
      "m_2": "7305319045799514268272120858355870310534978014147101007311454518968531958054",
      "a": "14844912702470868179521586926134181750894090773047541174695949535671215686561152817635530521869704307303241620428782997615160309203742555018212456152978218558221758570928590138933698774008719170411873022537960519441144504002906569198367737977937615785569935017718931523095844510994104217005719826748598404596206400802335265404233756620288226952475686634699683297431110507274700718547215878693135829962083762313871425803001646999209119309645813904842944499702484567760454564666147804073083555832116498460658343742999144644654083790859145459343410031643523956190170889716781177750508466903312732579813158445784265423377",
      "e": "259344723055062059907025491480697571938277889515152306249728583105665800713306759149981690559193987143012367913206299323899696942213235956742929938668687346298700943854241024183489",
      "v": "8387086101037151305185778556119457911798187212118149201882895432087746857116977574121346243552540856933316283870904265400051433189801970084732357652630551191280289814530004250646369779671677926876342065698678055020818715361339219851297869774270326070869379218860741080941219885760076650703657670342493007005020998285677196559419839635079282556340866370602592434668545433676191202501393312216725437181002777409778133006819702618382461451144768867892011851422649237232055425746841938384818894961237137930145139250225039028636834192319877970553202838304002448925108448011618294118164015842183517754309081270035248554288640329255836932244590610470857416778611921805021796007250433353379898135981810969089320238797076286558928530397515694944345969236224197190013996211670324736368085071986791310897512813454594932035015427059"
    },
    "r_credential": {
      "sigma": "1 01E93D930955B546C5E77DFA68A665293043A223AC1205E2D088A16FFFF68FD9 1 10F98B6EECA6AF330D31BC13EF9FBED03E3E043F8256D2B2D3B7F58270DE1EE8 2 095E45DDF417D05FB10933FFC63D474548B7FFFF7888802F07FFFFFF7D07A8A8",
      "c": "1E62DE3FC03A397FA5B0E5BA5C273E7DB9A9C1BBFCE46EE6D36925D2DA7BF920",
      "vr_prime_prime": "1C76C4D2174338F22E0383465D7ECFC242E99A2CB062AA66D3100AC9698C097C",
      "witness_signature": {
        "sigma_i": "1 07EA72F4064FF3F686886E0FAF4C6D2147714EFD13859E1980A96918CB3908F4 1 075CF56BC08547027C6E944360E404CC3F0E93126D80FAD75BBC097FFBA38F9C 1 1318F030FF17125E34C775B545F652917EBF94A8C52D547172AEF76B8DF1236F 1 06842DB53117BCA5D12E46C02947797D3B24CF18524A52B031A41D5F242AA155 2 095E45DDF417D05FB10933FFC63D474548B7FFFF7888802F07FFFFFF7D07A8A8 1 0000000000000000000000000000000000000000000000000000000000000000",
        "u_i": "1 1E072314806249C297F14C40D3E9A36BCF801845C2755EE0CEBA3E219F70BE79 1 215C68A9133392E4EE95D06B87E0527E8E91551EC25885C1CB44D3B045C2B0C2 1 0C3FEB9FA5812E3E827D78B1B7F585497362815E8F8738376206415701B57D68 1 0215EC379C9EC876ACF53704148A11EC2B3421EB9986A835D982E1C26E9077A2 2 095E45DDF417D05FB10933FFC63D474548B7FFFF7888802F07FFFFFF7D07A8A8 1 0000000000000000000000000000000000000000000000000000000000000000",
        "g_i": "1 22DA34344B797B365BC26C0AF63C967E6CA0E51689E5E2054B833BCD5DD22589 1 04ABB9AF319E46E47C3681F420179C2CE479B89CDCB1C71081A061E66F158631 2 095E45DDF417D05FB10933FFC63D474548B7FFFF7888802F07FFFFFF7D07A8A8"
      },
      "g_i": "1 22DA34344B797B365BC26C0AF63C967E6CA0E51689E5E2054B833BCD5DD22589 1 04ABB9AF319E46E47C3681F420179C2CE479B89CDCB1C71081A061E66F158631 2 095E45DDF417D05FB10933FFC63D474548B7FFFF7888802F07FFFFFF7D07A8A8",
      "i": 44,
      "m2": "1026A9FF5FC7537D0E218AD77501EB0ED3444ACB3F5D10FED1C3095EACE23926"
    }
  },
  "signature_correctness_proof": {
    "se": "6219587800254961292201550431492811823034436241314927998073286276305115636931901662751085349997880880391015795378287411770607013309694403951516553406518853087492054403788074106436923532055092885506047389459184816879685780357222857537219138520912880436522023752833052129901442047554691631307209798739098145843728774312475213640673865199775931882687457308171390891830840408194234886003046777737328890389667057844079428079471442022600389306320449269743101851712438940109680189127232597318314278397254826107750535280251908796135243962700353785986867870470261979597558713018700948536821357259132259131929237281745757292731",
    "c": "114316312050390703566577975573103219557354614093345274619034464414017402213123"
  },
  "rev_reg": {
    "accum": "21 125E18D45CB6A65AC183C3AEAD5193B253911ADF7D1A698DE405D22C305AE60C9 21 1414980198802934B7C14DE988957BA70BA9A40BA37B4E9B4FDCBC3BB5A4EDC88 6 64887278BD596A17F17430CED4CC972E07DBB40AB7D39CE2D33174E0F5F94A77 4 1B6F6588BBFD98D71E14ECEC21FBFD1E11C67299DA118B8C66CFE05FF2644298 6 7A67D9C53E2F7E5C3CD3B4543A52DF7E1DD799EB22A57B85DFA082BA57E13E18 4 21F7BDAEFB60DC00DBE77614B1DC0B65A78C82E273BC4FBBDC5A9795EAB20A8D"
  },
  "witness": {
    "omega": "21 1310F0EAFEFA7B9515E5F221090947B49A8C508779E6B2AFF09E7CB1EE5F4B064 21 135C3B62866A9226AF02884D7E3F64ED25C346B9C09D51377859464798E64C0F8 6 797DA355304B65AF9650F5B1DE3965B18880F720973A4137D5176EC2090BB4DE 4 3BA261017FC9AED454CC80501A3CD7FF4582B5F5F15F151828A5D9BE9ED387B6 6 8FB2082EA9A74E8CD4BAF67F977737E1D7B3531FB328EB3E15A860EBF48F3307 4 2BCB2E3BCDBEBB4D273629A2138AA88C666665896CADD72C0AA20E024B4EB6A3"
  }
}
```

packages/core/src/modules/credentials/repository/CredentialRecord.ts
JamesKEbert: Got it--I'd agree with Timo's suggestion to use `indyRevocationRegistryId` and `indyCredentialRevocationId` since this is somewhat specific to Indy. 

> Or do you think this way we'll be able to reuse the name for other formats

There might be overlap, however I don't think we have enough info on other methods right now to make an educated decision, so I think doing Indy specific names will give us the most flexibility in the future IMO.

TimoGlastra:
> Since we're still re-fetching the credential it should be fine querying it as I understand it,

Got it! I thought re-fetching the credential still needs to be added so that's why I was confused the code was already re-fetching it. But now I understand you meant to find a better alternative to re-fetching the credential. Makes sense.


If the tests pass and the regex is updated I'm good to merge this 👍 Thanks for the really great addition

TimoGlastra:
BTW -- it seems tests are failing due to the snapshots being updated. I think all changes are related to adding the indy metadata tags and look fine. You can update the snapshots by running `yarn test -u`

PR #580, opened 2021-12-14
feat: ledger connections happen on agent init in background

TheTreek:
Agent will try to connect to all ledger pools on initialization unless specifically told not to in `InitConfig.connectLedgersOnStart`.

Has a dependency on the next indy-sdk-react-native release https://github.com/hyperledger/indy-sdk-react-native/pull/24. May want to avoid merging until that has been made available.


packages/core/src/modules/ledger/IndyPool.ts
```diff
@@ -74,35 +75,39 @@ export class IndyPool {
   }
 
   public async connect() {
-    const poolName = this.poolConfig.id
-    const genesisPath = await this.getGenesisPath()
+    this.poolConnected = (async () => {
```
TimoGlastra, line 78: Maybe we can extract this to a private method, so we don't need a self executing function? 

I think that may look a bit cleaner

TimoGlastra: Also, maybe we should add a check in this method that we're not already connecting/connected and throw an error otherwise? Otherwise you can still call the method two times and it will try to open another connection

TimoGlastra (changes requested):
Nice @TheTreek! I left a few comments

TimoGlastra:
indy-sdk-react-native 0.1.16 has been released: https://github.com/hyperledger/indy-sdk-react-native/releases/tag/0.1.16

packages/core/src/modules/ledger/IndyPool.ts
TheTreek: Agreed, I added a check in the method that will return an error if `poolConnected` is defined. I also added a catch to `poolConnected` that will set `poolConnected` to undefined so we can retry the connection when an error occurs.

```TypeScript
  public async connect() {
    if (!this.poolConnected) {
      // Save the promise of connectToLedger to determine if we are done connecting
      this.poolConnected = this.connectToLedger()
      this.poolConnected.catch((error) => {
        // Set poolConnected to undefined so we can retry connection upon failure
        this.poolConnected = undefined
        this.logger.error('Connection to pool: ' + this.poolConfig.genesisPath + ' failed.', { error })
      })
      return this.poolConnected
    } else {
      throw new AriesFrameworkError('Cannot attempt connection to ledger, already connecting.')
    }
  }
```

packages/core/src/modules/ledger/LedgerModule.ts
```diff
@@ -19,6 +19,13 @@ export class LedgerModule {
     this.wallet = wallet
   }
 
+  /**
+   * Connect to all the ledger pools
+   */
+  public async connectToGenesis() {
+    await this.ledgerService.connectToPools()
+  }
```
TimoGlastra, line 27: Maybe we should also call this `connectToPools`? That would be consistent with other method names and makes more sense than `connectToGenesis` to me.


```suggestion
  public async connectToPools() {
    await this.ledgerService.connectToPools()
  }
```

TimoGlastra (commented):
This LGTM @TheTreek. One final method name change for consistency.

packages/core/src/modules/ledger/LedgerModule.ts
JamesKEbert: +1

TimoGlastra (approved):
LGTM 👍

TimoGlastra:
@TheTreek the branch is out of date, can you update?

TheTreek:
@TimoGlastra Okay, it's been updated.

TimoGlastra:
Almost forgot, but can you also update to the latest indy-sdk-react-native version? Otherwise this wouldn't work right?

PR #581, opened 2021-12-15
refactor(core): separate logic for inbound plaintext and encrypted message

jakubkoci:
Currently, we use `receiveMessage` for both encrypted and plaintext connection-less OOB messages. However, the plaintext message just passed through a lot of code related only to the encrypted message. I split the logic into separate methods.

I also did a few renamings according to [DIDComm spec](https://identity.foundation/didcomm-messaging/spec/) introducing `EncryptedMessage` and `PlaintextMessage`.


packages/core/src/agent/MessageReceiver.ts
```diff
@@ -59,21 +59,33 @@ export class MessageReceiver {
...
+    if (typeof inboundMessage !== 'object' || inboundMessage == null) {
       throw new AriesFrameworkError('Invalid message received. Message should be object')
     }
 
     this.logger.debug(`Agent ${this.config.label} received message`)
 
-    const unpackedMessage = await this.unpackMessage(inboundPackedMessage as WireMessage)
-    const senderKey = unpackedMessage.senderVerkey
-    const recipientKey = unpackedMessage.recipientVerkey
+    if (this.isPlaintextMessage(inboundMessage as Record<string, unknown>)) {
```
TimoGlastra, line 74: if you make the input to `isPlaintextMessage` unknown and do some additional checks in the method, we don't need to cast 

```suggestion
    if (this.isPlaintextMessage(inboundMessage)) {
```

packages/core/src/agent/MessageReceiver.ts
```diff
@@ -59,21 +59,33 @@ export class MessageReceiver {
...
+      await this.receivePlaintextMessage(inboundMessage as PlaintextMessage)
```
TimoGlastra, line 75: This shouldn't be needed as the `isPlaintextMessage` return `message is PlaintextMessage`

```suggestion
      await this.receivePlaintextMessage(inboundMessage)
```

packages/core/src/agent/MessageReceiver.ts
```diff
@@ -59,21 +59,33 @@ export class MessageReceiver {
...
+    } else {
+      await this.receiveEncryptedMessage(inboundMessage as EncryptedMessage, session)
+    }
+  }
+
+  private async receivePlaintextMessage(plaintextMessage: PlaintextMessage) {
+    const message = await this.transformAndValidate(plaintextMessage, null)
```
TimoGlastra, line 82: I'm not a big fan of required `null` parameters. Maybe we can make it an optional parameter?

packages/core/src/agent/TransportService.ts
```diff
@@ -64,5 +64,5 @@ export interface TransportSession {
   keys?: EnvelopeKeys
   inboundMessage?: AgentMessage
   connection?: ConnectionRecord
-  send(wireMessage: WireMessage): Promise<void>
+  send(wireMessage: EncryptedMessage): Promise<void>
```
TimoGlastra, line 67:
```suggestion
  send(encryptedMessage: EncryptedMessage): Promise<void>
```

packages/core/src/agent/__tests__/MessageSender.test.ts
```diff
@@ -41,7 +41,7 @@ class DummyOutboundTransport implements OutboundTransport {
 describe('MessageSender', () => {
   const EnvelopeService = <jest.Mock<EnvelopeServiceImpl>>(<unknown>EnvelopeServiceImpl)
 
-  const wireMessage: WireMessage = {
+  const wireMessage: EncryptedMessage = {
```
TimoGlastra, line 44:
```suggestion
  const encryptedMessage: EncryptedMessage = {
```

packages/node/src/transport/HttpInboundTransport.ts
```diff
@@ -75,7 +75,7 @@ export class HttpTransportSession implements TransportSession {
     this.res = res
   }
 
-  public async send(wireMessage: WireMessage): Promise<void> {
+  public async send(wireMessage: EncryptedMessage): Promise<void> {
```
TimoGlastra, line 78:
```suggestion
  public async send(encryptedMessage: EncryptedMessage): Promise<void> {
```

packages/node/src/transport/WsInboundTransport.ts
```diff
@@ -81,7 +81,7 @@ export class WebSocketTransportSession implements TransportSession {
     this.socket = socket
   }
 
-  public async send(wireMessage: WireMessage): Promise<void> {
+  public async send(wireMessage: EncryptedMessage): Promise<void> {
```
TimoGlastra, line 84:
```suggestion
  public async send(encryptedMessage: EncryptedMessage): Promise<void> {
```

tests/transport/SubjectInboundTransport.ts
```diff
@@ -52,7 +52,7 @@ export class SubjectTransportSession implements TransportSession {
     this.replySubject = replySubject
   }
 
-  public async send(wireMessage: WireMessage): Promise<void> {
+  public async send(wireMessage: EncryptedMessage): Promise<void> {
```
TimoGlastra, line 55:
```suggestion
  public async send(encryptedMessage: EncryptedMessage): Promise<void> {
```

packages/core/src/agent/MessageReceiver.ts
```diff
@@ -131,53 +127,65 @@ export class MessageReceiver {
       this.transportService.saveSession(session)
     }
 
+    const messageContext = new InboundMessageContext(message, {
+      // Only make the connection available in message context if the connection is ready
+      // To prevent unwanted usage of unready connections. Connections can still be retrieved from
+      // Storage if the specific protocol allows an unready connection to be used.
+      connection: connection?.isReady ? connection : undefined,
+      senderVerkey: senderKey,
+      recipientVerkey: recipientKey,
```
TimoGlastra, line 136:
```suggestion
      senderKey,
      recipientKey,
```

I see you updated verkey to key in a lot of places, maybe would be good to also remove the reference to verkey here.

However we do lose some context here, as a verkey is a base58 encoded public key (probably ed25519) while key is more general and doesn't specify the encoding of the key.

For now lets use `senderKey` etc but it may be good to use something like `senderKeyBase58`

TimoGlastra (approved):
Nice 👍

packages/core/src/agent/MessageReceiver.ts
jakubkoci: That's interesting... Do you mean that the sender key differs from the recipient keys? I see that `Wallet`, for example, contains `recipientKeys: string[], senderVerkey?: string`. Is the naming on purpose to distinguish different formats?

jakubkoci: Ah, I just realized that it should be the same because we're using the sender key as recipient and vice-versa. I think if it's currently the same format across the framework we can name it as `xxxKey`. But you have a good point, it can be eventually a different key format and we should distinguish between them.

packages/core/src/agent/MessageReceiver.ts
jakubkoci: Good point. Fixed.

packages/core/src/agent/MessageReceiver.ts
jakubkoci: I needed to move the validation from the first line of the method inside `isPlaintextMessage`. It works now also without `as PlaintextMessage`

PR #584, opened 2021-12-21
refactor: resolve feedback for problem report

nbAmit:
Resolve the feedback #560 

Signed-off-by: Amit Padmani <amit@northernblock.io>

### DIFF
--- a/packages/core/src/modules/connections/services/ConnectionService.ts
+++ b/packages/core/src/modules/connections/services/ConnectionService.ts
@@ -426,7 +426,7 @@ export class ConnectionService {
   public async processProblemReport(
     messageContext: InboundMessageContext<ConnectionProblemReportMessage>
   ): Promise<ConnectionRecord> {
-    const { message: connectionProblemReportMessage, recipientVerkey } = messageContext
+    const { message: connectionProblemReportMessage, recipientVerkey, senderVerkey } = messageContext
 
     this.logger.debug(`Processing connection problem report for verkey ${recipientVerkey}`)
 
@@ -440,6 +440,10 @@ export class ConnectionService {
       throw new AriesFrameworkError(
         `Unable to process connection problem report: connection for verkey ${recipientVerkey} not found`
       )
+    }
+
+    if (connectionRecord.theirKey && connectionRecord.theirKey !== senderVerkey) {
+      throw new AriesFrameworkError("Sender verkey doesn't match verkey of connection record")
     }
 
     connectionRecord.errorMessage = `${connectionProblemReportMessage.description.code} : ${connectionProblemReportMessage.description.en}`
--- a/packages/core/src/modules/connections/services/ConnectionService.ts
+++ b/packages/core/src/modules/connections/services/ConnectionService.ts
@@ -443,7 +443,7 @@ export class ConnectionService {
     }
 
     connectionRecord.errorMessage = `${connectionProblemReportMessage.description.code} : ${connectionProblemReportMessage.description.en}`
-    await this.updateState(connectionRecord, connectionRecord.state)
+    await this.update(connectionRecord)
     return connectionRecord
   }
 
@@ -536,6 +536,10 @@ export class ConnectionService {
         previousState,
       },
     })
+  }
+
+  public update(connectionRecord: ConnectionRecord) {
+    return this.connectionRepository.update(connectionRecord)
   }
 
   /**
--- a/packages/core/src/modules/connections/services/ConnectionService.ts
+++ b/packages/core/src/modules/connections/services/ConnectionService.ts
@@ -323,7 +323,16 @@ export class ConnectionService {
     connectionRecord.assertState(ConnectionState.Requested)
     connectionRecord.assertRole(ConnectionRole.Invitee)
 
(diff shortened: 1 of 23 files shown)


packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -433,8 +442,8 @@ export class ConnectionService {
       )
     }
 
-    connectionRecord.errorMsg = `${connectionProblemReportMessage.description.code} : ${connectionProblemReportMessage.description.en}`
-    await this.updateState(connectionRecord, ConnectionState.None)
+    connectionRecord.errorMessage = `${connectionProblemReportMessage.description.code} : ${connectionProblemReportMessage.description.en}`
+    await this.updateState(connectionRecord, connectionRecord.state)
```
TimoGlastra, line 446: The state won't change, so I think we should just call the connection repository with update directly as this will emit a state changed event.

TimoGlastra: Also, with the `None` state being removed, there is no way to determine that you received a problem report, so we should probably look into adding some events related to problem reports in follow up PRs

packages/core/src/modules/credentials/services/CredentialService.ts
```diff
@@ -743,8 +745,8 @@ export class CredentialService {
     )
 
     // Update record
-    credentialRecord.errorMsg = `${credentialProblemReportMessage.description.code}: ${credentialProblemReportMessage.description.en}`
-    await this.updateState(credentialRecord, CredentialState.None)
+    credentialRecord.errorMessage = `${credentialProblemReportMessage.description.code}: ${credentialProblemReportMessage.description.en}`
+    await this.updateState(credentialRecord, credentialRecord.state)
```
TimoGlastra, line 749: Same here

packages/core/tests/helpers.ts
```diff
@@ -93,6 +93,7 @@ export async function waitForProofRecord(
     state?: ProofState
     previousState?: ProofState | null
     timeoutMs?: number
+    errorMessage?: string
```
TimoGlastra, line 96: You also need to add a `filter` to the `waitForProofRecordSubject` and `waitForCredentialRecordSubject` methods for the `errorMessage` param to be taken into account

TimoGlastra (changes requested):
Nice! Really great improvements :)

Few nitpicks

packages/core/tests/helpers.ts
nbAmit: Sorry @TimoGlastra, I didn't get you.
Can you please explain in detail?

TimoGlastra: Sorry, let me rephrase.

Adding the `errorMessage` to the parameters of `waitForProofRecord` doesn't add `errorMessage` as a filter to the `waitForProofRecord` by itself. So passing `errorMessage` to the `waitForProofRecord` (as you do in the tests) doesn't do anything.

You would need to add an extra line to `waitForProofRecordSubject`:

```ts
.filter((e) => errorMessage === undefined || e.payload.proofRecord.errorMessage === errorMessage),
```

same for credentials

packages/core/src/modules/connections/services/ConnectionService.ts
nbAmit: Hi @TimoGlastra, If we update the `updateState` with an `update`, it will be difficult to capture the event in the e2e test case?

https://github.com/hyperledger/aries-framework-javascript/blob/1041aa77bdc1c441d99e7852d4062766520d6597/packages/core/tests/credentials.test.ts#L483

packages/core/src/modules/credentials/services/CredentialService.ts
nbAmit: Hi @TimoGlastra, If we update the `updateState` with an `update`, it will be difficult to capture the event in the e2e test case?

https://github.com/hyperledger/aries-framework-javascript/blob/1041aa77bdc1c441d99e7852d4062766520d6597/packages/core/tests/credentials.test.ts#L483

TimoGlastra: Yes, see my comment above:

> Also, with the None state being removed, there is no way to determine that you received a problem report, so we should probably look into adding some events related to problem reports in follow up PRs


We should probably add a new event (e.g. `ConnectionProblemReportReceived`, `CredentialProblemReportReceived` etc...) where the event looks something like:

```ts
{
   payload: {
     credentialRecord: { /* the record */ },
     problemReportMessage: { /* the problem report message */ }
}
}
```

TimoGlastra: You can either add it to the current PR, or you can bypass it by listening for the `AgentMessageProcessedEvent` for now and filtering on the problem report @type

nbAmit: @TimoGlastra , As of now I have removed the test case and I will add it into upcoming PR with the problem-report events

TimoGlastra (approved):
LGTM @nbAmit

PR #585, opened 2021-12-22
fix(core)!: Improved typing on metadata api

blu3beri:
BREAKING CHANGE: removed the `getAll()` function.

- Improved typings on the metadata functions. 
- Hidden ugly `_internal/...` and replaced with an enum
- Record metadata stays the same.

Would be nice if we can get this merged before 0.1.0 as it makes the public api quite a bit better

### DIFF
--- a/packages/core/src/modules/credentials/__tests__/CredentialRecord.test.ts
+++ b/packages/core/src/modules/credentials/__tests__/CredentialRecord.test.ts
@@ -1,8 +1,7 @@
 import { CredentialState } from '../CredentialState'
 import { CredentialPreviewAttribute } from '../messages'
 import { CredentialRecord } from '../repository/CredentialRecord'
-
-import { CredentialMetadataKeys } from '@aries-framework/core'
+import { CredentialMetadataKeys } from '../repository/credentialMetadataTypes'
 
 describe('CredentialRecord', () => {
   describe('getCredentialInfo()', () => {
--- a/packages/core/src/modules/credentials/__tests__/CredentialRecord.test.ts
+++ b/packages/core/src/modules/credentials/__tests__/CredentialRecord.test.ts
@@ -2,6 +2,8 @@ import { CredentialState } from '../CredentialState'
 import { CredentialPreviewAttribute } from '../messages'
 import { CredentialRecord } from '../repository/CredentialRecord'
 
+import { CredentialMetadataKeys } from '@aries-framework/core'
+
 describe('CredentialRecord', () => {
   describe('getCredentialInfo()', () => {
     test('creates credential info object from credential record data', () => {
@@ -17,7 +19,7 @@ describe('CredentialRecord', () => {
         ],
       })
 
-      credentialRecord.metadata.set('_internal/indyCredential', {
+      credentialRecord.metadata.set(CredentialMetadataKeys.IndyCredential, {
         credentialDefinitionId: 'Th7MpTaRZVRYnPiabds81Y:3:CL:17:TAG',
         schemaId: 'TL1EaPFCZ8Si5aUrqScBDt:2:test-schema-1599055118161:1.0',
       })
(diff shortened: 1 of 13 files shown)


packages/core/src/modules/credentials/__tests__/CredentialService.test.ts
```diff
@@ -38,6 +38,8 @@ import { CredentialService } from '../services'
 import { CredentialProblemReportMessage } from './../messages/CredentialProblemReportMessage'
 import { credDef, credOffer, credReq } from './fixtures'
 
+import { CredentialMetadataKeys } from '@aries-framework/core'
```
TimoGlastra, line 41: Is the package import (instead of relative) on purpose?

```suggestion
import { CredentialMetadataKeys } from '@aries-framework/core'
```

packages/core/src/modules/credentials/__tests__/CredentialRecord.test.ts
```diff
@@ -2,6 +2,8 @@ import { CredentialState } from '../CredentialState'
 import { CredentialPreviewAttribute } from '../messages'
 import { CredentialRecord } from '../repository/CredentialRecord'
 
+import { CredentialMetadataKeys } from '@aries-framework/core'
```
TimoGlastra, line 5: Same here

packages/core/src/modules/credentials/repository/credentialMetadataTypes.ts
```diff
@@ -0,0 +1,12 @@
...
+  IndyCredential = '_internal/indyCredential',
+  IndyRequest = '_internal/indyRequest',
+}
+
+export type CredentialMetadata = {
+  [CredentialMetadataKeys.IndyCredential]: {
+    schemaId?: string
+    credentialDefinitionId?: string
+  }
+  [CredentialMetadataKeys.IndyRequest]: Record<string, unknown>
```
TimoGlastra, line 11: Would still be good to use the indy cred req type, even though it is `Record<string, unknown>`
```suggestion
  [CredentialMetadataKeys.IndyRequest]: CredReqMetadata
```

packages/core/src/modules/credentials/services/CredentialService.ts
```diff
@@ -109,9 +109,9 @@ export class CredentialService {
     })
 
     // Set the metadata
-    credentialRecord.metadata.set('_internal/indyCredential', {
+    credentialRecord.metadata.set(CredentialMetadataKeys.IndyCredential, {
       schemaId: options.schemaId,
-      credentialDefinintionId: options.credentialDefinitionId,
+      credentialDefinitionId: options.credentialDefinitionId,
```
TimoGlastra, line 114: Thank god for typing :)

When was this typo introduced? Can it cause problems?

packages/core/src/storage/Metadata.ts
```diff
@@ -3,22 +3,17 @@ export type MetadataBase = {
...
- * update the record in the correct repository
+ * note: To add persistence to these records, you have to update the record in the correct repository
  *
  * @example
  *
- * ```ts
- * connectionRecord.metadata.set('foo', { bar: 'baz' })
- * connectionRepository.update(connectionRecord)
- * ```
+ * ```ts connectionRecord.metadata.set('foo', { bar: 'baz' }) connectionRepository.update(connectionRecord) ```
```
TimoGlastra, line 14: weird formating

TimoGlastra (changes requested):

packages/core/src/modules/credentials/services/CredentialService.ts
blu3beri: This can cause a problem if the metadata is set at the `createProposal` function and was never overwritten. I think the chances of that, and the metadata still being used, is slim to none.

TimoGlastra (approved):
LGTM! 👍

Issue #583, opened 2021-12-16
Allow enabling auto-accept when accepting connection invitation


jakubkoci:
I found the following `TODO` in `ConnectionResponseHandler`:

```
// TODO: should we only send ping message in case of autoAcceptConnection or always?
// In AATH we have a separate step to send the ping. So for now we'll only do it
// if auto accept is enable
```

I would expect that the connection will be completed after I call `acceptInvitation`. Maybe, there is a reason why we have a separate step in AATH. But, what if we allow to update `autoAcceptConnections` via a config parameter in `acceptInvitation`? Or, maybe, we should set `autoAcceptConnections` automatically when user calls `acceptInvitation`? 🤷‍♂️

jakubkoci:
To put more context... I'm trying to make the follwing test pass: [make a connection based on OOB invitation and process requests after the acceptation](https://github.com/hyperledger/aries-framework-javascript/pull/531/files#diff-261e45361be29b0d7c6678c64e2586f7c34cb9ab2e684e34727aa55fa06ed6c2R270), where I call `receiveInvitation` and then `acceptInvitation` expecting that after `acceptInvitation` is called, the connection will be finished and oob message request processed.

TimoGlastra:
The idea was that you set `autoAccept` when calling `createConnection` or `acceptInvitation` but I think allowing for `autoAcceptConnection` in `acceptInvitation` makes sense. Automatically setting auto accept when calling `acceptInvitation` feels like a bigger change as it changes how AFJ works. We then reduce the connection protocol a 2-step exchange from the POV of a framework user, which we haven't done with any of the other protocols. This is exactly where the agent auto accept config can be used for, to make it feel like connections only needes create invitation / accept invitation

jakubkoci:
Fixed by #589

PR #589, opened 2021-12-28
feat(core): allow to set auto accept connetion exchange when accepting invitation

jakubkoci:
Related to #583 

I'm already using these changes in OOB implementation #531 but I wanted to create separate PR.

### DIFF
--- a/packages/core/src/modules/connections/services/ConnectionService.ts
+++ b/packages/core/src/modules/connections/services/ConnectionService.ts
@@ -160,23 +160,29 @@ export class ConnectionService {
    */
   public async createRequest(
     connectionId: string,
-    config?: {
+    config: {
       myLabel?: string
       myImageUrl?: string
       autoAcceptConnection?: boolean
-    }
+    } = {}
   ): Promise<ConnectionProtocolMsgReturnType<ConnectionRequestMessage>> {
     const connectionRecord = await this.connectionRepository.getById(connectionId)
 
     connectionRecord.assertState(ConnectionState.Invited)
     connectionRecord.assertRole(ConnectionRole.Invitee)
 
+    const { myLabel, myImageUrl, autoAcceptConnection } = config
+
     const connectionRequest = new ConnectionRequestMessage({
-      label: config?.myLabel ?? this.config.label,
+      label: myLabel ?? this.config.label,
       did: connectionRecord.did,
       didDoc: connectionRecord.didDoc,
-      imageUrl: config?.myImageUrl ?? this.config.connectionImageUrl,
+      imageUrl: myImageUrl ?? this.config.connectionImageUrl,
     })
+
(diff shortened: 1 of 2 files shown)


packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -177,6 +178,7 @@ export class ConnectionService {
       imageUrl: config?.myImageUrl ?? this.config.connectionImageUrl,
     })
 
+    connectionRecord.autoAcceptConnection = config?.autoAcceptConnection
```
TimoGlastra, line 181: I think this could overwrite the previously set autoAcceptConnection state.

If receiveInvitation had autoAccept set to false and you don't provide it for acceptInvitation, it will be set to undefined.

If the global config is set to true, having false and undefined can have different meaning.

The problem then arises how do you clear autoAccept (null value?), but I think that's another issue to solve

jakubkoci: I see. What about adding the following condition?

```ts
if (config?.autoAcceptConnection !== undefined || config?.autoAcceptConnection !== null) {
  connectionRecord.autoAcceptConnection = config?.autoAcceptConnection
}
```

That wouldn't still solve clearing `autoAccept` but I could be good enough for now.

TimoGlastra: Yeah that looks good to me. I think just `!== undefined` should be fine, as `null` is not an allowed value

jakubkoci: Better safe than sorry 😄  Actually, I think `null` can appear there. What if the user of the framework doesn't use TS?

jakubkoci:
Updated

packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -163,20 +163,28 @@ export class ConnectionService {
...
       myImageUrl?: string
+      autoAcceptConnection?: boolean
     }
   ): Promise<ConnectionProtocolMsgReturnType<ConnectionRequestMessage>> {
     const connectionRecord = await this.connectionRepository.getById(connectionId)
 
     connectionRecord.assertState(ConnectionState.Invited)
     connectionRecord.assertRole(ConnectionRole.Invitee)
 
+    const { myLabel, myImageUrl, autoAcceptConnection } = config ?? {}
```
TimoGlastra, line 174: Nit: I think it looks cleaner to do the default empty method assignment in the method parameters

TimoGlastra (approved):

packages/core/src/modules/connections/services/ConnectionService.ts
jakubkoci: Good tip 👍 Let's try that

PR #590, opened 2021-12-28
docs(apple-silicon): update OpenSSL instructions

karimStekelenburg:
When following the instructions for installing indy for an Apple Silicon (M1 Pro) system, I found that the `brew install` command didn't work. After looking into how to install OpenSSL for M1, I found that the OpenSSL v1.1 on Homebrew also works for Apple Silicon chips (see: https://formulae.brew.sh/formula/openssl@1.1).

After following the rest of the instructions I ran the `is-indy-installed` test script, which resulted in `Libindy was installed correctly`. As a double check I have also installed `@aries-framework/node`, again with success.

> NOTE: When I ran `brew install openssl@1.1` it told me v1.1 was already on my system. I am unsure if this is because stuff I have installed prior to this attempt, or if this is because it's natively shipped with macOs. Therefore I have included the install command in the instructions, just to be sure. 

﻿Signed-off-by: Karim Stekelenburg <karim@animo.id>

### DIFF
--- a/docs/libindy/macos-apple.md
+++ b/docs/libindy/macos-apple.md
@@ -15,6 +15,11 @@ The first thing we'll do is install OpenSSL. Since Apple replaced OpenSSL with t
 brew install openssl@1.1 # possibly already installed on your system
 
 brew link openssl@1.1 --force
+
+echo 'export PATH="/opt/homebrew/opt/openssl@1.1/bin:$PATH"' >> ~/.zshrc
+
+source ~/.zshrc
+
 ```
 To double-check if the correct version is installed, you need to restart your terminal session and run the following command:
 
--- a/docs/libindy/macos-apple.md
+++ b/docs/libindy/macos-apple.md
@@ -21,6 +21,7 @@ echo 'export PATH="/opt/homebrew/opt/openssl@1.1/bin:$PATH"' >> ~/.zshrc
 source ~/.zshrc
 
 ```
+
 To double-check if the correct version is installed, you need to restart your terminal session and run the following command:
 
 ```sh
--- a/docs/libindy/macos-apple.md
+++ b/docs/libindy/macos-apple.md
@@ -12,22 +12,16 @@
 The first thing we'll do is install OpenSSL. Since Apple replaced OpenSSL with their own version of LibreSSL, we'll need to install it. Also, we need to install a specific version of OpenSSL that is compatible with Apples architecture. After the installation, we need to link it, so that it overrides the default openssl command (we have not noticed any issues with overriding this command, but be cautious).
 
 ```sh
-curl https://raw.githubusercontent.com/rbenv/homebrew-tap/e472b7861b49cc082d1db0f66f265368da107589/Formula/openssl%401.0.rb -o openssl@1.0.rb
+brew install openssl@1.1 # possibly already installed on your system
 
-brew install ./openssl@1.0.rb
-
-rm -rf ./openssl@1.0.rb
-
-brew link openssl@1.0 --force
+brew link openssl@1.1 --force
 ```
-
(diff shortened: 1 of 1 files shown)


blu3beri (approved):
Nice!

karimStekelenburg:
If anyone has any suggestions as to where I can find if OpenSSL v1.1 is natively installed on macOs or not, please shout! I also forgot to add the commands to add it to the PATH (just added them). I'm not sure if echoing to `.zshrc` is the proper way to do this (this was suggested by Homebrew).

blu3beri:
> If anyone has any suggestions as to where I can find if OpenSSL v1.1 is natively installed on macOs or not, please shout! I also forgot to add the commands to add it to the PATH (just added them). I'm not sure if echoing to `.zshrc` is the proper way to do this (this was suggested by Homebrew).

Echoing to `.zshrc` is the standard way to do it. However I do not understand why this is needed. `brew link --force` should override it.

Openssl@1.1 that is needed by rust is not the one supplied by Apple. Apple has their own fork of LibreSSL and is not compatible with `openssl` in rust.

karimStekelenburg:
Hmm, I'm not sure either. Here is the output, maybe I'm missing something...
CleanShot 2021-12-28 at 19 06 01@2x

blu3beri:
Hmm, alight yeah it seems that this is the way to do it. Not too bad to be fair.

docs/libindy/macos-apple.md
```diff
@@ -12,22 +12,22 @@
...
 
 ```sh
-curl https://raw.githubusercontent.com/rbenv/homebrew-tap/e472b7861b49cc082d1db0f66f265368da107589/Formula/openssl%401.0.rb -o openssl@1.0.rb
+brew install openssl@1.1 # possibly already installed on your system
 
-brew install ./openssl@1.0.rb
+brew link openssl@1.1 --force
 
-rm -rf ./openssl@1.0.rb
+echo 'export PATH="/opt/homebrew/opt/openssl@1.1/bin:$PATH"' >> ~/.zshrc
```
blu3beri, line 19: Not the end of the world but not everyone uses zsh. Maybe bash, ash ... or even fish?

karimStekelenburg: True, but since ZSH is macOs native I think anyone installing a custom shell will know what to do here...

blu3beri: alrighty!

TimoGlastra:
Wasn't there a limitation that we should use an older version of openssl?

Or why haven't we done this sooner? Seems like a more straight forward approach :)

blu3beri:
> Wasn't there a limitation that we should use an older version of openssl?
> 
> Or why haven't we done this sooner? Seems like a more straight forward approach :)

I think the brew version of `openssl@1.1` did not have arm64 support yet. that is why we needed to use openssl@1.0, because there was an arm64 build.

swcurran:
I assume this is for working with indy-sdk. The issue of OpenSSL 1.0 vs. 1.1 came up today on the Indy Contributors call.  Does these instructions imply that Indy SDK works fine with OpenSSL 1.1, or is that an open question?   Any guidance/experience would be appreciated.

blu3beri:
> I assume this is for working with indy-sdk. The issue of OpenSSL 1.0 vs. 1.1 came up today on the Indy Contributors call.  Does these instructions imply that Indy SDK works fine with OpenSSL 1.1, or is that an open question?   Any guidance/experience would be appreciated.

Ive been using afj with openssl 1.1 for a bit now, did not do extensive testing,  and it seems to work okay. If I encounter any issue I will report it here and at the indy-sdk. 


@karimStekelenburg it works okay for you aswell, right?

Issue #382, opened 2021-07-15
ProofService.createProof expects non-standard input object


TimoGlastra:
As a consequence of the changes made in #329, the `ProofService.createProof` method now expects the `RequestedCredentials` object to have a `credentialInfo` property. This is added in the `getRequestedCredentialsForProofRequest`, but is not a default property on the object.

In updating AATH we've noticed that we create the requested credentials outside of the framework because we already have the credential ids, meaning we don't call this method.

So my question is basically, is what were doing not the default approach and should thus handle this outside of the framework, OR should we take this into account in the createProof method and retrieve the credential info if not attached yet.

I'd lean toward the second option, but I'm interested to hear your opinion on this.

cc @TheTreek @JamesKEbert

JamesKEbert:
So if I understand correctly, the issue here isn't that `credentialInfo` is present, it's that the AATH doesn't have it in its tests. I am open to the second option, however I am curious if the AATH could use the AFJ methods available to get the requested credentials? 
And @TheTreek and I are working in this area already so we likely could make needed fixes here once decided upon a solution.

TimoGlastra:
Yes! AATH passes the requested credentials object to the backchannel, so no need to call the method that attaches `credentialInfo`. We've added a workaround in AATH for now.

I think we should extend the method in createProof to first check if `credentialInfo` is present (And make it optional on the class). If not present, fetch it in the method.

JamesKEbert:
Awesome, @TheTreek and I will look to make this change then. 👍

blu3beri:
@JamesKEbert any update on this?

JamesKEbert:
I am _actually_ working on revocation today, so yes this should be covered under that scope of work!

TimoGlastra:
Fixed in #592

Issue #349, opened 2021-07-03
Add support for receiving and proving revocable indy credentials


TimoGlastra:

TimoGlastra:
@JamesKEbert assigning you. This is only the receiving and proving part. #112 tracks the whole flow

swcurran:
A reminder -- because it's been a pain.

- A proof request that includes a revocation interval can be satisfied with a non-revocable credential.
- A proof request that does not include a revocation interval can be satisfied with a revocable credential.

LISSI and estatus support this correctly, Trinisic still does not.

JamesKEbert:
@swcurran revocation intervals in Indy can be specified on an attribute/predicate basis, is this something generally supported by the community, or is it sufficient to use the proof request's top level revocation interval?

swcurran:
I'm not aware of anyone using them, but all of that is handled in the Indy SDK (or indy-shared-rs) and so should "just work" if the conventions are followed.  See: https://github.com/hyperledger/aries-rfcs/tree/master/concepts/0441-present-proof-best-practices for what conventions are to be followed -- with the big thing being the from/to range to give the Indy proof request.

TimoGlastra:
@JamesKEbert what's the status of this? If you're not able to finish it, maybe you can post what's already there so someone else can pick it up?

JamesKEbert:
@TimoGlastra, sorry for any delays here--I actually have been making good progress on this, I expect to have a PR this week. I have a couple of things that might be good to discuss, so maybe during the AFJ call tomorrow some discussion time might be good?

TimoGlastra:
@JamesKEbert as part of the revocation work, are you also picking up the revocation notification protocol (#493)? There is someone else that might work on it, if you're not working on it already

JamesKEbert:
Yes, we are, we've only implemented the holder side of the protocol though. I figure when the efforts for the issuer-side happen we may be to a new iteration of the protocol anyways.

TimoGlastra:
That's great. Thanks James :)

PR #592, opened 2022-01-07
feat: indy revocation (prover & verifier)

JamesKEbert:
This PR adds support for querying the revocation status of credentials and presenting proof of non revocation as the Prover/Holder role in Indy revocation. 

For a presentation request, this PR will automatically fetch whether matching credentials are revoked given that the request specifies a non-revocation interval and the credentials are revocable. This is done for display to end users for their decision making or for app machine governance rules.
Additionally, when creating a presentation, the revocation states will be generated for the requestedCredentials automatically as part of the proof creation, if proof of non-revocation is requested and the credentials are revocable. 

This work was implemented in accordance with [Aries RFC 0441](https://github.com/hyperledger/aries-rfcs/blob/main/concepts/0441-present-proof-best-practices/README.md) and [Indy HIPE 0011](https://github.com/hyperledger/indy-hipe/blob/master/text/0011-cred-revocation/README.md). This PR addresses issues #349 & #382.

This PR is reliant on the merging of an indy-sdk typing fix PR (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/58026) & a corresponding PR for indy-sdk-react-native (https://github.com/hyperledger/indy-sdk-react-native/pull/27).

Additions by @TimoGlastra:
> Adds an downloadToFile method that directly download the file from a specified url to a specified path. The implementation is different for Node.JS and React Native, as it is hard to work with binary files directly in JS without touching Node.JS / RN specific API's.
> 
> This uses RNFS instead of expo file system. This makes the implementation easier, doesn't require everyone using AFJ to use expo and also doesn't mean we have to switch up any dependencies.

Future outstanding work:
- [ ] Addition of revocation tests (fairly problematic as we're only providing the Holder role, but can be supplemented by the AATH).
- [ ] In-memory caching of ledger transactions (this specifically will improve performance with revocation). 
  - [ ] Revocation Registry Definitions
  - [ ] Revocation Registry Deltas
- [ ] Persistent caching of ledger transactions (this specifically will improve performance with revocation). 
- [ ] Verification of downloaded tails file hash

A big thanks to other contributors: @TheTreek, @dbluhm, @TimoGlastra, @burdettadam, among many others :)


TimoGlastra:
indy-sdk-react-native 0.1.17 has been released: https://github.com/hyperledger/indy-sdk-react-native/releases/tag/0.1.17

JamesKEbert:
I have added the verifier portion: verifying proofs containing proof of non revocation. Addresses #350

TimoGlastra:
Awesome!

Indy SDK React Native 0.1.20 has been released: https://github.com/hyperledger/indy-sdk-react-native/releases/tag/0.1.20

The new Indy SDK types version should be released soon

packages/core/src/modules/indy/services/IndyRevocationService.ts
```diff
@@ -0,0 +1,197 @@
...
+
+    return {
+      revoked,
+      deltaTimestamp,
+    }
+  }
+
+  // TODO: Add Test
+  // Check revocation interval in accordance with https://github.com/hyperledger/aries-rfcs/blob/main/concepts/0441-present-proof-best-practices/README.md#semantics-of-non-revocation-interval-endpoints
+  private assertRevocationInterval(requestRevocationInterval: RevocationInterval) {
```
TimoGlastra, line 183: RFC 0441 defines best practices. Should we throw errors if those best practices are not followed?

packages/core/src/modules/indy/services/IndyVerifierService.ts
```diff
@@ -4,35 +4,64 @@ import { Lifecycle, scoped } from 'tsyringe'
...
+        revocationRegistryDefinitions[revRegId] = revocRegDef
+      }
+
+      //Fetch Revocation Registry by Timestamp if not already fetched
+      if (revRegId && timestamp && !revocationRegistryStates[revRegId]?.[timestamp]) {
+        if (!revocationRegistryStates[revRegId]) {
+          revocationRegistryStates[revRegId] = {}
+        }
+        const { revocReg } = await this.ledgerService.getRevocationRegistry(revRegId, timestamp)
+        revocationRegistryStates[revRegId][timestamp] = revocReg
```
TimoGlastra, line 61: It's complaining about prototype pollution and that we shouldn't create an object prone to prototype polution

packages/core/src/modules/ledger/LedgerModule.ts
```diff
@@ -72,4 +72,12 @@ export class LedgerModule {
   public async getCredentialDefinition(id: string) {
     return this.ledgerService.getCredentialDefinition(id)
   }
+
+  public async getRevocationRegistryDefinition(revocRegDefId: string) {
```
TimoGlastra, line 76: Nit: We use full names everywhere in AFJ, especially to avoid inconsistency in naming (e.g. this method uses `revocRegDefId` and the method below uses `revRegId`.

Maybe we can use the full `revocationRegistryDefinitionId` here? Is there a difference between `revocationRegistryId` and `revocationRegistryDefinitionId`? If not maybe we can use `revocationRegistryId` everywere?

packages/core/src/modules/ledger/LedgerModule.ts
```diff
@@ -72,4 +72,12 @@ export class LedgerModule {
...
+    return this.ledgerService.getRevocationRegistryDefinition(revocRegDefId)
+  }
+
+  public async getRevocationRegistryDelta(revRegId: string, from = 0, to = new Date().getTime()) {
```
TimoGlastra, line 80: Also nit: Having the time notation in the params names (and also making it an object can make more clear what the API expects.

E.g. if it needs milliseconds: 

```suggestion
  public async getRevocationRegistryDelta({ revocationRegistryId, fromMs, toMs = 0 }: {
  	revocationRegistryId: string;
  	fromMs: number,
  	toMs?: number
  }) {
```

packages/core/src/modules/ledger/services/IndyLedgerService.ts
```diff
@@ -369,6 +527,21 @@ export interface CredentialDefinitionTemplate {
   supportRevocation: boolean
 }
 
+export interface ParseRevRegDefTemplate {
+  revocRegDef: Indy.RevocRegDef
```
TimoGlastra, line 531: Same here, using `revocationRegistryDefinition` would be more clear IMO

TimoGlastra (approved):
Really nice work @JamesKEbert. Left a few nits, but overall LGTM.

Only thing is that this addition doesn't include a single test, meaning it's really easy to break something without us knowing. We've had discussions during the WG call on how to approach this. An easy scripts to test revocation with ACA-Py could be one. Another thing I was thinking is whether we can set up revocation in a way where we create a wallet with a revocable credential, export it and then import it everytime we run the tests. We can then mock all ledger responses and run through the flow of revocation. However that may be a bit too much work. It should be trivial to integrate this with the AATH, so maybe that's a good start

packages/core/src/modules/indy/services/IndyRevocationService.ts
JamesKEbert: Currently as the code is written it will throw errors if the RFC is not followed. 
But, ff your question was more along the lines of whether we should throw errors vs warnings, I would say errors based off of the inclusion of the 441 RFC in the [AIPv2 INDYCREDS sub-target](https://github.com/hyperledger/aries-rfcs/tree/main/concepts/0302-aries-interop-profile#indycred-indy-based-credentials). I'd also say that there's behavior that may be unexpected/not thought through if the intervals don't follow the RFC, so I'd be concerned with behavior past the assertion. However I'd be more than happy to discuss further, but that's my initial thoughts :)

packages/core/src/modules/indy/services/IndyVerifierService.ts
JamesKEbert: I had noticed this as well but had initially thought it was something specific to my environment. I'll look to reproduce.

JamesKEbert:
> An easy scripts to test revocation with ACA-Py could be one. Another thing I was thinking is whether we can set up revocation in a way where we create a wallet with a revocable credential, export it and then import it everytime we run the tests. We can then mock all ledger responses and run through the flow of revocation. However that may be a bit too much work. It should be trivial to integrate this with the AATH, so maybe that's a good start

Thanks for the review! For testing I'd definitely lean in the direction of first doing the AATH integration, and then following up with scripts for ACA-Py testing or wallet import testing. I'm definitely happy to help with either of those elements, although I admit my Docker-fu is a little lighter than I'd like 😅

packages/core/src/modules/indy/services/IndyVerifierService.ts
JamesKEbert: Interestingly I wasn't able to recreate this--I wonder if it was resolved with our last types PR addition?

packages/core/src/modules/ledger/LedgerModule.ts
JamesKEbert: And there's no difference between the two ids--they're identical 👍

packages/core/src/modules/ledger/LedgerModule.ts
JamesKEbert: I've updated it to `fromSeconds` as `fromS` looked a little peculiar to me.

JamesKEbert:
Note, we've opened a PR for the indy-sdk types to resolve the last sneaky type issue that I've found: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/58667

TimoGlastra:
Seems like revocation is working, but not for all test cases. However most seem related to test cases that test valid formats that do not follow the best practices. We should try to filter those out from running as we don't support them.

To run it yourself:

1. checkout [feature/afj-revocation branch out from TimoGlastra/aries-agent-test-harness](https://github.com/TimoGlastra/aries-agent-test-harness/tree/feature/afj-revocation)
2. [optional] If you made changes, update tar files in the aries-backchannels/javascript directory
3. build acapy-main and javascript backchannels
    - `./manage build -a acapy-main -a javascript`
4. Run revocation tests
    - `LEDGER_URL_CONFIG=http://test.bcovrin.vonx.io TAILS_SERVER_URL_CONFIG=https://tails.vonx.io ./manage run -d acapy-main -b javascript -t @revocation -t ~@wip -t ~@DIDExchangeConnection`

TimoGlastra:
Also `@types/indy-sdk@1.16.12` has been released

TimoGlastra:
I'm good with merging this and having it as a experimental feature. Then we can start tracking revocation test in AATH on a daily basis

JamesKEbert:
> I'm good with merging this and having it as an experimental feature. Then we can start tracking revocation test in AATH on a daily basis

Agreed--my only concern is in regards to merging today would be the prototype pollution mentioned, so if that's not something you're able to recreate than I'm in favor of merging.

> It's complaining about prototype pollution and that we shouldn't create an object prone to prototype polution

TimoGlastra:
It's still present: https://github.com/hyperledger/aries-framework-javascript/security/code-scanning/8?query=ref%3Arefs%2Fpull%2F592%2Fmerge+ref%3Arefs%2Fpull%2F592%2Fhead

PR #593, opened 2022-01-07
feat: add didcomm message record

TimoGlastra:
Adds a `DidCommMessageRecord` that can be used to store didcomm messages.

This can be useful in the OOB protocol, and with the addition of v2 to the credentials and proofs module. This way we don't have to store the messages inside the record class anymore. 

Instead of accessing and storing a message in the record like this:

```ts
// write
credentialRecord.proposalMessage = proposalMessage
await credentialRepository.save(credentialRecord)

// read
credentialRecord.proposalMessage
```

we can use it like this:

```ts
// write
await didcommMessageRepository.saveAgentMessage({ 
    agentMessage: proposalMessage, 
    role: DidCommMessageRole.Receiver,
    associatedRecordId: credentialRecord.id
})

// read
await didcommMessageRepository.getAgentMessage({
   associatedRecordId: credentialRecord.id,
   messageClass: V2CredentialProposalMessage
})
git 

```

The implementation is BETA, but I think it can be useful to already add this to the framework so we can start using it.

### DIFF
--- a/packages/core/src/storage/didcomm/DidCommMessageRecord.ts
+++ b/packages/core/src/storage/didcomm/DidCommMessageRecord.ts
@@ -1,11 +1,11 @@
 import type { AgentMessage } from '../../agent/AgentMessage'
-import type { JsonMap } from '../../types'
+import type { JsonObject } from '../../types'
 import type { DidCommMessageRole } from './DidCommMessageRole'
 
 import { AriesFrameworkError } from '../../error'
 import { JsonTransformer } from '../../utils/JsonTransformer'
 import { rightSplit } from '../../utils/string'
-import { isJsonMap } from '../../utils/type'
+import { isJsonObject } from '../../utils/type'
 import { uuid } from '../../utils/uuid'
 import { BaseRecord } from '../BaseRecord'
 
@@ -25,14 +25,14 @@ export type DefaultDidCommMessageTags = {
 
 export interface DidCommMessageRecordProps {
   role: DidCommMessageRole
-  message: JsonMap
+  message: JsonObject
   id?: string
   createdAt?: Date
   associatedRecordId?: string
 }
 
 export class DidCommMessageRecord extends BaseRecord<DefaultDidCommMessageTags> {
-  public message!: JsonMap
+  public message!: JsonObject
   public role!: DidCommMessageRole
 
   /**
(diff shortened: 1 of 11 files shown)


TimoGlastra:
Updated and ready for review

TimoGlastra:
Can someone review this PR please? Need this for issue credential / present proof v2

packages/core/src/types.ts
```diff
@@ -78,3 +78,9 @@ export interface OutboundPackage {
   endpoint?: string
   connectionId?: string
 }
+
+export type AnyJson = boolean | number | string | null | JsonArray | JsonMap
+export interface JsonMap {
+  [key: string]: AnyJson
+}
+export type JsonArray = Array<AnyJson>
```
jakubkoci, line 86: This is good. I was confused a bit with `Map` in `JsonMap`, 

Could we use the following naming:
```
type JsonValue = string | number | boolean | JsonObject | JsonArray;

interface JsonObject {
  [property: string]: JsonValue;
}

interface JsonArray extends Array<JsonValue> { }
```

It's actually according to the standard https://www.json.org/json-en.html where you also have object, array and value

packages/core/src/storage/didcomm/DidCommMessageRecord.ts
```diff
@@ -0,0 +1,99 @@
...
+  public associatedRecordId?: string
+
+  public static readonly type = 'DidCommMessageRecord'
+  public readonly type = DidCommMessageRecord.type
+
+  public constructor(props: DidCommMessageRecordProps) {
+    super()
+
+    if (props) {
+      this.id = props.id ?? uuid()
```
jakubkoci, line 53: I'm thinking if we shouldn't use message ID instead of creating an artificial one. But I guess it's actually a record ID and not a message ID in this case.

jakubkoci (approved):
I just added one note about better naming, otherwise, it's good to me 👍

packages/core/src/storage/didcomm/DidCommMessageRecord.ts
TimoGlastra: Problem with message id is that it can be provided by other agents, meaning it could be non-unique.... That's why I went with a new, generated, id.

packages/core/src/types.ts
TimoGlastra: Yes! This was just a copy paste from somewhere :)

PR #595, opened 2022-01-11
feat: add support for signed attachments

TimoGlastra:
Adds support for signed attachments as described in RFC 0017.

Usage:

```ts
 const attachment = new Attachment({
    id: 'some-uuid',
    data: new AttachmentData({
      base64: JsonEncoder.toBase64(didJwsz6Mkf.DATA_JSON),
    }),
  })

// sign
await attachment.data.sign(wallet, [verkey])

// verify
await attachment.data.verify(wallet)

// match verkeys of jws against other verkey (e.g. from invitation)
if (!attachment.data.jwsVerkeys.includes(connection.invitationKey)) {
   throw new Error("Attachment not signed with invitation key!")
}
```


jakubkoci (commented):
It looks good, but would not be better to separate logic from Attachment outside. Maybe, I'm missing something but it seems like there is a lot of logic related only to JWS and not to the attachment itself.

For example:
`AttachmentJwsHeaderOptions`, `AttachmentJwsHeader`, `AttachmentJwsGeneralOptions` represent general JWS data and are not related to attachment.

I understand the benefit of calling just `attachment.data.sign()`, though, the following code it's still quite good to use:
```ts
const  = signatureService.createJws(attachment)
attachment.addJws(jws)
```

It'll be convenient if there are other objects we would like to sign in the future.

I would also try to avoid passing wallet around and rather send data to sign to the wallet, but I understand there are some intertwining piecies  in `createJws`.  🤔  

From the security perspective, there is a good practice where you have a separate business service and a wallet service and you just pass the messages you want to sign to a queue. There is no direct access from a business service to the wallet service. I know we're in a different environment here, I'm just giving it as an example of how I think about it. :)

TimoGlastra:
Updated the PR with an `JwsService`.

For DIDExchange with attachment it would now look like this:

```ts
const { verkey } = await wallet.createDid({ seed: didJwsz6Mkf.SEED });

const kid = DidKey.fromPublicKeyBase58(verkey, KeyType.ED25519);
const payload = JsonEncoder.toBuffer(didJwsz6Mkf.DATA_JSON);

const attachment = new Attachment({
  data: new AttachmentData({
    base64: JsonEncoder.toBase64(didJwsz6Mkf.DATA_JSON),
  }),
});

const jws = await jwsService.createJws({
  payload,
  verkey,
  header: {
    kid,
  },
});

attachment.addJws(jws);

const { isValid, signerVerkeys } = await jwsService.verifyJws({
  jws: attachment.data.jws,
  payload,
});

if (!isValid || !signerVerkeys.includes(invitationKey)) {
  // err diddoc is not signed correctly
}

```

Thanks for the feedback @jakubkoci, looks better now :)

jakubkoci (approved):
Great. Thanks for the example with DID exchange 👍

PR #597, opened 2022-01-13
feat: add issue credential v2

NB-MikeRichardson:
This is the first draft PR for the AIP 2.0 stuff. Some notes:

1. There are MJR-TODO reminders in places. This is a reminder for me of remaining work that still needs to be done, and some questions in places.

2. The CredentialExchangeRecord stuff is not yet implemented pending further discussions on this.

3. There is both a V1CredentialService and V1LegacyCredentialService and in most cases one just delegates to the other. This is because the interfaces are different in the new architecture (and subject to change) so it made sense to touch the existing service code as little as possible (for now). The intention is to move the legacy code into the new V1CredentialService in due course. 
4. Similarly CredentialModules will be removed eventually, to be replaced by CredentialsAPI

5. The new DIDCommMessageRecord stuff is not done yet (but it looks like the PR for this has not been merged yet so perhaps I will do that a bit later)

All tests should run fine and lint checking has been done.


packages/core/src/modules/credentials/v2/CredentialExchangeRecord.ts
```diff
@@ -0,0 +1,86 @@
...
+export interface CredentialRecordTags {
+  threadId: string
+  protocolVersion: CredentialProtocolVersion
+  state: CredentialState
+  connectionId?: string
+}
+
+export enum CredentialRecordType {
+  INDY,
+  W3C,
```
TimoGlastra, line 18: I would assign string values to the enum values as we will be storing the values. If you don't assign a value it will be auto-assigned a number based on the order of declaration

So in this case INDY = 0, W3C = 1, if we later add an entry at the top of the array, INDY = 1, W3C = 2. 
 

Also we normally use PascalCase for enums

```suggestion
  Indy = "Indy",
  W3c = "W3c",
```

packages/core/src/modules/credentials/v2/CredentialMessageBuilder.ts
```diff
@@ -0,0 +1,273 @@
...
+import { V2OfferCredentialMessage } from './messages/V2OfferCredentialMessage'
+import { V2ProposeCredentialMessage } from './messages/V2ProposeCredentialMessage'
+import { V2RequestCredentialMessage } from './messages/V2RequestCredentialMessage'
+
+export interface CredentialProtocolMsgReturnType<MessageType extends AgentMessage> {
+  message: MessageType
+  credentialRecord: CredentialRecord
+}
+
+export class CredentialMessageBuilder {
```
TimoGlastra, line 29: The credential service feels quite light at the moment. I'm in favour of adding extra abstraction layers (such as the `CredentialMessageBuilder`), but I'm not sure if this is the right implementation of it. It is not clear to me exactly what should go into the CredentielMessageBuilder vs the V2CredentialService, and some methods have nothing to do with building messages (e.g. `getCredentialDefinitionId`). 

The methods that call this builder have a very small function body because most of the logic happens here, so my preference would be took keep most of the logic in the `V2CredentialService` for now. I'm open to other suggestions

packages/core/src/modules/credentials/v2/V2CredentialPreview.ts
```diff
@@ -0,0 +1,64 @@
...
+    }
+  }
+
+  @Expose({ name: '@type' })
+  @Equals(V2CredentialPreview.type)
+  @Transform(({ value }) => replaceLegacyDidSovPrefix(value), {
+    toClassOnly: true,
+  })
+  public type = V2CredentialPreview.type
+  public static type = `https://didcomm.org/issue-credential/2.0/credential-preview`
```
TimoGlastra, line 30:
```suggestion
  public readonly type = V2CredentialPreview.type
  public static readonly type = 'https://didcomm.org/issue-credential/2.0/credential-preview'
```

packages/core/src/modules/credentials/v2/V2CredentialPreview.ts
```diff
@@ -0,0 +1,64 @@
...
+    if (options) {
+      this.attributes = options.attributes
+    }
+  }
+
+  @Expose({ name: '@type' })
+  @Equals(V2CredentialPreview.type)
+  @Transform(({ value }) => replaceLegacyDidSovPrefix(value), {
+    toClassOnly: true,
+  })
```
TimoGlastra, line 28: `replaceLegacyDidSovPrefix` is not needed anymore for v2 messages. This was needed for v1 when we went from `did:sov:xxx` to `https://didcomm.org` prefixes. But the v2 issue credential protocol uses `https://didcomm.org` natively

```suggestion
```

TimoGlastra (changes requested):
Great work @NB-MikeRichardson!

I think the overall architecture and code structure looks good to me, but there are some things I'd approach differently. I've left some comments to share my thoughts.

I only looked at more overarching points. When doing a final review I'd be more thorough (code formatting, eslint rules, left over comments, unitTest logger, commented code, etc...) but I suspected those are temporary and therefore ignored those.

NB-MikeRichardson:
Thank you Timo

Yes there is much stuff that is only for “draft” review. I will go through your comments and make any necessary changes.

I think I would do things differently to a degree if I had as much knowledge of AFJ as I do now of course (so I adopted a conservative policy).

Talk to you tomorrow
Cheers
Mike

packages/core/src/modules/credentials/CredentialService.ts
```diff
@@ -0,0 +1,45 @@
...
+import type { RequestCredentialMessage } from './v1/messages'
+import type { CredentialRecordType } from './v2/CredentialExchangeRecord'
+import type { V2CredentialService } from './v2/V2CredentialService'
+import type { CredentialFormatService } from './v2/formats/CredentialFormatService'
+import type { AcceptProposalOptions, ProposeCredentialOptions, RequestCredentialOptions } from './v2/interfaces'
+import type { V2RequestCredentialMessage } from './v2/messages/V2RequestCredentialMessage'
+
+export type CredentialServiceType = V1CredentialService | V2CredentialService
+
+export abstract class CredentialService {
```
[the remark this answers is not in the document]
NB-MikeRichardson: CredentialServiceType not needed. I changed this to just return CredentialService (the interface). Not sure why I put that type in there.

packages/core/src/modules/credentials/CredentialsAPI.ts
```diff
@@ -0,0 +1,397 @@
+import type { AgentMessage } from '../../agent/AgentMessage'
+import type { CredentialService, CredentialServiceType } from './CredentialService'
+import type { CredentialRecord } from './repository'
+import type { CredentialRecordBinding, CredentialExchangeRecordProps } from './v2/CredentialExchangeRecord'
+import type {
+  AcceptOfferOptions,
+  AcceptProposalOptions,
+  ProposeCredentialOptions,
+  RequestCredentialOptions,
+} from './v2/interfaces'
```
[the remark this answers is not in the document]
NB-MikeRichardson: I moved interfaces.ts up one level. Now at top level for both v1 and v2

packages/core/src/modules/credentials/CredentialsAPI.ts
```diff
@@ -0,0 +1,397 @@
...
+  getById(credentialRecordId: string): Promise<CredentialRecord>
+  // findById(credentialRecordId: string): Promise<CredentialExchangeRecord | null>
+  // deleteById(credentialRecordId: string): Promise<void>
+  // findByQuery(query: Record<string, Tag | string[]>): Promise<CredentialExchangeRecord[]>
+}
+
+@scoped(Lifecycle.ContainerScoped)
+export class CredentialsAPI extends CredentialsModule implements CredentialsAPI {
+  private connService: ConnectionService
+  private msgSender: MessageSender
```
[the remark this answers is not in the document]
NB-MikeRichardson: This was done to distinguish the local objects (v2) from the ones passes up to the base class. I renamed them with a v2prefix. This can be changed later when the base class (CredentialsModule) is removed.

packages/core/src/modules/credentials/CredentialsAPI.ts
```diff
@@ -0,0 +1,397 @@
...
+  private dispatcher: Dispatcher
+  private agConfig: AgentConfig
+  private credentialResponseCoord: CredentialResponseCoordinator
+  private v1Service: V1CredentialService
+  private v2Service: V2CredentialService
+  private indyIssuerService: IndyIssuerService
+  private mediatorRecipientService: MediationRecipientService
+  private indyLedgerService: IndyLedgerService
+  private indyHolderService: IndyHolderService
+  private serviceMap: { '1.0': V1CredentialService; '2.0': V2CredentialService }
```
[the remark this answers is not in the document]
NB-MikeRichardson: Some typescript coolness!

packages/core/src/modules/credentials/v2/CredentialMessageBuilder.ts
NB-MikeRichardson: I have been grappling with this concept as well, ie functionality in the CredentialsService vs the MessageBuilder.

I think the message builder has grown more than I wanted. I will make these changes later.

PR #604, opened 2022-01-21
feat: cli demo

Annelein:
A CLI demo to get to know the AFJ flow together with agents Alice and Faber.

- ✅ Creating a connection
- ✅ Offering a credential
- ✅ Requesting a proof
- ✅ Sending basic messages


demo/src/base_agent.ts
```diff
@@ -0,0 +1,54 @@
+import type { InitConfig } from '@aries-framework/core'
+
+import { Agent, AutoAcceptCredential, AutoAcceptProof, HttpOutboundTransport } from '@aries-framework/core'
+import { agentDependencies, HttpInboundTransport } from '@aries-framework/node'
+
+import { Color } from './output_class'
+
+const bc_coverin = `{"reqSignature":{},"txn":{"data":{"data":{"alias":"Node1","blskey":"4N8aUNHSgjQVgkpm8nhNEfDf6txHznoYREg9kirmJrkivgL4oSEimFF6nsQ6M41QvhM2Z33nves5vfSn9n1UwNFJBYtWVnHYMATn76vLuL3zU88KyeAYcHfsih3He6UHcXDxcaecHVz6jhCYz1P2UZn2bDVruL5wXpehgBfBaLKm3Ba","blskey_pop":"RahHYiCvoNCtPTrVtP7nMC5eTYrsUA8WjXbdhNc8debh1agE9bGiJxWBXYNFbnJXoXhWFMvyqhqhRoq737YQemH5ik9oL7R4NTTCz2LEZhkgLJzB3QRQqJyBNyv7acbdHrAT8nQ9UkLbaVL9NBpnWXBTw4LEMePaSHEw66RzPNdAX1","client_ip":"138.197.138.255","client_port":9702,"node_ip":"138.197.138.255","node_port":9701,"services":["VALIDATOR"]},"dest":"Gw6pDLhcBcoQesN72qfotTgFa7cbuqZpkX3Xo6pLhPhv"},"metadata":{"from":"Th7MpTaRZVRYnPiabds81Y"},"type":"0"},"txnMetadata":{"seqNo":1,"txnId":"fea82e10e894419fe2bea7d96296a6d46f50f93f9eeda954ec461b2ed2950b62"},"ver":"1"}
```
TimoGlastra, line 8:
```suggestion
const bcovrin = `{"reqSignature":{},"txn":{"data":{"data":{"alias":"Node1","blskey":"4N8aUNHSgjQVgkpm8nhNEfDf6txHznoYREg9kirmJrkivgL4oSEimFF6nsQ6M41QvhM2Z33nves5vfSn9n1UwNFJBYtWVnHYMATn76vLuL3zU88KyeAYcHfsih3He6UHcXDxcaecHVz6jhCYz1P2UZn2bDVruL5wXpehgBfBaLKm3Ba","blskey_pop":"RahHYiCvoNCtPTrVtP7nMC5eTYrsUA8WjXbdhNc8debh1agE9bGiJxWBXYNFbnJXoXhWFMvyqhqhRoq737YQemH5ik9oL7R4NTTCz2LEZhkgLJzB3QRQqJyBNyv7acbdHrAT8nQ9UkLbaVL9NBpnWXBTw4LEMePaSHEw66RzPNdAX1","client_ip":"138.197.138.255","client_port":9702,"node_ip":"138.197.138.255","node_port":9701,"services":["VALIDATOR"]},"dest":"Gw6pDLhcBcoQesN72qfotTgFa7cbuqZpkX3Xo6pLhPhv"},"metadata":{"from":"Th7MpTaRZVRYnPiabds81Y"},"type":"0"},"txnMetadata":{"seqNo":1,"txnId":"fea82e10e894419fe2bea7d96296a6d46f50f93f9eeda954ec461b2ed2950b62"},"ver":"1"}
```

demo/src/faber.ts
```diff
@@ -0,0 +1,168 @@
...
+
+    try {
+      invitationJson = JsonEncoder.fromBase64(invitationJson)
+    } catch (e) {
+      console.log(`${Color.green}\nIt looks like your invitation link is not correctly formatted?\n${Color.reset}`)
+      return
+    }
+
+    const invitationMessage = JsonTransformer.fromJSON(invitationJson, ConnectionInvitationMessage)
+    return await this.agent.connections.receiveInvitation(invitationMessage)
```
TimoGlastra, line 54: I think we can simplify this with the `receiveInvitationFromUrl` method

```suggestion
    return await this.agent.connections.receiveInvitationFromUrl(invitationUrl)
```

demo/src/faber.ts
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  private async getConnectionRecord() {
+    if (!this.connectionRecordAliceId) {
+      throw Error(`${Color.red}${Output.missingConnectionRecord}${Color.reset}`)
+    }
+    return await this.agent.connections.getById(this.connectionRecordAliceId)
+  }
+
+  private async receiveConnectionRequest(invitation_url: string) {
```
TimoGlastra, line 42:
```suggestion
  private async receiveConnectionRequest(invitationUrl: string) {
```

demo/src/alice.ts
```diff
@@ -0,0 +1,79 @@
...
+import { BaseAgent } from './base_agent'
+import { Color, Output } from './output_class'
+
+export class Alice extends BaseAgent {
+  public connectionRecordFaberId?: string
+  public credDef: string
+
+  public constructor(port: number, name: string) {
+    super(port, name)
+    this.credDef = '7KuDTpQh3GJ7Gp6kErpWvM:3:CL:115269:latest'
```
TimoGlastra, line 14: Is credDef used?

demo/src/faber.ts
```diff
@@ -0,0 +1,168 @@
...
+      name: 'Alice Smith',
+      degree: 'Computer Science',
+      date: '01/01/2022',
+    })
+    return credentialPreview
+  }
+
+  public async issueCredential() {
+    const schema = await this.registerSchema()
+    const credentialDefinition = await this.registerCredentialDefiniton(schema)
```
TimoGlastra, line 113: creating a schema and credential definition only needs to happen once, I think it would be good to reflect that in the demo.

demo/src/faber_inquirer.ts
```diff
@@ -0,0 +1,133 @@
...
+
+export const runFaber = async () => {
+  clear()
+  console.log(textSync('Faber', { horizontalLayout: 'full' }))
+  const faber = await FaberInquirer.build()
+  faber.processAnswer()
+}
+
+enum PromptOptions {
+  Connection = 'setup connection',
```
TimoGlastra, line 18: Maybe we can make it more clear what kind of the flow each side will handle?

E.g. `Create connection invitation` for Alice, and `Receive connection invitation` for Faber. Currently both have `setup connection` which makes it hard to know which agent does what

demo/src/faber_inquirer.ts
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  public async processAnswer() {
+    const choice = await this.getPromptChoice()
+    if (this.listener.on === true) {
+      return
+    }
+    switch (choice.options) {
+      case PromptOptions.Connection:
+        await this.connection()
```
TimoGlastra, line 56: Nitpick, but it would be good to make methods as descriptive as possible. I had to look at the implementation to know that this is about asking for an invitation url and creating a connection with it. 

Maybe something like : `this.receiveInvitationPrompt()`

demo/src/listener.ts
```diff
@@ -0,0 +1,115 @@
...
+    this.turnListenerOff()
+    aliceInquirer.processAnswer()
+  }
+
+  public proofRequestListener(alice: Alice, aliceInquirer: AliceInquirer) {
+    alice.agent.events.on(ProofEventTypes.ProofStateChanged, async ({ payload }: ProofStateChangedEvent) => {
+      if (payload.proofRecord.state === ProofState.RequestReceived) {
+        await this.newProofRequestPrompt(payload.proofRecord, aliceInquirer)
+      }
+      return
```
TimoGlastra, line 96: This return statement is unnecessary I think?

```suggestion
```

demo/src/listener.ts
```diff
@@ -0,0 +1,115 @@
...
+    })
+  }
+
+  public proofAcceptedListener(faber: Faber, faberInquirer: FaberInquirer) {
+    faber.agent.events.on(ProofEventTypes.ProofStateChanged, async ({ payload }: ProofStateChangedEvent) => {
+      if (payload.proofRecord.state === ProofState.Done) {
+        faberInquirer.processAnswer()
+      }
+      return
```
TimoGlastra, line 105: Same here

demo/package.json
```diff
@@ -0,0 +1,51 @@
+{
+  "name": "demo",
+  "version": "1.0.0",
+  "main": "./lib/annelein.js",
+  "bin": {
+    "demo": "./lib/annelein.js"
+  },
+  "license": "MIT",
```
TimoGlastra, line 8: It is important to add

```
"private": true,
```

to make sure it doesn't get published to NPM

lerna.json
```diff
@@ -1,5 +1,5 @@
 {
-  "packages": ["packages/*"],
+  "packages": ["packages/*", "demo"],
```
TimoGlastra, line 2: I don't think this is neccesarry

TimoGlastra (changes requested):
Nice work @Annelein! :)

In terms of user flow the demo looks really good, just left some comments on the code. 

I did encounter sometimes that the screen will flicker / things will jump and get printed multiple times

Some other comments I didn't leave in the code comments:

- `Credential offer send!` -> `Credential offer sent!`

---

If I create a connection invitation url in the alice terminal and wait for more than 20 seconds the demo will error and crash. This is due to `returnWhenIsConnected` that has a timeout of 20s. Maybe add a catch and return to the home menu?

image


---

When I paste the connection invitation url it suddenly jumps a few lines down. Not too problematic, but feels a bit quircky

image

---

When offering a credential you get the question if the credential was accepted. When it is declined we can't know whether is still unanswered or really declined, but if the credential was accepted it would be nice to show "Credential accepted!" and go back to the home menu. Same for proofs

image

---

Sometimes text is showed twice. Such as this in the video, or the `Options` key.

https://user-images.githubusercontent.com/23165168/150543214-c521d875-1c1d-4b67-a1cc-a8d0046dd48c.mov

---

When issuing a credential or requesting a proof you never see the attributes of the proof or the credential. Would be nice to show the attributes like you did with registering a schema/cred def

---

TimoGlastra:
Also, would be good to add this to main readme so people know it exists! :)

demo/src/faber.ts
```diff
@@ -0,0 +1,168 @@
...
+    await this.agent.credentials.offerCredential(connectionRecord.id, {
+      credentialDefinitionId: credentialDefinition.id,
+      preview: credentialPreview,
+    })
+    this.ui.updateBottomBar(`\nCredential offer sent! \n${Color.reset}`)
+  }
+
+  private async printProofFlow(print: string) {
+    this.ui.updateBottomBar(print)
+    await new Promise((f) => setTimeout(f, 2000))
```
blu3beri, line 127: Why do we need to wait here?

demo/src/faber.ts
```diff
@@ -0,0 +1,168 @@
...
+          }),
+        ],
+      }),
+    }
+  }
+
+  public async sendProofRequest() {
+    const connectionRecord = await this.getConnectionRecord()
+    const proofAttribute = await this.newProofAttribute()
+    await this.printProofFlow(`${Color.green}\nRequesting proof...\n`)
```
blu3beri, line 147: Might be useful to create a util method that takes a color and a message for nicer printing. String interpolation like this everywhere is not very nice.

demo/src/faber.ts
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  public async sendMessage(message: string) {
+    const connectionRecord = await this.getConnectionRecord()
+    await this.agent.basicMessages.sendMessage(connectionRecord.id, message)
+  }
+
+  public async exit() {
+    console.log(Output.exit)
+    await this.agent.shutdown()
```
blu3beri, line 161: Do we also want to delete the wallet after this?

demo/src/faber_inquirer.ts
```diff
@@ -0,0 +1,133 @@
...
+        break
+      case PromptOptions.Credential:
+        await this.credential()
+        return
+      case PromptOptions.Proof:
+        await this.proof()
+        return
+      case PromptOptions.Message:
+        await this.message()
+        break
```
blu3beri, line 66: are all these return/break correct? Looks a bit random, but I might be wrong

demo/src/faber.ts
Annelein: Because if you don't, the program is too fast for printing to see what is going on.

demo/src/faber.ts
```diff
@@ -0,0 +1,168 @@
...
+
+  private async printProofFlow(print: string) {
+    this.ui.updateBottomBar(print)
+    await new Promise((f) => setTimeout(f, 2000))
+  }
+
+  private async newProofAttribute() {
+    await this.printProofFlow(`${Color.green}\nCreating new proof attribute for 'name' ...\n`)
+    return {
+      name: new ProofAttributeInfo({
```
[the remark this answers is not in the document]
Annelein: What do you mean with the name contains the proof attribute info object which is not correct?

demo/src/faber.ts
Annelein: I don't know, is that something you should do?

demo/src/faber_inquirer.ts
Annelein: Yeah because with the Credential and Proof use case, you get a new prompt afterwards (not the main menu), once that prompt is finished, you get the main menu but this is called from the other prompt.. I'll look into if it's possible to do this differently!

demo/src/faber.ts
```diff
@@ -0,0 +1,155 @@
+import type { ConnectionRecord } from '@aries-framework/core'
+import type { CredDef, Schema } from 'indy-sdk-react-native'
+import type BottomBar from 'inquirer/lib/ui/bottom-bar'
+
+import { CredentialPreview, ProofAttributeInfo, AttributeFilter } from '@aries-framework/core'
+import { uuid } from '@aries-framework/core/build/utils/uuid'
```
TimoGlastra, line 6: This still contains the build path, won't run for me like this

```suggestion
import { utils } from '@aries-framework/core'

utils.uuid()
```

TimoGlastra:
@Annelein, there's a conflict in the yarn.lock file, and the DCO sign off is failing again. Could you resolve these?

TimoGlastra:
I had some issues running locally because of changed file names. Git by default matches file name changes case insensitive. So faber => Faber didn't actually work: https://stackoverflow.com/questions/10523849/changing-capitalization-of-filenames-in-git/24979063#24979063

TimoGlastra:
If I press `Send message` and try to send a message in the Alice demo while no connection has been created yet, the demo crashes and exists. Maybe we should only add that option after the connection has been created?

image

demo/src/faber.ts
```diff
@@ -0,0 +1,155 @@
...
+    }
+    this.printSchema(schemaTemplate.name, schemaTemplate.version, schemaTemplate.attributes)
+    this.ui.updateBottomBar(greenText('\nRegistering schema...\n', false))
+    const schema = await this.agent.ledger.registerSchema(schemaTemplate)
+    this.ui.updateBottomBar('\nSchema registerd!\n')
+    return schema
+  }
+
+  private async registerCredentialDefiniton(schema: Schema) {
+    this.ui.updateBottomBar('\nRegistering credential defenition...\n')
```
TimoGlastra:
```suggestion
    this.ui.updateBottomBar('\nRegistering credential definition...\n')
```

lerna.json
TimoGlastra:
```suggestion
  "packages": ["packages/*"],
```

TimoGlastra:
Few comments but otherwise looks good to me @Annelein. 

One thing that I pointed out in my previous comment is that it feels a bit weird that you have to specify whether the credential/proof was accepted even if the credential/proof was accepted. This can be detected using a listener.

I see you already have the `proofAcceptedListener` but this isn't called anywhere. I think that should be sufficient. If we then also do that for credential, would make the flow more logical IMO.

image

PR #608, opened 2022-01-26
feat: add support for did:peer

TimoGlastra:
This PR adds support for did:peer dids as described in the peer did method specification. It supports all methods (0,1,2).

The implementation should be seen as experimental as the peer did spec is quite complex and there's a good chance I messed some things up.

One thing that needs fixing for sure is the encoding of didcomm services, but I need to wait for a reply on that: https://github.com/decentralized-identity/peer-did-method-spec/issues/39


packages/core/src/modules/dids/__tests__/peer-did.test.ts
```diff
@@ -0,0 +1,157 @@
+import type { IndyLedgerService } from '../../ledger'
+
+import { getAgentConfig } from '../../../../tests/helpers'
```
TimoGlastra, line 3: @jakubkoci I wrote these tests to approximately work out what we'll be needing for DIDExchange. The first test is when we create the did document. The second step is when we are receiving the did document

TimoGlastra:
What's still missing is the integration of the did resolver into the message sender. I'll take a look at that next, but this should be enough to start integrating with the peer did API.

samples/mediator.ts
```diff
@@ -13,6 +13,7 @@
  */
 
 import type { InitConfig } from '@aries-framework/core'
+import type { Socket } from 'net'
```
jakubkoci, line 16: It isn't the correct package to import `Socket`, is it?

TimoGlastra: It is for the version of ws we have: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/57590/files

This is fixed in v8 of ws, but there are some breaking changes that would need proper testing before updating (seems like there are some incompallities with the normal WS, but not sure)

So would like to keep it like this for now until we update to v8

packages/core/src/modules/dids/methods/indy/IndyDidResolver.ts
```diff
@@ -1,15 +1,15 @@
...
 
-import { BufferEncoder } from '../../../utils/BufferEncoder'
-import { getFullVerkey } from '../../../utils/did'
-import { DidCommService } from '../../connections'
-import { DidDocumentService } from '../domain'
-import { DidDocumentBuilder } from '../domain/DidDocumentBuilder'
-import { DidCommV2Service } from '../domain/service/DidCommV2Service'
+import { BufferEncoder } from '../../../../utils/BufferEncoder'
+import { getFullVerkey } from '../../../../utils/did'
+import { DidCommService } from '../../../connections'
```
jakubkoci, line 9: I think `connections` should depend on `dids` module

jakubkoci (approved):
Super awesome Timo. No time for proper deep-dive review, but overall looks good.

packages/core/src/modules/dids/methods/indy/IndyDidResolver.ts
TimoGlastra: Fixed

PR #609, opened 2022-01-27
feat: support new did document in didcomm message exchange

TimoGlastra:
Adds support for sending and receiving messages that are stored using peer dids and the new did record. The sending of messages is completely the same, mainly because I unified the `Service` classes from the old and new did document class as they were 100% identical. This made integrating it quite straight forward

Current exchange is still working, but haven't been able to properly test the message exchange using the new did documents yet as we don't have DID exchange etc yet. Will add proper tests over the next few days, but you should be fine to use for now.


packages/core/src/modules/dids/__tests__/peer-did.test.ts
```diff
@@ -11,15 +11,15 @@ import { convertPublicKeyToX25519, getEd25519VerificationMethod } from '../domai
 import { getX25519VerificationMethod } from '../domain/key-type/x25519'
 import { DidKey } from '../methods/key'
 import { DidPeer, PeerDidNumAlgo } from '../methods/peer/DidPeer'
-import { DidDocumentRecord, DidDocumentRepository } from '../repository'
```
TimoGlastra, line 14: @jakubkoci there are a few changes to how you should retrieve and store the did document

packages/core/src/modules/dids/__tests__/peer-did.test.ts
```diff
@@ -102,15 +102,20 @@ describe('peer dids', () => {
...
       id: didPeer1zQmY.id,
       role: DidDocumentRole.Created,
       // It is important to take the did document from the PeerDid class
       // as it will have the id property
       didDocument: peerDid.didDocument,
+      tags: {
+        // We need to save the recipientKeys, so we can find the associated did
+        // of a key when we receive a message from another connection.
+        recipientKeys: peerDid.didDocument.recipientKeys,
+      },
```
TimoGlastra, line 115: Main difference is that we need to store the recipientKeys so we can retrieve the connection associated with a received message

packages/core/src/modules/dids/__tests__/peer-did.test.ts
```diff
@@ -134,17 +139,20 @@ describe('peer dids', () => {
...
+      role: DidDocumentRole.Received,
+      // If the method is a genesis doc (did:peer:1) we should store the document
+      // Otherwise we only need to store the did itself (as the did can be generated)
+      didDocument: didPeer.numAlgo === PeerDidNumAlgo.GenesisDoc ? didPeer.didDocument : undefined,
+      tags: {
+        // We need to save the recipientKeys, so we can find the associated did
+        // of a key when we receive a message from another connection.
+        recipientKeys: didPeer.didDocument.recipientKeys,
+      },
+    })
```
TimoGlastra, line 153: This means we now always need to create a did record (that's why I changed it from did document record to did record), but only if using `numAlgo === PeerDidNumAlgo.GenesisDoc` we store the actual did document

jakubkoci (approved):
Ok, we can merge it.

PR #615, opened 2022-01-31
feat(core): did-exchange protocol

jakubkoci:


packages/core/src/modules/connections/DidExchangeProtocol.ts
```diff
@@ -0,0 +1,462 @@
...
+
+  public async createRequest(
+    connectionRecord: ConnectionRecord,
+    params: DidExchangeRequestParams
+  ): Promise<DidExchangeRequestMessage> {
+    this.logger.debug(`Create message ${DidExchangeRequestMessage.type} start`, connectionRecord)
+
+    if (!connectionRecord.invitation) {
+      throw new AriesFrameworkError('Connection invitation is missing.')
+    }
```
TimoGlastra, line 72: This implementation is still leaning on the 0160 invitation. It feels limiting, as we don't store the OOB invitation anywhere now even though it is richer than the connection invitation message

packages/core/src/modules/connections/DidExchangeProtocol.ts
```diff
@@ -0,0 +1,462 @@
...
+    if (!connectionRecord) {
+      throw new AriesFrameworkError('No connection record in message context.')
+    }
+
+    DidExchangeStateMachine.assertProcessMessageState(DidExchangeRequestMessage.type, connectionRecord)
+
+    // check corresponding invitation ID is the request's ~thread.pthid
+
+    if (connectionRecord.invitation?.id !== message.thread?.parentThreadId) {
+      throw new ProblemReportError('Missing reference to invitation.', { problemCode: 'request_not_accepted' })
```
TimoGlastra, line 124: Throwing a `ProblemReportError` will send a generic problem report message. According to the DIDExchange RFC we should send an adopted DID Exchange or OOB problem report:

> The problem-report may be an adopted Out of Band protocol message or an adopted DID Exchange protocol message, depending on where in the processing of the invitation the error was detected.

See e.g. `ConnectionProblemReportError.ts` for an example of an adopted problem report error usage

packages/core/src/modules/connections/DidExchangeProtocol.ts
```diff
@@ -0,0 +1,462 @@
...
+        multiUseInvitation: false,
+        routing,
+        autoAcceptConnection: connectionRecord.autoAcceptConnection,
+        invitation: connectionRecord.invitation,
+        tags: connectionRecord.getTags(),
+        protocol: connectionRecord.protocol,
+      })
+    }
+
+    const peerDid = DidPeer.fromDid(message.did)
```
TimoGlastra, line 149: Maybe we should add a guard here that checks whether the did start with `did:peer:` and otherwise throw an error that public dids are not supported?

packages/core/src/modules/connections/DidExchangeProtocol.ts
```diff
@@ -0,0 +1,462 @@
...
+    routing?: Routing
+  ): Promise<DidExchangeResponseMessage> {
+    this.logger.debug(`Create message ${DidExchangeResponseMessage.type} start`, connectionRecord)
+    DidExchangeStateMachine.assertCreateMessageState(DidExchangeResponseMessage.type, connectionRecord)
+
+    // They will then either update the provisional service information to rotate the key, or provision a new DID entirely.
+    // The choice here will depend on the nature of the DID used in the invitation.
+
+    // if reuse did from invitation then do ...
+    // otherwise create new did and didDoc
```
TimoGlastra, line 197: is this outdated?

packages/core/src/modules/connections/DidExchangeProtocol.ts
```diff
@@ -0,0 +1,462 @@
...
+    // Sign message attachment
+    // Use invitationKey by default, fall back to verkey (?)
+    const [verkey] = connectionRecord.invitation?.recipientKeys || []
+
+    if (!verkey) {
+      throw new AriesFrameworkError('Connection invitation does not contain recipient key.')
+    }
+
+    const peerDidRouting = routing || {
+      endpoints: connectionRecord.invitation?.serviceEndpoint ? [connectionRecord.invitation?.serviceEndpoint] : [],
```
TimoGlastra, line 215: This should not be possible

packages/core/src/modules/connections/DidExchangeProtocol.ts
```diff
@@ -0,0 +1,462 @@
...
+      verkey,
+      did,
+      routingKeys: connectionRecord.invitation?.routingKeys || [],
```
TimoGlastra, line 218: Same here, if we don't have an invitation we should throw an error

packages/core/src/modules/connections/DidExchangeProtocol.ts
```diff
@@ -0,0 +1,462 @@
...
+    }
+
+    // TODO Currently, we just reuse invitation to create a peer did, we should also add option to create new keys
+    const peerDid = await this.createPeerDidDoc(peerDidRouting)
+    connectionRecord.did = peerDid.did
+
+    const message = new DidExchangeResponseMessage({ did: peerDid.did, threadId })
+
+    // TODO
+    // As I understood, a numAlgo 0 doesn't include service inside encoded peer did therefore we should also create a did doc attachment for it
```
TimoGlastra, line 228: numAlgo 0 shoudn't be used for DIDExchange as it can't contain a service if using static peer dids (which we do).

packages/core/src/modules/connections/DidExchangeProtocol.ts
```diff
@@ -0,0 +1,462 @@
...
+    this.logger.debug(`Process message ${DidExchangeCompleteMessage.type} start`, messageContext)
+    const { connection: connectionRecord, message } = messageContext
+
+    if (!connectionRecord) {
+      throw new AriesFrameworkError('No connection record in message context.')
+    }
+
+    DidExchangeStateMachine.assertProcessMessageState(DidExchangeCompleteMessage.type, connectionRecord)
+
+    if (connectionRecord.invitation?.id !== message.thread?.parentThreadId) {
```
TimoGlastra, line 320: Do we only need to check for parentThreadId?

packages/core/src/modules/connections/DidExchangeProtocol.ts
```diff
@@ -0,0 +1,462 @@
...
+      // TODO we should perhaps use keyAgreement instead of authentication key in here, then it must be changed also in connection record verkey
+      recipientKeys: [ed25519Key.publicKeyBase58],
+      serviceEndpoint,
+      accept: ['didcomm/aip2;env=rfc19'],
+      // It is important that we encode the routing keys as key references.
+      // So instead of using plain verkeys, we should encode them as did:key dids
+      routingKeys: mediatorRoutingKey ? [mediatorRoutingKey] : [],
+    })
+
+    const didDocument = new DidDocumentBuilder('')
```
TimoGlastra, line 381: We should remove the required id property from the did document builder, it's ugly to pass an empty string. I'll make sure to create a PR for that.

packages/core/src/modules/connections/DidExchangeProtocol.ts
```diff
@@ -0,0 +1,462 @@
...
+
+  private async createSignedAttachment(didDoc: DidDocument, verkey: string) {
+    const didDocAttach = new Attachment({
+      mimeType: 'application/json',
+      data: new AttachmentData({
+        base64: JsonEncoder.toBase64(didDoc),
+      }),
+    })
+
+    const kid = Key.fromPublicKeyBase58(verkey, KeyType.Ed25519)
```
TimoGlastra, line 416: I see I put this wrong in the example I made. 

The kid should be a did:key

```suggestion
    const key = Key.fromPublicKeyBase58(verkey, KeyType.Ed25519)
    const kid = new DidKey(key).did
```

packages/core/src/modules/connections/DidExchangeProtocol.ts
```diff
@@ -0,0 +1,462 @@
...
+
+    const didDocKeys = didDocument.authentication
+      .map((a) => {
+        if (typeof a === 'string') {
+          return a
+        }
+        // TODO return key according to type property?
+        return a.publicKeyBase58
+      })
+      .concat(invitationKeys)
```
TimoGlastra, line 454: As discusses during our call yesterday. If `a` is of type string it means it is a reference to an item in the `verificationMethod` array. We should dereference it.

But while we're at it, it would be good to use the generic way to get the base58Key:

```suggestion
    const didDocKeys = didDocument.authentication
      .map((authentication) => {
        const dereferenced = typeof authentication === 'string' ? didDocument.dereferenceKey(authentication) : authentication))
        const { getKeyFromVerificationMethod } = getKeyDidMappingByVerificationMethod(entry)
        const key = getKeyFromVerificationMethod(entry)
        
        return key.publicKeyBase58
      })
      .concat(invitationKeys)
```

packages/core/src/modules/connections/DidExchangeProtocol.ts
```diff
@@ -0,0 +1,462 @@
...
+
+    this.logger.trace('JWS verification result', { isValid, signerVerkeys, didDocKeys })
+
+    if (!isValid || signerVerkeys.length > 1 || !signerVerkeys.every((verkey) => didDocKeys.includes(verkey))) {
```
TimoGlastra, line 458: Why is it only allowed to be signed by 1 verkey? What if the invitation key was different from the key used in the exchange request/response?

packages/core/src/modules/connections/ConnectionsModule.ts
```diff
@@ -86,6 +96,7 @@ export class ConnectionsModule {
       autoAcceptConnection?: boolean
       alias?: string
       mediatorId?: string
+      protocol?: string
```
TimoGlastra, line 99: How will this work when protocol is didexchange?

I only accepts an 0160 invitation, but allows a string value protocol.

TimoGlastra (changes requested):
Great work @jakubkoci! :)

I left a bunch of comments, maybe we don't have to resolve them all in this PR, but I do think there are some important things to fix.

packages/core/src/modules/connections/DidExchangeProtocol.ts
jakubkoci: Updated.

packages/core/src/modules/connections/DidExchangeProtocol.ts
jakubkoci: Yeah, thanks for that code snippet 👍  It would take me much longer to understand how to compose those functions to get the key. Updated.

packages/core/src/modules/connections/handlers/DidExchangeResponseHandler.ts
```diff
@@ -0,0 +1,63 @@
...
+    // A connection request message is the only case when I can use the connection record found
+    // only based on recipient key without checking that `theirKey` is equal to sender key.
+    //
+    // The question is if we should do it here in this way or rather somewhere else to keep
+    // responsibility of all handlers aligned.
+    //
+    messageContext.connection = connectionRecord
+    const connection = await this.didExchangeProtocol.processResponse(messageContext)
+
+    // TODO: should we only send ping message in case of autoAcceptConnection or always?
```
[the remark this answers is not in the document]
jakubkoci: Good point, but shouldn't it be actually `complete` instead of `ack`?

packages/core/src/modules/connections/DidExchangeProtocol.ts
jakubkoci: 👍  Updated.

packages/core/src/modules/connections/DidExchangeProtocol.ts
```diff
@@ -0,0 +1,486 @@
...
+    })
+    const x25519VerificationMethod = getX25519VerificationMethod({
+      id: uuid(),
+      key: x25519Key,
+      controller: '#id',
+    })
+
+    let mediatorRoutingKey
+    if (routing.routingKeys.length > 0) {
+      const [mediatorPublicKeyBase58] = routing.routingKeys
```
jakubkoci, line 351: As I was working on iteration over endpoints as mentioned below I found out that this is a similar problem. I should perhaps transform all routing keys in this way, right? I'm just not sure why we're using did:key for routing keys and not for recipient keys. 🤔

TimoGlastra: Hmm you're right that doesn't make sense. From looking at the did:key RFC we shouldn't use did:key encoding at all when using did:peer did docs, but rather use a key reference. However what should we put there in that case? We don't have a key reference for the mediator routing keys.

I'm confused with the mediator coordination protocol, which requires the routing keys to be encoded as did:keys: https://github.com/hyperledger/aries-rfcs/blob/main/features/0211-route-coordination/README.md#mediation-grant.

However the [did:key RFC](https://github.com/hyperledger/aries-rfcs/blob/main/features/0360-use-did-key/README.md) mentions:

> Service entries in did:peer DIDDocs (such as in RFCs [0094-cross-domain-messaging](https://github.com/hyperledger/aries-rfcs/tree/main/concepts/0094-cross-domain-messaging) and [0067-didcomm-diddoc-conventions](https://github.com/hyperledger/aries-rfcs/tree/main/features/0067-didcomm-diddoc-conventions)) should NOT use a did:key public key representation. Instead, service entries in the DIDDoc should reference keys defined internally in the DIDDoc where appropriate.

It's not always possible to include key references from did docs, as that would mean the mediator needs to use a public did. So in that case I'm guessing we should use a did:key reference so we can still make a resolvable key from the routing keys. If we directly use the publicKeyBase58 there's no way to determine which key type is used. 

See also this issue for a relevant discussion: https://github.com/hyperledger/aries-rfcs/issues/704

Basically boils down to:
- DIDComm v2 will support any type of did for the routingKeys, allowing mediators to rotate keys etc.. without needing to communicate with the recipient
- DIDComm v2 doesn't have the recipientKeys anymore, so it's mostly a leftover legacy thing. Also having recipientKeys outside of the did document makes less sense than having routingKeys outside of the did document 


The encoding of the `recipientKeys` should be key references.  So instead of this:
```
recipientKeys: [ed25519Key.publicKeyBase58],
```

it should be this:

```
recipientKeys: [`#${ed25519VerificationMethod.id}`],
```

Well actually, this:

```
recipientKeys: [`#${x25519VerificationMethod.id}`],
```

but that will cause other problems

packages/core/src/modules/routing/RecipientModule.ts
```diff
@@ -288,13 +289,15 @@ export class RecipientModule {
         routing,
       })
       this.logger.debug('Processed mediation invitation', {
-        connectionId: invitationConnectionRecord,
```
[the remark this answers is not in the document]
jakubkoci: I think it's not bad to use a module from another module but only when it's one-way communication through module API.

packages/core/src/modules/connections/DidExchangeProtocol.ts
```diff
@@ -0,0 +1,462 @@
...
+      const mediatorEd25519DidKey = new DidKey(mediatorEd25519Key)
+      const mediatorX25519Key = Key.fromPublicKey(
+        convertPublicKeyToX25519(mediatorEd25519Key.publicKey),
+        KeyType.X25519
+      )
+      // Use ed25519 did:key, which also includes the x25519 key used for didcomm
+      mediatorRoutingKey = `${mediatorEd25519DidKey.did}#${mediatorX25519Key.fingerprint}`
+    }
+
+    // TODO Iterate over all endpoints
```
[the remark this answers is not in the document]
jakubkoci: Updated.

packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -73,15 +75,29 @@ export class ConnectionService {
...
     myLabel?: string
     myImageUrl?: string
+    protocol?: string
   }): Promise<ConnectionProtocolMsgReturnType<ConnectionInvitationMessage>> {
     // TODO: public did
+
+    let role
+    let state
+
+    if (config?.protocol === HandshakeProtocol.DidExchange) {
```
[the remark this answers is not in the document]
jakubkoci: I agree. But that would mean we would need `state` and `role` empty, right? Or should we set it to `DidExchange` from now on?

packages/core/tests/oob.test.ts
```diff
@@ -141,10 +143,9 @@ describe('out of band', () => {
 
     test('create OOB message only with requests', async () => {
       const { offerMessage } = await faberAgent.credentials.createOutOfBandOffer(credentialTemplate)
-      const { outOfBandMessage, connectionRecord } = await faberAgent.oob.createMessage(
-        { label: 'test-connection', handshake: false },
-        [offerMessage]
-      )
+      const { outOfBandMessage, connectionRecord } = await faberAgent.oob.createMessage({ label: 'test-connection' }, [
+        offerMessage,
+      ])
```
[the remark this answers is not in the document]
jakubkoci: 👍  Updated.

packages/core/src/modules/connections/handlers/DidExchangeResponseHandler.ts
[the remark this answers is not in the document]
TimoGlastra: Yes! :)

packages/core/src/modules/connections/services/ConnectionService.ts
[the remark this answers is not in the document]
TimoGlastra: Yeah that sounds good! Maybe we can do a transformer for now (map 0160 state to didexchange state, same with role)?

Something like the metadata transformer for credential metadata: https://github.com/hyperledger/aries-framework-javascript/blob/main/packages/core/src/utils/transformers.ts#L54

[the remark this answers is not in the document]
TimoGlastra: Or we should have the migration script ready at the same time we merge this PR

packages/core/src/modules/connections/DidExchangeProtocol.ts
jakubkoci: Do you mean that we should throw an error when there is no `routing` and no `connectionRecord.invitation`? Or something else?

packages/core/src/modules/connections/DidExchangeProtocol.ts
jakubkoci: It could be responsibility of `DidPeer` class. I added it anyway, because we would still need some error handling here so that check inside `DidPeer` wouldn't help so much.

packages/core/src/modules/connections/repository/ConnectionRecord.ts
```diff
@@ -20,8 +19,8 @@ export interface ConnectionRecordProps {
   theirDidDoc?: DidDoc
   theirLabel?: string
   invitation?: ConnectionInvitationMessage
-  state: ConnectionState
-  role: ConnectionRole
+  state: ConnectionState | DidExchangeState
+  role: ConnectionRole | DidExchangeRole
```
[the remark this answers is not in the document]
jakubkoci: Hmm. That does not seem so good. These are part of the protocol and protocol defines values of them. If I'm using older `connections` protocol I would expect the same values as they're defined in the spec. What about something like protocol metadata? 

I realize that in this context, we don't assume the older protocol to be there forever, it's hard to say how long it will be, hard to predict from my perspecitve.

packages/core/src/modules/connections/DidExchangeProtocol.ts
```diff
@@ -0,0 +1,462 @@
...
+      .build()
+
+    const peerDid = DidPeer.fromDidDocument(didDocument, PeerDidNumAlgo.GenesisDoc)
+
+    const didRecord = new DidRecord({
+      id: peerDid.did,
+      role: DidDocumentRole.Created,
+      // It is important to take the did document from the PeerDid class
+      // as it will have the id property
+      // Should not we also resolve and store document for inline peer did?
```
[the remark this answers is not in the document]
jakubkoci: I don't remember why I wrote it 😄 But removed

packages/core/src/modules/connections/DidExchangeProtocol.ts
jakubkoci: 👍  Removed

packages/core/src/modules/connections/DidExchangeProtocol.ts
jakubkoci: Yes, updated 👍

packages/core/src/modules/connections/DidExchangeProtocol.ts
jakubkoci: Yes, I updated it to support only numalgo 1 for now.

packages/core/src/modules/connections/DidExchangeProtocol.ts
jakubkoci: Good catch. I added check for `threadId` and also added the same check to `processResponse` method.

packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -662,7 +690,7 @@ export class ConnectionService {
     // IndyAgentService is old service type
     const services = endpoints.map(
       (endpoint, index) =>
-        new IndyAgentService({
+        new DidCommService({
```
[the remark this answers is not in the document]
jakubkoci: Reverted

packages/core/src/modules/connections/repository/ConnectionRecord.ts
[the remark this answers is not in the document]
TimoGlastra: I tend to disagree. As a framework user I want the API to be as simple as possible. Having two state values that basically mean the same but have different values is not helping towards making a simple API. I think having a single state (our framework state) that is the default that can map to the 0160 state would be more helpful here. Especially having a single value to store will reduce complexity around the framework (then we don't need the `state === Complete || state === Completed` which is pointless).

If someone still wants to work with the 0160 state values, we could add a property that returns the 0160 state derived from the 0023 state if the user would want that. I would like to be able to use the DIDExchange and Connection protocol with as minimal API differences as possible. We call it the connections module because it allows me to create connections, whether that is through the DIDExchange or Connection protocol is irrelevant most of the time.

So maybe we can look at a way where:
- We only store a single value
- We only use that value around the framework to reduce complexity of state checks
- We add convenience methods to get the state in 0160 notation

What do you think of this approach?

packages/core/src/modules/oob/OutOfBandModule.ts
```diff
@@ -67,19 +67,28 @@ export class OutOfBandModule {
...
+   *  }
+   *  const message = outOfBandModule.createMessage(config)
+   * ```
+   *
+   * Then, the out-of-band will use its keys and will work as a connection invitation.
+   *
+   * @param config Configuration and other attributes of out-of-band message
    * @param messages Messages that will be sent inside out-of-band message
-   * @returns Out-of-band message and connection record if it has been created.
+   * @returns Out-of-band message and optionally connection record created based on `handshakeProtocol`
```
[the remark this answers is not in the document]
jakubkoci: Updated. Now, if `handshake` is set it assigns handhsake protocols by default and it's possible to override it with `handshakeProtocols` array.

I still need to figure out how to avoid creating the connection record with a particular protocol. But that's a different topic being discussed elsewhere.

packages/core/src/modules/connections/repository/ConnectionRecord.ts
```diff
@@ -30,12 +29,13 @@ export interface ConnectionRecordProps {
   multiUseInvitation: boolean
   mediatorId?: string
   errorMessage?: string
+  protocol?: string
```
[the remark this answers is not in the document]
jakubkoci: Maybe for now. But, don't we eventually want to get into a situation when protocols are pluggable in runtime?

packages/core/src/modules/connections/repository/ConnectionRecord.ts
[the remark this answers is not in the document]
jakubkoci: Let me just reiterate if I understand what you suggest. You also have a related suggestion here https://github.com/hyperledger/aries-framework-javascript/pull/615/files#r796613625, so I want to make it clear and in one place:

* We use `DidExchange` roles and states.
* OOB methods sets initial values in the following way:
`createMessage` **Responder**, invitation-sent
`receiveMessage` **Requester**, invitation-received
* We transform old Connection values to DidExchange values. There are 2 options:
	1. We prepare migration script to transform old roles and state to the new ones. This will force any current user to update event listeners to new API.
	2. We transform old Connection protocol values to DidExchange values if user requires to do that. How do we do that?

Transition table from DidExchange to Connection will look like this:
* start -> **null**
No connection exists or is in progress

* invitation-received, invitation-sent -> **invited**
The invitation has been shared with the intended *invitee*(s), and they have not yet sent a *connection_request*.

* request-sent, request-received -> **requested**
A *connection_request* has been sent by the *invitee* to the *inviter* based on the information in the *invitation*.

* response-sent, response-received  -> **responded**
A *connection_response* has been sent by the *inviter* to the *invitee* based on the information in the *connection_request*.

* completed -> **complete**

* abandoned -> ??? (`null`?)

packages/core/src/modules/connections/repository/ConnectionRecord.ts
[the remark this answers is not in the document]
jakubkoci: Ok, I'm good with that. 👍

packages/core/src/modules/connections/handlers/DidExchangeResponseHandler.ts
```diff
@@ -0,0 +1,63 @@
...
+    this.connectionService = connectionService
+    this.agentConfig = agentConfig
+  }
+
+  public async handle(messageContext: HandlerInboundMessage<DidExchangeResponseHandler>) {
+    if (!messageContext.recipientVerkey || !messageContext.senderVerkey) {
+      throw new AriesFrameworkError('Unable to process connection request without senderVerkey or recipientVerkey')
+    }
+
+    const connectionRecord = await this.connectionService.findByVerkey(messageContext.recipientVerkey)
```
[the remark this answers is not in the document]
jakubkoci: But we need a whole connection record, don't we? 🤔 Do you mean something like this:

```ts
const didRecord = await this.dids.findByVerkey(messageContext.recipientVerkey)
const connectionRecord =  await this.connectionService.findByDid(didRecord.did)
```

packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -639,9 +662,13 @@ export class ConnectionService {
...
   }
 
-  private async createConnection(options: {
-    role: ConnectionRole
-    state: ConnectionState
+  public async findByTheirDid(did: string): Promise<ConnectionRecord | null> {
+    return this.connectionRepository.findSingleByQuery({ theirDid: did })
+  }
+
+  public async createConnection(options: {
```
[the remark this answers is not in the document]
jakubkoci: It's currently used in OOB createMessage, but I agree we should get rid of that and then eventually update the createConnection method

packages/core/src/modules/oob/OutOfBandModule.ts
```diff
@@ -307,14 +337,18 @@ export class OutOfBandModule {
...
       options = {
-        ...service,
+        recipientKeys: service.recipientKeys,
+        serviceEndpoint: service.serviceEndpoint,
+        routingKeys: service.routingKeys,
       }
     }
 
-    const invitation = new ConnectionInvitationMessage({ label, ...options })
+    const invitation = new ConnectionInvitationMessage({ id: outOfBandMessage.id, label, ...options })
```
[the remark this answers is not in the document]
jakubkoci: I think it's more related to OOB record and DIDComm message storage

packages/core/tests/oob.test.ts
```diff
@@ -341,6 +371,13 @@ describe('out of band', () => {
         reuseConnection: true,
       })
 
+      await aliceAgent.connections.returnWhenIsConnected(secondAliceFaberConnection?.id || '')
```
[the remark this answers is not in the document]
jakubkoci: Agree, I'll look at it. I don't think it's crucial for this PR.

packages/core/src/modules/connections/handlers/DidExchangeResponseHandler.ts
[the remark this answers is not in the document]
TimoGlastra: Yes, exactly!

packages/core/tests/oob.test.ts
```diff
@@ -371,10 +408,21 @@ describe('out of band', () => {
...
       expect(faberConnections).toHaveLength(3)
       expect(faberAliceConnection?.multiUseInvitation).toBe(true)
-      expect(faberAliceConnection?.state).toBe(ConnectionState.Invited)
+      // expect(faberAliceConnection?.state).toBe(ConnectionState.Invited)
+      expect(faberAliceConnection?.state).toBe(DidExchangeState.InvitationSent)
 
       expect(firstConnection.getTag('invitationKey')).toEqual(faberAliceConnection?.verkey)
       expect(secondConnection.getTag('invitationKey')).toEqual(faberAliceConnection?.verkey)
+
+      // We need to wait for the faber connection to be completed because there is a clean up
```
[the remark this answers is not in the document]
jakubkoci: The test calls `returnWhenIsConnected` only for the Alice side and not the Faber side.

packages/core/src/modules/connections/repository/ConnectionRecord.ts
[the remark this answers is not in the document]
TimoGlastra: Yes that's what I'm suggesting. Re the transformation to connection state I would suggest adding a new computed property that returns the 0160 state

So this:

```
agent.events.on<ConnectionStateChangedEvent>(ConnectionEventTypes.StateChanged, (event) => {
    event.payload.record.state == ConnectionState.Requested
})
```

becomes this:

```
agent.events.on<ConnectionStateChangedEvent>(ConnectionEventTypes.StateChanged, (event) => {
    event.payload.record.rfc160State == ConnectionState.Requested
})
```

packages/core/src/modules/connections/DidExchangeProtocol.ts
```diff
@@ -0,0 +1,462 @@
...
+    // TODO Iterate over all endpoints
+    const [serviceEndpoint] = routing.endpoints
+    const service = new DidCommService({
+      id: '#service-0',
+      // Fixme: can we use relative reference (#id) instead of absolute reference here (did:example:123#id)?
+      // We don't know the did yet
+      // TODO we should perhaps use keyAgreement instead of authentication key in here, then it must be changed also in connection record verkey
+      recipientKeys: [ed25519Key.publicKeyBase58],
+      serviceEndpoint,
+      accept: ['didcomm/aip2;env=rfc19'],
```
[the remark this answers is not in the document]
jakubkoci: Updated

packages/core/src/modules/connections/DidExchangeProtocol.ts
jakubkoci: What other problems do you have on your mind? I guess one is that we cannot convert x25519 to ed25519 needed by Indy?

It seems that ACA-Py supports just [ed25519 base58](https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/connections/base_manager.py#L139).

I aligned the recipient and routing keys, it's using `ed25519Key.publicKeyBase58` in service for both. I also added the routing key into the did doc as a verification method.

This is tested with added oob mediation test. I think it's a good starting point before we finalize how it should work according to the current and future crypto libs and protocols.

jakubkoci:
Hi, @TimoGlastra I addressed many of your suggestions. Thank you for the great feedback and for filling the gaps in my understanding of protocols 👍  

The rest is mainly about:
* Better separation of OOB with DID exchange and Connections protocol. Replacing old invitation with the new one, introducing OOB record and/or storing OOB message.
* Better did-exchange vs. connection role and state management.

I suggest merging this PR to `feat/out-of-band` and continuing in separate PRs.

TimoGlastra (approved):
Great! Sounds good @jakubkoci

Issue #587, opened 2021-12-22
check credential preview attributes against schema attributes before making credential offer


TimoGlastra:
When issuing a credential for a schema/credential definition with attribute names that do not match the schema, a vague `CommonInvalidStructure` error will be thrown when issuing the credential.

We should add a check when offering a credential (`CredentialService.createOffer` and `CredentialService.createOfferAsResponse`) to make sure the attributes from the preview match the attributes from the schema. 

Implementation in ACA-Py: https://github.com/hyperledger/aries-cloudagent-python/blob/cd5588f79beb7314f03d78bdfb979d1d6b99acd8/aries_cloudagent/protocols/issue_credential/v1_0/manager.py#L274-L283

```
ERROR [undefined.<anonymous>] Error processing inbound message: IndyError(CommonInvalidStructure): CommonInvalidStructure 
IndySdkError: IndyError(CommonInvalidStructure): CommonInvalidStructure
    at IndyIssuerService.createCredential (/Users/annelein/Animo/afj-demo/demo/node_modules/@aries-framework/core/src/modules/indy/services/IndyIssuerService.ts:127:15)
    at async CredentialService.createCredential (/Users/annelein/Animo/afj-demo/demo/node_modules/@aries-framework/core/src/modules/credentials/services/CredentialService.ts:571:26)
    at async RequestCredentialHandler.createCredential (/Users/annelein/Animo/afj-demo/demo/node_modules/@aries-framework/core/src/modules/credentials/handlers/RequestCredentialHandler.ts:41:43)
    at async RequestCredentialHandler.handle (/Users/annelein/Animo/afj-demo/demo/node_modules/@aries-framework/core/src/modules/credentials/handlers/RequestCredentialHandler.ts:29:14)
    at async Dispatcher.dispatch (/Users/annelein/Animo/afj-demo/demo/node_modules/@aries-framework/core/src/agent/Dispatcher.ts:47:25)
    at async MessageReceiver.receiveEncryptedMessage (/Users/annelein/Animo/afj-demo/demo/node_modules/@aries-framework/core/src/agent/MessageReceiver.ts:134:5)
    at async MessageReceiver.receiveMessage (/Users/annelein/Animo/afj-demo/demo/node_modules/@aries-framework/core/src/agent/MessageReceiver.ts:73:7)
[...]

PR #625, opened 2022-02-07
fix: credential preview attributes mismatch schema attributes

Annelein:
Solves #587 

TO DO:
- [ ] Write tests

### DIFF
--- a/packages/core/src/modules/credentials/services/CredentialService.ts
+++ b/packages/core/src/modules/credentials/services/CredentialService.ts
@@ -244,6 +244,17 @@ export class CredentialService {
       }),
     })
 
+    // Check if credential preview attributes match the schema attributes
+    const schema = await this.ledgerService.getSchema(credOffer.schema_id)
+    const credAttributes = credentialTemplate.preview.attributes
+    const schemaAttribute = schema.attrNames
+
+    credAttributes.forEach((credAtrr) => {
+      if (schemaAttribute.indexOf(credAtrr.name) === -1) {
+        throw new AriesFrameworkError(`The credential preview attributes do not match the schema attributes`)
+      }
+    })
+
     const credentialOfferMessage = new OfferCredentialMessage({
       comment,
       offerAttachments: [offerAttachment],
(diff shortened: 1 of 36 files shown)


packages/core/src/modules/credentials/services/CredentialService.ts
```diff
@@ -244,6 +244,24 @@ export class CredentialService {
...
+      if (schemaAttributes.indexOf(credAtrr) === -1) {
+        difference.push(credAtrr)
+      }
+    })
+
+    if (difference.length > 0) {
+      throw new AriesFrameworkError(
+        `The credential preview attributes do not match the schema attributes (difference is: ${difference}, needs: ${schemaAttributes})`
+      )
+    }
```
TimoGlastra, line 263: Might be good to extract this to a separate private method that is used in both createOffer and createOfferAsResponse to avoid duplication

TimoGlastra (changes requested):
One small comment, and as discussed would be good to also add a test for `createOfferAsResponse`.

TimoGlastra:
Also: tests are failing. Seems like an error with credential-auto-accept.test.ts that is now thrown because of this new check

packages/core/src/modules/credentials/services/CredentialService.ts
```diff
@@ -244,6 +244,24 @@ export class CredentialService {
...
     })
 
+    // Check if credential preview attributes match the schema attributes
+    const schema = await this.ledgerService.getSchema(credOffer.schema_id)
+    const credAttributes = credentialTemplate.preview.attributes.map((a) => a.name)
+    const schemaAttributes = schema.attrNames
+
+    const difference: string[] = []
+    credAttributes.forEach((credAtrr) => {
+      if (schemaAttributes.indexOf(credAtrr) === -1) {
```
jakubkoci, line 254: Small suggestion
```suggestion
      if (!schemaAttributes.includes(credAtrr)) {
```

packages/core/src/modules/credentials/services/CredentialService.ts
```diff
@@ -297,6 +315,24 @@ export class CredentialService {
...
+        difference.push(credAtrr)
+      }
+    })
```
TimoGlastra, line 328: Looking at this again this is not where we landed yesterday right? I think we took the example of "symmetric difference". Currently if you have a property missing in the credAttributes array, it won't throw an error.

https://stackoverflow.com/questions/1187518/how-to-get-the-difference-between-two-arrays-in-javascript

This will find _any_ differences between the two arrays:

```suggestion
    const difference = credAttributes
                 .filter(x => !schemaAttributes.includes(x))
                 .concat(schemaAttributes.filter(x => !credAttributes.includes(x)));
```

TimoGlastra (approved):

Issue #569, opened 2021-12-08
Check for "REQNACK" response from indy ledger


TimoGlastra:
Currently we check if a transaction failed by checking for the `REJECT` value on response.op, but I noticed that it is also possible to get a `REQNACK` response if you try to write something to the ledger while that did is not registered on the ledger.

See logs for more info

```
server_1  | 2021-12-08 08:35:06.777  DEBUG [IndyLedgerService.registerCredentialDefinition] Registered credential definition 'SK8di1uRdeC8e1zMksoHUz:3:CL:95154:Crypto Wallet' on ledger 'BCOVRIN_TEST_GENESIS'
server_1  | {
server_1  |   response: {
server_1  |     op: 'REQNACK',
server_1  |     identifier: 'SK8di1uRdeC8e1zMksoHUz',
server_1  |     reqId: 1638952506203019800,
server_1  |     reason: 'client request invalid: could not authenticate, verkey for SK8di1uRdeC8e1zMksoHUz cannot be found'
server_1  |   },
server_1  |   credentialDefinition: {
server_1  |     ver: '1.0',
server_1  |     id: 'SK8di1uRdeC8e1zMksoHUz:3:CL:95154:Crypto Wallet',
server_1  |     schemaId: '95154',
server_1  |     type: 'CL',
server_1  |     tag: 'Crypto Wallet',
server_1  |     value: {
server_1  |       primary: {
server_1  |         n: '97743451841326778227641653730537624191340665687172518469820133433815801000797936061661012472341235246949095122217732134658777546977245327358514710072351592210846621971323039750446364590824983220814323674668208979123426482465763971908429674125375262077508259867958928969835469223077106244655103284044268209307669206155967742710267167093819266178336358107283136000277023496128496566969144343710489058530163886756406417723937553385321530288380419605941224464397998770862506756847680765563246449553960409132429261481014057069890877186503339169247734719573259692689960149406054935083621200297829711769841873461908009439301',
[...]

PR #626, opened 2022-02-07
feat: check for "REQNACK" response from indy ledger 

Annelein:
Solves #569 

### DIFF
--- a/packages/core/src/modules/indy/services/IndyHolderService.ts
+++ b/packages/core/src/modules/indy/services/IndyHolderService.ts
@@ -201,7 +201,7 @@ export interface StoreCredentialOptions {
   credential: Indy.Cred
   credentialDefinition: Indy.CredDef
   credentialId?: Indy.CredentialId
-  revocationRegistryDefinitions?: Indy.RevRegsDefs
+  revocationRegistryDefinitions?: Indy.RevocRegDef
 }
 
 export interface CreateProofOptions {
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -26,7 +26,7 @@
     "@multiformats/base-x": "^4.0.1",
     "@stablelib/ed25519": "^1.0.2",
     "@stablelib/sha256": "^1.0.1",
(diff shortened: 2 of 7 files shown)


TimoGlastra (approved):
LGTM!

packages/core/src/modules/indy/services/IndyHolderService.ts
```diff
@@ -201,7 +201,7 @@ export interface StoreCredentialOptions {
   credential: Indy.Cred
   credentialDefinition: Indy.CredDef
   credentialId?: Indy.CredentialId
-  revocationRegistryDefinitions?: Indy.RevRegsDefs
+  revocationRegistryDefinitions?: Indy.RevocRegDef
```
TimoGlastra, line 204:
```suggestion
  revocationRegistryDefinitions?: Indy.RevocRegDefs
```

TimoGlastra (changes requested):
Just spotted a small typo

packages/core/src/modules/indy/services/IndyHolderService.ts
Annelein: It should actually be `RevocRegDef` otherwise there is an error saying type `RevocRegDefs` of `revocationRegistryDefinitions ?? null `does not match `RevocRegDef`
`  /**
   * Store a credential in the wallet.
   *
   * @returns The credential id
   */
  public async storeCredential({
    credentialRequestMetadata,
    credential,
    credentialDefinition,
    credentialId,
    revocationRegistryDefinitions,
  }: StoreCredentialOptions): Promise<Indy.CredentialId> {
    try {
      return await this.indy.proverStoreCredential(
        this.wallet.handle,
        credentialId ?? null,
        credentialRequestMetadata,
        credential,
        credentialDefinition,
        revocationRegistryDefinitions ?? null
      )
    } catch (error) {
      throw new IndySdkError(error)
    }
  }`

TimoGlastra (approved):

PR #632, opened 2022-02-08
fix: leading zeros in credential value encoding

JamesKEbert:
This PR resolves an interop issue between ACA-Py & AFJ—in credential issuance we verify that the credential did not change between what was offered and what is actually being issued. 

If the credential attribute value to be encoded (as described [here](https://github.com/hyperledger/aries-rfcs/blob/be4ad0a6fb2823bb1fc109364c96f077d5d8dffa/features/0036-issue-credential/README.md#encoding-claims-for-indy-based-verifiable-credentials)) is a string of numbers that contains a leading zero, such as an account number like: `08888`, then ACA-Py removes the leading zero (`8888`), while AFJ instead encodes it with the zero (`08888`). This PR adjusts AFJ’s behavior to bring it in line with ACA-Py’s implementation.

Signed-off-by: James Ebert <jamesebert.k@gmail.com>

### DIFF
--- a/packages/core/src/modules/credentials/CredentialUtils.ts
+++ b/packages/core/src/modules/credentials/CredentialUtils.ts
@@ -153,7 +153,7 @@ export class CredentialUtils {
 
     // If value is an int32 number string return as number string
     if (isString(value) && !isEmpty(value) && !isNaN(Number(value)) && this.isInt32(Number(value))) {
-      return value
+      return Number(value).toString()
     }
 
     if (isNumber(value)) {
--- a/packages/core/src/modules/credentials/__tests__/CredentialUtils.test.ts
+++ b/packages/core/src/modules/credentials/__tests__/CredentialUtils.test.ts
@@ -74,6 +74,10 @@ const testEncodings: { [key: string]: { raw: string | number | boolean | null; e
     raw: '0.1',
     encoded: '9382477430624249591204401974786823110077201914483282671737639310288175260432',
   },
+  'leading zero number string': {
+    raw: '012345',
+    encoded: '12345',
+  },
   'chr 0': {
     raw: String.fromCharCode(0),
     encoded: '49846369543417741186729467304575255505141344055555831574636310663216789168157',


blu3beri (approved):
Don't you think we should change this in aca-py? Keeping a number as a string might be done to keep the leading zeros for some, no clue which, use cases. The RFC does not specify to remove leading zeros if provided in a number string. 

Maybe I miss something, but it seems more to me that removing characters from a string is unexpected.

JamesKEbert:
It is worth while noting that this is just the encoding, vs the final value, so as long as the encodings match I think it'd be okay. 
Also, while not the most clear, the following grabbed from the RFC does indicate that it should convert to an int (or number in the JS case) and then stringify as a number, which is not the behavior as currently implemented (which I believe would result in the leading zero being dropped). 
> * convert the resulting digest bytes, big-endian, to integer
> * stringify the integer as a decimal.

I am all ears if I am mistaken though! :)

TimoGlastra (approved):
That makes sense @JamesKEbert

TimoGlastra:
Could you maybe add a test case to [this](https://github.com/hyperledger/aries-framework-javascript/blob/main/packages/core/src/modules/credentials/__tests__/CredentialUtils.test.ts#L8) file so this won't happen again in the future? :)

JamesKEbert:
Can't believe I missed that file as a thing lol--test case added, happy to adjust based off of nit-picks if any.

TimoGlastra:
@JamesKEbert I just encountered an issue with this when testing for interop between AFJ 0.1.0 and 0.2.0. Now that I understand the issue better I'm not sure if we actually need to drop the 0 in this case. 

E.g. in our demo we issue a "credit card" credential where the security code has a value `063`. AFJ 0.1.0 will encode that as `063`, while AFJ 0.2.0 and ACA-Py will encode it as `63`. After thinking about it more, aren't both ACA-Py and AFJ 0.1.0 + AFJ 0.2.0 wrong? Shouldn't we take the 'this is not a number' path and create a hash of the value and convert that to a number?

In this case we should really threat the value as a string. This is of course horrible for interop with ACA-Py and older AFJ versions, but curious to hear what you think. Anyway by fixing this we now have broken interop with older versions of AFJ which is of course not great (although only if you issue credentials with number strings that have leading 0's)

Issue #586, opened 2021-12-22
proof request attribute groupName is the same as another one


blu3beri:
When we receive a proofrequest it should not be possible for any `groupName` to be the same. e.g. When we receive a proof request with groupName `foo`, `bar` and `foo` it will not retrieve the correct credentials from Indy as we include `foo` twice.

This should be fixed IMO when generating a proof request on the verifier side. However it might be very useful to check if there are non-unique keys when receiving a proof request and logging a warning or handling it at least in some way. 

tl;dr 

We can receive proof request with non-unique `groupNames`. This will result in unexpected behaviour and we will not be able to fulfil  the proof request.

Places we need to check for the proof request keys: (proofService)
- When we receive a proof request
  - processRequest in service
- When we send a proof request
   - createRequest
   - createRequestAsResponse 

What to check:
- Check whether proof request attributes keys and predicate keys do not overlap.

PR #638, opened 2022-02-14
fix: check groupnames

Annelein:
Solves #586

### DIFF
--- a/packages/core/src/modules/proofs/services/ProofService.ts
+++ b/packages/core/src/modules/proofs/services/ProofService.ts
@@ -254,8 +254,8 @@ export class ProofService {
       comment?: string
     }
   ): Promise<ProofProtocolMsgReturnType<RequestPresentationMessage>> {
-    // Check if Attribute names match Predicate names
-    this.checkAttributePredicateMatch(proofRequest)
+    // Assert attribute and predicate group names do not match
+    this.assertAttributePredicateGroupNamesDoNotMatch(proofRequest)
 
     // Assert
     proofRecord.assertState(ProofState.ProposalReceived)
@@ -302,8 +302,8 @@ export class ProofService {
   ): Promise<ProofProtocolMsgReturnType<RequestPresentationMessage>> {
     this.logger.debug(`Creating proof request`)
 
-    // Check if Attribute names match Predicate names
-    this.checkAttributePredicateMatch(proofRequest)
+    // Assert attribute and predicate group names do not match
+    this.assertAttributePredicateGroupNamesDoNotMatch(proofRequest)
 
     // Assert
     connectionRecord?.assertReady()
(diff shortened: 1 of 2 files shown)


packages/core/src/modules/proofs/services/ProofService.ts
```diff
@@ -1037,6 +1046,16 @@ export class ProofService {
...
+
+  public checkAttributePredicateMatch(proofRequest: ProofRequest) {
+    for (const attribute of proofRequest.requestedAttributes.keys()) {
+      for (const predicate of proofRequest.requestedPredicates.keys()) {
+        if (attribute === predicate) {
+          throw new AriesFrameworkError(`The proof request contains attributes that match the predicates: ${attribute}`)
+        }
+      }
+    }
+  }
```
TimoGlastra, line 1058: For each key it will loop through the keys of the other array. Maybe we can use `.includes` here?

```suggestion
    const predicateGroupNames = proofRequest.requestedPredicates.keys()
    for (const attribute of proofRequest.requestedAttributes.keys()) {
      if (predicateGroupNames.includes(attribute)) {
        throw new AriesFrameworkError(`The proof request contains attributes that match the predicates:  ${attribute}`)
        }
      }
  }
```

packages/core/tests/proofs.test.ts
```diff
@@ -331,4 +331,40 @@ describe('Present Proof', () => {
       presentationMessage: expect.any(PresentationMessage),
     })
   })
+
+  test('Faber starts with proof request to Alice, but the attributes match with the predicates', async () => {
```
TimoGlastra, line 335:
```suggestion
  test('Faber starts with proof request to Alice, but an attribute group name matches with a predicate group name so an error is thrown', async () => {
```

packages/core/src/modules/proofs/services/ProofService.ts
```diff
@@ -1037,6 +1046,16 @@ export class ProofService {
 
     return credentialDefinitions
   }
+
+  public checkAttributePredicateMatch(proofRequest: ProofRequest) {
+    for (const attribute of proofRequest.requestedAttributes.keys()) {
+      for (const predicate of proofRequest.requestedPredicates.keys()) {
+        if (attribute === predicate) {
+          throw new AriesFrameworkError(`The proof request contains attributes that match the predicates: ${attribute}`)
```
TimoGlastra, line 1054:
```suggestion
          throw new AriesFrameworkError(`The proof request contains an attribute group name that matches a predicate group name: ${attribute}`)
```

packages/core/src/modules/proofs/services/ProofService.ts
```diff
@@ -360,6 +366,9 @@ export class ProofService {
     }
     await validateOrReject(proofRequest)
 
+    // Check if Attribute names match Predicate names
+    this.checkAttributePredicateMatch(proofRequest)
```
TimoGlastra, line 370: This comment makes me believe we want them to match. 

Maybe something like:

```suggestion
    // Assert attribute and predicate group names do not match
    this.assertAttributePredicateGroupNamesDoNotMatch(proofRequest)
```

Not sure if doNotMatch is the best, but is is explicit

TimoGlastra (changes requested):
Some suggestions on wording and structure, but overall looks good!

TimoGlastra (approved):
Nice

Issue #550, opened 2021-11-25
Check if message has structure of packed message before trying to unpack


TimoGlastra:
Some agents return errors as the HTTP response when something went wrong. We try to unpack all received messages which can result in weird errors:

```
 ERROR  ERROR: error while unpacking message {
  "error": {},
  "packedMessage": {
    "error": "Unhandled exception: Couldn't locate a message handler for type https://didcomm.org/issue-credential/1.0/ack",
    "errorType": "AriesFrameworkException"
  },
  "errorMessage": "Error unpacking message"
}
```

We should try to see if the received message has the structure of a packed message (correct JWE keys) before trying to unpack. We could throw a dedicated error if the structure is not a JWE and notfiy the framework user we received an error message from the other agent

Relevant HTTP code: https://github.com/hyperledger/aries-framework-javascript/blob/main/packages/core/src/transport/HttpOutboundTransport.ts#L78-L81

Encrypted message structure: https://github.com/hyperledger/aries-framework-javascript/blob/main/packages/core/src/types.ts#L15

```
function A(message: any): message is EncryptedMessage {
  return message.protected
}
```

Example encrypted message:

```json
    {
[...]

PR #639, opened 2022-02-14
fix: error if unpacked message does not match JWE structure

Annelein:
Solves #550 


packages/core/src/transport/HttpOutboundTransport.ts
```diff
@@ -76,6 +77,11 @@ export class HttpOutboundTransport implements OutboundTransport {
         this.logger.debug(`Response received`, { responseMessage, status: response.status })
 
         try {
+          if (validJWEStucture(responseMessage) === false) {
```
blu3beri, line 80:
```suggestion
          if (!validJWEStucture(responseMessage)) {
```

packages/core/src/utils/ValidJWEStucture.ts
```diff
@@ -0,0 +1,5 @@
+import type { EncryptedMessage } from '../types'
+
+export function validJWEStucture(message: any): message is EncryptedMessage {
```
blu3beri, line 3: This would allow it to accept a string, but I think it might be better to send the parsed object to this function.

```suggestion
export function validJWEStucture(message: string): boolean {
  const parsedMessage = JSON.parse(message)
  return parsedMessage.protected && parsedMessage.iv && parsedMessage.ciphertext && parsedMessage.tag
}
```

packages/core/src/utils/ValidJWEStucture.ts
```diff
@@ -0,0 +1,5 @@
...
+  return message.protected && message.iv && message.ciphertext && message.tag
```
TimoGlastra, line 4: As the type of message is any, we can't be certain it is an object.

```suggestion
  return message && typeof message === 'object' && message.protected && message.iv && message.ciphertext && message.tag
```

packages/core/src/transport/HttpOutboundTransport.ts
TimoGlastra, line 80:
```suggestion
          if (!validJWEStucture(responseMessage)) {
```

packages/core/src/utils/__tests__/ValidJWEStructure.test.ts
```diff
@@ -0,0 +1,19 @@
...
+describe('ValidJWEStructure', () => {
+  test('throws error when the response message has an invalid JWE structure', async () => {
+    const responseMessage = 'invalid JWE structure'
+    await expect(validJWEStucture(responseMessage)).toBeFalsy()
+  })
+
+  test('valid JWE structure', async () => {
+    const responseMessage = {
+      protected:
+        'eyJlbmMiOiJ4Y2hhY2hhMjBwb2x5MTMwNV9pZXRmIiwidHlwIjoiSldNLzEuMCIsImFsZyI6IkF1dGhjcnlwdCIsInJlY2lwaWVudHMiOlt7ImVuY3J5cHRlZF9rZXkiOiJNYUNKa3B1YzltZWxnblEtUk8teWtsQWRBWWxzY21GdFEzd1hjZ3R0R0dlSmVsZDBEc2pmTUpSWUtYUDA0cTQ2IiwiaGVhZGVyIjp7ImtpZCI6IkJid2ZCaDZ3bWdZUnJ1TlozZXhFelk2RXBLS2g4cGNob211eDJQUjg5bURlIiwiaXYiOiJOWVJGb0xoUG1EZlFhQ3czUzQ2RmM5M1lucWhDUnhKbiIsInNlbmRlciI6IkRIQ0lsdE5tcEgwRlRrd3NuVGNSWXgwZmYzTHBQTlF6VG1jbUdhRW83aGU5d19ERkFmemNTWFdhOEFnNzRHVEpfdnBpNWtzQkQ3MWYwYjI2VF9mVHBfV2FscTBlWUhmeTE4ZEszejhUTkJFQURpZ1VPWi1wR21pV3FrUT0ifX1dfQ==',
```
blu3beri, line 12: Maybe GitHub diff viewer does it wrong but is there an empty line here?

packages/core/src/transport/HttpOutboundTransport.ts
```diff
@@ -76,6 +77,11 @@ export class HttpOutboundTransport implements OutboundTransport {
...
+            throw new Error(
+              `Received an error from the other agent but the structure of the incoming packed message is not a JWE`
```
TimoGlastra, line 82:
```suggestion
              `Received a response from the other agent but the structure of the incoming message is not a DIDComm message: ${responseMessage}`
```

packages/core/src/utils/ValidJWEStucture.ts
TimoGlastra, line 3: For methods that return booleans we usually start with isXXX. Also for abbreviations (JWE) we only capitalize the first letter (so Jwe)

```suggestion
export function isValidJweStucture(message: any): message is EncryptedMessage {
```

blu3beri (changes requested):

TimoGlastra (changes requested):
Great work!

Could you also add this check to the WS outbound transporter?

packages/core/src/utils/ValidJWEStucture.ts
Annelein: @TimoGlastra

packages/core/src/transport/WsOutboundTransport.ts
```diff
@@ -101,7 +102,13 @@ export class WsOutboundTransport implements OutboundTransport {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   private handleMessageEvent = (event: any) => {
     this.logger.trace('WebSocket message event received.', { url: event.target.url, data: event.data })
-    const payload = JSON.parse(Buffer.from(event.data).toString('utf-8'))
+    const responseMessage = Buffer.from(event.data).toString('utf-8')
```
blu3beri, line 105: Not your pr, but `toString()` on a buffer is `UTF-8` by default.

packages/core/src/transport/HttpOutboundTransport.ts
```diff
@@ -76,6 +77,11 @@ export class HttpOutboundTransport implements OutboundTransport {
         this.logger.debug(`Response received`, { responseMessage, status: response.status })
 
         try {
+          if (!isValidJweStucture(responseMessage)) {
+            throw new Error(
+              `Received a response from the other agent but the structure of the incoming message is not a DIDComm message: ${responseMessage}`
+            )
+          }
```
TimoGlastra, line 84: After thinking about it for a bit, maybe we should log n error here and return? It's not fatal and we can just move on.

TimoGlastra: Also, `responseMessage` is of type string here, not json.

packages/core/src/transport/WsOutboundTransport.ts
TimoGlastra: same here, you pass a string but don't parse the json in the isValidJweStructure body.

packages/core/src/transport/HttpOutboundTransport.ts
```diff
@@ -77,6 +78,12 @@ export class HttpOutboundTransport implements OutboundTransport {
 
         try {
           const encryptedMessage = JSON.parse(responseMessage)
+          if (!isValidJweStucture(responseMessage)) {
```
TimoGlastra, line 81:
```suggestion
          if (!isValidJweStucture(encryptedMessage)) {
```

packages/core/src/utils/ValidJWEStucture.ts
blu3beri: I also think that we should use `JsonEncoder.fromString()` and not `JSON.parse()`. Right now, `JsonEncoder.fromString()` just wraps `JSON.parse()`, but it might be nice to expand this function in the future to take a generic and do a type comparison (if the conversion went correctly).

Annelein: So is the conclusion that we leave it like it is right now?

TimoGlastra: Yes, but I agree with @blu3beri's suggestion to use JsonEncoder.fromString() instead of JSON.parse

packages/core/src/transport/WsOutboundTransport.ts
```diff
@@ -101,7 +102,13 @@ export class WsOutboundTransport implements OutboundTransport {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   private handleMessageEvent = (event: any) => {
     this.logger.trace('WebSocket message event received.', { url: event.target.url, data: event.data })
-    const payload = JSON.parse(Buffer.from(event.data).toString('utf-8'))
+    const responseMessage = Buffer.from(event.data).toString()
+    const payload = JSON.parse(responseMessage)
```
TimoGlastra, line 106: For this we can also use the JsonEncoder:

```suggestion
    const payload = JsonEncoder.fromBuffer(event.data)
```

packages/core/src/utils/ValidJWEStucture.ts
Annelein: Implemented!

TimoGlastra (commented):
LGTM

TimoGlastra (approved):

Issue #582, opened 2021-12-15
Present proof same attribute as predicate


berendsliedrecht:
When responding to a proof request with an attribute and predicate of the same value, e.g. requesting attribute `age` and predicate `age` `>` `5` the request fails. This makes sense (ish) as it should not happen that we request a value as well as a predicate variant of this. Indy-sdk throws an error here with `CommonInvalidStructure`. Since this is not logged at the agent it is very difficult to debug. (Even with the log of `CommonInvalidStructure` it is difficult to debug).

I propose to check in the framework if the requested attributes and requested predicates do not match in some way.  After the check throw an error, as we cannot warn this, that the requested predicate cannot be in the attributes as well.

However... all of this should not even happen at the prover. The verifier should never be able to send this request. My verifier is ACA-Py so it might also be an issue there. I also don't think we check this when creating a proof in AFJ.

I am not a 100% that this is the exact reason why IndySdk errors. It might be something else that also happens when we have this kind of proof request structure.

- Write some tests cases that test the following:
	- Can we parse a proof request that having matching attribute names with the same cred def id filter
	- Can we parse a proof request that have a attribute name and predicate name matching with the same cred def id filter
	- Can we parse a proof request that having matching attribute names with different cred def id filter
	- Can we parse a proof request that have a attribute name and predicate name matching with the different cred def id filter

PR #640, opened 2022-02-15
fix: issue where attributes and predicates match

Annelein:
Solves #582


packages/core/src/modules/proofs/services/ProofService.ts
```diff
@@ -1137,6 +1143,27 @@ export class ProofService {
       )
     }
   }
+
+  public assertAttributeNamesDoNotMatch(proofRequest: ProofRequest) {
+    const attributes = Array.from(proofRequest.requestedAttributes.values()).map((a) => a.name)
+    const duplicates = attributes.filter((item, index) => attributes.indexOf(item) != index)
```
blu3beri, line 1149: Maybe I don't see it but it looks like you are comparing each items index with its own index.

Couldn't you create a `Set` of the array and compare those lengths?

e.g.

```js
const attributes = Array.from(proofRequest.requestedAttributes.values()).map((a) => a.name)
const uniqueAttributes = new Set(attributes)
// Ignore the bad naming
const areAttributesTrulyUnique = attributes.length === uniqueAttributes.size
```

blu3beri: Ah I see now how it is supposed to be. `indexOf` returns the first index and if it occured a second time those indices will not be the same. 

Ignore my comment :).

packages/core/src/modules/proofs/services/ProofService.ts
```diff
@@ -1137,6 +1143,27 @@ export class ProofService {
...
+
+    if (duplicates.length > 0) {
+      throw new AriesFrameworkError(`The proof request contains a duplicate attribute name: ${duplicates}`)
```
blu3beri, line 1152: Hopefully soon we will add some stricter typing. But arrays will not be allowed anymore in template literals. Could you replace `${duplicates}` to `${duplicates.toString()}` at all occurrences?

packages/core/tests/helpers.ts
```diff
@@ -664,3 +664,26 @@ export async function setupProofsTest(faberName: string, aliceName: string, auto
...
+    issuerConnectionId: faberConnection.id,
+    holderAgent: aliceAgent,
+    credentialTemplate: {
+      credentialDefinitionId: secCredDefId,
+      comment: 'some comment about credential',
+      preview: credentialPreview,
+    },
+  })
+
+  return { secCredDefId: definition.id }
```
blu3beri, line 688: We generally do not abbriviate variables

packages/core/tests/proofs.test.ts
```diff
@@ -369,4 +372,288 @@ describe('Present Proof', () => {
...
+        name: 'age',
+        restrictions: [
+          new AttributeFilter({
+            credentialDefinitionId: credDefId,
+          }),
+        ],
+      }),
+    }
+
+    await expect(
```
blu3beri, line 396: I dont think `expect` has to be async. I think the function called inside the expect has to be marked `void` though.

blu3beri (approved):

packages/core/tests/proofs.test.ts
Annelein: @TimoGlastra you told me some times tests fail because the expect is not awaited, right?

blu3beri: Ah I just looked at the jest docs and this is definitely how we are supposed to do it. We did not mark this like this in some other tests. Will fix that in a different PR.

packages/core/tests/helpers.ts
Annelein: I actually took this from the function above
https://github.com/hyperledger/aries-framework-javascript/pull/640/files/da58c95310b499ea2ae3a4b84c6d728fb916e487#diff-ea192c0081e4b56d18cea33ca16a6b4d7cb2ccc5b4b23c183ba511843e30b60bL659

should I change it to return just the `definition.id` and take a string as a return type?

packages/core/src/modules/proofs/services/ProofService.ts
Annelein: Yes!

packages/core/tests/helpers.ts
blu3beri: Yes, I think that is even better.

TimoGlastra (changes requested):
Nice, great addition @Annelein. some suggestions on tests and the checks

packages/core/src/modules/proofs/services/ProofService.ts
```diff
@@ -369,8 +373,10 @@ export class ProofService {
     }
     await validateOrReject(proofRequest)
 
-    // Assert attribute and predicate group names do not match
+    // Assert attribute and predicate (group) names do not match
     this.assertAttributePredicateGroupNamesDoNotMatch(proofRequest)
+    this.assertAttributeNamesDoNotMatch(proofRequest)
+    this.assertAttributePredicateNamesDoNotMatch(proofRequest)
```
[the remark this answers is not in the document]
blu3beri: I think one method would suffice for all three checks (although it might feel bit hacky)

```ts
const assertDuplicatesInArray(arr: string[]): void {
  const arrayLength = arr.length
  const uniqueArrayLength = new Set(arr).size

  // Minor optimalization
  if(arrayLength === uniquearrayLength) return;

  const duplicates = arr.filter((item, index) => arr.indexOf(item) != index)
  throw new AriesFrameworkError(`The proof request contains a duplicate items: ${duplicates.toString()}`)
}

// And call like so

// replicate assertAttributeNamesDoNotMatch()
 const attributes = Array.from(proofRequest.requestedAttributes.values()).map((a) => a.name)
assertDuplicatesInArray(attributes)

// replicate assertPredicateNamesDoNotMatch()
const predicates = Array.from(proofRequest.requestedPredicates.values()).map((a) => a.name)
assertDuplicatesInArray(predicates)

// replicate assertAttributePredicateNamesDoNotMatch()
assertDuplicatesInArray(attributes.concat(predicates))
```

The function could also take a `proofRequest` and get the arrays from there, I have no preference really.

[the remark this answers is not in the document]
Annelein: I like it! Will implement it. 
One question:  `if(arrayLength === uniquearrayLength) return;` on this line you say that if the arrays are evenly long, it is fine, but there could still be a duplicate right?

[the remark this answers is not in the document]
blu3beri: When you create a Set in Javascript from an array it will remove all the duplicates instances of items. like so

```ts
const arr = [1,2,3,4,4,4,4];
const setOfArray = new Set(arr);

console.log(setOfArray);
// output: Set(1,2,3,4);
```

So if the array has a length of lets say 5 and the Set aswell it is safe to assume that there are no duplicates removed when creating the set.

packages/core/src/utils/assertDuplicates.ts
```diff
@@ -0,0 +1,23 @@
+import type { ProofRequest } from '../modules/proofs/models/ProofRequest'
+
+import { AriesFrameworkError } from '../error/AriesFrameworkError'
+
+import { attributesToArray, predicatesToArray } from './toArray'
+
+export function checkProofRequestForDuplicates(proofRequest: ProofRequest) {
```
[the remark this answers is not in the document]
TimoGlastra: I'm fine with proof specific utils in the proofs module 👍 

However this is going to be moved to the indy proof format service soon, so should it then be in the proofs utils, or rather in the indy proof format service?

packages/core/src/utils/toArray.ts
```diff
@@ -0,0 +1,12 @@
...
+export function attributesToArray(proofRequest: ProofRequest) {
+  return Array.from(proofRequest.requestedAttributes.values()).reduce<string[]>(
+    (names, a) => [...names, ...(a.name ? [a.name] : a.names ? a.names : [])],
+    []
+  )
+}
+
+export function predicatesToArray(proofRequest: ProofRequest) {
+  return Array.from(proofRequest.requestedPredicates.values()).map((a) => a.name)
+}
```
TimoGlastra, line 12: This is also very proof request specific, so maybe we should host all the utils (assertDuplicates,  toArray) in a util file called `indyProofRequest.ts` (either in the root util directory, or the proofs specific util directory)

packages/core/src/utils/assertDuplicates.ts
```diff
@@ -0,0 +1,23 @@
...
+  const attributes = attributesToArray(proofRequest)
+  const predicates = predicatesToArray(proofRequest)
+  assertDuplicatesInArray(attributes)
+  assertDuplicatesInArray(predicates)
+  assertDuplicatesInArray(attributes.concat(predicates))
+}
+
+export function assertDuplicatesInArray(arr: string[]) {
```
TimoGlastra, line 15: The name of this method makes me assume we want to assert there are duplicates in the array. While we want to make sure there are not duplicates in the array

```suggestion
export function assertNoDuplicatesInArray(array: string[]) {
```

packages/core/src/utils/assertDuplicates.ts
```diff
@@ -0,0 +1,23 @@
...
+import { AriesFrameworkError } from '../error/AriesFrameworkError'
+
+import { attributesToArray, predicatesToArray } from './toArray'
+
+export function checkProofRequestForDuplicates(proofRequest: ProofRequest) {
+  const attributes = attributesToArray(proofRequest)
+  const predicates = predicatesToArray(proofRequest)
+  assertDuplicatesInArray(attributes)
+  assertDuplicatesInArray(predicates)
+  assertDuplicatesInArray(attributes.concat(predicates))
```
TimoGlastra, line 12: I think we can get away with only the last check. We check for any duplicates across the two arrays, making the first two checks redundant

packages/core/src/utils/toArray.ts
```diff
@@ -0,0 +1,12 @@
+import type { ProofRequest } from '@aries-framework/core'
+
+export function attributesToArray(proofRequest: ProofRequest) {
+  return Array.from(proofRequest.requestedAttributes.values()).reduce<string[]>(
```
TimoGlastra, line 4:
```suggestion
  // Attributes can contain either a `name` string value or an `names` string array. We reduce it to a single array
  // containing all attribute names from the requested attributes.
  return Array.from(proofRequest.requestedAttributes.values()).reduce<string[]>(
```

packages/core/src/utils/assertDuplicates.ts
blu3beri: Very good point.

packages/core/tests/proofs.test.ts
```diff
@@ -365,8 +371,162 @@ describe('Present Proof', () => {
         requestedAttributes: attributes,
         requestedPredicates: predicates,
       })
-    ).rejects.toThrowError(
-      `The proof request contains an attribute group name that matches a predicate group name: age`
-    )
+    ).rejects.toThrowError(`The proof request contains duplicate items: age`)
+  })
+
+  test('attribute names match, same cred def filter', async () => {
```
TimoGlastra, line 377: These tests are great! I think it would be good to move them to a util test in the utils __tests__ directory. Then add tests to this file that makes sure the method is called where it should be (createRequest, etc...)

TimoGlastra (commented):
One comment on the tests, but implementation looks good!

packages/core/src/utils/__tests__/indyProofRequest.test.ts
```diff
@@ -0,0 +1,240 @@
...
+  return credentialRecord
+}
+
+describe('Present Proof', () => {
+  let secCredDef: CredentialRecord
+  let credDef: CredentialRecord
+
+  beforeAll(async () => {
+    credDef = mockCredentialRecord({ state: CredentialState.OfferSent })
+    secCredDef = mockCredentialRecord({ state: CredentialState.OfferSent })
```
TimoGlastra, line 98: Because we're doing unit tests we can just create a random credential definition id (took this one from https://indyscan.io).

Also because the `checkProofRequestForDuplicates` method doesn't use the credential definition at all, we don't have to provide a second credential definition id. This should suffice (we also don't need the `mockCredentialRecord` etc... as the credential record is not used, we only need an id):

```suggestion
    const credentialDefinitionId = '9vPXgSpQJPkJEALbLXueBp:3:CL:57753:tag1'
```

Annelein: Do we still want to test if it passes with an attribute and a predicate with different credential id's? or should I remove those tests then as well?

TimoGlastra: I think that could be useful, but this method doesn't test that. It never touches the credentialDefinitionId (note credentialDefinitionId is different from the credentialId. )


We should do an actual proof exchange to be able to test that (and should be done in proofs.test.ts)

packages/core/src/utils/__tests__/indyProofRequest.test.ts
```diff
@@ -0,0 +1,90 @@
...
+  ProofPredicateInfo,
+  ProofRequest,
+} from '@aries-framework/core'
+
+export const INDY_CREDENTIAL_OFFER_ATTACHMENT_ID = 'libindy-cred-offer-0'
+
+describe('Present Proof', () => {
+  let credDefId: string
+
+  beforeAll(async () => {
```
blu3beri, line 17: Does not have to be async

packages/core/src/utils/__tests__/indyProofRequest.test.ts
```diff
@@ -0,0 +1,90 @@
...
+    credDefId = '9vPXgSpQJPkJEALbLXueBp:3:CL:57753:tag1'
+  })
+
+  test('attribute names match, same cred def filter', async () => {
```
blu3beri, line 21: Does not have to be async

packages/core/src/utils/__tests__/indyProofRequest.test.ts
```diff
@@ -0,0 +1,90 @@
...
+      name: 'proof-request',
+      version: '1.0',
+      nonce,
+      requestedAttributes: attributes,
+    })
+
+    expect(() => checkProofRequestForDuplicates(proofRequest)).toThrowError(AriesFrameworkError)
+  })
+
+  test('attribute names match with predicates name, same cred def filter', async () => {
```
blu3beri, line 53: Does not have to be async

packages/core/src/utils/__tests__/indyProofRequest.test.ts
```diff
@@ -0,0 +1,90 @@
...
+        name: 'age',
+        restrictions: [
+          new AttributeFilter({
+            credentialDefinitionId: credDefId,
+          }),
+        ],
+      }),
+    }
+
+    const nonce = 'testtesttest12345'
```
blu3beri, line 41: @TimoGlastra I created an issue for this, #646. Is this solution fine or do we need to expose the generateNonce function in a more accessible way?

packages/core/src/utils/indyProofRequest.ts
```diff
@@ -0,0 +1,22 @@
...
+export function attributesToArray(proofRequest: ProofRequest) {
+  // Attributes can contain either a `name` string value or an `names` string array. We reduce it to a single array
+  // containing all attribute names from the requested attributes.
+  return Array.from(proofRequest.requestedAttributes.values()).reduce<string[]>(
+    (names, a) => [...names, ...(a.name ? [a.name] : a.names ? a.names : [])],
+    []
+  )
+}
+
+export function predicatesToArray(proofRequest: ProofRequest) {
```
blu3beri, line 14: I think `predicateNamesToArray`, and same for attributes, Is a better name.

packages/core/tests/helpers.ts
```diff
@@ -664,3 +664,29 @@ export async function setupProofsTest(faberName: string, aliceName: string, auto
     aliceReplay,
   }
 }
+
+export async function setupSecondCredential(
```
blu3beri, line 668: Is this still being used somewhere?

blu3beri (approved):

TimoGlastra (approved):
LGTM

PR #641, opened 2022-02-15
feat: filter retrieved credential by revocation state

TimoGlastra:
Filter retrieved credentials by revoked status. If revoked = true it means non-revocation was requested and the credentials can't be used to to satisfy the proof request. This is configurable in the `getRequestedCredentialsForProofRequest` method using the `filterByNonRevocationRequirements` property. 

Made it filter by non revocation requirements by default, but figured it may be dersirable to retrieve credentials that match except for the non-revocation requirement (same as the `filterByPresentationPreview` config option). The question is whether we want `filterByNonRevocationRequirements` to be `true` or `false` by default. Currently have it to false, but maybe true makes more sense?

@JamesKEbert you input is highly appreciated here. Also if there's anything incorrect here please let me know. I'm not that familiar with revocation.


packages/core/src/modules/proofs/services/ProofService.ts
```diff
@@ -1067,6 +1042,43 @@ export class ProofService {
...
+        `Presentation is requesting proof of non revocation, getting revocation status for credential`,
+        {
+          requestNonRevoked,
+          credentialRevocationId,
+          revocationRegistryId,
+        }
+      )
+
+      // Note presentation from-to's vs ledger from-to's: https://github.com/hyperledger/indy-hipe/blob/master/text/0011-cred-revocation/README.md#indy-node-revocation-registry-intervals
+      const status = await this.indyRevocationService.getRevocationStatus(
```
blu3beri, line 1070: minor: status can be deconstructed here to `revoked` and `deltaTimestamp`

blu3beri (approved):

JamesKEbert (approved):
This looks great to me! 
As to the question of whether the config should be by default or not, I tend to think it should be, as using a revoked credential for a proof request will fail on the verifier side. The primary/standard use case for having the revoked credentials present is to add contextual information to a user why they can't use a credential in a presentation (which was part of our use case), which is a bit more complex and requires the end-developer using it correctly.

JamesKEbert:
> I also created a PR in AATH that fixes the broken revocation tests: https://github.com/hyperledger/aries-agent-test-harness/pull/436

> It is not dependant on this PR and already works (all tests that align with RFC0441 now pass 🎉 )

That's awesome to hear! Thanks for getting those fixed/sorted out! :)

PR #649, opened 2022-02-24
feat: issue credential v2

NB-MikeRichardson:


packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts
```diff
@@ -0,0 +1,750 @@
...
+  }
+  /**
+   * Get attributes for indy format from a proposal message. This allows attributes
+   * to be copied across to old style credential records
+   *
+   * @param proposal ProposeCredentialOptions object containing (optionally) the attributes
+   * @return array of attributes or undefined if none present
+   */
+  public getCredentialAttributes(proposal: ProposeCredentialOptions): CredentialPreviewAttribute[] | undefined {
+    const credPropose: CredPropose = proposal.credentialFormats.indy?.payload as CredPropose
```
TimoGlastra, line 255: With the more strict types we wouldn't have to cast everytime. Looser typing can easily lead to bugs! (Typescript won't complain when the type changes)

packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts
```diff
@@ -0,0 +1,750 @@
...
+      return {
+        indy: {
+          payload: {
+            credentialPayload: credOffer, // old v1 object from Indy SDK
+          },
+        },
+      }
+    }
+    if (!this.indyIssuerService) {
+      throw new Error('Missing Indy Issuer Service')
```
TimoGlastra, line 301: Use AFJ error

packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts
```diff
@@ -0,0 +1,750 @@
...
+
+  public async processProposal(
+    options: AcceptProposalOptions,
+    credentialRecord: CredentialExchangeRecord
+  ): Promise<AcceptProposalOptions> {
+    let proposalMessage
+    try {
+      proposalMessage = await this.didCommMessageRepository.getAgentMessage({
+        associatedRecordId: credentialRecord.id,
+        messageClass: V2ProposeCredentialMessage,
```
TimoGlastra, line 365: V2 specific code should be in v2 credential service

packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts
```diff
@@ -0,0 +1,750 @@
...
+      })
+    } catch (RecordNotFoundError) {
+      // can happen in normal processing
```
TimoGlastra, line 368: To prevent the try/catch you can use findAgentMessage

packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts
```diff
@@ -0,0 +1,750 @@
...
+    if (proposal.credentialFormats.indy) {
+      proposal.credentialFormats.indy.attributes = preview.attributes
+    }
+    return proposal
+  }
+
+  public async processProposal(
+    options: AcceptProposalOptions,
+    credentialRecord: CredentialExchangeRecord
+  ): Promise<AcceptProposalOptions> {
```
TimoGlastra, line 360: Why are we already returning the accept options here? I think we can maybe leave that to the create request method

packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts
```diff
@@ -0,0 +1,750 @@
...
+   * @returns The Attachment if found or undefined
+   */
+  public getAttachment(
+    message:
+      | V2RequestCredentialMessage
+      | V2ProposeCredentialMessage
+      | V2OfferCredentialMessage
+      | V2IssueCredentialMessage
+  ): Attachment | undefined {
+    const indyFormat = message.formats.find((f) => f.format.includes('indy'))
```
TimoGlastra, line 423: Would be nice to be a bit more strict here and match in the exact format names

packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts
```diff
@@ -0,0 +1,750 @@
...
+    }
+  }
+  /**
+   * Gets the attachment object for a given attachId. We need to get out the correct attachId for
+   * indy and then find the corresponding attachment (if there is one)
+   * @param message Gets the
+   * @returns The Attachment if found or undefined
+   */
+  public getAttachment(
+    message:
```
TimoGlastra, line 417: This seems v2 specific. Maybe this can be moved to the v2 service and the format service will only include a method supportsFormatId. For all indy ones it will return true, but the v2credentialservice is still responsible for the extraction of the attachment (as it can change between versions, in v3 attachments will be placed in the appended/generic attachments array)

packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts
```diff
@@ -0,0 +1,750 @@
...
+  /**
+   * Create a {@link AttachmentFormats} object dependent on the message type.
+   *
+   * @param requestOptions The object containing all the options for the credential request
+   * @param credentialRecord the credential record containing the offer from which this request
+   * is derived
+   * @returns object containing associated attachment, formats and requestAttach elements
+   *
+   */
+  // public async createIssueCredentialAttachFormats(credentialRecord: CredentialRecord, options: AcceptRequestOptions) {
```
TimoGlastra, line 437:
```suggestion
```

packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts
```diff
@@ -0,0 +1,750 @@
...
+  ): boolean {
+    const autoAccept = CredentialResponseCoordinator.composeAutoAccept(
+      credentialRecord.autoAcceptCredential,
+      autoAcceptType
+    )
+
+    if (autoAccept === AutoAcceptCredential.Always) {
+      return true
+    } else if (autoAccept === AutoAcceptCredential.ContentApproved) {
+      return this.isRequestDefinitionIdValidNEW(requestPayload, offerPayload, proposePayload)
```
TimoGlastra, line 632:
```suggestion
      return this.isRequestDefinitionIdValid(requestPayload, offerPayload, proposePayload)
```

packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts
```diff
@@ -0,0 +1,750 @@
...
+
+    if (autoAccept === AutoAcceptCredential.Always) {
+      return true
+    } else if (autoAccept === AutoAcceptCredential.ContentApproved) {
+      return this.areCredentialValuesValidNEW(credentialRecord, credentialPayload)
+    }
+    return false
+  }
+
+  private areProposalValuesValidNEW(
```
TimoGlastra, line 662: Why the NEW?

TimoGlastra (changes requested):
I didn't get to review all files (there were A LOT of changes), but going to put this in as a first review.

Overall the structure looks really good, but in some places I think we can approach things a bit different. Interested to hear what you have to say!

TimoGlastra:
Ok -- I gave the PR another look @NB-MikeRichardson. Much better, but still a lot of minor things to be improved. Quite some comments from my first review haven't been addressed. I've marked all that are addressed as resolved. Maybe you can start giving thumbs up or something to keep track of what you have resolved.

For some of the smaller items I encountered I thought it would be quicker to fix and so made a PR with changes. Hope it is helpful: https://github.com/NB-MikeRichardson/aries-framework-javascript/pull/13. I tried to make a lot of small and scoped commits so it's easier to see what kind of changes I've made.

 I've also left some new comments and have made some notes while going through the code in my local editor:


- The v1 service still calls the indy issuer / holder / verifier services directly. I think with the indy format service this won't be needed anymore? What does the indy format service not do that is needed in v1, but not in v2?
- I've found quite a lot of commented code in tests. Either uncomment and make it work or remove the comments
- I see the v2 service imports v1 messages? V1 and v2 should be completely separated
- There is no v2 credential problem report as I can see
- Log messages are not specific enough. Either provide messages that show the context or remove then (`version = {}` is too vague)
- You often end with throwing errors and do all logic in if statements. To avoid nesting of if statements, it can sometimes help to first make guard statements and keep the happy flow top level
- I've seen multiple places with empty string values. A value is either required or optional but we should never initialize it with an empty string value (see some examples below, might be more):
	 - `const comment: string = proposal.comment ? proposal.comment : ''`
	 - `replacementId: ''`
	 - ```     
	   const options: AcceptProposalOptions = {       
          connectionId: '',
          credentialRecordId: '',
           credentialFormats: {},
         }
- I've found quite some unused interfaces (mainly in `interfaces.ts` but also other places). This adds complexity to understanding the code
- I'd remove the default protocol version. In the future we may detect this using the discover features protocol in which case we can pick a sensible default. But there's not an obvious reason to pick either v1 or v2 so we should let the user decide
- I've found multiple `MJR TODO`. Please fix, or make it a more general todo
- We should probably check the protocolVersion after retrieving a record from storage with the protocol version from the received message (so in v2 service check that the record protocol version is also v2). This can prevent errors where another agent is sending messages for both protocols.
- I've found multiple places where ProtocolVersion needs to be passed even though we can extract it from the record. This can lead to errors if the user provides a different version than is stored on the record!
- I'm not so sure how the extension to the attachment decorator works with format, but those are specific to this protocol and should not be added to the base decorator. Please remove. If you need to reuse logic across all credential messages you can create a base message that is used for all issue credential messages
- Handlers should not call format services directly. This should always go through the credential service.
- I found some checks in the wrong places:
	- v2 credential sevice has this check, which is indy specific and must live in the indy format service: `credentialOptions.credentialFormats.indy?.credentialDefinitionId`
- `RequestCredentialOptions` needs a `credentialDefinition` key which is indy specific? Shouldn't this live under the format specific key?
- Found quite some typos in comments and variable names. If you're using VSCode I can highly recommend this plugin: https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker
- The interfaces are a mess. Maybe we can look at creating a separate file for all the interfaces per level. So a file for module interfaces. File for service interfaces and a file for format service interfaces.
- The interfaces should be better separated per level. The module exposes too much properties that are often only needed on the service/format service layer. This confuses the user of the framework and we should make sure public API only exposes the absolutely needed properties (e.g. `acceptCredentialProposal` takes an `offerAttachment` which shouldn't be exposed. This also leads to lot of optional parameters which makes it harder to catch bugs. We don't get type errors anymore about missing parameters.
- There is some uncessary nesting present in some interfaces. 
E.g. the following interface:
```ts
        credentialFormats: {
          indy: {
            payload: {
              credentialPayload: {
                attributes: credentialPreview.attributes,
                schemaIssuerDid: faberAgent.publicDid?.did,
                schemaName: schema.name,
                schemaVersion: schema.version,
                schemaId: schema.id,
                issuerDid: faberAgent.publicDid?.did,
              },
            },
          },
        },
``` 

Could be reduced to:

```
        credentialFormats: {
          indy: {
            attributes: credentialPreview.attributes,
            schemaIssuerDid: faberAgent.publicDid?.did,
            schemaName: schema.name,
            schemaVersion: schema.version,
            schemaId: schema.id,
            issuerDid: faberAgent.publicDid?.did,
          },
        },

**Comment by TimoGlastra** [comment]
Doing some local testing with the ICv2 apis. Looks really great but still have some notes left on how the current API works. It's mainly about interfaces requiring properties they don't need, or making properties that are actually required optional.

- The `agent.credentials.offerCredential` takes an optional `payload` in for the indy format. I think we can leave this out for the module?
- Protocol version is marked as optional in the offer credential method, but an error is thrown if the version is not provided. Can we make it required in the interface?
- From the offer credential method in the module all the way down to the `this.indyIssuerService.createCredentialOffer` call in the indy credential format service there's not a single try catch. This leads to the following error being thrown if something goes wrong. This is probably better addressed in a separate PR as this is also present in the current implementation (just want to highlight it so we don't forget).
```
/Users/timo/Developer/Work/Projects/AFJ/aries-framework-javascript/packages/core/src/modules/indy/services/IndyIssuerService.ts:93
      throw isIndyError(error) ? new IndySdkError(error) : error
                                 ^
IndySdkError: IndyError(CommonInvalidStructure): CommonInvalidStructure
    at IndyIssuerService.createCredentialOffer (/Users/timo/Developer/Work/Projects/AFJ/aries-framework-javascript/packages/core/src/modules/indy/services/IndyIssuerService.ts:93:34)
    at async IndyCredentialFormatService.createCredentialOffer (/Users/timo/Developer/Work/Projects/AFJ/aries-framework-javascript/packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts:259:34)
    at async IndyCredentialFormatService.createOffer (/Users/timo/Developer/Work/Projects/AFJ/aries-framework-javascript/packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts:138:19)
    at async V1CredentialService.createOfferProcessing (/Users/timo/Developer/Work/Projects/AFJ/aries-framework-javascript/packages/core/src/modules/credentials/protocol/v1/V1CredentialService.ts:485:42)
    at async V1CredentialService.createOffer (/Users/timo/Developer/Work/Projects/AFJ/aries-framework-javascript/packages/core/src/modules/credentials/protocol/v1/V1CredentialService.ts:1121:43)
    at async CredentialsModule.offerCredential (/Users/timo/Developer/Work/Projects/AFJ/aries-framework-javascript/packages/core/src/modules/credentials/CredentialsModule.ts:378:43)
    at async run (/Users/timo/Developer/Work/Projects/AFJ/aries-framework-javascript/samples/extension-module/requester.ts:44:30)
```
- I had to import `CredentialProtocolVersion` and `CredentialPreviewAttribute` from a nested path. Could you export types that are needed to interact with the credentials module from the root package?
- This PR lost the additions from this PR https://github.com/hyperledger/aries-framework-javascript/pull/625 that adds the `checkAttributesMatch` check on a credential offer
- The following interface is defined, but I'm not sure why we need to define the cred def here. It seems to be set inside the indy credential format service and therefore doesn't have to be exposed in the credential format service public api?
```ts
export interface FormatServiceRequestCredentialOptions extends ServiceRequestCredentialOptions {
  credentialDefinition?: {
    credDef: CredDef
  }
}
```
- There is both a `acceptCredentialOffer` and a `acceptOffer` method. I think we can just do with only `acceptOffer` (having two methods that do the same may confuse people)
- module method input are sometimes called `options`, sometimes method speicfic without options (e.g. `offer`) and sometimes method specific with options (e.g. `credentialOptions`). Would be good to make this consistent
- Maybe we can leave out the credential part from the method names in the module as we're already in the credential module. Some methods are without `Credential` (e.g. `acceptRequest`) while other are with (e.g. `acceptCredentialProposal`). I think for all `acceptXXX` and `negotiateXXX` methods we can just use `acceptOffer`, etc... instead of `acceptCredentialOffer`
- the interface of `negotiateCredentialOffer` extends the propose interface which takes a connection id, but for negotiation this is not required as we already have the connection id stored in the credential exchange record. We should probably not extend the interface of the propose method in this case. Same is the case with `NegotiateProposalOptions`, this shouldn't take a connectionId
- The same point is valid for the `protocolVersion`, we can take this from the existing credential exchange record when negotiating and don't need to make this available through the public api
- The `CredPropose` takes a `comment` property. There is a comment property on higher levels and it doesn't seem to be used. I think we should remove the comment property here.
- `acceptCredentialProposal` takes an optional `connectionId` (not needed). It also requires an input on the format payload (same interface as `createOffer`). If you look at the `acceptProposal` method in the current main branch in takes a lot less parameters as we can construct the offer based on the proposal. If we don't do that the `negotiateProposal` and `acceptProposal` are basically the same. Can you look at making this reflect the behaviour of the current implementation?
- When using negotiation the framework will error out with the following erorr. Instead of using `saveAgentMessage` you should rather use `saveOrUpdateAgentMessage` to avoid two messages of the same type being saved. Would be good to add some tests for this as the negotiation feature is not working at the moment:
```
RecordDuplicateError: DidCommMessageRecord: Multiple records found for given query '{"associatedRecordId":"7c88742a-0ad0-47ca-beb2-490768679829","messageType":"https://didcomm.org/issue-credential/2.0/offer-credential"}
```
- The `createOutOfBandOffer` method requires a connection id to be passed. This method is intended to be used in connectionless exchange where no connection id is available

TimoGlastra:
It also seems to undo the additions in this PR: https://github.com/hyperledger/aries-framework-javascript/pull/691

TimoGlastra:
Did another walk through the code, this time focussed on the service and credential format service

- CredentialRole is unused
- The credential format service describes the `attachment` property in the return type of the `createXXX` method as optional. This means the service has to do a check if the attachment is returned and otherwise throw an error. Would it be possible to directly throw an error instead in the credential format service? I don't see the case where an attachment shouldn't be returned.
- The interface of the `createOffer` method in the format service is not making sense to me. The method suggests we're creating an offer, the property suggests we're passing a proposal, and the type of the property suggest we're accepting an offer.

```
abstract createOffer(proposal: ServiceAcceptOfferOptions): Promise<FormatServiceOfferAttachmentFormats>
```

- I found the following statement in the `CredentialMessageBuilder.createOffer`. Using `as unknown as XXX` should only be used as a last resort. If we change anything about the input type it will be ignored by TypeScript, while the addition of the strict type checking is to help prevents such changes from causing issues. Looking at the input of the `CredentialMessageBuilder.createOffer` options it's defined as `ServiceOfferCredentialOptions`. While the `CredentialFormatService.createOffer` takes an `AcceptProposalOptions`. Shouldn't this be something like `OfferCredentialOptions` instead?

```
const offerOptions = options as unknown as AcceptProposalOptions
```

- The `ServiceOfferCredentialOptions` defines an offer attachment. I'm not 100% following why the offer credential interface needs the offer attachment? Looking at the code this is set in the `CredentialMessageBuilder.createOffer` method. This is where I'm more lost as it's set for each call to the format service (so only the last one actually stored). Then it calls `CredentialFormatService.processOffer` only _if_ `options.offerAttachment` is defined. I think we can remove the `options.offerOfferAttachment`, and also the call to `CredentialFormatService.processOffer` from here. `CredentialFormatService.processOffer` should be called when we _receive_ a credential offer message, not when we create it.
- You're passing an empty string value (`''`) for replacementId. This should be undefined if we're not using it.
- The `id` property is marked as required in the `V2OfferCredentialMessageOptions` interface. Normally we make the `id` optional and generate it in the constructor (which it is already doing, should just make it optional) -> Please check this for all the message classes, I've seen other methods also have this.
- credential preview has been marked as optional in the v2 offer credential message but looking at the RFC it seems required. Otherwise you won't know what variables the credential will contain when using the indy format. This is different for JSON-LD credentials, but we should still probably make it required? What do you think?
- Maybe we can handle the `CredentialFormatService.checkPreviewAttributesMatchSchemaAttributes` method inside the indy credential format service and not be a public method. The indy format service is the only service that needs that check
- `CredentialMessageBuilder.createCredential` declares `offerMessage` as optional but it then throws an error if this is not defined. Could you please do a check on all types and make sure to not declare fields as optional that are actually required. This makes the code harder to understand.
- You quite often modify the input options objects and attach properties to it in other methods. The disadvantage of this approach is that you never know when fields are defined or undefined even though they are required in the end. E.g. the `CredentialMessageBuilder.createCredential` takes an `options` object that declares the `offerAttachment` and `requestAttachment` as optional. This is misleading as the value will always be added in the `CredentialMessageBuilder.createCredential`, so we could also take them out and reduce complexity of the `CredentialMessageBuilder.createCredential`. Then the fields are also declared as optional for the `CredentialFormatService.createCredential` but they're actually required. I propose the following:
  1. Remove the `offerAttachment` and `requestAttachment` from the `ServiceAcceptRequestOptions`.
  2. Make the `requestAttachment` required and the `offerAttachment` optional in the `CredentialFormatService.createCredential` method
  3. Instead of doing `options.offerAttachments = ` we construct a new options object that maps from the service to the credential format service types:
     1. `const formatServiceOptions: TheType = { ...serviceOptions, offerAttachment: theOfferAttachment, requestAttachment: theRequestAttachment }`
- There's a lot of checks for values that we know are defined because they've been made required in the types (e.g. `offerMessage` in `CredentialMessageBuilder.createRequest` is required so we don't have to do extra checks)
- I found the following comment. What is the error being thrown?

```
// @IsInstance(CredentialFormatSpec, { each: true }) -> this causes message validation to fail
```

- I see you used the `assert` method from the `console` library. Is this intended? I think we should just throw an error here instead.
- The `V2CredentialService.processProposal` calls `CredentialMessageBuilder.acceptProposal`. This feels quite misleading as it isn't actually accepting the proposal, it's just creating an exchange record after receiving a proposal.
- I found the following code. By using `findAgentMessage` we can remove the `RecordNotFoundError` handling, that is only needed for get. (find will return null if not found)
  Before

```ts
// this may not be the first proposal message...
let proposalCredentialMessage, offerCredentialMessage;
try {
  proposalCredentialMessage =
    await this.didCommMessageRepository.findAgentMessage({
      associatedRecordId: credentialRecord.id,
      messageClass: V2ProposeCredentialMessage,
    });
  offerCredentialMessage = await this.didCommMessageRepository.findAgentMessage(
    {
      associatedRecordId: credentialRecord.id,
      messageClass: V2OfferCredentialMessage,
    }
  );
} catch (RecordNotFoundError) {
  // record not found - expected (sometimes)
}
```

After

```ts
const proposalCredentialMessage =
  await this.didCommMessageRepository.findAgentMessage({
    associatedRecordId: credentialRecord.id,
    messageClass: V2ProposeCredentialMessage,
  });
const offerCredentialMessage =
  await this.didCommMessageRepository.findAgentMessage({
    associatedRecordId: credentialRecord.id,
    messageClass: V2OfferCredentialMessage,
  });
```

- In the `V2CredentialFormatService` I see you're calling `setMetaDataFor` in both the `createProposal` and `processProposal` methods. The `setMetaDataFor` calls `CredentialFormatService.processProposal` under the hood. It doesn't make sense to me to call `CredentialFormatService.processProposal` from both the `create` and `process` proposal methods. Shouldn't we _only_ call `processXXX` methods in the credential format service when we're processing a _received_ message (please check other methods for this, I've seen it being called in create method multiple times)? Should we rename `processXXX` to something different to make this more clear? In addition `setMetadataFor` is maybe not the best name. It's correct that for indy we now may be setting some metadata, but we don't know what the format service does with the format data. I'd say call `CredentialFormatService.processProposal` directly from the `V2CredentialService.processProposal` and remove the `setMetaDataFor` method
- I found the following code. If the connection.id is defined the search would never return a record where the connection id is then undefined (because it searches for a record where the connection id is the same). So the assignment of the connection id in this case does nothing and can be removed. (This happens in multiple places I believe, please check)

```ts
// Credential record already exists
credentialRecord = await this.getByThreadAndConnectionId(
  proposalMessage.threadId,
  connection?.id
);
credentialRecord.connectionId = connection?.id;
```

- I see there's two methods in the `V2CredentialService` that take exactly the same parameters but have them swapped (`acceptProposal` and `createOfferAsResponse`). We should try to be as consistent as possible. I'd say record takes precedence over the options.
- Nit: I see a lot of uses of `offerCredentialMessage ? offerCredentialMessage : undefined`. This can be shortened to `offerCredentialMessage ?? undefined` I think
- I've seen the following code. Again, I think mentioning metadata here is an assumption. we just know that we're calling processOffer. If the `processOffer` actually updates the metadata it could log so inside the method itself.

```ts
this.logger.debug("Save metadata for offer");
await format.processOffer(attachment, credentialRecord);
```

- Nit: The `credentialMessageBuilder.createRequest` method takes 5 parameters. Usually when something needs that much input we make it an object so you can explicitly declare each parameter with a key
- Instead of calling `createAcceptProposalOptions` and `createOfferAsResponse` from the `V2ProposeCredentialHandler` couldn't we call `acceptProposal` and remove the public method `createAcceptProposalOptions`? `acceptProposal` could do the same the `createAcceptProposalOptions` method does but without needing to expose another method.
- There's still some places that call `getAgentMessage` with a catch that does nothing. Could you update these to `findAgentMessage` instead?
- Found the following code. Instead of adding a method `getRevocationRegistry` on the formatService and then passing it directly to the credential format service again, why not remove the public method `getRevocationRegistry` and just calling it internally in the `processCredential` method?

```ts
const revocationRegistry = await formatService.getRevocationRegistry(
  issueAttachment
);

const options: ServiceAcceptCredentialOptions = {
  credentialAttachment: issueAttachment,
  revocationRegistry,
};
await formatService.processCredential(options, credentialRecord);
```

- `deleteAssociatedCredential` -> `deleteAssociatedCredentials` now that it can be multiple
- `options?.deleteAssociatedCredential && credentialRecord` -> `&& credentialRecord` is redundant
- `shouldAutoRespondToProposal` is not implemented for v1. Why is that?
- should auto respond seems to only take the agent config into account and not the record auto accept config? The record config should take precedence over the agent config (would be good to have a test case that covers this). Or am I missing something?
- Why is the `shouldAutoRespondToProposal` method checking for `appendedAttachments` (`proposeMessage && proposeMessage.appendedAttachments`). I think this is unrelated?
- Most methods in the credential response coordinator are unused. I've added a note below that we should use the response coordinator more, but for now if the method isn't used, please remove it
- Maybe we can make the `getFormatService` method non-public in the `V2CredentialService`? I don't think there's a need for it to be public. There's one usage of it but that's from the `V1RequestCredentialHandler` that should probably call `credentialService.shouldAutoRespondToRequest()` right?
- We're almost nowhere doing checks that we're working with at least one format. If you look at ACA-Py this check is done regularly: https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/protocols/issue_credential/v2_0/manager.py#L234-L237
- `getFormats` method can be made private for `V2CredentialService` and removed from the `V1CredentialService` I think.
- The methods in the services are all over the place. We have a `shouldAutoRespond` method followed by a `deleteById` method followed by another `shouldAutoRespond` method. This is all before some private properties and the public constructor. It would be great if we could group methods. Previously we had it grouped from top to bottom with first the private / public properties, then the public contructor, then the `createXXX` and `processXXX` (+ other message specific methods) at the top, starting with the first message of the protocol (`createProposal`, `processProposal`, `createRequest`, `processRequest` ,etc...). Then the repository methods, then private internal methods. Maybe we can add the should auto respond between the message methods and the repository methods?
- `V1CredentialService.shouldAutoRespondToRequest` is using appended attachment as the proposal attachment, but v1 credential proposal doesn't have any attachments?
- `deleteById` is implemented in both `V1CredentialService` and `V2CredentialService. I think this can be handled like the rest of the repository methods? (add it to the base class)
- The following code in the V1CredentialService feels very inconsistent. All try to access the attachment format. All do it differently

```
    if (proposeMessage && proposeMessage.appendedAttachments) {
      proposalAttachment = proposeMessage.appendedAttachments[0]
    }
    if (offerMessage) {
      offerAttachment = offerMessage.getAttachmentById(INDY_CREDENTIAL_OFFER_ATTACHMENT_ID)
    }
    if (requestMessage) {
      requestAttachment = requestMessage.getAttachmentIncludingFormatId(INDY_CREDENTIAL_REQUEST_ATTACHMENT_ID)
    }
```

- Maybe we can make `holderDid` part of the options for the `CredentialFormatService`? It feels like an option to me
- `V1CredentialService.processRequest` doesn't call `IndyCredentialFormatService.processRequest`
- `V1CredentialService.processOffer` doesn't call `IndyCredentialFormatService.processOffer`. We should call _ALL_ format service methods for a few reasons. This is happening for quite a lot of methods, please reuse the credential format service for this:
  1. If we make update to the indy format service processing it will be effective for both v1 and v2
  2. We don't have to repeat the logic (which we're currently doing with the setting of metadata in the v1 process offer method)
- `credentialRecordId: ''` -> Don't use empty strings as values

--- 

Also noted down some things we can probably leave for future PRs:

- Credential record is retrieved multiple times (in module and in service)
   - Maybe we can look at either passing the connection to the service or call `connection.assertReady` on the module level (module level removes connections a bit more from the service which is not bad)
- We could optimize the auto accept flow a bit by returning early if one of the format services returns false. In addition. You do a check if autoAcceptCredential is never to return false directly. We could do the same for always and return true directly. Only `ContentApproved` should be handled by the format service. Also the `CredentialResponseCoordinator` seems mostly unused now? In the present proof branch the handler calls the response coordinator which only calls the service if the type is content approved. I think we could also use that approach here.
- `getFormatsFromMessage` currently throws an error if it detects an unrecongized format. We should just ignore this, as that's the nice thing of sending more formats. We can pick the ones we support
- credential format identifiers are used multiple times, would be good to declare these as constant if they're used multiple times (e.g. `hlindy/proof-req@v2.0`)

jakubkoci:
I see that we're using a naming convention `V1XxxMessage`, `V2XxxMessage`. Are there any problems removing this prefix and relying only on a place in a folder structure when importing?

packages/core/src/modules/credentials/protocol/v2/messages/V2OfferCredentialMessage.ts
```diff
@@ -0,0 +1,63 @@
...
+      this.credentialPreview = options.credentialPreview
+      this.messageAttachment = options.offerAttachments
+    }
+  }
+
+  @Type(() => CredentialFormatSpec)
+  @ValidateNested()
+  @IsArray()
+  // @IsInstance(CredentialFormatSpec, { each: true }) -> this causes message validation to fail
+  public formats!: CredentialFormatSpec[]
```
TimoGlastra, line 34: I think this issue is happening because the value is not always a class. I've seen multiple occurences of the following:

```
    const formats: CredentialFormatSpec = {
      attachId: this.generateId(),
      format: 'hlindy/cred-filter@v2.0',
    }
```

which are not a class instance but rather a plain object. To make it an actual class instance you'd have to do this:

```ts
    const formats = new CredentialFormatSpec({
      attachId: this.generateId(),
      format: 'hlindy/cred-filter@v2.0',
    })
```

And add the following constructor to the `CredentialFormatSpec` class:

```
  public constructor(options: { attachId: string; format: string }) {
    if (options) {
      this.attachId = options.attachId
      this.format = options.format
    }
  }
```

Typescript is fine with only an object that matches the interface, while we lose the decorators in that case... It's an annoying issue

packages/core/src/modules/credentials/protocol/v2/handlers/V2OfferCredentialHandler.ts
```diff
@@ -0,0 +1,118 @@
...
+    this.agentConfig.logger.info(
+      `Automatically sending request with autoAccept on ${this.agentConfig.autoAcceptCredentials}`
+    )
+
+    if (messageContext.connection) {
+      const { message, credentialRecord } = await this.credentialService.createRequest(
+        record,
+        {},
+        messageContext.connection.did
+      )
```
TimoGlastra, line 78: As mentioned before, I think ti would look cleaner to add the did to the options object:

```suggestion
      const { message, credentialRecord } = await this.credentialService.createRequest(
        record,
        {
          holderDid: messageContext.connection.did
        } 
      )
```

packages/core/src/modules/credentials/protocol/v2/__tests__/v2credentials-architecture.test.ts
```diff
@@ -0,0 +1,121 @@
...
+        multiFormatProposal.credentialFormats as FormatServiceProposeCredentialFormats
+      const formats: CredentialFormatService[] = service.getFormats(credFormats)
+      expect(formats.length).toBe(1) // for now will be added to with jsonld
+      const messageBuilder: CredentialMessageBuilder = new CredentialMessageBuilder()
+
+      const v2Proposal = messageBuilder.createProposal(formats, multiFormatProposal)
+
+      expect(v2Proposal.message.formats.length).toBe(1)
+      expect(v2Proposal.message.formats[0].format).toEqual('hlindy/cred-filter@v2.0')
+      // expect(v2Proposal.message.formats[1].format).toEqual('aries/ld-proof-vc-detail@v1.0')
```
TimoGlastra, line 118:
```suggestion
```

packages/core/src/modules/credentials/protocol/v1/V1CredentialPreview.ts
```diff
@@ -46,20 +14,20 @@ export interface CredentialPreviewOptions {
...
   @Expose({ name: '@type' })
-  @Equals(CredentialPreview.type)
+  @Equals(V1CredentialPreview.type)
   @Transform(({ value }) => replaceLegacyDidSovPrefix(value), {
     toClassOnly: true,
   })
-  public readonly type = CredentialPreview.type
-  public static readonly type = 'https://didcomm.org/issue-credential/1.0/credential-preview'
+  public type = V1CredentialPreview.type
+  public static type = `https://didcomm.org/issue-credential/1.0/credential-preview`
```
TimoGlastra, line 30:
```suggestion
  public readonly type = V1CredentialPreview.type
  public static readonly type = 'https://didcomm.org/issue-credential/1.0/credential-preview'
```

TimoGlastra (approved):
I've left a few more comments. I'm going to accept this PR and think it may be good to merge on the account that the outstanding comments just raised will be resolved as soon as possible.

I've opened this pr to address some small issues, also with the migration script: https://github.com/NB-MikeRichardson/aries-framework-javascript/pull/27

In addition I have one more general comment:
- Offer credential interface is shared between offer credential and out of band offer. They should have separate interfaces as one requires a connectionId and the other explicitly doesn't need one.

packages/core/src/modules/credentials/protocol/v2/messages/V2RequestCredentialMessage.ts
```diff
@@ -0,0 +1,52 @@
+import { Expose, Type } from 'class-transformer'
```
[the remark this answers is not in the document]
NB-MikeRichardson: I assume you mean RevocationNoficationMessage, rather than RequestCredentialMessage.

Moved the two messages and two handlers into folders unter protocol/revocation-notification

packages/core/src/modules/credentials/protocol/v2/handlers/V2OfferCredentialHandler.ts
NB-MikeRichardson: I could have sworn we decided not to put it in the options object as it is Indy specific, so from a previous review the decision was made to make it (temporarily) a parameter while we decided how to address this issue.

That said, I have made the change anyway.

PR #652, opened 2022-02-28
feat: add wallet module with import export

TimoGlastra:
Add a wallet module that exposes the public api of the wallet. This is in preparation of the wallet upgrade work, but better addressed in a separate PR. 

After creating a wallet we should store the framework storage version in the wallet. If we do this directly in the wallet we depend on a repository from the wallet, which in turn depends on the wallet, creating a dependency cycle. I first created a hacky solution with tsyringe.

However it probably makes more sense to add another layer (the WalletModule) that exposes the public api of the wallet and allows to add extra logic to wallet methods without adding it to the wallet methods itself. Advantage is that wallet class keeps dependencies lighter, and we also can perform tasks that are generic to all wallet types (such as storing the framework storage version after creating a wallet). I will add this in a separate PR.

In addition this PR also adds support for import / export of the wallet which we'll also be using for the upgrade work. Now for NodeJS and will soon update rn-indy-sdk to support import / export


packages/core/tests/wallet.test.ts
```diff
@@ -101,4 +102,39 @@ describe('=== wallet', () => {
...
+    }
+
+    const backupKey = 'someBackupKey'
+    const backupWalletName = `backup-${uuid()}`
+    const backupPath = path.join(tmpdir(), backupWalletName)
+
+    // Create backup and delete wallet
+    await bobAgent.wallet.export({ path: backupPath, key: backupKey })
+    await bobAgent.wallet.delete()
+
```
blu3beri, line 132: Might be quite redundant but could we add an `expect` to see that the messageRecord does not exist after deletion?

MosCD3:
@TimoGlastra  will that allow for changing the wallet key? like export the wallet then import it with different key?

TimoGlastra:
> @TimoGlastra will that allow for changing the wallet key? like export the wallet then import it with different key?

Yes it should allow for changing the wallet key, but indy provides a native way to change the wallet key. We can look at making that API available

TimoGlastra:
Updated and ready for review again

packages/core/src/crypto/__tests__/JwsService.test.ts
```diff
@@ -18,7 +18,7 @@ describe('JwsService', () => {
     const config = getAgentConfig('JwsService')
     wallet = new IndyWallet(config)
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-    await wallet.initialize(config.walletConfig!)
+    await wallet.create({ ...config.walletConfig!, keepOpenAfterCreate: true })
```
jakubkoci, line 21: Isn't it a parameter of the `create` method rather than `walletConfig`? Those flag params are harder to understand and use, we're already using them quite a lot. I'm thinking if `createAndOpen` wouldn't be better. No strong opinion just small feedback.

packages/core/src/wallet/IndyWallet.ts
```diff
@@ -127,6 +107,8 @@ export class IndyWallet implements Wallet {
         throw new WalletError(errorMessage, { cause: error })
```
jakubkoci, line 107: I know it's not part of your changes but I think we should close the wallet when `createMasterSecret` throws an error, otherwise, we can get into an inconsistent state.

jakubkoci (approved):
Just added a few notes but it seems good overall 👍

packages/core/src/agent/Agent.ts
```diff
@@ -93,7 +95,7 @@ export class Agent {
     this.messageSender = this.container.resolve(MessageSender)
     this.messageReceiver = this.container.resolve(MessageReceiver)
     this.transportService = this.container.resolve(TransportService)
-    this.wallet = this.container.resolve(InjectionSymbols.Wallet)
+    this.walletService = this.container.resolve(InjectionSymbols.Wallet)
```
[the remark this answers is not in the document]
TimoGlastra: > It seems a bit weird it's named `walletService` although it's not a service.

Agreed. Do you have any suggestions of the naming?

> What if we expose initPublicDid and publicDid via the wallet "module"?

I would rather not do that at the moment. With the work happening on other did methods and the did module, I'd like to move away from the `initPublicDid` and `publicDid`. With supporting multiple types of dids, it'll be more common to have different kind of public dids, so `publicDid` and `initPublicDid` won't make a lot of sense anymore. So I'd rather not expose that API right now, but rather keep it as is currently (wallet api has only been exposed after 0.1.0 release, so not a breaking change to not make them public).

We can then start looking at how to handle public dids in general through the dids module. The wallet can then focus more on providing the actual crypto instead of dealing with DIDs.

TimoGlastra:
Thanks for the suggestions @jakubkoci. The wallet is now closed if an error occurs while creating the wallet, and a new `createAndOpen` method is added. Could you take another look?

MosCD3:
> > @TimoGlastra will that allow for changing the wallet key? like export the wallet then import it with different key?
> 
> Yes it should allow for changing the wallet key, but indy provides a native way to change the wallet key. We can look at making that API available

Thanks @TimoGlastra if you give me a lead about that "native indy" I can make our team work on it.

TimoGlastra:
@MosCD3 If you look at the `openWallet` api in the NodeJS wrapper it supports a `rekey` property: https://github.com/hyperledger/indy-sdk/tree/master/wrappers/nodejs#openwallet--config-credentials----handle

I assume this is also supported by both RN/NodeJS wrappers (as it just passes the config to the native libindy library). You would need to make this property available in the AFJ config, after which it should work

packages/core/src/agent/Agent.ts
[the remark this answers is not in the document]
jakubkoci: I see. Yes, dids module is a better place for this functionality. That can help with `walletService` naming issue. We wouldn't need to have a class member referring to`Wallet` instance.

jakubkoci:
@TimoGlastra Good to go. I'm sure we will solve the naming issue somehow.

jakubkoci:
> @MosCD3 If you look at the openWallet api in the NodeJS wrapper it supports a rekey property:

What if we add a new method `rotateKey(walletConfig)`. It would close the wallet and open it with the new key.

tranducduong1994:
Hello @TimoGlastra @jakubkoci when can we release this feature? I would like to use import/export method. But the latest release is v0.1.0

TimoGlastra:
Soon!

PR #655, opened 2022-03-03
feat(core): add OOB record

jakubkoci:
There are still some places where it's not so nice because of preserving compatibility with old connection protocol initialization.

I added `protocolCreateRequest` and `protocolProcessRequest` to `ConnectionService` but I want to separate them to`ConnectionProtocol` class later together with other "protocol" methods as `DidExchangeProtocol` class is organized.

We can also add some events around the OOB record, but I also leave it out, for now, to keep it simple.


packages/core/src/modules/oob/OutOfBandService.ts
```diff
@@ -0,0 +1,36 @@
...
+  public constructor(outOfBandRepository: OutOfBandRepository) {
+    this.outOfBandRepository = outOfBandRepository
+  }
+
+  public async save(outOfBandRecord: OutOfBandRecord) {
+    return this.outOfBandRepository.save(outOfBandRecord)
+  }
+
+  public async updateState(outOfBandRecord: OutOfBandRecord, newState: OutOfBandState) {
+    outOfBandRecord.state = newState
```
TimoGlastra, line 21: I see this doesn't include a state yet. Is this something you intend to add in an upcoming PR?

packages/core/src/modules/oob/OutOfBandModule.ts
```diff
@@ -218,36 +218,63 @@ export class OutOfBandModule {
...
+          autoAcceptConnection,
+        })
       }
 
       if (messages) {
         this.logger.debug('Out of band message contains request messages.')
-        if (!connectionRecord.isReady) {
+        if (connectionRecord.isReady) {
+          await this.emitWithConnection(connectionRecord, messages)
+        } else {
```
TimoGlastra, line 277: Not for PR, but now that we have an out of band record, we could update this to an event based flow. So instead of waiting for the connection to be finished, we can add an internal event listener that whenever a connection is completed checks whether there is any oob flow to be finished.

packages/core/src/modules/oob/OutOfBandModule.ts
```diff
@@ -218,36 +218,63 @@ export class OutOfBandModule {
       )
     }
 
+    const outOfBandRecord = new OutOfBandRecord({
+      role: OutOfBandRole.Receiver,
+      state: OutOfBandState.PrepareResponse,
+      outOfBandMessage: outOfBandMessage,
+      autoAcceptConnection: autoAcceptConnection ?? this.agentConfig.autoAcceptConnections,
```
TimoGlastra, line 225: For the connections record we don't store the agentConfig in the connection record. If we change the auto accept config parameter it should be reflected on all flows that don't have it expliclitly set. So I think to be consistent we should just use the autoAcceptConnection value (which may be undefined).

packages/core/src/modules/oob/OutOfBandModule.ts
```diff
@@ -167,7 +157,16 @@ export class OutOfBandModule {
       })
     }
 
-    return { outOfBandMessage, connectionRecord }
+    const outOfBandRecord = new OutOfBandRecord({
+      role: OutOfBandRole.Sender,
+      state: OutOfBandState.Initial,
+      outOfBandMessage: outOfBandMessage,
+      reusable: multiUseInvitation,
+      autoAcceptConnection: autoAcceptConnection ?? this.agentConfig.autoAcceptConnections,
```
TimoGlastra, line 165: Same here

packages/core/src/modules/oob/OutOfBandModule.ts
```diff
@@ -89,73 +101,51 @@ export class OutOfBandModule {
...
-      }
-      outOfBandMessage = new OutOfBandMessage(options)
-    } else {
-      const routing = await this.mediationRecipientService.getRouting()
-      const service = new DidCommService({
-        id: '#inline',
+    const services = routing.endpoints.map((endpoint, index) => {
+      return new DidCommService({
+        id: `#inline-${index}`,
         priority: 0,
```
TimoGlastra, line 135:
```suggestion
        priority: index,
```

packages/core/src/modules/connections/repository/ConnectionRecord.ts
```diff
@@ -26,10 +26,11 @@ export interface ConnectionRecordProps {
   threadId?: string
   tags?: CustomConnectionTags
   imageUrl?: string
-  multiUseInvitation: boolean
+  multiUseInvitation?: boolean
   mediatorId?: string
   errorMessage?: string
   protocol?: HandshakeProtocol
+  outOfBandId?: string
```
TimoGlastra, line 33: Thinking whether it may be better to give the oob record an optional connectionId? A connection can have multiple oob records associated with it (due to connection reuse), while each oob record can have at most 1 connection.

packages/core/src/modules/connections/handlers/ConnectionResponseHandler.ts
```diff
@@ -1,22 +1,44 @@
...
+    if (!messageContext.recipientVerkey || !messageContext.senderVerkey) {
+      throw new AriesFrameworkError('Unable to process connection request without senderVerkey or recipientVerkey')
+    }
+
+    const connectionRecord = await this.connectionService.findByVerkey(messageContext.recipientVerkey)
+    if (!connectionRecord) {
+      throw new AriesFrameworkError(`Connection for verkey ${messageContext.recipientVerkey} not found!`)
+    }
+
+    const outOfBandRecord =
```
TimoGlastra, line 37:
```suggestion
    const outOfBandRecord = connectionRecord.outOfBandId ? (await this.outOfBandService.findById(connectionRecord.outOfBandId)) : undefined
```

packages/core/src/modules/connections/handlers/ConnectionResponseHandler.ts
```diff
@@ -1,22 +1,44 @@
...
+      connectionRecord.outOfBandId && (await this.outOfBandService.findById(connectionRecord.outOfBandId))
+
+    // The presence of outOfBandRecord is not mandatory when the old connection invitation is used
+    const connection = await this.connectionService.processResponse(messageContext, outOfBandRecord || undefined)
```
TimoGlastra, line 41:
```suggestion
    const connection = await this.connectionService.processResponse(messageContext, outOfBandRecord)
```

packages/core/src/modules/connections/ConnectionsModule.ts
```diff
@@ -38,19 +43,59 @@ export class ConnectionsModule {
...
+
+    let result
+    if (protocol === HandshakeProtocol.DidExchange) {
+      result = await this.didExchangeProtocol.createRequest(outOfBandRecord, {
+        label: label || this.agentConfig.label,
+        routing,
+        autoAcceptConnection,
+      })
+    } else if (protocol === HandshakeProtocol.Connections) {
+      result = await this.connectionService.protocolCreateRequest(outOfBandRecord, {
```
TimoGlastra, line 81: What's the difference in naming of `createRequest` and `protocolCreateRequest`?

packages/core/src/modules/oob/repository/OutOfBandRecord.ts
```diff
@@ -0,0 +1,81 @@
...
+      recipientKey,
+    }
+  }
+
+  public getRecipientKeys() {
+    return this.outOfBandMessage.services
+      .filter((s): s is DidCommService => typeof s !== 'string')
+      .map((s) => s.recipientKeys)
+      .reduce((acc, curr) => [...acc, ...curr], [])
+  }
```
TimoGlastra, line 68: This won't work well with public dids. But as we're not supporting that right now I guess it's fine

packages/core/src/modules/oob/repository/OutOfBandRecord.ts
```diff
@@ -0,0 +1,81 @@
...
+      this.role = props.role
+      this.state = props.state
+      this.autoAcceptConnection = props.autoAcceptConnection
+      this.reusable = props.reusable ?? false
+      this._tags = props.tags ?? {}
+    }
+  }
+
+  public getTags() {
+    const [recipientKey] = this.getRecipientKeys()
```
TimoGlastra, line 52: We're only storing the first key here. Is this the key always used to send a message to?

Also, `recipientKey` feels a bit ambiguous to me. With the connection record we used `invitationKey` which was indepent of the role (just the key used for the invitation). But with `recipientKey` I'm wondering who's the recipient. As both sides are a recipient?

packages/core/src/modules/oob/repository/OutOfBandRecord.ts
```diff
@@ -0,0 +1,81 @@
...
+  tags?: TagsBase
+  outOfBandMessage: OutOfBandMessage
+  role: OutOfBandRole
+  state: OutOfBandState
+  autoAcceptConnection: boolean
+  reusable?: boolean
+}
+
+export class OutOfBandRecord extends BaseRecord<TagsBase> {
+  @Type(() => OutOfBandMessage)
```
TimoGlastra, line 26: I see you're using `OutOfBandMessage` instead of `OutOfBandInvitationMessage`. Is there a specific reason in deviating from the RFC?

packages/core/src/modules/connections/DidExchangeProtocol.ts
```diff
@@ -61,28 +62,38 @@ export class DidExchangeProtocol {
...
+  ): Promise<{ message: DidExchangeRequestMessage; connectionRecord: ConnectionRecord }> {
+    this.logger.debug(`Create message ${DidExchangeRequestMessage.type} start`, { outOfBandRecord, params })
 
-    if (!connectionRecord.invitation) {
-      throw new AriesFrameworkError('Connection invitation is missing.')
-    }
+    const { outOfBandMessage } = outOfBandRecord
+    const { goal, goalCode, routing, autoAcceptConnection } = params
+
+    const connectionRecord = await this.connectionService.createConnection({
```
TimoGlastra, line 73: Is the idea to extract the connection protocol stuff to a separate `ConnectionProtocol` class?

TimoGlastra (approved):
Some notes, but overall looks good to me. I'm fine with merging this now, and having these points adressed  in separate PRs

packages/core/src/modules/oob/OutOfBandService.ts
jakubkoci: I'm sorry, I don't understand what do you mean by this "doesn't include a state yet."?

TimoGlastra: Whoops, I mean event

packages/core/src/modules/connections/repository/ConnectionRecord.ts
jakubkoci: Hmm. Isn't it actually vice-versa? I can create a multi-use OOB record. Therefore I'll have more connections related to one OOB. If I reuse a connection, it doesn't create a new one.

packages/core/src/modules/connections/ConnectionsModule.ts
jakubkoci: The old one is to preserve the implementation relying on an existing connection record. It'll be removed as we discussed on the last AFJ call.

packages/core/src/modules/connections/DidExchangeProtocol.ts
jakubkoci: Yes.

packages/core/src/modules/connections/repository/ConnectionRecord.ts
TimoGlastra: Hmm, maybe you're right.

But in the following scenario it's the other way around:
1. I create an oob invitation with handshake protocols
2. Other party receices the invitation, a connection is made.
3. I make another oob invitation with handshake.
The other party receives it and sends an connection reuse.
4. I now have 2 oob record that are associated witht the same connection.


With multi use invitations there's indeed the possibility of multiple connections having the same oob record.....

packages/core/src/modules/oob/OutOfBandService.ts
jakubkoci: I thought so :) Yes, I plan to add it eventually.

packages/core/src/modules/oob/repository/OutOfBandRecord.ts
jakubkoci: Fair point. No particular reason for that. Maybe, we can name it just `InvitationMessage` because it's already in oob module and also `HandshakeReuseMessage` is not `OutOfBandHandshakeReuseMessage`.

packages/core/src/modules/connections/handlers/ConnectionResponseHandler.ts
jakubkoci: This won't work because `findById` can return `null`. However, I hope I'll get rid of it when OOB replaces the old invitation.

packages/core/src/modules/oob/OutOfBandModule.ts
jakubkoci: I see. 👍

packages/core/src/modules/oob/OutOfBandModule.ts
jakubkoci: Both updated.

packages/core/src/modules/oob/repository/OutOfBandRecord.ts
jakubkoci: Good point. I just did it in the same way we used `invitationKey` in connection record. We could change it to an array. Is it possible to search for records by tags containing array?

> Also, recipientKey feels a bit ambiguous to me.

I hear you, but I'm not sure if it helps. If it's `invitationKey` I still can wonder who's the invited, or not? It would definitely help to communicate that this key is part of oob invitation. On the other hand, we're introducing new terminology outside of protocol spec.

PR #669, opened 2022-03-11
feat(routing): allow to discover mediator pickup strategy

### LINKED ISSUE #490: Detect implicit or explitict mediation pickup strategy
I have discussed this with @TimoGlastra and we found it to be a nice-to-have feature.

The flow is as follows:

1. Aries Framework JavaScript agent adds a `mediationConnectionInvite` in their `agentConfig` without specifying the `mediationPickupStrategy`.
2. The agent connects with the mediator and queries its features.
3. When the mediator supports `https://didcomm.org/messagepickup/1.0`, we could set `mediationPickupStrategy` to `explicit`. If it does not support `https://didcomm.org/messagepickup/1.0/batch-pickup` we could set it to `implicit`.

I would love to hear some feedback or any other methods that could do the same, but more efficiently.

TimoGlastra:
Add option to detect the pickup strategy that should be used for picking up messages using the discover features protocol. 

This is only used if no pickup strategy is defined in the startup config. If fetched, it will be stored in the mediation record so it can be reused later.

Fixes #490

BREAKING CHANGE: the default `mediatorPickupStrategy` value has been removed. This allows for the value to be determined using the discover features protocol. If you would like the original behaviour, make sure to set the `mediatorPickupStrategy` in the agent config to `MediatorPickupStrategy.Explicit`


TimoGlastra:
@jakubkoci @JamesKEbert @blu3beri can someone please review? :)

packages/core/src/modules/routing/RecipientModule.ts
```diff
@@ -181,6 +192,60 @@ export class RecipientModule {
...
+  }
+
+  private async isBatchPickupSupportedByMediator(mediator: MediationRecord) {
+    const { protocolUri } = parseMessageType(BatchPickupMessage.type)
+
+    // Listen for response to our feature query
+    const replaySubject = new ReplaySubject(1)
+    this.eventEmitter
+      .observable<AgentMessageProcessedEvent>(AgentEventTypes.AgentMessageProcessed)
+      .pipe(
```
blu3beri, line 223: I am not too sure what will be listened to here, but these are a lot of operations. Can't we combine the maps and filter? This might make it a bit less readable but performance should go "up" (probably a very very minor increase though).

blu3beri (approved):

packages/core/src/modules/routing/RecipientModule.ts
TimoGlastra: Yeah sure. This method is really only called once per mediator you connect to as s recipient so was not too worried about adding a few extra method calls. rxjs can be hard to read so slicing it into small steps may help future readers (including me!)

Will update

genaris:
@TimoGlastra this seems interesting. So basically it will use explicit strategy only if mediator supports Batch Pickup message, and implicit otherwise?

I guess if mediator is AFJ-based, it will take explicit strategy, isn't it?

TimoGlastra:
> So basically it will use explicit strategy only if mediator supports Batch Pickup message, and implicit otherwise?

Correct!

> I guess if mediator is AFJ-based, it will take explicit strategy, isn't it?

Yes indeed, AFJ only supports the explicit method. In the future we can add support for pickup v2 which we can also detect if the mediator supports it

Issue #671, opened 2022-03-13
Adding the ability to rotate wallet key


MosCD3:
Reference to the conversation here #652 
I am working on adding a function to AFJ that will allow rotating key as part of security module work on Bifold or any other implementation wallet

MosCD3:
Created a PR to reflect new attribute in type definition 
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/59256

TimoGlastra:
Fixed in #672

PR #672, opened 2022-03-14
feat: support wallet key rotation

MosCD3:
Signed-off-by: Mostafa <mostafa.youssef@ontario.ca>

Related issue: #671 

The type also has to be updated in indy-sdk types repo
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/59256

I tested it e2e using [ComX](https://github.com/MosCD3/ComX4) test bed and it worked fine and I was able to rotate the wallet key
Although I faced challenges parsing decoding incoming messages using the latest AFJ but will post that on different ticket


packages/core/src/agent/Agent.ts
```diff
@@ -214,6 +214,28 @@ export class Agent {
     this._isInitialized = false
   }
 
+  public async rotateKey(walletConfig: WalletConfig) {
```
TimoGlastra, line 217: I think this is better located in the `WalletModule` and `Wallet` interface.

Maybe not every wallet handles this the same, so we should probably add a `rotateKey` method to the `Wallet` interface and require to specify the new key and the new key derivation method.

packages/core/src/agent/Agent.ts
```diff
@@ -214,6 +214,28 @@ export class Agent {
...
+    
+    // close wallet if still initialized
```
TimoGlastra, line 219: Could you make sure your code passes validation using `yarn validate`?

packages/core/src/types.ts
```diff
@@ -19,6 +19,7 @@ export const enum KeyDerivationMethod {
 export interface WalletConfig {
   id: string
   key: string
+  rekey?: string
```
TimoGlastra, line 22: I think this shouldn't be part of the open method, but rathe as a separate rotate wallet key config

TimoGlastra (changes requested):
Thanks for this addition @MosCD3. Could you add a test for the changes? Would be good to have a test that creates a wallet, rotates the key and then opens the wallet using the new key.

packages/core/src/agent/Agent.ts
MosCD3: OK I agree, I moved the function to wallet interface then added an implementation in WalletModule and of course IndyWallet, however, we still need to access that from the agent side cuz you still need to shutdown the agent correctly like stop transports and you know all those buzz and whistles going on the "shutdown()" which is accessible on the Agent level. 

So the current implementation would be 

Access rekey from the Agent -> the Agent will make sure it shutdown properly -> execute the "rekey" method of the Wallet module
Is that correct? excuse me here am like a baby learning first steps :)

packages/core/src/types.ts
MosCD3: Agree but we have a problem here, I could create a different interface or type which will move around as long as I am in AFJ but the "this.indy.openWallet( ..." in IndyWallet module take a "WalletConfig" type and I can't go lower than that its part of the indy-sdk types repo

Or do you mean we create the lets say "WalletConfigRekey" in indy-sdk types repo? I don't think this is good idea since we still only have one function to open the wallet "openWallet" in the LibIndy wrappers and it already takes an optional argument "rekey" as you mentioned before. lots of workI guess

MosCD3: Agree but we have a problem here, I could create a different interface or type which will move around as long as I am in AFJ but the "this.indy.openWallet( ..." in IndyWallet module take a "WalletConfig" type and I can't go lower than that its part of the indy-sdk types repo

Or do you mean we create the lets say "WalletConfigRekey" in indy-sdk types repo? I don't think this is good idea since we still only have one function to open the wallet "openWallet" in the LibIndy wrappers and it already takes an optional argument "rekey" as you mentioned before. lots of workI guess

packages/core/src/agent/Agent.ts
TimoGlastra: No I'd say you need to shutdown the agent yourself.

So rather something like:

```ts
await agent.shutdown()
await agent.wallet.rotateKey()
await agent.initialize()
```

packages/core/src/types.ts
TimoGlastra: Yeah agree changing indy-sdk is not a good idea, but I think we can hide it in AFJ.

The implementation of the rotateKey method could like something like this:

```ts
// could optionally call this.close() here, but this makes it more explicit on how to do key rotation
if (this.wallet.isInitialized) {
  throw new AriesFrameworkError("Wallet must be closed to perform key rotation. Call agent.shutdown() before rotating your wallet key")
}

// we create a private method ._open in the indy wallet that is called by .open() and .rotateKey()
// This way .open() in the public interface doesn't have the rekey property
this._open({

})
```

MosCD3: Sure, 
But the 'rekey' has to be in the 'OpenWalletCredentials' interface so if you can review the Pr [there](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/59256)

packages/core/src/agent/Agent.ts
MosCD3: Got it, so the shutting down will occur on the wallet side (controller)

TimoGlastra:
The types PR has been merged @MosCD3

TimoGlastra:
What is the version of `@types/indy-sdk` in the `yarn.lock` file? 

To make sure it works you should updates all `@types/indy-sdk` and `@types/indy-sdk-react-native` version in the package.json to `^1.16.15`

packages/core/src/agent/Agent.ts
```diff
@@ -1,7 +1,7 @@
 import type { Logger } from '../logger'
 import type { InboundTransport } from '../transport/InboundTransport'
 import type { OutboundTransport } from '../transport/OutboundTransport'
-import type { InitConfig } from '../types'
+import type { InitConfig, WalletConfig } from '../types'
```
TimoGlastra, line 4: Is the `WalletConfig` import unused?

packages/core/src/wallet/IndyWallet.ts
```diff
@@ -129,9 +157,13 @@ export class IndyWallet implements Wallet {
...
         { id: walletConfig.id },
-        { key: walletConfig.key, key_derivation_method: walletConfig.keyDerivationMethod }
+        { key: walletConfig.key, rekey: rekey, key_derivation_method: walletConfig.keyDerivationMethod }
       )
-      this.walletConfig = walletConfig
+      if (this.walletConfig) {
+        this.walletConfig = { ...this.walletConfig, ...walletConfig }
+      } else {
+        this.walletConfig = walletConfig
+      }
```
TimoGlastra, line 166: If rekey is set we can maybe set the wallet config with the new key as the key:

```suggestion
      if (walletConfig.rekey) {
        const { rekey, key, keyDerivationMethod, rekeyDerivationMethod, ...rest } = walletConfig
        this.walletConfig = { ...rest, key: rekey, keyDerivationMethod: rekeyDerivationMethod }
      } else {
        this.walletConfig = walletConfig
      }
```

packages/core/src/types.ts
```diff
@@ -22,6 +22,13 @@ export interface WalletConfig {
   keyDerivationMethod?: KeyDerivationMethod
 }
 
+export interface WalletConfigRekey {
+  id: string
+  key: string
+  rekey?: string
```
TimoGlastra, line 28: I think we should make `rekey` required in the rotate key method.

packages/core/src/types.ts
```diff
@@ -22,6 +22,13 @@ export interface WalletConfig {
   keyDerivationMethod?: KeyDerivationMethod
 }
 
+export interface WalletConfigRekey {
```
TimoGlastra, line 25: can you add rekeyDerivationMethod also?

MosCD3:
This will fail 

 if (walletConfig.rekey) {
        const { rekey, key, keyDerivationMethod, rekeyDerivationMethod, ...rest } = walletConfig
        this.walletConfig = { ...rest, key: rekey, keyDerivationMethod: rekeyDerivationMethod }
      } else {
        this.walletConfig = walletConfig
      }
      
   It failed after I merged it 

'walletConfig.rekey' wallet config does not have 'rekey' remember you said to create new type for key rotation
and also there is no 'rekeyDerivationMethod'

I will apply new solution and let me know if it works

TimoGlastra:
Ah my bad. But how you did it now looks good.

We should include the rekeyDerivationMethod: https://github.com/hyperledger/indy-sdk/blob/master/wrappers/nodejs/README.md#openwallet--config-credentials----handle

MosCD3:
@TimoGlastra what happened to this PR, it has mixed up changes from other PRs

MosCD3:
> Ah my bad. But how you did it now looks good.
> 
> We should include the rekeyDerivationMethod: https://github.com/hyperledger/indy-sdk/blob/master/wrappers/nodejs/README.md#openwallet--config-credentials----handle

Done

packages/core/tests/wallet.test.ts
```diff
@@ -157,4 +157,27 @@ describe('wallet', () => {
     // Expect same basic message record to exist in new wallet
     expect(await bobBasicMessageRepository.getById(basicMessageRecord.id)).toMatchObject(basicMessageRecord)
   })
+
+  test('changing wallet key', async () => {
```
TimoGlastra, line 161: This is using the same key to rekey the wallet. Would be good to add a test that does the following:

- create and open the wallet
- close the wallet
- rotate the wallet key to a new key
- open the wallet (using the agent.wallet so you can provide the new key directly -> more explicit)

MosCD3:
> @TimoGlastra The test function is indeed using a new key not the same key, if take a close look you have to pass the old key + the new key for indy to rotate the key and that's what am doing
> 
> So let me revise the test again
> 
> 1- create and open the wallet // already implemented but do I need 'initialize()'?
> 
> 2- close the wallet //Already implimented await aliceAgent.shutdown() //or do you mean Walle.close? but that will leave the agent initialized
> 
> 3- rotate the wallet key to a new key //Already implemented passing old key and another new key in rekey attribute
> 
> 4- open the wallet (using the agent.wallet so you can provide the new key directly -> more explicit) Remember by the time you rotate the key, wallet is already opened so you might add 'agent.initialize()'

@TimoGlastra any thoughts about that?

TimoGlastra (approved):
You're right @MosCD3. LGTM

PR #675, opened 2022-03-15
refactor(core): renamed BufferEncoder to TypedArrayEncoder

blu3beri:
BufferEncoder is incorrectly named as it accepts Uint8Arrays (and buffers). The name TypedArrayEncoder is more fitting as it fits both versions.


TimoGlastra (approved):
I'm not so sure on the name `TypedArray`, but from googling a bit it seems as the most fitting name. The reason why it's called `BufferEncoder` is because the return is always Buffer (if decoding).

But I'm good with this change 👍

blu3beri:
> The reason why it's called `BufferEncoder` is because the return is always Buffer (if decoding).

Is there actually something specific we need from the buffer? Most likely there is, but if not we might just want to work with uint8arrays.

blu3beri:
Yeah okay, that makes sense. Thanks

TimoGlastra:
Can you resolve conflicts @blu3beri?

PR #679, opened 2022-03-23
feat: regex for schemaVersion, issuerDid, credDefId, schemaId, schemaIssuerDid

Annelein:


packages/core/src/modules/credentials/models/IndyCredentialInfo.ts
```diff
@@ -29,10 +29,12 @@ export class IndyCredentialInfo {
 
   @Expose({ name: 'schema_id' })
   @IsString()
+  @Matches(/^[a-zA-Z0-9]{21,22}:2:.+:[0-9.]+$/)
   public schemaId!: string
 
   @Expose({ name: 'cred_def_id' })
   @IsString()
+  @Matches(/^([a-zA-Z0-9]{21,22}):3:CL:(([1-9][0-9]*)|([a-zA-Z0-9]{21,22}:2:.+:[0-9.]+)):(.+)?$/)
```
TimoGlastra, line 37: These are still inline

packages/core/src/utils/__tests__/regex.test.ts
```diff
@@ -0,0 +1,35 @@
+import { credDefIdRegex, issuerDidRegex, schemaIdRegex, schemaIssuerDidRegex, schemaVersionRegex } from '../regex'
+
+describe('Valid Regular Expression', () => {
+  const invalidTest = 'test'
+
+  test('test for credDefIdRegex', async () => {
+    const test = 'q7ATwTYbQDgiigVijUAej:3:CL:160971:1.0.0'
+    expect(test).toMatch(credDefIdRegex)
+    expect(credDefIdRegex.test(invalidTest)).toBeFalsy
```
TimoGlastra, line 9: Will this work without `()`? If you do `toBeTruthy` will that throw an error?

```suggestion
    expect(credDefIdRegex.test(invalidTest)).toBeFalsy()
```

TimoGlastra: Bot are fine! I just didn't know you could call it without the `()` :)

TimoGlastra (approved):

PR #680, opened 2022-03-23
test: reuse connection create and process request tests

jakubkoci:
Just small PR adding ConnectionService tests I removed in the previous PR. I also added some checks for the OOB role and state.

### DIFF
--- a/packages/core/src/modules/connections/services/ConnectionService.ts
+++ b/packages/core/src/modules/connections/services/ConnectionService.ts
@@ -75,21 +75,13 @@ export class ConnectionService {
    * @param config config for creation of connection request
    * @returns outbound message containing connection request
    */
-  public async protocolCreateRequest(
+  public async createRequest(
     outOfBandRecord: OutOfBandRecord,
     config: ConnectionRequestParams
   ): Promise<ConnectionProtocolMsgReturnType<ConnectionRequestMessage>> {
-    if (outOfBandRecord.role !== OutOfBandRole.Receiver) {
-      throw new AriesFrameworkError(
-        `Invalid out-of-band record role ${outOfBandRecord.role}, expected is ${OutOfBandRole.Receiver}.`
-      )
-    }
-
-    const state = outOfBandRecord.state
-    const expectedStates = [OutOfBandState.Initial, OutOfBandState.PrepareResponse]
-    if (state && !expectedStates.includes(state)) {
-      throw new AriesFrameworkError(`Invalid out-of-band record state ${state}. Valid states are: ${[expectedStates]}.`)
-    }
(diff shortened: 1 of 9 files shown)


packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -74,12 +75,14 @@ export class ConnectionService {
...
-  public async protocolCreateRequest(
+  public async createRequest(
     outOfBandRecord: OutOfBandRecord,
     config: ConnectionRequestParams
   ): Promise<ConnectionProtocolMsgReturnType<ConnectionRequestMessage>> {
-    // TODO check oob role is sender
-    // TODO check oob state is await-response
+    this.logger.debug(`Create message ${ConnectionRequestMessage.type} start`, outOfBandRecord)
+    outOfBandRecord.assertRole(OutOfBandRole.Receiver)
+    outOfBandRecord.assertState([OutOfBandState.Initial, OutOfBandState.PrepareResponse])
```
TimoGlastra, line 84: In what case is it initial and what case is it prepareResponse?

packages/core/src/modules/oob/repository/OutOfBandRecord.ts
```diff
@@ -67,14 +67,20 @@ export class OutOfBandRecord extends BaseRecord<TagsBase> {
       .reduce((acc, curr) => [...acc, ...curr], [])
   }
 
+  public assertRole(expectedRole: OutOfBandRole | undefined) {
```
TimoGlastra, line 70: Why can expectedRole be undefined?

TimoGlastra (approved):

packages/core/src/modules/oob/repository/OutOfBandRecord.ts
jakubkoci: I copy-pasted from `connectionRecord` where I put it and left it by accident I guess :) Updated in both places.

packages/core/src/modules/connections/services/ConnectionService.ts
jakubkoci: It didn't make sense. I updated it.

jakubkoci: I wasn't using initial and setting it directly to one of prepare/await response. So the check should be more strict. But I updated oob module to set oob state to `initial` in `createInvitation` and then update it in `acceptInvitation`. The method `recieveInvitation` stays same, setting `await-response` directly to a new oob record.

jakubkoci:
@TimoGlastra I did updates based on your suggestions, can please check if it makes sense?

PR #684, opened 2022-03-28
feat: bbs createKey, sign and verify

blu3beri:
**READY FOR REVIEW**

Almost every change in this PR will be reverted when support for `aries-askar` is added, as we can then do all the functionality inside the wallet instead of the AFJ level. 

- Adds bls12381g1 and bls12381g2 to `indyWallet.createKey`
- Adds bls12381g2 to `indyWallet.sign`
- Adds bls12381g2 to `indyWallet.verify`
- Store and retrieve the keyPair in the wallet via the public key
- Added tests for the indy wallet


docs/setup-react-native.md
```diff
@@ -81,3 +81,23 @@ try {
...
+When using AFJ inside the React Native environment, temporarily, a dependency for creating keys, sigining and verifying
+with bbs keys must be swapped. Inside your package.json the following must be added:
+
+```diff
++   "resolutions": {
++     "@mattrglobal/bbs-signatures": "react-native-bbs-signatures@0.1.0",
++   },
+    "dependencies": {
+      ...
++     "react-native-bbs-signatures": "0.1.0",
```
TimoGlastra, line 96: Can we add `react-native-bbs-signatures` as a peer dependency to the RN wrapper?

packages/core/src/crypto/BbsService.ts
```diff
@@ -0,0 +1,151 @@
...
+import { KeyType } from './KeyType'
+
+export interface BlsKeyPair {
+  publicKeyBase58: string
+  privateKeyBase58: string
+  keyType: Exclude<KeyType, KeyType.Ed25519 | KeyType.X25519>
+}
+
+interface BbsCreateKeyOptions extends CreateKeyOptions {
+  keyType: Exclude<KeyType, KeyType.X25519 | KeyType.Ed25519 | KeyType.Bls12381g1g2>
```
TimoGlastra, line 25: Maybe we can invert this type so it still works if new key types are added?

```suggestion
  keyType: Pick<KeyType, KeyType.Bls12381g1 | KeyType.Bls12381g2>
```

packages/core/src/utils/TypedArrayEncoder.ts
```diff
@@ -60,4 +60,18 @@ export class TypedArrayEncoder {
...
+   *
+   * @param array unknown The array that has to be checked
+   *
+   * @returns A boolean if the array is a byte array
+   */
+  public static isTypedArray(array: unknown): boolean {
+    // Checks whether the static property 'BYTES_PER_ELEMENT' exists on the provided array.
+    // This has to be done, since the TypedArrays, e.g. Uint8Array and Float32Array, do not
+    // extend a single base class
+    return 'BYTES_PER_ELEMENT' in (array as Record<string, unknown>)
```
TimoGlastra, line 75: Does this also work with buffer?

packages/core/src/wallet/IndyWallet.ts
```diff
@@ -324,6 +328,110 @@ export class IndyWallet implements Wallet {
...
+   * @returns a Key instance with a publicKeyBase58
+   *
+   * @throws {WalletError} When an unsupported keytype is requested
+   * @throws {WalletError} When the key could not be created
+   */
+  public async createKey({ seed, keyType }: CreateKeyOptions): Promise<Key> {
+    try {
+      if (keyType === KeyType.Ed25519) {
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        //@ts-ignore
```
TimoGlastra, line 349: Might be good to add this to the definitely typed package

packages/core/src/wallet/IndyWallet.ts
```diff
@@ -324,6 +328,110 @@ export class IndyWallet implements Wallet {
...
+   * @param data Buffer The data that needs to be signed
+   * @param key Key The key that is used to sign the data
+   *
+   * @returns A signature for the data
+   */
+  public async sign({ data, key }: SignOptions): Promise<Buffer> {
+    try {
+      if (key.keyType === KeyType.Ed25519) {
+        // Checks to see if it is an not an Array of messages, but just a single one
+        if (typeof data[0] !== 'number') {
```
TimoGlastra, line 380: Can we also use `IsTypedArray()` here?

packages/core/src/wallet/IndyWallet.ts
```diff
@@ -324,6 +328,110 @@ export class IndyWallet implements Wallet {
...
+   * @returns A boolean whether the signature was created with the supplied data and key
+   *
+   * @throws {WalletError} When it could not do the verification
+   * @throws {WalletError} When an unsupported keytype is used
+   */
+  public async verify({ data, key, signature }: VerifyOptions): Promise<boolean> {
+    try {
+      if (key.keyType === KeyType.Ed25519) {
+        // Checks to see if it is an not an Array of messages, but just a single one
+        if (typeof data[0] !== 'number') {
```
TimoGlastra, line 418: Same here

TimoGlastra (approved):
Nice!

packages/core/src/wallet/IndyWallet.ts
blu3beri: Yes, added some tests.

packages/react-native/package.json
```diff
@@ -42,6 +42,7 @@
   "peerDependencies": {
     "indy-sdk-react-native": "^0.1.16",
     "react-native-fs": "^2.18.0",
-    "react-native-get-random-values": "^1.7.0"
+    "react-native-get-random-values": "^1.7.0",
+    "react-native-bbs-signatures": "0.1.0"
```
TimoGlastra, line 46: This should be `@animo-id/react-native-bbs-signatures` right? And the version should be `0.0.3`

TimoGlastra: To make sure the package actually exist, maybe we can add it as a dev-dependency to the RN package.json?

docs/setup-react-native.md
```diff
@@ -81,3 +81,25 @@ try {
...
 ```
+
+## Using BBS Signatures
+
+When using AFJ inside the React Native environment, temporarily, a dependency for creating keys, sigining and verifying
+with bbs keys must be swapped. Inside your package.json the following must be added:
+
+```diff
++   "resolutions": {
++     "@mattrglobal/bbs-signatures": "react-native-bbs-signatures@0.1.1",
```
TimoGlastra, line 92: should be `@animo-id/react-native-bbs-signatures` with version `0.0.3` (we probably want to release 0.1.0 right?)

TimoGlastra, line 92: Resolution field only work when using yarn. We should add an example for the NPM equivalent: https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides (would be good to test if this works).

I think bifold uses npm. We also need to mention you need at least NPM 8 for the `overrides` to work (if using NPM)

DEVREADME.md
```diff
@@ -2,6 +2,11 @@
 
 This file is intended for developers working on the internals of the framework. If you're just looking how to get started with the framework, see the [docs](./docs)
 
+## Installing dependencies
+
+Right now, as a patch that will later be changed, some platforms will have an "error" when installing the dependencies with yarn. This is because the BBS signatures library that we use is built for Linux x86 and MacOS x86 (and not Windows and MacOS arm). This means that it will show that it could not download the binary.
+This is not an error for developers, the library that fails is `node-bbs-signaturs` and is an optional dependency for perfomance improvements. It will fallback to a, slower, wasm build.
+
 ## Running tests
```
TimoGlastra, line 10: We should probably add this to the installation docs instead of the developer docs right?

docs/setup-react-native.md
blu3beri: Yes when I finished some work on it I will release a 0.1.0

DEVREADME.md
blu3beri: Yeah we can move it there and reference it from here.

docs/setup-react-native.md
blu3beri: Confirmed to work:


```
"node_modules/@mattrglobal/bbs-signatures": {
      "resolved": "node_modules/@mattrglobal/jsonld-signatures-bbs/@animo-id/react-native-bbs-signatures@0.1.0",
      "link": true
},

```

blu3beri:
Not quite sure what happened to the commits, but since we squash it is okay. (every commit is double)

blu3beri:
> I looked at the W3C PR and there are parts of this PR missing in that PR.
> 
> I think we want to merge this into a `next` branch for now. What do you think? Then the w3c pr can also target the next branch. Don't want to add all these dependencies in 0.2.0 if not used yet. Maybe we'll be able to solve this issue (#730) before 0.3.0 release so users are not required to bundle the react native bbs signatures if they're only using indy credentials for example.

I agree with not adding it in before the 0.2.0 release. We can merge it after that release or move it to the `next` branch. I do not really have a big preference.

blu3beri:
@TimoGlastra base is now `next`.

PR #687, opened 2022-03-29
refactor: remove verkeys and did docs from connection record

jakubkoci:


packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -143,15 +167,23 @@ export class ConnectionService {
...
       autoAcceptConnection: outOfBandRecord.autoAcceptConnection,
       tags: {
         invitationKey: outOfBandRecord.getTags().recipientKey,
       },
     })
 
-    connectionRecord.theirDid = message.connection.did
-    connectionRecord.theirDidDoc = message.connection.didDoc
+    // TODO Could we use authentication key from did doc instead of did comm services?
+    const theirVerkeys = message.connection.didDoc?.didCommServices[0].recipientKeys ?? []
```
TimoGlastra, line 179: This is a tricky one. RecipientKeys is where we should _send_ messages to, but doesn't necesarily mean we will receive messages using these keys. So for completeness we should probably use all authentication keys (as legacy ed25519) and all keyagreement keys?

packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -170,15 +202,51 @@ export class ConnectionService {
...
     connectionRecord: ConnectionRecord,
-    outOfBandRecord: OutOfBandRecord
+    outOfBandRecord: OutOfBandRecord,
+    routing?: Routing
   ): Promise<ConnectionProtocolMsgReturnType<ConnectionResponseMessage>> {
     this.logger.debug(`Create message ${ConnectionResponseMessage.type} start`, connectionRecord)
     connectionRecord.assertState(ConnectionState.Requested)
     connectionRecord.assertRole(ConnectionRole.Inviter)
 
+    const { did } = routing ? routing : outOfBandRecord
```
TimoGlastra, line 212: Will this work?

```suggestion
    const { did } = routing ?? outOfBandRecord
```

I'm a bit unsure of what this does. In what case is routing undefined, and in what case is the did on the oob record undefined?

packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -170,15 +202,51 @@ export class ConnectionService {
...
+          routingKeys: Array.from(
+            new Set(
+              outOfBandRecord.outOfBandMessage.services
+                .filter((s): s is DidCommService => typeof s !== 'string')
+                .map((s) => s.routingKeys)
+                .filter((r): r is string[] => r !== undefined)
+                .reduce((acc, curr) => acc.concat(curr), [])
+            )
+          ),
+        }
```
TimoGlastra, line 238:
```suggestion
    let didDocRouting = routing
    
    if (!didRouting) {
    didDocRouting = {
          did,
          endpoints: Array.from(
            new Set(
              outOfBandRecord.outOfBandMessage.services
                .filter((s): s is DidCommService => typeof s !== 'string')
                .map((s) => s.serviceEndpoint)
            )
          ),
          verkey: outOfBandRecord.getTags().recipientKey,
          routingKeys: Array.from(
            new Set(
              outOfBandRecord.outOfBandMessage.services
                .filter((s): s is DidCommService => typeof s !== 'string')
                .map((s) => s.routingKeys)
                .filter((r): r is string[] => r !== undefined)
                .reduce((acc, curr) => acc.concat(curr), [])
            )
          ),
        }
   }
```

packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -249,10 +313,15 @@ export class ConnectionService {
...
+      throw error
     }
 
     const connection = JsonTransformer.fromJSON(connectionJson, Connection)
-    await MessageValidator.validate(connection)
+    try {
+      await MessageValidator.validate(connection)
+    } catch (error) {
+      throw new Error(error)
+    }
```
TimoGlastra, line 324: We should probably add this wrapper in the `MessageValidator` itself

packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -267,14 +336,20 @@ export class ConnectionService {
...
-    connectionRecord.theirDidDoc = connection.didDoc
-    connectionRecord.threadId = message.threadId
-
-    if (!connectionRecord.theirKey) {
-      throw new AriesFrameworkError(`Connection with id ${connectionRecord.id} has no recipient keys.`)
+    if (!connection.didDoc) {
+      throw new AriesFrameworkError('DID Document is missing.')
     }
 
+    const theirVerkeys = connection.didDoc?.didCommServices[0].recipientKeys ?? []
```
TimoGlastra, line 343: We should take all keys from all services I think?

packages/core/src/modules/connections/services/helpers.ts
```diff
@@ -0,0 +1,49 @@
...
+      didDocumentBuilder.addAuthentication(ed25519VerificationMethod)
+    }
+  })
+
+  didDoc.publicKey.forEach((pk) => {
+    if (pk.type === 'Ed25519VerificationKey2018' && pk.value) {
+      const ed25519VerificationMethod = convertPublicKeyToVerificationMethod(pk)
+      didDocumentBuilder.addVerificationMethod(ed25519VerificationMethod)
+    }
+  })
```
TimoGlastra, line 26: How does this work with references?

packages/core/src/modules/oob/OutOfBandModule.ts
```diff
@@ -464,20 +466,18 @@ export class OutOfBandModule {
...
-            existingConnection = await this.connectionsModule.findByDid(theirDidRecord.id)
-          }
+          existingConnection = await this.connectionsModule.findByDid(theirDidRecord.id)
+          // TODO what if we have more connections?
+          if (existingConnection) return existingConnection
         }
-
-        return existingConnection
       }
+      return existingConnection
```
TimoGlastra, line 480: We shouldn't find existing connections based on the recipientKeys. We should either transform the service into a peer did method 2, or use the did string.

Not sure we should add this logic as it's not conformant to the RFC

packages/core/src/modules/connections/repository/ConnectionRecord.ts
```diff
@@ -1,19 +1,16 @@
...
-import { ConnectionRole, DidExchangeRole, DidDoc, ConnectionState, DidExchangeState } from '../models'
+import { ConnectionRole, DidExchangeRole, ConnectionState, DidExchangeState } from '../models'
 
 export interface ConnectionRecordProps {
   id?: string
   createdAt?: Date
   did: string
-  didDoc: DidDoc
-  verkey: string
+  didDoc?: DidDoc
```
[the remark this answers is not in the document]
jakubkoci: Removed

packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -87,31 +96,42 @@ export class ConnectionService {
 
     const { outOfBandMessage } = outOfBandRecord
 
+    const { did, mediatorId } = config.routing
+    const didDoc = this.createDidDoc(config.routing)
```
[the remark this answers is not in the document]
jakubkoci: Yes, something like that could work 👍

packages/core/src/agent/MessageSender.ts
```diff
@@ -192,16 +194,20 @@ export class MessageSender {
       outOfBand
     )
 
+    const senderDidDocument = await this.resolveDidDocument(connection.did)
+    const senderAuthenticationKeys = getAuthenticationKeys(senderDidDocument)
+    // TODO We're selecting just the first authentication key. Is it ok?
+    const [firstSenderAuthenticationKey] = senderAuthenticationKeys
+    const shouldUseReturnRoute = !this.transportService.hasInboundEndpoint(senderDidDocument)
```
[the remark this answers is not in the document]
jakubkoci: Somehow seemed obvious to name it `sender` while I was writing :) 🤷‍♂️  but no problem with `outDid` we use usually it for similar cases across the framework, I agree with renaming.

What do you think about the `TODO`: "We're selecting just the first authentication key. Is it ok?"

packages/core/src/agent/MessageSender.ts
```diff
@@ -318,10 +324,6 @@ export class MessageSender {
       this.logger.debug(`Resolving services for connection theirDid ${connection.theirDid}.`)
```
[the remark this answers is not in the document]
jakubkoci: Goot catch 👍  Removed

packages/core/src/modules/connections/services/ConnectionService.ts
jakubkoci: I somehow ended up creating a new routing when there is a reusable oob invitation, but maybe it's not needed.

packages/core/src/modules/connections/ConnectionsModule.ts
```diff
@@ -212,26 +216,26 @@ export class ConnectionsModule {
...
-   * @param verkey the verkey to search for
-   * @returns the connection record, or null if not found
-   * @throws {RecordDuplicateError} if multiple connections are found for the given verkey
-   */
-  public findByVerkey(verkey: string): Promise<ConnectionRecord | null> {
-    return this.connectionService.findByVerkey(verkey)
-  }
+  public async findByKeys({ senderKey, recipientKey }: { senderKey: string; recipientKey: string }) {
+    let connectionRecord
+    const theirDidRs = await this.didRepository.findMultipleByVerkey(senderKey)
```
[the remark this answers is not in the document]
jakubkoci: It's currently possible because we don't check if there is did doc for a given sender key. We should add this check. In the same way, it doesn't make sense to have more did doc for the same recipient key, does it?

My previous probably bad assumption was that I can create a connection request with a different did doc but signed by the same sender key as a previously created connection (for reusable invitation), then the creation of such did doc record won't fail on duplicated id error.

packages/core/src/modules/connections/services/ConnectionService.ts
jakubkoci: Hmm, I see. To me it seems the reason we're storing the recipient keys here is to later find did doc based on sender keys. We can always find other keys via did doc services if needed somewhere else. I agree we should use keys from more didcomm services but I would rather store just authentication keys.

packages/core/src/modules/routing/RecipientModule.ts
```diff
@@ -98,9 +102,9 @@ export class RecipientModule {
     })
 
     const websocketSchemes = ['ws', 'wss']
-    const hasWebSocketTransport = connectionRecord.theirDidDoc?.didCommServices?.some((s) =>
-      websocketSchemes.includes(s.protocolScheme)
-    )
+    const didRecord = connectionRecord.theirDid && (await this.dids.findByDid(connectionRecord.theirDid))
+    const services = didRecord && didRecord.didDocument?.didCommServices
+    const hasWebSocketTransport = services && services.some((s) => websocketSchemes.includes(s.protocolScheme))
```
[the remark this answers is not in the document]
jakubkoci: Good point. Updated.

packages/core/src/modules/connections/services/helpers.ts
jakubkoci: It doesn't. How can I create `DidDoc` with referenced auth? if I do the following it doesn't work:
```js
const didDoc = new DidDoc({
  authentication: [
    new EmbeddedAuthentication(
      new Ed25119Sig2018({
        id: 'did:sov:SKJVx2kn373FNgvff1SbJo#4',
        controller: 'did:sov:SKJVx2kn373FNgvff1SbJo',
        publicKeyBase58: 'EoGusetSxDJktp493VCyh981nUnzMamTRjvBaHZAy68d',
      })
    ),
    new ReferencedAuthentication(
      new Ed25119Sig2018({
        id: 'did:sov:SKJVx2kn373FNgvff1SbJo#5',
        controller: 'did:sov:SKJVx2kn373FNgvff1SbJo',
        publicKeyBase58: 'EoGusetSxDJktp493VCyh981nUnzMamTRjvBaHZAy68d',
      }),
      'Ed25519VerificationKey2018'
    ),
  ],
  id: 'test-id',
  publicKey: [
    new Ed25119Sig2018({
      id: 'did:sov:SKJVx2kn373FNgvff1SbJo#4',
      controller: 'did:sov:SKJVx2kn373FNgvff1SbJo',
      publicKeyBase58: 'EoGusetSxDJktp493VCyh981nUnzMamTRjvBaHZAy68d',
    }),
    new Ed25119Sig2018({
      id: 'did:sov:SKJVx2kn373FNgvff1SbJo#5',
      controller: 'did:sov:SKJVx2kn373FNgvff1SbJo',
      publicKeyBase58: 'EoGusetSxDJktp493VCyh981nUnzMamTRjvBaHZAy68d',
    }),
  ],
  service: [
    new IndyAgentService({
      id: 'did:sov:SKJVx2kn373FNgvff1SbJo#service-1',
      serviceEndpoint: 'did:sov:SKJVx2kn373FNgvff1SbJo',
      recipientKeys: ['EoGusetSxDJktp493VCyh981nUnzMamTRjvBaHZAy68d'],
      routingKeys: ['EoGusetSxDJktp493VCyh981nUnzMamTRjvBaHZAy68d'],
      priority: 5,
    }),
  ],
})
```

I would expect that it returns this JSON:
```js
{
  "context": "https://w3id.org/did/v1",
  "publicKey": [
    {
      "id": "did:sov:SKJVx2kn373FNgvff1SbJo#4",
      "controller": "did:sov:SKJVx2kn373FNgvff1SbJo",
      "type": "Ed25519VerificationKey2018",
      "value": "EoGusetSxDJktp493VCyh981nUnzMamTRjvBaHZAy68d"
    },
    {
      "id": "did:sov:SKJVx2kn373FNgvff1SbJo#5",
      "controller": "did:sov:SKJVx2kn373FNgvff1SbJo",
      "type": "Ed25519VerificationKey2018",
      "value": "EoGusetSxDJktp493VCyh981nUnzMamTRjvBaHZAy68d"
    }
  ],
  "service": [
    {
      "id": "did:sov:SKJVx2kn373FNgvff1SbJo#service-1",
      "serviceEndpoint": "did:sov:SKJVx2kn373FNgvff1SbJo",
      "type": "IndyAgent",
      "priority": 5,
      "recipientKeys": ["EoGusetSxDJktp493VCyh981nUnzMamTRjvBaHZAy68d"],
      "routingKeys": ["EoGusetSxDJktp493VCyh981nUnzMamTRjvBaHZAy68d"]
    }
  ],
  "authentication": [
    {
      "publicKey": {
        "id": "did:sov:SKJVx2kn373FNgvff1SbJo#4",
        "controller": "did:sov:SKJVx2kn373FNgvff1SbJo",
        "type": "Ed25519VerificationKey2018",
        "value": "EoGusetSxDJktp493VCyh981nUnzMamTRjvBaHZAy68d"
      },
      "type": "Ed25519VerificationKey2018"
    },
    "did:sov:SKJVx2kn373FNgvff1SbJo#5"
  ],
  "id": "test-id"
}

```

But it returns this one
```
{
  "context": "https://w3id.org/did/v1",
  "publicKey": [
    {
      "id": "3",
      "controller": "did:sov:LjgpST2rjsoxYegQDRm7EL",
      "type": "RsaVerificationKey2018",
      "value": "-----BEGIN PUBLIC X..."
    },
    {
      "id": "did:sov:SKJVx2kn373FNgvff1SbJo#4",
      "controller": "did:sov:SKJVx2kn373FNgvff1SbJo",
      "type": "Ed25519VerificationKey2018",
      "value": "EoGusetSxDJktp493VCyh981nUnzMamTRjvBaHZAy68d"
    },
    {
      "id": "did:sov:SKJVx2kn373FNgvff1SbJo#5",
      "controller": "did:sov:SKJVx2kn373FNgvff1SbJo",
      "type": "Ed25519VerificationKey2018",
      "value": "EoGusetSxDJktp493VCyh981nUnzMamTRjvBaHZAy68d"
    },
    {
      "id": "6",
      "controller": "did:sov:LjgpST2rjsoxYegQDRm7EL",
      "type": "Secp256k1VerificationKey2018",
      "value": "-----BEGIN PUBLIC A..."
    }
  ],
  "service": [
    {
      "id": "did:sov:SKJVx2kn373FNgvff1SbJo#service-1",
      "serviceEndpoint": "did:sov:SKJVx2kn373FNgvff1SbJo",
      "type": "IndyAgent",
      "priority": 5,
      "recipientKeys": ["EoGusetSxDJktp493VCyh981nUnzMamTRjvBaHZAy68d"],
      "routingKeys": ["EoGusetSxDJktp493VCyh981nUnzMamTRjvBaHZAy68d"]
    }
  ],
  "authentication": [
    {
      "publicKey": {
        "id": "did:sov:SKJVx2kn373FNgvff1SbJo#4",
        "controller": "did:sov:SKJVx2kn373FNgvff1SbJo",
        "type": "Ed25519VerificationKey2018",
        "value": "EoGusetSxDJktp493VCyh981nUnzMamTRjvBaHZAy68d"
      }
    },
    {
      "publicKey": {
        "id": "did:sov:SKJVx2kn373FNgvff1SbJo#5",
        "controller": "did:sov:SKJVx2kn373FNgvff1SbJo",
        "type": "Ed25519VerificationKey2018",
        "value": "EoGusetSxDJktp493VCyh981nUnzMamTRjvBaHZAy68d"
      },
      "type": "Ed25519VerificationKey2018"
    }
  ],
  "id": "test-id"
}
```

packages/core/src/modules/connections/services/ConnectionService.ts
jakubkoci: Do you think we should reuse keys from oob invitation even if it's reusable?

packages/core/src/modules/oob/OutOfBandModule.ts
jakubkoci: Agree, as we discussed in a call we should use peer did method 2. I'll do it in a separate PR.

packages/core/src/modules/connections/handlers/ConnectionResponseHandler.ts
```diff
@@ -29,7 +33,14 @@ export class ConnectionResponseHandler implements Handler {
...
     }
 
-    const connectionRecord = await this.connectionService.findByVerkey(messageContext.recipientVerkey)
+    const { recipientVerkey } = messageContext
+
+    let connectionRecord
+    const ourDidRecords = await this.didRepository.findMultipleByVerkey(recipientVerkey)
+    for (const ourDidRecord of ourDidRecords) {
+      connectionRecord = await this.connectionService.findByOurDid(ourDidRecord.id)
+    }
```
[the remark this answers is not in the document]
jakubkoci: Good point. I changed it. We use thread ID now, but shouldn't we still check if recipient keys are the same as we have in our did record?

I added this:
```ts
    if (!ourDidDocument.recipientKeys.includes(recipientVerkey)) {
      throw new AriesFrameworkError(`Recipient key ${recipientVerkey} not found in did document recipient keys.`)
    }
```

I thought, what if someone has a connection with an agent (therefore known sender key) and somehow gets the thread ID, then he could send a response with different recipient keys.

packages/core/src/modules/connections/services/ConnectionService.ts
jakubkoci: Agree. But probably in separate PR.

packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -87,31 +96,42 @@ export class ConnectionService {
 
     const { outOfBandMessage } = outOfBandRecord
 
+    const { did, mediatorId } = config.routing
```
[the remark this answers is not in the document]
jakubkoci: Good idea. We can do it. I assume you meant "We should NOT be now passing unqualified dids", right?

packages/core/src/modules/connections/handlers/ConnectionRequestHandler.ts
```diff
@@ -39,11 +39,15 @@ export class ConnectionRequestHandler implements Handler {
...
+    // TODO: Allow rotation of keys used in the invitation for new ones not only when out-of-band is reusable
+    let routing
+    if (outOfBandRecord.reusable) {
+      routing = await this.mediationRecipientService.getRouting()
+    }
+    const connectionRecord = await this.connectionService.processRequest(messageContext, outOfBandRecord, routing)
 
     if (connectionRecord?.autoAcceptConnection ?? this.agentConfig.autoAcceptConnections) {
-      const { message } = await this.connectionService.createResponse(connectionRecord, outOfBandRecord)
+      const { message } = await this.connectionService.createResponse(connectionRecord, outOfBandRecord, routing)
```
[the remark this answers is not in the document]
jakubkoci: It's not needed when the oob is not reusable. But, as I asked somewhere else, if we want to reuse routing even if it's reusable, I'll updated it and we can remove it from here.

PR #688, opened 2022-03-30
feat: extension module creation

genaris:
Export some convenience types and classes to facilitate extension/plug-in development and add a fully-working dummy extension module in samples directory.

I'm not sure if it's the right place to put all this stuff, but I believe it could help developers to add custom protocols and wallet storage to their controllers, so it's worth for it to be somewhere :-).


karimStekelenburg:
This is a really cool addition! Nice job @genaris!

genaris:
Thanks guys! You're right @TimoGlastra , it's very frustrating when you want to use some examples and they do not work out-of-the-box because they are outdated, So in a further PR I'll add some tests to be run alongside framework's test suite.

JamesKEbert:
Also, just wanted to say good work @genaris! I appreciate the addition here, I've been wanting to get this functionality working in AFJ for a while now.

PR #689, opened 2022-03-31
fix: did sov service type resolving

JamesKEbert:
This PR fixes behavior with the service resolution & types behavior with did:sov resolution. This additionally renames the `IndyDidResolver` to `SovDidResolver` to reduce confusion and prevent collision with the did:indy method in the future. Happy for any feedback/thoughts here.

@TimoGlastra if you could take a look at this one to ensure I didn't miss something here that'd be appreciated!


TimoGlastra (approved):
LGTM!

TimoGlastra:
@jameskebert Could you add a BREAKING CHANGE section to the PR body describing the change?

JamesKEbert:
I'd be happy to add one, but I guess since this is just adding services where they wouldn't be previously it doesn't at least initially strike me as a breaking change, just an addition?

TimoGlastra:
Ah you're right! nvm

TimoGlastra:
@JamesKEbert  could you fix the broken tests so we can merge? Thanks!

Issue #602, opened 2022-01-21
Delete operations does not delete stuff from wallet


jakubkoci:
I found that there are `deleteById` methods for connections, credentials, and proofs. However, these methods just remove a record from non-secrets storage and not from a wallet. Therefore, there will be related items like keys and credentials still stored in the wallet.

As far as I know, wallet API doesn't have any delete operations available. Does anyone know if it's designed in that way on purpose? 

We should at least document this behavior for now.

blu3beri:
I am not too sure about the design specifics. But it is possible those completely delete them via `indy-sdk` or `indy-sdk-react-native`. 

I do think it is a good idea to add a way to delete creds/conns/proofs. I had once where in a wallet I deleted the credential(record) and I could still fullfill a proof request with that credential. 

We could expand the `deleteById` with a param that takes a `bool` whether to delete the cred/conn/proof aswell or create a new method. My preference goes to the extra parameter.

TimoGlastra:
I think for proof there's not anything we'd like to delete. 

For connections it's a bit more complicated (we should make sure the key is only used for that connection) but would be good to do some housekeeping sometimes. Also with connectionless exchanges would be good to throw away keys after the exchange.

@blu3beri could you make a PR so that the credential is also deleted from indy storage (after we've discussed whether it should be a separate method or a parameter. I also like the parameter approach the most I think)? Then we can tackle the connection keys in a separate PR

PR #691, opened 2022-03-31
feat: delete credential from wallet

janrtvld:
As discussed in #602, deleting a credential can now also be deleted from the wallet.


packages/core/src/modules/indy/services/IndyHolderService.ts
```diff
@@ -205,6 +206,24 @@ export class IndyHolderService {
...
 
+  /**
+   * Delete a credential stored in the wallet by id.
+   *
+   * @param credentialId the id (referent) of the credential
+   *
+   */
+  public async deleteCredential(credentialId: Indy.CredentialId): Promise<void> {
+    try {
+      return await this.indy.deleteWalletRecord(this.wallet.handle, CredentialRecord.type, credentialId)
```
TimoGlastra, line 217: You need to call `proverDeleteCredential`

https://github.com/hyperledger/indy-sdk/blob/master/wrappers/nodejs/src/index.js#L188-L192

janrtvld: `indy.proverDeleteCredential` is not exposed.. ?

packages/core/src/modules/credentials/__tests__/CredentialService.test.ts
```diff
@@ -1087,6 +1090,27 @@ describe('CredentialService', () => {
...
+    it('should call delete from repository', async () => {
+      const credential = mockCredentialRecord()
+      mockFunction(credentialRepository.getById).mockReturnValue(Promise.resolve(credential))
+
+      const repositoryDeleteSpy = jest.spyOn(credentialRepository, 'delete')
+      await credentialService.deleteById(credential.id)
+      expect(repositoryDeleteSpy).toHaveBeenNthCalledWith(1, credential)
+    })
+
+    it('deleteFromWallet parameter should call deleteCredential in indyHolderService with credentialId', async () => {
```
TimoGlastra, line 1103:
```suggestion
    it('deleteAssociatedCredential parameter should call deleteCredential in indyHolderService with credentialId', async () => {
```

TimoGlastra (approved):

Issue #553, opened 2021-11-28
Add support for postgres wallet type


TimoGlastra:

sairanjitAW:
Hi @TimoGlastra I am working on adding a support for postgres

TimoGlastra:
That's great @sairanjitAW!

swcurran:
Not that I want to discourage any efforts in building the platform, but is this a good use of time?  This implies using AFJ on the server-side. AFAIK there is a fair amount still to add to AFJ to make it useful/scalable for operation on a server platform.  Is there a reason not to push that effort into ACA-Py for server-side uses and focus AFJ efforts towards mobile functionality?

swcurran:
Sounds good!  Happy with that answer.  :-)

genaris:
> I strongly disagree with this, and we're really interested in building out AFJ as a platform for both mobile and server side use cases. ACA-Py has had _A LOT_ more investment for server side applications, but there's multiple reasons why we're interested in using AFJ for the server side (Being able to share the same codebase and protocols between environments, AFJ is more strictly typed which is a big + IMO, TS is a very popular language, a lot more libraries available in TS for SSI/Blockchain related projects than python, etc...).
> 
> So I'd definitely say this is a good use of time!

+1 to Timo's words. For small teams it could be very beneficial to be able to share the same code base and use a single language/toolchain, especially when implementing custom features such as specific protocols over DIDComm. Also I like the 'framework' approach in terms that `Agent` class can be easily embedded within an existing application instead of using a separate module for handling Aries stuff.

TimoGlastra:
Fixed

PR #699, opened 2022-04-07
feat: add support for postgres wallet type

sairanjitAW:
Signed-off-by: Sai Ranjit Tummalapalli <sairanjit.tummalapalli@ayanworks.com>

Related issue: #553


packages/core/src/wallet/IndyWallet.ts
```diff
@@ -33,6 +34,14 @@ export class IndyWallet implements Wallet {
   public constructor(agentConfig: AgentConfig) {
     this.logger = agentConfig.logger
     this.indy = agentConfig.agentDependencies.indy
+
+    const { walletConfig, fileSystem } = agentConfig
+
+    if (walletConfig?.storageType === WalletStorageType.Postgres) {
+      if (walletConfig.storageConfig && walletConfig.storageCreds && fileSystem?.loadPostgresPlugin) {
+        fileSystem?.loadPostgresPlugin(walletConfig.storageConfig, walletConfig.storageCreds)
+      }
```
TimoGlastra, line 43: If the wallet storage is set to postgres, I think we should throw an error if the required parameters are not set. I think we should also throw an error if `loadPostgresPlugin` is not defined and wallet storage type is postgres.

packages/core/tests/agents.test.ts
```diff
@@ -77,3 +83,65 @@ describe('agents', () => {
     expect(aliceAgent.isInitialized).toBe(true)
   })
 })
+
+describe('postgres agents', () => {
```
TimoGlastra, line 87: Maybe we can extract this to a separate postgres.test.ts?

packages/core/src/storage/FileSystem.ts
```diff
@@ -1,8 +1,34 @@
...
+export interface WalletStorageCreds {
+  [key: string]: unknown
+}
+
+export interface StorageCreds {
+  account: string
+  password: string
+  admin_account: string
+  admin_password: string
 }
```
TimoGlastra, line 34: The postgres plugin is related to the environment and the wallet type (in this case indy-sdk). I'm not sure the file system is the best place to add this, mainly because the filesystem stays the same if we support something different than indy-sdk.


Maybe we can keep it a bit more generic. Maybe later we can integrate more, but for now maybe can approach it like this:

```ts
import { Agent } from '@aries-framework/core'
import { agentDependencies, loadPostgresPlugin } from '@aries-framework/node'

async function run() {
  const storageConfig = {}
  const storageCreds = {}

  await loadPostgresPlugin(storageConfig, storageCreds)

  const agent = new Agent({
    walletConfig: {
      id: 'id',
      key: 'key',
      storageConfig,
      storageCreds,
      storageType: 'postgres'
  })
}

run()
```

packages/core/src/types.ts
```diff
@@ -20,6 +22,9 @@ export interface WalletConfig {
   id: string
   key: string
   keyDerivationMethod?: KeyDerivationMethod
+  storageType?: WalletStorageType
+  storageConfig?: WalletStorageConfig
+  storageCreds?: WalletStorageCreds
```
TimoGlastra, line 27: Maybe we can chamge the interface of this to be not 100% what indy expects, but more how we would want the user of the framework to provide the values. Mainly with the addition of other wallet types I'm not sure if this approach fits nicely. I was thinking about something like this:

```ts
// defined in types.ts
  id: string
  key: string
  keyDerivationMethod?: KeyDerivationMethod
  storage?: {
    type: string
    [key: string]: unknown
  }
}

// defined in @aries-framework/node package
interface IndyPostgresStorageConfig {
  type: 'postgres',
  config: {
    // .. parameters needed for postgres plugin
  },
   credentials: {
    // .. parameters needed for postgres plugin
  }
}
```

It can then be used like this:

```ts
import { IndyPostgresStorageConfig } from '@aries-framework/core'

const storageConfig: IndyPostgresStorageConfig = {
  type: 'postgres',
  // other required fields
}

await loadPostgresPlugin()

const agent = new Agent({
  walletConfig: {
     // other required fields
    storage: storageConfig
  }
})

await agent.initialize()
```

packages/node/package.json
```diff
@@ -30,12 +30,16 @@
   "dependencies": {
     "@aries-framework/core": "0.1.0",
     "express": "^4.17.1",
+    "ffi-napi": "^4.0.3",
     "indy-sdk": "^1.16.0-dev-1636",
     "node-fetch": "^2.6.1",
-    "ws": "^7.5.3"
+    "ref-napi": "^3.0.3",
+    "ws": "^7.5.3",
+    "os": "^0.1.2"
```
TimoGlastra, line 38: This is a node core module, we don't need to add it to the dependencies

```suggestion
```

TimoGlastra (changes requested):
Thanks for the great contribution @sairanjitAW! I left a few notes, mainly about keeping the api generic for other wallet types / storage types coming in the near future.

@blu3beri could you take a look at the postgres binary initialization and see if that looks good?

packages/node/package.json
```diff
@@ -30,12 +30,16 @@
...
   },
   "devDependencies": {
     "@types/express": "^4.17.13",
```
blu3beri, line 41: missing `@types/ref-napi`

packages/node/src/NodeFileSystem.ts
```diff
@@ -69,4 +72,10 @@ export class NodeFileSystem implements FileSystem {
         })
     })
   }
+
+  public async loadPostgresPlugin(storageConfig: WalletStorageConfig, storageCreds: WalletStorageCreds) {
+    await storagePlugin.postgresstorage_init()
```
blu3beri, line 77: I don't believe this is asynchronous. Could be wrong, but by default `ffi-napi` is sync.

blu3beri (changes requested):
Great addition! Just 1 comment on the initialisation of the postgres plugin.

packages/core/src/types.ts
```diff
@@ -16,10 +16,24 @@ export const enum KeyDerivationMethod {
   Raw = 'RAW',
 }
 
+export enum WalletScheme {
+  DatabasePerWallet = 'DatabasePerWallet',
+  MultiWalletSingleTable = 'MultiWalletSingleTable',
+  MultiWalletSingleTableSharedPool = 'MultiWalletSingleTableSharedPool',
+}
```
TimoGlastra, line 23: This is  a postgres wallet feature, so maybe we can also move this to the node package?

packages/core/src/types.ts
```diff
@@ -16,10 +16,24 @@ export const enum KeyDerivationMethod {
...
   id: string
   key: string
   keyDerivationMethod?: KeyDerivationMethod
+  storage?: {
+    type: string
+    [key: string]: unknown
+  }
+}
+
+export interface WalletStorageCreds {
```
TimoGlastra, line 35: Same here, I think we can leave this out

packages/core/src/types.ts
```diff
@@ -47,6 +61,11 @@ export enum DidCommMimeType {
   V1 = 'application/didcomm-envelope-enc',
 }
 
+export enum WalletStorageType {
```
TimoGlastra, line 64: This can also be removed now

packages/core/src/wallet/IndyWallet.ts
```diff
@@ -67,6 +68,41 @@ export class IndyWallet implements Wallet {
...
+      key_derivation_method: walletConfig.keyDerivationMethod,
+    }
+    if (rekey) {
+      walletCredentials.rekey = rekey
+    }
+    if (rekeyDerivation) {
+      walletCredentials.rekey_derivation_method = rekeyDerivation
+    }
+    if (walletConfig.storage?.credentials) {
+      walletCredentials.storage_credentials = walletConfig.storage?.credentials as WalletStorageCreds
```
TimoGlastra, line 100:
```suggestion
      walletCredentials.storage_credentials = walletConfig.storage?.credentials as Record<string, unknown>
```

packages/node/src/PostgresPlugin.ts
```diff
@@ -0,0 +1,88 @@
...
+  postgresstorage_init: () => number
+  init_storagetype: (arg0: string, arg1: string) => number
+}
+
+export interface IndyPostgresStorageConfig {
+  type: 'postgres_storage'
+  config: WalletStorageConfig
+  credentials: {
+    [key: string]: unknown
+  }
```
TimoGlastra, line 79: I think we can type this more strictly as this is for postgres only? So add the config and credential parameters strictly

TimoGlastra (commented):
Really nice improvements @sairanjitAW, left some minor nits, but should be good to go afterwards 👍

TimoGlastra (approved):
Nice, this looks go to me! 

I'll leave to @blu3beri to give the final review on the loading of the postgres library

PR #700, opened 2022-04-07
feat: use did keys for oob and did exchange protocols

jakubkoci:

### DIFF
--- a/packages/core/src/agent/EnvelopeService.ts
+++ b/packages/core/src/agent/EnvelopeService.ts
@@ -5,6 +5,7 @@ import type { AgentMessage } from './AgentMessage'
 import { inject, scoped, Lifecycle } from 'tsyringe'
 
 import { InjectionSymbols } from '../constants'
+import { didKeyToVerkey, verkeyToDidKey } from '../modules/dids/helpers'
 import { ForwardMessage } from '../modules/routing/messages'
 import { Wallet } from '../wallet/Wallet'
 
@@ -17,7 +18,7 @@ export interface EnvelopeKeys {
 }
 
 @scoped(Lifecycle.ContainerScoped)
-class EnvelopeService {
+export class EnvelopeService {
   private wallet: Wallet
   private logger: Logger
   private config: AgentConfig
@@ -29,38 +30,44 @@ class EnvelopeService {
   }
 
   public async packMessage(payload: AgentMessage, keys: EnvelopeKeys): Promise<EncryptedMessage> {
-    const { routingKeys, senderKey } = keys
-    let recipientKeys = keys.recipientKeys
+    const { recipientKeys, routingKeys, senderKey } = keys
+    let recipientVerkeys = recipientKeys.map(didKeyToVerkey)
+    const routingVerkeys = routingKeys.map(didKeyToVerkey)
+    const senderVerkey = senderKey && didKeyToVerkey(senderKey)
 
     // pass whether we want to use legacy did sov prefix
     const message = payload.toJSON({ useLegacyDidSovPrefix: this.config.useLegacyDidSovPrefix })
 
     this.logger.debug(`Pack outbound message ${message['@type']}`)
 
-    let encryptedMessage = await this.wallet.pack(message, recipientKeys, senderKey ?? undefined)
+    let encryptedMessage = await this.wallet.pack(message, recipientVerkeys, senderVerkey ?? undefined)
 
     // If the message has routing keys (mediator) pack for each mediator
-    for (const routingKey of routingKeys) {
+    for (const routingVerkey of routingVerkeys) {
       const forwardMessage = new ForwardMessage({
         // Forward to first recipient key
-        to: recipientKeys[0],
+        to: recipientVerkeys[0],
         message: encryptedMessage,
       })
-      recipientKeys = [routingKey]
+      recipientVerkeys = [routingVerkey]
       this.logger.debug('Forward message created', forwardMessage)
 
       const forwardJson = forwardMessage.toJSON({ useLegacyDidSovPrefix: this.config.useLegacyDidSovPrefix })
 
       // Forward messages are anon packed
-      encryptedMessage = await this.wallet.pack(forwardJson, [routingKey], undefined)
+      encryptedMessage = await this.wallet.pack(forwardJson, [routingVerkey], undefined)
     }
 
     return encryptedMessage
   }
 
   public async unpackMessage(encryptedMessage: EncryptedMessage): Promise<DecryptedMessageContext> {
-    return this.wallet.unpack(encryptedMessage)
+    const decryptedMessage = await this.wallet.unpack(encryptedMessage)
+    const { recipientKey, senderKey, plaintextMessage } = decryptedMessage
+    return {
+      recipientKey: recipientKey && verkeyToDidKey(recipientKey),
+      senderKey: senderKey && verkeyToDidKey(senderKey),
+      plaintextMessage,
+    }
   }
 }
-
-export { EnvelopeService }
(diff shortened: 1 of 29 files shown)


packages/core/src/agent/EnvelopeService.ts
```diff
@@ -29,38 +30,44 @@ class EnvelopeService {
   }
 
   public async packMessage(payload: AgentMessage, keys: EnvelopeKeys): Promise<EncryptedMessage> {
-    const { routingKeys, senderKey } = keys
-    let recipientKeys = keys.recipientKeys
+    const { recipientKeys, routingKeys, senderKey } = keys
+    let recipientVerkeys = recipientKeys.map(didKeyToVerkey)
```
TimoGlastra, line 34: I would rather not re-introduce the term verkey into more places of the codebase. It feels to me like something that is left-over from indy and hides what the value actually is: a base58 encoded public key (took me really long to understand a verkey is just that).

Maybe we can instead start introducing the encoding in the variable name?

So `senderPublicKeyBase58`, etc... What do you think?

packages/core/src/agent/EnvelopeService.ts
```diff
@@ -29,38 +30,44 @@ class EnvelopeService {
...
+      encryptedMessage = await this.wallet.pack(forwardJson, [routingVerkey], undefined)
     }
 
     return encryptedMessage
   }
 
   public async unpackMessage(encryptedMessage: EncryptedMessage): Promise<DecryptedMessageContext> {
-    return this.wallet.unpack(encryptedMessage)
+    const decryptedMessage = await this.wallet.unpack(encryptedMessage)
+    const { recipientKey, senderKey, plaintextMessage } = decryptedMessage
```
TimoGlastra, line 66: Not necessarily for now, but what if we instead start passing around `Key` instances instead of the string? I think that may help with all the transformations we need to do with `verkeyToDidKey` and `didKeyToVerkey`? If we just add one helper method on the didcomm service to access the actual key instances

packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -495,20 +495,20 @@ export class ConnectionService {
         type: message.type,
       })
 
+      const recipientVerkey = messageContext.recipientVerkey && didKeyToVerkey(messageContext.recipientVerkey)
```
TimoGlastra, line 498: The more I go through this code, the more I'm reluctant to this approach. Basically we now say a value of `recipientKey`, `recipientVerkey`, `routingKey` or `routingVerkey` can always be either a did:key or a base58 encoded key.

What if we start attaching an instance of `Key` as the `recipientKey` to the message context, so we don't have to do this dance everywhere? It really makes it uncertain which value to expect

TimoGlastra (commented):
This is great jakub, however have some doubts about the approach. What do you think of my comments?

packages/core/src/agent/EnvelopeService.ts
jakubkoci: Agree, although `senderPublicKeyBase58` sems terrible but it's better naming. What about just `senderKeyBase58`? There are no private keys in other places. Just idea.

TimoGlastra: Yeah sure `senderKeyBase58` sounds fine to me, it's explicit enough

packages/core/src/agent/EnvelopeService.ts
TimoGlastra: What do you think of this idea @jakubkoci? I quite like it, and as we're using `Key` in more and more places it starts to make more sense.

e.g. in this PR we update the wallet interface to use `Key` instances instead of leveraging `verkeys` : https://github.com/hyperledger/aries-framework-javascript/pull/684/files#diff-3789ea30a1c820026da4fd9d886aa730250fc2a99a8dc62bb6c3143f8b798aefR41-R56

packages/core/src/wallet/Wallet.ts
```diff
@@ -24,7 +24,7 @@ export interface Wallet {
   initPublicDid(didConfig: DidConfig): Promise<void>
   createDid(didConfig?: DidConfig): Promise<DidInfo>
   pack(payload: Record<string, unknown>, recipientKeys: string[], senderVerkey?: string): Promise<EncryptedMessage>
-  unpack(encryptedMessage: EncryptedMessage): Promise<DecryptedMessageContext>
```
TimoGlastra, line 27: We still use the term `EncryptedMessage` so if we want to rename this to `Unpacked` we should probably also call it `PackedMessage`. Any reason you want to go for pack vs encrypt? I'd be leaning to encrypted as it's a more commonly used term over packed, but fine to go with either of them, as long as we're consistent

packages/core/src/modules/routing/__tests__/mediation.test.ts
```diff
@@ -7,18 +7,21 @@ import { SubjectInboundTransport } from '../../../../../../tests/transport/Subje
...
 import { ConnectionRecord, HandshakeProtocol } from '../../connections'
 import { MediatorPickupStrategy } from '../MediatorPickupStrategy'
 import { MediationState } from '../models/MediationState'
 
-const recipientConfig = getBaseConfig('Mediation: Recipient')
+const recipientConfig = getBaseConfig('Mediation: Recipient', { logger: new ConsoleLogger(LogLevel.debug) })
 const mediatorConfig = getBaseConfig('Mediation: Mediator', {
   autoAcceptMediationRequests: true,
   endpoints: ['rxjs:mediator'],
+  logger: new ConsoleLogger(LogLevel.debug),
```
TimoGlastra, line 19: console loggers

TimoGlastra (approved):

packages/core/src/wallet/Wallet.ts
jakubkoci: Ah, you got me. I just needed to do a small naming workaround to distinguish between the following data structures:

```
export interface DecryptedMessageContext {
  plaintextMessage: PlaintextMessage
  senderKey?: Key
  recipientKey?: Key
}
```

and 

```
export interface UnpackedMessageContext {
  plaintextMessage: PlaintextMessage
  senderKey?: string
  recipientKey?: string
}
```

I hesitated a bit where to draw the line between Indy types and AFJ types. Now, I see the `Wallet` interface as part of the AFJ so its types should be also part of the AFJ. Implementors, like `IndyWallet` can use different types inside. So we can use `DecryptedMessageContext` in the `Wallet` interface.

Regarding the `unpack/pack` vs `decrypt/encrypt`. I also tended towards using decrypt/encrypt, it seems more natural to me. But if I remember correctly, there were discussions about this in Indy/Aries community before and it has been changed from encrypt to pack. I don't know the exact reason. 

One thing where I see a small benefit of pack/unpack is that the message can be encrypted many times with the use of a forward message.

jakubkoci: There can be also a problem with dependencies. If we use `DecryptedMessageContext` we should move conversion from `publicKeyBase58` to instance of `Key` inside `IndyWallet` and that means a dependency on `modules/dids`

jakubkoci: I gave it another look and maybe we can live with the dependency for now.

TimoGlastra: Looking at DIDComm v2 they're using encrypted:

Authcrypted and/or anoncrypted | application/didcomm-encrypted+json
-- | --
Signed and then anoncrypted | application/didcomm-encrypted+json
Signed | application/didcomm-signed+json
Plaintext | application/didcomm-plain+json


--- 

Re the dependency on `modules/dids` the `Key` class has been moved to the crypto directory in a separate PR (as it's not only related to dids). Would that resolve the issue?

Issue #685, opened 2022-03-29
Adding non wallet related records to the wallet (Non-secret)


MosCD3:
This is an enhancement ticket to be as a reference for my work on adding the ability to use indy non-secret APIs to add custom wallet records.

Plan 
- Create module 'NonSecret'
- Module will handle adding and retrieving custom records to the designated wallet using 'Indy.Framework -> addRecordInWallet(..)' 

Work will start soon
Suggestions are welcomed along the way

PR #702, opened 2022-04-09
feat: ability to add generic records

MosCD3:
Summary of Changes

Added a non-secret module with minimal API just to expose the on-secret APIs for Indy library [here](https://hyperledger-indy.readthedocs.io/projects/sdk/en/latest/docs/design/003-wallet-storage/README.html#) 
This will enable business logic that requires saving custom records like wallet history to be saved to the wallet and retrieved back when possible

Current Status

- Change all module names and APIs to "Generic Module" name (Done)
- Implement Typing constraints (Done)
- Implement generic get (Pending help from community .. )
- Implement modify and delete record
- Add unit tests

Related Issue #685


packages/core/src/modules/non-secret/repository/NonSecretRecord.ts
```diff
@@ -0,0 +1,47 @@
+import type { RecordTags, TagsBase } from '../../../storage/BaseRecord'
+
+import { BaseRecord } from '../../../storage/BaseRecord'
+import { uuid } from '../../../utils/uuid'
+
+export type NonSecretRecordTags = TagsBase
+export type DefaultNonSecretRecordTags = {
+  connectionId: string
```
blu3beri, line 8: I think the `connectionId` is slightly opinionated. I can imagine that a lot of use cases do not benefit from this. I do understand that it can be used for getting this record via a connection, but then why would you not use the `connectonRecord.metadata`?

packages/core/src/modules/non-secret/repository/NonSecretRecord.ts
```diff
@@ -0,0 +1,47 @@
...
+      this.content = props.content
+      this.connectionId = props.connectionId
+      this._tags = props.tags ?? {}
+    }
+  }
+
+  public getTags() {
+    return {
+      ...this._tags,
+      connectionId: this.connectionId ?? 'NONE',
```
blu3beri, line 44: I don't think we should use a string `NONE` here. NULL would be better.

packages/core/src/modules/non-secret/service/NonSecretRecordService.ts
```diff
@@ -0,0 +1,40 @@
...
+
+@scoped(Lifecycle.ContainerScoped)
+export class NonSecretRecordService {
+  private nonSecretRepository: NonSecretRecordRepository
+
+  public constructor(nonSecretRepository: NonSecretRecordRepository) {
+    this.nonSecretRepository = nonSecretRepository
+  }
+
+  public async saveRecord(message: string, tags?: NonSecretRecordTags, connectionRecord?: ConnectionRecord) {
```
blu3beri, line 18: It would be better for the api to take an object so that we have named parameters.

Somehting like this: 

```suggestion
public async saveRecord({message, tags, connectionRecord}: SaveNonSecretRecordOption) {
```

packages/core/src/modules/non-secret/service/NonSecretRecordService.ts
```diff
@@ -0,0 +1,40 @@
...
+    const nonsecretRecord = new NonSecretRecord({
+      content: message,
+      connectionId: connectionRecord?.id,
+      tags: tags,
```
blu3beri, line 22: `tags: tags` can be `tags`

packages/core/src/modules/non-secret/service/NonSecretRecordService.ts
```diff
@@ -0,0 +1,40 @@
...
+    })
+
+    try {
+      await this.nonSecretRepository.save(nonsecretRecord)
+      return nonsecretRecord
+    } catch (error) {
+      throw new AriesFrameworkError(`${error}`)
```
blu3beri, line 29: This error should get some context

```suggestion
      throw new AriesFrameworkError(`Unable to store the nonSecret record with id ${nonSecretRecord.id}. Message: ${error}`)
```

packages/core/src/modules/non-secret/service/NonSecretRecordService.ts
```diff
@@ -0,0 +1,40 @@
...
+@scoped(Lifecycle.ContainerScoped)
+export class NonSecretRecordService {
+  private nonSecretRepository: NonSecretRecordRepository
+
+  public constructor(nonSecretRepository: NonSecretRecordRepository) {
+    this.nonSecretRepository = nonSecretRepository
+  }
+
+  public async saveRecord(message: string, tags?: NonSecretRecordTags, connectionRecord?: ConnectionRecord) {
+    const nonsecretRecord = new NonSecretRecord({
```
blu3beri, line 19:
```suggestion
    const nonSecretRecord = new NonSecretRecord({
```

packages/core/src/modules/non-secret/service/NonSecretRecordService.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    try {
+      await this.nonSecretRepository.save(nonsecretRecord)
+      return nonsecretRecord
+    } catch (error) {
+      throw new AriesFrameworkError(`${error}`)
+    }
+  }
+
+  public async findAllByQuery(query: Partial<NonSecretRecordTags>) {
```
blu3beri, line 33: I think that we should keep this as `findByQuery` to keep the same name or we should rename all `findByQuery` to `findAllByQuery` I feel like `findAllByQuery` is a better name, as it is more explicit and `Array.find()` returns a single and not multiple so it can be confusing. Curious to here some opinions about this.

packages/core/src/modules/non-secret/service/NonSecretRecordService.ts
blu3beri, line 18: I think for the public api `saveRecord` and the getters should accept a `Record<string, unknown>`. This would allow AFJ to handle the ser/de side and would make it a bit easier for the user as I can imagine that JSON is the preferred storage structure. 

If we want to make the api even nicer, imo, we should add a generic that extends `Record<string, unknown>`. Like so (half pseudo code):

```ts
public async get<R extends Record<string, unknown>(id: string): Promise<NonSecretRecord<R>> {
  const record = /* query */
  record.content = JsonEncoder.fromString(record.content) as R
  return record
}

```

blu3beri: I also noticed that a getById, or something similiar is missing. Was this intentional?

packages/core/src/modules/non-secret/repository/NonSecretRecord.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export type BasicMessageTags = RecordTags<NonSecretRecord>
+
+export interface NonSecretRecordStorageProps {
+  id?: string
+  createdAt?: Date
+  connectionId?: string
+  tags?: NonSecretRecordTags
+
+  content: string
```
blu3beri, line 19: See the comment at the `NonSecretStorageService`. 

The interface `NonSecretStorageProps` would no take a generic as type of content. 

I am not too sure if this is the best option and would be happy to hear any arguments against it or alternatives.

blu3beri (changes requested):
I like the addition and can be useful in many cases. I have some minor points that I feel like would make it a bit more generic.

Also if you want to storage some data alongside a `connectionRecord`, which is why I think you added the `connectionId`, you can also use `connectionRecord.metadata` to add some quick JSON to a record.

TimoGlastra (approved):
Thanks for the PR @MosCD3. 

I'm not sure if we should call this non secrets. Indy called it like this, but I would rather think of a better name. People not familiar with the indy ecyosystem have no idea what non secrets means and as we're moving to Aries Askar it'll make less sense. A connection record is as much a non secrets record than the records created with this new API. (I'm still thinking of a suitable name, will get back if I can think of something. Please let me know if you have any suggestions)

also in addition to the e2e tests, could you add some unit tests? This has bitten us in the past, so would like make sure new additions are well-covered.

as a last note, can you look to remove the changes to the `yarn.lock` file? I think they're unrelated to the additions of this PR.

TimoGlastra (changes requested):
.

packages/core/src/modules/non-secret/repository/NonSecretRecord.ts
genaris: When I first saw this PR, I thought that I could use to store some wallet-specific preferences, which of course are not tied to any connection, so I agree that leveraging the use of metadata would make this module more generic.

packages/core/src/modules/non-secret/repository/NonSecretRecord.ts
```diff
@@ -0,0 +1,47 @@
...
+
+import { BaseRecord } from '../../../storage/BaseRecord'
+import { uuid } from '../../../utils/uuid'
+
+export type NonSecretRecordTags = TagsBase
+export type DefaultNonSecretRecordTags = {
+  connectionId: string
+}
+
+export type BasicMessageTags = RecordTags<NonSecretRecord>
```
genaris, line 11: Maybe this comes from Basic Message module?

packages/core/src/modules/non-secret/service/NonSecretRecordService.ts
genaris: > I think for the public api `saveRecord` and the getters should accept a `Record<string, unknown>`. This would allow AFJ to handle the ser/de side and would make it a bit easier for the user as I can imagine that JSON is the preferred storage structure.
> 
> If we want to make the api even nicer, imo, we should add a generic that extends `Record<string, unknown>`. Like so (half pseudo code):
> 
> ```ts
> public async get<R extends Record<string, unknown>(id: string): Promise<NonSecretRecord<R>> {
>   const record = /* query */
>   record.content = JsonEncoder.fromString(record.content) as R
>   return record
> }
> ```

That would be very useful to make this module even more generic!

packages/core/src/modules/non-secret/service/NonSecretRecordService.ts
genaris: I had this issue before! I think `findAllByQuery` is more clear. Also I had some confusions regarding the usage of 'find' and 'get' (e.g. get(record) will throw and exception if not found while getAll() returns empty array if no record found. Why not using findAll() in that case?)

packages/core/src/modules/non-secret/service/NonSecretRecordService.ts
genaris, line 18: I really like that we can assign custom tags to each record, so they can be efficiently searchable!

However I think it would be nice if there is an API at module/service level to **update** tags and contents for existing records. And also to **delete** them.

genaris (commented):
This will be a great addition to the framework, as it will give the opportunity to store in the wallet those things that do not fit in other record types!

It's true that the naming could be too 'Indy-centric'. At first sigh I would call it 'Generic Records Module', but probably it's too long.

By the way, following conventions from other modules I think the naming of repository and service should be `NonSecretRepository` and `NonSecretService` respectively (unless we use 'Record' suffix as part of module name).

packages/core/src/modules/non-secret/service/NonSecretRecordService.ts
MosCD3: this function returns Array even if it is a single item so it should be findAll but any way the final word is for the maintainers

packages/core/src/modules/non-secret/repository/NonSecretRecord.ts
MosCD3: While I know that you can manually add the connectionId to the tg but remember most of the business relies on adding custom history messages that depends on interaction with some sort of a connection with other agent, so why not automate this and just have the connectionID as part of the tags after all, 
The other reason is when removing the connectionId, if found that the getTags function fails to comply with base class so its kind of glued to the framework.
For metadata I do not recommend shuffling much of metadata into the tags and let that be on the controller side to pickup from. I would rather make as simple as the id of the connection that had this data

packages/core/src/modules/non-secret/service/NonSecretRecordService.ts
MosCD3: Good point, can we do that enhancement as next level? I not quit familiar with that and it seems causing lots of errors when adding it

packages/core/src/modules/non-secret/repository/NonSecretRecord.ts
TimoGlastra: We've had multiple use cases where external data is not tied to a specific connection. For now we've always used the record metadata, but for larger objects I can understand you would use an external record. For this reason I think it is important to keep this general and not tied to any connection. Implementations that want to tie this to a connection can do some by storing the connectionId in the record payload.

packages/core/src/modules/non-secret/repository/NonSecretRecord.ts
TimoGlastra: Agreed. It should be an optional tag

packages/core/src/modules/non-secret/service/NonSecretRecordService.ts
TimoGlastra: Happy to change `getAll` to `findAll` for consistency, but probably better addressed in a separate PR. 

So we would use the following terms:

`findById` / `getById`
`findSingleByQuery` / `getSingleByQuery`
`findAll`
`findAllByQuery`

Anyone volunteering to address this in a separate PR? :)

packages/core/src/modules/non-secret/service/NonSecretRecordService.ts
TimoGlastra: Agreed! We should probably add some methods to the public api to update and delete records

TimoGlastra:
> It's true that the naming could be too 'Indy-centric'. At first sigh I would call it 'Generic Records Module', but probably it's too long.

I like the direction of this. So the module will used like `genericRecords`. Don't think that's too long.

So it would be something like

```ts
agent.genericRecords.saveRecord()

// another thought?
agent.genericStorage.saveRecord()
```

genaris:
> @TimoGlastra that's a good idea 'genericRecords'. I will change to that

Nice! So in that case `saveRecord` method would become just `save` to avoid verbosity?

packages/core/src/modules/non-secret/service/NonSecretRecordService.ts
MosCD3: I can't find any similar implementation in any of the current modules using this Generic way. 
@blu3beri  Can you suggest please the change to make to "GenericRecord" so that it can be wrapped in the function you suggested?

MosCD3:
@TimoGlastra @blu3beri Any idea about the dummy module that've been mixed with our changes? should we create a new clean PR?

TimoGlastra:
Not sure what happened here. Can you try to merge with main?

packages/core/src/modules/generic-record/GenericRecordModule.ts
```diff
@@ -0,0 +1,77 @@
...
+@scoped(Lifecycle.ContainerScoped)
+export class GenericRecordModule {
+  private genericRecordsService: GenericRecordService
+  private logger: Logger
+  public constructor(agentConfig: AgentConfig, genericRecordsService: GenericRecordService) {
+    this.genericRecordsService = genericRecordsService
+    this.logger = agentConfig.logger
+  }
+
+  public async save({ message, tags }: SaveGenericRecordOption) {
```
TimoGlastra, line 23: I see it's called content in the service, would be good to keep it consistent

packages/core/src/modules/generic-record/GenericRecordModule.ts
```diff
@@ -0,0 +1,77 @@
+import type { Logger } from '../../logger'
```
TimoGlastra, line 1: Maybe we can call the module generic-records? All other modules are plural

MosCD3: Done

packages/core/src/modules/generic-record/GenericRecordModule.ts
MosCD3: Done

packages/core/src/modules/generic-records/repository/GenericRecord.ts
```diff
@@ -0,0 +1,45 @@
...
+export interface SaveGenericRecordOption {
+  content: Record<string, unknown>
+  id?: string
+  tags?: GenericRecordTags
+}
+
+export class GenericRecord extends BaseRecord<GenericRecordTags> {
+  public content!: Record<string, unknown>
+
+  public static readonly type = 'NonSecretRecord'
```
TimoGlastra, line 26:
```suggestion
  public static readonly type = 'GenericRecord'
```

packages/core/tests/generic-records.test.ts
```diff
@@ -0,0 +1,111 @@
...
+import { getBaseConfig } from './helpers'
+
+const aliceConfig = getBaseConfig('Agents Alice', {
+  endpoints: ['rxjs:alice'],
+})
+
+describe('genericRecords', () => {
+  let aliceAgent: Agent
+
+  // const fooString: FooString = { foo: 'Some data saved' }
```
TimoGlastra, line 14:
```suggestion
```

TimoGlastra (changes requested):
Can you add back in the yarn.lock file?

packages/core/src/modules/generic-records/GenericRecordsModule.ts
```diff
@@ -0,0 +1,77 @@
...
+    }
+  }
+
+  public async delete(record: GenericRecord): Promise<void> {
+    try {
+      await this.genericRecordsService.delete(record)
+    } catch (error) {
+      this.logger.error('Error while saving generic-record', {
+        error,
+        encryptedMessage: record.content,
```
TimoGlastra, line 46:
```suggestion
        content: record.content,
```

packages/core/src/modules/generic-records/GenericRecordsModule.ts
```diff
@@ -0,0 +1,77 @@
...
+    try {
+      const record = await this.genericRecordsService.save({
+        content: content,
+        tags: tags,
+      })
+      return record
+    } catch (error) {
+      this.logger.error('Error while saving generic-record', {
+        error,
+        encryptedMessage: content,
```
TimoGlastra, line 33:
```suggestion
        content,
```

packages/core/src/modules/generic-records/GenericRecordsModule.ts
```diff
@@ -0,0 +1,77 @@
...
+    }
+  }
+
+  public async update(record: GenericRecord): Promise<void> {
+    try {
+      await this.genericRecordsService.update(record)
+    } catch (error) {
+      this.logger.error('Error while update generic-record', {
+        error,
+        encryptedMessage: record.content,
```
TimoGlastra, line 59:
```suggestion
        content: record.content,
```

TimoGlastra (approved):

blu3beri (approved):
Looks good! I think there are some very minor code cleanup things, but I would be happy to pick that up myself Later on.

blu3beri:
closes #685

PR #704, opened 2022-04-13
fix: disallow floating promises

TimoGlastra:
Helps catch nasty bugs (and even caught some places where we didn't call await!), and as #563 has been around forever it's easier to already pick some important rules.

### DIFF
--- a/packages/core/src/agent/Agent.ts
+++ b/packages/core/src/agent/Agent.ts
@@ -175,11 +175,11 @@ export class Agent {
     }
 
     for (const transport of this.inboundTransports) {
-      transport.start(this)
+      await transport.start(this)
     }
 
     for (const transport of this.outboundTransports) {
-      transport.start(this)
+      await transport.start(this)
     }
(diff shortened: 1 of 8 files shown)


blu3beri (approved):

packages/core/src/agent/Agent.ts
```diff
@@ -201,10 +201,10 @@ export class Agent {
 
     // Stop transports
     for (const transport of this.outboundTransports) {
-      transport.stop()
+      await transport.stop()
     }
     for (const transport of this.inboundTransports) {
-      transport.stop()
+      await transport.stop()
     }
```
jakubkoci, line 208: Wouldn't be better to do something like this `await Promises.all(this.inboundTransports.map(t => t.stop()))`? I think we just need to wait until all transports finish

TimoGlastra: Yes! I'll address in a seperate PR

PR #705, opened 2022-04-13
refactor: start and stop transports in parallel

TimoGlastra:
start and stop transports in parallel as suggested by @jakubkoci  in https://github.com/hyperledger/aries-framework-javascript/pull/704#discussion_r849258130

I've grouped the inbound / outbound so we they're all running in parallel

### DIFF
--- a/packages/core/src/agent/Agent.ts
+++ b/packages/core/src/agent/Agent.ts
@@ -174,13 +174,10 @@ export class Agent {
       })
     }
 
-    for (const transport of this.inboundTransports) {
-      await transport.start(this)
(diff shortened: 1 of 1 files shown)


blu3beri (approved):
This does add that if one rejects they all reject right? Is that wanted?

TimoGlastra:
currently if one fails the agent startup will fail, so I think it should be fine. But as you say, we may want to make this a bit more sophisticated still stopping other transports if one fails. Let's address in a separate PR

jakubkoci (approved):

PR #714, opened 2022-04-20
feat: support handling messages with different minor version

TimoGlastra:
Adds support for processing messages with a different minor version.

Some notes about the changes:
- The static `type` property on the message classes are now `ParsedMessageType` containing info about the message type. I've done this to make the info readily available and not have to parse it every time we need this info (would need to be done for every message for each message received). This is how declaring the type now looks. `praseMessageType` is an utitliy method that makes it easier to created the `ParsedMessageType` object.
```ts
class CustomProtocolMessage extends AgentMessage {
  @IsValidMessageType(CustomProtocolMessage.type)
  public readonly type = CustomProtocolMessage.type.messageTypeUri
  public static readonly type = parseMessageType('https://didcomm.org/fake-protocol/1.5/message')
}
```
- Had to refactor the decorators a bit as it was incorrectly typed which was now causing errors with typescript
- Added an e2e test where we send a message with a lower minor version to another agent with a higher minor version, and also a test where we send a message with a higher minor version to another agent with a lower minor version.
- Added transformations from string values `1` and `0` to `n__1` and `n__0`. See here for the relevant discussion: https://github.com/hyperledger/aries-framework-javascript/pull/690#discussion_r850942240. Other approaches very welcome.

~~This PR doesn't include handling the error case where we receive a message with a higher minor version. I'll like to address that in a separate PR (should be straightforward, we just have to change the reason in the problem report), but this at least unblocks us now.~~ -> Higher versions are now allowed, still need to send the problem report though in a future pr


swcurran:
Per discussion today on the Aries WG Call, and the RRC 0003 Protocols --- handling a higher minor version is not an error, and an agent should respond with the highest version that the agent does support.  So if you only support OOB 1.0, then you should happily receive an OOB 1.1 and respond with an OOB 1.0 message.  Same for present-proof 2.1 / 2.0.

TimoGlastra:
Interesting, I only allowed to receive a lower version. I'll update it to also support higher versions in that case.

packages/core/src/agent/Dispatcher.ts
```diff
@@ -91,17 +92,21 @@ class Dispatcher {
...
   private getHandlerForType(messageType: string): Handler | undefined {
+    const incomingMessageType = parseMessageType(messageType)
+
     for (const handler of this.handlers) {
       for (const MessageClass of handler.supportedMessages) {
-        if (MessageClass.type === messageType) return handler
+        if (canHandleMessageType(MessageClass, incomingMessageType)) return handler
       }
     }
   }
```
blu3beri, line 102: I think this might be a big cleaner


```ts
return this.handlers.forEach((handler) => handler.supportedMessages.find((message) => message.type === messageType))
```

packages/core/src/agent/Dispatcher.ts
```diff
@@ -91,17 +92,21 @@ class Dispatcher {
...
   public getMessageClassForType(messageType: string): typeof AgentMessage | undefined {
+    const incomingMessageType = parseMessageType(messageType)
+
     for (const handler of this.handlers) {
       for (const MessageClass of handler.supportedMessages) {
-        if (MessageClass.type === messageType) return MessageClass
+        if (canHandleMessageType(MessageClass, incomingMessageType)) return MessageClass
       }
     }
   }
```
blu3beri, line 112: Same here, could be a bit cleaner.

packages/core/src/modules/connections/messages/ConnectionInvitationMessage.ts
```diff
@@ -61,12 +61,12 @@ export class ConnectionInvitationMessage extends AgentMessage {
...
 
-  @Equals(ConnectionInvitationMessage.type)
+  @IsValidMessageType(ConnectionInvitationMessage.type)
   @Transform(({ value }) => replaceLegacyDidSovPrefix(value), {
     toClassOnly: true,
   })
-  public readonly type = ConnectionInvitationMessage.type
-  public static readonly type = 'https://didcomm.org/connections/1.0/invitation'
+  public readonly type = ConnectionInvitationMessage.type.messageTypeUri
+  public static readonly type = parseMessageType('https://didcomm.org/connections/1.0/invitation')
```
blu3beri, line 69: Not really part of your PR, but more just curious, why do we have a `public static readonly type` and `public readonly type`?

blu3beri (approved):
Again some minor nits. Tests look good :)

packages/core/src/agent/Dispatcher.ts
TimoGlastra: This won't work because the forEach doesn't return anything. I could change it to find, but then it would return the handler and not the message class. 

Any other suggestions on how to approach?

packages/core/src/agent/Dispatcher.ts
TimoGlastra: Same as above

packages/core/src/modules/connections/messages/ConnectionInvitationMessage.ts
TimoGlastra: Because we sometimes need the static property and sometimes the instance property. When transforming from/to a json object we need to set the `@type` property on the class.

However when we need to find the correct message class for an incoming message we don't have an instance yet, so we use the static  property. It's not ideal, but I'm not sure of other approaches. Any ideas? Would love to remove the need for both.

packages/core/src/storage/IndyStorageService.ts
```diff
@@ -42,9 +42,16 @@ export class IndyStorageService<T extends BaseRecord> implements StorageService<
...
+      }
+      // Transform '1' and '0' to boolean
+      else if (value === '1' || value === '0') {
         transformedTags[key] = value === '1'
       }
+      // If 1 or 0 is prefixed with 'n__' we need to remove it. This is to prevent
+      // casting the value to a boolean
+      else if (value === 'n__1' || value === 'n__0') {
+        transformedTags[key] = value === 'n__1' ? '1' : '0'
+      }
```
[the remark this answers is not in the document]
TimoGlastra: Hmm not so sure on the benefit of this. As you say it will be less safe, with no clear advantages.

I'm going to leave it like this for now, if you feel strongly about this point, feel free to convince me

TimoGlastra:
I'll wait with merging this PR until #649 has been merged to not cause more merge conflicts for @NB-MikeRichardson 😄

packages/core/src/agent/Dispatcher.ts
blu3beri: I believe that swapping the `forEach` with a find would work.

packages/core/src/modules/connections/messages/ConnectionInvitationMessage.ts
blu3beri: I don't think I have another solution for this right now.

packages/core/src/storage/IndyStorageService.ts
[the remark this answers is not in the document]
blu3beri: I do not feel strongly about this. Just not a big fan of these switch cases. It could still be a ternary expressions, but I don't think that that will improve readability.

packages/core/src/agent/Dispatcher.ts
blu3beri: return this.handlers().flatMap((handler) => handler.supportedMessages).find((MessageClass) => canHandleMessageType(MessageClass, incomingMessageType))

blu3beri: If `flatMap` is not supported:

```ts
return Array.prototype.concat(...this.handlers.map((handler) => handler.supportedMessages).find((MessageClass) => canHandleMessageType(MessageClass, incomingMessageType))
```

PR #716, opened 2022-04-22
feat: delete events

Przytua:
Added delete events to `Connection`, `Credential` and `Proof` services so that they notify about deletion in similar way to state change events.

### DIFF
--- a/packages/core/src/modules/connections/ConnectionEvents.ts
+++ b/packages/core/src/modules/connections/ConnectionEvents.ts
@@ -4,6 +4,7 @@ import type { ConnectionRecord } from './repository/ConnectionRecord'
 
 export enum ConnectionEventTypes {
   ConnectionStateChanged = 'ConnectionStateChanged',
+  ConnectionDeleted = 'ConnectionDeleted',
 }
 
 export interface ConnectionStateChangedEvent extends BaseEvent {
@@ -11,5 +12,11 @@ export interface ConnectionStateChangedEvent extends BaseEvent {
   payload: {
     connectionRecord: ConnectionRecord
     previousState: ConnectionState | null
+  }
+}
+export interface ConnectionDeletedEvent extends BaseEvent {
+  type: typeof ConnectionEventTypes.ConnectionDeleted
+  payload: {
+    connectionRecord: ConnectionRecord
   }
 }
--- a/packages/core/src/modules/connections/__tests__/ConnectionService.test.ts
+++ b/packages/core/src/modules/connections/__tests__/ConnectionService.test.ts
@@ -1,4 +1,5 @@
 import type { Wallet } from '../../../wallet/Wallet'
+import type { ConnectionDeletedEvent } from '../ConnectionEvents'
 import type { Routing } from '../services/ConnectionService'
 
 import { getAgentConfig, getMockConnection, mockFunction } from '../../../../tests/helpers'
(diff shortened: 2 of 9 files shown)


TimoGlastra (changes requested):
Thanks, this is a great contribution! 

One thing I'm thinking about is whether we could maybe solve this generically by adding events to the repository level. This way we can have `RecordDeletedEvent`, `RecordSavedEvent`, `RecordUpdatedEvent`, etc... What do you think of this?

I understand this is more work to implement and is maybe something for the future. So I'm fine with merging this for now and replacing it with the more generic api later on. 

One thing I'd like to see is if you could add some tests to see if deleting a record actually calls the deleted event?

Przytua:
@TimoGlastra I've added tests.

I could add the generic Repository events (in a separate PR), but am wondering if this would be very useful. Currently you can listen to separate events for different records, which is very convenient, as you don't need to filter in the callback by record type, to handle event properly. Anyway if you feel this would be useful, I can add it.

TimoGlastra (approved):
Thanks for adding the tests @Przytua! This looks good to me.

TimoGlastra:
> I could add the generic Repository events (in a separate PR), but am wondering if this would be very useful. Currently you can listen to separate events for different records, which is very convenient, as you don't need to filter in the callback by record type, to handle event properly. Anyway if you feel this would be useful, I can add it.

I agree, that makes working with the event less convenient. To me it still seems useful as you can filter in the event handler, and it will be easier for us to stay consistent across all crud events. This also allows to add generic processing when any record is deleted.

We  could also add some default filter methods to help with the event handler filters

```ts
// We could add some util operators that help with the filtering of the desired record type
agent.events.observables(RepositoryEventTypes.RecordDeleted).pipe(
    filter(e => e.payload.record.type === ConnectionRecord.type)
).subscribe(e => {

})
```

Przytua:
I've added the #723 issue for generic events. I'm unable to assign myself to it, maybe you can do this?

What is the process now with this one? Someone else needs to merge it?

TimoGlastra:
> I've added the #723 issue for generic events. I'm unable to assign myself to it, maybe you can do this?

I've added it to the agenda for the WG call today to see if this is a desired feature. Maybe others also share your opinion. If we agree this is desired, I'll make sure to assign you!

> What is the process now with this one? Someone else needs to merge it?

Let's wait a bit to see if someone else reviews, and otherwise I'll merge it in the next few days

Przytua:
Great, thank you!

Przytua:
@TimoGlastra I've also created a PR in `react-hooks` https://github.com/hyperledger/aries-framework-javascript-ext/pull/113 that uses these delete events. I noticed that react-hooks use `v0.1.0` of core. Maybe there would be some way to also add these changes in a fix to `v0.1.0` so that it could be used in `react-hooks` before `0.2.0` is mature enough to be released? (sorry for abusing your help)

TimoGlastra:
Yes that would be a good idea, but I'm not 100% on how to do this at the moment. It would require us to create a v1 branch I think and only merge fixes that should cut the 0.1.0 release. I would have hoped we would reach the 0.2.0 release sooner to avoid this problem.

amanji (approved):

amanji:
Can this be merged? If conflicts need to be resolved I can help with that.

TimoGlastra:
@[Przytua](https://github.com/Przytua) Any update on the more generic repository events?

Przytua:
> @Przytua Any update on the more generic repository events?

@TimoGlastra sorry it took so long (had a busy time, and I forgot about it). I've created a PR with generic repository events: https://github.com/hyperledger/aries-framework-javascript/pull/842

PR #720, opened 2022-04-23
feat: Updated initPublicDid method

dinkar-jain:
This PR reuses the public did instead of generating a new every time the wallet is opened.

Signed-off-by: Dinkar Jain <62498436+dinkar-jain@users.noreply.github.com>

### DIFF
--- a/packages/core/src/wallet/IndyWallet.ts
+++ b/packages/core/src/wallet/IndyWallet.ts
@@ -354,7 +354,17 @@ export class IndyWallet implements Wallet {
   }
 
   public async initPublicDid(didConfig: DidConfig) {
+    (await this.indy.listMyDidsWithMeta(this.handle)).forEach(element => {
+      if (element.metadata == "Public") {
+        this.publicDidInfo = {
+          element.did,
+          element.verkey,
+        }
+        return
+      }
+    });
     const { did, verkey } = await this.createDid(didConfig)
+    this.indy.setDidMetadata(this.handle, did, "Public")
     this.publicDidInfo = {
       did,
       verkey,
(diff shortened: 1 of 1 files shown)


TimoGlastra (commented):
Hi @dinkar-jain, thanks for this contribution! 

Were you running into a specific issue with the method being called multiple times? We're currently working on an overhaul of how public dids work in AFJ by making it a lot more general and removing the default public did. This would also mean this method would be removed. 

Is this just a performance optimization, or were you actually running into errors that the did already existed?

dinkar-jain:
Hello @TimoGlastra 

I was actually trying to keep the Public DID's limited since the role of the public did is of a global identifier. For e.g., a public did of an Issuer of a credential used by any Verifier for the purpose of trust. Issuing multiple global identifiers will not be of any good since generally public DIDs will be also required to be issued on the ledger too only increasing the workload.

TimoGlastra:
Your reasoning makes sense to me.  Although maybe not everyone will need multiple dids, it's not something we should restrict on the framework level I think (maybe there's different areas where different dids are used). 

The current approach you took will ignore the `seed` parameter, which is not desired I think.

For now, would it be possible to use the optional `seed` parameter to deterministically generate the did?

dinkar-jain:
@TimoGlastra Thanks for your insights. 
Sure, it is possible to use the optional seed parameter to deterministically generate the did but i don't think it will be an optimal solution.
You are right, we should not restrict it on framework level. But shouldn't there be a method too for generating only a single Public DID, a single global identifier so as to increase the scope and applicability of the framework too??

PR #736, opened 2022-05-02
fix: mediation record checks for pickup v2

TimoGlastra:
Adds checks in the mediation recipient service for mediation record state and role. Also splits up the sending of the status request message to be more in line with separation of service/module in the rest of the framework. 

Finally, this updates the processing of the message to be using the event emitter. We also use this approach for pickup v1 and it was causing issues in the oob branch due to circular dependencies.


TimoGlastra:
@KolbyRKunz could you give this PR a look and see if the changes look good to you? Thanks!

KolbyRKunz (approved):
This all looks good to me, thank you for catching the `mediatorrecord.connectionId` issue in `sendStatusRequest`

TheTreek (approved):
This looks good.

blu3beri (approved):
Looks good :)

PR #757, opened 2022-05-10
fix(node): allow to import node package without postgres

TimoGlastra:

### DIFF
--- a/packages/core/tests/postgres.test.ts
+++ b/packages/core/tests/postgres.test.ts
@@ -55,7 +55,7 @@ describe('postgres agents', () => {
     }
 
     // loading the postgres wallet plugin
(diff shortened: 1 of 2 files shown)


TimoGlastra:
@blu3beri could calling `getLibrary()` multiple lead to issues? It was called once by default, but this was causing issues if you don't have the postgres binary

TimoGlastra:
FYI @sairanjitAW

blu3beri:
> @blu3beri could calling `getLibrary()` multiple lead to issues? 

I don't believe so. However it will create the bindings once again (there are only 2), which might cause some perf drops. I don't think we have to restrict it, but we probably should if we use a bigger FFI library.

PR #766, opened 2022-05-11
Fix delete credentials

NB-MikeRichardson:

### DIFF
--- a/packages/core/src/modules/credentials/protocol/v2/__tests__/v2credentials.propose-offer.test.ts
+++ b/packages/core/src/modules/credentials/protocol/v2/__tests__/v2credentials.propose-offer.test.ts
@@ -12,10 +12,11 @@ import type {
 } from '../../../CredentialsModuleOptions'
 import type { CredPropose } from '../../../formats/models/CredPropose'
 
-import { AriesFrameworkError } from '../../../../../../src/error/AriesFrameworkError'
-import { DidCommMessageRepository } from '../../../../../../src/storage'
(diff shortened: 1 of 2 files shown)


packages/core/src/modules/credentials/protocol/v2/__tests__/v2credentials.propose-offer.test.ts
```diff
@@ -381,6 +382,140 @@ describe('credentials', () => {
     }
   })
 
+  test('Faber Issues Credential which is the deleted from Alice`s wallet', async () => {
```
TimoGlastra, line 385: I think this test can be simplified a lot by storing the credential record without actually issuing the credential:

In addition it would be good to add these tests for both v1 and v2 of the protocol (wouldn't it be easier to add an unit test to the v1 and v2 credential service?)

```ts
const credentialRecord = getMockCredential({
  protocolVersion: CredentialProtocolVersion.V1
  credentials: [
   type: TheIndyType,
   credentialId: 'an-id'
]
})

await credentialRepository.save(credentialRecord)

  const deleteCredentialSpy = jest.spyOn(holderService, 'deleteCredential')
  deleteCredentialSpy.mockResolvedValue()
  await agent.credentials.deleteById(credentialRecord.id, { deleteAssociatedCredentials: true })
  expect(deleteCredentialSpy).toHaveBeenNCalledWith(1, 'an-id')
```

Keeping tests light and simple means they're easier to maintain and update

TimoGlastra (approved):

PR #767, opened 2022-05-11
refactor: delete credentials by default when deleting exchange

NB-MikeRichardson:

BREAKING CHANGE: the credentials associated with a credential exchange record are now deleted by default when deleting a credential exchange record. If you only want to delete the credential exchange record and not the associated credentials, you can pass the `deleteAssociatedCredentials` to the `deleteById` method:

```ts
await agent.credentials.deleteById('credentialExchangeId', { 
	deleteAssociatedCredentials: false 
})
```


packages/core/src/modules/credentials/services/CredentialService.ts
```diff
@@ -247,16 +247,25 @@ export abstract class CredentialService {
...
+    const credentialRecord = await this.getById(credentialRecordId)
 
     await this.credentialRepository.delete(credentialRecord)
 
-    if (options?.deleteAssociatedCredentials) {
-      for (const credential of credentialRecord.credentials) {
-        const formatService: CredentialFormatService = this.getFormatService(credential.credentialRecordType)
-        await formatService.deleteCredentialById(credentialRecord, options)
-      }
+    const deleteOptions = options ? options : { deleteAssociatedCredentials: true }
```
TimoGlastra, line 264:
```suggestion
    const deleteOptions = options ?? { deleteAssociatedCredentials: true }
```

packages/core/src/modules/credentials/services/CredentialService.ts
```diff
@@ -247,16 +247,25 @@ export abstract class CredentialService {
...
-    const credentialRecord = await this.getById(credentialId)
+  /**
+   * Delete a credential exchange record from the repository and (optionally) 
+   * not bound to any connection. The offer must be delivered out-of-band to the holder
+   * Note that the default is now to delete from the wallet
+   * 
+   * @param credentialRecordId The id of the credential record
+   * @param options The credential options to use for the delete
+
+   */
```
TimoGlastra, line 258: This comment seems outdated

packages/core/src/modules/credentials/services/CredentialService.ts
```diff
@@ -247,16 +247,25 @@ export abstract class CredentialService {
...
+    const credentialRecord = await this.getById(credentialRecordId)
 
     await this.credentialRepository.delete(credentialRecord)
 
-    if (options?.deleteAssociatedCredentials) {
-      for (const credential of credentialRecord.credentials) {
-        const formatService: CredentialFormatService = this.getFormatService(credential.credentialRecordType)
-        await formatService.deleteCredentialById(credentialRecord, options)
-      }
+    const deleteOptions = options ?? { deleteAssociatedCredentials: true }
```
TimoGlastra, line 264: After thinking about this a bit more, this will break when more config options are added. We should probably do something like this:

``` 
  public async deleteById(credentialRecordId: string, { deleteAssociatedCredentials = true }: DeleteCredentialOptions = {}): Promise<void> {
    const credentialRecord = await this.getById(credentialRecordId)
```

Because currently if you call it like this (with an empty object):

```ts
agent.credentials.deleteById('id', {})
```

the value won't be set to true because options exist and the associated credential won't be deleted

packages/core/src/modules/credentials/services/CredentialService.ts
```diff
@@ -247,16 +247,29 @@ export abstract class CredentialService {
...
-      }
+    let deleteOptions = options
+
+    if (!deleteOptions || !deleteOptions.deleteAssociatedCredentials) {
+      deleteOptions = { ...options, deleteAssociatedCredentials: true }
+    }
+
+    for (const credential of credentialRecord.credentials) {
+      const formatService: CredentialFormatService = this.getFormatService(credential.credentialRecordType)
+      await formatService.deleteCredentialById(credentialRecord, deleteOptions)
```
TimoGlastra, line 272: Sorry -- I should have taken better notice the first review. But why do we need to provide the deleteAssociatedCredential parameter to the formatService? The format service should only be called if the associated credential needs to be deleted.

```
const deleteAssociatedCredentials = options?.deleteAssociatedCredentials ?? true

if (deleteAssociatedCredentials) {
    for (const credential of credentialRecord.credentials) {
      const formatService = this.getFormatService(credential.credentialRecordType)
     // this takes the `credential.credentialId`, not the credentialRecord!
      await formatService.deleteCredentialById(credential.credentialId)
  }
}
```

packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts
```diff
@@ -529,17 +529,14 @@ export class IndyCredentialFormatService extends CredentialFormatService {
...
       return binding.credentialRecordType == CredentialFormatType.Indy
     })
     if (indyCredential.length != 1) {
       throw new AriesFrameworkError(`Could not find Indy record id for credential record ${credentialRecord.id}`)
     }
-    if (options?.deleteAssociatedCredentials && indyCredential[0].credentialRecordId) {
+    if (indyCredential[0].credentialRecordId) {
       await this.indyHolderService.deleteCredential(indyCredential[0].credentialRecordId)
     }
   }
```
TimoGlastra, line 542: I see you didn't add this in this PR, but I think we can make this a lot simpler by just passing the actual credentialRecordId  form the service. We already loop through the credentials there and find the correct format service (see my last suggestion on how to call this method from there, only the id is passed.). The body of this function will then be as simple as:

```ts
  public async deleteCredentialById(credentialRecordId: string): Promise<void> {
    await this.indyHolderService.deleteCredential(credentialRecordId)
  }
```

TimoGlastra (commented):
Can you also fix the eslint warnings?

packages/core/src/modules/credentials/services/CredentialService.ts
```diff
@@ -250,10 +250,12 @@ export abstract class CredentialService {
...
     await this.credentialRepository.delete(credentialRecord)
 
-    if (options?.deleteAssociatedCredentials) {
+    const deleteAssociatedCredentials = options?.deleteAssociatedCredentials ?? true
+    
+    if (deleteAssociatedCredentials) {
       for (const credential of credentialRecord.credentials) {
         const formatService: CredentialFormatService = this.getFormatService(credential.credentialRecordType)
-        await formatService.deleteCredentialById(credential.credentialRecordId)
+        await formatService.deleteCredentialById(credentialRecord)
```
TimoGlastra, line 258:
```suggestion
        await formatService.deleteCredentialById(credential.credentialRecordId)
```

packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts
```diff
@@ -546,8 +546,16 @@ export class IndyCredentialFormatService extends CredentialFormatService {
...
+  public async deleteCredentialById(credentialRecord: CredentialExchangeRecord): Promise<void> {
+    const indyCredential = credentialRecord.credentials.filter((binding) => {
+      return binding.credentialRecordType == CredentialFormatType.Indy
+    })
+    if (indyCredential.length != 1) {
+      throw new AriesFrameworkError(`Could not find Indy record id for credential record ${credentialRecord.id}`)
+    }
+    if (indyCredential[0].credentialRecordId) {
+      await this.indyHolderService.deleteCredential(indyCredential[0].credentialRecordId)
+    }
```
TimoGlastra, line 558:
```suggestion
  public async deleteCredentialById(credentialRecordId: string): Promise<void> {
    await this.indyHolderService.deleteCredential(credentialRecordId)
```

packages/core/src/modules/credentials/formats/CredentialFormatService.ts
```diff
@@ -65,7 +65,7 @@ export abstract class CredentialFormatService {
   abstract shouldAutoRespondToRequest(options: HandlerAutoAcceptOptions): boolean
   abstract shouldAutoRespondToCredential(options: HandlerAutoAcceptOptions): boolean
 
-  abstract deleteCredentialById(credentialRecordId: string): Promise<void>
+  abstract deleteCredentialById(credentialRecord: CredentialExchangeRecord): Promise<void>
```
TimoGlastra, line 68:
```suggestion
  abstract deleteCredentialById(credentialRecordId: string): Promise<void>
```

TimoGlastra:
@NB-MikeRichardson updated the PR to only update it to delete by default. If you could add a tests that the default value is true (so it will be deleted automatically when we don't pass a value) we can get this merged!

packages/core/src/modules/credentials/__tests__/V1CredentialServiceCred.test.ts
```diff
@@ -826,6 +826,17 @@ describe('CredentialService', () => {
       })
       expect(deleteCredentialMock).toHaveBeenNthCalledWith(1, credentialRecord.credentials[0].credentialRecordId)
     })
+
+    it('deleteAssociatedCredentials not set - defaults to true , credential still deleted by default', async () => {
+      const deleteCredentialMock = indyHolderService.deleteCredential as jest.Mock<Promise<void>, [string]>
```
TimoGlastra, line 831:
```suggestion
      const deleteCredentialMock = mockFunction(indyHolderService.deleteCredential)
```

packages/core/src/modules/credentials/__tests__/V2CredentialServiceCred.test.ts
```diff
@@ -816,6 +816,17 @@ describe('CredentialService', () => {
       })
       expect(storeCredentialMock).toHaveBeenNthCalledWith(1, credentialRecord.credentials[0].credentialRecordId)
     })
+
+    it('deleteAssociatedCredentials not set - defaults to true , credential still deleted by default', async () => {
+      const deleteCredentialMock = indyHolderService.deleteCredential as jest.Mock<Promise<void>, [string]>
```
TimoGlastra, line 821:
```suggestion
      const deleteCredentialMock = mockFunction(indyHolderService.deleteCredential)
```

TimoGlastra (approved):

PR #769, opened 2022-05-12
fix: miscellaneous issue credential v2 fixes

NB-MikeRichardson:
Changes from ICV2 PR review comments which were left to do after the merge


packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts
```diff
@@ -129,10 +132,10 @@ export class IndyCredentialFormatService extends CredentialFormatService {
    *
    */
   public async createOffer(options: ServiceOfferCredentialOptions): Promise<FormatServiceOfferAttachmentFormats> {
-    const formats: CredentialFormatSpec = {
+    const formats = new CredentialFormatSpec({
       attachId: this.generateId(),
-      format: 'hlindy/cred-abstract@v2.0',
-    }
+      format: 'hlindy/cred-filter@v2.0',
```
TimoGlastra, line 137: This should be cred-abstract

packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts
```diff
@@ -181,24 +184,29 @@ export class IndyCredentialFormatService extends CredentialFormatService {
...
-    const { credReq, credReqMetadata } = await this.createIndyCredentialRequest(offer, credDef, holderDid)
+    const { credReq, credReqMetadata } = await this.createIndyCredentialRequest(offer, credDef, options.holderDid)
     credentialRecord.metadata.set(CredentialMetadataKeys.IndyRequest, credReqMetadata)
 
-    const formats: CredentialFormatSpec = {
+    const formats = new CredentialFormatSpec({
       attachId: this.generateId(),
-      format: 'hlindy/cred-req@v2.0',
-    }
+      format: 'hlindy/cred-filter@v2.0',
```
TimoGlastra, line 208: copy pasta? 😉

packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts
```diff
@@ -358,10 +366,10 @@ export class IndyCredentialFormatService extends CredentialFormatService {
       credentialValues: CredentialUtils.convertAttributesToValues(credentialAttributes),
     })
 
-    const formats: CredentialFormatSpec = {
+    const formats = new CredentialFormatSpec({
       attachId: this.generateId(),
-      format: 'hlindy/cred-abstract@v2.0',
-    }
+      format: 'hlindy/cred-filter@v2.0',
```
TimoGlastra, line 371: same

packages/core/src/modules/credentials/repository/CredentialExchangeRecord.ts
```diff
@@ -98,9 +96,8 @@ export class CredentialExchangeRecord extends BaseRecord<
   public getTags() {
     const metadata = this.metadata.get(CredentialMetadataKeys.IndyCredential)
     let Ids: string[] = []
-    if (this.credentials) {
-      Ids = this.credentials.map((c) => c.credentialRecordId)
-    }
+    Ids = this.credentials.map((c) => c.credentialRecordId)
```
TimoGlastra, line 99:
```suggestion
    const credentialIds = this.credentials.map((c) => c.credentialRecordId)
```

packages/core/src/modules/credentials/repository/CredentialExchangeRecord.ts
```diff
@@ -46,8 +46,6 @@ export interface CredentialRecordBinding {
   credentialRecordType: CredentialFormatType
   credentialRecordId: string
   credentialId?: string
```
TimoGlastra, line 48: credential Id can also be removed

```suggestion
```

packages/core/src/modules/credentials/repository/CredentialExchangeRecord.ts
```diff
@@ -46,8 +46,6 @@ export interface CredentialRecordBinding {
...
-  indyRevocationRegistryId?: string
-  indyCredentialRevocationId?: string
```
TimoGlastra, line 50: `indyRevocationRegistryId ` and `indyCredentialRevocationId ` were originally in the `DefaultCredentialTags`, and I think this got mixed up with the merge from main. Can you move them back?

TimoGlastra (commented):
Nice improvements @NB-MikeRichardson! 

Left some notes

packages/core/src/modules/credentials/repository/CredentialExchangeRecord.ts
TimoGlastra: We can then remove the previous let Ids: string[] line

packages/core/src/modules/credentials/repository/CredentialExchangeRecord.ts
```diff
@@ -96,7 +96,11 @@ export class CredentialExchangeRecord extends BaseRecord<
   public getTags() {
     const metadata = this.metadata.get(CredentialMetadataKeys.IndyCredential)
     let Ids: string[] = []
-    Ids = this.credentials.map((c) => c.credentialRecordId)
+
+    // without this check we get test failures
+    if (this.credentials) {
+      Ids = this.credentials.map((c) => c.credentialRecordId)
+    }
```
TimoGlastra, line 103: according to the types `this.credentials` is always defined. So we should probably look at the root cause why this isn't the case.

Which tests are failing? Maybe we should do `this.credentials = props.credentials || []` instead of `this.credentials = props.credentials ?? []`?

packages/core/src/modules/credentials/protocol/v2/V2CredentialService.ts
```diff
@@ -220,8 +219,7 @@ export class V2CredentialService extends CredentialService {
     credentialRecord: CredentialExchangeRecord
   ): Promise<CredentialProtocolMsgReturnType<V2OfferCredentialMessage>> {
     const options: ServiceOfferCredentialOptions = {
-      connectionId: proposal.connectionId ?? undefined,
-      protocolVersion: proposal.protocolVersion,
+      protocolVersion: credentialRecord.protocolVersion,
```
TimoGlastra, line 222: Why do we need protocolVersion here? It's already defined on the record and we're in v2 context here.

packages/core/src/modules/credentials/repository/CredentialExchangeRecord.ts
```diff
@@ -91,16 +89,14 @@ export class CredentialExchangeRecord extends BaseRecord<
...
     }
   }
 
   public getTags() {
     const metadata = this.metadata.get(CredentialMetadataKeys.IndyCredential)
-    let Ids: string[] = []
-    if (this.credentials) {
-      Ids = this.credentials.map((c) => c.credentialRecordId)
-    }
+    const Ids = this.credentials.map((c) => c.credentialRecordId)
```
TimoGlastra, line 98:
```suggestion
    const ids = this.credentials.map((c) => c.credentialRecordId)
```

packages/core/src/utils/CredentialResponseCoordinator.ts
```diff
@@ -1,24 +1,13 @@
...
-
-import { AutoAcceptCredential } from './CredentialAutoAcceptType'
+import { AutoAcceptCredential } from '../modules/credentials/CredentialAutoAcceptType'
 
 /**
  * This class handles all the automation with all the messages in the issue credential protocol
  * Every function returns `true` if it should automate the flow and `false` if not
  */
 @scoped(Lifecycle.ContainerScoped)
 export class CredentialResponseCoordinator {
```
TimoGlastra, line 10: What I meant with making it an util is to remove the whole class. The class used to have a lot of methods, but now it's just a simple function, so I don't see the need for it anymore. We can just call `composeAutoAccept` as a function instead of as `CredentialResponseCoordinator.composeAutoAccept`. It doesn't coordinate anything anymore

packages/core/src/modules/credentials/CredentialsModuleOptions.ts
```diff
@@ -41,7 +39,7 @@ interface NegotiateProposalOptions extends BaseOptions {
 interface OfferCredentialOptions extends BaseOptions {
   credentialRecordId?: string
   connectionId?: string
-  protocolVersion: CredentialProtocolVersion
+  protocolVersion?: CredentialProtocolVersion
```
TimoGlastra, line 42: protocolVersion shouldn't be optional for offerCredential, as it's the start of the protocol in this case

packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts
TimoGlastra: This doesn't seem resolved

packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts
TimoGlastra: This also doesn't seem resolved

packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts
TimoGlastra: This also doesn't seem resolved

packages/core/src/modules/credentials/repository/CredentialExchangeRecord.ts
```diff
@@ -91,24 +90,22 @@ export class CredentialExchangeRecord extends BaseRecord<
...
   public getTags() {
     const metadata = this.metadata.get(CredentialMetadataKeys.IndyCredential)
-    let credentialIds: string[] = []
-
+    let ids: string[] = []
     if (this.credentials) {
-      credentialIds = this.credentials.map((c) => c.credentialRecordId)
+      ids = this.credentials.map((c) => c.credentialRecordId)
     }
-
```
TimoGlastra, line 105: With the change to `|| []` I think we can just do:

```ts
const credentialIds = this.credentials.map(c => c.credentialRecordId)
```

packages/core/src/utils/composeAutoAccept.ts
```diff
@@ -0,0 +1,14 @@
+import { AutoAcceptCredential } from '../modules/credentials/CredentialAutoAcceptType'
+/**
+ * Returns the credential auto accept config based on priority:
+ *	- The record config takes first priority
+ *	- Otherwise the agent config
+ *	- Otherwise {@link AutoAcceptCredential.Never} is returned
+ */
+
+export function composeAutoAccept(
```
TimoGlastra, line 9: As this is a credential specific util (it uses `AutoAcceptCredential` enum type) can we move this to the credentials module directory?

packages/core/src/modules/credentials/repository/CredentialExchangeRecord.ts
NB-MikeRichardson: That change doesn't work
Test fails

packages/core/src/modules/credentials/CredentialsModuleOptions.ts
NB-MikeRichardson: That change breaks loads of stuff. I suggest leaving this for now

TimoGlastra: What is the reason for this? It is required in the current main branch.

Not keen on merging this change as it's now not clear you're required to pass the protocolVersion

NB-MikeRichardson: I have reverted it back to being mandatory for now.

As stated I would like to allot a PR just for this one Protocol Version field as I find it bewilderingly confusing and I want to fully understand it and do it properly.

packages/core/src/modules/credentials/protocol/v2/V2CredentialService.ts
```diff
@@ -221,6 +221,7 @@ export class V2CredentialService extends CredentialService {
     const options: ServiceOfferCredentialOptions = {
       credentialFormats: proposal.credentialFormats,
       comment: proposal.comment,
+      protocolVersion: credentialRecord.protocolVersion,
     }
     const message = await this.createOfferAsResponse(credentialRecord, options)
```
TimoGlastra, line 226: When creating an offer as response we can take the protocol version from the credential record

packages/core/src/modules/credentials/repository/CredentialExchangeRecord.ts
TimoGlastra: Then we should look at why it is failing. The types indicate the array is always defined, so somewhere it's assigning an undefined value where it should be an empty array I think. Please investigate.

packages/core/src/modules/credentials/protocol/v2/V2CredentialService.ts
TimoGlastra: This also seems unresolved @NB-MikeRichardson

TimoGlastra (approved):

PR #770, opened 2022-05-12
fix: delete credentials

NB-MikeRichardson:


packages/core/src/modules/credentials/protocol/v2/__tests__/v2credentials.propose-offer.test.ts
```diff
@@ -381,6 +382,38 @@ describe('credentials', () => {
...
+      },
+    })
+    // test that delete credential removes from both repository and wallet
+    // latter is tested by spying on holder service (Indy) to
+    // see if deleteCredential is called
+    const holderService = aliceAgent.injectionContainer.resolve(IndyHolderService)
+
+    const deleteCredentialSpy = jest.spyOn(holderService, 'deleteCredential')
+    await aliceAgent.credentials.deleteById(holderCredential.id, { deleteAssociatedCredentials: true })
+    expect(deleteCredentialSpy).toHaveBeenCalledTimes(1)
```
TimoGlastra, line 410: It would be good to check the deleteCredentialSpy has been called with the expected id

packages/core/src/modules/credentials/services/CredentialService.ts
```diff
@@ -247,18 +247,19 @@ export abstract class CredentialService {
...
+  public async deleteById(credentialRecordId: string, options?: DeleteCredentialOptions): Promise<void> {
+    const credentialRecord = await this.getById(credentialRecordId)
 
     await this.credentialRepository.delete(credentialRecord)
 
     if (options?.deleteAssociatedCredentials) {
       for (const credential of credentialRecord.credentials) {
         const formatService: CredentialFormatService = this.getFormatService(credential.credentialRecordType)
-        await formatService.deleteCredentialById(credentialRecord, options)
+        await formatService.deleteCredentialById(credentialRecordId)
```
TimoGlastra, line 258: This is incorrect. We should be passing the credentialId / credentialRecordId, not the credential exchange id. (so `credential.credentialRecordId`). The indy credential service doesn't know about the `CredentialExchangeRecord.id`)

packages/core/src/modules/credentials/__tests__/V2CredentialService.cred.test.ts
```diff
@@ -806,11 +806,11 @@ describe('CredentialService', () => {
 
       const credential = mockCredentialRecord()
       mockFunction(credentialRepository.getById).mockReturnValue(Promise.resolve(credential))
-
+      credential.credentials[0].credentialRecordId
```
TimoGlastra, line 809:
```suggestion
```

packages/core/src/modules/credentials/__tests__/V2CredentialService.cred.test.ts
```diff
@@ -806,11 +806,11 @@ describe('CredentialService', () => {
...
       await credentialService.deleteById(credential.id, {
         deleteAssociatedCredentials: true,
       })
-      expect(storeCredentialMock).toHaveBeenNthCalledWith(1, credential.credentials[0].credentialRecordId)
+      expect(storeCredentialMock).toHaveBeenNthCalledWith(1, credential.id)
```
TimoGlastra, line 813: this should still be the actual credential record id as used for the indy credential, not the `CredentialExchangeRecord.id`

```suggestion
      expect(storeCredentialMock).toHaveBeenNthCalledWith(1, credential.credentials[0].credentialRecordId)
```

TimoGlastra, line 813: also, it should probably be called deleteCredentialMock?

```suggestion
      expect(deleteCredentialMock).toHaveBeenNthCalledWith(1, credential.id)
```

packages/core/src/modules/credentials/__tests__/V1CredentialService.cred.test.ts
```diff
@@ -820,7 +820,7 @@ describe('CredentialService', () => {
       await credentialService.deleteById(credential.id, {
         deleteAssociatedCredentials: true,
       })
-      expect(storeCredentialMock).toHaveBeenNthCalledWith(1, credential.credentials[0].credentialRecordId)
+      expect(storeCredentialMock).toHaveBeenNthCalledWith(1, credential.id)
```
TimoGlastra, line 823: same comments as v2 service

packages/core/src/modules/credentials/CredentialsModule.ts
```diff
@@ -501,14 +502,17 @@ export class CredentialsModule implements CredentialsModule {
...
    *
    * @param credentialId the credential record id
+   * @param options the delete credential options for the delete operation
    */
-  public async deleteById(credentialId: string) {
+  public async deleteById(credentialId: string, options?: DeleteCredentialOptions) {
     const credentialRecord = await this.getById(credentialId)
-    return this.credentialRepository.delete(credentialRecord)
+    const service = this.getService(credentialRecord.protocolVersion)
+    return service.deleteById(credentialId, options)
```
TimoGlastra, line 515: This is a nit, but we're retrieving the record two times now just to delete it (which wasn't the case before this PR). Could we make this a `delete` and pass the record instead of the id?

packages/core/src/modules/credentials/services/CredentialService.ts
```diff
@@ -247,18 +247,20 @@ export abstract class CredentialService {
     return this.credentialRepository.findById(connectionId)
   }
 
-  public async deleteById(credentialId: string, options?: DeleteCredentialOptions): Promise<void> {
-    const credentialRecord = await this.getById(credentialId)
-
+  public async deleteById(
+    credentialRecord: CredentialExchangeRecord,
```
TimoGlastra, line 251: It doesn't make sense to call this `deleteById` anymore, as we're deleting it by record.

```suggestion
  public async delete(
    credentialRecord: CredentialExchangeRecord,
```

TimoGlastra (approved):

PR #772, opened 2022-05-15
refactor: do not add ~service in createOOBOffer method

TimoGlastra:
When calling createOutOfBand offer and then attaching it to the oob invitation would create two set of keys. To overcome this the creating of an offer does not create the keys for the exchange anymore. Instead you just get the message that is not associated with a connection yet. Then you can:
- call oob.createInvitation to create an oob invitation
- call oob.createLegacyConnectionlessInvitation to create a pre-aip connectionless invitation by attaching the ~service decorator.

BREAKING CHANGE: with the addition of the out of band module `credentials.createOutOfBandOffer` is renamed to `credentials.createOffer` and no longer adds the `~service` decorator to the message. You need to call `oob.createLegacyConnectionlessInvitation` afterwards to use it for AIP-1 style connectionless exchanges. See [Migrating from AFJ 0.1.0 to 0.2.x](https://github.com/hyperledger/aries-framework-javascript/blob/main/docs/migration/0.1-to-0.2.md) for detailed migration instructions.


TimoGlastra:
@NB-MikeRichardson please take a look at the changes, this touches the credentials module

packages/core/src/modules/oob/OutOfBandModule.ts
```diff
@@ -240,6 +240,35 @@ export class OutOfBandModule {
...
+  public async createLegacyConnectionlessInvitation<Message extends AgentMessage>(config: {
+    recordId: string
+    message: Message
+    domain: string
+  }): Promise<{ message: Message; invitationUrl: string }> {
+    // Create keys (and optionally register them at the mediator)
+    const routing = await this.mediationRecipientService.getRouting()
+
+    // Set the service on the message
+    config.message.service = new ServiceDecorator({
```
blu3beri, line 252: I am not the biggest fan of reassigning a parameter variable like here. Since we pass objects by reference in JavaScript, it will override and could cause unexpected behaviour. I suggest just dereferencing with addition:

```typescript
const messageService = new ServiceDecorator({...})

const foo = {
  ...config,
  message: {...config.message, service: serviceMessage}
} 
```

It might be slightly messy code, but it makes sure we don't edit the params.

packages/core/src/storage/didcomm/DidCommMessageRepository.ts
```diff
@@ -27,9 +28,12 @@ export class DidCommMessageRepository extends Repository<DidCommMessageRecord> {
   }
 
   public async saveOrUpdateAgentMessage(options: SaveAgentMessageOptions) {
+    const parsedMessageType = parseMessageType(options.agentMessage.type)
```
blu3beri, line 31: might be nice to destruct the return type here.

blu3beri (approved):
Nothing major, just two code cleanup things.

packages/core/src/modules/oob/OutOfBandModule.ts
TimoGlastra: Hmm it's a class instance, so we'd have to clone it somehow. I'll look into it

TimoGlastra: I understand your point, but couldn't find an easy way to do this. We do this in a lot of places and therefore we should look at fixing this a separate PR. Ignoring for now

PR #773, opened 2022-05-15
feat: 0.2.0 migration script for connections

TimoGlastra:
Adds migration script for the changes introduced by the out of band and did exchange pr. See the markdown file for a description of the changes.


packages/core/src/modules/oob/messages/OutOfBandInvitation.ts
```diff
@@ -98,6 +100,15 @@ export class OutOfBandInvitation extends AgentMessage {
     return dids
   }
 
+  // TODO: this only takes into account inline didcomm services, won't work for public dids
+  public getRecipientKeys(): Key[] {
+    return this.services
+      .filter((s): s is OutOfBandDidCommService => typeof s !== 'string')
+      .map((s) => s.recipientKeys)
+      .reduce((acc, curr) => [...acc, ...curr], [])
```
blu3beri, line 108: What does this reducer do? it seems like a copy without propose.

blu3beri (approved):

packages/core/src/modules/oob/messages/OutOfBandInvitation.ts
TimoGlastra: It's basically a flatmap. We have an array of arrays

PR #775, opened 2022-05-17
fix: propose payload attachment in in snake_case JSON format

NB-MikeRichardson:

### DIFF
--- a/packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts
+++ b/packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts
@@ -81,7 +81,7 @@ export class IndyCredentialFormatService extends CredentialFormatService {
    * @returns object containing associated attachment, formats and filtersAttach elements
    *
    */
-  public createProposal(options: ProposeCredentialOptions): FormatServiceProposeAttachmentFormats {
+  public async createProposal(options: ProposeCredentialOptions): Promise<FormatServiceProposeAttachmentFormats> {
     const formats: CredentialFormatSpec = {
       attachId: this.generateId(),
       format: 'hlindy/cred-filter@v2.0',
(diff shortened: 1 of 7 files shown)


packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts
```diff
@@ -89,7 +90,12 @@ export class IndyCredentialFormatService extends CredentialFormatService {
       throw new AriesFrameworkError('Missing payload in createProposal')
     }
 
-    const attachment: Attachment = this.getFormatData(options.credentialFormats.indy?.payload, formats.attachId)
+    // Use class instance instead of interface, otherwise this causes interoperability problems
+    let proposal = new CredPropose(options.credentialFormats.indy?.payload)
```
TimoGlastra, line 94: Would be good to also validate the proposal class instance using `MessageValidator.validate` (and add an exception test)

packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts
```diff
@@ -89,7 +90,12 @@ export class IndyCredentialFormatService extends CredentialFormatService {
...
+    proposal = JsonTransformer.toJSON(proposal)
+
+    const attachment: Attachment = this.getFormatData(proposal, formats.attachId)
```
TimoGlastra, line 97: In TS, different than most programming languages like e.g. Java, C#, ..., common behaviour is to only use type annotations when you need to. the `getFormatData` method has a return type of `Attachment` making the explicit definition of `Attachment` unnecessary. This in generally also the pattern we use in AFJ.

```suggestion
    const attachment = this.getFormatData(proposal, formats.attachId)
```

TimoGlastra (commented):
The code looks good to me, but could you add a test case that checks the proposal message and makes sure the attributes values are encoded as should be according to the RFC?

TimoGlastra (approved):

TimoGlastra (approved):

PR #779, opened 2022-05-19
docs: fix demo for oob

TimoGlastra:


demo/src/Alice.ts
```diff
@@ -20,33 +26,61 @@ export class Alice extends BaseAgent {
   }
 
   private async getConnectionRecord() {
-    if (!this.connectionRecordFaberId) {
+    if (!this.outOfBandId) {
+      throw Error(redText(Output.MissingConnectionRecord))
+    }
+
+    const [connection] = await this.agent.connections.findAllByOutOfBandId(this.outOfBandId)
```
blu3beri, line 33: This looks to me like we want a find single in the public api aswell.

blu3beri (approved):

demo/src/Alice.ts
TimoGlastra: Well there could be multiple, so not sure we should add that method. For demo purposes it's fine to pick the first one I guess

PR #787, opened 2022-05-20
fix(routing): sending of trustping in pickup v2

KolbyRKunz:
There was a bug discovered that would cause the first message in the pick up queue to be delivered when the trust ping was sent to create the web socket connection. This PR addresses that by only sending the trust ping when the request status receives a status of zero, then indicating to the mediator that we are ready to actively receive messages.


TimoGlastra (approved):
This is getting quite complex and hard to follow, we should probably look at splitting up the mediator service in different pickup services and let each manage their own routing lifecycle.

TimoGlastra:
can you update @KolbyRKunz

KolbyRKunz:
I updated this to main @TimoGlastra, I think separating the the pick up services would be a good idea to de-tangle what is going on currently.

PR #789, opened 2022-05-23
fix: add BBS context to DidDoc

karimStekelenburg:
BBS context was missing in the DidDocumentBuilder, resulting in some value being dropped during frameing. This PR adds the context.


blu3beri (approved):
Not really sure if it is really needed, but some tests might be nice (could be included in the other PR)

PR #795, opened 2022-05-25
feat: Issue Credentials V2: W3C/ jsonld credentials

NB-MikeRichardson:
Merge of jsonld-credentials -> 0.3.0-pre


packages/core/src/modules/credentials/protocol/v2/V2CredentialService.ts
```diff
@@ -1082,7 +1101,7 @@ export class V2CredentialService extends CredentialService {
       if (msg.format.includes('indy')) {
         formats.push(this.getFormatService(CredentialFormatType.Indy))
       } else if (msg.format.includes('aries')) {
-        // todo
+        formats.push(this.getFormatService(CredentialFormatType.JsonLd))
       } else {
         throw new AriesFrameworkError(`Unknown Message Format: ${msg.format}`)
       }
```
TimoGlastra, line 1107: We should remove this method. It requires the credential service to be aware of the different format services. In the proof PR this is solved by adding a method to the format service asking whether it supporst the format (just like you have the `getAttachment`on the format service that knows which format identifier to look for. This makes it possible to add completely new formats in the future without needing to modfiy the credential service (and would even allow for people to add them dynamically by implementing the format service interface) this would prevent that from working.

TimoGlastra, line 1107: We should remove this method. It requires the credential service to be aware of the different format services. In the proof PR this is solved by adding a method to the format service asking whether it supporst the format (just like you have the `getAttachment`on the format service that knows which format identifier to look for. This makes it possible to add completely new formats in the future without needing to modfiy the credential service (and would even allow for people to add them dynamically by implementing the format service interface) this would prevent that from working.

packages/core/src/modules/credentials/protocol/v2/__tests__/v2.ldproof.connectionless-credentials.test.ts
```diff
@@ -0,0 +1,204 @@
...
+  test('Faber starts with V2 W3C connection-less credential offer to Alice', async () => {
+    testLogger.test('Faber sends credential offer to Alice')
+
+    const offerOptions: OfferCredentialOptions = {
+      comment: 'V2 Out of Band offer (W3C)',
+      credentialFormats: {
+        jsonld: signCredentialOptions,
+      },
+      protocolVersion: CredentialProtocolVersion.V2,
+      connectionId: '',
```
TimoGlastra, line 138: connectionId should not be an allowed value for `createOutOfBandOffer` and definitely shouldn't be allowed as an empty string (this should blow up!)

```suggestion
```

TimoGlastra, line 138: connectionId should not be an allowed value for `createOutOfBandOffer` and definitely shouldn't be allowed as an empty string (this should blow up!)

```suggestion
```

packages/core/src/modules/credentials/protocol/v2/__tests__/v2.ldproof.connectionless-credentials.test.ts
```diff
@@ -0,0 +1,204 @@
...
+    let aliceCredentialRecord = await waitForCredentialRecordSubject(aliceReplay, {
+      threadId: faberCredentialRecord.threadId,
+      state: CredentialState.OfferReceived,
+    })
+
+    testLogger.test('Alice sends credential request to Faber')
+    const acceptOfferOptions: AcceptOfferOptions = {
+      credentialRecordId: aliceCredentialRecord.id,
+    }
+    aliceCredentialRecord.protocolVersion = CredentialProtocolVersion.V2
```
TimoGlastra, line 156: This shouldn't be needed in a test
```suggestion
```

TimoGlastra, line 156: This shouldn't be needed in a test
```suggestion
```

packages/core/src/modules/credentials/protocol/v2/__tests__/v2.ldproof.connectionless-credentials.test.ts
```diff
@@ -0,0 +1,204 @@
...
+    aliceCredentialRecord = await waitForCredentialRecordSubject(aliceReplay, {
+      threadId: faberCredentialRecord.threadId,
+      state: CredentialState.CredentialReceived,
+    })
+
+    testLogger.test('Alice sends credential ack to Faber')
+    aliceCredentialRecord = await aliceAgent.credentials.acceptCredential(aliceCredentialRecord.id)
+
+    testLogger.test('Faber waits for credential ack from Alice')
+    faberCredentialRecord = await waitForCredentialRecordSubject(faberReplay, {
```
TimoGlastra, line 183: We should add a check to this test that the holder has the actual w3c credential in their wallet. (and check for all the values)

TimoGlastra, line 183: We should add a check to this test that the holder has the actual w3c credential in their wallet. (and check for all the values)

packages/core/src/modules/credentials/protocol/v2/__tests__/v2.ldproof.connectionless-credentials.test.ts
```diff
@@ -0,0 +1,204 @@
...
+      threadId: expect.any(String),
+    })
+
+    expect(faberCredentialRecord).toMatchObject({
+      type: CredentialExchangeRecord.type,
+      id: expect.any(String),
+      createdAt: expect.any(Date),
+      state: CredentialState.Done,
+      threadId: expect.any(String),
+    })
```
TimoGlastra, line 202: Can we add some more property check here? E.g. `connectionId` is undefined, `threadId` is the same for both record, `credentials` array contains the credential record binding on the holder side, the protocol version is set to the correct version, etc...

TimoGlastra, line 202: Can we add some more property check here? E.g. `connectionId` is undefined, `threadId` is the same for both record, `credentials` array contains the credential record binding on the holder side, the protocol version is set to the correct version, etc...

packages/core/src/modules/credentials/protocol/v2/__tests__/v2.ldproof.credentials-auto-accept.test.ts
```diff
@@ -0,0 +1,481 @@
...
+        'faber agent: always v2 jsonld',
+        'alice agent: always v2 jsonld',
+        AutoAcceptCredential.Always
+      ))
+      wallet = faberAgent.injectionContainer.resolve(IndyWallet)
+      const issuerDidInfo = await wallet.createDid({ seed })
+      const issuerKey = Key.fromPublicKeyBase58(issuerDidInfo.verkey, KeyType.Ed25519)
+      issuerDidKey = new DidKey(issuerKey)
+      verificationMethod = `${issuerDidKey.did}#${issuerDidKey.key.fingerprint}`
+      const inputDoc = {
```
TimoGlastra, line 55: Maybe we can extract this to a `__fixtures__` folder and store them as json. Then we store both the unsigned and the signed variant and we can make sure the output matches the stored fixtures. This makes the tests files smaller and we have all the test data in a single place. We do this in other places in the framework (look for json files / fixtures folders etc..)

TimoGlastra, line 55: Maybe we can extract this to a `__fixtures__` folder and store them as json. Then we store both the unsigned and the signed variant and we can make sure the output matches the stored fixtures. This makes the tests files smaller and we have all the test data in a single place. We do this in other places in the framework (look for json files / fixtures folders etc..)

packages/core/src/modules/credentials/protocol/v2/__tests__/v2.ldproof.credentials-auto-accept.test.ts
```diff
@@ -0,0 +1,481 @@
...
+      const offerOptions: OfferCredentialOptions = {
+        comment: 'V2 Offer Credential',
+        connectionId: faberConnection.id,
+        credentialFormats: {
+          indy: {
+            attributes: credentialPreview.attributes,
+            credentialDefinitionId: credDefId,
+          },
+        },
+        protocolVersion: CredentialProtocolVersion.V2,
```
TimoGlastra, line 148: this looks like an indy credential in the ldproof test file?

TimoGlastra, line 148: this looks like an indy credential in the ldproof test file?

packages/core/src/modules/credentials/protocol/v2/__tests__/v2.ldproof.credentials-auto-accept.test.ts
```diff
@@ -0,0 +1,481 @@
...
+
+        expect(faberCredentialRecord).toMatchObject({
+          type: CredentialExchangeRecord.type,
+          id: expect.any(String),
+          createdAt: expect.any(Date),
+          state: CredentialState.Done,
+        })
+      } else {
+        throw new AriesFrameworkError('missing alice connection id')
+      }
```
TimoGlastra, line 372: If we just do the throw in the if, we don't need the else (and have less nesting)

```suggestion
      if (!aliceCredentialRecord.connectionId)  throw new AriesFrameworkError('missing alice connection id')
      
	    // we do not need to specify connection id in this object
	    // it is either connectionless or included in the offer message
	    const acceptOfferOptions: AcceptOfferOptions = {
	    credentialRecordId: aliceCredentialRecord.id,
	    }
	    testLogger.test('Alice sends credential request to faber')
	    const faberCredentialExchangeRecord: CredentialExchangeRecord = await aliceAgent.credentials.acceptOffer(
	    acceptOfferOptions
	    )
	    
	    testLogger.test('Alice waits for credential from Faber')
	    aliceCredentialRecord = await waitForCredentialRecord(aliceAgent, {
	    threadId: faberCredentialExchangeRecord.threadId,
	    state: CredentialState.CredentialReceived,
	    })
	    
	    testLogger.test('Faber waits for credential ack from Alice')
	    
	    const faberCredentialRecord = await waitForCredentialRecord(faberAgent, {
	    threadId: faberCredentialExchangeRecord.threadId,
	    state: CredentialState.Done,
	    })
	    
	    expect(aliceCredentialRecord).toMatchObject({
	    type: CredentialExchangeRecord.type,
	    id: expect.any(String),
	    createdAt: expect.any(Date),
	    metadata: {},
	    state: CredentialState.Done,
	    })
	    
	    expect(faberCredentialRecord).toMatchObject({
	    type: CredentialExchangeRecord.type,
	    id: expect.any(String),
	    createdAt: expect.any(Date),
	    state: CredentialState.Done,
	    })
```

TimoGlastra, line 372: If we just do the throw in the if, we don't need the else (and have less nesting)

```suggestion
      if (!aliceCredentialRecord.connectionId)  throw new AriesFrameworkError('missing alice connection id')
      
	    // we do not need to specify connection id in this object
	    // it is either connectionless or included in the offer message
	    const acceptOfferOptions: AcceptOfferOptions = {
	    credentialRecordId: aliceCredentialRecord.id,
	    }
	    testLogger.test('Alice sends credential request to faber')
	    const faberCredentialExchangeRecord: CredentialExchangeRecord = await aliceAgent.credentials.acceptOffer(
	    acceptOfferOptions
	    )
	    
	    testLogger.test('Alice waits for credential from Faber')
	    aliceCredentialRecord = await waitForCredentialRecord(aliceAgent, {
	    threadId: faberCredentialExchangeRecord.threadId,
	    state: CredentialState.CredentialReceived,
	    })
	    
	    testLogger.test('Faber waits for credential ack from Alice')
	    
	    const faberCredentialRecord = await waitForCredentialRecord(faberAgent, {
	    threadId: faberCredentialExchangeRecord.threadId,
	    state: CredentialState.Done,
	    })
	    
	    expect(aliceCredentialRecord).toMatchObject({
	    type: CredentialExchangeRecord.type,
	    id: expect.any(String),
	    createdAt: expect.any(Date),
	    metadata: {},
	    state: CredentialState.Done,
	    })
	    
	    expect(faberCredentialRecord).toMatchObject({
	    type: CredentialExchangeRecord.type,
	    id: expect.any(String),
	    createdAt: expect.any(Date),
	    state: CredentialState.Done,
	    })
```

packages/core/src/modules/credentials/protocol/v2/__tests__/v2.ldproof.credentials-auto-accept.test.ts
```diff
@@ -0,0 +1,481 @@
...
+        state: CredentialState.ProposalReceived,
+      })
+
+      // Check if the state of faber credential record did not change
+      const faberRecord = await faberAgent.credentials.getById(faberCredentialRecord.id)
+      faberRecord.assertState(CredentialState.ProposalReceived)
+
+      aliceCredentialRecord = await aliceAgent.credentials.getById(aliceCredentialRecord.id)
+      aliceCredentialRecord.assertState(CredentialState.ProposalSent)
+      await sleep(5000)
```
TimoGlastra, line 478: Why is this needed? We shouldn't add sleeps to tests. 

```suggestion
```

TimoGlastra, line 478: Why is this needed? We shouldn't add sleeps to tests. 

```suggestion
```

packages/core/tests/helpers.ts
```diff
@@ -185,7 +185,7 @@ export function waitForCredentialRecordSubject(
     threadId,
     state,
     previousState,
-    timeoutMs = 10000,
+    timeoutMs = 15000,
```
TimoGlastra, line 188: Why is this needed?

TimoGlastra, line 188: Why is this needed?

packages/core/src/storage/migration/__tests__/__snapshots__/0.1.test.ts.snap
```diff
@@ -53,10 +53,8 @@ Object {
       },
       "id": "574b2a37-1db1-4af1-a3bf-35c6cb9e1d7a",
       "metadata": Object {
-        "_internal/indyCredential": Object {
```
TimoGlastra, line 56: Something is not correct here. this is back to the old format and shouldn't be changed.

TimoGlastra, line 56: Something is not correct here. this is back to the old format and shouldn't be changed.

packages/core/src/modules/credentials/protocol/v2/__tests__/v2credentials-architecture.test.ts
```diff
@@ -108,14 +158,14 @@ describe('V2 Credential Architecture', () => {
       const credFormats: FormatServiceProposeCredentialFormats =
         multiFormatProposal.credentialFormats as FormatServiceProposeCredentialFormats
       const formats: CredentialFormatService[] = service.getFormats(credFormats)
-      expect(formats.length).toBe(1) // for now will be added to with jsonld
+      expect(formats.length).toBe(2) // for now will be added to with jsonld
```
TimoGlastra, line 161:
```suggestion
      expect(formats.length).toBe(2) 
```

TimoGlastra, line 161:
```suggestion
      expect(formats.length).toBe(2) 
```

packages/core/src/modules/credentials/protocol/v2/__tests__/v2credentials-architecture.test.ts
```diff
@@ -108,14 +158,14 @@ describe('V2 Credential Architecture', () => {
       const credFormats: FormatServiceProposeCredentialFormats =
```
TimoGlastra, line 158: Can we add an e2e flow where we issue credential with multiple formats? So propose both formats, offer both formats, request both formats, issue both formats

TimoGlastra, line 158: Can we add an e2e flow where we issue credential with multiple formats? So propose both formats, offer both formats, request both formats, issue both formats

packages/core/src/modules/credentials/protocol/v2/__tests__/v2.ldproof.credentials.propose-offerED25519.test.ts
```diff
@@ -0,0 +1,540 @@
...
+    testLogger.test('Faber sends credential offer to Alice')
+    await faberAgent.credentials.acceptProposal(options)
+
+    testLogger.test('Alice waits for credential offer from Faber')
+    aliceCredentialRecord = await waitForCredentialRecord(aliceAgent, {
+      threadId: faberCredentialRecord.threadId,
+      state: CredentialState.OfferReceived,
+    })
+
+    didCommMessageRepository = faberAgent.injectionContainer.resolve<DidCommMessageRepository>(DidCommMessageRepository)
```
TimoGlastra, line 150:
```suggestion
    didCommMessageRepository = faberAgent.injectionContainer.resolve(DidCommMessageRepository)
```

TimoGlastra, line 150:
```suggestion
    didCommMessageRepository = faberAgent.injectionContainer.resolve(DidCommMessageRepository)
```

packages/core/src/modules/credentials/protocol/v2/__tests__/v2.ldproof.credentials.propose-offerED25519.test.ts
```diff
@@ -0,0 +1,540 @@
...
+
+    const offerMessage = await didCommMessageRepository.findAgentMessage({
+      associatedRecordId: faberCredentialRecord.id,
+      messageClass: V2OfferCredentialMessage,
+    })
+
+    expect(JsonTransformer.toJSON(offerMessage)).toMatchObject({
```
TimoGlastra, line 157: 👍

TimoGlastra, line 157: 👍

packages/core/src/modules/credentials/protocol/v2/__tests__/v2.ldproof.credentials.propose-offerED25519.test.ts
```diff
@@ -0,0 +1,540 @@
...
+        {
+          attach_id: expect.any(String),
+          format: 'aries/ld-proof-vc-detail@v1.0',
+        },
+      ],
+      'offers~attach': [
+        {
+          '@id': expect.any(String),
+          'mime-type': 'application/json',
+          data: expect.any(Object),
```
TimoGlastra, line 171: Can we also check the contents of the attachment?

TimoGlastra, line 171: Can we also check the contents of the attachment?

packages/core/src/modules/credentials/protocol/v2/__tests__/v2.ldproof.credentials.propose-offerED25519.test.ts
```diff
@@ -0,0 +1,540 @@
...
+        '~please_ack': { on: ['RECEIPT'] },
+        '~service': undefined,
+        '~attach': undefined,
+        '~timing': undefined,
+        '~transport': undefined,
+        '~l10n': undefined,
+      })
+    } else {
+      throw new Error('Missing Connection Id')
+    }
```
TimoGlastra, line 286: Can we invert these again? Then we don't need the else clause and we can keep most code top level

TimoGlastra, line 286: Can we invert these again? Then we don't need the else clause and we can keep most code top level

packages/core/src/modules/credentials/protocol/v2/__tests__/v2.ldproof.credentials.propose-offerED25519.test.ts
```diff
@@ -0,0 +1,540 @@
...
+          }),
+        }),
+      ],
+      payload: {
+        schemaIssuerDid: 'GMm4vMw8LLrLJjp81kRRLp',
+        schemaName: 'ahoy',
+        schemaVersion: '1.0',
+        schemaId: '1560364003',
+        issuerDid: 'GMm4vMw8LLrLJjp81kRRLp',
+      },
```
TimoGlastra, line 326: Why do we have the same attributes multiple times (top level and in the payload object)? Can we make sure we're reusing the `schemaIssuerDid` / `schemaName` / `schemaVersion` / `schemaId` / `issuerDid` / `credentialDefinitionId` values?

TimoGlastra, line 326: Why do we have the same attributes multiple times (top level and in the payload object)? Can we make sure we're reusing the `schemaIssuerDid` / `schemaName` / `schemaVersion` / `schemaId` / `issuerDid` / `credentialDefinitionId` values?

TimoGlastra: I think we can just remove the payload object and use the top level properties everywhere

TimoGlastra: I think we can just remove the payload object and use the top level properties everywhere

packages/core/src/modules/credentials/protocol/v2/__tests__/v2.ldproof.credentials.propose-offerED25519.test.ts
```diff
@@ -0,0 +1,540 @@
...
+    testLogger.test('Faber sends credential offer to Alice')
+    await faberAgent.credentials.acceptProposal(options)
+
+    testLogger.test('Alice waits for credential offer from Faber')
+    aliceCredentialRecord = await waitForCredentialRecord(aliceAgent, {
+      threadId: faberCredentialRecord.threadId,
+      state: CredentialState.OfferReceived,
+    })
+
+    didCommMessageRepository = faberAgent.injectionContainer.resolve<DidCommMessageRepository>(DidCommMessageRepository)
```
TimoGlastra, line 376:
```suggestion
    didCommMessageRepository = faberAgent.injectionContainer.resolve(DidCommMessageRepository)
```
```suggestion
    didCommMessageRepository = faberAgent.injectionContainer.resolve<DidCommMessageRepository>(DidCommMessageRepository)
```

TimoGlastra, line 376:
```suggestion
    didCommMessageRepository = faberAgent.injectionContainer.resolve(DidCommMessageRepository)
```

packages/core/src/modules/credentials/protocol/v2/__tests__/v2.ldproof.credentials.propose-offerED25519.test.ts
```diff
@@ -0,0 +1,540 @@
...
+          '@id': expect.any(String),
+          'mime-type': 'application/json',
+          data: expect.any(Object),
+          lastmod_time: undefined,
+          byte_count: undefined,
+        },
+        {
+          '@id': expect.any(String),
+          'mime-type': 'application/json',
+          data: expect.any(Object),
```
TimoGlastra, line 411: Same here, can we check the attachment data ?

TimoGlastra, line 411: Same here, can we check the attachment data ?

packages/core/src/modules/credentials/protocol/v2/__tests__/v2.ldproof.credentials.propose-offerED25519.test.ts
```diff
@@ -0,0 +1,540 @@
...
+      '~please_ack': undefined,
+      '~timing': undefined,
+      '~transport': undefined,
+      '~l10n': undefined,
+      replacement_id: undefined,
+    })
+    expect(aliceCredentialRecord.id).not.toBeNull()
+    expect(aliceCredentialRecord.type).toBe(CredentialExchangeRecord.type)
+
+    if (aliceCredentialRecord.connectionId) {
```
TimoGlastra, line 433: again, can we inverse so we don't have the nesting / the happy path top level?

TimoGlastra, line 433: again, can we inverse so we don't have the nesting / the happy path top level?

packages/core/src/modules/credentials/protocol/v2/__tests__/v2.ldproof.credentials.propose-offerED25519.test.ts
```diff
@@ -0,0 +1,540 @@
...
+  })
+
+  afterAll(async () => {
+    await faberAgent.shutdown()
+    await faberAgent.wallet.delete()
+    await aliceAgent.shutdown()
+    await aliceAgent.wallet.delete()
+  })
+
+  // -------------------------- V2 TEST BEGIN --------------------------------------------
```
TimoGlastra, line 102:
```suggestion
```

TimoGlastra, line 102:
```suggestion
```

packages/core/src/modules/credentials/protocol/v2/__tests__/v2.ldproof.credentials.propose-offerED25519.test.ts
```diff
@@ -0,0 +1,540 @@
...
+        '~timing': undefined,
+        '~transport': undefined,
+        '~l10n': undefined,
+      })
+    } else {
+      throw new Error('Missing Connection Id')
+    }
+  })
+})
+// -------------------------- V2 TEST END --------------------------------------------
```
TimoGlastra, line 540:
```suggestion
```

TimoGlastra, line 540:
```suggestion
```

packages/core/src/modules/credentials/protocol/v2/__tests__/v2.ldproof.credentials.propose-offerBBS.test.ts
```diff
@@ -0,0 +1,250 @@
...
+      '~timing': undefined,
+      '~transport': undefined,
+      '~l10n': undefined,
+      credential_preview: expect.any(Object),
+      replacement_id: undefined,
+    })
+    expect(aliceCredentialRecord.id).not.toBeNull()
+    expect(aliceCredentialRecord.type).toBe(CredentialExchangeRecord.type)
+
+    if (aliceCredentialRecord.connectionId) {
```
TimoGlastra, line 156: same here, please use if to throw error and make the happy path top level

TimoGlastra, line 156: same here, please use if to throw error and make the happy path top level

packages/core/src/modules/credentials/protocol/v2/__tests__/v2.ldproof.credentials.propose-offerBBS.test.ts
```diff
@@ -0,0 +1,250 @@
...
+let aliceConnection: ConnectionRecord
+let aliceCredentialRecord: CredentialExchangeRecord
+let faberCredentialRecord: CredentialExchangeRecord
+let wallet: IndyWallet
+let issuerDidKey: DidKey
+let didCommMessageRepository: DidCommMessageRepository
+let signCredentialOptions: SignCredentialOptions
+let verificationMethod: string
+const seed = 'testseed000000000000000000000001'
+describe('credentials, BBS+ signature', () => {
```
TimoGlastra, line 37: I don't see any checks for the actual credential that is issued. I think we should add more tests adding the actual format specific data (as this is a test specific to jsonld bbs). For all we know we're just exchanging empty json objects here.

TimoGlastra, line 37: I don't see any checks for the actual credential that is issued. I think we should add more tests adding the actual format specific data (as this is a test specific to jsonld bbs). For all we know we're just exchanging empty json objects here.

TimoGlastra (changes requested):
Nice @NB-MikeRichardson. Left a bunch of comments.

I seem to be missing a `JsonLdCredentialFormatService.test.ts` (i think we're also missing this for indy). I think this is important to test the format in isolation of the protocol. It also make testing a lot easier as we don't have to executre the whole protocol. We can just check whether it creates  /processes attachments as it should according to the RFC. This will become even more useful as we start extraciting these services into separate packages

TimoGlastra:
Hmm it looks like github somehow added all my comments twice, please ignore the duplication! :)

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
```diff
@@ -0,0 +1,344 @@
...
+    const attachmentId = options.attachId ? options.attachId : formats.attachId
+
+    // sign credential here. credential to be signed is received as the request attachment
+    // (attachment in the request message from holder to issuer)
+    const credentialOptions = requestAttachment?.getDataAsJson<SignCredentialOptions>()
+    const signCredentialOptions: SignCredentialOptions = {
+      credential: JsonTransformer.fromJSON(credentialOptions.credential, W3cCredential),
+      proofType: credentialOptions.proofType,
+      verificationMethod: credentialOptions.verificationMethod,
+    }
```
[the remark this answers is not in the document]
karimStekelenburg: The properties you mentioned are part of the W3cCredentialService API as optional properties as defined in the spec. Please see the interface [here](https://github.com/hyperledger/aries-framework-javascript/blob/be57da5725fe10802fc6cd12a75bf963c4a839fd/packages/core/src/modules/vc/models/W3cCredentialServiceOptions.ts#L9).

I could throw an error in the `sign` function of the `W3cCredentialService` if `credentialStatus` is provided, but I feel this is more appropriate at a higher level since that is tightly coupled to the protocol itself. Any thoughts on this @TimoGlastra?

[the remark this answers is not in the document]
karimStekelenburg: (same answer as in the other comment)

The properties you mentioned are part of the W3cCredentialService API as optional properties as defined in the spec. Please see the interface [here](https://github.com/hyperledger/aries-framework-javascript/blob/be57da5725fe10802fc6cd12a75bf963c4a839fd/packages/core/src/modules/vc/models/W3cCredentialServiceOptions.ts#L9).

I could throw an error in the `sign` function of the `W3cCredentialService` if `credentialStatus` is provided, but I feel this is more appropriate at a higher level since that is tightly coupled to the protocol itself. Any thoughts on this @TimoGlastra?

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
```diff
@@ -0,0 +1,344 @@
...
+   * Method called on reception of a propose credential message
+   * We do the necessary processing here to accept the proposal and do the state change, emit event etc.
+   * @param options the options needed to accept the proposal
+   */
+  public async processProposal(options: ServiceAcceptProposalOptions): Promise<void> {
+    const credPropose = options.proposalAttachment?.getDataAsJson<SignCredentialOptions>()
+
+    if (!credPropose) {
+      throw new AriesFrameworkError('Missing jsonld credential proposal data payload')
+    }
```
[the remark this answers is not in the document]
NB-MikeRichardson: Yes I made this change in Indy but missed it in jsonld.

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
```diff
@@ -0,0 +1,344 @@
...
+
+    options.credentialFormats = {
+      jsonld: credPropose,
+    }
```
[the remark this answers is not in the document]
NB-MikeRichardson: Same as earlier comment, see PR#829, I should be able to incorporate changes from there.

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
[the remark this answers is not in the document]
NB-MikeRichardson: I am checking for credential status...I throw an error if defined.

I don't see that those other properties are being used in the api (I can see they are optional params on the signCredential method).

What needs to be done here?

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
```diff
@@ -0,0 +1,344 @@
...
+        `JsonLd processCredential - Missing credential attachment for record id ${credentialRecord.id}`
+      )
+    }
+    const credentialAsJson = options.credentialAttachment.getDataAsJson<W3cVerifiableCredential>()
+
+    const credential = JsonTransformer.fromJSON(credentialAsJson, W3cVerifiableCredential)
+
+    const verifiableCredential: W3cCredentialRecord = await this.w3cCredentialService.storeCredential({
+      record: credential,
+    })
```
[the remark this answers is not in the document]
NB-MikeRichardson: Doesn't look like it does at the moment. One for @karimStekelenburg

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
```diff
@@ -0,0 +1,344 @@
...
+    proposal?: AttachmentData,
+    offer?: AttachmentData
+  ): boolean {
+    if (!request) {
+      return false
+    }
+    if (proposal || offer) {
+      const previousCredential = offer ? offer : proposal
+
+      if (previousCredential) {
```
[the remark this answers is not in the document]
NB-MikeRichardson: The compiler seems to disagree which is surprising. If I take out the 'if' it complains about `AttachmentData | undefined not assignable to previousCredential`

JamesKEbert:
Hey folks! I've just tried to run this within a React Native environment--has that been recently tested? I've ran into a number of issues. The most problematic though is that the documentLoaderNode `require`'s `https`, which doesn't work in a RN context. I noticed [Karim's addition](https://github.com/hyperledger/aries-framework-javascript/pull/795/commits/c4fd640a8c8dcd0d3aa5d68cda88b19879a91664) that provided a conditional that would load the appropriate documentLoader, however the imports are already occurring so therefore the dependency still fails at bundle time. 

Is this likely to work on RN right now, or am I troubleshooting at a problematic point in development and it's on the list to fix RN support? 

I appreciate all the effort happening here :) Thanks!

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
```diff
@@ -0,0 +1,316 @@
...
+  FormatAutoRespondProposalOptions,
+  FormatAutoRespondRequestOptions,
+  FormatCreateProposalReturn,
+  FormatCreateRequestOptions,
+} from '../CredentialFormatServiceOptions'
+import type { JsonLdCredentialFormat } from './JsonLdCredentialFormat'
+
+import { Lifecycle, scoped } from 'tsyringe'
+
+import { AriesFrameworkError } from '../../../../../src/error'
```
JamesKEbert, line 24:
```suggestion
import { AriesFrameworkError } from '../../../../error'
```
Goes one level too high, which is problematic once bundled :)

TimoGlastra:
@NB-MikeRichardson what is the status of this PR? It seems it is undoing a lot of other changes from previous PRs

genaris:
Hey @NB-MikeRichardson, I see there are some changes that seem not to be related with your PR (e.g. docs/setup-*, src/crypto/BbsService.ts) that I think were actually updated by some other PRs merged after your branch was first created. :thinking: Could you check if everything is in place?

NB-MikeRichardson:
> Hey @NB-MikeRichardson, I see there are some changes that seem not to be related with your PR (e.g. docs/setup-*, src/crypto/BbsService.ts) that I think were actually updated by some other PRs merged after your branch was first created. 🤔 Could you check if everything is in place?

I have taken out the files mentioned above; for the life of me I have no idea how they ended up in this PR but good spot anyway!

packages/core/src/modules/credentials/formats/CredentialFormatService.ts
```diff
@@ -88,4 +90,12 @@ export abstract class CredentialFormatService<CF extends CredentialFormat = Cred
 
     return attachment
   }
+
+  public areCredentialsEqual = (message1: Attachment, message2: Attachment): boolean => {
```
TimoGlastra, line 94: I don't think this needs to be public, and is something specific to the jsonld credential format service. Can we make this a private method of that class?

Also is the FIXME resolved?

packages/core/src/modules/credentials/formats/CredentialFormatServiceOptions.ts
```diff
@@ -89,6 +93,14 @@ export interface FormatAcceptRequestOptions<CF extends CredentialFormat> {
   offerAttachment?: Attachment
 }
 
+export interface FormatAcceptCredentialOptions<CF extends CredentialFormat> {
+  credentialRecord: CredentialExchangeRecord
+  credentialFormats?: CredentialFormatPayload<[CF], 'acceptRequest'>
```
TimoGlastra, line 98: `acceptCredential` doesn't need any input right? If we accept the credential it means we'll send an acknowledgement.

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
```diff
@@ -0,0 +1,440 @@
+import type {
+  FormatAcceptOfferOptions,
+  FormatAcceptProposalOptions,
+  FormatCreateOfferOptions,
+  FormatCreateOfferReturn,
+  FormatCreateProposalOptions,
+  FormatCreateReturn,
+  FormatProcessOptions,
+} from '..'
```
TimoGlastra, line 9: import from '..'

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialOptions.ts
```diff
@@ -0,0 +1,37 @@
...
+ */
+export class JsonLdCredential {
+  public constructor(options: JsonLdCredentialOptions) {
+    if (options) {
+      this.credential = options.credential
+      this.options = options.options
+    }
+  }
+
+  @Expose({ name: 'credential' })
```
TimoGlastra, line 25:
```suggestion
```

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialOptions.ts
```diff
@@ -0,0 +1,37 @@
...
+  @IsObject()
```
TimoGlastra, line 26:
```suggestion
  @Type(W3cCredential)
```

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialOptions.ts
```diff
@@ -0,0 +1,37 @@
...
+  public credential!: W3cCredential
+
+  @Expose({ name: 'options' })
+  @IsObject()
```
TimoGlastra, line 30:
```suggestion
  @Type(() => JsonLdOptionsRFC0593)
```

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialOptions.ts
```diff
@@ -0,0 +1,37 @@
...
+  public options!: JsonLdOptionsRFC0593
+
+  @Expose({ name: 'verificationMethod' })
+  @IsString()
+  @IsOptional()
+  public verificationMethod!: string
```
TimoGlastra, line 36: This isn't part of the RFC I think?

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormat.ts
```diff
@@ -0,0 +1,27 @@
+import type { W3cCredential } from '../../../vc/models/credential/W3cCredential'
+import type { CredentialFormat } from '../CredentialFormat'
+import type { JsonLdOptionsRFC0593 } from './JsonLdOptionsRFC0593'
+
+export interface JsonLdAcceptRequestOptions extends SignCredentialOptionsRFC0593 {
+  verificationMethod?: string
+}
+
+// feel free to rename this, the RFC suffix is useful during dev
```
TimoGlastra, line 9:
```suggestion
```

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormat.ts
```diff
@@ -0,0 +1,27 @@
...
+export interface JsonLdCredentialFormat extends CredentialFormat {
+  formatKey: 'jsonld'
+  credentialRecordType: 'w3c'
+  credentialFormats: {
+    createProposal: SignCredentialOptionsRFC0593
+    acceptProposal: SignCredentialOptionsRFC0593
+    createOffer: SignCredentialOptionsRFC0593
+    acceptOffer: SignCredentialOptionsRFC0593
+    createRequest: SignCredentialOptionsRFC0593
+    acceptRequest: JsonLdAcceptRequestOptions
```
TimoGlastra, line 25: Should these all be the same for each call? E.g. if we have the acceptProposal call, we should probably only need to provide some values. Otherwise there's no different between the `acceptProposal` and `createOffer` methods.

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
```diff
@@ -0,0 +1,440 @@
...
+
+    if (!jsonLdFormat) {
+      throw new AriesFrameworkError('Missing jsonld payload in createProposal')
+    }
+
+    const jsonLdCredential = new JsonLdCredential(jsonLdFormat)
+    MessageValidator.validateSync(jsonLdCredential)
+
+    // FIXME: this doesn't follow RFC0593
+    // jsonLdFormat is now of type SignCredentialOptionsRFC0593
```
TimoGlastra, line 83: Can you remove the FIXME's that are fixed?

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
```diff
@@ -0,0 +1,440 @@
...
+    agentContext: AgentContext,
+    { attachId, credentialFormats, proposalAttachment }: FormatAcceptProposalOptions<JsonLdCredentialFormat>
+  ): Promise<FormatCreateOfferReturn> {
+    // if the offer has an attachment Id use that, otherwise the generated id of the formats object
+    const format = new CredentialFormatSpec({
+      attachId,
+      format: JSONLD_VC_DETAIL,
+    })
+
+    const jsonLdFormat = credentialFormats?.jsonld
```
TimoGlastra, line 117: we should validate this

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
```diff
@@ -0,0 +1,440 @@
...
+    { credentialFormats, attachId }: FormatCreateOfferOptions<JsonLdCredentialFormat>
+  ): Promise<FormatCreateOfferReturn> {
+    // if the offer has an attachment Id use that, otherwise the generated id of the formats object
+    const format = new CredentialFormatSpec({
+      attachId,
+      format: JSONLD_VC_DETAIL,
+    })
+
+    // FIXME: SignCredentialOptions doesn't follow RFC0593
+    const jsonLdFormat = credentialFormats?.jsonld
```
TimoGlastra, line 147: we should validate this

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
```diff
@@ -0,0 +1,440 @@
...
+  public async acceptOffer(
+    agentContext: AgentContext,
+    { credentialFormats, attachId, offerAttachment }: FormatAcceptOfferOptions<JsonLdCredentialFormat>
+  ): Promise<FormatCreateReturn> {
+    const jsonLdFormat = credentialFormats?.jsonld
+
+    // FIXME: SignCredentialOptions doesn't follow RFC0593
+    // FIXME: Add validation of the jsonLdFormat data (if present)
+    const credentialOffer = offerAttachment.getDataAsJson<SignCredentialOptionsRFC0593>()
+    const requestData = jsonLdFormat ?? credentialOffer
```
TimoGlastra, line 176: we should validate if using the input data (jsonLdFormat)

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
```diff
@@ -0,0 +1,440 @@
...
+    agentContext: AgentContext,
+    { credentialFormats, attachId, requestAttachment }: FormatAcceptRequestOptions<JsonLdCredentialFormat>
+  ): Promise<FormatCreateReturn> {
+    const jsonLdFormat = credentialFormats?.jsonld
+
+    // sign credential here. credential to be signed is received as the request attachment
+    // (attachment in the request message from holder to issuer)
+    const credentialRequest = requestAttachment.getDataAsJson<SignCredentialOptionsRFC0593>()
+
+    let verificationMethod = credentialFormats?.jsonld?.verificationMethod
```
TimoGlastra, line 241:
```
const verificationMethod = credentialFormats?.jsonld?.verificationMethod ?? await this.deriveVerificationMethod(
   agentContext,
   credentialData.credential,
    credentialRequest
)

  if (!verificationMethod) {
    throw new AriesFrameworkError('Missing verification method in credential data')
  }
```

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
```diff
@@ -0,0 +1,440 @@
...
+
+    // FIXME: Need to transform from json to class instance
+    // FIXME: SignCredentialOptions doesn't follow RFC0593
+    // FIXME: Add validation of the jsonLdFormat data (if present)
+    const credentialData = jsonLdFormat ?? credentialRequest
```
TimoGlastra, line 246: we should validate jsonLdFormat if we're using that.

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
```diff
@@ -0,0 +1,440 @@
...
+    const format = new CredentialFormatSpec({
+      attachId,
+      format: JSONLD_VC,
+    })
+
+    const verifiableCredential = await this.w3cCredentialService.signCredential(agentContext, {
+      credential: JsonTransformer.fromJSON(credentialData.credential, W3cCredential),
+      proofType: credentialData.options.proofType,
+      verificationMethod: verificationMethod,
+    })
```
TimoGlastra, line 269: We're ignoring most of the options from the RFC (proofPurpose, created, challenge, domain, credentialStatus). If we arent' using these, they shouldn't be in the interface

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
```diff
@@ -0,0 +1,440 @@
...
+    const attachment = this.getFormatData(verifiableCredential, format.attachId)
+    return { format, attachment }
+  }
+
+  /**
+   * Derive a verification method using the issuer from the given verifiable credential
+   * @param credential the verifiable credential we want to sign
+   * @return the verification method derived from this credential and its associated issuer did, keys etc.
+   */
+  public async deriveVerificationMethod(
```
TimoGlastra, line 280: This can be private I think?

packages/core/src/modules/dids/domain/key-type/keyDidMapping.ts
```diff
@@ -24,6 +26,24 @@ const keyDidMapping: Record<KeyType, KeyDidMapping> = {
...
+//   [ProofType.EcdsaSecp256k1Signature2019]: '',
+//   [ProofType.EcdsaSecp256k1RecoverySignature2020]: '',
+//   [ProofType.JsonWebSignature2020]: '',
+//   [ProofType.RsaSignature2018]: '',
+//   [ProofType.GpgSignature2020]: '',
+//   [ProofType.JcsEd25519Signature2020]: '',
+//   [ProofType.BbsBlsSignatureProof2020]: VERIFICATION_METHOD_TYPE_BLS12381G2_KEY_2020,
+//   [ProofType.BbsBlsBoundSignatureProof2020]: '',
+// }
+
```
TimoGlastra, line 41:
```suggestion
```

packages/core/src/modules/dids/domain/key-type/keyDidMapping.ts
```diff
@@ -24,6 +26,24 @@ const keyDidMapping: Record<KeyType, KeyDidMapping> = {
...
+export const proofTypeKeyTypeMapping: Record<string, string> = {
+  ['Ed25519Signature2018']: VERIFICATION_METHOD_TYPE_ED25519_VERIFICATION_KEY_2018,
+  ['BbsBlsSignature2020']: VERIFICATION_METHOD_TYPE_BLS12381G2_KEY_2020,
```
TimoGlastra, line 44:
```suggestion
  Ed25519Signature2018: VERIFICATION_METHOD_TYPE_ED25519_VERIFICATION_KEY_2018,
  BbsBlsSignature2020: VERIFICATION_METHOD_TYPE_BLS12381G2_KEY_2020,
```

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
```diff
@@ -0,0 +1,440 @@
...
+    // this will throw an error if the issuer did is invalid
+    const issuerDidDocument = await this.didResolver.resolveDidDocument(agentContext, issuerDid)
+
+    // find first key which matches proof type
+    const proofType = credentialRequest.options.proofType
+
+    const keyType: string | undefined = proofTypeKeyTypeMapping[proofType]
+    if (!keyType) {
+      throw new AriesFrameworkError(`No Key Type found for proofType ${proofType}`)
+    }
```
TimoGlastra, line 300: The w3cCredentialService now has a method `getKeyTypesByProofType`, it also has a `getVerificationMethodTypesByProofType` which you could use to get the verification method type and pass that to `findVerficationMethodByVerificationMethodType` (modify from `findVerificationMethodByKeyType`)

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
```diff
@@ -0,0 +1,440 @@
...
+  }
+
+  public supportsFormat(format: string): boolean {
+    const supportedFormats = [JSONLD_VC_DETAIL, JSONLD_VC]
+
+    return supportedFormats.includes(format)
+  }
+
+  public async deleteCredentialById(): Promise<void> {
+    throw new Error('Method not implemented.')
```
TimoGlastra, line 387: not implemented

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
```diff
@@ -0,0 +1,440 @@
...
+    // https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/protocols/issue_credential/v2_0/formats/ld_proof/handler.py#L492
+    // TODO don't call areCredentialsEqual, call compareStuff() as per aca-py
+
+    const credentialAsJson = credentialAttachment.getDataAsJson<W3cVerifiableCredential>()
+    const credential = JsonTransformer.fromJSON(credentialAsJson, W3cVerifiableCredential)
+
+    const requestAsJson = requestAttachment.getDataAsJson<SignCredentialOptionsRFC0593>()
+    const request = JsonTransformer.fromJSON(requestAsJson, JsonLdCredential)
+
+    if (Array.isArray(credential.proof)) {
```
TimoGlastra, line 426: this needs to be reworked

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
```diff
@@ -0,0 +1,440 @@
...
+    // FIXME: we should do a lot of checks to verify if the credential we received is actually the credential
+    // we requested. We can take an example of the ACA-Py implementation:
+    // https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/protocols/issue_credential/v2_0/formats/ld_proof/handler.py#L492
+
+    // compare stuff in the proof object of the credential and request...based on aca-py
+
+    const requestAsJson = requestAttachment.getDataAsJson<SignCredentialOptionsRFC0593>()
+    const request = JsonTransformer.fromJSON(requestAsJson, JsonLdCredential)
+
+    if (Array.isArray(credential.proof)) {
```
TimoGlastra, line 334: same here. Maybe we can look at the ACA-Py implementation?

packages/core/src/modules/credentials/protocol/v2/V2CredentialService.ts
```diff
@@ -942,22 +943,26 @@ export class V2CredentialService<CFs extends CredentialFormat[] = CredentialForm
...
     }
 
     // not all formats use the proposal and preview, we only check if they're present on
     // either or both of the messages
-    if (proposalMessage.credentialPreview || offerMessage.credentialPreview) {
+    const noProposePreview =
+      !proposalMessage.credentialPreview || proposalMessage.credentialPreview.attributes.length === 0
+    const noOfferPreview = !offerMessage.credentialPreview || offerMessage.credentialPreview.attributes.length === 0
+
+    if (noProposePreview || noOfferPreview) {
```
TimoGlastra, line 956: why is this change needed?

packages/core/src/modules/vc/models/credential/W3cCredentialRecord.ts
```diff
@@ -0,0 +1,58 @@
+import type { TagsBase } from '../../../../storage/BaseRecord'
```
TimoGlastra, line 1: There's still trailing files left (w3c credential record already exists)

packages/core/src/modules/vc/models/credential/W3cCredentialRepository.ts
```diff
@@ -0,0 +1,18 @@
+import { inject, scoped, Lifecycle } from 'tsyringe'
+
+import { EventEmitter } from '../../../../agent/EventEmitter'
+import { InjectionSymbols } from '../../../../constants'
+import { Repository } from '../../../../storage/Repository'
+import { StorageService } from '../../../../storage/StorageService'
+
+import { W3cCredentialRecord } from './W3cCredentialRecord'
```
TimoGlastra, line 8: same here

packages/core/types/jsonld-signatures.ts
```diff
@@ -0,0 +1,23 @@
...
+  suites as JsonLdSuites,
+  purposes as JsonLdPurposes,
+  constants as JsonLdConstants,
+  //@ts-ignore
+} from '@digitalcredentials/jsonld-signatures'
+
+interface Suites {
+  LinkedDataSignature: any
+  LinkedDataProof: any
+}
```
TimoGlastra, line 11: same here

packages/core/types/jsonld.ts
```diff
@@ -0,0 +1,29 @@
...
+import jsonld from '@digitalcredentials/jsonld'
+//@ts-ignore
+import nodeDocumentLoader from '@digitalcredentials/jsonld/lib/documentLoaders/node'
+//@ts-ignore
+import xhrDocumentLoader from '@digitalcredentials/jsonld/lib/documentLoaders/xhr'
+
+interface JsonLd {
+  compact(document: any, context: any, options?: any): any
+  fromRDF(document: any): any
+  frame(document: any, revealDocument: any, options?: any): any
```
TimoGlastra, line 11: same here

packages/core/types/vc.ts
```diff
@@ -0,0 +1,12 @@
+// @ts-ignore
+import vc from '@digitalcredentials/vc'
+
+interface VC {
+  issue(options: any): Promise<Record<string, unknown>>
+  verifyCredential(options: any): Promise<Record<string, unknown>>
+  createPresentation(options: any): Promise<Record<string, unknown>>
+  signPresentation(options: any): Promise<Record<string, unknown>>
```
TimoGlastra, line 8: same here

packages/core/src/plugins/DependencyManager.ts
```diff
@@ -2,6 +2,8 @@ import type { ModulesMap } from '../agent/AgentModules'
 import type { Constructor } from '../utils/mixins'
 import type { DependencyContainer } from 'tsyringe'
 
+import 'reflect-metadata'
```
TimoGlastra, line 5: this shouldn't be needed

```suggestion
```

TimoGlastra (changes requested):
Didn't finish the full review yet, but there's still some things to fix before we can merge

NB-MikeRichardson:
> There's a lot of checks failing @NB-MikeRichardson, could you address those?

I can't see any which are a cause of this PR. Which ones do you think I need to fix?

packages/core/src/modules/credentials/formats/CredentialFormatService.ts
NB-MikeRichardson: Yes, it uses the new deepEqual method

NB-MikeRichardson: The method is public because it is accessed from the FormatService test. Is there a better way to allow that access?

TimoGlastra:
CI is only failing for Node 17/18 for the BBS tests, all other tests should pass. The 0.3.0-pre does not have these same errors so I think it is something to do with your branch specifically

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialOptions.ts
NB-MikeRichardson: Legacy of when we were passing the verificationMethod in the message. 

Removed.

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormat.ts
NB-MikeRichardson: I'm not sure how to resolve this one. The two methods acceptProposal and createOffer are identical but I can't see where the values provided are any different either, otherwise I could pass in a different interface.

Thoughts?

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
NB-MikeRichardson: This would entail a change in the W3cCredentialService (signCredential method), would it not?

Do you want me to go ahead and make that change?

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
NB-MikeRichardson: Again, it is public because there is a test for it.

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
NB-MikeRichardson: (Note that `created `and `proofPurpose `are referred to inside this method so presumably we don't want to take those out).

I propose to remove `challenge`, `domain `and `credentialStatus `which are not used.

NB-MikeRichardson:
The only test failing is the 18.x BBS test as far as I can see.

If there are no further comments can we get this merged?

packages/core/src/modules/credentials/formats/CredentialFormatService.ts
TimoGlastra: I think you can test it as a black box, through another method. So if the method that calls it receives false, it should trhow an error or something. I avoid making methods public for tests, as that now means it's part of the api

TimoGlastra:
Can you fix the DCO?

.gitignore
```diff
@@ -8,5 +8,4 @@ coverage
 .DS_Store
 logs.txt
 logs/
-packages/core/src/__tests__/genesis-von.txn
-lerna-debug.log
\ No newline at end of file
+packages/core/src/__tests__/genesis-von.txn
```
TimoGlastra, line 11:
```suggestion
packages/core/src/__tests__/genesis-von.txn
lerna-debug.log
```

NB-MikeRichardson:
I can do even though none of the commits are actually mine. I’ll do it anyway.

packages/core/src/modules/credentials/formats/CredentialFormatService.ts
TimoGlastra: As mentiond, this is only needed for the JsonLD Credential Format Service, can we move it to that class?

packages/core/src/modules/credentials/formats/CredentialFormatServiceOptions.ts
```diff
@@ -89,6 +93,14 @@ export interface FormatAcceptRequestOptions<CF extends CredentialFormat> {
   offerAttachment?: Attachment
 }
 
+export interface FormatAcceptCredentialOptions {
+  credentialRecord: CredentialExchangeRecord
+  // credentialFormats?: CredentialFormatPayload<[CF], 'acceptRequest'>
```
TimoGlastra, line 98:
```suggestion
```

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
TimoGlastra: this hasn't been resolved

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
TimoGlastra: this hasn't been resolved

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
TimoGlastra: this hasn't been resolved

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
TimoGlastra: Yes we should probably add them later on, but for now it's fine to leave them out I think

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
TimoGlastra: Shouldn't make methods public just for tests. You can just call it with a ts-ignore

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
TimoGlastra: This isn't resovled yet

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
TimoGlastra: We should just do nothing if the proof is an array. If we don't support this at the moment we should throw an error. This can lead to security issues

TimoGlastra:
> I can do even though none of the commits are actually mine. I’ll do it anyway.

I can't merge otherwise 🤷

TimoGlastra:
> If there are no further comments can we get this merged?

There's quite some comments that haven't been resolved yet, some could lead to security issues

genaris:
> > I can do even though none of the commits are actually mine. I’ll do it anyway.
> 
> I can't merge otherwise 🤷 

Regarding this particular issue, that happens regularly even if commits were actually signed off (but the DCO script detects an inconsistency), I noticed that there is a new "Set DCO to pass" button (only available to maintainers). Could this be a solution to these typical problems?

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormat.ts
NB-MikeRichardson: I can't see any way to change this.

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
NB-MikeRichardson: Please take a look at the current method: I have made a couple of minor changes; is it ok now? If not, what do we want to change?

NB-MikeRichardson: I now throw an error if it's an array. That should be ok I think.

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
NB-MikeRichardson: ditto

packages/core/src/modules/credentials/protocol/v2/V2CredentialService.ts
NB-MikeRichardson: Not sure what that is. I have taken it out and re-tested. Seems to be fine.

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
NB-MikeRichardson: I can't find a method `findVerficationMethodByVerificationMethodType`. Where is that?

packages/core/src/modules/credentials/protocol/v2/__tests__/v2credentials-architecture.test.ts
NB-MikeRichardson: There is a test which does this in v2.ldproof.credentials-auto-accet.test.ts

morrieinmaas:
This is urgent (to be merged) as it costs @NB-MikeRichardson a lot of overhead.

packages/core/src/agent/Agent.ts
```diff
@@ -45,7 +45,7 @@ export class Agent<AgentModules extends AgentModulesInput = ModulesMap> extends
   public constructor(options: AgentOptions<AgentModules>, dependencyManager = new DependencyManager()) {
     const agentConfig = new AgentConfig(options.config, options.dependencies)
     const modulesWithDefaultModules = extendModulesWithDefaultModules(agentConfig, options.modules)
-
+    extendModulesWithDefaultModules
```
genaris, line 48: I think this is not used

packages/core/src/modules/dids/domain/key-type/keyDidMapping.ts
```diff
@@ -24,6 +24,11 @@ const keyDidMapping: Record<KeyType, KeyDidMapping> = {
   [KeyType.Bls12381g1g2]: keyDidBls12381g1g2,
 }
 
+export const proofTypeKeyTypeMapping: Record<string, string> = {
+  Ed25519Signature2018: VERIFICATION_METHOD_TYPE_ED25519_VERIFICATION_KEY_2018,
+  BbsBlsSignature2020: VERIFICATION_METHOD_TYPE_BLS12381G2_KEY_2020,
+}
```
TimoGlastra, line 30: There is a `getVerificationMethodTypesByProofType` method on the SignatureSuiteRegistry/W3cCredentialService available. Can we leverage that over this? Otherwise dynamically adding new signature suites won't work

packages/core/tests/helpers.ts
```diff
@@ -25,6 +25,7 @@ import { catchError, filter, map, timeout } from 'rxjs/operators'
 
 import { SubjectInboundTransport } from '../../../tests/transport/SubjectInboundTransport'
 import { SubjectOutboundTransport } from '../../../tests/transport/SubjectOutboundTransport'
+import { BbsModule } from '../../bbs-signatures/src/BbsModule'
```
TimoGlastra, line 28: Why do we need the dependency on bbs here? It's fine for now, but ideally we only depend from bbs on core, not vice versa (also in tests)

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
```diff
@@ -0,0 +1,430 @@
...
+
+    if (!verificationMethod) {
+      throw new AriesFrameworkError('Missing verification method in credential data')
+    }
+    const format = new CredentialFormatSpec({
+      attachId,
+      format: JSONLD_VC,
+    })
+
+    const verifiableCredential = await this.w3cCredentialService.signCredential(agentContext, {
```
TimoGlastra, line 256: We're ignoring some properties here (domain, challenge, credentialStatus) from the `credentialData.options` object here. Ideally we support them, but we can add that in a later PR also. What is important is to throw an error when any of those fields are present in the options that we don't support them. Because this can lead to issuance of credentials that don't match what is negotiated. Having a challenge can be an important aspect of preventing replay-attacks so we need to make sure we're throwing errors if we're not following what the exchagne discussed.

packages/core/tests/helpers.ts
NB-MikeRichardson: I am using the BbsModule here to set up agent modules in setting up e2e credential tests.

Is there some way to avoid this? Maybe we need to move the agent set up out of helper.ts...??

NB-MikeRichardson:
I will get these done asap. Please can you raise an issue with all these small issues that need to be done 'post merge.'

I am happy to work on them.

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
NB-MikeRichardson: > > This is urgent (to be merged) as it costs @NB-MikeRichardson a lot of overhead.
> 
> Yes this has been outstanding for way too long. Sorry on this from my side for reviewing it slowly. I think as discussed before: smaller PRs that are easier to review. As a tip to you in the future Mike, please read the RFC carefully and handle all edge cases before making a PR. This will save me a lot of time in reviewing, and save you a lot of time in doing multiple round trips.

No need to apologise; I appreciate the amount of time you've put into this and I know how busy you are.. And it helps me become a better Aries/AFJ/Typescript developer.

NB-MikeRichardson: Moved this into Issue #1089 for separate PR

https://github.com/hyperledger/aries-framework-javascript/issues/1089

packages/core/src/modules/dids/domain/key-type/keyDidMapping.ts
NB-MikeRichardson: SignCredential takes a single verification method but the above method returns an array of verification method.

I will just use the first one found for now. Is that ok?

PR #798, opened 2022-05-25
fix(credentials): add missing issue credential v1 proposal attributes

NB-MikeRichardson:


packages/core/src/modules/credentials/__tests__/V1CredentialService.propose.offer.test.ts
```diff
@@ -1,7 +1,7 @@
 import type { AgentConfig } from '../../../agent/AgentConfig'
```
TimoGlastra, line 1: It's getting quite crazy with the dots. Should we create a `V1CredentialService.proposal.test.ts` or should we maybe just call it `V1CredentialService.proposal-offer.test.ts` Ideally we remove the dots all together, it's more meant as a file type, while here it's just part of the filename IMO

packages/core/src/modules/credentials/__tests__/V1CredentialService.propose.offer.test.ts
```diff
@@ -126,6 +126,108 @@ describe('CredentialService', () => {
...
+        },
+        '~attach': [
+          {
+            '@id': expect.any(String),
+            'mime-type': 'application/json',
+            data: {
+              base64: expect.any(String),
+            },
+          },
+        ],
```
TimoGlastra, line 226: Where is the attachment for? I don't think the v1 message should have an attachment.

packages/core/src/modules/credentials/__tests__/V2CredentialService.offer.test.ts
```diff
@@ -1,7 +1,7 @@
 import type { AgentConfig } from '../../../agent/AgentConfig'
 import type { ConnectionService } from '../../connections/services/ConnectionService'
 import type { CredentialStateChangedEvent } from '../CredentialEvents'
-import type { OfferCredentialOptions } from '../CredentialsModuleOptions'
+import type { OfferCredentialOptions, ProposeCredentialOptions } from '../CredentialsModuleOptions'
```
TimoGlastra, line 4: unused

packages/core/src/modules/credentials/protocol/v1/V1CredentialService.ts
```diff
@@ -125,10 +125,10 @@ export class V1CredentialService extends CredentialService {
     }
 
     const config: CredentialProposeOptions = {
```
TimoGlastra, line 127: why are we calling the format service here? We don't use the attachment, is it for the linked attachments? I'm not sure

packages/core/src/modules/credentials/__tests__/V1CredentialService.propose.offer.test.ts
NB-MikeRichardson: I have removed the attachment and it passes.

packages/core/src/modules/credentials/protocol/v1/V1CredentialService.ts
NB-MikeRichardson: Yes it looks like the createProposal method in the format service adds the linked attachments to the proposal but it also validates the proposal payload...I assume this is required as it relates to the interoperability issue we hit last week.

TimoGlastra: Ok, could we maybe add a comment explaining this? Future readers of the code can maybe be confused that we aren't using the result

packages/core/src/modules/credentials/__tests__/V1CredentialService.propose.offer.test.ts
NB-MikeRichardson: Files renamed

packages/core/src/modules/credentials/__tests__/V1CredentialServiceProposeOffer.test.ts
```diff
@@ -126,6 +126,99 @@ describe('CredentialService', () => {
...
+            state: CredentialState.ProposalSent,
+          }),
+        },
+      })
+    })
+
+    test('returns credential proposal message', async () => {
+      const { message: credentialProposal } = await credentialService.createProposal(proposeOptions)
+
+      // console.log(credentialProposal.toJSON())
```
TimoGlastra, line 193:
```suggestion
```

packages/core/src/modules/credentials/__tests__/V1CredentialServiceProposeOffer.test.ts
```diff
@@ -126,6 +126,99 @@ describe('CredentialService', () => {
...
+      expect(credentialProposal.toJSON()).toMatchObject({
```
TimoGlastra, line 194: 👍

TimoGlastra (approved):
If you can fix the merge conflicts we can merge, thanks!

PR #801, opened 2022-05-26
refactor: move signatures suites to vc module

TimoGlastra:
This moves the signature suites, the jsonld util and other files to the vc module. This will help with the separation of credential formats we're going to look at soon. It also fixes an issue with type files being places outside of the src directory causing build issues. We didn't catch this because CI was not enabled on the 0.3.0 branch. CI is now enabled which gives us more assurance when merging PRs.

This doesn't change anything about the functionality. It's just relocation, and fixing of type issues.


TimoGlastra:
@karimStekelenburg the CI seems to fail on the BBS derive proof tests. It works for me locally, but as we haven't had CI enabled on this branch it could be this has been failing from the start. Have you tried to run the tests with the node binaries instead of the wasm wrapper? Maybe there's something going wrong there. If you could take a look at this and test it works on linux with the actual binaries that would be great!

blu3beri (approved):

TimoGlastra:
@karimStekelenburg I'm going to merge this PR already because it's a hassle to keep up to date with the refactoring, but the CI issue is still present

karimStekelenburg:
I can see the CI is failing, but I'm not sure about the cause. I see two different errors: [one related to the agent not having an outbound transport](https://github.com/hyperledger/aries-framework-javascript/runs/6665958885?check_suite_focus=true#step:9:369) and many other ones related to the Indy pool genesis transaction.

I'm experiencing the issue related to Indy also when I run the tests locally. Because these errors occur in tests that are unrelated to this PR, I'm not sure where it comes from.

Did the Docker Indy pool we use for testing change recently by any chance?

TimoGlastra:
I'm not talking about the errors being logged (we should probably just turn them off), but about failing tests. If you scroll all the way down in test repo it shows just the errors for the failing test

Issue #647, opened 2022-02-18
Unable to accept credential offer as of 0.2.0-alpha.20


jleach:
As of  commit `c5febf5d` AriesBifold is not able to accept credential offers. This commit is where AFJ was updated:

```
-    "@aries-framework/core": "^0.1.0-alpha.322",
+    "@aries-framework/core": "^0.2.0-alpha.20",
     "@aries-framework/react-hooks": "^0.1.2",
-    "@aries-framework/react-native": "^0.1.0-alpha.322",
+    "@aries-framework/react-native": "^0.2.0-alpha.20",
```

To the latest alpha release. To reproduce the issue locally we run `npm i && cd ios && pod install && cd -` then do a clean and build in Xcode with a release config. The previous commit, `efac5e3` seems to work fine. The error message I see in the Xcode log is:

```
2022-02-18 11:09:35.050327-0800 AriesBifold[6696:1728891] [javascript] 'TRACE: Error retrieving did \'TG2TSLVkqfJ5X5X53MtY5i\' from ledger \'CandyDev\'', '{\n  "error": {\n    "name": "IndyError",\n    "message": "IndySdk: Error Domain=IndyErrorDomain Code=309 \\"(null)\\" UserInfo={message=Error: Item not found on ledger\\n  Caused by: Nym not found\\n, indy_backtrace=}",\n    "indyCode": 309,\n    "indyName": "LedgerNotFound",\n    "indyCurrentErrorJson": null\n  },\n  "did": "TG2TSLVkqfJ5X5X53MtY5i"\n}'
2022-02-18 11:09:35.050904-0800 AriesBifold[6696:1728891] [javascript] 'TRACE: Retrieved did \'TG2TSLVkqfJ5X5X53MtY5i\' from ledger \'BCovrinTest\'', '{\n  "did": "TG2TSLVkqfJ5X5X53MtY5i",\n  "verkey": "FK6puHKtLXdZfyz1bXtkxnyc1qtJcBeZ6jqVsvMgjxow",\n  "role": "101"\n}'
[...]

PR #803, opened 2022-05-26
fix(indy): async ledger connection issues on iOS

TheTreek:
Change `connectToPools` function to sequentially connect to pools. Previously there were [issues on iOS](https://github.com/hyperledger/aries-framework-javascript/issues/647) because of too many open sockets on this function.


TimoGlastra:
How is this PR related to https://github.com/hyperledger/aries-framework-javascript/pull/804? Should we add a note to this code explaining why we're doing it in sequence instead of in parallel? Or is this change unrelated to that change?

TimoGlastra:
Will this also fix the issue with `connectToIndyLedgersOnStartup`?

TheTreek:
> How is this PR related to #804? Should we add a note to this code explaining why we're doing it in sequence instead of in parallel? Or is this change unrelated to that change?

The issue we were having is we tried to connect to all ledgers in parallel. This would cause us to  reach the resource limit very quickly, so instead of doing it in parallel we do it in sequence so we don't use too many resources.

> Will this also fix the issue with `connectToIndyLedgersOnStartup`?

Yes, this is the main goal of this PR.

TimoGlastra:
> The issue we were having is we tried to connect to all ledgers in parallel. This would cause us to reach the resource limit very quickly, so instead of doing it in parallel we do it in sequence so we don't use too many resources

Can we add a comment to the relevant code explaining this? I'm worried this will get lost and 'fixed' to parallel opening of pools in the future again.

TimoGlastra:
> Yes, this is the main goal of this PR

That is really great to hear!

TheTreek:
> > The issue we were having is we tried to connect to all ledgers in parallel. This would cause us to reach the resource limit very quickly, so instead of doing it in parallel we do it in sequence so we don't use too many resources
> 
> Can we add a comment to the relevant code explaining this? I'm worried this will get lost and 'fixed' to parallel opening of pools in the future again.

Done!

TimoGlastra:
Thansk @TheTreek, if you can update your branch I can merge!

TimoGlastra (approved):

TimoGlastra:
Sorry it's out of date again :( Can you update again?

PR #806, opened 2022-05-27
fix: default credentials to [] to fix broken test

NB-MikeRichardson:
Signed-off-by: Mike Richardson <mike.richardson@ontario.ca>


packages/core/src/storage/migration/__tests__/__snapshots__/0.1.test.ts.snap
```diff
@@ -2488,443 +2488,3 @@ Object {
   },
 }
 `;
-
```
TimoGlastra, line 2491: These shouldn't be removed

packages/core/src/storage/migration/__tests__/__snapshots__/0.1.test.ts.snap
```diff
@@ -2893,6 +2893,2796 @@ Object {
       "threadId": "e9aeea8f-2c7a-4fd0-9353-f8b5b76094e7",
     },
   },
+  "STORAGE_VERSION_RECORD_ID": Object {
```
TimoGlastra, line 2896: I'm not sure why there are so much changes? It should only add a few new keys.....

TimoGlastra (changes requested):

Issue #807, opened 2022-05-27
[alpha.88+] Websocket connection to mediator not working


Iskander508:
@KolbyRKunz, @TimoGlastra 

It seems this PR broke our setup: https://github.com/hyperledger/aries-framework-javascript/pull/787
We cannot create a connection with latest alpha versions in a react-native application, using websocket connection to an aca-py mediator.

I have digged a bit deeper and the problem seems to be on this line:
https://github.com/Indicio-tech/aries-framework-javascript/blob/f6d70b92ded4070d310e14c9775ad0a4ebef92ec/packages/core/src/modules/routing/RecipientModule.ts#L176
The `mediator` object doesn't have the `pickupStrategy` filled, so `mediator.pickupStrategy === undefined`.
Is it our mistake (passing wrong configuration somewhere) or a bug?

It is working fine with version alpha.87.

our agent config in the app:
```
const agentConfig: InitConfig = {
...
  mediatorConnectionsInvite: mediatorUrl,
  autoAcceptMediationRequests: true,
  mediatorPickupStrategy: MediatorPickupStrategy.Implicit,
};
```

TimoGlastra:
Thanks for raising this issue @Iskander508. I've made a quick fix in this PR: https://github.com/hyperledger/aries-framework-javascript/pull/808

Will look soon to make the code a bit more robust and add test cases for these cases. The problem is that we assumed the mediation record would always store the pickup strategy, but in your case you're using the agent wide config parameter.

Iskander508:
@TimoGlastra: I think I have found the issue in the code. You can check the https://github.com/hyperledger/aries-framework-javascript/pull/809

KolbyRKunz:
Sorry for the confusion, I just have a follow up question about how this might affect things in the future. It might not even be possible but, if for some reason the mediator and the agent wide config parameter were somehow set to different strategies wouldn't we want to make sure those got synced and which one should we say is the source of truth in this instance? I am not sure if that is even possible or not but if one can be undefined while the other is defined it begs the question I think.

PR #809, opened 2022-05-27
fix(routing): updating mediator pickupStrategy

Iskander508:
Fixed bug with updating mediator pickupStrategy

Fixes https://github.com/hyperledger/aries-framework-javascript/issues/807

### DIFF
--- a/packages/core/src/modules/routing/RecipientModule.ts
+++ b/packages/core/src/modules/routing/RecipientModule.ts
@@ -249,7 +249,9 @@ export class RecipientModule {
           ? MediatorPickupStrategy.PickUpV1
           : MediatorPickupStrategy.Implicit
       }
+    }
 
+    if (mediator.pickupStrategy !== mediatorPickupStrategy) {
       // Store the result so it can be reused next time
       mediator.pickupStrategy = mediatorPickupStrategy
       await this.mediationRepository.update(mediator)


TimoGlastra:
Your approach is cleaner than my approach so inclined to merge this one. However it does have some impact when you change configurations.

In your approach the following happens:
1. Agent config specifies mediator pickup strategy
2. agent config value is stored in mediator record
3. Next time you use it the mediator record value will be used, regardless whether the agent config value is changed

in my approach the following happens:
1. Agent config specifies mediator pickup strategy
2. agent config value **is not** stored in mediator record
3. Next time you use it the agent config value will be used -> this matters when you change the agent config value

Any thoughts on which of those would be preferred here? For others protocols we normally use the second appraoch (so not storing a global config in a specific record). If I start the agent with autoAcceptConnections enabled, it won't actually store that in the connectino record. If I restart the agent next time with autoAcceptConnections false it will take it from the agent config, and doesn't have stored the autoAcceptConnection true on the connection record

Iskander508:
> Any thoughts on which of those would be preferred here?

No preference TBH, both should fix our case. If there's some pattern used in other places, it might be better to follow it, keeping better consistency and readability.

TimoGlastra:
I merged #808 to keep it consistent with the rest of the framework. Could you check with the latest alpha if your problem has been resolved?

PR #813, opened 2022-05-30
fix(core): allow JSON as input for indy attributes

blu3beri:
Right now, you have to create an instance of the class `CredentialPreviewAttribute` in the public API, like so:

```typescript
const attribute = new CredentialPreview({name: 'foo', value: 'bar'})
```

This PR will still allow for this but it is extended to also include:

```typescript
const attribute = { name: 'foo', value: 'bar' }
```

And it will tranform this into a class instance.

The reason for this PR is that is very weird for a public API to accept a class that only adds a `toJSON` method.

(I did not pickup the proofs module, as V2 is not merged and we can include it in there or I will pick it up after the merge).


network/indy-pool.dockerfile
```diff
@@ -105,4 +105,4 @@ COPY network/indy-cli-config.json /etc/indy/indy-cli-config.json
 
 EXPOSE 9701 9702 9703 9704 9705 9706 9707 9708
 
-CMD ["/usr/bin/supervisord"]
\ No newline at end of file
+CMD ["/usr/bin/supervisord"]
```
TimoGlastra, line 108: This seems unrelated?

```suggestion
CMD ["/usr/bin/supervisord"] 
```

TimoGlastra (approved):

Issue #812, opened 2022-05-30
OOB invitations using public did not supported


Iskander508:
@TimoGlastra maybe just a question: If we try to use the out-of-band invitations (created in an aca-py instance) and using the public did, the AFJ client cannot receive such invitation right now.

The `agent.oob.receiveInvitationFromUrl` currently fails with: `TypeError: undefined is not an object (evaluating 'iter[Symbol.iterator]`
We've identified the failure coming from this line: https://github.com/hyperledger/aries-framework-javascript/blob/main/packages/core/src/modules/oob/messages/OutOfBandInvitation.ts#L108

When adjusting the code to overcome the failure, we get:
`AriesFrameworkError: Dids are not currently supported in out-of-band invitation services attribute.`

Is it planned to add the support in the 0.2.0 release? I couldn't find any mention in https://github.com/hyperledger/aries-framework-javascript/issues/764
Thank you

TimoGlastra:
Ah we should update the code snippet to the following to avoid the error posted above:

```ts
  public getRecipientKeys(): Key[] {
    return this.services
      .filter((s): s is OutOfBandDidCommService => typeof s !== 'string' && !(s instanceof String))
      .map((s) => s.recipientKeys)
      .reduce((acc, curr) => [...acc, ...curr], [])
      .map((didKey) => DidKey.fromDid(didKey).key)
  }
```

But you're correct that receiving invitations from public dids are not supported yet. It's currently not top of my priority to work on this feature. @JamesKEbert mentioned that he'd be interested in working on this, but not sure of the status of this.

PR #814, opened 2022-05-30
fix(oob): check service is string instance

TimoGlastra:

Adds a check whether the service is instanceof String as mentioned in  #812. this is needed due to the class-validator logic sadly.

### DIFF
--- a/packages/core/src/modules/oob/messages/OutOfBandInvitation.ts
+++ b/packages/core/src/modules/oob/messages/OutOfBandInvitation.ts
@@ -103,7 +103,7 @@ export class OutOfBandInvitation extends AgentMessage {
   // TODO: this only takes into account inline didcomm services, won't work for public dids
   public getRecipientKeys(): Key[] {
     return this.services
-      .filter((s): s is OutOfBandDidCommService => typeof s !== 'string')
+      .filter((s): s is OutOfBandDidCommService => typeof s !== 'string' && !(s instanceof String))
       .map((s) => s.recipientKeys)
       .reduce((acc, curr) => [...acc, ...curr], [])
       .map((didKey) => DidKey.fromDid(didKey).key)


packages/core/src/modules/oob/messages/OutOfBandInvitation.ts
```diff
@@ -103,7 +103,7 @@ export class OutOfBandInvitation extends AgentMessage {
   // TODO: this only takes into account inline didcomm services, won't work for public dids
   public getRecipientKeys(): Key[] {
     return this.services
-      .filter((s): s is OutOfBandDidCommService => typeof s !== 'string')
+      .filter((s): s is OutOfBandDidCommService => typeof s !== 'string' && !(s instanceof String))
```
blu3beri, line 106: Any reason why the `s is OutOfBandDidCommService` is required?

blu3beri (approved):

packages/core/src/modules/oob/messages/OutOfBandInvitation.ts
```diff
@@ -103,7 +103,7 @@ export class OutOfBandInvitation extends AgentMessage {
...
       .map((s) => s.recipientKeys)
```
Iskander508, line 107: Just a comment. The error mentioned in https://github.com/hyperledger/aries-framework-javascript/issues/812 actually came from the `...curr`, because the previous map returns `[undefined]` in case of public did is used in the invitation.

So I'd suggest to modify the second mapping something like this:
```suggestion
      .map((s) => s.recipientKeys || [])
```

TimoGlastra: Hmm the filter method should filter out all non-string values, so it shouldn't be possible there is a type undefined in the array. It think by adding this to the filter it won't have the undefined anymore.

Iskander508: `s.recipientKeys` was undefined. The json parser converts the string did into some weird object, maybe that will get resolved in the future though.

PR #816, opened 2022-05-31
fix(credentials): default for credentials in exchange record

NB-MikeRichardson:


packages/core/src/modules/credentials/repository/CredentialExchangeRecord.ts
```diff
@@ -61,7 +61,7 @@ export class CredentialExchangeRecord extends BaseRecord<
   public revocationNotification?: RevocationNotification
   public errorMessage?: string
   public protocolVersion!: CredentialProtocolVersion
-  public credentials!: CredentialRecordBinding[]
+  public credentials: CredentialRecordBinding[] = [] // default in case constructor not used (as can happen in tests)
```
TimoGlastra, line 64:
```suggestion
  public credentials: CredentialRecordBinding[] = []
```

TimoGlastra (approved):

TimoGlastra:
@NB-MikeRichardson it looks like the tests are failing now but only because the `credentials` is added to the snapshot. This is an expected change. If you can try to run the tests with `-u` again locally it should work.

NB-MikeRichardson:
The tests are still failing. No idea why.

TimoGlastra (approved):

PR #819, opened 2022-05-31
refactor(credentials): separate offer interfaces

NB-MikeRichardson:
This is to avoid having connection and connection id in the same object and be able to remove protocol version
from the accept proposal / create offer functionality


packages/core/src/modules/credentials/CredentialsModule.ts
```diff
@@ -252,7 +252,7 @@ export class CredentialsModule implements CredentialsModule {
           'Invalid DidDocument: Missing verification method with type Ed25519VerificationKey2018 to use as indy holder did'
         )
       }
-      const indyDid = getIndyDidFromVerficationMethod(verificationMethod)
+      const indyDid = getIndyDidFromVerificationMethod(verificationMethod)
```
TimoGlastra, line 255: good catch 👍

packages/core/src/modules/credentials/protocol/v2/CredentialMessageBuilder.ts
```diff
@@ -243,8 +243,10 @@ export class CredentialMessageBuilder {
     // Construct v2 offer message
     const message = new V2OfferCredentialMessage(messageProps)
 
+    const id = connection ? connection.id : undefined
+
     const recordProps: CredentialExchangeRecordProps = {
-      connectionId,
+      connectionId: id,
```
TimoGlastra, line 249: I think we use optional chaining here

```suggestion
      connectionId: connection?.id,
```

packages/core/src/modules/credentials/protocol/v2/CredentialMessageBuilder.ts
```diff
@@ -243,8 +243,10 @@ export class CredentialMessageBuilder {
     // Construct v2 offer message
     const message = new V2OfferCredentialMessage(messageProps)
 
+    const id = connection ? connection.id : undefined
```
TimoGlastra, line 246:
```suggestion
```

TimoGlastra (commented):
some small nits

TimoGlastra (approved):
Nice 👍

PR #826, opened 2022-06-01
fix: process ws return route messages serially

TimoGlastra:
We were already processing messages serially for pickup v1 and v2. This pr updates to also use this for websocket inbound transport. This can help prevent some race conditions and is consistent with the other pickup approaches


TimoGlastra:
@amanji This should fix the issue with messages being processed in parallel

packages/core/src/transport/WsOutboundTransport.ts
```diff
@@ -1,3 +1,4 @@
+import type { AgentMessageReceivedEvent } from '..'
```
blu3beri, line 1: Could you update the import to not use '..'

PR #828, opened 2022-06-02
fix(oob): export messages in oob

blu3beri:
Noticed that the `OutofBandInvitation` was not exported to the public and it can be quite useful to have

- Exported the `messages` folder inside the `oob` module
- There might be more items we want to export from here, but I am not too sure about which.


TimoGlastra:
@blu3beri I've started to use modules as topics to make the commit topics more specific (so oob,credentials,routing,etc.. instead of core)

Issue #827, opened 2022-06-02
Accept Proposal should reuse credential definition id from the Proposal


NB-MikeRichardson:
accepting a proposal should take the credential definition id from the proposal to send a credential offer (if availalbe on the proposal). This functionality was there for v1 (and still is: https://github.com/hyperledger/aries-framework-javascript/blob/main/packages/core/src/modules/credentials/protocol/v1/V1CredentialService.ts#L204-L205), but itsn't present in v2. We should probably extract this from the proposal in the indy credential format service so we can reuse it across the v1 and v2 implementation. Currently throws an error that it is missing the credential definition id. It also means the user has to retrieve the message and extract the credential definition from the proposal themselves.
Same for the attributes, it shouldn't be required to pass the attributes when accepting a proposal as we can take them from the proposal (but should be optional for the case where there was no proposal preview and thus no attributes)

NB-MikeRichardson:
As per comment from @swcurran we need to ensure that this is valid, ie if the cred-def-id in the proposal is not one the issuer owns, they can't issue such a credential, so that check is a minimum.

TimoGlastra:
This is fixed

PR #829, opened 2022-06-02
refactor: Reuse cred def id and attributes from Proposal message in Indy Format Service

NB-MikeRichardson:
See Issue #837 https://github.com/hyperledger/aries-framework-javascript/issues/827

### DIFF
--- a/packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts
+++ b/packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts
@@ -78,7 +78,10 @@ export class IndyCredentialFormatService extends CredentialFormatService {
    * @returns object containing associated attachment, formats and filtersAttach elements
    *
    */
-  public async createProposal(options: ProposeCredentialOptions): Promise<FormatServiceProposeAttachmentFormats> {
+  public async createProposal(
+    options: ProposeCredentialOptions,
+    credentialRecord: CredentialExchangeRecord
+  ): Promise<FormatServiceProposeAttachmentFormats> {
     const formats: CredentialFormatSpec = {
       attachId: this.generateId(),
       format: 'hlindy/cred-filter@v2.0',
@@ -103,6 +106,10 @@ export class IndyCredentialFormatService extends CredentialFormatService {
 
     const { previewWithAttachments } = this.getCredentialLinkedAttachments(options)
 
+    credentialRecord.metadata.set(CredentialMetadataKeys.IndyCredential, {
+      schemaId: proposal.schemaId,
+      credentialDefinitionId: proposal.credentialDefinitionId,
+    })
     return { format: formats, attachment, preview: previewWithAttachments }
   }
 
@@ -125,22 +132,6 @@ export class IndyCredentialFormatService extends CredentialFormatService {
       options.credentialFormats.indy?.credentialDefinitionId ?? credProposal?.credentialDefinitionId
 
     options.credentialFormats.indy.attributes = credentialRecord.credentialAttributes ?? []
-  }
-
-  public async setProposalMetaData(
-    options: ServiceAcceptProposalOptions,
-    credentialRecord: CredentialExchangeRecord
-  ): Promise<void> {
-    const credProposalJson = options.proposalAttachment?.getDataAsJson<CredPropose>()
-    if (!credProposalJson) {
-      throw new AriesFrameworkError('Missing indy credential proposal data payload')
-    }
-    const credProposal = JsonTransformer.fromJSON(credProposalJson, CredPropose)
-
-    credentialRecord.metadata.set(CredentialMetadataKeys.IndyCredential, {
-      schemaId: credProposal.schemaId,
-      credentialDefinitionId: credProposal.credentialDefinitionId,
-    })
   }
   /**
    * Create a {@link AttachmentFormats} object dependent on the message type.
--- a/packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts
+++ b/packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts
@@ -117,21 +117,31 @@ export class IndyCredentialFormatService extends CredentialFormatService {
     const credProposal = JsonTransformer.fromJSON(credProposalJson, CredPropose)
     await MessageValidator.validate(credProposal)
 
-    if (credProposal.credentialDefinitionId) {
-      options.credentialFormats = {
-        indy: {
-          credentialDefinitionId: credProposal?.credentialDefinitionId,
-          attributes: [],
-        },
-      }
+    options.credentialFormats.indy = {
+      attributes: [],
     }
+    // reuse credential definition id (and attributes) for purposes of sending offer as response
+    options.credentialFormats.indy.credentialDefinitionId =
+      options.credentialFormats.indy?.credentialDefinitionId ?? credProposal?.credentialDefinitionId
+
+    options.credentialFormats.indy.attributes = credentialRecord.credentialAttributes ?? []
+  }
+
+  public async setProposalMetaData(
+    options: ServiceAcceptProposalOptions,
+    credentialRecord: CredentialExchangeRecord
+  ): Promise<void> {
+    const credProposalJson = options.proposalAttachment?.getDataAsJson<CredPropose>()
+    if (!credProposalJson) {
+      throw new AriesFrameworkError('Missing indy credential proposal data payload')
+    }
+    const credProposal = JsonTransformer.fromJSON(credProposalJson, CredPropose)
 
     credentialRecord.metadata.set(CredentialMetadataKeys.IndyCredential, {
       schemaId: credProposal.schemaId,
       credentialDefinitionId: credProposal.credentialDefinitionId,
     })
   }
-
   /**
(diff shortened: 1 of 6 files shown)


packages/core/src/modules/credentials/protocol/v1/V1CredentialService.ts
```diff
@@ -140,8 +140,19 @@ export class V1CredentialService extends CredentialService {
 
     const options = { ...config }
 
+    // Create record
+    const credentialRecord = new CredentialExchangeRecord({
```
blu3beri, line 144: Is this really the best way to create a `CredentialExchangeRecord`? I am not the biggest fan of passing empty strings, such as with `threadId`, and passing empty lists, like with `credentials` and credentialAttributes`. These should default to these values and do not have to be passed in.

packages/core/src/modules/credentials/protocol/v2/__tests__/v2credentials-architecture.test.ts
```diff
@@ -85,7 +91,11 @@ describe('V2 Credential Architecture', () => {
       const version: CredentialProtocolVersion = CredentialProtocolVersion.V2
       const service: CredentialService = api.getService(version)
       const formatService: CredentialFormatService = service.getFormatService(CredentialFormatType.Indy)
-      const { format: formats, attachment: filtersAttach } = await formatService.createProposal(proposal)
+
+      const { format: formats, attachment: filtersAttach } = await formatService.createProposal(
```
blu3beri, line 95: I think it is a bit cleaner to just use the same name as the return value.

It also confuses me a bit if the return value is `format` as a signular and we bind it to a variable as `formats`, plural. Which is it?

```suggestion
      const { format, attachment } = await formatService.createProposal(
```

packages/core/src/modules/credentials/protocol/v2/__tests__/v2credentials-architecture.test.ts
```diff
@@ -95,7 +105,10 @@ describe('V2 Credential Architecture', () => {
       const version: CredentialProtocolVersion = CredentialProtocolVersion.V2
       const service: CredentialService = api.getService(version)
       const formatService: CredentialFormatService = service.getFormatService(CredentialFormatType.Indy)
-      const { format: formats, attachment: filtersAttach } = await formatService.createProposal(proposal)
+      const { format: formats, attachment: filtersAttach } = await formatService.createProposal(
```
blu3beri, line 108: Same here as my comment above.

packages/core/src/modules/credentials/protocol/v2/__tests__/v2credentials-architecture.test.ts
```diff
@@ -9,13 +9,19 @@ import type { CredentialService } from '../../../services/CredentialService'
...
+import { CredentialState } from '../../../CredentialState'
 import { CredentialsModule } from '../../../CredentialsModule'
 import { CredentialFormatType } from '../../../CredentialsModuleOptions'
+import { CredentialExchangeRecord } from '../../../repository'
 import { V1CredentialPreview } from '../../v1/V1CredentialPreview'
 import { CredentialMessageBuilder } from '../CredentialMessageBuilder'
 
 const { config, agentDependencies: dependencies } = getBaseConfig('Format Service Test')
-
+const credentialRecord = new CredentialExchangeRecord({
```
blu3beri, line 20: Same here as my comment on instantiating credentialexchangerecords.

blu3beri (changes requested):

packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts
```diff
@@ -117,21 +124,15 @@ export class IndyCredentialFormatService extends CredentialFormatService {
...
     await MessageValidator.validate(credProposal)
 
-    if (credProposal.credentialDefinitionId) {
-      options.credentialFormats = {
-        indy: {
-          credentialDefinitionId: credProposal?.credentialDefinitionId,
-          attributes: [],
-        },
-      }
+    options.credentialFormats.indy = {
```
TimoGlastra, line 127: We shouldn't modify input objects like this. The only exception is the credential record

packages/core/src/modules/credentials/protocol/v2/CredentialMessageBuilder.ts
```diff
@@ -56,8 +56,19 @@ export class CredentialMessageBuilder {
...
+    const props: CredentialExchangeRecordProps = {
+      connectionId: proposal.connectionId,
+      state: CredentialState.ProposalSent,
+      autoAcceptCredential: proposal?.autoAcceptCredential,
+      protocolVersion: CredentialProtocolVersion.V2,
+      credentials: [],
+    }
+
+    // Create the v2 record
+    const credentialRecord = new CredentialExchangeRecord(props)
```
TimoGlastra, line 68: any reason over doing this? Saves an assignment, making the overall code more readable IMO

```suggestion
    // Create the v2 record
    const credentialRecord = new CredentialExchangeRecord({
      connectionId: proposal.connectionId,
      state: CredentialState.ProposalSent,
      autoAcceptCredential: proposal?.autoAcceptCredential,
      protocolVersion: CredentialProtocolVersion.V2,
      credentials: [],
    })
```

packages/core/src/modules/credentials/protocol/v2/V2CredentialService.ts
```diff
@@ -143,13 +143,15 @@ export class V2CredentialService extends CredentialService {
...
         credentialRecordId: credentialRecord.id,
         credentialFormats: {},
       }
-      options.proposalAttachment = format.getAttachment(proposalMessage.formats, proposalMessage.messageAttachment)
-      await format.processProposal(options, credentialRecord)
+      options.proposalAttachment = formatService.getAttachment(
+        proposalMessage.formats,
+        proposalMessage.messageAttachment
+      )
     }
```
TimoGlastra, line 155: this for loop is not doing anything anymore, can we remove it?

packages/core/src/modules/credentials/repository/CredentialExchangeRecord.ts
```diff
@@ -80,11 +80,11 @@ export class CredentialExchangeRecord extends BaseRecord<
       this.createdAt = props.createdAt ?? new Date()
       this.state = props.state
       this.connectionId = props.connectionId
-      this.threadId = props.threadId
+      this.threadId = props.threadId ?? ''
```
TimoGlastra, line 83:
```suggestion
      this.threadId = props.threadId
```

packages/core/src/modules/credentials/repository/CredentialExchangeRecord.ts
```diff
@@ -22,7 +22,7 @@ export interface CredentialExchangeRecordProps {
   createdAt?: Date
   state: CredentialState
   connectionId?: string
-  threadId: string
+  threadId?: string
```
TimoGlastra, line 25: Making this optional has some downsides. What if we create the threadId beforehand for the message?

So:
1. create the message (without any attachments, just empty proposal message)
2. create the credential record (with the threadId from the previous step)
3. call the format services and add the attachment formats to the message

Or:
1. generate an uuid
2. create the credential record with threadId set to uuid from previous step
3. call the format services and get the attachment formats
4. create the proposal message, passing the uuid from step 1 as the message id

My preference would go to the first appraoch, but if for some reason that's not possible the second appraoch still has my preference to making this property optional

packages/core/src/modules/credentials/repository/CredentialExchangeRecord.ts
```diff
@@ -80,11 +80,11 @@ export class CredentialExchangeRecord extends BaseRecord<
       this.createdAt = props.createdAt ?? new Date()
       this.state = props.state
       this.connectionId = props.connectionId
-      this.threadId = props.threadId
+      this.threadId = props.threadId ?? ''
       this.protocolVersion = props.protocolVersion
       this._tags = props.tags ?? {}
 
-      this.credentialAttributes = props.credentialAttributes
+      this.credentialAttributes = props.credentialAttributes ?? []
```
TimoGlastra, line 87:
```suggestion
      this.credentialAttributes = props.credentialAttributes
```

TimoGlastra (changes requested):

packages/core/src/modules/credentials/protocol/v2/CredentialMessageBuilder.ts
NB-MikeRichardson: I always prefer to create a const for the props; just a personal preference. But I will make the change, not a big deal.

TimoGlastra: In that case leave it be, it's one personal preference over the other

packages/core/src/modules/credentials/repository/CredentialExchangeRecord.ts
NB-MikeRichardson: note that threadId is now optional in the props, which was a change suggested by @blu3beri.

(it has to be if we create the credential record before the message for the createProposal)

TimoGlastra:
These changes have been resolved as part of #841

PR #830, opened 2022-06-02
feat: add validation to JSON transformer

### LINKED ISSUE #742: Add validation to `JsonTransformer.fromJson()`
Currently, the only place were JSON is being validated is inside the `MessageReceiver`. However, there are many cases where we manually transform JSON into class instances. This currently requires us to manually call the `class-validator`, which is easy to forget.

One way to make this process less error prone is to add this validation step to the `JsonTransformer.fromJson()` method as default behaviour. Because this adds extra processing time and not all use cases require this, we should add a flag to skip this validation step if desired, e.g. `JsonTransformer.fromJson(json, Class, { validate: false })`.

morrieinmaas:
closes #742 


packages/core/src/modules/proofs/services/ProofService.ts
```diff
@@ -905,7 +907,7 @@ export class ProofService {
    *
    */
   public async verifyProof(proofJson: IndyProof, proofRequest: ProofRequest): Promise<boolean> {
-    const proof = JsonTransformer.fromJSON(proofJson, PartialProof)
+    const proof = await JsonTransformer.fromJSON(proofJson, PartialProof, { validate: true })
```
blu3beri, line 910: I see you added `{validate: true}` to every function call, but it is also the default value. I'd suggest to rely on the default value to keep it a bit more terse.

packages/core/src/utils/JsonTransformer.ts
```diff
@@ -1,23 +1,61 @@
...
+    Class: { new (...args: any[]): T },
+    { validate = true }: Validate
+  ): Promise<T> {
+    if (!validate) {
+      return plainToInstance(Class, json, { exposeDefaultValues: true })
+    } else {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      const plainInstance = plainToInstance(Class, json, { exposeDefaultValues: true }) as any
+      try {
+        await MessageValidator.validate(plainInstance)
```
blu3beri, line 34: I dont think this is async, I might be wrong though.

packages/core/src/utils/__tests__/JsonTransformer.test.ts
```diff
@@ -67,18 +69,20 @@ describe('JsonTransformer', () => {
...
-      const didDocument = JsonTransformer.deserialize(didDocumentString, DidDocument)
+    //   const didDocument = await JsonTransformer.deserialize(didDocumentString, DidDocument, { validate: true })
 
-      const keyAgreement = didDocument.keyAgreement ?? []
+    //   const keyAgreement = didDocument.keyAgreement ?? []
 
-      expect(keyAgreement[0]).toBeInstanceOf(VerificationMethod)
-    })
+    //   await expect(async () => keyAgreement[0]).resolves.toBeInstanceOf(VerificationMethod)
+    // })
```
karimStekelenburg, line 86: Not sure if you need this still as the PR is WIP, but be aware of commented code :)

packages/core/src/utils/JsonTransformer.ts
karimStekelenburg: `MessageValidator.validate` indeed isn't async, so we can probably keep all `JsonTransformer` methods sync :)

TimoGlastra: It returns a promise so it is async... but maybe we can update it to use `validateSync` instead: https://github.com/typestack/class-validator#synchronous-validation

I agree it would be nice to keep this method sync.

Only downside is that it will ignore all async validators, but we don't have any async validators so we're probably good

morrieinmaas:
Thanks for all the feedback!

packages/core/src/utils/JsonTransformer.ts
morrieinmaas: ALso agree that keeping this sync (non-async) is preferable. also for people already using afj so they don't have to change their code

morrieinmaas: I'll have a go at using the sync validation from the suggested link

packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -286,11 +285,6 @@ export class ConnectionService {
     }
 
     const connection = JsonTransformer.fromJSON(connectionJson, Connection)
-    try {
```
morrieinmaas, line 289: Validation is in fromJSON now. Not sure we want to (re)raise the error here. Left out for now

packages/core/src/agent/MessageReceiver.ts
```diff
@@ -169,7 +169,8 @@ export class MessageReceiver {
     try {
       message = await this.transformMessage(plaintextMessage)
       await this.validateMessage(message)
-    } catch (error) {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
TimoGlastra, line 172: I don't think this is needed (it wasn't needed before this PR)

packages/core/src/agent/MessageReceiver.ts
```diff
@@ -209,7 +210,7 @@ export class MessageReceiver {
     }
 
     // Cast the plain JSON object to specific instance of Message extended from AgentMessage
-    return JsonTransformer.fromJSON(message, MessageClass)
+    return JsonTransformer.fromJSON(message, MessageClass, { validate: false })
```
TimoGlastra, line 213: Do you get any errors when calling this, or did you skip it because we're already validating it above? Maybe we can combine it with the validation above (so only a call to the transformer) and catch the validation error that is throw from the transformer?

packages/core/src/agent/__tests__/AgentMessage.test.ts
```diff
@@ -65,7 +65,7 @@ describe('AgentMessage', () => {
         '@type': 'https://didcomm.org/fake-protocol/2.0/message',
       }
 
-      const message = JsonTransformer.fromJSON(json, CustomProtocolMessage)
+      const message = JsonTransformer.fromJSON(json, CustomProtocolMessage, { validate: false })
 
       await expect(MessageValidator.validate(message)).rejects.toMatchObject([
         {
```
TimoGlastra, line 71: Maybe we can call it like this (to stay consistent with how you updated it with the other tests):

```suggestion
      await expect(JsonTransformer.fromJSON(json, CustomProtocolMessage)).rejects.toMatchObject([
        {
```

packages/core/src/decorators/ack/AckDecorator.test.ts
```diff
@@ -24,7 +24,7 @@ describe('Decorators | AckDecoratorExtension', () => {
   })
 
   test('transforms Json to AckDecorator class', () => {
-    const transformed = JsonTransformer.fromJSON({ '~please_ack': {} }, TestMessage)
+    const transformed = JsonTransformer.fromJSON({ '~please_ack': {} }, TestMessage, { validate: false })
```
TimoGlastra, line 27: What kind of validation errors do you get? Maybe we can add the required fields so we have validation included with the tests. I think adding an `@type` and `@id` should be enough (+ the required fields that TestMessage defines)

packages/core/src/decorators/transport/TransportDecorator.test.ts
```diff
@@ -4,7 +4,7 @@ import { MessageValidator } from '../../utils/MessageValidator'
 import { TransportDecorator, ReturnRouteTypes } from './TransportDecorator'
 
 const validTransport = (transportJson: Record<string, unknown>) =>
-  MessageValidator.validate(JsonTransformer.fromJSON(transportJson, TransportDecorator))
+  MessageValidator.validate(JsonTransformer.fromJSON(transportJson, TransportDecorator, { validate: false }))
```
TimoGlastra, line 7: Same here, if we keep validation would be nice

packages/core/src/decorators/service/ServiceDecorator.test.ts
```diff
@@ -25,7 +25,7 @@ describe('Decorators | ServiceDecoratorExtension', () => {
   })
 
   test('transforms Json to ServiceDecorator class', () => {
-    const transformed = JsonTransformer.fromJSON({ '~service': service }, TestMessage)
+    const transformed = JsonTransformer.fromJSON({ '~service': service }, TestMessage, { validate: false })
```
TimoGlastra, line 28: same here

packages/core/src/error/ValidationErrorUtils.ts
```diff
@@ -0,0 +1,14 @@
+import { ValidationError } from 'class-validator'
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function isValidationErrorArray(e: any): boolean {
```
TimoGlastra, line 4: Can you add some tests for this?

packages/core/src/modules/connections/DidExchangeProtocol.ts
```diff
@@ -479,7 +479,7 @@ export class DidExchangeProtocol {
     const payload = JsonEncoder.toBuffer(json)
     const { isValid, signerVerkeys } = await this.jwsService.verifyJws({ jws, payload })
 
-    const didDocument = JsonTransformer.fromJSON(json, DidDocument)
+    const didDocument = JsonTransformer.fromJSON(json, DidDocument, { validate: false })
```
TimoGlastra, line 482: What kind of errors do we get here, as this isn't a test file we should try to fix the issues instead of disabling validation

packages/core/src/modules/connections/messages/ConnectionInvitationMessage.ts
```diff
@@ -127,9 +126,7 @@ export class ConnectionInvitationMessage extends AgentMessage {
 
     if (typeof encodedInvitation === 'string') {
       const invitationJson = JsonEncoder.fromBase64(encodedInvitation)
-      const invitation = JsonTransformer.fromJSON(invitationJson, ConnectionInvitationMessage)
-
-      await MessageValidator.validate(invitation)
+      const invitation = JsonTransformer.fromJSON(invitationJson, ConnectionInvitationMessage, { validate: false })
```
TimoGlastra, line 129: This was validating before and now it has validate false. Can we keep validation enabled here?

TimoGlastra: Also, I think we can make this method sync now!

packages/core/src/modules/connections/services/ConnectionService.ts
TimoGlastra: You've disabled validation, so we're losing the validation here now

packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts
```diff
@@ -114,7 +114,7 @@ export class IndyCredentialFormatService extends CredentialFormatService {
     if (!credProposalJson) {
       throw new AriesFrameworkError('Missing indy credential proposal data payload')
     }
-    const credProposal = JsonTransformer.fromJSON(credProposalJson, CredPropose)
+    const credProposal = JsonTransformer.fromJSON(credProposalJson, CredPropose, { validate: false })
```
TimoGlastra, line 117: Shouldn't disable validation in non-test files

packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts
```diff
@@ -523,7 +523,7 @@ export class IndyCredentialFormatService extends CredentialFormatService {
   private areProposalAndOfferDefinitionIdEqual(proposalAttachment?: Attachment, offerAttachment?: Attachment) {
     const credOffer = offerAttachment?.getDataAsJson<CredOffer>()
     let credPropose = proposalAttachment?.getDataAsJson<CredPropose>()
-    credPropose = JsonTransformer.fromJSON(credPropose, CredPropose)
+    credPropose = JsonTransformer.fromJSON(credPropose, CredPropose, { validate: false })
```
TimoGlastra, line 526: same

packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts
```diff
@@ -571,7 +571,7 @@ export class IndyCredentialFormatService extends CredentialFormatService {
   ) {
     const indyCredentialRequest = requestAttachment?.getDataAsJson<CredReq>()
     let indyCredentialProposal = proposeAttachment?.getDataAsJson<CredPropose>()
-    indyCredentialProposal = JsonTransformer.fromJSON(indyCredentialProposal, CredPropose)
+    indyCredentialProposal = JsonTransformer.fromJSON(indyCredentialProposal, CredPropose, { validate: false })
```
TimoGlastra, line 574: same

packages/core/src/modules/dids/domain/__tests__/DidDocument.test.ts
```diff
@@ -152,100 +152,11 @@ describe('Did | DidDocument', () => {
...
 
-  it('validation should throw an error if the did document is invalid', async () => {
-    const didDocument = JsonTransformer.fromJSON(didExample456Invalid, DidDocument)
-
+  it('validation should throw an error if the did document is invalid', () => {
     try {
-      await MessageValidator.validate(didDocument)
+      JsonTransformer.fromJSON(didExample456Invalid, DidDocument)
     } catch (error) {
-      expect(error).toMatchObject([
```
TimoGlastra, line 161: we're losing some important test data here. Can we add an `errors` property to the class validation error class and match whether the errors match this object? I'd like to keep the test for specific validation errors in here

packages/core/src/modules/dids/domain/key-type/__tests__/bls12381g1g2.test.ts
```diff
@@ -70,7 +70,8 @@ describe('bls12381g1g2', () => {
   it('throws an error for getKeyFromVerificationMethod as it is not supported for bls12381g1g2 key types', () => {
     const verificationMethod = JsonTransformer.fromJSON(
       keyBls12381g1g2Fixture.verificationMethod[0],
-      VerificationMethod
+      VerificationMethod,
+      { validate: false }
```
TimoGlastra, line 74: What is the validation error? Would like to keep validation enabled

packages/core/src/modules/dids/domain/verificationMethod/VerificationMethodTransformer.ts
```diff
@@ -45,7 +45,7 @@ function VerificationMethodTransformer() {
         }
 
         // embedded verification method
-        return JsonTransformer.fromJSON(auth, VerificationMethod)
+        return JsonTransformer.fromJSON(auth, VerificationMethod, { validate: false })
```
TimoGlastra, line 48: same here, this is not a test file so shouldn't disable validation

packages/core/src/modules/dids/methods/peer/peerDidNumAlgo2.ts
```diff
@@ -66,7 +66,7 @@ export function didToNumAlgo2DidDocument(did: string) {
 
         service.id = `${did}#${service.type.toLowerCase()}-${serviceIndex++}`
 
-        didDocument.addService(JsonTransformer.fromJSON(service, DidDocumentService))
+        didDocument.addService(JsonTransformer.fromJSON(service, DidDocumentService, { validate: false }))
```
TimoGlastra, line 69: same here

packages/core/src/modules/dids/methods/web/WebDidResolver.ts
```diff
@@ -28,7 +28,7 @@ export class WebDidResolver implements DidResolver {
 
     let didDocument = null
     if (result.didDocument) {
-      didDocument = JsonTransformer.fromJSON(result.didDocument, DidDocument)
+      didDocument = JsonTransformer.fromJSON(result.didDocument, DidDocument, { validate: false })
```
TimoGlastra, line 31: same here

packages/core/src/modules/oob/__tests__/OutOfBandInvitation.test.ts
```diff
@@ -1,5 +1,6 @@
 import type { ValidationError } from 'class-validator'
 
+import { ClassValidationError } from '../../../error/ClassValidationError'
```
TimoGlastra, line 3: Unused import

packages/core/src/modules/proofs/__tests__/ProofRequest.test.ts
```diff
@@ -39,38 +40,34 @@ describe('ProofRequest', () => {
...
+              schema_id: 'q7ATwTYbQDgiigVijUAej:2:test:1.0',
+            },
+          ],
+        },
+      ],
+    }
 
-    // Expect 2 top level validation errors
-    expect(MessageValidator.validate(proofRequest)).rejects.toHaveLength(2)
+    expect(() => JsonTransformer.fromJSON(proofRequest, ProofRequest)).toThrowError(ClassValidationError)
```
TimoGlastra, line 71: losing that we have two validation errors (and just check if it is a class validatino error)

packages/core/src/utils/JsonTransformer.ts
```diff
@@ -1,15 +1,53 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    json: any,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    Class: { new (...args: any[]): T },
+    options: Validate = { validate: true }
+  ): T {
+    if (!options.validate) {
+      return plainToInstance(Class, json, { exposeDefaultValues: true })
+    } else {
+      const plainInstance = plainToInstance(Class, json, { exposeDefaultValues: true })
```
TimoGlastra, line 31:
```suggestion
   const instance = plainToInstance(Class, json, { exposeDefaultValues: true })
   
   // skip validation
   if (!options.validate) return instance
   
   // the validation flow top level
```

packages/core/src/utils/JsonTransformer.ts
```diff
@@ -1,15 +1,53 @@
...
 
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  public static fromJSON<T>(json: any, Class: { new (...args: any[]): T }): T {
-    return plainToInstance(Class, json, { exposeDefaultValues: true })
+  public static fromJSON<T>(
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    json: any,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    Class: { new (...args: any[]): T },
+    options: Validate = { validate: true }
```
TimoGlastra, line 26: This will mean the default value is not set when you call it with an object (e.g. `fromJson({ })` will mean validate is undefined which means false.)

We can always set a default (even if an empty object is provided) using the following syntax:
```suggestion
    { validate = true }: Validate = { }
```

--- 
or we can update it like this


```suggestion
    options?: Validate
```

so we can use it like this in the function body:
```
const validate = options?.validate ?? true
```

packages/core/src/utils/JsonTransformer.ts
```diff
@@ -1,15 +1,53 @@
...
+        return plainInstance
+      }
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      const errors = MessageValidator.validateSync(plainInstance as any)
+      // NOTE: validateSync (strangely) return an Array of errors so we
+      // have to transform that into an error of choice and throw that.
+      if (isValidationErrorArray(errors)) {
+        throw new ClassValidationError('Failed to validate class.', {
+          recordType: typeof Class,
+          cause: new Error(JSON.stringify(errors)),
```
TimoGlastra, line 43: We never put stringified errors in the cause. The cause should either be an error, or it should be undefined. In this case we can keep it undefined and add a custom errors/validationErrors property to the class which stores the errors.

packages/core/src/utils/JsonTransformer.ts
```diff
@@ -1,15 +1,53 @@
...
+        })
+      } else if (errors.length !== 0) {
+        throw new ClassValidationError('An unknown validation error occurred.', { recordType: typeof Class })
```
TimoGlastra, line 46: Have you seen cases where the array is not all ValidationError type?

TimoGlastra, line 46: recordType is something we use for storage reords (`ConnectionRecord`, etc..). Maybe we can update it to a name that doesn't conflict (and confuse) with existing things. Maybe `classType`?

packages/core/src/utils/JsonTransformer.ts
```diff
@@ -1,15 +1,53 @@
...
+      } else {
+        return plainInstance as T
+      }
```
TimoGlastra, line 49: Nit: I like to keep happy flow top level:

```suggestion

     return plainInstance as T
```

packages/core/src/utils/MessageValidator.ts
```diff
@@ -11,4 +11,8 @@ export class MessageValidator {
   public static validate<T extends object>(classInstance: T) {
     return validateOrReject(classInstance)
   }
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  public static validateSync<T extends object>(classInstance: T) {
+    return validateSync(classInstance)
+  }
```
TimoGlastra, line 17: Instead of adding the validation logic to the json transformer, what if we add it to here. This way if you use the validator directly you'll get the same error types as if you were using the json transformer. By handling this at the lowest level (the validator class) we provide more consistency in how validation errors work. This will also keep the json transformer simpler and focused on transformation

TimoGlastra (changes requested):
This is a great addition @morrieinmaas! Left some notes mainly on the disabling of validation. The goal to integrate validation with transformation is to get rid of these hidden errors, so would be great if we could fix as much validation errors as possible

morrieinmaas:
Thanls for all the feedback @TimoGlastra. sorry hadn't managed to write up some more context. From running the tests it became clear the the sync check method doesn't validate as reliably as the async one. That resulted in some tests failed as they were. Was sort of a catch 22 situation when writing this last Friday. 

Also, I'll add more in detail to your comments above, but essentially I was a little unsure of how consistent with the previous implementation we want to keep it aka not introducing breaking changes. I'll have another fresh look. Maybe I was over-complicating things and missed the elegant way

packages/core/src/modules/connections/services/ConnectionService.ts
morrieinmaas: yes eventually i disabled it, because it got a little messy with the current tests.

morrieinmaas:
I'll have another go at it today to incorporate your suggestions

packages/core/src/utils/MessageValidator.ts
morrieinmaas: ahhh yes, excellent idea!

packages/core/src/utils/JsonTransformer.ts
morrieinmaas: yes. I once got an empty array

TimoGlastra:
> I was a little unsure of how consistent with the previous implementation we want to keep it aka not introducing breaking changes

I think the previous implementation has some issues so it's fine to introduce some breaking changes if they improve behaviour

packages/core/src/decorators/ack/AckDecorator.test.ts
morrieinmaas: if i enable validation for 
```json
{
          '~please_ack': {},
          '@id': undefined,
          '@type': undefined,
        }
```
I get:
```json
{"children": [], "constraints": {"matches": "id must match /[-_./a-zA-Z0-9]{8,64}/ regular expression"}, "property": "id", "target": {"~please_ack": {}}}, {"children": [], "constraints": {"matches": "type must match /(.*?)([a-zA-Z0-9._-]+)\\/(\\d[^/]*)\\/([a-zA-Z0-9._-]+)$/ regular expression"}, "property": "type", "target": {"~please_ack": {}}}, {"children": [{"children": [], "constraints": {"isArray": "on must be an array", "isEnum": "each value in on must be a valid enum value"}, "property": "on", "target": {}}], "property": "pleaseAck", "target": {"~please_ack": {}}, "value": {}}
```

packages/core/src/decorators/transport/TransportDecorator.test.ts
morrieinmaas: similar to above - it doesn't pass validation. Happy correct this somehow. Will need to do some digging though because I'm not familiar enough with AFJ to be able to tell what's going on/wrong (the test says it should transform this - so that makes me assume it should pass validation also). Maybe I didn't get the validation quite right...

morrieinmaas:
Also random question - I've had some tests failing locally that then worked on remote due to snapshot obsolete. How can I fix this?

TimoGlastra:
> Also random question - I've had some tests failing locally that then worked on remote due to snapshot obsolete. How can I fix this?

Can you share the error log?

packages/core/src/decorators/ack/AckDecorator.test.ts
TimoGlastra: I think using something like this should work: 

```ts
{
  "@id": "1234-1234-1234-1234",
  "@type": "https://didcomm.org/test-protocol/1.0/test-message",
  "~please_ack": {
    "on": ["RECEIPT"]
   }
}
```

packages/core/src/decorators/transport/TransportDecorator.test.ts
TimoGlastra: Can you share the error(s) you get here?

morrieinmaas: Here the output for validation enabled:

```
FAIL   @aries-framework/core  packages/core/src/decorators/transport/TransportDecorator.test.ts
  Decorators | TransportDecorator
    ✓ should correctly transform Json to TransportDecorator class (2 ms)
    ✓ should correctly transform TransportDecorator class to Json
    ✕ should only allow correct return_route values (2 ms)
    ✕ should require return_route_thread when return_route is thread

  ● Decorators | TransportDecorator › should only allow correct return_route values

    ClassValidationError: function: Failed to validate class.

      26 |     const errors = validateSync(classInstance)
      27 |     if (isValidationErrorArray(errors)) {
    > 28 |       throw new ClassValidationError('Failed to validate class.', {
         |             ^
      29 |         classType: typeof Class,
      30 |         validationErrors: errors,
      31 |       })

      at Function.validateSync (src/utils/MessageValidator.ts:28:13)
      at Function.fromJSON (src/utils/JsonTransformer.ts:34:24)
      at validTransport (src/decorators/transport/TransportDecorator.test.ts:7:45)
      at expectInvalid (src/decorators/transport/TransportDecorator.test.ts:11:10)
      at Object.<anonymous> (src/decorators/transport/TransportDecorator.test.ts:68:11)

  ● Decorators | TransportDecorator › should only allow correct return_route values

    expect.assertions(4)

    Expected four assertions to be called but received three assertion calls.

      62 |
      63 |   it('should only allow correct return_route values', async () => {
    > 64 |     expect.assertions(4)
         |            ^
      65 |     await expectValid(valid.all)
      66 |     await expectValid(valid.none)
      67 |     await expectValid(valid.thread)

      at Object.<anonymous> (src/decorators/transport/TransportDecorator.test.ts:64:12)

  ● Decorators | TransportDecorator › should require return_route_thread when return_route is thread

    ClassValidationError: function: Failed to validate class.

      26 |     const errors = validateSync(classInstance)
      27 |     if (isValidationErrorArray(errors)) {
    > 28 |       throw new ClassValidationError('Failed to validate class.', {
         |             ^
      29 |         classType: typeof Class,
      30 |         validationErrors: errors,
      31 |       })

      at Function.validateSync (src/utils/MessageValidator.ts:28:13)
      at Function.fromJSON (src/utils/JsonTransformer.ts:34:24)
      at validTransport (src/decorators/transport/TransportDecorator.test.ts:7:45)
      at expectInvalid (src/decorators/transport/TransportDecorator.test.ts:11:10)
      at Object.<anonymous> (src/decorators/transport/TransportDecorator.test.ts:74:11)

Test Suites: 1 failed, 1 total
Tests:       2 failed, 2 passed, 4 total
Snapshots:   0 total
Time:        0.294 s, estimated 1 s
Ran all test suites matching /packages\/core\/src\/decorators\/transport\/TransportDecorator.test.ts/i.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

TimoGlastra: Can you share the actual validation errors? Also can we look at extending the error message to include the actual validation errors? It would be really useful to see the valiadtion error in the message/stack.

morrieinmaas: yes will do

TimoGlastra:
I think your environement got corrupted. To clean you can do the following to delete all indy wallets:

`rm -rf ~/.indy_client`

And as you don't have a postgres database running locally you can run the tests with:

```
yarn test --testPathIgnorePatterns ./packages/core/tests/postgres.test.ts
```

Can you see if the errors still persist?

morrieinmaas:
> I think your environement got corrupted. To clean you can do the following to delete all indy wallets:
> 
> `rm -rf ~/.indy_client`
> 
> And as you don't have a postgres database running locally you can run the tests with:
> 
> ```
> yarn test --testPathIgnorePatterns ./packages/core/tests/postgres.test.ts
> ```
> 
> Can you see if the errors still persist?

Ah okay thanks a bunch. teh rm -rf ~/.indy_client was exactly the kind of thing i was looking for. will try your suggestions in a minute

packages/core/src/decorators/transport/TransportDecorator.test.ts
morrieinmaas: alright i figured it out. Will push in a second.

morrieinmaas: Regarding the `It would be really useful to see the validation error in the message/stack.` if I understand that correctly, I added that already last night. You can now either use the validationErrors attribute on ClassValidationError object or you can call the inspect() method which also provides the errors

morrieinmaas:
> `yarn test --testPathIgnorePatterns ./packages/core/tests/postgres.test.ts`

Hmm also weird - my postgres container died...

morrieinmaas:
> I think your environement got corrupted. To clean you can do the following to delete all indy wallets:
> 
> `rm -rf ~/.indy_client`
> 
> And as you don't have a postgres database running locally you can run the tests with:
> 
> ```
> yarn test --testPathIgnorePatterns ./packages/core/tests/postgres.test.ts
> ```
> 
> Can you see if the errors still persist?

That unfortunately doesn't change much. here the output https://gist.github.com/morrieinmaas/d2f7c8795eddf9052893bee2bec82202

packages/core/src/decorators/transport/TransportDecorator.test.ts
TimoGlastra: I mean it would be useful to somehow print the validation error when the error/stack is printed. 

See this thread for an example of what I mean: https://github.com/typestack/class-validator/issues/26#issuecomment-374059419

It appears validation error exposes an `toString()` method that you can call. If we loop over the array and print it for each error that would be nice I think.

TimoGlastra:
Maybe it fails because the backups already exists... Can you try to remove the backups also?

This one is a bit more complex (will look at a way to nuke this more easily) but you need to add a log to the `0.1.test.ts` file on line 77 with the following: 

```ts
console.log(`${agent.config.fileSystem.basePath}/afj/migration/backup`)
```

then run 

```ts
yarn test packages/core/src/storage/migration/__tests__/0.1.test.ts
```

and then delete the folder that is printend to the console

```
rm -rf /var/folders/l3/xy8jzyvj4p5_d9g1123rt4bw0000gn/T/afj/migration/backup
```

Also make sure to delete your indy wallet and then try again. If that doesn't work I'm not sure why this doesn't work on your end :shrug:

morrieinmaas:
Maybe up for discussion for future reference: 

* Maybe we can scope the issue a bit better before starting?!
* Happy to make all the changes but think in hindsight it would've been better to:
  * Be more clear in the issue what this PR should do an
  * What it shouldn't do
  * This would've been useful because there were async and sync parts of code (cos .validate is async method and to JSON was sync method). Wasn't clear to what the desired way was. We discussed that here in code review then, but I think this would be nice-to-have and easily possible in the ticket description. Also regarding tests and breaking changes:
* Would have implemented the validator in this PR but left it disabled by default so:
  * The logic is there, but optional -> no breaking changes to how the code works. tests should mostly pass (which also means people's code won't break (much)
* open another PR to set the default validation to true.
  * This would then introduce more changes in logic/break more tests

morrieinmaas:
@TimoGlastra - ready for re-review.

packages/core/src/modules/proofs/__tests__/ProofRequest.test.ts
```diff
@@ -39,38 +40,41 @@ describe('ProofRequest', () => {
...
+    let caughtError
```
blu3beri, line 72: Same here.

packages/core/src/modules/proofs/services/ProofService.ts
```diff
@@ -700,7 +700,9 @@ export class ProofService {
       //Get credentialInfo
       if (!requestedAttribute.credentialInfo) {
         const indyCredentialInfo = await this.indyHolderService.getCredential(requestedAttribute.credentialId)
-        requestedAttribute.credentialInfo = JsonTransformer.fromJSON(indyCredentialInfo, IndyCredentialInfo)
+        requestedAttribute.credentialInfo = JsonTransformer.fromJSON(indyCredentialInfo, IndyCredentialInfo, {
+          validate: true,
```
blu3beri, line 704: Default value

packages/core/src/storage/migration/__tests__/backup.test.ts
```diff
@@ -130,7 +130,8 @@ describe('UpdateAssistant | Backup', () => {
 
     try {
       await updateAssistant.update()
-    } catch (error) {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } catch (error: any) {
```
blu3beri, line 134: Any chance we can avoid this? unknown or just no type?

packages/core/src/utils/JsonTransformer.ts
```diff
@@ -1,15 +1,41 @@
...
       exposeDefaultValues: true,
     })
   }
 
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  public static fromJSON<T>(json: any, Class: { new (...args: any[]): T }): T {
-    return plainToInstance(Class, json, { exposeDefaultValues: true })
+  public static fromJSON<T>(
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    json: any,
```
blu3beri, line 22: I'd assume this could be `Record<string, unknown>`

packages/core/src/utils/JsonTransformer.ts
```diff
@@ -1,15 +1,41 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    Class: { new (...args: any[]): T },
+    { validate = true }: Validate = {}
```
blu3beri, line 25:
```suggestion
    { validate = true }: Validate
```

What is the empty object assignment there? It would collide with the `{validate: true}` and one would take precedence (not sure which one).

packages/core/src/utils/JsonTransformer.ts
```diff
@@ -1,15 +1,41 @@
...
+  ): T {
+    const instance = plainToInstance(Class, json, { exposeDefaultValues: true })
+
+    // Skip validation
+    if (!validate === true) return instance
```
blu3beri, line 30:
```suggestion
    if (!validate) return instance
```

packages/core/src/utils/JsonTransformer.ts
```diff
@@ -1,15 +1,41 @@
...
+
+    if (instance === undefined || instance === null) {
```
blu3beri, line 32:
```suggestion
    if (!instance) {
```

packages/core/src/utils/JsonTransformer.ts
```diff
@@ -1,15 +1,41 @@
...
+      throw new ClassValidationError('Cannot validate instance of ', { classType: typeof instance })
+    }
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    MessageValidator.validateSync(instance, Class as any)
```
blu3beri, line 36:
```suggestion
    MessageValidator.validateSync(instance, Class as T)
```

I am not too sure if this would work

packages/core/src/utils/JsonTransformer.ts
```diff
@@ -25,8 +51,12 @@ export class JsonTransformer {
...
   }
 
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  public static deserialize<T>(jsonString: string, Class: { new (...args: any[]): T }): T {
-    return this.fromJSON(JSON.parse(jsonString), Class)
+  public static deserialize<T>(
+    jsonString: string,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    Class: { new (...args: any[]): T },
+    { validate = true }: Validate = {}
```
blu3beri, line 58: Same here with the default object assignment.

packages/core/src/utils/JsonTransformer.ts
```diff
@@ -25,8 +51,12 @@ export class JsonTransformer {
...
+  ): T {
+    return this.fromJSON(JSON.parse(jsonString), Class, { validate: validate })
```
blu3beri, line 60:
```suggestion
    return this.fromJSON(JSON.parse(jsonString), Class, { validate })
```

packages/core/src/utils/JsonTransformer.ts
```diff
@@ -25,8 +51,12 @@ export class JsonTransformer {
     return JSON.stringify(this.toJSON(classInstance))
   }
 
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  public static deserialize<T>(jsonString: string, Class: { new (...args: any[]): T }): T {
-    return this.fromJSON(JSON.parse(jsonString), Class)
+  public static deserialize<T>(
+    jsonString: string,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    Class: { new (...args: any[]): T },
```
blu3beri, line 57:
```suggestion
    class: { new (...args: any[]): T },
```

packages/core/src/utils/MessageValidator.ts
```diff
@@ -11,4 +14,24 @@ export class MessageValidator {
   public static validate<T extends object>(classInstance: T) {
     return validateOrReject(classInstance)
   }
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  public static validateSync<T extends object>(
```
blu3beri, line 18:
```suggestion
  public static validateSync<T extends Record<string, unknown>>(
```

packages/core/src/utils/MessageValidator.ts
```diff
@@ -11,4 +14,24 @@ export class MessageValidator {
...
+    // eslint-disable-next-line @typescript-eslint/ban-types
+    classInstance: T & {},
```
blu3beri, line 20:
```suggestion
    classInstance: T,
```

I believe this would suffice.

packages/core/src/utils/MessageValidator.ts
```diff
@@ -11,4 +14,24 @@ export class MessageValidator {
...
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    Class: { new (...args: any[]): T }
```
blu3beri, line 22:
```suggestion
    class: { new (...args: any[]): T }
```

packages/core/src/utils/JsonTransformer.ts
```diff
@@ -1,15 +1,41 @@
...
   }
 
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  public static fromJSON<T>(json: any, Class: { new (...args: any[]): T }): T {
-    return plainToInstance(Class, json, { exposeDefaultValues: true })
+  public static fromJSON<T>(
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    json: any,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    Class: { new (...args: any[]): T },
```
blu3beri, line 24:
```suggestion
    class: { new (...args: any[]): T },
```

blu3beri (changes requested):
Nice changes, especially some less async code :). Comments are mainly about cleanup and less `eslint-disable-next-line` 

Disabling the next line should only be used as a last resort and we should really avoid it as much as possible.

morrieinmaas:
> Nice changes, especially some less async code :). Comments are mainly about cleanup and less `eslint-disable-next-line`
> 
> Disabling the next line should only be used as a last resort and we should really avoid it as much as possible.

Thanks for all the feedback. I'll address the requested changes/suggestions and push them asap.

packages/core/src/error/ValidationErrorUtils.ts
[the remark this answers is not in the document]
morrieinmaas: I'm not so deep into Javascript/Typescript internals but the reason I went for any (unknwon is a good suggestion) is that when we do catch we don't know the type of what is thrown (well I do but the linter doesn't). So when something else occurs than Array<type> I'd keep it generic to just return false. Note that if it was type Array<ValidationError | unknown> ts would still complain in some possible(?) cases of the input not being a list (which I cover in a test.

Not sure right now if there is a better solution, but please let me know if you can think of any (no pun intended). I'll make it unknown for now.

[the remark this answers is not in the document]
morrieinmaas: Maybe `e: ValidationError[] | unknown` ?

packages/core/src/storage/migration/__tests__/backup.test.ts
morrieinmaas: Just no type I guess. There's also the option to add "useUnknownInCatchVariables": true to tsconfig (But that gets a little weird sldo depending on typescript version aka works for me locally but fails in the gh pipeline. I'll just revert it to previous for now.

packages/core/src/utils/JsonTransformer.ts
morrieinmaas: hmm good shout lemme give that a try

packages/core/src/utils/JsonTransformer.ts
morrieinmaas: `{ validate = true }: Validate` does not work with `    TypeError: Cannot read properties of undefined (reading 'validate')` whereas ` { validate = true }: Validate = { validate: true }` gives the same (correct) behaviour as `    { validate = true }: Validate = {}`. not 100% sure what shorthand quirk of typescript that is exactly - but it works😅

packages/core/src/utils/JsonTransformer.ts
morrieinmaas: Happy to change that otherwise but casting to T gives:

```
Argument of type 'T' is not assignable to parameter of type 'new (...args: any[]) => object'
```
ts(2345)

packages/core/src/utils/JsonTransformer.ts
morrieinmaas: same as above

packages/core/src/utils/JsonTransformer.ts
morrieinmaas: I assume you mean `cls`?

morrieinmaas: `'class' is not allowed as a parameter name.ts(1390)`

packages/core/src/utils/MessageValidator.ts
morrieinmaas: Maybe there is a good way to keep it `T` only? Added & {} because that is the type returned by `plainToInstance`

packages/core/src/utils/MessageValidator.ts
morrieinmaas: This also won't work when parsing an `AgentMessage`

packages/core/src/utils/JsonTransformer.ts
morrieinmaas: see below/same as below

packages/core/src/error/ValidationErrorUtils.ts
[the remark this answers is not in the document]
blu3beri: I am not entirely sure what you mean, but if `unknown` is possible always use that instead of `any`.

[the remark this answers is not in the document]
morrieinmaas: yes, agreed! I meant maybe `e: ValidationError[] | unknown` is better than just `unknown`?

Just `unknown` certainly works. Happy to use either

packages/core/src/utils/JsonTransformer.ts
morrieinmaas: same as above

packages/core/src/utils/JsonTransformer.ts
blu3beri: Hmmm, I will take a look at this later as this is not really how it is supposed to work haha.

packages/core/src/utils/JsonTransformer.ts
blu3beri: Anything other than pascalcase is fine by my :)

packages/core/src/error/ValidationErrorUtils.ts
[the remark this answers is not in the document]
blu3beri: `e: ValidationError[] | unknown` is okay.

packages/core/src/agent/__tests__/AgentMessage.test.ts
```diff
@@ -43,32 +43,51 @@ describe('AgentMessage', () => {
...
+      expect(() => JsonTransformer.fromJSON(json, CustomProtocolMessage)).toThrowError(ClassValidationError)
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      let thrownError: any
+      try {
+        JsonTransformer.fromJSON(json, CustomProtocolMessage)
+      } catch (e) {
+        thrownError = e
+      }
+      expect(thrownError.message).toContain('Failed to validate class.')
+      expect(thrownError.validationErrors).toMatchObject([
```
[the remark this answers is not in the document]
morrieinmaas: Could you elaborate on why?

packages/core/src/utils/MessageValidator.ts
blu3beri: Thats odd. The only reason I can imagine this would not work is when the key is `number` or `Symbol`. Could you try `Record<string | number | Symbol, unknown>`?

If that works we should create a new type alias for that and use that. We use `Record<string, unknown>` everywhere as an `anyObject` type, but it does not catch all objects.

packages/core/src/agent/__tests__/AgentMessage.test.ts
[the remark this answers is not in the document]
morrieinmaas: Just curious

packages/core/src/utils/MessageValidator.ts
morrieinmaas: I'll give that a try but am suspecting that at least for the .reduce in transformer this will still complain - but lemme check.

packages/core/src/agent/__tests__/AgentMessage.test.ts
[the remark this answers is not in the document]
morrieinmaas: ah alright thanks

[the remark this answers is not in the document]
morrieinmaas: As a side note `toThrow()` doesn't accept assymetric matchers (eg from `expect.objectContaining`)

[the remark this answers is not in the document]
morrieinmaas: That then makes it a bit tricky to catch the `ValidationError[]` on .validationError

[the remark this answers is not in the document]
morrieinmaas: alright, got an approach that doesn't use let: empty - will push in a bit

packages/core/src/utils/MessageValidator.ts
morrieinmaas: Other PR - due to complexity

packages/core/src/modules/dids/domain/__tests__/DidDocument.test.ts
TimoGlastra: weird this is losing all this data. If you solve this one error, will it give a new error? (i.e. it will deliver them one by one instead)

packages/core/src/utils/JsonTransformer.ts
```diff
@@ -1,15 +1,41 @@
...
 
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  public static fromJSON<T>(json: any, Class: { new (...args: any[]): T }): T {
-    return plainToInstance(Class, json, { exposeDefaultValues: true })
+  public static fromJSON<T>(
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    json: any,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    cls: { new (...args: any[]): T },
+    { validate }: Validate = { validate: true }
```
TimoGlastra, line 25: If we set it right from the `=`, it means if I pass an empty object to  (`fromJson(json, Class, {})`) it will be set to false. Setting it left of the `=` means it will be assigned whenever the property `validate` is undefined.

```suggestion
    { validate = true }: Validate = {  }
```

packages/core/src/utils/__tests__/JsonTransformer.test.ts
```diff
@@ -72,9 +72,9 @@ describe('JsonTransformer', () => {
...
+        '{"@context":["https://w3id.org/did/v1"],"id":"did:peer:1zQmRYBx1pL86DrsxoJ2ZD3w42d7Ng92ErPgFsCSqg8Q1h4i","controller": "nowYouAreUnderMyControl", "keyAgreement":[{"id":"#6MkqRYqQiSgvZQdnBytw86Qbs2ZWUkGv22od935YF4s8M7V", "controller": "#id", "type":"Ed25519VerificationKey2018","publicKeyBase58":"ByHnpUCFb1vAfh9CFZ8ZkmUZguURW8nSw889hy6rD8L7"}],"service":[{"id":"#service-0","type":"did-communication","serviceEndpoint":"https://example.com/endpoint","recipientKeys":["#6MkqRYqQiSgvZQdnBytw86Qbs2ZWUkGv22od935YF4s8M7V"],"routingKeys":["did:key:z6MkpTHR8VNsBxYAAWHut2Geadd9jSwuBV8xRoAnwWsdvktH#z6MkpTHR8VNsBxYAAWHut2Geadd9jSwuBV8xRoAnwWsdvktH"],"accept":["didcomm/v2","didcomm/aip2;env=rfc587"]}]}'
 
-      const didDocument = JsonTransformer.deserialize(didDocumentString, DidDocument)
+      const didDocument = JsonTransformer.deserialize(didDocumentString, DidDocument, { validate: false })
```
TimoGlastra, line 77: This passes for me locally with validation enabled
```suggestion
      const didDocument = JsonTransformer.deserialize(didDocumentString, DidDocument)
```

packages/core/src/utils/__tests__/MessageValidator.test.ts
```diff
@@ -0,0 +1,49 @@
...
+    })
+    it('throws an error for invalid class instance', () => {
+      const invitation = new ConnectionInvitationMessage({
+        did: 'did:sov:test1234',
+        id: 'afe2867e-58c3-4a8d-85b2-23370dd9c9f0',
+        label: 'test-label',
+      })
+      invitation.did = 'jnsjcd'
+
+      expect(MessageValidator.validateSync(invitation, ConnectionInvitationMessage)).toThrow
```
TimoGlastra, line 46: If the method is not sync it should always be wrapped, so I'm not sure if this is actually validating. If I change it to the following it is not actually throwing:

```suggestion
      expect(() => MessageValidator.validateSync(invitation, ConnectionInvitationMessage)).toThrow()
```

I don't think you can use `toThrow` without the `()`. Same with `.rejects` with nothing following it (should be `.rejects.toThrow()`

You can make the tests (all in this file are not correct) working by setting the `.did = undefined` this will trigger a validation error

packages/core/src/utils/MessageValidator.ts
```diff
@@ -11,4 +14,26 @@ export class MessageValidator {
   public static validate<T extends object>(classInstance: T) {
     return validateOrReject(classInstance)
```
TimoGlastra, line 15: this will still use the `class-validator` error while the `validateSync` method now uses our custom error handling. We should either remove this method, or give it the same care we did in the `validateSync` method

packages/core/src/agent/__tests__/AgentMessage.test.ts
```diff
@@ -43,32 +43,51 @@ describe('AgentMessage', () => {
...
-
-      await expect(MessageValidator.validate(message)).rejects.toMatchObject([
+      expect(() => JsonTransformer.fromJSON(json, CustomProtocolMessage)).toThrowError(ClassValidationError)
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      let thrownError: any
+      try {
+        JsonTransformer.fromJSON(json, CustomProtocolMessage)
+      } catch (e) {
+        thrownError = e
+      }
```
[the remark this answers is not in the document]
TimoGlastra: This will mainly help in other place where we are calling `expect` in a catch blocks. If the try block never throws the expects will never get called and the tests will succeed. We could also solve this using `expect.assertions(n)`.

packages/core/src/decorators/transport/TransportDecorator.test.ts
```diff
@@ -4,11 +4,11 @@ import { MessageValidator } from '../../utils/MessageValidator'
 import { TransportDecorator, ReturnRouteTypes } from './TransportDecorator'
 
 const validTransport = (transportJson: Record<string, unknown>) =>
-  MessageValidator.validate(JsonTransformer.fromJSON(transportJson, TransportDecorator))
+  MessageValidator.validate(JsonTransformer.fromJSON(transportJson, TransportDecorator, { validate: true }))
```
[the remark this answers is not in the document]
morrieinmaas: outdated

packages/core/src/utils/JsonTransformer.ts
morrieinmaas: ah sorry already sorted - haven't pushed

packages/core/src/utils/MessageValidator.ts
morrieinmaas: I'll remove it

packages/core/src/decorators/ack/AckDecorator.test.ts
```diff
@@ -25,107 +27,56 @@ describe('Decorators | AckDecoratorExtension', () => {
...
-      },
-      TestMessage
-    )
+        TestMessage,
+        {
+          validate: true,
+        }
+      )
 
-    expect(transformed).toEqual({
```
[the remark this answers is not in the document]
morrieinmaas: I'm slightly confused as to why this is showing outdated changes

packages/core/src/decorators/ack/AckDecorator.test.ts
```diff
@@ -25,107 +27,56 @@ describe('Decorators | AckDecoratorExtension', () => {
...
+          '@type': undefined,
         },
-        '@id': '7517433f-1150-46f2-8495-723da61b872a',
-        '@type': 'https://didcomm.org/test-protocol/1.0/test-message',
-      },
-      TestMessage
-    )
+        TestMessage,
+        {
+          validate: true,
```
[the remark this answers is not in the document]
morrieinmaas: this is outdated (pushed the changes but still appears here?!)

packages/core/src/modules/credentials/formats/CredentialFormatService.ts
```diff
@@ -35,7 +35,7 @@ export abstract class CredentialFormatService<CF extends CredentialFormat = Cred
 
   // proposal methods
   abstract createProposal(options: FormatCreateProposalOptions<CF>): Promise<FormatCreateProposalReturn>
-  abstract processProposal(options: FormatProcessOptions): Promise<void>
+  abstract processProposal(options: FormatProcessOptions): void
```
TimoGlastra, line 38: this should stay async (it's an interface that could have different implementation in the future that may need async logic)

packages/core/src/agent/__tests__/AgentMessage.test.ts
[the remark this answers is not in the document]
morrieinmaas: maybe we can make that a separate issue?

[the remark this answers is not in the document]
morrieinmaas: and then also go through the code and see where this is applicable and make changes so we keep it consistent

packages/core/src/agent/MessageReceiver.ts
```diff
@@ -168,7 +168,7 @@ export class MessageReceiver {
     let message: AgentMessage
     try {
       message = await this.transformMessage(plaintextMessage)
-      await this.validateMessage(message)
+      this.validateMessage(message)
```
TimoGlastra, line 171: `transformMessage` will also validate, so we're validating twice now. I think we can remove the `this.validateMessage` method completely

TimoGlastra: Also this method uses `AgentMessage` for validation which looses the message specific properties, the validation done in the transform message method will cover this correctly

TimoGlastra (changes requested):

packages/core/src/utils/MessageValidator.ts
```diff
@@ -8,7 +11,24 @@ export class MessageValidator {
    * @throws array of validation errors {@link ValidationError}
    */
   // eslint-disable-next-line @typescript-eslint/ban-types
-  public static validate<T extends object>(classInstance: T) {
-    return validateOrReject(classInstance)
+  public static validateSync<T extends object>(
+    // eslint-disable-next-line @typescript-eslint/ban-types
+    classInstance: T & {},
```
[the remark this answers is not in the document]
morrieinmaas: Had a discussion about that with @blu3beri - this doesn't work with only `T`. I'll have another go at it

packages/core/src/utils/__tests__/MessageValidator.test.ts
morrieinmaas: ah sorry - yes, i missunderstood that!😅

packages/core/src/utils/MessageValidator.ts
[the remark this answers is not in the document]
morrieinmaas: the type check for this fails with only `T` with `Argument of type 'T' is not assignable to parameter of type 'object'.`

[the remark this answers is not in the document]
morrieinmaas: I tried using `NonNullable<T>` to no avail

packages/core/src/utils/JsonTransformer.ts
```diff
@@ -1,15 +1,40 @@
...
```
TimoGlastra, line 25: validate should be set to true on the left side of the equation

See the output of this script for why we want this (https://www.typescriptlang.org/play?#code/PTAEHUFMBsGMHsC2lQBd5oBYoCoE8AHSAZVgCcBLA1UABWgEM8BzM+AVwDsATAGiwoBnUENANQAd0gAjQRVSQAUCEmYKsTKGYUAbpGF4OY0BoadYKdJMoL+gzAzIoz3UNEiPOofEVKVqAHSKymAAmkYI7NCuqGqcANag8ABmIjQUXrFOKBJMggBcISGgoAC0oACCbvCwDKgU8JkY7p7ehCTkVDQS2E6gnPCxGcwmZqDSTgzxxWWVoASMFmgYkAAeRJTInN3ymj4d-jSCeNsMq-wuoPaOltigAKoASgAywhK7SbGQZIIz5VWCFzSeCrZagNYbChbHaxUDcCjJZLfSDbExIAgUdxkUBIursJzCFJtXydaiKDIKMjJBhLABqDGgFG4dRQAG9FCUSjoGUyWQB+fLjeDwFqcRQAX0UCE4gho7mSNAAvKAABSs0DcxnMhSgZWoMjsFDiwX0rUs3WgVnigCUuoAfGiZSLIAFoPBmCqAETy1Ce-ia3kKa3BH0q4Ohq3hyAKtUannayCC6nQQRG4NSxqy0CUZiYJWq9UBhOgY2gU2BlDKwvxlmC-WGm32x2CZ2u91enN5v1xs1B4Kd1BhxQDtU24cUXOD6u9xM4hmpkvBlQ4ZoeMheRDwPoMYHsGhfNxmL5kd6p-jMSBHVCOBSuaR4bOQTc6YZYfQoBCIaHCS4HiRb6IRAMDhsQWJhWA4HgAiAA)

morrieinmaas: @blu3beri have a look at the link.

packages/core/src/agent/__tests__/AgentMessage.test.ts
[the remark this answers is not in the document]
morrieinmaas: opened #874 for this yesterday

TimoGlastra (approved):
A lot of work went into this PR, great work @morrieinmaas!

TimoGlastra:
@blu3beri can you give a review to unblock or leave more feedback?

morrieinmaas:
> A lot of work went into this PR, great work @morrieinmaas!

haha puuh that's come a long way. thanks for all the feedback and assistance.🙏

blu3beri (approved):
Unblocked.

Issue #739, opened 2022-05-04
Complex storage queries (Wallet Query Language (WQL))


karimStekelenburg:
For some use cases it's currently required to query the `StorageService` multiple times in a row. This would not be necesarry if the `StorageService` would expose an API that allows for complex queries that include `$and`, `$or` and `$not` clauses.

We should implement the Wallet Query Language (WQL) as specified by [Indy doc 011](https://github.com/hyperledger/indy-sdk/tree/master/docs/design/011-wallet-query-language).

blu3beri:
Do you think we should just allow a WQL string or should we create a builder for this? Would be quite a nice feature to have, and with a builder we could also expose it publicly.

PR #831, opened 2022-06-02
feat: support advanced wallet query

TimoGlastra:
Adds support for some of the most important wallet query features as defined by the indy-sdk (and aries-askar). https://hyperledger-indy.readthedocs.io/projects/sdk/en/latest/docs/design/011-wallet-query-language/README.html

Currently only added support for `$and`, `$or` and `$not`, which is enough for most use cases. We need this in particular for querying credentials based on a DIF presentation definition. 

Fixes https://github.com/hyperledger/aries-framework-javascript/issues/739

@blu3beri I just used the default syntax for this, so you'd use it like this:

```ts
const query: Query<TheRecord> = {
  $and: [
	{ $or: [{ expandedTypes: 'https://url1.com' }, { context: 'https://url1.com' }] },
	{ $or: [{ expandedTypes: 'https://url2.com' }, { context: 'https://url2.com' }] },
  ]
}
```

We can look at a builder, but not sure what that would look like exactly? 

@NB-PrasadKatkar this should unblock you for the presentation exchange work


karimStekelenburg (approved):
LGMT. However, may I ask why we're not implementing all operators in one go? Are the others more work or something?

I can see these cover most use cases, but I can also think of some use cases where `$gt` and `$lt` are useful for example.

TimoGlastra:
The `$gt` and `$lt` only work on unencrypted tags in indy and askar, but we only support encrypted tags in afj. Would like to keep it minimal so it's easier for new storage methods to implement, especially if we aren't using some features

blu3beri:
Why is inMemory storage not supported for querying?

PR #833, opened 2022-06-02
fix: clone record before emitting event

### LINKED ISSUE #832: record emitted in events are modified afterwards, changing the payload of an event
Take the scenario where a request is received for a new connection. Once the request has been processed we emit an event where the connection record has state `request-received`. We then return the connection record. If auto accept is enabled (which is basically always for connections) we immediate create a response and update the state on the record to response-sent.

Because we're using the same object for the whole the state in the record that was emitted in the event is now also response-sent. If you immediately process the events this is oftentimes not a problem, but it you put them in some kind of queue (like we do in AATH), this can result in unexpected behaviour as it may seem the event for request-received has never been fired.

I was thinking how to quickly fix this, but this seems deeply rooted in all protocols of AFJ. The correct option would probably be to always refetch the record / clone the record before emitting events, but this is probably easier said than done. 

Cloning would be desired over refetching, so maybe we can look into cloning the object using `classToClass` (https://github.com/typestack/class-transformer#classtoclass) before emitting a record in an event. Opening this issue to be aware of this kind of behaviour, and make sure we handle this for new protocols being added.

TimoGlastra:
Clone records before emitting them in events to avoid mutation later on. 

Fixes #832

### DIFF
--- a/packages/core/src/modules/connections/services/ConnectionService.ts
+++ b/packages/core/src/modules/connections/services/ConnectionService.ts
@@ -181,13 +181,7 @@ export class ConnectionService {
     })
 
     await this.connectionRepository.update(connectionRecord)
-    this.eventEmitter.emit<ConnectionStateChangedEvent>({
-      type: ConnectionEventTypes.ConnectionStateChanged,
-      payload: {
-        connectionRecord,
-        previousState: null,
-      },
-    })
+    this.emitStateChangedEvent(connectionRecord, null)
 
     this.logger.debug(`Process message ${ConnectionRequestMessage.type} end`, connectionRecord)
     return connectionRecord
@@ -509,10 +503,17 @@ export class ConnectionService {
     connectionRecord.state = newState
     await this.connectionRepository.update(connectionRecord)
 
+    this.emitStateChangedEvent(connectionRecord, previousState)
+  }
+
+  private emitStateChangedEvent(connectionRecord: ConnectionRecord, previousState: DidExchangeState | null) {
+    // Connection record in event should be static
+    const clonedConnection = JsonTransformer.clone(connectionRecord)
+
     this.eventEmitter.emit<ConnectionStateChangedEvent>({
       type: ConnectionEventTypes.ConnectionStateChanged,
       payload: {
-        connectionRecord,
+        connectionRecord: clonedConnection,
         previousState,
       },
     })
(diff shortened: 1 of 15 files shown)


packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -509,10 +503,17 @@ export class ConnectionService {
     connectionRecord.state = newState
     await this.connectionRepository.update(connectionRecord)
 
+    this.emitStateChangedEvent(connectionRecord, previousState)
+  }
+
+  private emitStateChangedEvent(connectionRecord: ConnectionRecord, previousState: DidExchangeState | null) {
```
blu3beri, line 509: Why not 
```suggestion
  private emitStateChangedEvent(connectionRecord: ConnectionRecord, previousState?: DidExchangeState) {
```

blu3beri (approved):
I like the addition of adding the `emitStateChangedEvent` and cloning/copying, but would it not make more sense to implement clone on the `BaseRecord` class? This way, we would avoid calling `const clonedRecord = record.clone()`. 

I also think that this will be forgotten in the future and cause problems again. I am sure of a solution but it might be nice to revisit this later on.

packages/core/src/modules/connections/services/ConnectionService.ts
TimoGlastra: Events are emitted with previousState value null (which has a different meaning than undefined). If we change this now it would be a breaking change to events (because `previousState` will now be undefined)

TimoGlastra:
> This way, we would avoid calling const clonedRecord = record.clone().

How could we avoid calling record.clone? I like the idea of adding a clone to the base record, but instead of calling `JsonTransformer.clone` we'd now call `record.clone()` right? Or am I missing something?

blu3beri:
> > This way, we would avoid calling const clonedRecord = record.clone().
> 
> How could we avoid calling record.clone? I like the idea of adding a clone to the base record, but instead of calling `JsonTransformer.clone` we'd now call `record.clone()` right? Or am I missing something?

Yeah record.clone(). 

Maybe clone adds a property `clone: true` and where we want to passed a cloned object we require that property? This does not avoid calling it but we would not forget to call it at least.

TimoGlastra:
I'm not 100% following you. Where would we add the clone property to? Could you give an example of how this would work?

I'm going to merge this now to unblock me in the AATH work, but happy to make a follow up PR

blu3beri:
> I'm not 100% following you. Where would we add the clone property to? Could you give an example of how this would work?
> 
> I'm going to merge this now to unblock me in the AATH work, but happy to make a follow up PR

Yeah my apologies, not the best explanantion. 

I want to enforce calling `clone()` on the record so we dont encounter more issues like this. My idea was that cloning would add a property to the record called `shouldBeCloned` and set it to `true`. Then every function that would modify the record would basically check if a property on the record called `shouldBeCloned` would be `true`.

This is for sure not the best solution and just an idea. Ignore this if its still too vague :).

PR #834, opened 2022-06-03
fix(oob): expose parseInvitation publicly

blu3beri:
- Quite a nice function to have outside of the framework for parsing qrcodes.


packages/core/src/utils/parseInvitation.ts
```diff
@@ -0,0 +1,27 @@
...
+import { OutOfBandInvitation } from '../modules/oob/messages'
+
+/**
+ * Parses URL containing encoded invitation and returns invitation message.
+ *
+ * @param invitationUrl URL containing encoded invitation
+ *
+ * @returns OutOfBandInvitation
+ */
+export const parseInvitation = async (invitationUrl: string): Promise<OutOfBandInvitation> => {
```
TimoGlastra, line 15:
```suggestion
export const parseInvitationUrl = async (invitationUrl: string): Promise<OutOfBandInvitation> => {
```

packages/core/src/modules/oob/OutOfBandModule.ts
```diff
@@ -280,18 +274,8 @@ export class OutOfBandModule {
...
-      return outOfBandInvitation
-    } else if (parsedUrl['c_i'] || parsedUrl['d_m']) {
-      const invitation = await ConnectionInvitationMessage.fromUrl(invitationUrl)
-      return convertToNewInvitation(invitation)
-    }
-    throw new AriesFrameworkError(
-      'InvitationUrl is invalid. It needs to contain one, and only one, of the following parameters: `oob`, `c_i` or `d_m`.'
-    )
+  public async parseInvitation(invitationUrl: string): Promise<OutOfBandInvitation> {
+    return await this.parseInvitation(invitationUrl)
```
TimoGlastra, line 278: This will cause an endless loop it seems?

PR #835, opened 2022-06-06
fix: support pre-aip2 please ack decorator

TimoGlastra:
The update to the please ack decorator broke for pre-aip 2 ack decorators. This adds some sane defaults to make sure AFJ doesn't break with old please ack decorators


amanji (approved):
LGTM. Also tested locally on Bifold and it looks like things are working as expected

blu3beri (approved):

PR #838, opened 2022-06-07
fix: send message to service

niallshaw-absa:
Fixed issue where message sender would incorrectly return early from loop (inside `sendMessage`) even if service was unable to be satisfied by outbound transports.

Signed-off-by: Niall Shaw <niall.shaw@absa.africa>


TimoGlastra (approved):
Thanks for this PR @niallshaw-absa! If possible, could you add a test case for this so it won't happen again?

PR #840, opened 2022-06-08
Feature/shortenedUrl

KolbyRKunz:
Adds support for attempting to parse a URL as a shortened URL after failing to parse a URL normally, This follows the most recent changes to the [`out-of-band` rfc](https://github.com/hyperledger/aries-rfcs/tree/main/features/0434-outofband) with the know caveats involved.


KolbyRKunz:
This branch was for some reason based on a old branch that has been merged already and that is the reason behind the commits mentioning pickup v2

packages/core/src/modules/oob/messages/OutOfBandInvitation.ts
```diff
@@ -121,7 +174,13 @@ export class OutOfBandInvitation extends AgentMessage {
   public readonly goal?: string
 
   public readonly accept?: string[]
-
+  @Transform(
```
blu3beri, line 177:
```ts
  @Transform(
     ({ value }) =>
       value.map(replaceLegacyDidSovPrefix),
     { toClassOnly: true }
   )
```

slightly cleaner imo

packages/core/src/utils/parseInvitation.ts
```diff
@@ -13,15 +10,11 @@ import { OutOfBandInvitation } from '../modules/oob/messages'
  * @returns OutOfBandInvitation
  */
 export const parseInvitationUrl = async (invitationUrl: string): Promise<OutOfBandInvitation> => {
-  const parsedUrl = parseUrl(invitationUrl).query
-  if (parsedUrl['oob']) {
+  try {
     const outOfBandInvitation = await OutOfBandInvitation.fromUrl(invitationUrl)
     return outOfBandInvitation
-  } else if (parsedUrl['c_i'] || parsedUrl['d_m']) {
+  } catch (error) {
```
blu3beri, line 16: Unused error right?

packages/core/src/modules/routing/services/MediationRecipientService.ts
```diff
@@ -69,6 +69,31 @@ export class MediationRecipientService {
...
   }
 
+  public async requestStatus(
+    config: {
+      mediatorId?: string
+      recipientKey?: string
+    } = {}
+  ) {
+    let mediator
+    let mediatorRecord
```
blu3beri, line 79: I am not a big fan of a `let` like this at the top without a declaration. Is there a way we can rewrite this function to make it a bit cleaner? I can give it a go aswell.

packages/core/src/modules/routing/services/MediationRecipientService.ts
```diff
@@ -69,6 +69,31 @@ export class MediationRecipientService {
     this.messageSender = messageSender
   }
 
+  public async requestStatus(
+    config: {
```
blu3beri, line 73: `config` can be `config?` to avoid assigning a default empty object to it.

packages/core/src/modules/oob/messages/OutOfBandInvitation.ts
```diff
@@ -71,17 +73,65 @@ export class OutOfBandInvitation extends AgentMessage {
...
+        throw new AriesFrameworkError(
+          'InvitationUrl is invalid. It needs to contain one, and only one, of the following parameters; `oob`'
+        )
+      }
+    } catch (error) {
+      return await this.fromShortUrl(invitationUrl)
+    }
+  }
+  //This currently does not follow the RFC because of issues with fetch, currently uses a janky work around
+  public static async fromShortUrl(invitationUrl: string) {
```
blu3beri, line 92: This seems almost identical to the connection fromShortUrl, which makes sense, but is there are way to unify them so we do not repeat themselves.

blu3beri (dismissed):

packages/core/src/utils/parseInvitation.ts
```diff
@@ -13,15 +10,11 @@ import { OutOfBandInvitation } from '../modules/oob/messages'
  * @returns OutOfBandInvitation
  */
 export const parseInvitationUrl = async (invitationUrl: string): Promise<OutOfBandInvitation> => {
```
TimoGlastra, line 12: I think the flow right now is:
1. check inline oob invitation url
2. try to fetch shortened oob invitation url
3. check inline connection invitation url
4. try to fetch shortened connection invitation url

Could we maybe reorder this to:
1. check inline oob invitation url
2. check inline connection invitation url
3. try to fetch shortened oob invitation url
4. try to fetch shortened connection invitation url

TimoGlastra (changes requested):
This will be a great addition to AFJ @KolbyRKunz! However I think there's some things we should clean up and improve on in this PR

Could you maybe write some tests for the parsing of the invitation with the differnt flows (oob vs connection vs inline vs shortened). You can mock `node-fetch` response object and process those, or you could set up a simple server that returns mock objects/urls, but I think it's important that we test this functionality.

packages/core/src/modules/oob/messages/OutOfBandInvitation.ts
```diff
@@ -71,17 +73,65 @@ export class OutOfBandInvitation extends AgentMessage {
...
+    // eslint-disable-next-line no-restricted-globals
+    const abortController = new AbortController()
+    const id = setTimeout(() => abortController.abort(), 15000)
+    let response
+    try {
+      response = await fetch(invitationUrl, {
+        method: 'GET',
+        headers: {
+          'Content-Type': 'application/json',
```
Iskander508, line 101: This should probably be `Accept` instead of `Content-Type`

packages/core/src/modules/connections/messages/ConnectionInvitationMessage.ts
```diff
@@ -119,24 +121,73 @@ export class ConnectionInvitationMessage extends AgentMessage {
...
+  public static async fromShortUrl(invitationUrl: string) {
+    // eslint-disable-next-line no-restricted-globals
+    const abortController = new AbortController()
+    const id = setTimeout(() => abortController.abort(), 15000)
+    let response
+    try {
+      response = await fetch(invitationUrl, {
+        method: 'GET',
+        headers: {
+          'Content-Type': 'application/json',
```
Iskander508, line 157: Here as well: `Accept` instead of `Content-Type`

packages/core/src/modules/routing/services/MediationRecipientService.ts
```diff
@@ -69,6 +69,31 @@ export class MediationRecipientService {
     this.messageSender = messageSender
   }
 
+  public async requestStatus(
```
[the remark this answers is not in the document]
KolbyRKunz: That was my bad, when I create this branch locally it was not based on main, looks like I created my own problem to fix with that.

packages/core/src/modules/routing/services/MediationRecipientService.ts
KolbyRKunz: This was a result of me not being careful when creating the branch that brought this method into this PR, Timo mentioned below that this was removed because it causes issues with the pick up process. I am removing it from this PR as well

packages/core/src/modules/oob/messages/OutOfBandInvitation.ts
```diff
@@ -121,7 +174,7 @@ export class OutOfBandInvitation extends AgentMessage {
   public readonly goal?: string
 
   public readonly accept?: string[]
-
+  @Transform(({ value }) => value.map(replaceLegacyDidSovPrefix), { toClassOnly: true })
```
TimoGlastra, line 177: BTW just noticed this value is optional so I think we may need to do something like:

```suggestion
  @Transform(({ value }) => value?.map(replaceLegacyDidSovPrefix), { toClassOnly: true })
```

Could you maybe add a test for this (test with value both defined and undefined also with legacy dids) in `OutOfBandInvitation.test.ts`?

packages/core/src/modules/routing/services/MediationRecipientService.ts
KolbyRKunz: This code was deprecated, my branch being based on an old branch brought it back in, I have removed this section.

packages/core/src/modules/oob/messages/OutOfBandInvitation.ts
KolbyRKunz: These have been combined into one function

packages/core/src/modules/connections/messages/ConnectionInvitationMessage.ts
```diff
@@ -1,5 +1,7 @@
 import { Transform } from 'class-transformer'
 import { ArrayNotEmpty, IsArray, IsOptional, IsString, IsUrl, ValidateIf } from 'class-validator'
+// eslint-disable-next-line import/no-extraneous-dependencies
+import fetch from 'node-fetch'
```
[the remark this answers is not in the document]
KolbyRKunz: I have moved the fromShortUrl function into parseInvitation and am passing the agentDependencies into the function to avoid the fetch error.

packages/core/src/modules/oob/messages/OutOfBandInvitation.ts
KolbyRKunz: I am not entirely sure how to write tests for this scenario, some direction on how to write this would be helpful if you want me to attempt to add this to this PR. @TimoGlastra

KolbyRKunz: I was able to get some assistance on this matter and believe that the test case was covered. @TimoGlastra

TimoGlastra:
I've added this to the Agenda for the WG call today, would be great if one of you @KolbyRKunz @JamesKEbert could attend!

KolbyRKunz:
After testing the trinsic shortener we found that if short url leads to a deep link fetch does not automatically follow the redirect and the workflow is at intended in the RFC, we added a statement that checks whether we followed the redirect or not and grabs the encrypted URL according to redirect status. This was a small change but also has us follow the RFC more closely. 

Additionally we have noted that the trinsic shortener doe not follow the rfc in that if you supply the `content-type` as `application/json` we DO NOT get the invitation as json back from the fetch.

packages/core/src/utils/__tests__/shortenedUrl.test.ts
```diff
@@ -0,0 +1,114 @@
+import type { Response } from 'node-fetch'
+
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { Headers } from 'node-fetch'
```
TimoGlastra, line 4: If this doesn't work we should add it to the dev dependencies

```suggestion
import type { Headers } from 'node-fetch'
```

packages/core/src/utils/__tests__/shortenedUrl.test.ts
```diff
@@ -0,0 +1,114 @@
...
+    const short = await fromShortUrl(mockedResponseConnectionJson)
+    expect(short).toEqual(connectionInvitationToNew)
+  })
+})
+describe('shortUrlConnectionUrl', () => {
+  test('connectionFromUrl', async () => {
+    const short = await fromShortUrl(mockedResponseConnectionUrl)
+    expect(short).toEqual(connectionInvitationToNew)
+  })
+})
```
TimoGlastra, line 114: can we make the descriptions of the tests a bit more descriptive? I find it hard to understand what each of the different tests does

packages/core/src/utils/parseInvitation.ts
```diff
@@ -1,10 +1,122 @@
...
+  try {
+    response = await dependencies.fetch(invitationUrl, {
+      method: 'GET',
+      headers: {
+        Accept: 'application/json',
+        'Content-Type': 'application/json',
+      },
+    })
+  } catch (error) {
+    throw new AriesFrameworkError('Get request failed on provided Url')
```
TimoGlastra, line 29: We'll lose all information on why this failed here. Maybe we can include some of the error details here:

```suggestion
    throw new AriesFrameworkError(`Get request failed on provided url: ${error.message}`, { cause: error })
```

packages/core/src/utils/parseInvitation.ts
```diff
@@ -1,10 +1,122 @@
...
+          const invitation = JsonTransformer.fromJSON(invitationJson, ConnectionInvitationMessage)
+
+          await MessageValidator.validateSync(invitation)
+
+          return convertToNewInvitation(invitation)
+        }
+        throw new AriesFrameworkError(
+          'InvitationUrl is invalid. Needs to be encrypted with either c_i, d_m, or oob or must be valid shortened URL'
+        )
+      }
```
TimoGlastra, line 84: Can't we use the already available `parseInvitationUrl` for this?

TimoGlastra (changes requested):
This is great. Left some nits on the naming / structure of the code, but the contents look good and look forward to merging this PR!

packages/core/src/utils/parseInvitation.ts
```diff
@@ -1,10 +1,122 @@
...
+/**
+ * Parses URL containing encoded invitation and returns invitation message.
+ *
+ * @param invitationUrl URL containing encoded invitation
+ *
+ * @param dependencies Agent dependicies containing fetch
+ *
+ * @returns OutOfBandInvitation
+ */
+export const parseInvitationShortUrl = async (
```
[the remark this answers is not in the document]
KolbyRKunz: I can see doing that in a update that bring in breaking changes to the framework but for this PR I see having the two different functions to bring in the functionality as the best route. I do agree that providing an options parameter would be the best way to merge the two methods into one though. @TimoGlastra

packages/core/src/utils/parseInvitation.ts
KolbyRKunz: Yes, that is a good point

packages/core/src/utils/parseInvitation.ts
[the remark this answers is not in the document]
TimoGlastra: Ah right, we discussed that during the WG call. Makes sense 👍. So what would be the approach for receiving an shortened invitation url? I think we did say we could add it to the receiveInvitationFromUrl method right?

packages/core/package.json
```diff
@@ -49,6 +49,7 @@
     "web-did-resolver": "^2.0.8"
   },
   "devDependencies": {
+    "node-fetch": "^3.2.6",
```
TimoGlastra, line 52: node-fetch 3 uses esm only, so we should probably stick to 2.x (the version that is used in the @aries-framework/node package)

KolbyRKunz:
@TimoGlastra Got the the receiveInvitationFromUrl has been updated and the node version changed to greater than ^2.0. If there is not anything else I think this is ready to be merged.

TimoGlastra (approved):

PR #843, opened 2022-06-09
fix: Fix for Credential records created before adding protocol v2

Przytua:
Credential records created before adding protocol v2 don't have protocol version set, so `credentialRecord.protocolVersion` returns `undefined`, and `getService` also returned undefined. Old credentials couldn't be deleted due to this issue.

### DIFF
--- a/packages/core/src/modules/credentials/CredentialsModule.ts
+++ b/packages/core/src/modules/credentials/CredentialsModule.ts
@@ -113,7 +113,7 @@ export class CredentialsModule implements CredentialsModule {
     this.logger.debug(`Initializing Credentials Module for agent ${this.agentConfig.label}`)
   }
 
-  public getService(protocolVersion: CredentialProtocolVersion): CredentialService {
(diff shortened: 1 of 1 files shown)


TimoGlastra:
@Przytua There's been quite some breaking changes to storage in 0.2.0. We resolve those using a migration script that can be run once updated to 0.2.0.

See https://github.com/hyperledger/aries-framework-javascript/blob/main/docs/migration/updating.md and https://github.com/hyperledger/aries-framework-javascript/blob/main/docs/migration/0.1-to-0.2.md for more information

Przytua:
> @Przytua There's been quite some breaking changes to storage in 0.2.0. We resolve those using a migration script that can be run once updated to 0.2.0.
> 
> See https://github.com/hyperledger/aries-framework-javascript/blob/main/docs/migration/updating.md and https://github.com/hyperledger/aries-framework-javascript/blob/main/docs/migration/0.1-to-0.2.md for more information

Thanks for letting me know! Didn't notice that earlier.

PR #870, opened 2022-06-16
feat: delete associated didCommMessages when deleting credential records

### LINKED ISSUE #728: Deleting a credential record should delete the associated didcomm messages
When deleting a record from the service or module level it should delete all associated didcomm messages. 

We can do this by finding all `DidCommMessageRecord` where the `associatedRecordId` matches with the record that will be deleted.

TODO:
- [ ] Add a new config property `deleteAssociatedDidCommMessages` to `DeleteCredentialOptions`.
- [ ] In `CredentialService.delete` check if `deleteAssociatedDidCommMessage` is true (should be true by default), if so delete the assocaited didcomm messages by first retrieving them:
	- `didCommMessageRepository.findByQuery({ associatedRecordId: credentialRecord.id })`
	- delete all records returned from the query

morrieinmaas:

closes #728

### DIFF
--- a/packages/core/src/modules/credentials/protocol/v1/__tests__/V1CredentialServiceCred.test.ts
+++ b/packages/core/src/modules/credentials/protocol/v1/__tests__/V1CredentialServiceCred.test.ts
@@ -11,7 +11,7 @@ import { EventEmitter } from '../../../../../agent/EventEmitter'
 import { InboundMessageContext } from '../../../../../agent/models/InboundMessageContext'
 import { Attachment, AttachmentData } from '../../../../../decorators/attachment/Attachment'
 import { AriesFrameworkError } from '../../../../../error'
-import { DidCommMessageRole } from '../../../../../storage'
+import { DidCommMessageRecord, DidCommMessageRole } from '../../../../../storage'
 import { DidCommMessageRepository } from '../../../../../storage/didcomm/DidCommMessageRepository'
 import { JsonTransformer } from '../../../../../utils'
 import { JsonEncoder } from '../../../../../utils/JsonEncoder'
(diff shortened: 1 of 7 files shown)


packages/core/src/modules/credentials/services/CredentialService.ts
```diff
@@ -190,13 +194,23 @@ export abstract class CredentialService<CFs extends CredentialFormat[]> {
...
         const formatService = this.getFormatServiceForRecordType(credential.credentialRecordType)
         await formatService.deleteCredentialById(credential.credentialRecordId)
       }
     }
+
+    if (deleteAssociatedDidCommMessages) {
+      const didCommMessages = await this.didCommMessageRepository.findByQuery({
+        associatedRecordId: credentialRecord.id,
+      })
+      for (const msgRecord of didCommMessages) {
```
TimoGlastra, line 210:
```suggestion
      for (const didCommMessage of didCommMessages) {
```

packages/core/src/modules/credentials/protocol/v2/__tests__/V2CredentialServiceCred.test.ts
```diff
@@ -790,6 +803,16 @@ describe('CredentialService', () => {
...
+    })
+    it('deleteAssociatedDidCommMessages should default to true', async () => {
+      const deleteCredentialMock = mockFunction(indyCredentialFormatService.deleteCredentialById)
+
+      const credentialRecord = mockCredentialRecord()
+      mockFunction(credentialRepository.getById).mockResolvedValue(credentialRecord)
+
+      await credentialService.delete(credentialRecord)
+
       expect(deleteCredentialMock).toHaveBeenNthCalledWith(1, credentialRecord.credentials[0].credentialRecordId)
```
TimoGlastra, line 816: This doesn't check if the didcomm messages are deleted. We can check this using `expect(didCommMessageRepository.delete).toHaveBeenCalledTimes(3)` and then checking if it has called it for each of the record returned by findByQuery

packages/core/src/modules/credentials/protocol/v1/__tests__/V1CredentialServiceCred.test.ts
```diff
@@ -846,6 +859,16 @@ describe('V1CredentialService', () => {
...
+    })
+    it('deleteAssociatedDidCommMessages should default to true', async () => {
+      const deleteCredentialMock = mockFunction(indyCredentialFormatService.deleteCredentialById)
+
+      const credentialRecord = mockCredentialRecord()
+      mockFunction(credentialRepository.getById).mockResolvedValue(credentialRecord)
+
+      await credentialService.delete(credentialRecord)
+
       expect(deleteCredentialMock).toHaveBeenNthCalledWith(1, credentialRecord.credentials[0].credentialRecordId)
```
TimoGlastra, line 872: same here doesn't check whether didcomm messages are deleted

PR #878, opened 2022-06-17
ci: fix lerna version detection based on tags

TimoGlastra:
Fixes lerna version detection for next release. Lerna will use the latest tag to determine the latest released version. As we create a tag for each commit we need to remove all pre-release tags so we can determine the last stable release. It also excludes the private packages that have been added since 0.1.0 (extension-module and demo) from the release / changelog creation.


karimStekelenburg (approved):
Nice job! Very needed 👏

PR #881, opened 2022-06-19
refactor!: add agent context

TimoGlastra:
This is a big pull request, but doesn't change any functionality of the agent. All services, repositories etc.. now use agent context so we can start using them as stateless services for multitenancy (as described here: https://hackmd.io/vGLVlxLvQR6jsEEjzNcL8g#Agent-Context).

This doesn't change the public API of the framework, but because of the large number of changes and also the implications for custom modules I've marked this as a breaking change.

Will add more detailed migration docs later on, but generally you have to update your modules to now inject the agent context object and pass that around to services. See the `DummyModule` for an example: https://github.com/TimoGlastra/aries-framework-javascript/blob/fb26a4dfe656b38da984ed46faf5b5a5f3cda1df/samples/extension-module/dummy/DummyModule.ts

### DIFF
--- a/packages/core/src/agent/Dispatcher.ts
+++ b/packages/core/src/agent/Dispatcher.ts
@@ -1,13 +1,13 @@
-import type { Logger } from '../logger'
 import type { OutboundMessage, OutboundServiceMessage } from '../types'
 import type { AgentMessage } from './AgentMessage'
 import type { AgentMessageProcessedEvent } from './Events'
 import type { Handler } from './Handler'
 import type { InboundMessageContext } from './models/InboundMessageContext'
 
-import { AgentConfig } from '../agent/AgentConfig'
+import { InjectionSymbols } from '../constants'
 import { AriesFrameworkError } from '../error/AriesFrameworkError'
-import { injectable } from '../plugins'
+import { Logger } from '../logger'
+import { injectable, inject } from '../plugins'
 import { canHandleMessageType, parseMessageType } from '../utils/messageType'
 
 import { ProblemReportMessage } from './../modules/problem-reports/messages/ProblemReportMessage'
@@ -23,10 +23,14 @@ class Dispatcher {
   private eventEmitter: EventEmitter
   private logger: Logger
 
-  public constructor(messageSender: MessageSender, eventEmitter: EventEmitter, agentConfig: AgentConfig) {
+  public constructor(
+    messageSender: MessageSender,
+    eventEmitter: EventEmitter,
+    @inject(InjectionSymbols.Logger) logger: Logger
+  ) {
     this.messageSender = messageSender
     this.eventEmitter = eventEmitter
-    this.logger = agentConfig.logger
+    this.logger = logger
   }
 
   public registerHandler(handler: Handler) {
@@ -70,7 +74,7 @@ class Dispatcher {
     }
 
     if (outboundMessage && isOutboundServiceMessage(outboundMessage)) {
-      await this.messageSender.sendMessageToService({
+      await this.messageSender.sendMessageToService(messageContext.agentContext, {
         message: outboundMessage.payload,
         service: outboundMessage.service,
         senderKey: outboundMessage.senderKey,
@@ -78,11 +82,11 @@ class Dispatcher {
       })
     } else if (outboundMessage) {
       outboundMessage.sessionId = messageContext.sessionId
-      await this.messageSender.sendMessage(outboundMessage)
+      await this.messageSender.sendMessage(messageContext.agentContext, outboundMessage)
     }
 
     // Emit event that allows to hook into received messages
-    this.eventEmitter.emit<AgentMessageProcessedEvent>({
+    this.eventEmitter.emit<AgentMessageProcessedEvent>(messageContext.agentContext, {
       type: AgentEventTypes.AgentMessageProcessed,
       payload: {
         message: messageContext.message,
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -1,6 +1,7 @@
 // reflect-metadata used for class-transformer + class-validator
 import 'reflect-metadata'
 
+export { AgentContext } from './agent/AgentContext'
 export { Agent } from './agent/Agent'
 export { EventEmitter } from './agent/EventEmitter'
 export { Handler, HandlerInboundMessage } from './agent/Handler'
--- a/packages/core/src/modules/connections/ConnectionsModule.ts
+++ b/packages/core/src/modules/connections/ConnectionsModule.ts
@@ -1,10 +1,9 @@
 import type { DependencyManager } from '../../plugins'
-import type { Key } from '../dids'
 import type { OutOfBandRecord } from '../oob/repository'
 import type { ConnectionRecord } from './repository/ConnectionRecord'
 import type { Routing } from './services'
 
-import { AgentConfig } from '../../agent/AgentConfig'
+import { AgentContext } from '../../agent'
 import { Dispatcher } from '../../agent/Dispatcher'
 import { MessageSender } from '../../agent/MessageSender'
 import { createOutboundMessage } from '../../agent/helpers'
@@ -17,14 +16,14 @@ import { RoutingService } from '../routing/services/RoutingService'
 
 import { DidExchangeProtocol } from './DidExchangeProtocol'
 import {
+  AckMessageHandler,
   ConnectionRequestHandler,
   ConnectionResponseHandler,
-  AckMessageHandler,
-  TrustPingMessageHandler,
-  TrustPingResponseMessageHandler,
+  DidExchangeCompleteHandler,
   DidExchangeRequestHandler,
   DidExchangeResponseHandler,
-  DidExchangeCompleteHandler,
+  TrustPingMessageHandler,
+  TrustPingResponseMessageHandler,
 } from './handlers'
 import { HandshakeProtocol } from './models'
 import { ConnectionRepository } from './repository'
@@ -34,7 +33,6 @@ import { TrustPingService } from './services/TrustPingService'
 @module()
 @injectable()
 export class ConnectionsModule {
-  private agentConfig: AgentConfig
   private didExchangeProtocol: DidExchangeProtocol
   private connectionService: ConnectionService
   private outOfBandService: OutOfBandService
@@ -43,10 +41,10 @@ export class ConnectionsModule {
   private routingService: RoutingService
   private didRepository: DidRepository
   private didResolverService: DidResolverService
+  private agentContext: AgentContext
 
   public constructor(
     dispatcher: Dispatcher,
-    agentConfig: AgentConfig,
     didExchangeProtocol: DidExchangeProtocol,
     connectionService: ConnectionService,
     outOfBandService: OutOfBandService,
@@ -54,9 +52,9 @@ export class ConnectionsModule {
     routingService: RoutingService,
     didRepository: DidRepository,
     didResolverService: DidResolverService,
-    messageSender: MessageSender
+    messageSender: MessageSender,
+    agentContext: AgentContext
   ) {
-    this.agentConfig = agentConfig
     this.didExchangeProtocol = didExchangeProtocol
     this.connectionService = connectionService
     this.outOfBandService = outOfBandService
@@ -65,6 +63,8 @@ export class ConnectionsModule {
     this.didRepository = didRepository
     this.messageSender = messageSender
     this.didResolverService = didResolverService
+    this.agentContext = agentContext
+
     this.registerHandlers(dispatcher)
   }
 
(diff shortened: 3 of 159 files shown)


packages/core/src/agent/Dispatcher.ts
```diff
@@ -24,10 +24,14 @@ class Dispatcher {
...
-  public constructor(messageSender: MessageSender, eventEmitter: EventEmitter, agentConfig: AgentConfig) {
+  public constructor(
+    messageSender: MessageSender,
+    eventEmitter: EventEmitter,
+    @inject(InjectionSymbols.Logger) logger: Logger
+  ) {
     this.messageSender = messageSender
     this.eventEmitter = eventEmitter
-    this.logger = agentConfig.logger
+    this.logger = logger
```
blu3beri, line 34: I am a bit confused by this. Maybe I am missing something about the context, but could we not get the logger from `agentContext.logger`? 

Checking a bit through this PR it seems a bit "messy" on how we deal with the logger instance. We get it via the injection but it is also in the `agentConfig`. I am not too knowledgeable on dependency injection with tsyringe, but is this really the best way to deal with the logger instance?

Also, do we want each agent to have their instance of a logger or do we want to use a `RootContainer` (if I understand your design correctly) logger instance?

packages/core/src/index.ts
```diff
@@ -1,6 +1,7 @@
 // reflect-metadata used for class-transformer + class-validator
 import 'reflect-metadata'
 
+export { AgentContext } from './agent/AgentContext'
```
blu3beri, line 4: Why do we want to expose the `AgentContext` publicly?

packages/core/src/modules/connections/ConnectionsModule.ts
```diff
@@ -217,31 +231,11 @@ export class ConnectionsModule {
    * @param connectionId the connection record id
    */
   public async deleteById(connectionId: string) {
-    return this.connectionService.deleteById(connectionId)
-  }
-
-  public async findByKeys({ senderKey, recipientKey }: { senderKey: Key; recipientKey: Key }) {
```
blu3beri, line 223: Is there any reason why this method is deleted? I assume it does not really have anything to do with the `AgentContext` PR?

packages/core/src/modules/connections/DidExchangeProtocol.ts
```diff
@@ -164,7 +166,7 @@ export class DidExchangeProtocol {
       )
     }
 
-    const didDocument = await this.extractDidDocument(message)
+    const didDocument = await this.extractDidDocument(messageContext.agentContext, message)
```
blu3beri, line 169: Why exactly is the agentContext inside the messageContext? Just curious.

packages/core/src/modules/connections/handlers/DidExchangeResponseHandler.ts
```diff
@@ -95,10 +97,14 @@ export class DidExchangeResponseHandler implements Handler {
     // TODO: should we only send complete message in case of autoAcceptConnection or always?
     // In AATH we have a separate step to send the complete. So for now we'll only do it
     // if auto accept is enable
```
blu3beri, line 99: Not in your pr, but: 

```suggestion
    // if auto accept is enabled
```

packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -550,22 +560,22 @@ export class ConnectionService {
...
-    const connectionRecord = await this.getById(connectionId)
-    return this.connectionRepository.delete(connectionRecord)
+  public async deleteById(agentContext: AgentContext, connectionId: string) {
+    const connectionRecord = await this.getById(agentContext, connectionId)
+    return this.connectionRepository.delete(agentContext, connectionRecord)
   }
 
-  public async findSingleByQuery(query: { did: string; theirDid: string }) {
-    return this.connectionRepository.findSingleByQuery(query)
+  public async findByDids(agentContext: AgentContext, query: { ourDid: string; theirDid: string }) {
```
blu3beri, line 577: Is there a reason why we call it `ourDid`? I believe ACA-Py uses `my_did`? I have no preference, more just curious.

packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -576,33 +586,56 @@ export class ConnectionService {
...
    */
-  public getByThreadId(threadId: string): Promise<ConnectionRecord> {
-    return this.connectionRepository.getByThreadId(threadId)
+  public async getByThreadId(agentContext: AgentContext, threadId: string): Promise<ConnectionRecord> {
+    return this.connectionRepository.getByThreadId(agentContext, threadId)
   }
 
-  public async findByTheirDid(did: string): Promise<ConnectionRecord | null> {
-    return this.connectionRepository.findSingleByQuery({ theirDid: did })
+  public async findByTheirDid(agentContext: AgentContext, did: string): Promise<ConnectionRecord | null> {
```
blu3beri, line 593: For consistency ;)

```suggestion
  public async findByTheirDid(agentContext: AgentContext, theirDid: string): Promise<ConnectionRecord | null> {
```

packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -576,33 +586,56 @@ export class ConnectionService {
...
+    return this.connectionRepository.findSingleByQuery(agentContext, { theirDid: did })
   }
 
-  public async findByOurDid(did: string): Promise<ConnectionRecord | null> {
-    return this.connectionRepository.findSingleByQuery({ did })
+  public async findByOurDid(agentContext: AgentContext, did: string): Promise<ConnectionRecord | null> {
```
blu3beri, line 597:
```suggestion
  public async findByOurDid(agentContext: AgentContext, ourDid: string): Promise<ConnectionRecord | null> {
```

packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -576,33 +586,56 @@ export class ConnectionService {
...
   }
 
-  public async findByInvitationDid(invitationDid: string) {
-    return this.connectionRepository.findByQuery({ invitationDid })
+  public async findByInvitationDid(agentContext: AgentContext, invitationDid: string) {
+    return this.connectionRepository.findByQuery(agentContext, { invitationDid })
   }
 
-  public async createConnection(options: ConnectionRecordProps): Promise<ConnectionRecord> {
+  public async findByKeys(
```
blu3beri, line 609: I see the method is just moved to the service. Ignore my comment from before :)

blu3beri (commented):
I have reviewed the first 36 files and will continue somewhere tomorrow (it is quite a bit).

But I do have a comment and that is about the function signatures. I believe we can make every function signature way more consistent by having a common pattern.

Seeing that most functions now take an `AgentContext` it would nice to have something like this:


```ts
const afjFunc = (agentContext: AgentContext, parameters: {...}, options?: {...}) => { ... }


// or 
const afjFunc = (parameters: {...}, options?: {...}) => { ... }
```

With this we will always have named parameters and have a nice division between the parameters of a function and the configuration/options we supply to it (like `shouldDeleteCredential` or `timeOut`). I am interested to hear what you think about this structure and whether it is something we want to implement in AFJ. I think it has value, but it does require a big breaking change. 

I am not too sure if we want to also add this to the public api, might be worth discussing.

packages/core/src/modules/connections/services/ConnectionService.ts
TimoGlastra: Not sure, @jakubkoci added this

packages/core/src/modules/connections/DidExchangeProtocol.ts
TimoGlastra: Because each message can have their own context. Handlers, services and repositories are now all stateless. The agent context gives the context to process the message. So if we receive a message we'll find the correct tenant, assign the agent context for that tenant to the message context and then process the message. 

We could make it a separate parameter, but it kinda made sense to me to tie the agent context to a specific message context. Would you prefer to make it a separate parameter?

packages/core/src/index.ts
TimoGlastra: If we want to inject it (in extension modules for example) we need it. So it's not necessary for the public api for users, but is needed if you want to write extension modules.

packages/core/src/agent/Dispatcher.ts
TimoGlastra: Yeah that's something I also wasn't sure about, so good that you're calling it out. 

It's in the agent config because we configure it in the agent init config. However my thought was that the logger would be shared between the different tenants. I'm not sure if it would work well if we have a lot of logger instances, and there's ways to different context of the different agents in logs.

As mentioned in the design document we should probably look at a way to separate agent config (global, used by all tenants) and tenant config (specific to that tenant).

What would be your preferred approach here?

TimoGlastra:
> Seeing that most functions now take an `AgentContext` it would nice to have something like this:
> 
> ```ts
> const afjFunc = (agentContext: AgentContext, parameters: {...}, options?: {...}) => { ... }
> 
> 
> // or 
> const afjFunc = (parameters: {...}, options?: {...}) => { ... }
> ```

Not 100% sure what you mean by this. This is a pattern already used by some methods, do you mean we should enforce it for all methods? How would we do something like that?

I'm also not sure it will work for all methods. It can be difficult sometimes to differentiate between what parameters and what options are (in the case of the `shouldDeleteCredential` it's clear, but I think there's a lot of methods where it's not that clear. I'm also not sure if it will improve the API for all methods. E.g. this:

```ts
agent.credentials.offerCredential({
  connectionId: 'the-connection-id',
  autoAcceptCredential: AutoAcceptCredential.Always,
  credentialFormats: {
  /* formats */
  }
})
```

than this:

```ts
agent.credentials.offerCredential({
  connectionId: 'the-connection-id',
  credentialFormats: {
  /* formats */
  }
}, {
  autoAcceptCredential: AutoAcceptCredential.Always
})
```

Anyway it seems this would be better addressed to me in a separate issue, except if this involves moving the `AgentContext` around. I made it the first parameter to each method now for consistency, would you change that?

blu3beri:
> Anyway it seems this would be better addressed to me in a separate issue, except if this involves moving the `AgentContext` around. I made it the first parameter to each method now for consistency, would you change that?

No the agentContext as first param is a good choice. My point is to get more consistency within the framework. TBH, I do think that the second method with the object for `autoAccept` is better. And 100% it will be difficult, but I think that finding a clear division between parameters and config/options might cleanup the functions aswell.

packages/core/src/agent/Dispatcher.ts
blu3beri: Yeah I think a global configuration, maybe a Provider/manager agent aswell, and tenant specific configuration would be a good solution.

Quickly looking through the `AgentConfig` I do not see a lot that should be globally defined. That being said, I still think it has value for future configuration.

packages/core/src/modules/connections/DidExchangeProtocol.ts
blu3beri: Its a difficult one. I think for consistency a seperate `agentContext` might be nicer, but I see your point of containing the agentContext inside the messageContext. I am okay with keeping it like this.

packages/core/src/modules/ledger/IndyPool.ts
```diff
@@ -31,14 +32,20 @@ export class IndyPool {
...
   public authorAgreement?: AuthorAgreement | null
 
-  public constructor(agentConfig: AgentConfig, poolConfig: IndyPoolConfig) {
-    this.indy = agentConfig.agentDependencies.indy
+  public constructor(
+    poolConfig: IndyPoolConfig,
+    agentDependencies: AgentDependencies,
+    logger: Logger,
+    stop$: Subject<boolean>,
+    fileSystem: FileSystem
```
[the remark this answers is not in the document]
TimoGlastra: This is the fileSystem instance, while the agentDependencies holds the fileSystem class. I'm slowly trying to move away from the agent dependencies as it's not a modular approach, we should inject dependencies separately.

packages/core/src/storage/migration/updates/0.1-0.2/mediation.ts
```diff
@@ -1,6 +1,6 @@
-import type { V0_1ToV0_2UpdateConfig } from '.'
 import type { Agent } from '../../../../agent/Agent'
 import type { MediationRecord } from '../../../../modules/routing'
+import type { V0_1ToV0_2UpdateConfig } from './index'
```
blu3beri, line 3: Does this type come from the index file?

packages/core/tests/mocks/MockWallet.ts
```diff
@@ -0,0 +1,61 @@
...
+import type { Buffer } from '../../src/utils/buffer'
+import type { DidInfo, UnpackedMessageContext, DidConfig } from '../../src/wallet'
+
+export class MockWallet implements Wallet {
+  public publicDid = undefined
+  public isInitialized = true
+  public isProvisioned = true
+
+  public create(walletConfig: WalletConfig): Promise<void> {
+    throw new Error('Method not implemented.')
```
blu3beri, line 13: Are the not defined on purpose and jest will be used with mocking?

blu3beri (approved):
Great work @TimoGlastra! I do think that we should not mark this as `refactor` as there are many many changes to functions and the way we use them.

I left a couple of notes but nothing blocking. 

We should also merge it probably after the 0.2.0 release.

packages/core/tests/mocks/MockWallet.ts
TimoGlastra: These are not defined on purpose indeed. A lot of tests we just mock the calls to the wallet and it shouldn't call the actual methods.

packages/core/src/storage/migration/updates/0.1-0.2/mediation.ts
TimoGlastra: Yes, do you think we should extract it to a seperate file?

TimoGlastra:
Superseded by #920 (which is based on 0.3.0-pre instead of main)

PR #882, opened 2022-06-19
fix(connections): allow ; to convert legacy did

TimoGlastra:
Older dids use `;` as an identifier separator instead of the now used `#`. So an id of `did:sov:123;indy` would be normalized to `#did:sov:123;indy`. We now check when converting a legacy did doc to a new did doc whether an `;` character is in place and the `#` isn't and we will handle it correclty so it parses to `#indy`


packages/core/src/modules/connections/services/helpers.ts
```diff
@@ -82,8 +82,14 @@ export function convertToNewDidDocument(didDoc: DidDoc): DidDocument {
 }
 
 function normalizeId(fullId: string): `#${string}` {
-  const [, id] = fullId.split('#')
+  // Some old dids use `;` as the delimiter for the id. If we can't find a `#`
+  // and a `;` exists, we will parse everything after `;` as the id.
+  if (!fullId.includes('#') && fullId.includes(';')) {
```
blu3beri, line 87: This parser only works if there is a single `;` in the id. Is that correct?

packages/core/src/modules/connections/__tests__/ConnectionService.test.ts
```diff
@@ -209,7 +209,7 @@ describe('ConnectionService', () => {
       const processedConnection = await connectionService.processRequest(messageContext, outOfBand)
 
       expect(processedConnection.state).toBe(DidExchangeState.RequestReceived)
-      expect(processedConnection.theirDid).toBe('did:peer:1zQmfPPbuG8vajHvYjGUW8CN5k9rLuuMmYSGBYwJqJDDUS72')
+      expect(processedConnection.theirDid).toBe('did:peer:1zQmW2esSyEVGzrh3CFt1eQZUHEAb3Li1hyPudPhSoFevrFY')
```
blu3beri, line 212: Why did this DID change?

blu3beri (approved):

packages/core/src/modules/connections/services/helpers.ts
TimoGlastra: The only usage I've seen is with one `;`. I think it was used before the `#` and you can only include one. What we could do is leave the other ones in place (like url parsers do):

image

So we would get `#indy;anotherword` instead of `#indy` if the id was `something;indy;anotherword`. Do you think that is better?

blu3beri: I think leaving them behind, which follows the url parser you showed, is good. I am very unaware of the security implications (likely none) here though.

PR #886, opened 2022-06-19
feat(vc): delete w3c credential record

karimStekelenburg:
This PR adds a method to the W3cCredentialService to remove a credential record from storage. I also renamed a few properties and methods because their previous names were somewhat misleading.

- feat: add method to remove w3c credential
- refactor: rename methods and props for consistancy


karimStekelenburg:
Because the CI is still testing against Node v14 the tests are failing. Since the 0.3.0-pre branch itself already contains the Node v14 incompatible code I guess it's OK to merge this one, but I'd like to double check. @TimoGlastra, thoughts?

TimoGlastra:
You mean Node 12?

This has already been removed in main, so if we rebase 0.3.0 with main we can see if CI now fully passess

packages/core/src/modules/vc/W3cCredentialService.ts
```diff
@@ -336,23 +336,27 @@ export class W3cCredentialService {
     return w3cCredentialRecord
   }
 
-  public async getAllCredentials(): Promise<W3cVerifiableCredential[]> {
-    const allRecords = await this.w3cCredentialRepository.getAll()
-    return allRecords.map((record) => record.credential)
+  public async removeCredentialRecord(id: string) {
+    const cred = await this.w3cCredentialRepository.getById(id)
+    await this.w3cCredentialRepository.delete(cred)
```
TimoGlastra, line 341: nit, but we try to avoid abbreviations as much as possible

```suggestion
    const credential = await this.w3cCredentialRepository.getById(id)
    await this.w3cCredentialRepository.delete(credential)
```

packages/core/src/modules/vc/W3cCredentialService.ts
```diff
@@ -336,23 +336,27 @@ export class W3cCredentialService {
...
   }
 
-  public async getCredentialById(id: string): Promise<W3cVerifiableCredential> {
-    return (await this.w3cCredentialRepository.getById(id)).credential
+  public async getAllCredentialRecords(): Promise<W3cCredentialRecord[]> {
```
TimoGlastra, line 344: We don't use this in all other services/modules (just getAll instead of getAllCredentials, getAllCredentialRecords). Do you think this helps? If so we should probably change this in all methods

packages/core/src/modules/vc/__tests__/W3cCredentialService.test.ts
```diff
@@ -249,27 +264,62 @@ describe('W3cCredentialService', () => {
         expect(result.verified).toBe(true)
       })
     })
-    describe('storeCredential', () => {
-      it('should store a credential', async () => {
+    describe('Credential Storage', () => {
```
TimoGlastra, line 267: can you also add a test for the delete method?

packages/core/src/modules/vc/W3cCredentialService.ts
karimStekelenburg: I get your point in terms of consistency. However, I consciously  did this because the `W3cCredentialService` deals with more than credential records. For example, there is a `signPresentation()` method and a `getSignatureSuitesForCredential()`. It seems a bit odd to me to specify the data model in some method names, but leave it out in others.

I'm fine with changing it though. However, in that case it would make sense to me to rename  `W3cCredentialService` to `W3cCredentialRecordService`. Then at least the class name gives you a hint of what `getAll()` actually returns.

TimoGlastra:
Merging as only failing test is the bbs test

PR #887, opened 2022-06-20
feat(credentials): find didcomm message methods

TimoGlastra:
Adds `findProposalMessage`, `findOfferMessage`, `findRequestMessage`, `findCredentialMessage` methods to the credentials module to be able to access the did comm message associated with a credential exchange.


packages/core/src/modules/credentials/CredentialsModule.ts
```diff
@@ -69,6 +69,12 @@ export interface CredentialsModule<CFs extends CredentialFormat[], CSs extends C
   getById(credentialRecordId: string): Promise<CredentialExchangeRecord>
   findById(credentialRecordId: string): Promise<CredentialExchangeRecord | null>
   deleteById(credentialRecordId: string, options?: DeleteCredentialOptions): Promise<void>
+
+  // DidComm Message Records
+  findProposalMessage(credentialExchangeId: string): Promise<ReturnType<CSs[number]['findProposalMessage']>>
```
blu3beri, line 74: Thats quite a return type. Do you think an alias would be a worthwhile benefit? It looks quite ambigous about what it returns here (I am sure any IDE will supply a nice return type).

blu3beri (approved):

packages/core/src/modules/credentials/CredentialsModule.ts
```diff
@@ -548,4 +554,38 @@ export class CredentialsModule<
...
+    const service = this.getService(credentialExchangeRecord.protocolVersion)
+
+    return service.findRequestMessage(credentialExchangeId)
+  }
+
+  public async findCredentialMessage(
+    credentialExchangeId: string
+  ): Promise<ReturnType<CSs[number]['findProposalMessage']>> {
+    const credentialExchangeRecord = await this.getById(credentialExchangeId)
+    const service = this.getService(credentialExchangeRecord.protocolVersion)
```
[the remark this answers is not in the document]
morrieinmaas: maybe `getService(credentialExchangeId)` ?

morrieinmaas (commented):
left one suggestion but overall LGTM!

packages/core/src/modules/credentials/CredentialsModule.ts
TimoGlastra: Yes we can probably make this an alias. Let me take a look. It is indeed to make the return type return `V1ProposalMessage | V2ProposalMessage` instead of `AgentMessage`

PR #888, opened 2022-06-20
fix: remove usage of const enum

### LINKED ISSUE #884: const enums can't be used after code is transpiled
We've started using const enums in a few places throughout the framework. Although lighter, I've noticed they can't be used after code is transpiled (in react native which uses a custom ts compiler I think).

Due to this I think it may be better to keep using normal enums for now (to stay consistent across the framework). We can then later look at using other types of enum methods that can be used outside of AFJ (in consumer code).

more info: https://ncjamieson.com/dont-export-const-enums/

TimoGlastra:
Fixes #884, see the issue for why we can't use const enums


blu3beri (approved):

blu3beri:
Interesting, never knew about this. Is there maybe an eslint rule to disable constant enums?

TimoGlastra:
Added an eslint rule to disallow const enums

morrieinmaas (commented):
lgtm

blu3beri (approved):

PR #890, opened 2022-06-20
chore: move to docs repo

TimoGlastra:
Should not be merged until https://aries.js.org is live.

Removes all docs that have been moved to the Aries JavaScript docs repo. Have kept some getting started docs in place until we can add sections for that in the aries javascript docs repo.

Related PR in https://github.com/animo/aries-javascript-docs/pull/30 (should be merged first)


blu3beri (approved):
Approved. You can merge it when aries.js.org is live :).

PR #893, opened 2022-06-20
test(credentials): fix flaky tests with events

TimoGlastra:
Fixes the flaky e2e test for icv2. We only start event listeners action the has been performed (sendProposal -> start proposal received listener) while we should be starting the listener before performing the action (start proposal received listener -> sendProposal) to prevent flaky tests.

I've fixed it for one file now, but we should look at all ICv2 and PPv2 tests (that's where we use evens A LOT).


morrieinmaas (commented):
LGTM. Nice n tidy

morrieinmaas:
@TimoGlastra needs some TLC regarding the checks pipeline

blu3beri (approved):

TimoGlastra:
@2byrds will take a look, @TimoGlastra will provide some pointers

2byrds:
The concept/changes seem straightforward. Go through our tests and make sure we start the listener receiver before executing the action.
A broader topic for me is the e2e testing in AFJ (and this module).  I'm very familiar with unit/integration/e2e testing in other languages. I've done some initial learning on Jest.  Are there any specific guides about the AFJ e2e testing environment that I should be aware of? If not, should I create a guide for the JS docs as I come up to speed?

2byrds:
> The concept/changes seem straightforward. Go through our tests and make sure we start the listener receiver before executing the action. A broader topic for me is the e2e testing in AFJ (and this module). I'm very familiar with unit/integration/e2e testing in other languages. I've done some initial learning on Jest. Are there any specific guides about the AFJ e2e testing environment that I should be aware of? If not, should I create a guide for the JS docs as I come up to speed?

I'll start here https://github.com/hyperledger/aries-framework-javascript/blob/main/DEVREADME.md

2byrds:
I'm comfortable with the test setup, etc.  So I started a Pull Request against @TimoGlastra branch associated with this PR.  Once his branch is up-to-date there should be no difference (yet) since i added his repo as a remote and merged.

Happy to submit to his repo.  Or let me know if you would prefer that we modify this PR or start a new PR for my branch.

See https://github.com/TimoGlastra/aries-framework-javascript/pull/17

packages/core/src/modules/credentials/protocol/v2/__tests__/v2-credentials.e2e.test.ts
```diff
@@ -91,7 +100,7 @@ describe('v2 credentials', () => {
     })
 
     testLogger.test('Faber waits for credential proposal from Alice')
-    faberCredentialRecord = await waitForCredentialRecord(faberAgent, {
+    faberCredentialRecord = await waitForCredentialRecordSubject(faberReplay, {
```
2byrds, line 103: I've been considering your changes in this test and the test below where 'Faber waits for credential proposal from Alice'.  Here you use **waitForCredentialRecordSubject(faberReplay**.... vs. in the test below where you made the change to create the Promise from **waitForCredentialRecord(faberAgent**.... why the different calls?

TimoGlastra:
Sorry for the delayed response! I think it would be better to open a PR directly to AFJ with all changes. This way the contribution will be attributed to your GH account instead of mine :)

2byrds:
> Sorry for the delayed response! I think it would be better to open a PR directly to AFJ with all changes. This way the contribution will be attributed to your GH account instead of mine :)

Sounds good.  I assume the signed commits would still track the author properly but a separate PR is easy enough.

packages/core/src/modules/credentials/protocol/v2/__tests__/v2-credentials.e2e.test.ts
2byrds: Okay thank you for clarification! Yes, makes sense.  I'll look through how we query/test for events between the tests so that i can generalize the pattern to other tests.

TimoGlastra:
Closing in favour of https://github.com/hyperledger/aries-framework-javascript/pull/966

PR #898, opened 2022-06-21
feat!: add agent context provider

TimoGlastra:
> - Dependant on #881 (until that one is merged this is kinda hard to review)
> - Design: https://hackmd.io/vGLVlxLvQR6jsEEjzNcL8g#Agent-Context-Provider

Adds an `AgentContextProvider` with default `DefaultAgentContextProvider` implementation that always returns the same agent context (for a single tenant agent). 

This allows the tenant module (will be added as a separate module) to implement its own `TenantAgentContextProvider` that will find provide the agent context for a specific tenant. This allows to keep the core lean, while providing more advanced usages of providing the agent context.

I added a `contextCorrelationId` to the `AgentContext` interface. For now this is not really needed, but this will become really useful when we add tenants. The `contextCorrelationId` allows to correlate the context with an identifier to make it easier to provide the context for an incoming message. By default we find the tenant based on the recipient public keys of a message, but there's numerous cases where we are processing plaintext messages. By providing the contextCorrelationId we can use this in the `AgentContextProvider` to correlate the message to a specific context. In the case of the `TenantContextProvider` the `contextCorrelationId` will be the `tenantId`.

BREAKING CHANGE: the `agent.receiveMessage` method should only be used if you want to process the message in the context of the current agent. This means it fine to use for e.g. processing plaintext connection-less messages, but it shouldn't be used anymore by outbound and inbound transports to process messages. Instead, inject the `MessageReceiver` class and call `receiveMessage` on that class instead.


TimoGlastra:
Superseded by #921

PR #906, opened 2022-06-24
feat(credentials): added credential sendProblemReport method

Iskander508:

Adding a convenience method to send problem report as part of RFC's 0036 and 0453 unhappy flow.
It is a complementary implementation to the `agent.proofs.sendProblemReport`. But for credentials it was still missing.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -99,7 +99,7 @@ module.exports = {
       },
     },
     {
-      files: ['*.test.ts', '**/__tests__/**', '**/tests/**', 'jest.*.ts', 'samples/**', 'demo/**'],
+      files: ['*.test.ts', '**/__tests__/**', '**/tests/**', 'jest.*.ts', 'samples/**', 'demo/**', 'scripts/**'],
       env: {
         jest: true,
         node: false,
--- a/.github/workflows/continuous-deployment.yml
+++ b/.github/workflows/continuous-deployment.yml
@@ -31,7 +31,9 @@ jobs:
 
       # On push to main, release unstable version
       - name: Release Unstable
-        run: yarn lerna publish --loglevel=verbose --canary minor --exact --force-publish --yes --no-verify-access --dist-tag alpha
+        run: |
+          export NEXT_VERSION_BUMP=$(./node_modules/.bin/ts-node ./scripts/get-next-bump.ts)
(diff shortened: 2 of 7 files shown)


packages/core/src/modules/credentials/CredentialsModule.ts
```diff
@@ -67,6 +67,7 @@ export interface CredentialsModule<CFs extends CredentialFormat[], CSs extends C
 
   // Credential
   acceptCredential(options: AcceptCredentialOptions): Promise<CredentialExchangeRecord>
+  sendProblemReport(credentialRecordId: string, message: string): Promise<CredentialExchangeRecord>
```
TimoGlastra, line 70: Maybe we can also make this an object like we do with the other methods that contain both the `credentialRecordId` and `message`? Mainly for two reasons:
1. Consistency with other methods
2. It would allow to add more options in the future without needing to make a breaking change (we'll probably want to make it an options object when adding more options)

packages/core/src/modules/credentials/services/CredentialService.ts
```diff
@@ -84,6 +84,9 @@ export abstract class CredentialService<CFs extends CredentialFormat[] = Credent
   // methods for ack
   abstract processAck(messageContext: InboundMessageContext<AgentMessage>): Promise<CredentialExchangeRecord>
 
+  // methods for problem-report
+  abstract createProblemReport(message: string): ProblemReportMessage
```
TimoGlastra, line 88: Same here, if we make it an object it would be easier to add new properties later on (such as exposing the `code`)

TimoGlastra (commented):
This is great @Iskander508!

packages/core/src/modules/credentials/CredentialsModule.ts
Iskander508: @TimoGlastra refactored

packages/core/src/modules/credentials/services/CredentialService.ts
Iskander508: refactored

TimoGlastra (approved):
🚀

TimoGlastra:
@Iskander508 I'll merge this once #908 is merged

PR #909, opened 2022-06-28
feat(routing): add routing service

TimoGlastra:
Adds a routing service that handles the creating of keys for routing and integrates with the mediation recipient service to add routing for the mediator.

The extraction is just to make sure we don't have to depend on a mediator dependency all across the framework. 

What this also adds is an event for when routing keys are created. This allows the tenant module to listen for created routing keys and can then create a mapping of it in the base wallet. I first started with a middleware approach (As described in the design document: https://hackmd.io/vGLVlxLvQR6jsEEjzNcL8g), but this added _ A LOT_ of complexity and didn't abstract away the mediator functionality as that is a part of the core API. So I went for the less generic but simpler API, that is just fine I think.

Again no changes to the public api, so we can merge this without issues


blu3beri (approved):
LGTM.

PR #913, opened 2022-06-29
feat: fetch verification method types by proof type

karimStekelenburg:

### DIFF
--- a/packages/core/src/modules/vc/SignatureSuiteRegistry.ts
+++ b/packages/core/src/modules/vc/SignatureSuiteRegistry.ts
@@ -1,3 +1,5 @@
+import type { KeyType } from '../../crypto'
+
 import { AriesFrameworkError } from '../../error'
 import { injectable, injectAll } from '../../plugins'
 
@@ -9,7 +11,8 @@ export const SignatureSuiteToken = Symbol('SignatureSuiteToken')
 export interface SuiteInfo {
   suiteClass: typeof LinkedDataSignature
   proofType: string
-  keyTypes: string[]
+  verificationMethodTypes: string[]
+  keyTypes: KeyType[]
 }
 
 @injectable()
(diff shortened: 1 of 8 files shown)


packages/core/src/modules/vc/W3cCredentialService.ts
```diff
@@ -352,6 +352,10 @@ export class W3cCredentialService {
     return result.map((record) => record.credential)
   }
 
+  public getKeyTypeByProofType(proofType: string): KeyType {
+    return this.suiteRegistry.getKeyTypeByProofType(proofType)
+  }
```
TimoGlastra, line 357: a proof type could potentially have multiple supported key types associated with it. Should we make this an array?

TimoGlastra (changes requested):
Can you add tests for the added methods?

packages/core/src/modules/vc/W3cCredentialService.ts
karimStekelenburg: Good point, that will cause issues later on. Resolved now.

karimStekelenburg:
> Can you add tests for the added methods?

@TimoGlastra good catch, thanks. Time for a PR checklist.. I'll open a discussion about it.

packages/core/src/modules/vc/SignatureSuiteRegistry.ts
```diff
@@ -11,7 +11,7 @@ export interface SuiteInfo {
   suiteClass: typeof LinkedDataSignature
   proofType: string
   requiredKeyType: string
-  keyType: string
+  keyTypes: [KeyType]
```
TimoGlastra, line 14: This should be `KeyType[]`. This actually means an array with one item of type `KeyType` (you can use this to create tuple types)

```suggestion
  keyTypes: KeyType[]
```

TimoGlastra: Hmm after thinking about it a bit more. This is the key type as we use it in the framework, but not the key as is used in did documents. Those key types are e.g. Ed25519Key2018 (or something like it). So we still have to make the transformation, but only the signature suites know which actual types is supports. 

Should this be updated to include the supported key types as used by did documents? E.g. if the suites only supports 2018 and 2020 ed25519 keys and there's a new 2022 one that isn't supported we should only extract the 2018 and 2022 keys from the did document, as that's the only ones we can sign with.

TimoGlastra:
BBS tests failing in Node.JS 17/18 But otherwise tests are passing. Merging

PR #919, opened 2022-07-03
feat: add base agent class

TimoGlastra:
Adds a `BaseAgent` class that does most of the heavy lifting for the agent. `Agent` extends this and registers all dependencies in the `registerDependencies` method. 

When adding the tenant module, I'll also extend the `BaseAgent` for the `TenantAgent` class. This allows the `Agent` class to do all setup that only needs to happen once for the base agent, and the tenant agent can expose the same api without doing this registration setup.

Relatively small PR, but a lot of changes due to dependency on #898 and #881


TimoGlastra:
Superseded by #921

PR #920, opened 2022-07-03
refactor!: add agent context

TimoGlastra:

This is the same PR as #881, but based off the 0.3.0-pre branch instead of main. There's no changes compared to #881 except for making it work with this branch (making the new w3c services stateless). Due to the breaking changes we don't want to have this in main just yet. 

BREAKING CHANGE: To make AFJ multi-tenancy ready, all services and repositories have been made stateless. A new `AgentContext` is introduced that holds the current context, which is passed to each method call. The public API hasn't been affected, but due to the large impact of this change it is marked as breaking.

### DIFF
--- a/packages/core/src/agent/Agent.ts
+++ b/packages/core/src/agent/Agent.ts
@@ -2,13 +2,13 @@ import type { Logger } from '../logger'
 import type { InboundTransport } from '../transport/InboundTransport'
 import type { OutboundTransport } from '../transport/OutboundTransport'
 import type { InitConfig } from '../types'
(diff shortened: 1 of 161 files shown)


blu3beri (approved):

TimoGlastra:
> I assume the test will succeed once #935 is merged.

Yes, locally that's the only tests failing

TimoGlastra:
Merging as the only failed test is the one as fixed in https://github.com/hyperledger/aries-framework-javascript/pull/935 so we can keep the PR review going

PR #921, opened 2022-07-03
feat: add agent context provider

TimoGlastra:
This is the same PR as #898, but based off the 0.3.0-pre branch instead of main. There's no changes compared to #898

Dependant on:
- #920


packages/core/src/agent/context/AgentContext.ts
```diff
@@ -0,0 +1,45 @@
...
+  /**
+   * Dependency manager holds all dependencies for the current context. Possibly a child of a parent dependency manager,
+   * in which case all singleton dependencies from the parent context are also available to this context.
+   */
+  public readonly dependencyManager: DependencyManager
+
+  /**
+   * An identifier that allows to correlate this context across usages. An example of the contextCorrelationId could be
+   * the id of the `TenantRecord` that is associated with this context. The AgentContextProvider can use this identifier to
+   * correlate an inbound message to a specific context (if the message is not encrypted, it's impossible to correlate it to a tenant)
```
blu3beri, line 17: `(if the message is not encrypted, it's impossible to correlate it to a tenant)`

Could you eloborate on this a bit more, why not?

blu3beri (approved):
The main of this PR seems the addition of the `contextCorrelationId` which I do not understand fully. 

Like, could it be the GitHub name of some tenant (terrible example)? So that you pass `blu3beri` as the `contextCorrelateId` to resolve the correct `agentContext` quicker? (I assume the key word here is quicker, right?)

TimoGlastra:
> Hmm, it seems that the tests here also fail like in #935

Yes, this is still the issue as described in #923. All PRs against the 0.3.0-pre branch will fail CI currently

packages/core/src/agent/context/AgentContext.ts
TimoGlastra: If we have a plaintext didcomm message like the following it's impossible to know which tenant this message is for. There's no information that correlates this to a specific tenant. 

```json
{
    "@type": "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/issue-credential/1.0/propose-credential",
    "@id": "<uuid-of-propose-message>",
    "comment": "some comment",
    "credential_proposal": <json-ld object>,
    "schema_id": "Schema ID string",
    "cred_def_id": "Credential Definition ID string"
}
```

If we have an encrypted message however, we can look at the recipient public keys of the messages. We have a mapping internally that registers which public keys belong to which tenant so we can route the message to the correct tenant.

In cases of out of band messages we often already know which context we're in because we receive it through the agent API, not through the inbound transport, so that allows us to just pass the contextCorrelationId

blu3beri: Got it. That makes sense to me.

blu3beri:
> If we can agree on the best appraoch I'll make sure to update the document of the AgentContext to make it more clear.

I think that this approach is good and maybe a bit more docs, maybe with an example, will be enough.

TimoGlastra:
Merging as the only failing test is the one as described in #923

PR #930, opened 2022-07-05
feat: OOB public did

Iskander508:

First attempt to add support for receiving OOB invitations with public did.

Currently this is working together with the aca-py agent, when  oob invitation is created using:
```
POST /out-of-band/create-invitation
{
  "handshake_protocols": [
    "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/connections/1.0"
  ],
  "use_public_did": true,
  ...
}
```

* refactored `OutOfBandInvitation.services` to tackle the issues with `String` coming from the transformer. Exposing only a getter `getServices` method to get simplified representation.
* resolving public DID services procedure moved from the `MessageSender` into the `DidResolverService` as it needs to be performed when sending the connection request as well as handling the connection response
  * We should maybe think about some caching to not run the resolution multiple times
* The service `recipientKey` lookup procedure had to be adjusted, to work together with aca-py. There's a strange Ed25519/X25519 key specification historical convention (https://sovrin-foundation.github.io/sovrin/spec/did-method-spec-template.html#did-document-notes)

TODO: 
* check OOB invitation with `didexchange` protocol
* fix tests


packages/core/src/modules/oob/messages/OutOfBandInvitation.ts
```diff
@@ -99,13 +99,21 @@ export class OutOfBandInvitation extends AgentMessage {
...
-    return this.services
-      .filter((s): s is OutOfBandDidCommService => typeof s !== 'string' && !(s instanceof String))
+    return this.getServices()
+      .filter((s): s is OutOfBandDidCommService => typeof s !== 'string')
       .map((s) => s.recipientKeys)
       .reduce((acc, curr) => [...acc, ...curr], [])
       .map((didKey) => DidKey.fromDid(didKey).key)
   }
 
+  // shorthand for services without the need to deal with the String DIDs
```
TimoGlastra, line 109: Great addition! I was struggling with the class transformer way too much and settled for String instances...

packages/core/src/agent/MessageSender.ts
```diff
@@ -341,49 +332,6 @@ export class MessageSender {
     throw new AriesFrameworkError(`Unable to send message to service: ${service.serviceEndpoint}`)
   }
 
-  private async retrieveServicesFromDid(did: string) {
```
TimoGlastra, line 344: Ideally I would like to keep the did module separated as much as possible from didcomm. Maybe we can look at adding a didcomm directory and a didcomm did document service that handles this?

packages/core/src/modules/connections/DidExchangeProtocol.ts
```diff
@@ -222,9 +222,9 @@ export class DidExchangeProtocol {
     if (routing) {
       services = this.routingToServices(routing)
     } else if (outOfBandRecord) {
-      const inlineServices = outOfBandRecord.outOfBandInvitation.services.filter(
-        (service) => typeof service !== 'string'
-      ) as OutOfBandDidCommService[]
+      const inlineServices = outOfBandRecord.outOfBandInvitation
+        .getServices()
+        .filter((service) => typeof service !== 'string') as OutOfBandDidCommService[]
```
TimoGlastra, line 227: maybe we can also add a `getDidServices` and `getInlineServices` (or `getDidCommServices`). This way we don't have to do the string filter here, and we already get the correct type back (string[])

packages/core/src/modules/connections/handlers/ConnectionResponseHandler.ts
```diff
@@ -65,8 +66,26 @@ export class ConnectionResponseHandler implements Handler {
     }
 
     messageContext.connection = connectionRecord
-    // The presence of outOfBandRecord is not mandatory when the old connection invitation is used
-    const connection = await this.connectionService.processResponse(messageContext, outOfBandRecord)
+
+    const invitationRecipientKeys = outOfBandRecord
+      .getTags()
+      .recipientKeyFingerprints.map((fingerprint) => Key.fromFingerprint(fingerprint))
```
TimoGlastra, line 72: I think we should store the invitation recipient keys with public did keys from the start rather than in the .

My proposal would be to move the `recipientKeyFingerprints` from the default tags (generated on save) to custom tags that we must set ourselves. When processing the invitation we can set all recipient keys. The advantage is that we only need to do it once for both connection and didexchagne protocol, and we also don't need this or functionality (we now only add the did recipient keys if we don't have any inline didcomm service keys).

packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -121,22 +134,9 @@ export class ConnectionService {
...
-    const connectionRequest = new ConnectionRequestMessage({
-      label: label ?? this.config.label,
-      did: didDoc.id,
-      didDoc,
-      imageUrl: imageUrl ?? this.config.connectionImageUrl,
-    })
-
-    if (autoAcceptConnection !== undefined || autoAcceptConnection !== null) {
-      connectionRecord.autoAcceptConnection = config?.autoAcceptConnection
-    }
```
TimoGlastra, line 137: It looks like we're losing this line

packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -257,7 +257,7 @@ export class ConnectionService {
    */
   public async processResponse(
     messageContext: InboundMessageContext<ConnectionResponseMessage>,
-    outOfBandRecord: OutOfBandRecord
+    invitationRecipientKeys: Key[]
```
TimoGlastra, line 260: if we already set the `recipientKeyFingerprints` tag correctly, we don't have to pass the invitationRecipientkeys here, we can just extract them from the oob record

packages/core/src/modules/oob/OutOfBandModule.ts
```diff
@@ -563,13 +567,12 @@ export class OutOfBandModule {
...
-      // We need to check if the service is an instance of string because of limitations from class-validator
-      if (typeof didOrService === 'string' || didOrService instanceof String) {
-        // TODO await this.connectionsModule.findByTheirDid()
-        throw new AriesFrameworkError('Dids are not currently supported in out-of-band invitation services attribute.')
+      let did: string
+      if (typeof didOrService === 'string') {
+        did = didOrService
+      } else {
+        did = outOfBandServiceToNumAlgo2Did(didOrService)
       }
```
TimoGlastra, line 575: We could also call invitationDids for this: 

```ts
const invitationDids = outOfBandInvitation.invitationDids

for (const invitationDid of invitationDids) {
....
}
```

Otherwise this may look a bit cleaner (but the 1st approach should work fine):

```suggestion
      const did = typeof didOrService === 'string' ? didOrService : outOfBandServiceToNumAlgo2Did(didOrService)
```

packages/core/src/modules/oob/OutOfBandModule.ts
```diff
@@ -622,19 +625,34 @@ export class OutOfBandModule {
...
     // The framework currently supports only older OOB messages with `~service` decorator.
     // TODO: support receiving messages with other services so we don't have to transform the service
     // to ~service decorator
     const [service] = services
 
     if (typeof service === 'string') {
-      throw new AriesFrameworkError('Dids are not currently supported in out-of-band invitation services attribute.')
+      const didDocument = await this.didResolverService.resolveDidDocument(service)
+      const documentService = didDocument?.service?.[0]
+      serviceEndpoint = documentService?.serviceEndpoint
```
TimoGlastra, line 640:
```suggestion
      const documentService = didDocument.service?.[0]
      serviceEndpoint = documentService?.serviceEndpoint
```

packages/core/src/modules/oob/OutOfBandModule.ts
```diff
@@ -622,19 +625,34 @@ export class OutOfBandModule {
...
+      recipientKeys = didDocument?.recipientKeys?.map((key) => key.publicKeyBase58)
```
TimoGlastra, line 641: This is missing some handling of cases (routingKeys, resolving recipient keys). Maybe we can extract the logic from this method into smaller methods and reuse that in both cases: https://github.com/Iskander508/aries-framework-javascript/blob/oob_public_did/packages/core/src/modules/dids/services/DidResolverService.ts#L79-L134

packages/core/tests/oob.test.ts
```diff
@@ -658,12 +658,11 @@ describe('out of band', () => {
...
-      const { message } = await faberAgent.credentials.createOffer(credentialTemplate)
-      const { outOfBandInvitation } = await faberAgent.oob.createInvitation({
-        ...issueCredentialConfig,
-        messages: [message],
+      const outOfBandInvitation = new OutOfBandInvitation({
+        label: 'test-connection',
+        handshakeProtocols: [HandshakeProtocol.DidExchange],
+        services: ['somedid'],
       })
-      outOfBandInvitation.services = ['somedid']
```
TimoGlastra, line 666: This test is not applicable anymore?

TimoGlastra (changes requested):
Thanks for this @Iskander508! Left some notes / comments on how we may approach it a bit differently, but overall I like the direction of this.

packages/core/src/modules/connections/services/ConnectionService.ts
Iskander508: This was probably a coding mistake. The condition `autoAcceptConnection !== undefined || autoAcceptConnection !== null` is always true. And the `connectionRecord.autoAcceptConnection` was already initialized with the same value `config?.autoAcceptConnection` before.

packages/core/src/agent/MessageSender.ts
Iskander508: I have tried to split it into a `packages/core/src/modules/didcomm/services/DidCommDocumentService`. Have a look if that's what you mean

packages/core/src/modules/oob/OutOfBandModule.ts
Iskander508: changed to reuse the `resolveServicesFromDid` method

packages/core/src/modules/connections/handlers/ConnectionResponseHandler.ts
Iskander508: Yes, I get the point and fully agree that it would be better to store the recipient keys at the start of the connection process. However I'm not sure where exactly to do that. To get the recipient keys, we need to run the did resolution for services from the ledger (which is quite costly operation).
I thought of 2 options:
1. Either to do it right when creating the OOB record in `OutOfBandModule.receiveInvitation`. Which would mean to run an additional resolution. Also I'm not sure what to put there in case when we are creating the invitation. `OutOfBandModule.createInvitation`
2. Store the recipientKeys into the record once we have already resolved them (when sending the connection request message in `MessageSender.retrieveServicesByConnection`): https://github.com/hyperledger/aries-framework-javascript/blob/49f8a1b9f9e1c9223755f39f5445ff2fc89e8627/packages/core/src/agent/MessageSender.ts#L363)

I'm also not familiar with the concept of the custom/default tags to be honest. Currently the recipientKeyFingerprints are retrieved from the OOB invitation object: https://github.com/hyperledger/aries-framework-javascript/blob/49f8a1b9f9e1c9223755f39f5445ff2fc89e8627/packages/core/src/modules/oob/repository/OutOfBandRecord.ts#L69 Does it mean they are not saved as tags? Or am I missing something?

packages/core/tests/oob.test.ts
Iskander508: true, this case is not applicable anymore

packages/core/src/agent/MessageSender.ts
TimoGlastra: Yes, this is great. We can now move more items from the agent directory to the didcomm module over time.

packages/core/src/modules/connections/handlers/ConnectionResponseHandler.ts
TimoGlastra: > Which would mean to run an additional resolution

Where would we need the additional resolution for?

> I'm also not familiar with the concept of the custom/default tags to be honest. Currently the recipientKeyFingerprints are retrieved from the OOB invitation object:

We have two types of tags in AFJ; default and custom. Default tags are set in the `getTags` method on a record. This can contain derived properties (such as the recipientKeysFingerprints`, but can't be async. This worked with inline services, but if we want to get the recipientKeyFingerprints for the public dids we need to do the async resolving. What we can do is remove it from the default tags (so we're not setting it anymore in the getTags method), and rather set it manaually using `record.setTag('recipientKeyFingerprints', theKeys)`.

You would have to create a new type `CustomOutOfBandRecordTags` type, and add it here: https://github.com/hyperledger/aries-framework-javascript/blob/49f8a1b9f9e1c9223755f39f5445ff2fc89e8627/packages/core/src/modules/oob/repository/OutOfBandRecord.ts#L26-L33

```
type DefaultOutOfBandRecordTags = {
  role: OutOfBandRole
  state: OutOfBandState
  invitationId: string
}

type CustomOutOfBandRecordTags = {
  recipientKeyFingerprints?: string[]
}

export class OutOfBandRecord extends BaseRecord<DefaultOutOfBandRecordTags, CustomOutOfBandRecordTags> {
```

packages/core/src/modules/connections/services/ConnectionService.ts
TimoGlastra: You're right 👍

packages/core/src/modules/oob/OutOfBandModule.ts
```diff
@@ -644,19 +642,37 @@ export class OutOfBandModule {
...
+
     // The framework currently supports only older OOB messages with `~service` decorator.
     // TODO: support receiving messages with other services so we don't have to transform the service
     // to ~service decorator
     const [service] = services
 
     if (typeof service === 'string') {
-      throw new AriesFrameworkError('Dids are not currently supported in out-of-band invitation services attribute.')
+      const didServices = await this.didCommDocumentService.resolveServicesFromDid(service)
+      const didService = didServices?.[0]
```
TimoGlastra, line 656:
```suggestion
      const [didService] = await this.didCommDocumentService.resolveServicesFromDid(service)
```

packages/core/src/modules/oob/messages/OutOfBandInvitation.ts
```diff
@@ -100,13 +100,26 @@ export class OutOfBandInvitation extends AgentMessage {
 
   // TODO: this only takes into account inline didcomm services, won't work for public dids
   public getRecipientKeys(): Key[] {
-    return this.services
-      .filter((s): s is OutOfBandDidCommService => typeof s !== 'string' && !(s instanceof String))
+    return this.getInlineServices()
       .map((s) => s.recipientKeys)
       .reduce((acc, curr) => [...acc, ...curr], [])
       .map((didKey) => DidKey.fromDid(didKey).key)
   }
```
TimoGlastra, line 107: We should probably remove this method as it only takes into account inline didcomm services.

packages/core/tests/oob.test.ts
```diff
@@ -172,7 +172,7 @@ describe('out of band', () => {
       expect(outOfBandInvitation.getRequests()).toBeUndefined()
 
       // expect contains services
-      const [service] = outOfBandInvitation.services as OutOfBandDidCommService[]
+      const [service] = outOfBandInvitation.getServices() as OutOfBandDidCommService[]
```
TimoGlastra, line 175:
```suggestion
      const [service] = outOfBandInvitation.getInlineServices()
```

packages/core/tests/oob.test.ts
```diff
@@ -220,7 +220,7 @@ describe('out of band', () => {
       expect(outOfBandInvitation.getRequests()).toHaveLength(1)
 
       // expect contains services
-      const [service] = outOfBandInvitation.services as OutOfBandDidCommService[]
+      const [service] = outOfBandInvitation.getServices() as OutOfBandDidCommService[]
```
TimoGlastra, line 223:
```suggestion
      const [service] = outOfBandInvitation.getInlineServices()
```

packages/core/tests/oob.test.ts
```diff
@@ -467,8 +467,8 @@ describe('out of band', () => {
       const outOfBandRecord2 = await faberAgent.oob.createInvitation(makeConnectionConfig)
 
       // Take over the recipientKeys from the first invitation so they match when encoded
-      const firstInvitationService = outOfBandRecord.outOfBandInvitation.services[0] as OutOfBandDidCommService
-      const secondInvitationService = outOfBandRecord2.outOfBandInvitation.services[0] as OutOfBandDidCommService
+      const firstInvitationService = outOfBandRecord.outOfBandInvitation.getServices()[0] as OutOfBandDidCommService
```
TimoGlastra, line 470:
```suggestion
      const [firstInvitationService] = outOfBandRecord.outOfBandInvitation.getInlineServices()
```

packages/core/tests/oob.test.ts
```diff
@@ -467,8 +467,8 @@ describe('out of band', () => {
...
+      const secondInvitationService = outOfBandRecord2.outOfBandInvitation.getServices()[0] as OutOfBandDidCommService
```
TimoGlastra, line 471:
```suggestion
      const [secondInvitationService] = outOfBandRecord2.outOfBandInvitation.getInlineServices()
```

packages/core/src/modules/connections/handlers/ConnectionResponseHandler.ts
Iskander508: Great, thank you for the explanation of the tags.

> > Which would mean to run an additional resolution
>
> Where would we need the additional resolution for?

I meant if we now additionally run a resolution right in the beggining when creating the oob record, that would be the additional resolution. I'll try with the approach n°2, to set the recipientKeyFingerprints after sending the credential request message (for the case of public did invite), otherwise set it immediately when creating the oob record from the invitation's inline services.

TheTreek:
> @TheTreek @JamesKEbert Please take a look at this PR

All of this looks really good to me.

packages/core/src/modules/oob/messages/OutOfBandInvitation.ts
Iskander508: It is still used on a few places. I'm not sure if it makes sense to remove it at this moment. What would be the replacement for the few spots, where it is used?

TimoGlastra: Where is it used? The reason I'd like to remove it is that we should be taking into account public dids everywhere now. This methods makes it quite easy to forget the public did recipient keys, which can cause issues later

Iskander508:
> It would be good to add some more tests, but otherwise this is close to being ready to merge. Did you make any breaking changes in this PR? If we can avoid any breaking changes and postpone that to 0.3.0 we can get this in 0.2.x and have it released a lot sooner

I plan to take a look at the tests to increase coverage of the new cases.
I don't think there are any breaking changes in the API, unless we count the changes in the tags.

packages/core/src/modules/oob/messages/OutOfBandInvitation.ts
Iskander508: Maybe it's better if you check for yourself:
1. `DidExchangeProtocol.processResponse`: https://github.com/hyperledger/aries-framework-javascript/blob/main/packages/core/src/modules/connections/DidExchangeProtocol.ts#L303 (not sure how to deal with this one)
2. `OutOfBandRecord` initial tags in constructor: Here we store the recipient keys from the inline services (this we could probably refactor by expanding the function in place)
3. `migrateToOobRecord`: https://github.com/hyperledger/aries-framework-javascript/blob/main/packages/core/src/storage/migration/updates/0.1-0.2/connection.ts#L315 (here it looks like we have to deal with inline services only as well, so probably easy to expand)

TimoGlastra: Ok so I think we can remove / refactor them all:

1. This checks if the response is 'authorized' by the keys from the invitation. I think we can use the recipientKeyFingerprints from the tag for this and transform it into keys (`recipientKeysFingerprints.map(fingerprint => Key.fromFingerprint(fingerprint).publicKeyBase58)` )
2. I think we can remove the default tags assignment in the constructor. We should just set in once with the processing of both the dids and the inline services (it's redundant to set it in the constructor I think?)
3. You're right we only have to deal with inline services here. Maybe we can copy the implementation directly to the migration file.

packages/core/src/agent/MessageSender.ts
```diff
@@ -399,14 +356,29 @@ export class MessageSender {
...
+            if (!oldRecipientKeyFingerprints?.length) {
+              const allRecipientKeys = didCommServices.reduce<Key[]>(
+                (aggr, { recipientKeys }) => [...aggr, ...recipientKeys],
+                []
+              )
+              outOfBand.setTag(
+                'recipientKeyFingerprints',
+                allRecipientKeys.map((key) => key.fingerprint)
+              )
+              await this.outOfBandRepository.update(outOfBand)
```
TimoGlastra, line 380: Did you place this here to prevent needing to resolve the did(s) multiple times? If so we should probably add some priority to caching

It feels like this maybe better placed inside the createInvitation method. What do you think?

packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -291,7 +287,10 @@ export class ConnectionService {
     // as the recipient key(s) in the connection invitation message
     const signerVerkey = message.connectionSig.signer
 
-    const invitationKey = Key.fromFingerprint(outOfBandRecord.getTags().recipientKeyFingerprints[0]).publicKeyBase58
+    const invitationRecipientKeys = outOfBandRecord
+      .getTags()
+      .recipientKeyFingerprints.map((fingerprint) => Key.fromFingerprint(fingerprint))
+    const invitationKey = invitationRecipientKeys[0]?.publicKeyBase58
```
TimoGlastra, line 293: This will transform all keys, but we'll only use the first one. Is there a reason for this change?

packages/core/src/modules/didcomm/services/DidCommDocumentService.ts
```diff
@@ -0,0 +1,78 @@
...
+                recipientKeyReference !== method.id ? keyReferenceToKey(didDocument, method.id) : null
+              )
+              .find((matchingKey) => {
+                if (matchingKey?.keyType !== KeyType.Ed25519) return false
+                const keyX25519 = Key.fromPublicKey(convertPublicKeyToX25519(matchingKey.publicKey), KeyType.X25519)
+                return keyX25519.publicKeyBase58 === key.publicKeyBase58
+              })
+            if (matchingEd25519Key) return matchingEd25519Key
+          }
+          return key
```
TimoGlastra, line 64: This is great!  Some small notes on the implementation:

- I think the `verificationMethod` as well as the `keyAgreement` can contain x25519 keys 
- Maybe we can extract this into a separate util in the didcomm module so its' easier to test and keeps the `resolveServicesFromDid` "simple"

packages/core/src/agent/MessageSender.ts
Iskander508: yes, it feels unclean to put it here, but this is the actual place where the service did resolution is performed currently. Caching somewhere in the resolver might be a viable solution. But it is always tricky to say when the cache should get invalidated.

packages/core/src/modules/connections/services/ConnectionService.ts
Iskander508: reverted

packages/core/src/modules/didcomm/services/DidCommDocumentService.ts
Iskander508: lookup extracted, keyAgreement references added

packages/core/src/modules/oob/messages/OutOfBandInvitation.ts
Iskander508: I have removed the `getRecipientKeys` method and refactored the 3 cases

packages/core/src/modules/oob/repository/OutOfBandRecord.ts
```diff
@@ -56,17 +60,22 @@ export class OutOfBandRecord extends BaseRecord<DefaultOutOfBandRecordTags> {
...
+          .getInlineServices()
+          .map((s) => s.recipientKeys)
+          .reduce((acc, curr) => [...acc, ...curr], [])
+          .map((didKey) => DidKey.fromDid(didKey).key.fingerprint),
+      }
     }
   }
 
-  public getTags() {
+  public getTags(): Tags<DefaultOutOfBandRecordTags, CustomOutOfBandRecordTags> {
```
TimoGlastra, line 73: This should be inferred I think. Is it giving issues?
```suggestion
  public getTags() {
```

packages/core/src/modules/oob/repository/OutOfBandRecord.ts
```diff
@@ -56,17 +60,22 @@ export class OutOfBandRecord extends BaseRecord<DefaultOutOfBandRecordTags> {
...
       this.mediatorId = props.mediatorId
       this.reuseConnectionId = props.reuseConnectionId
-      this._tags = props.tags ?? {}
+      this._tags = props.tags ?? {
+        recipientKeyFingerprints: props.outOfBandInvitation
+          .getInlineServices()
+          .map((s) => s.recipientKeys)
+          .reduce((acc, curr) => [...acc, ...curr], [])
+          .map((didKey) => DidKey.fromDid(didKey).key.fingerprint),
+      }
```
TimoGlastra, line 69: I'm not sure we want to set this here. It doesn't take the public dids into account and will only be set if no custom tags were provided. I.e. If I create the record like this:

```ts
const record = new OutOfBandRecord({
  anotherTag: '10'
})
```

It won't set the fingerprints , which I think is not desired.

packages/core/src/agent/MessageSender.ts
```diff
@@ -399,14 +356,29 @@ export class MessageSender {
...
-        for (const service of outOfBand.outOfBandInvitation.services) {
+        for (const service of outOfBand.outOfBandInvitation.getServices()) {
           // Resolve dids to DIDDocs to retrieve services
           if (typeof service === 'string') {
-            didCommServices = await this.retrieveServicesFromDid(service)
+            this.logger.debug(`Resolving services for did ${service}.`)
+            didCommServices = await this.didCommDocumentService.resolveServicesFromDid(service)
+
+            // store recipientKeyFingerprints in the oob record
+            const oldRecipientKeyFingerprints = outOfBand.getTags().recipientKeyFingerprints
```
TimoGlastra, line 370: This will only be set if the recipientKeys haven't been set yet, however the recipient keys are set in the out of band record constructor. Does this mean it will only work if there is only a did in the service array? If we have one inline service and one did this won't work I think.

If possible I'd like to make sure we're covering all cases. After thinking about it a bit more this probably isn't the best place to store this, and maybe we're better off doing it in the out of band module and take our loss with doing two resolves of the dids for now? (so method no 1 you described). If we want to keep it this way we should somehow take into account a mix of dids and inline services

TimoGlastra (commented):
Some final remarks. What is the status of the did exchange intergration. Is that something we want to add in a separate PR?

Iskander508:
> What is the status of the did exchange intergration. Is that something we want to add in a separate PR?

I guess so. I don't know how to test the didexchange protocol. So I cannot verify if it's already working or what needs to be done.

packages/core/src/modules/oob/repository/OutOfBandRecord.ts
Iskander508: explicit types removed

packages/core/src/modules/oob/repository/OutOfBandRecord.ts
Iskander508: The resolution of recipient key fingerprints moved out of the constructor, and now the values are passed on each spot a new OOB record is created.

packages/core/src/agent/MessageSender.ts
Iskander508: This part is now removed. The recipient keys are resolved already when creating the OOB record as suggested. No caching for the resolved services is done now. Maybe it can be added later as an optimisation.

JamesKEbert:
@TheTreek has done some testing with this branch & ACA-Py, and everything is working as expected (using connections v1), except that there is a minor version mis-match between AFJ & ACA-Py (1.1 vs 1.0). IIRC you had an idea to solve this in mind @TimoGlastra? I think it probably falls outside of the scope of this PR, but it is definitely related.

swcurran:
Can you characterize the 1.0/1.1 issue? Which library is doing what?

At the Interop-a-thon, we should come out of it with all of the OOB-DID Exchange tests working between AFJ and ACA-Py, and perhaps some new tests added -- or at least defined.

TimoGlastra:
>  except that there is a minor version mis-match between AFJ & ACA-Py (1.1 vs 1.0). IIRC you had an idea to solve this in mind @TimoGlastra?

Is the version mismatch causing issues? AFJ should just allow the different version, as should ACA-Py.

TimoGlastra:
> I guess so. I don't know how to test the didexchange protocol. So I cannot verify if it's already working or what needs to be done.

Fair point. Let's merge this and add support for public dids in didexchange later on

TimoGlastra (approved):
LGTM

JamesKEbert:
> Can you characterize the 1.0/1.1 issue? Which library is doing what?

Yeah, here's my best understanding of the issue @swcurran, as described in [RFC 0003](https://github.com/hyperledger/aries-rfcs/tree/main/concepts/0003-protocols#semver-rules-for-protocols).

According to my testing--the behavior is:
* That AFJ is able to process an OOB 1.0 invitation 
  * We may want to send a warning problem report `version-with-degraded-features`). 
  * We also could have responded with a `version-not-supported` problem report if the mismatch was critical (not so in this case)
* We then use connections v1, which is support as expected within AFJ & ACA-Py. 
* Upon processing an OOB 1.0 invitation that we already have a corresponding connection for, AFJ attempts to perform a handshake reuse using an OOB 1.1 handshake-reuse. 
* ACA-Py receives this and throws an error due to not being OOB 1.0. We would expect that ACA-Py would either need to: 
  * Respond with an OOB 1.1 handshake-reuse-accepted message (if implemented in ACA-Py) 
  * Respond with a warning problem report `fields-ignored-due-to-version-mismatch` **and** respond with an OOB handshake-reuse-accepted message (although it's not clear to me in the 0003 RFC whether this should be a 1.0 or 1.1).


Here are the error logs from ACA-Py when performing connection reuse with the OOB minor version mismatch:
```
demo-agent-alice-1   | 2022-08-26 19:43:46,429 indy.crypto DEBUG unpack_message: <<< res: b'{"message":"{\\"@type\\":\\"https://didcomm.org/out-of-band/1.1/handshake-reuse\\",\\"@id\\":\\"e95b89c6-6fdc-45f0-bb8b-9c4c3117f76d\\",\\"~thread\\":{\\"thid\\":\\"e95b89c6-6fdc-45f0-bb8b-9c4c3117f76d\\",\\"pthid\\":\\"6cca1b0b-4f00-436b-8550-a18d382365d8\\"}}","recipient_verkey":"9KDHrCvstES8ia4hUL4tinX6r61qWDN74v9wu8PreHMH","sender_verkey":"2sZHisiyu4SzTyzih9ziJKcRdxE37otXnj918jGx9Cp3"}'
demo-agent-alice-1   | 2022-08-26 19:43:46,429 aries_cloudagent.transport.pack_format DEBUG Expanded message: {'@type': 'https://didcomm.org/out-of-band/1.1/handshake-reuse', '@id': 'e95b89c6-6fdc-45f0-bb8b-9c4c3117f76d', '~thread': {'thid': 'e95b89c6-6fdc-45f0-bb8b-9c4c3117f76d', 'pthid': '6cca1b0b-4f00-436b-8550-a18d382365d8'}}
demo-agent-alice-1   | 2022-08-26 19:43:46,430 aiohttp.access INFO 172.19.0.3 [26/Aug/2022:19:43:46 +0000] "POST / HTTP/1.1" 200 174 "-" "okhttp/3.12.12"
demo-agent-alice-1   | 2022-08-26 19:43:46,430 aries_cloudagent.core.conductor ERROR Exception in message handler:
demo-agent-alice-1   | Traceback (most recent call last):
demo-agent-alice-1   |   File "/home/indy/.venv/lib/python3.6/site-packages/aries_cloudagent/core/dispatcher.py", line 142, in handle_message
demo-agent-alice-1   |     message = await self.make_message(inbound_message.payload)
demo-agent-alice-1   |   File "/home/indy/.venv/lib/python3.6/site-packages/aries_cloudagent/core/dispatcher.py", line 234, in make_message
demo-agent-alice-1   |     message_cls = registry.resolve_message_class(message_type)
demo-agent-alice-1   |   File "/home/indy/.venv/lib/python3.6/site-packages/aries_cloudagent/core/protocol_registry.py", line 170, in resolve_message_class
demo-agent-alice-1   |     return ClassLoader.load_class(msg_cls)
demo-agent-alice-1   |   File "/home/indy/.venv/lib/python3.6/site-packages/aries_cloudagent/utils/classloader.py", line 97, in load_class
demo-agent-alice-1   |     if "." in class_name:
demo-agent-alice-1   | TypeError: argument of type 'NoneType' is not iterable
demo-agent-alice-1   | 2022-08-26 19:43:46,433 aries_cloudagent.core.conductor ERROR DON'T shutdown on TypeError argument of type 'NoneType' is not iterable
demo-agent-alice-1   | 2022-08-26 19:43:46,434 aries_cloudagent.core.dispatcher ERROR Handler error: Dispatcher.handle_message
demo-agent-alice-1   | Traceback (most recent call last):
demo-agent-alice-1   |   File "/home/indy/.venv/lib/python3.6/site-packages/aries_cloudagent/core/dispatcher.py", line 142, in handle_message
demo-agent-alice-1   |     message = await self.make_message(inbound_message.payload)
demo-agent-alice-1   |   File "/home/indy/.venv/lib/python3.6/site-packages/aries_cloudagent/core/dispatcher.py", line 234, in make_message
demo-agent-alice-1   |     message_cls = registry.resolve_message_class(message_type)
demo-agent-alice-1   |   File "/home/indy/.venv/lib/python3.6/site-packages/aries_cloudagent/core/protocol_registry.py", line 170, in resolve_message_class
demo-agent-alice-1   |     return ClassLoader.load_class(msg_cls)
demo-agent-alice-1   |   File "/home/indy/.venv/lib/python3.6/site-packages/aries_cloudagent/utils/classloader.py", line 97, in load_class
demo-agent-alice-1   |     if "." in class_name:
demo-agent-alice-1   | TypeError: argument of type 'NoneType' is not iterable
```

swcurran:
Dang -- that's not good. I'll get someone looking at this ASAP.  I'll open an issue in ACA-Py.

For RFC 0003, if ACA-Py doesn't support OOB 1.1, it should be respond with an OOB 1.0 handshake-reuse-accepted message, perhaps with a warning (but not likely).  

However, I would expect that ACA-Py should support OOB 1.1, but perhaps not.

The only thing I wonder is since AFJ received an OOB 1.0, should it respond with an OOB 1.0?  That would require that a protocol be able to support all "less than current" minor versions, and know what to do for each one.  Basically, that would be -- in the final processing of the message, dropping fields corresponding to each lower version.  Perhaps too much, but this conversation may happen.

JamesKEbert:
> The only thing I wonder is since AFJ received an OOB 1.0, should it respond with an OOB 1.0? That would require that a protocol be able to support all "less than current" minor versions, and know what to do for each one. Basically, that would be -- in the final processing of the message, dropping fields corresponding to each lower version. Perhaps too much, but this conversation may happen.

That's a good point, according to the RFC, `message types received with a minor version at or higher than the minimum supported and less than the current minor version are processed` (AFJ's minimum version is 0 and our current version being 1), and that's followed by: 
`ideally with a response using the same minor version of the received message`, which would indicate yes, we _should_ be responding with a 1.0 in AFJ. 


> However, I would expect that ACA-Py should support OOB 1.1, but perhaps not.

Fortunately I _think_ the changes from 1.0 to 1.1 are fairly straightforward in OOB, although I had to hunt them down in the commit history (https://github.com/hyperledger/aries-rfcs/commit/4a57f664f1e66922fe55d02c376af34636f9169c)

swcurran:
Re: hunting down the changes. I've had on my list for a while to update that RFC to put a proper change log in it. 😞

TimoGlastra:
> The only thing I wonder is since AFJ received an OOB 1.0, should it respond with an OOB 1.0? That would require that a protocol be able to support all "less than current" minor versions, and know what to do for each one. Basically, that would be -- in the final processing of the message, dropping fields corresponding to each lower version. Perhaps too much, but this conversation may happen.

That would be a lot more complex than just responding with a 1.1 oob message (so we don't have to support multiple versions). As it's allowed I think it as nice way to keep interoperabillity without overly complicating the protocol implementation

TimoGlastra:
Wohoo! Thanks for this great contribution @Iskander508!! Let's try to get a new release out this week

PR #922, opened 2022-07-03
feat: add base agent class 

TimoGlastra:
This is the same PR as #919 , but based off the 0.3.0-pre branch instead of main. There's no changes compared to #919 

Dependant on:
- #920
- #921 

Signed-off-by: Timo Glastra <timo@animo.id>

Issue #923, opened 2022-07-03
CI is broken in 0.3.0-pre branch


TimoGlastra:
Opening a dedicated issue to track progress on fixing CI in 0.3.0-pre. It's failing consistently.

blu3beri:
I will take a look at this today. 

It seems that the node-bbs-signatures fails on Linux + node 17/18 and possibly more node versions.

TimoGlastra:
Do you know if these node versions are supported? If node 17/18 are not supported there's nothing we can do about it 🤷. We should just be clear to document it in that case and skip tests for bbs in node 17/18

karimStekelenburg:
Adding the test results I previously posted in #935 here for context:

### Ubuntu
#### Specs
![CleanShot 2022-07-08 at 03 06 55](https://user-images.githubusercontent.com/18397186/177896428-f31cc549-80ce-484d-839d-f9f825c0881b.png)

#### Results for Node v14.19.3
![CleanShot 2022-07-08 at 02 52 42](https://user-images.githubusercontent.com/18397186/177896526-0a858c49-a5af-432f-b22a-b962a8639a34.png)

#### Results for Node v16.16.0
![CleanShot 2022-07-08 at 02 55 12](https://user-images.githubusercontent.com/18397186/177896604-76b5f86b-fbf1-401e-8b74-ea7be9631aad.png)

#### Results for Node v17.9.1
![CleanShot 2022-07-08 at 02 57 11](https://user-images.githubusercontent.com/18397186/177896680-d26e79be-9bc0-4533-a7c9-1576d8e59789.png)

#### Results for Node v18.5.0
![CleanShot 2022-07-08 at 03 12 44](https://user-images.githubusercontent.com/18397186/177896748-6abb69f3-9c8b-4bd8-a6a9-7d4d76c23efe.png)

### macOs (M1)
#### Specs
![CleanShot 2022-07-08 at 03 14 21](https://user-images.githubusercontent.com/18397186/177897043-985e01dc-c968-4a5d-9c83-affadca9c613.png)

#### Results for Node v14.20.0
![CleanShot 2022-07-08 at 03 16 40](https://user-images.githubusercontent.com/18397186/177897108-08024949-e93b-4a3d-bf3f-cf8ed94c68c3.png)

#### Results for Node v16.16.0
[...]

PR #932, opened 2022-07-05
feat(tenants): initial tenants module 

TimoGlastra:
Initial version of the tenants module. This is not fully finished yet and ready for prime time, but it contains most of the things needed for multi-tenancy. It doesn't introduce any breaking changes (those were addressed in the PRs this PR depends on).

Dependant on:
- #920
- #921 
- #922 

Until the modularization is fully here the API for registering module is not really nice, but I see that as a separate task that we can address after multi-tenancy has been merged. This is how you use it:

```ts
import { DependencyManager, Agent } from '@aries-framework/core'
import { TenantsApi, TenantsModule } from '@aries-framework/module-tenants'
import { agentDependencies } from '@aries-framework/node'

// create dependency manager and register module
const dependencyManager = new DependencyManager()
dependencyManager.registerModules(TenantsModule)

// create agent, and resolve tenants api
const agent = new Agent({ /* config */, agentDependencies , dependencyManager)
const tenantsApi = agent.dependencyManager.resolve(TenantsApi)

const tenantRecord = await tenantsApi.createTenant({
  config: {
   label: 'Tenant 1',
  },
})

const tenantAgent = await tenantsApi.getTenantAgent({
  tenantId: tenantRecord.id,
})

// Create oob invitation in scope of tenant
const outOfBandRecord = await tenantAgent.oob.createInvitation()
```

I would like to already merge so we can keep PRs reasonable (this is already getting quite large).

Basically the only thing that's left to do is correctly managing the tenant lifecycle and managing tenant sessions (see https://hackmd.io/vGLVlxLvQR6jsEEjzNcL8g?view#Tenant-Lifecycle). Currently, wallets are being opened when needed and will not be closed afterwards. That means over time the number of open wallets keeps increasing. Correctly opening wallets and avoiding race conditions if two sessions would like to open the same wallet is handled (using `async-mutex` library). This does need some improvements in not waiting indefinitely.

Concrete what's still todo:
- Agent lifecycle. Currently you call `getTenantAgent` with no proper way to dispose of the agent (session count won't be reduced)
- Check we're not using context after a process is complete (e.g. by setting an event listener). This should always start a new session instead of using the session from the upper scope

I'm quite happy with how it turned out and how multi-tenancy is an optional addon without having the code in the core package.


packages/module-tenants/src/context/TenantAgentContextProvider.ts
```diff
@@ -0,0 +1,144 @@
...
+    this.rootAgentContext = rootAgentContext
+    this.eventEmitter = eventEmitter
+    this.tenantSessionCoordinator = tenantSessionCoordinator
+    this.logger = logger
+
+    // Start listener for newly created routing keys, so we can register a mapping for each new key for the tenant
+    this.listenForRoutingKeyCreatedEvents()
+  }
+
+  public async getAgentContextForContextCorrelationId(tenantId: string) {
```
karimStekelenburg, line 48: What  is a 'correlation id'? Looks like this just fetches the context for a specific tenant id. `tenantId` is more clear imho.

packages/module-tenants/src/repository/TenantRecord.ts
```diff
@@ -0,0 +1,37 @@
+import type { TenantConfig } from '../models/TenantConfig'
```
karimStekelenburg, line 1: Maybe it's useful to add some kind of `usedAt` or `latestUsage` kind of property to indicate when the tenant was used for the last time. This might come in handy later on when doing a clean-up of old/unused tenants.

karimStekelenburg (approved):
Nice work! Looks good to me. I left two comments, but those are minor things.

Additionally, maybe it's nice to add methods for creating and getting multiple tenants. E.g. `createTenants([{...}, {...}, {...}, {...}])`. Maybe that is too difficult regarding sessions management tho. Idk, just a thought.

packages/module-tenants/src/context/TenantAgentContextProvider.ts
TimoGlastra: Yes this seems to be causing some confusion. See my reply to berend: https://github.com/hyperledger/aries-framework-javascript/pull/921#issuecomment-1177253640. The tldr is that multi-tenancy is an optional addon so I'd like to keep the word tenant/multi-tenancy out of core

karimStekelenburg:
> > Additionally, maybe it's nice to add methods for creating and getting multiple tenants. E.g. `createTenants([{...}, {...}, {...}, {...}])`. Maybe that is too difficult regarding sessions management tho. Idk, just a thought.
> 
> We could. Do you think it will be an often used use case to create a lot of tenants at the same time? I'd like to keep the API minimal, so if you have a good use case for it, I'd be happy to add it.

So I can't think of a good use case for creating multiple tenants atm, but I can imagine wanting to execute some kind of action in batch for multiple tenants at the same time (e.g. creating a connection invite). In this case a `getTenants` method could be useful.

packages/module-tenants/src/context/TenantAgentContextProvider.ts
karimStekelenburg: Ok, clear. Not sure I like the term `correlationId`, but I also can't think of anything better atm...

packages/module-tenants/src/repository/TenantRecord.ts
karimStekelenburg: Hmm ok, I get what you're saying. Maybe postpone this a bit indeed. It's useful but not a necessity either.

packages/module-tenants/src/context/TenantAgentContextProvider.ts
TimoGlastra: Yes feel free to make suggestions on the name. I'm also not a fan of it, but it was the best I can come up with. It's different than a contextId (as multiple contexts could have the same tenant id).

TimoGlastra:
> So I can't think of a good use case for creating multiple tenants atm, but I can imagine wanting to execute some kind of action in batch for multiple tenants at the same time (e.g. creating a connection invite). In this case a `getTenants` method could be useful.

Ok, I've made some changes to these method in a next PR. Let me look at it in that branch and see If I can come up with nice API

TimoGlastra:
Merging as only failing test is the one from #923

PR #935, opened 2022-07-05
fix(vc): change pubKey input from Buffer to Uint8Array

karimStekelenburg:
This PR addresses an issue that occurred when running the `deriveProof` test case in the `W3cCredentialService` tests. Strangely enough this issue only occurred when running in a NodeJS environment (which uses node-bbs-signatures) instead of the WASM fallback.

Huge thanks to @blu3beri for helping debug this issue. This one wasn't easy to catch.


TimoGlastra:
Hmm after looking at it again, this looks like a different issue than the flaky tests as it's only failing on BBS tests. Maybe a linux + node 18 issue?

karimStekelenburg:
Ok, I've added the environment variable in 49c35d4 which should hopefully make sure the tests succeed for all Node versions (waiting on CI to finish atm, wish me luck 😄). One thing I'd like to double check: in order to not have to overhaul the entire workflow structure, I've added the environment variable top level. This means that the tests for Node 14 and 16 (which did succeed using `node-bbs-signatures`) will also use the WASM build.

Please let me know if there are any objections regarding this.

@TimoGlastra, @blu3beri, would love to hear your thoughts on this.

TimoGlastra:
Documenting would be good. In addition I think there's two approaches here (maybe more):
- set the environment variable by default before importing the bbs library
- logging a warning if the node version is used and we're in node 17/18. The hard part is, how do we know which version is used? We could check if the node bbs signatures install succeeded (by trying to import it) and then logging the warning
- logging a warning if the env variable is not set and we're in node 17/18. This will push the user to always set the env variable explicitly

TimoGlastra:
Could you maybe open an issue in the node bbs signatures repo so the maintainers are aware of the issue?

karimStekelenburg:
Ahh great! Now we broke Node 14 again! 💯

TimoGlastra:
> Ahh great! Now we broke Node 14 again! 💯

Lol. Maybe we should take the first route and only set it for node 17 & 18?

karimStekelenburg:
> > Ahh great! Now we broke Node 14 again! 💯
> 
> Lol. Maybe we should take the first route and only set it for node 17 & 18?

Yeah... I tried running the tests on my machine (M1) locally with 14 and it succeeds. I will try to run it with Node 14 on an Ubuntu system now, just to get a better understanding where the problem lies, but I agree, that solution is probably better.

However, should we really set this environment variable just before importing the package? Isn't it cleaner to do this more 'top level' e.g. in the `Agent` class?

blu3beri:
I will try to debug the issue tomorrow with the node library, however a temporary fix would be best for now. 

WASM should not be that much slower so it's for sure not the end of the world.

karimStekelenburg:
> > > > Ahh great! Now we broke Node 14 again! 💯
> > > 
> > > 
> > > Lol. Maybe we should take the first route and only set it for node 17 & 18?
> > 
> > 
> > Yeah... I tried running the tests on my machine (M1) locally with 14 and it succeeds. I will try to run it with Node 14 on an Ubuntu system now, just to get a better understanding where the problem lies, but I agree, that solution is probably better.
> > However, should we really set this environment variable just before importing the package? Isn't it cleaner to do this more 'top level', e.g. in the `Agent` class?
> 
> If we do it in the agent class the bbs library is probably already imported in which case setting the env variable doesn't matter anymore. So we need to make sure it happens before the import of bbs library

Even if we do this at the very top of the `Agent` class (before importing any of the containers)? I'm OK with anything tho, but this seems like a more evident place. I just want to avoid forgetting this in the future because it's being done in a deeply nested file.

Anyhow, will do it right before the import for now. We can change it in a follow up PR :)

TimoGlastra:
Maybe the index.ts file is a better suited place? That's the entry point of the framework. I'm almost certain the agent class won't work

karimStekelenburg:
Ok, I ran some local tests. I ran it on Ubuntu hoping to reproduce the error, and on macOs (M1) for future reference. On both platforms all tests succeeded. This is going to be fun!

### Ubuntu
#### Specs
CleanShot 2022-07-08 at 03 06 55

#### Results for Node v14.19.3
CleanShot 2022-07-08 at 02 52 42

#### Results for Node v16.16.0
CleanShot 2022-07-08 at 02 55 12

#### Results for Node v17.9.1
CleanShot 2022-07-08 at 02 57 11

#### Results for Node v18.5.0
CleanShot 2022-07-08 at 03 12 44

### macOs (M1)
#### Specs
CleanShot 2022-07-08 at 03 14 21

#### Results for Node v14.20.0
CleanShot 2022-07-08 at 03 16 40

#### Results for Node v16.16.0
CleanShot 2022-07-08 at 03 18 27

#### Results for Node v17.9.1
CleanShot 2022-07-08 at 03 20 11

#### Results for Node v18.5.0
CleanShot 2022-07-08 at 03 21 54

TimoGlastra:
> Ok, I ran some local tests. I ran it on Ubuntu hoping to reproduce the error, and on macOs (M1) for future reference. On both platforms all tests succeeded. This is going to be fun!
> 
> ### Ubuntu
> #### Specs
> CleanShot 2022-07-08 at 03 06 55
> 
> #### Results for Node v14.19.3
> CleanShot 2022-07-08 at 02 52 42
> 
> #### Results for Node v16.16.0
> CleanShot 2022-07-08 at 02 55 12
> 
> #### Results for Node v17.9.1
> CleanShot 2022-07-08 at 02 57 11
> 
> #### Results for Node v18.5.0
> CleanShot 2022-07-08 at 03 12 44
> 
> ### macOs (M1)
> #### Specs
> CleanShot 2022-07-08 at 03 14 21
> 
> #### Results for Node v14.20.0
> CleanShot 2022-07-08 at 03 16 40
> 
> #### Results for Node v16.16.0
> CleanShot 2022-07-08 at 03 18 27
> 
> #### Results for Node v17.9.1
> CleanShot 2022-07-08 at 03 20 11
> 
> #### Results for Node v18.5.0
> CleanShot 2022-07-08 at 03 21 54

But isn't that expected with WASM? We're having issues with Node.JS. Or do I mistunderstand the issue?

PR #943, opened 2022-07-11
feat(credentials): credential module config

TimoGlastra:
Adds a CredentialModuleConfig to showcase how module configuration would work. For now the module is still internal and takes config from the agent config, but it shows how custom module configuration works. This is one step closer to configurable modules.

Basically how you'd use it (if we have full modularization):

```ts
const agent = new Agent({
  modules: {
    credentials: new CredentialsModule({
        autoAcceptCredentials: AutoAcceptCredential.always
    })
  }
})
```

The credentials api can then depend on it like this:

```ts
CredentialsApi {
  public constructor(
    credentialsModuleConfig: CredentialsModuleConfig
  ) {
   const autoAcceptCredentials = this.credentialsModuleConfig.autoAcceptCredentials
  }
}
```

@karimStekelenburg @blu3beri @genaris @2byrds @JamesKEbert Thoughts? If we can agree on the approach for configuration, I can update it for all modules. We can still keep it in init config (so there won't be much breaking changes for now). But from then on it will be trivial to start extracting modules out of core


karimStekelenburg:
I'm a bit confused on what the intent is of this addition. Is my understanding correct that we're currently in a sort of 'hybrid state' where we can register modules both the old and the new way?

Regardless, a showcase/reference implementation is useful indeed. Probably needless to say, but I would prefer to refactor the configuration for all modules before releasing `0.3.0`.

TimoGlastra:
> Is my understanding correct that we're currently in a sort of 'hybrid state' where we can register modules both the old and the new way?

Well this PR is just to demonstrate how it would look and get some feedback (hence draft PR). If we think this is a good approach I'll update this PR to refactor all modules to the new format with config. However I won't immediately deprecate the old way, I'd like to do that in a separate PR

packages/core/src/agent/BaseAgent.ts
```diff
@@ -82,7 +82,7 @@ export abstract class BaseAgent {
 
     // We set the modules in the constructor because that allows to set them as read-only
     this.connections = this.dependencyManager.resolve(ConnectionsModule)
-    this.credentials = this.dependencyManager.resolve(CredentialsModule) as CredentialsModule
+    this.credentials = this.dependencyManager.resolve(CredentialsApi) as CredentialsApi
```
blu3beri, line 85: Any idea why this cast is required here?

packages/core/src/modules/credentials/CredentialsApi.ts
```diff
@@ -0,0 +1,626 @@
...
+  findProposalMessage(credentialExchangeId: string): Promise<FindProposalMessageReturn<CSs>>
+  findOfferMessage(credentialExchangeId: string): Promise<FindOfferMessageReturn<CSs>>
+  findRequestMessage(credentialExchangeId: string): Promise<FindRequestMessageReturn<CSs>>
+  findCredentialMessage(credentialExchangeId: string): Promise<FindCredentialMessageReturn<CSs>>
+}
+
+@injectable()
+export class CredentialsApi<
+  CFs extends CredentialFormat[] = [IndyCredentialFormat],
+  CSs extends CredentialService<CFs>[] = [V1CredentialService, V2CredentialService<CFs>]
```
blu3beri, line 90: Will these kinds of generics only be used with the `credentialsApi`? I am worried about an api class with many generics and it becomes very difficult to comprehend what is going on.

packages/core/src/modules/credentials/CredentialsModule.ts
```diff
@@ -1,638 +1,29 @@
...
   }
 
   /**
    * Registers the dependencies of the credentials module on the dependency manager.
    */
-  public static register(dependencyManager: DependencyManager) {
+  public register(dependencyManager: DependencyManager) {
     // Api
-    dependencyManager.registerContextScoped(CredentialsModule)
+    dependencyManager.registerContextScoped(CredentialsApi)
```
blu3beri, line 26: I am still not the biggest fan of these register function names. I think that `registerContextScoped` is a bit ambiguous and does not tell me when to use this method in the context of AFJ.

aliassing it to something like: `registerPublicApi` or whatever fits better might be worth it to make writing external plugins a lot easier.

packages/core/src/modules/credentials/CredentialsModuleConfig.ts
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * CredentialsModuleConfigOptions defines the interface for the options of the CredentialsModuleConfig class.
+ * This can contain optional parameters that have default values in the config class itself.
+ */
+export interface CredentialsModuleConfigOptions {
+  autoAcceptCredentials: AutoAcceptCredential
+}
+
+export class CredentialsModuleConfig {
```
blu3beri, line 11: This is awesome! Way more clear which properties from the agent config are with that module. 

Just more curious, how do we deal with configuration shared between modules, if it ever occurs?

blu3beri (commented):
Really nice! Just let some comments, mainly on usability.

packages/core/src/modules/credentials/CredentialsApi.ts
TimoGlastra: It will be used for protocols where this is needed. So I suspect it will only be credentials and proofs. Understand the note on generics being tricky. I'm not sure however how to make it both generic and simple

packages/core/src/modules/credentials/CredentialsModule.ts
TimoGlastra: I see your point. What I'm not sure about is that we're not sure yet what types of components people want to register. Maybe we can come up with a set of components, but it would need to cover all cases. 

- `registerApi` for the api instead of `registerContextScoped`
- `registerStatelessService` instead of `registerSingleton` 
- `registerInstance` stays as is

We would still need another name for the `registerContextScoped` if you have something that is not an API but needs to be unique per context. Do you like this direction better?

packages/core/src/modules/credentials/CredentialsModuleConfig.ts
TimoGlastra: Good question. I think either declare it in both modules and provide it twice, or one module can depend on the configuration of the other module (if they are dependant on each other).

E.g. the outOfBandModule can depend on the `autoAcceptConnections` from the `ConnectionsModule` (as you can't use the oob module without the connections module). Does that make sense?

packages/core/src/modules/credentials/CredentialsModule.ts
blu3beri: Yeah this makes it a lot easier to understand. Maybe for the `registerContextScoped` without an api and needs to be unique, we can use `registerStatefulService`? I believe that that is what you want to register, right?

packages/core/src/modules/credentials/CredentialsModuleConfig.ts
blu3beri: Yeah that makes sense. Is it obvious to the user that the `autoAcceptConnections` from the `ConnectionsModule` will be used in the `outOfBandModule`?

packages/core/src/modules/credentials/CredentialsModule.ts
TimoGlastra: I'll raise it during the WG call and gather some more input on this. If we agree we want to change something, I'll raise another PR

packages/core/src/modules/credentials/CredentialsModuleConfig.ts
TimoGlastra: Not sure. We could add it to the documentation of the out of band module. So in the receiveInvitation method documentation we can mention it will use the `autoAcceptConnections` property from the connections module config to determine whether to auto accept the config. 

I've added the config of each module as a public readonly property to the api class, which means it's public api and makes sense it can be used by other modules.

So `agent.connections.config` is a public property that exposes the config of the connections module

blu3beri (approved):
Nice stuff in here! Also a good idea to mark stuff with `@deprecated`. However, do we have any backwards compatibility with the agent config from before?

TimoGlastra:
> Nice stuff in here! Also a good idea to mark stuff with `@deprecated`. However, do we have any backwards compatibility with the agent config from before?

What do you mean exactly with `from before`?

blu3beri:
> What do you mean exactly with `from before`?

Well, you added a deprecation notice like so:

```ts
  /**
    * @deprecated use indyLedgers from the `LedgerModuleConfig` class
    */
   public get indyLedgers() {
     return this.initConfig.indyLedgers ?? []
   }
```

This would mean that this is still works and will get the value from the initConfig. Is that correct?

TimoGlastra:
Yes, currently all values still come from the init config. This is mainly because there's no way to configure the internal modules yet. So the main init config is the easiest to make sure it keeps working, but already allowing custom config for custom modules.

I'm still not sure what the best approach is for internal module configuration. My thoughts were to maybe allow to define the module yourself and we'll only register it internally with default config if you don't provide the module.

Without custom config the connections module is registerd internally and auto accept connections will have default value false.

```ts
const agent = new Agent({
  modules: {

  }
})
```


If i register it, it won't be registered internally and we'll use the config as passed by the user.

```ts
const agent = new Agent({
  connections: new ConnectionsModule({
     autoAcceptConnections: true
  })
})
```

But maybe there's other / better appraoches?

blu3beri:
> But maybe there's other / better appraoches?

I quite like this approach and I think that we should do it like that.

One thing that might become slightly annoying here is that shared configuration, I don't think it will happen a lot so it might just be ignored for now.

One quick fix that I have for that is to do it like so:

```ts
const agent = new Agent({
  config: {
     autoAcceptConnections: true
  },
  connections: new ConnectionsModule()
})
```

We can create some cool type that infers the `config` types from the modules you have added, but this will leave us again with a large object where all the configuration is mashed together... 

TBH I prefer your method, but maybe this can give some inspiration for another solution that fixes every problem :)

TimoGlastra:
Merging as only failing test is BBS+ test

PR #945, opened 2022-07-12
Feat/indy namespace

### LINKED ISSUE #944: add `indyNamespace` to the indy ledgers configuration (need to document this: https://hyperledger.github.io/indy-did-method/#indy-did-method-identifiers)
- add `indyNamespace` to the indy ledgers configuration (need to document this: https://hyperledger.github.io/indy-did-method/#indy-did-method-identifiers)
- add fully qualified identifier to the anon creds (schema|credential definition) records (replace the non qualified identifier before storing and after retrieving it). For indy we use the did:indy spec identifiers to store.
  - Note: Records stores fully qualified identifier as identifier.
  - Need to transform unqualified to qualified before storing or querying
  - Need to transform qualified to unqualified after retrieving
- Need to return the fully qualified id somehow from the service (either by adding qualifiedId return value, or overriding the id before returning and the setting it as the unqualified again after storing it). 
- Ledger module construct fully qualified string/name. Check with that against record store

_Originally posted by @TimoGlastra in https://github.com/hyperledger/aries-framework-javascript/issues/900#issuecomment-1181718542_

morrieinmaas:
closes #944


TimoGlastra:
I think this can be closed in favour of #965 @morrieinmaas?

PR #946, opened 2022-07-13
fix: no return routing and wait for ping

TimoGlastra:
This PR adds some improvements to return routing. Mainly it adds two things:
1. Messages returned as a http response are now emitted using an event, meaning the sendMessage await won't wait for the returned message to be processed. This could lead to a chain of send / receive messages that would take a long time to resolve. Especially if the socket was kept open because no response was returned this could lead to unwanted behaviour. (see https://github.com/hyperledger/aries-cloudagent-python/pull/1853 for more context on socket staying open)
2. Do not include the return route decorator if not response is expected. The return route decorator is automatically added if we don't have an endpoint. This only happens with the mediator in most cases. What would happen is that the trust ping was sent to complete the connection, and we didn't receive a response. This would lead to the socket staying open for 15 seconds, before the mediation request message was sent. Now if the return route is already set (in this case to none) we won't add it in the message sender.. 

As mentioned in https://github.com/hyperledger/aries-cloudagent-python/pull/1853, this reduces the connection + request mediation flow time from ~17 seconds to around 1.5 seconds :)

### DIFF
--- a/packages/core/src/agent/MessageSender.ts
+++ b/packages/core/src/agent/MessageSender.ts
@@ -227,7 +227,9 @@ export class MessageSender {
     // as the `from` field in a received message will identity the did used so we don't have to store all keys in tags to be able to find the connections associated with
     // an incoming message.
     const [firstOurAuthenticationKey] = ourAuthenticationKeys
-    const shouldUseReturnRoute = !this.transportService.hasInboundEndpoint(ourDidDocument)
+    // If the returnRoute is already set we won't override it. This allows to set the returnRoute manually if this is desired.
+    const shouldAddReturnRoute =
+      payload.transport?.returnRoute === undefined && !this.transportService.hasInboundEndpoint(ourDidDocument)
 
     // Loop trough all available services and try to send the message
     for await (const service of services) {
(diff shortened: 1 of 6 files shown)


JamesKEbert (approved):
This is great Timo!
You mentioned during the WG call this morning that it does seem a little hacky. Would it be appropriate to add a new decorator or add to the return route decorator to address this specific use case, as that way it might not be as hacky? Thoughts?

TimoGlastra:
> Would it be appropriate to add a new decorator or add to the return route decorator to address this specific use case, as that way it might not be as hacky?

Could you elaborate a bit on what you mean?

TimoGlastra:
Interesting. Well the question arises for me, what does keep-alive mean then. This implies a message is always instantly processed and a message may be returned directly, or the socket will be closed. Maybe we should approach it more as the keep alive header in HTTP: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Keep-Alive?

I think ACA-Py's implementation shouldn't hang on to the http connection after a message has been processed, which means this wouldn't be an issue anymore.

Issue #948, opened 2022-07-14
Extract bbs dependencies into @aries-framework/bbs-signatures packages


TimoGlastra:
See https://hackmd.io/6eXZcMhdQR-QCNPMFYTvuA?view#W3C-Credentials

All code that uses the mattr bbs libraries should be extracted to the `@aries-framework/bbs-signatures` which can then be dynamically added if bbs support is desired. This way you're not forced to add yet another native binary to your mobile app (this is where it counts the most).

PR #949, opened 2022-07-14
feat: add dynamic suite and signing provider 

TimoGlastra:
Extends the `SignatureSuiteRegistry` and adds a `SigningProviderRegistry` that allows to dynamically register signers for key types and signature suites. This is needed in preparation of #948, so we can extract the bbs signing provider and signature suites to it's own package.

Extracting into a separate package will be done in a separate PR.

BbsService has been moved into the generic `SigningProvider` interface. I've removed support for Bls12381g1 keys for now as they're not really used for anything. It will be easy to add back in.


packages/core/src/crypto/signing-provider/SigningProvider.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export interface CreateKeyOptions {
+  seed?: string
+}
+
+export interface SigningProvider {
+  readonly keyType: KeyType
+
+  createKey(options: CreateKeyOptions): Promise<KeyPair>
```
blu3beri, line 29:
```suggestion
  createKeyPair(options: CreateKeyPairOptions): Promise<KeyPair>
```

For `createKey` I would expect a symmetric key.

packages/core/src/modules/vc/SignatureSuiteRegistry.ts
```diff
@@ -1,40 +1,27 @@
...
 
 import { suites } from './libraries/jsonld-signatures'
-import { Ed25519Signature2018 } from './signature-suites'
-import { BbsBlsSignature2020, BbsBlsSignatureProof2020 } from './signature-suites/bbs'
 
 const LinkedDataSignature = suites.LinkedDataSignature
 
+export const SignatureSuiteToken = Symbol('SignatureSuiteToken')
 export interface SuiteInfo {
   suiteClass: typeof LinkedDataSignature
```
blu3beri, line 12: I would assume this is correct, but if is supposed to be generic between, for now, ed25519 and BBS+ then the `LinkedDataSignature` is not really a good name I think.

packages/core/src/wallet/IndyWallet.ts
```diff
@@ -467,17 +470,21 @@ export class IndyWallet implements Wallet {
...
         const verkey = await this.indy.createKey(this.handle, { seed, crypto_type: 'ed25519' })
         return Key.fromPublicKeyBase58(verkey, keyType)
       }
 
-      if (keyType === KeyType.Bls12381g1 || keyType === KeyType.Bls12381g2) {
-        const blsKeyPair = await BbsService.createKey({ keyType, seed })
-        await this.storeKeyPair(blsKeyPair)
-        return Key.fromPublicKeyBase58(blsKeyPair.publicKeyBase58, keyType)
+      // Check if there is a signing key provider for the specified key type.
+      if (this.signingKeyProviderRegistry.hasProviderForKeyType(keyType)) {
```
blu3beri, line 482: Isn't this check redundant as we do this in a try catch and use the `this.signingKeyProviderRegistry.getProviderForKeyType` method, which can error and checks whether it exists or not.

blu3beri (approved):
Nice abstraction. Just left small comments.

packages/core/src/wallet/IndyWallet.ts
TimoGlastra: Yeah sort of. I thought this way we can keep the custom error message that this key type isn't supported in the indy wallet (As a wallet can support key types itself, as well as using the singing key provider). Do you think it's better to remove it?

packages/core/src/modules/vc/SignatureSuiteRegistry.ts
TimoGlastra: I think the bbs suites also extend linked data signature.

packages/core/src/crypto/signing-provider/SigningProvider.ts
TimoGlastra: Makes sense 👍

packages/core/src/wallet/IndyWallet.ts
blu3beri: No, I am okay with this. Just more curious behind the reasoning.

packages/core/src/modules/vc/SignatureSuiteRegistry.ts
TimoGlastra: Nope it extends `LinkedDataProof`, and `LinkedDataSignature` also extends `LinkedDataProof`. 

@karimStekelenburg any reason you used `LinkedDataSignature` instead of `LinkedDataProof`?

karimStekelenburg: Hmm not sure tbh. Probably by accident

PR #955, opened 2022-07-20
feat!: agent module registration api

TimoGlastra:

Add agent module registration api. With this you can now register custom modules on the agent like this:

```ts
const agent = new Agent({
  config: { /* agent config */ },
  modules: {
    tenants: new TenantsModule(),
  },
})
```

This will register the tenants module and make the tenants api available. There's some nice typescript magic going on here that automatically infers the api type from the tenants module, allowing the agent to be fully typed with custom modules:

```ts
// agent.modules.tenants is fully typed
const tenantAgent = await agent.modules.tenants.getTenantAgent({ tenantId: 'the-tenant-id' })

// tenantAgent is also fully typed, but doesn't have the `agent.api.tenants` available
await tenantAgent.oob.createInvitation()
```

As you may have already noticed from the example above, all the custom module api's are now nested under the `agent.modules` property. I've taken this approach for two reasons:

You can define the api for a module using the api key on a module:

```ts
class MyModule {
  public api = MyApi

    dependencyManager.registerContextScoped(MyApi)
  }
}
```

All modules that were registered by default on the agent are still registered by default. So `agent.oob` is present, as well as `agent.connections`, etc... For now the default modules are still initialized based on the configuration in the agent config, but in a next PR I'll remove that logic and you will only be able to configure modules using the module config itself. If you want to provide custom configuration for a default module you can do so by registering the module. This will just register the module with you custom config, instead of using the default config:

```ts
const agent = new Agent({
  config: { /* agent config */ },
  modules: {
    // provide custom config for the default registered connections module.
    connections: new ConnectionsModule({
        autoAcceptConnections: true
    })
  },
})
```

---

BREAKING CHANGE: custom modules have been moved to the .modules namespace. In addition the agent constructor has been updated to a single options object that contains the `config` and `dependencies` properties. Instead of constructing the agent like this:
```ts
const agent = new Agent({ /* config */ }, agentDependencies)
```

You should now construct it like this:
```ts
const agent = new Agent({
  config: { /* config */ },
  dependencies: agentDependencies
})
```

This allows for the new custom modules to be defined in the agent constructor.


karimStekelenburg:
Nice job @TimoGlastra! My only remark here is that the `agent.api` naming doesn't make total sense to me (maybe I'm seeing it wrong though). Wouldn't `agent.modules` make more sense?

TimoGlastra:
> My only remark here is that the `agent.api` naming doesn't make total sense to me (maybe I'm seeing it wrong though). Wouldn't `agent.modules` make more sense?

Yeah have been thinking about using module, but that also didn't fully make sense to me. It only exposes the api of the module, not the whole module object (so `ConnectionsApi` class instance, not the `ConnectionsModule` instance)

karimStekelenburg:
Haha fair, this is a very semantic thingy.

My thought here is this: as a user I will pass modules to the agent constructor. Therefore, I already know that a module adds certain functionality. However, I might not be familiar with the internal structure of modules, so I might not be aware of the concept of `FooApi` classes.

Since I'm passing modules, I wouldn't be surprised to have to access their APIs through `agent.modules`.

Anyway, just my thought! This maybe is a good candidate for a WG discussion ;)

TimoGlastra:
Input from WG call:

- add `custom` property that only contains the custom modules. Default modules will still be top level.
- use `modules` instead of `api`.

I think I agree with the `agent.modules` part over the `agent.api`. Not sure I'm a fan of the `custom` namespace. More feedback is welcome!

blu3beri:
> Input from WG call:
> 
> 
> 
> - add `custom` property that only contains the custom modules. Default modules will still be top level.
> 
> - use `modules` instead of `api`.
> 
> 
> 
> I think I agree with the `agent.modules` part over the `agent.api`. Not sure I'm a fan of the `custom` namespace. More feedback is welcome!

I am slightly in favour of the top-level modules and custom modules defined in `agent.modules`. I think doing something like `agent.modules.credentials.offerCredential` becomes so long that it will be annoying to use. Custom modules are IMHO second-class and can be defined under a specific namespace. 

Keeping the core-defined modules top level, which in the end might not be a lot, makes a cleaner API. 

Having said that, doing `agent.oob.XXX` For an invitation and then `agent.modules.credential.XXX` For issuance is very weird. 

I am fine with either for different reasons. Maybe we can look at different frameworks and how they solved this issue?

blu3beri:
I thought about this issue a bit more, and it might not be such a big issue to have breaking changes when we add new modules into core. Adding a module into core, after modularisation, is very unlikely to happen when we have the did core modules in there. Unless they expand their spec or we want add seperation, we would not have the add any modules. Of course, it might happen and bundling it with some other breaking changes would be fine IMHO.

Also, as we mentioned in the last WG call, the `agent.module.XXX` does not really solve the key-collision issue vs the `agent.XXX`. With this my preference is for `agent.XXX` unless I am missing something that was another purpose of `agent.module.XXX`.

TimoGlastra:
Well we need to change the call to Agent.create (or something like that such as a separate method `createAgent`) instead of calling `new Agent()`, but otherwise there's no difference

genaris:
I don't see any problem on using an agent creator method, especially considering that we currently need to call `initialize` method after the constructor to make it usable. Could be done anything there to simplify this initialization in cases where applicable?

A question that arises from this approach is if there would be a chance of overriding a core module. We currently do so with connections module to add some project-specific capabilities (we sub-class `Agent` and define a `connections` property assigned to a sub-class of `ConnectionModule`), and even if it feels a bit hacky, it works quite smoothly.

TimoGlastra:
Yes as berend mentioned it would be possible to override the core module by providing it in the agent config, however the typing is currently set to only allow the same module. You would have to write your `CustomModule` and register it under a different key. We could loosen the type a bit to allow different modules, but I think it makes sense to use a different key if you're not using the connections module. So `agent.myConnections` or something that contains your business name etc.. would be a bit more explicit maybe? Otherwise we can look at making it accept the same interface (as it extends the connections module)

TimoGlastra:
> it's not a big deal if this idea gets too complicated. Just a 'nice to have' from my point of view.

I think it's doable, but maybe for a future PR. This can be added as a non-breaking change, so we could maybe push it to 0.3.1

TimoGlastra:
This PR is now updated and should be ready to merge. I've updated the description of the PR but it boils down to:
- default modules are still nested under agent.xxx
- custom modules are nested under agent.modules.xxx.

I think this is one of the blockers for the 0.3.0 release, so let's get this merged :)

PR #959, opened 2022-07-21
feat(routing): support promise in message repo

TimoGlastra:


packages/core/src/storage/MessageRepository.ts
```diff
@@ -1,6 +1,6 @@
 import type { EncryptedMessage } from '../types'
 
 export interface MessageRepository {
-  takeFromQueue(connectionId: string, limit?: number): EncryptedMessage[]
-  add(connectionId: string, payload: EncryptedMessage): void
+  takeFromQueue(connectionId: string, limit?: number): EncryptedMessage[] | Promise<EncryptedMessage[]>
+  add(connectionId: string, payload: EncryptedMessage): void | Promise<void>
```
blu3beri, line 5: Can this still be sync? I thought all repository operations are async.

blu3beri (approved):

packages/core/src/storage/MessageRepository.ts
TimoGlastra: The message repository is a special type of repository. We should probably rework this in the near future if we want to start using AFJ as a mediator. For now this makes it possible to make a custom repo implementation based on the indy wallet storage (which is indeed async).

TimoGlastra: It should probably be called something like queue (as that's what is is based on the methods e.g. takeFromQueue)

PR #960, opened 2022-07-22
refactor(credential): update repository after metadata update

blu3beri:
- Updates the credentialRecord in the repository after a metadata change. (This probably did not cause any issues before as we modify the object by reference).
- Processing a credential offer now also adds the schema + cred def in the metadata to show the schema as the name in a wallet before accepting the credential.

### DIFF
--- a/packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts
+++ b/packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts
@@ -129,7 +129,7 @@ export class IndyCredentialFormatService extends CredentialFormatService<IndyCre
       schemaId: proposal.schemaId,
       credentialDefinitionId: proposal.credentialDefinitionId,
     })
-    this.credentialRepository.update(credentialRecord)
+    await this.credentialRepository.update(credentialRecord)
 
     return { format, attachment, previewAttributes }
   }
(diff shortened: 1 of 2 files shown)


packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts
```diff
@@ -211,13 +212,19 @@ export class IndyCredentialFormatService extends CredentialFormatService<IndyCre
...
+    if (!credentialOffer.schema_id || !credentialOffer.cred_def_id) {
       throw new CredentialProblemReportError('Invalid credential offer', {
         problemCode: CredentialProblemReportReason.IssuanceAbandoned,
       })
     }
+
+    credentialRecord.metadata.set(CredentialMetadataKeys.IndyCredential, {
+      credentialDefinitionId: credentialOffer.cred_def_id,
+      schemaId: credentialOffer.schema_id,
+    })
```
TimoGlastra, line 226: This was removed recently as we discussed that the record will only contain the values we have accept. When we process an offer we haven't accepted the offer yet.

This makes it so that I can always check for the values I last accepted in the metadata, as otherwise it can be either the data I received, or the data I sent.

If we want to revert this (which I'd like to have a more in depth discussion about) it'll be a breaking change.

packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts
```diff
@@ -211,13 +212,19 @@ export class IndyCredentialFormatService extends CredentialFormatService<IndyCre
...
+    await this.credentialRepository.update(credentialRecord)
```
TimoGlastra, line 227: Updating is done in the credential service after tje credential format has been called. The formats services just modify it. Were you running into issues with things not being saved?

TimoGlastra (changes requested):

PR #963, opened 2022-07-25
build: remove did peer git cache file

TimoGlastra:

This was added in #717, but was probably a mistake and is causing issues


blu3beri (approved):
Ignore my comment :) I did not read it correctly.

PR #965, opened 2022-07-26
feat: add indynamespace for ledger id for anoncreds

morrieinmaas:


morrieinmaas:
closes #944 


Note to self: update [docs](https://hyperledger.github.io/indy-did-method/#indy-did-method-identifiers) when all review feedback is addressed.

packages/core/src/utils/indyIdentifiers.ts
```diff
@@ -0,0 +1,24 @@
...
+}
+
+export function createQualifiedIdentifier(didIndyNamespace: string, namespaceIdentifier: string): IndyNamespace {
+  return `did:indy:${didIndyNamespace}:${namespaceIdentifier}`
+}
+
+export function getDidFromSchemaOrCredentialDefinitionId(schemaId: string): string {
+  const subStrings = schemaId.split(':')
+  return subStrings[0]
+}
```
TimoGlastra, line 24: We already have `didFromSchemaId` and `didFromCredentialDefinitionId`. Can we reuse those?

packages/core/src/utils/indyIdentifiers.ts
```diff
@@ -0,0 +1,24 @@
...
+
+// TODO: This template literal type can possibly be improved. This version leaves the substrings as potentially undefined
+export type IndyNamespace = `${Did}:${DidIndyMethod}:${DidIndyNamespace}:${NamespaceIdentifier}`
+
+export function createUnqualifiedIdentifier(qualifiedIdentifier: string): string {
+  const subStrings = qualifiedIdentifier.split(':')
+  return subStrings[subStrings.length - 1]
+}
+
+export function createQualifiedIdentifier(didIndyNamespace: string, namespaceIdentifier: string): IndyNamespace {
```
TimoGlastra, line 17:
```suggestion
export function getQualifiedIndyDid(didIndyNamespace: string, namespaceIdentifier: string): IndyNamespace {
```

packages/core/src/utils/indyIdentifiers.ts
```diff
@@ -0,0 +1,24 @@
...
+export type DidIndyMethod = 'indy'
+// Maybe this can be typed more strictly than string. Choosing string for now as this can be eg just `sovrin` or eg `sovrin:staging`
+export type DidIndyNamespace = string
+//  NOTE: because of the ambiguous nature - whether there is a colon or not within DidIndyNamespace this is the substring after the ***last*** colon
+export type NamespaceIdentifier = string
+
+// TODO: This template literal type can possibly be improved. This version leaves the substrings as potentially undefined
+export type IndyNamespace = `${Did}:${DidIndyMethod}:${DidIndyNamespace}:${NamespaceIdentifier}`
+
+export function createUnqualifiedIdentifier(qualifiedIdentifier: string): string {
```
TimoGlastra, line 12: or maybe `unqualifyIndyDid` if it is getting to long. but identifier is maybe a bit misleading, we're working with dids here

```suggestion
export function getUnqualifiedDidFromQualifiedIndyDid(qualifiedIndyDid: string): string {
```

packages/core/tests/ledger.test.ts
```diff
@@ -145,6 +145,7 @@ describe('ledger', () => {
       indyLedgers: [
         {
           id: 'pool-Faber Ledger Genesis Transactions',
+          didIndyNamespace: 'pool-Faber Ledger Genesis Transactions',
```
TimoGlastra, line 148: I'm not sure namespaces can contain spaces?

packages/core/src/modules/ledger/services/IndyLedgerService.ts
```diff
@@ -109,6 +109,8 @@ export class IndyLedgerService {
     try {
       this.logger.debug(`Set endpoints for did '${did}' on ledger '${pool.id}'`, endpoints)
 
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
```
TimoGlastra, line 113: This should work if you have the latest @types/indy-sdk version

```suggestion
```

packages/core/src/modules/ledger/__tests__/IndyPoolService.test.ts
```diff
@@ -24,30 +24,35 @@ const CacheRepositoryMock = CacheRepository as jest.Mock<CacheRepository>
...
-    id: 'sovrinStaging',
+    id: 'sovringStaging',
+    didIndyNamespace: 'sovringStaging',
     isProduction: false,
     genesisTransactions: 'xxx',
     transactionAuthorAgreement: { version: '1', acceptanceMechanism: 'accept' },
   },
   {
     id: 'indicioMain',
+    didIndyNamespace: 'indicioMain',
```
TimoGlastra, line 48:
```suggestion
    didIndyNamespace: 'indicio',
```

packages/core/src/modules/ledger/__tests__/IndyPoolService.test.ts
```diff
@@ -24,30 +24,35 @@ const CacheRepositoryMock = CacheRepository as jest.Mock<CacheRepository>
...
     id: 'sovrinBuilder',
+    didIndyNamespace: 'sovrinBuilder',
     isProduction: false,
     genesisTransactions: 'xxx',
     transactionAuthorAgreement: { version: '1', acceptanceMechanism: 'accept' },
   },
   {
-    id: 'sovrinStaging',
+    id: 'sovringStaging',
+    didIndyNamespace: 'sovringStaging',
```
TimoGlastra, line 41:
```suggestion
    didIndyNamespace: 'sovrin:staging',
```

packages/core/src/modules/ledger/__tests__/IndyPoolService.test.ts
```diff
@@ -24,30 +24,35 @@ const CacheRepositoryMock = CacheRepository as jest.Mock<CacheRepository>
...
   {
     id: 'sovrinMain',
+    didIndyNamespace: 'sovrinMain',
     isProduction: true,
     genesisTransactions: 'xxx',
     transactionAuthorAgreement: { version: '1', acceptanceMechanism: 'accept' },
   },
   {
     id: 'sovrinBuilder',
+    didIndyNamespace: 'sovrinBuilder',
```
TimoGlastra, line 34:
```suggestion
    didIndyNamespace: 'sovrin:builder',
```

packages/core/src/modules/ledger/__tests__/IndyPoolService.test.ts
```diff
@@ -24,30 +24,35 @@ const CacheRepositoryMock = CacheRepository as jest.Mock<CacheRepository>
 const pools: IndyPoolConfig[] = [
   {
     id: 'sovrinMain',
+    didIndyNamespace: 'sovrinMain',
```
TimoGlastra, line 27:
```suggestion
    didIndyNamespace: 'sovrin',
```

packages/core/src/modules/ledger/__tests__/IndyLedgerService.test.ts
```diff
@@ -24,6 +24,7 @@ const CacheRepositoryMock = CacheRepository as jest.Mock<CacheRepository>
 const pools: IndyPoolConfig[] = [
   {
     id: 'sovrinMain',
+    didIndyNamespace: 'sovrinMain',
```
TimoGlastra, line 27:
```suggestion
    didIndyNamespace: 'sovrin',
```

packages/core/src/modules/ledger/IndyPool.ts
```diff
@@ -43,6 +49,7 @@ export class IndyPool {
     this.fileSystem = fileSystem
     this.poolConfig = poolConfig
     this.logger = logger
+    this.indyNamespace = createQualifiedIdentifier(this.poolConfig.didIndyNamespace, this.poolConfig.id)
```
TimoGlastra, line 52: This doesn't seem right. The namespace is just the didIndyNamespace value right? This will create a full did.

packages/core/src/modules/ledger/IndyPool.ts
```diff
@@ -20,6 +23,7 @@ export interface IndyPoolConfig {
   genesisTransactions?: string
   id: string
   isProduction: boolean
+  didIndyNamespace: DidIndyNamespace
```
TimoGlastra, line 26: Why do we use `didIndyNamespace` in some places and `indyNamespace` in others? I think it would be nice to stay consistent across the codebase

TimoGlastra (commented):
This is great @morrieinmaas! I'm not sure on the direction of adding the unqualified identifiers and methods to the repositories and records. As descried in #944 the records would be stored with the qualified indy dids that would overwrite the unqualified indy dids:

>add fully qualified identifier to the anon creds (schema|credential definition) records (replace the non qualified identifier before storing and after retrieving it). For indy we use the did:indy spec identifiers to store.
> - Note: Records stores fully qualified identifier as identifier.
> - Need to transform unqualified to qualified before storing or querying
> - Need to transform qualified to unqualified after retrieving

Was there a specific reason why you deviated from the described approach?

blu3beri:
> This is great @morrieinmaas! I'm not sure on the direction of adding the unqualified identifiers and methods to the repositories and records. As descried in #944 the records would be stored with the qualified indy dids that would overwrite the unqualified indy dids:
> Was there a specific reason why you deviated from the described approach?

I am partly to blame for this and I am certain that there is a better way to do it but here was my reasoning:

1. We never store the unqualified identifier, we just add id as a tag to query by `credentialDefinitionId` or `schemaId`. 
2. When we have one of those identifiers we do not have the namespace and querying by qualified identifier is impossible then. 
3. Converting to only support qualified identifiers would simply swap the usafe of `(find|get)byunqualifiedXXX`to `(find|get)byqualified` in the repository and instead of getting the `unqualifiedIdentifier` on the record you now get the `qualifiedIdentifier` on the record.

Swapping the identifier to qualified before storing and after retrieving I think comes with inherent issues. Every possible way of updating/adding/getting in the repository has to be checked and changed and will cause "weird" code.

I am 10000% open for better implementations. This to me however looked like the quickest way to later convert to `qualifiedIdentifiers` but I would not be suprised if I overcomplicated things here. :)

packages/core/src/modules/indy/repository/AnonCredsCredentialDefinitionRecord.ts
```diff
@@ -1,39 +1,54 @@
...
 export type DefaultAnonCredsCredentialDefinitionTags = {
-  credentialDefinitionId: string
-  issuerDid: string
-  schemaId: string
-  tag: string
+  qualifiedIdentifier: string
+  unqualifiedIdentifier: string
 }
 
 export class AnonCredsCredentialDefinitionRecord extends BaseRecord<DefaultAnonCredsCredentialDefinitionTags> {
```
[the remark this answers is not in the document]
blu3beri: I am not sure how I feel about using the `AnonCredsCredentailDefinitionRecord.id` for the qualifiedIdentifier. Every other record, that I am aware of, uses it as an internal tracker and has no actual relation to a "real value".

In the record itself it is stored as `identifier` which could be changed to `resourceIdentifier` to specify that that is the identifier for whatever resource it is stored upon. Having it like this does not make it indy specific right?

packages/core/tests/ledger.test.ts
blu3beri: As per the did-core spec [here](https://www.w3.org/TR/did-core/#did-syntax)

Its defined like so (IMO not the most understandle way to define it...): 

```
did                = "did:" method-name ":" method-specific-id
method-name        = 1*method-char
method-char        = %x61-7A / DIGIT
method-specific-id = *( *idchar ":" ) 1*idchar
idchar             = ALPHA / DIGIT / "." / "-" / "_" / pct-encoded
pct-encoded        = "%" HEXDIG HEXDIG
```

The `didIndyNamespace` would be in the `method-specific-id` which does not allow spaces.

tl;dr no they can not.

packages/core/src/utils/indyIdentifiers.ts
```diff
@@ -0,0 +1,24 @@
...
+  const subStrings = qualifiedIdentifier.split(':')
+  return subStrings[subStrings.length - 1]
```
blu3beri, line 14: I am a bit confused here so please help me understand this. 

How can we go from a qualified identifier to a credential definition id? I thought that that was the purpose of this method, along with unqualify a did and schemaId.

This method will not unqualify a credentialDefinitionId properly but maybe we are doing that somewhere else (I can't find where though).

TimoGlastra:
Also after thinking about it a bit more this doesn't seem to follow the indy did method specification for schema / credentials definitions ids:
https://hyperledger.github.io/indy-did-method/#did-urls-for-indy-object-identifiers

TimoGlastra:
@blu3beri I'm not on the same page with you on this one.  The goal of the AnonCreds records is to make them generic already, so we can get away from the endless unqualified legacy indy dids issues. By replacing the id before storing on the object itself, we make the id already follow the did:indy specification. The nice thing is, once we add support for did:indy it will just work and we won't have to transform the records anymore because we're already following the spec. So I would argue strongly against any tags or properties that have to do with unqualified dids.


If we have the unqualified schema/cred def id and the namespace we can transform that into the did:indy qualified shcema/cred def id and find it by the qualified id. if we have the qualified id we can reduce it to an unqualified id.

As an example the steps of creating / retrieving a schema:
1. Construct qualified indy did schema id from the input parameters:
  - name: myschema, version: 1.0.0, did: 1234. pool: { id: 'test', indyNamespace: 'test' }
  - qualified indy did: `did:indy:test:1234/anoncreds/v0/SCHEMA/myschema/1.0.0`
2. Query for existing anon creds schema record where id tag is `did:indy:test:1234/anoncreds/v0/SCHEMA/myschema/1.0.0`
3. a record exists:
  - Transorm qualified indy did schema id into uqualified schema id `1234:2:myschema:1.0.0`
  - Replace id property on record schema with unqualified shcema id
  - return record
3b Record does not exist
	- Create schema using indy sdk
	- Create record, replace id with qualified indy did schema from step 1 before returning
	- Store record with qualified did
	- Return schema with unqualified did

Maybe I'm not fully understanding what you're saying, but I don't see the need for these properties?

blu3beri:
> @blu3beri I'm not on the same page with you on this one. The goal of the AnonCreds records is to make them generic already, so we can get away from the endless unqualified legacy indy dids issues. By replacing the id before storing on the object itself, we make the id already follow the did:indy specification. The nice thing is, once we add support for did:indy it will just work and we won't have to transform the records anymore because we're already following the spec. So I would argue strongly against any tags or properties that have to do with unqualified dids.

Yeah I agree with you. I could not really find a way to retrieve a anoncredsrecord ONLY via the qualified identifier when we do not know the namespace (if this is even required). If, in every context, we are aware of the namespace, 100% remove every reference to unqualified identifiers.

> If we have the unqualified schema/cred def id and the namespace we can transform that into the did:indy qualified shcema/cred def id and find it by the qualified id. if we have the qualified id we can reduce it to an unqualified id.
> 
> As an example the steps of creating / retrieving a schema:
> 
> 1. Construct qualified indy did schema id from the input parameters:

Where do we get these input parameters from?

> Maybe I'm not fully understanding what you're saying, but I don't see the need for these properties?

Yeah I don't think it was the best explanation, and I probably don't understand the exact situation we are in. Where we use unqualified or qualified and what we get back from the ledger.

TimoGlastra:
> Where do we get these input parameters from?

From the api, except for the pool, currently that is always the first one. Later on we can allow to specify the qualified did you'd like to use to register it and we can extract it from the qualified did

blu3beri:
> From the api, except for the pool, currently that is always the first one. Later on we can allow to specify the qualified did you'd like to use to register it and we can extract it from the qualified did

Ahh okay now I think I fully understand it. I assume with first one you mean ledger in your configuration?

morrieinmaas:
wow. thx for all the feedback guys🙏🚀. Really nice to have hit some hot topic that triggered a good discussion. I'll have a proper read through and will reply more in depth tomorrow

packages/core/src/modules/ledger/IndyPool.ts
morrieinmaas: ah yeah you're right

packages/core/src/utils/indyIdentifiers.ts
morrieinmaas: hmm maybe I'm getting this wrong here but I thought extracting the `6cgbu8ZPoWTnR5Rv5JcSMB` part from sth like `did:indy:sovrin:staging:6cgbu8ZPoWTnR5Rv5JcSMB` would be what we want here?

morrieinmaas:
> @blu3beri I'm not on the same page with you on this one. The goal of the AnonCreds records is to make them generic already, so we can get away from the endless unqualified legacy indy dids issues. By replacing the id before storing on the object itself, we make the id already follow the did:indy specification. The nice thing is, once we add support for did:indy it will just work and we won't have to transform the records anymore because we're already following the spec. So I would argue strongly against any tags or properties that have to do with unqualified dids.
> 
> If we have the unqualified schema/cred def id and the namespace we can transform that into the did:indy qualified shcema/cred def id and find it by the qualified id. if we have the qualified id we can reduce it to an unqualified id.
> 
> As an example the steps of creating / retrieving a schema:
> 
>     1. Construct qualified indy did schema id from the input parameters:
> 
> 
>     * name: myschema, version: 1.0.0, did: 1234. pool: { id: 'test', indyNamespace: 'test' }
> 
>     * qualified indy did: `did:indy:test:1234/anoncreds/v0/SCHEMA/myschema/1.0.0`
> 
> 
>     2. Query for existing anon creds schema record where id tag is `did:indy:test:1234/anoncreds/v0/SCHEMA/myschema/1.0.0`
> 
>     3. a record exists:
> 
> 
>     * Transorm qualified indy did schema id into uqualified schema id `1234:2:myschema:1.0.0`
> 
>     * Replace id property on record schema with unqualified shcema id
> 
>     * return record
>       3b Record does not exist
>       
>       * Create schema using indy sdk
>       * Create record, replace id with qualified indy did schema from step 1 before returning
>       * Store record with qualified did
>       * Return schema with unqualified did
> 
> 
> Maybe I'm not fully understanding what you're saying, but I don't see the need for these properties?

Ahh thx for the exmaple @TimoGlastra - I was missing some info there which is also why you correctly were puzzled about [this](https://hyperledger.github.io/indy-did-method/#did-urls-for-indy-object-identifiers) part. I was following [this](https://hyperledger.github.io/indy-did-method/#indy-did-method-identifiers) instead but I get it now

packages/core/src/modules/indy/repository/AnonCredsCredentialDefinitionRecord.ts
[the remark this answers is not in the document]
morrieinmaas: Any more thoughts on this?

packages/core/src/utils/indyIdentifiers.ts
blu3beri: Well yes that is what this method is doing. But is there anywehere in the code where we go from a qualified indentifier, for cred def or something, to an unqualified indentifier?

packages/core/tests/ledger.test.ts
blu3beri: I think I have the same point as @TimoGlastra mentioned here: https://github.com/hyperledger/aries-framework-javascript/pull/965#issuecomment-1195815884

packages/core/src/modules/indy/repository/AnonCredsCredentialDefinitionRecord.ts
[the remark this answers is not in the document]
TimoGlastra: I mean modifying the actual id of the schema / cred def.

The schem object has the following structure:

```ts
export interface Schema {
    id: SchemaId;
    attrNames: string[];
    name: string;
    version: string;
    ver: string;
    seqNo: number;
}
```

I'm proposing we modify the schema.id property itself, not the record property. Because this is what the schema.id property will look like once we have did:indy identifiers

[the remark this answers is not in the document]
TimoGlastra: So not adding any new properties besides the schema / cred def object itself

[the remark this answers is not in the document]
blu3beri: Yeah thats okay by me.

packages/core/src/utils/indyIdentifiers.ts
morrieinmaas: I'm for unqualifyIndyDid

packages/core/src/modules/ledger/services/IndyLedgerService.ts
morrieinmaas: ah yesssss. forgot to update the deps

packages/core/src/utils/indyIdentifiers.ts
morrieinmaas: we do that when we store the as anonCreds with qualified id and retrieve it via the LedgerApi. The api then returns the unqualified one. Updated/refactored this now anyway. Will push in a bit

packages/core/src/modules/dids/methods/sov/SovDidRegistrar.ts
```diff
@@ -102,9 +102,9 @@ export class SovDidRegistrar implements DidRegistrar {
       // Build did document.
       const didDocument = didDocumentBuilder.build()
 
-      // FIXME: we need to update this to the `indyNamespace` once https://github.com/hyperledger/aries-framework-javascript/issues/944 has been resolved
-      const indyNamespace = this.indyPoolService.ledgerWritePool.config.id
-      const qualifiedIndyDid = `did:indy:${indyNamespace}:${unqualifiedIndyDid}`
+      // FIXME: create qualified identifier
```
TimoGlastra, line 105: This is already fixed right?

packages/core/src/modules/indy/repository/AnonCredsCredentialDefinitionRecord.ts
```diff
@@ -1,39 +1,47 @@
...
 
 import { BaseRecord } from '../../../storage/BaseRecord'
-import { didFromCredentialDefinitionId } from '../../../utils/did'
+import { getQualifiedIdentifier } from '../../../utils/indyIdentifiers'
+import { CredentialDefinitionTransformer } from '../../../utils/transformers'
+import { uuid } from '../../../utils/uuid'
 
 export interface AnonCredsCredentialDefinitionRecordProps {
   credentialDefinition: CredDef
+  didIndyNamespace: DidIndyNamespace
```
TimoGlastra, line 11: I think this isn't needed

packages/core/src/modules/indy/repository/AnonCredsCredentialDefinitionRecord.ts
```diff
@@ -1,39 +1,47 @@
...
-}
-
-export class AnonCredsCredentialDefinitionRecord extends BaseRecord<DefaultAnonCredsCredentialDefinitionTags> {
+export class AnonCredsCredentialDefinitionRecord extends BaseRecord {
   public static readonly type = 'AnonCredsCredentialDefinitionRecord'
   public readonly type = AnonCredsCredentialDefinitionRecord.type
-  public readonly credentialDefinition!: CredDef
+
+  @CredentialDefinitionTransformer()
+  public readonly credentialDefinition!: CredDef & { didIndyNamespace?: DidIndyNamespace; schemaSeqNo?: number }
```
TimoGlastra, line 20: I don't think we need this. We can extract the did indy namespace from the identifier and the schemaSeqNo from the schemaID property

```suggestion
  public readonly credentialDefinition!: CredDef
```

packages/core/src/modules/indy/repository/AnonCredsCredentialDefinitionRecord.ts
```diff
@@ -1,39 +1,47 @@
...
+import { uuid } from '../../../utils/uuid'
 
 export interface AnonCredsCredentialDefinitionRecordProps {
   credentialDefinition: CredDef
+  didIndyNamespace: DidIndyNamespace
+  schemaSeqNo: number
 }
 
-export type DefaultAnonCredsCredentialDefinitionTags = {
-  credentialDefinitionId: string
```
TimoGlastra, line 11: I think we want to keep the credentialDefinition as a tag?

packages/core/src/modules/indy/repository/AnonCredsCredentialDefinitionRecord.ts
```diff
@@ -1,39 +1,47 @@
...
+      this.credentialDefinition = {
+        ...props.credentialDefinition,
+        didIndyNamespace: props.didIndyNamespace,
+        schemaSeqNo: props.schemaSeqNo,
+      }
+      this.schemaSeqNo = props.schemaSeqNo
+      this._tags.id = getQualifiedIdentifier(props.didIndyNamespace, {
+        ...this.credentialDefinition,
+        schemaSeqNo: props.schemaSeqNo,
+      })
```
TimoGlastra, line 38: I think it is better to do the transformation outside of the record class as the record will be used by other anoncreds services.


```suggestion
      this.credentialDefinition = props.credentialDefinition
```

packages/core/src/modules/indy/repository/AnonCredsCredentialDefinitionRecord.ts
```diff
@@ -1,39 +1,47 @@
...
     }
   }
 
   public getTags() {
     return {
       ...this._tags,
-      credentialDefinitionId: this.credentialDefinition.id,
```
TimoGlastra, line 33: I think we should keep the cred def id tag

packages/core/src/modules/indy/repository/AnonCredsSchemaRecord.ts
```diff
@@ -1,39 +1,36 @@
...
 
-export type DefaultAnonCredsSchemaTags = {
-  schemaId: string
-  schemaIssuerDid: string
-  schemaName: string
-  schemaVersion: string
-}
-
-export class AnonCredsSchemaRecord extends BaseRecord<DefaultAnonCredsSchemaTags> {
+export class AnonCredsSchemaRecord extends BaseRecord {
```
TimoGlastra, line 14: same suggestions as the cred def class

packages/core/src/modules/ledger/IndyPool.ts
```diff
@@ -31,6 +33,8 @@ export class IndyPool {
   private _poolHandle?: number
   private poolConnected?: Promise<number>
   public authorAgreement?: AuthorAgreement | null
+  // aka the qualified identifier
+  public didIndyNamespace: string
```
TimoGlastra, line 37: We can make this a getter (like we did for the id property). This way we don't have to create another property

packages/core/src/modules/ledger/LedgerApi.ts
```diff
@@ -75,14 +84,33 @@ export class LedgerApi {
 
     const schemaId = generateSchemaId(did, schema.name, schema.version)
 
+    // Generate teh qualified ID
+    // FIXME: How to handle indyLedgers[0]? This should not require selecting the first on eby default
```
TimoGlastra, line 88: We should get the ledgerWritePool from the ledgerservice

packages/core/src/modules/ledger/LedgerApi.ts
```diff
@@ -75,14 +84,33 @@ export class LedgerApi {
...
+
     // Try find the schema in the wallet
-    const schemaRecord = await this.anonCredsSchemaRepository.findBySchemaId(this.agentContext, schemaId)
-    //  Schema in wallet
-    if (schemaRecord) return schemaRecord.schema
+    const schemaRecord = await this.anonCredsSchemaRepository.findById(this.agentContext, qualifiedIdentifier)
+    // Schema in wallet
+    if (schemaRecord) {
+      // Transform qualified to unqualified
+      // schemaRecord.schema.id = unqualifyIndyDid(schemaRecord.schema.id)
```
TimoGlastra, line 99: This shouldn't be commented right?

packages/core/src/modules/ledger/LedgerApi.ts
```diff
@@ -75,14 +84,33 @@ export class LedgerApi {
...
     if (schemaFromLedger) return schemaFromLedger
-    return this.ledgerService.registerSchema(this.agentContext, did, schema)
+    const createdSchema = await this.ledgerService.registerSchema(this.agentContext, did, schema)
+
+    const anonCredsSchema = new AnonCredsSchemaRecord({
+      schema: createdSchema,
+      didIndyNamespace: this.config.indyLedgers[0].didIndyNamespace,
+    })
+    await this.anonCredsSchemaRepository.save(this.agentContext, anonCredsSchema)
+    return createdSchema
```
TimoGlastra, line 113: We should update the id before storing it here

packages/core/src/modules/ledger/LedgerApi.ts
```diff
@@ -137,10 +179,20 @@ export class LedgerApi {
...
 
     // Register the credential
-    return await this.ledgerService.registerCredentialDefinition(this.agentContext, did, {
+    const registeredCredential = await this.ledgerService.registerCredentialDefinition(this.agentContext, did, {
       ...credentialDefinitionTemplate,
       signatureType: 'CL',
     })
+    // Replace the unqualified with qualified Identifier
+    // credentialDefinitionTemplate.schema.id = qualifiedIdentifier
+    const anonCredCredential = new AnonCredsCredentialDefinitionRecord({
```
TimoGlastra, line 188: same here, should update the id before storing

packages/core/src/modules/ledger/__tests__/IndyLedgerService.test.ts
TimoGlastra, line 27:
```suggestion
    didIndyNamespace: 'sovrin',
```

packages/core/src/modules/ledger/ledgerUtil.ts
```diff
@@ -15,3 +16,32 @@ export function generateSchemaId(did: string, name: string, version: string) {
...
+    type: 'BaseRecord', // Irrelevant for qualifiedIdentifier construction
+    tag: credentialDefinitionTemplate.tag,
+    version: credentialDefinitionTemplate.schema.version,
+    ver: credentialDefinitionTemplate.schema.ver,
+    value: {
+      primary: { ...credentialDefinitionTemplate.schema },
+      revocation: credentialDefinitionTemplate.supportRevocation,
+    },
+  } as Indy.CredDef
+}
```
TimoGlastra, line 36: Why is this needed? It creates an invalid object so would rather skip the step to go to cred def

packages/core/src/modules/ledger/ledgerUtil.ts
```diff
@@ -15,3 +16,32 @@ export function generateSchemaId(did: string, name: string, version: string) {
...
+
+export function generateSchemaFromTemplate(schemaId: string, schema: SchemaTemplate): Indy.Schema {
```
TimoGlastra, line 38: Same here

packages/core/src/utils/indyIdentifiers.ts
```diff
@@ -0,0 +1,91 @@
...
+export type DidIndyMethod = 'indy'
+// Maybe this can be typed more strictly than string. Choosing string for now as this can be eg just `sovrin` or eg `sovrin:staging`
+export type DidIndyNamespace = string
+//  NOTE: because of the ambiguous nature - whether there is a colon or not within DidIndyNamespace this is the substring after the ***last*** colon
+export type NamespaceIdentifier = string
+
+// TODO: This template literal type can possibly be improved. This version leaves the substrings as potentially undefined
+export type IndyNamespace = `${Did}:${DidIndyMethod}:${DidIndyNamespace}:${NamespaceIdentifier}`
+
+export function isQualifiedIdentifier(identifier: string | undefined): boolean {
```
TimoGlastra, line 23:
```suggestion
export function isQualifiedIndyIdentifier(identifier: string): boolean {
```

packages/core/src/utils/indyIdentifiers.ts
```diff
@@ -0,0 +1,91 @@
...
+  } else if (txType === 'CLAIM_DEF') {
+    // did:indy:sovrin:5nDyJVP1NrcPAttP3xwMB9/anoncreds/v0/CLAIM_DEF/56495/npdb -> 5nDyJVP1NrcPAttP3xwMB9:3:CL:56495:npbd
+    const [id, , , , seqNo, name] = identifierTrunk.split('/')
+    return generateCredentialDefinitionId(id, +seqNo, name)
+  } else {
+    // indy Union Indy Ledger w/o url syntax
+    // did:indy:idunion:test:2MZYuPv2Km7Q1eD4GCsSb6 -> 2MZYuPv2Km7Q1eD4GCsSb6
+    return identifierTrunk
+  }
+}
```
TimoGlastra, line 48: We always know if we're working with a schema or cred def I think. My preference here would be to create two methods; `getLegacyIndyCredentialDefinitionId` and `getLegacyIndySchemaId`.

packages/core/src/utils/indyIdentifiers.ts
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ *
+ * @see https://hyperledger.github.io/indy-did-method/#schema
+ *
+ */
+export function getDidUrlTrunkFromSchema(schema: Schema): string {
```
TimoGlastra, line 55: `schemaToQualifiedIndySchemaId`

packages/core/src/utils/indyIdentifiers.ts
```diff
@@ -0,0 +1,91 @@
...
+  const namespaceIdentifier = didFromSchemaId(schema.id)
+  const didUrl = `${namespaceIdentifier}/anoncreds/v0/SCHEMA/${schema.name}/${schema.version}`
```
TimoGlastra, line 57: The schema id should already contain this part or it should be a legacy identifier, in which case it won't include the indy namspace identifier. I think this state we're assuming here is invalid

packages/core/src/utils/indyIdentifiers.ts
```diff
@@ -0,0 +1,91 @@
...
+  return didUrl
+}
+
+/**
+ *
+ * @see https://hyperledger.github.io/indy-did-method/#cred-def
+ *
+ */
+export function getDidUrlTrunkFromCredDef(credDef: CredDef & { schemaSeqNo: number }): string {
+  const namespaceIdentifier = didFromCredentialDefinitionId(credDef.id)
```
TimoGlastra, line 67: Same here

packages/core/src/utils/indyIdentifiers.ts
```diff
@@ -0,0 +1,91 @@
...
+  return `${namespaceIdentifier}/anoncreds/v0/CLAIM_DEF/${credDef.schemaSeqNo}/${credDef.tag}`
+}
+
+export function getDidUrlTrunk(data: Schema | (CredDef & { schemaSeqNo: number })): string {
```
TimoGlastra, line 71: I think this method tries to be too generic, and I think we can just call the specific resource methods when needed (We always know whether we have a schema or cred def)

packages/core/src/utils/transformers.ts
```diff
@@ -45,6 +46,49 @@ export function RecordTransformer<T>(Class: { new (...args: any[]): T }) {
   })
 }
 
+export function CredentialDefinitionTransformer() {
```
TimoGlastra, line 49: Maybe we can place these next to the record classes so they're bundled with the correct module

morrieinmaas:
Thanks for the feedback @TimoGlastra

packages/core/src/modules/indy/repository/AnonCredsCredentialDefinitionRecord.ts
morrieinmaas: let me double-check this. Added it for the transformer.

packages/core/src/modules/indy/repository/AnonCredsCredentialDefinitionRecord.ts
morrieinmaas: will also double-check whether/how to refactor this out.

packages/core/src/modules/indy/repository/AnonCredsCredentialDefinitionRecord.ts
morrieinmaas: I'll have a look

packages/core/src/modules/ledger/LedgerApi.ts
morrieinmaas: ahhhh alright. Thx!

packages/core/src/modules/ledger/LedgerApi.ts
morrieinmaas: yes indeed

packages/core/src/utils/indyIdentifiers.ts
morrieinmaas: hmmm ok. Let me have a look at this. Personally unsure right now but you're probably right

packages/core/src/modules/indy/repository/AnonCredsCredentialDefinitionRecord.ts
morrieinmaas: hmm not sure why this si showing here. ALready removed this in my last commit (locally)

packages/core/src/modules/indy/repository/AnonCredsCredentialDefinitionRecord.ts
morrieinmaas: Might be missing sth here but how do I get the schemaSeqNo from the schemaId?

morrieinmaas: as discussed (just for the record) cannot get schemaSeqNo from schemaId but from extra ledger request to resolve the schema via the credDef

packages/core/src/modules/indy/repository/AnonCredsCredentialDefinitionRecord.ts
morrieinmaas: hmm ok. Happy to add this back in. Just out of curiousity why do you prefer this over credentialDefinition.id?

packages/core/src/modules/ledger/ledgerUtil.ts
morrieinmaas: refactoring out. Indeed not needed. had this because some function was expecting credDef instead of template mainly because of the .id prop on it but we can just pass this as params instead of creating a dummie cred def from `Omit<CredentialDefinitionTemplate, "signatureType">`

packages/core/src/utils/indyIdentifiers.ts
morrieinmaas: Not sure I fully get your remark here. As per https://hyperledger.github.io/indy-did-method/#schema this gets `F72i3Y3Q4i466efjYJYCHM` from `F72i3Y3Q4i466efjYJYCHM:2:npdb:4.3.4` and creates the 'trunk' (`F72i3Y3Q4i466efjYJYCHM/anoncreds/v0/SCHEMA/npdb/4.3.4`) to be added to `did:indy:sovrin` creating the entire `did:indy:sovrin:F72i3Y3Q4i466efjYJYCHM/anoncreds/v0/SCHEMA/npdb/4.3.4`

What am I missing/misunderstanding here?

packages/core/src/utils/indyIdentifiers.ts
morrieinmaas: same question as above

packages/core/src/utils/indyIdentifiers.ts
blu3beri: What @TimoGlastra means, I believer, is that the `schema.id` should already have this `didUrl` part as it is the schema id.

A legacy schema id does have the `namespace` in the `schema.id` so we can never extract this as is done in line 56.

Also, `didFromSchemaId(schema.id)`, for legacy and qualified, will never return a namespace idenfitier as it just does `"5nDyJVP1NrcPAttP3xwMB9:3:CL:56495:npdb".split(":")[0]` which returns the DID ONLY for unqualified identifiers.

TimoGlastra: Ah I think I misunderstood the terms here. namespaceIdentifier IS the did, correct?

morrieinmaas: yes in this case! Maybe i should rename it did. Just followed the naming from did:indy:<namespace>:<namespace identifier> (also in https://hyperledger.github.io/indy-did-method/#indy-did-method-identifiers, bottom of the section)

packages/core/src/modules/indy/repository/AnonCredsCredentialDefinitionRecord.ts
morrieinmaas: Left schemaSeqNo in for now as I'm unsure where to best do the ledger request here to resolve it from the cred def.

packages/core/src/modules/ledger/LedgerApi.ts
morrieinmaas: Not sure what is going on but my pushed changes are not showing up here

morrieinmaas:
generally not sure what is going on but my git status is up to date and pushed but the change set here is not what I've pushed

packages/core/src/modules/credentials/formats/indy/IndyCredentialFormatService.ts
```diff
@@ -232,6 +232,7 @@ export class IndyCredentialFormatService extends CredentialFormatService<IndyCre
         credentialDefinition,
       }
     )
+
```
TimoGlastra, line 235:
```suggestion
```

packages/core/src/modules/ledger/ledgerUtil.ts
```diff
@@ -15,3 +16,14 @@ export function generateSchemaId(did: string, name: string, version: string) {
 export function generateCredentialDefinitionId(did: string, seqNo: number, tag: string) {
   return `${did}:3:CL:${seqNo}:${tag}`
 }
+
+export function generateSchemaFromTemplate(schemaId: string, schema: SchemaTemplate): Indy.Schema {
```
TimoGlastra, line 20: this method is unused I think

packages/core/src/utils/indyIdentifiers.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  if (txType === 'CLAIM_DEF') {
+    // did:indy:sovrin:5nDyJVP1NrcPAttP3xwMB9/anoncreds/v0/CLAIM_DEF/56495/npdb -> 5nDyJVP1NrcPAttP3xwMB9:3:CL:56495:npbd
+    const [id, , , , seqNo, name] = identifierTrunk.split('/')
+    return generateCredentialDefinitionId(id, +seqNo, name)
+  } else {
+    // indy Union Indy Ledger w/o url syntax
+    // did:indy:idunion:test:2MZYuPv2Km7Q1eD4GCsSb6 -> 2MZYuPv2Km7Q1eD4GCsSb6
+    return identifierTrunk
+  }
```
TimoGlastra, line 43: This isn't a credential definition id right? Should we throw an error instead?

packages/core/src/utils/indyIdentifiers.ts
```diff
@@ -0,0 +1,114 @@
...
+// TODO: This template literal type can possibly be improved. This version leaves the substrings as potentially undefined
+export type IndyNamespace = `${Did}:${DidIndyMethod}:${DidIndyNamespace}:${NamespaceIdentifier}`
+
+export function isQualifiedIndyIdentifier(identifier: string | undefined): boolean {
+  if (!identifier || identifier === '') return false
+  return identifier.startsWith('did:indy:')
+}
+
+export function getLegacyIndyCredentialDefinitionId(qualifiedIdentifier: string) {
+  if (!isQualifiedIndyIdentifier(qualifiedIdentifier)) return qualifiedIdentifier
```
TimoGlastra, line 29: should we throw an error here? Or do you think it it better to just silently continue?

packages/core/src/utils/indyIdentifiers.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  if (txType === 'SCHEMA') {
+    // did:indy:sovrin:F72i3Y3Q4i466efjYJYCHM/anoncreds/v0/SCHEMA/npdb/4.3.4 -> F72i3Y3Q4i466efjYJYCHM:2:npdb:4.3.4
+    const [id, , , , name, version] = identifierTrunk.split('/')
+    return generateSchemaId(id, name, version)
+  } else {
+    // indy Union Indy Ledger w/o url syntax
+    // did:indy:idunion:test:2MZYuPv2Km7Q1eD4GCsSb6 -> 2MZYuPv2Km7Q1eD4GCsSb6
+    return identifierTrunk
+  }
```
TimoGlastra, line 61: Same here, would be better to throw an error in this case I think

packages/core/src/utils/indyIdentifiers.ts
```diff
@@ -0,0 +1,114 @@
...
+
+export function getQualifiedIdentifierSchema(
+  indyNamespace: string,
+  schemaTemplate: SchemaTemplate | Schema,
+  schemaId: string
+): IndyNamespace {
+  if (isQualifiedIndyIdentifier(schemaId)) return schemaId as IndyNamespace
+
+  const didUrl = schemaToQualifiedIndySchemaId(schemaTemplate, schemaId)
+  return `did:indy:${indyNamespace}:${didUrl}`
```
TimoGlastra, line 113: This is a nit, but I would expect the `schemaToQualifiedIndySchemaId` to return the fully qualified schema id, which includes the `did:indy:${indyNamespace}` part. Maybe we can call it `schemaToQualifiedIndySchemaTrunk` and only return the trunk (after the identifier part), or we just return the fully qualified schema id from `schemaToQualifiedIndySchemaId`

TimoGlastra (commented):
Some final remarks, but we're getting there! This is looking good!

packages/core/src/utils/indyIdentifiers.ts
morrieinmaas: hmm good question - tbh I'm unsure

packages/core/src/utils/indyIdentifiers.ts
morrieinmaas: Also rather unsure. Gut feeling is silently continue but I do see the benfit of throwing an error

packages/core/src/modules/ledger/IndyPool.ts
```diff
@@ -31,6 +33,7 @@ export class IndyPool {
   private _poolHandle?: number
   private poolConnected?: Promise<number>
   public authorAgreement?: AuthorAgreement | null
+  // aka the qualified identifier
```
TimoGlastra, line 36:
```suggestion
```

packages/core/src/modules/ledger/LedgerApi.ts
```diff
@@ -75,14 +83,31 @@ export class LedgerApi {
...
+
     // Try find the schema in the wallet
-    const schemaRecord = await this.anonCredsSchemaRepository.findBySchemaId(this.agentContext, schemaId)
-    //  Schema in wallet
-    if (schemaRecord) return schemaRecord.schema
+    const schemaRecord = await this.anonCredsSchemaRepository.findById(this.agentContext, qualifiedIdentifier)
+    // Schema in wallet
+    if (schemaRecord) {
+      // Transform qualified to unqualified
+      schemaRecord.schema.id = getLegacyIndySchemaId(schemaRecord.schema.id)
```
TimoGlastra, line 94:
```suggestion
```

packages/core/src/modules/ledger/LedgerApi.ts
```diff
@@ -75,14 +83,31 @@ export class LedgerApi {
...
+      return {
+        ...schemaRecord?.schema,
```
TimoGlastra, line 96:
```suggestion
        ...schemaRecord.schema,
```

packages/core/src/modules/ledger/LedgerApi.ts
```diff
@@ -121,12 +146,27 @@ export class LedgerApi {
...
     )
-    if (credentialDefinitionRecord) return credentialDefinitionRecord.credentialDefinition
+
+    // Credential Definition in wallet
+    if (credentialDefinitionRecord) {
+      // Transform qualified to unqualified
+      credentialDefinitionRecord.credentialDefinition.id = getLegacyIndyCredentialDefinitionId(
+        credentialDefinitionRecord.credentialDefinition.id
+      )
+      return credentialDefinitionRecord.credentialDefinition
```
TimoGlastra, line 168:
```suggestion
      return {
        ...credentialDefinitionRecord.credentialDefinition,
        id: getLegacyIndyCredentialDefinitionId(credentialDefinitionRecord.credentialDefinition.id)
```

packages/core/src/utils/indyIdentifiers.ts
```diff
@@ -0,0 +1,117 @@
...
+
+export function isQualifiedIndyIdentifier(identifier: string | undefined): boolean {
+  if (!identifier || identifier === '') return false
+  return identifier.startsWith('did:indy:')
+}
+
+export function getLegacyIndyCredentialDefinitionId(qualifiedIdentifier: string) {
+  if (!isQualifiedIndyIdentifier(qualifiedIdentifier))
+    throw new AriesFrameworkError(
+      `Identifier ${qualifiedIdentifier} not a qualified indy identifier. Hint: Needs to start with 'did:ind:'`
```
TimoGlastra, line 32:
```suggestion
      `Identifier ${qualifiedIdentifier} not a qualified indy identifier. Hint: Needs to start with 'did:indy:'`
```

packages/core/src/utils/indyIdentifiers.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  if (txType === 'SCHEMA') {
+    // did:indy:sovrin:F72i3Y3Q4i466efjYJYCHM/anoncreds/v0/SCHEMA/npdb/4.3.4 -> F72i3Y3Q4i466efjYJYCHM:2:npdb:4.3.4
+    const [id, , , , name, version] = identifierTrunk.split('/')
+    return generateSchemaId(id, name, version)
+  } else {
+    // indy Union Indy Ledger w/o url syntax
+    // did:indy:idunion:test:2MZYuPv2Km7Q1eD4GCsSb6 -> 2MZYuPv2Km7Q1eD4GCsSb6
+    return identifierTrunk
+  }
```
TimoGlastra, line 64: This should also be updated to throw errors like with the cred def to be consistent

TimoGlastra:
Ah damnit, the Github UI signoff feature is broken. S

blu3beri:
Not sure why it resubmitted my previous comments...

morrieinmaas:
Thanks for the feedback @TimoGlastra 🙏 (and sorry for this being less straight-forward than could be 😅)

packages/core/src/modules/ledger/LedgerApi.ts
TimoGlastra, line 94: This is unused with the ... return below

```suggestion
```

packages/core/src/modules/ledger/LedgerApi.ts
```diff
@@ -121,12 +146,27 @@ export class LedgerApi {
...
```
TimoGlastra, line 168:
```suggestion
      // Transform qualified to unqualified
      return {
        ...credentialDefinitionRecord.credentialDefinition,
        id: getLegacyIndyCredentialDefinitionId(credentialDefinitionRecord.credentialDefinition.id)
     }
```

packages/core/src/utils/indyIdentifiers.ts
```diff
@@ -0,0 +1,116 @@
...
+    const [id, , , , seqNo, name] = identifierTrunk.split('/')
+    return generateCredentialDefinitionId(id, +seqNo, name)
+  } else {
+    // indy Union Indy Ledger w/o url syntax
+    throw new AriesFrameworkError(`Provided identifier ${qualifiedIdentifier} has invalid format.`)
+  }
+}
+
+export function getLegacyIndySchemaId(qualifiedIdentifier: string) {
+  if (!isQualifiedIndyIdentifier(qualifiedIdentifier)) return qualifiedIdentifier
```
TimoGlastra, line 50: We should also throw an error here to be consistent with the `getLegacyIndyCredentialDefinitionId` method

packages/core/src/utils/indyIdentifiers.ts
```diff
@@ -0,0 +1,116 @@
...
+/**
+ *
+ * @see https://hyperledger.github.io/indy-did-method/#schema
+ *
+ */
+export function schemaToQualifiedIndySchemaTrunk(schema: SchemaTemplate | Schema, schemaId: string): string {
+  const did = didFromSchemaId(schemaId)
+  const didUrl = `${did}/anoncreds/v0/SCHEMA/${schema.name}/${schema.version}`
+  return didUrl
+}
```
TimoGlastra, line 75: I'm still not too keen on this method, as it returns a variant of the schemaId that we will never need. Maybe we can combine the `getQualifiedIdentifierSchema` and `schemaToQualifiedIndySchemaTrunk` methods?

packages/core/src/utils/indyIdentifiers.ts
```diff
@@ -0,0 +1,116 @@
...
+  credDefId: string,
+  credDef: Omit<CredentialDefinitionTemplate, 'signatureType'> | (CredDef & { schemaSeqNo?: string })
+): string {
+  const did = didFromCredentialDefinitionId(credDefId)
+  let seqNo
+  if ('schema' in credDef) {
+    seqNo = credDef.schema.seqNo
+  } else if ('schemaSeqNo' in credDef) {
+    seqNo = credDef.schemaSeqNo
+  }
```
TimoGlastra, line 92: Where does `schemaSeqNo` come from? I don't think this is still used?

morrieinmaas: this is used to create the anonCreds trunk `${did}/anoncreds/v0/CLAIM_DEF/${seqNo}/${credDef.tag}` so that's needed I reckon. Maybe I'm missing sth here though?

packages/core/src/utils/indyIdentifiers.ts
morrieinmaas: already addressed ?!

packages/core/src/utils/indyIdentifiers.ts
TimoGlastra: I don't think the schemaSeqNo is defined anywhere? It's added to the signature of this method, but I think we shouldn't add custom properties to existing objects.

packages/core/src/utils/indyIdentifiers.ts
TimoGlastra: It's addressed for `getLegacyIndyCredentialDefinitionId`, not `getLegacyIndySchemaId`.

packages/core/src/utils/indyIdentifiers.ts
TimoGlastra: I don't think the schemaSeqNo is defined anywhere? It's added to the signature of this method, but I think we shouldn't add custom properties to existing objects.

morrieinmaas: hmm ok fair enough. was thinking to do it this way to safe a call to getSchema but rather just have it at hand directly. Maybe you have another/better idea?

TimoGlastra: Where do we get the scheqSeqNo from?

TimoGlastra: Maybe it can be a separate property that we pass separately to this method and store separately?

morrieinmaas: we get it from CredentialDefinitionTemplate.schema.seqno which is the type in ledgerApi registerCredentialDefinition

morrieinmaas: > Maybe it can be a separate property that we pass separately to this method and store separately?

yeah that might be possible. Unsure right now where to store that then tho -I'll have a think

morrieinmaas: @TimoGlastra this is the one bit I'm actually unsure how to do better. Happy to implement suggestions. 

Otherwise all feedback addressed

morrieinmaas: @TimoGlastra if you have some time maybe we can work this out (together) and get this merged

TimoGlastra:
@morrieinmaas added a PR to resolve my comment. It seems DCO is failing. Could you squash all commits into one so we can merge?

TimoGlastra (approved):

blu3beri:
> @morrieinmaas added a PR to resolve my comment. It seems DCO is failing. Could you squash all commits into one so we can merge?

We can use the "set dco to pass" and squash when we merge. No need to squash before hand.

TimoGlastra:
Yes that makes sense @blu3beri :-)

morrieinmaas:
ALright, thx. Anything else that needs doing then?

PR #966, opened 2022-07-27
test(credentials): fix flaky tests with events

2byrds:
Include similar fixes for other ice2e tests like @TimoGlastra
started in https://github.com/hyperledger/aries-framework-javascript/pull/893


2byrds:
> Include similar fixes for other ice2e and ppe2e tests like @TimoGlastra started in #893

@TimoGlastra how does auto-accept test look? i'll learn from your suggestions before continuing to other test files. packages/core/src/modules/credentials/protocol/v2/__tests__/v2-credentials-auto-accept.e2e.test.ts

TimoGlastra:
Sorry for the delayed response again. But looks perfect!

2byrds:
> Sorry for the delayed response again. But looks perfect!

No sorry necessary.  You get a mega-gold-star for replying so quickly after returning from vacation.  I thought it was high risk to put that on your queue while you were away. But you got to it faster than I would have guessed :). Okay I'll keep chugging in this direction.

2byrds:
@TimoGlastra in your original ticket you said we should handle the ppv2 e2e tests as well (as the icv2 e2e tests).  I don't see a similar directory and naming for proof tests (like there is for icv2 e2e tests).  Looks like I should probably modify packages/core/tests/proofs.test.ts and proofs-auto-accept.test.ts but I wanted to verify with you, first.
If you want to list all the flaky tests that I should modify that's fine too.

TimoGlastra:
I think we should make these updates to the 0.3.0-pre as that introcuded a lot of changes to the present proof module. If the credential tests are done we can merge this PR and address that in a separate PR.

2byrds:
> I think we should make these updates to the 0.3.0-pre as that introcuded a lot of changes to the present proof module. If the credential tests are done we can merge this PR and address that in a separate PR.

Sounds good.  Triggering review.  I'll create another ticket for PPv2 tests and link it here.

TimoGlastra (approved):

PR #967, opened 2022-07-28
feat(ledger): handle REQNACK response for write request

jakubkoci:
I'm not sure if the check for REQNACK response was omitted in the write request method on purpose, but `registerSchema` is throwing a confusing error.

Previously:
```
TypeError: Cannot read property 'txnMetadata' of undefined
```

After the update:
```
Ledger 'pool-localhost' rejected write transaction request: client request invalid: could not authenticate, verkey for 9gFCquotxSS7ctKG1GJatU cannot be found
```


blu3beri:
I think it was just forgotten to do when it was added in this PR: https://github.com/hyperledger/aries-framework-javascript/pull/626 

I don't see a reason why this would be omitted.

blu3beri (approved):

PR #971, opened 2022-07-29
feat(ledger): Add ledgerService interface

DaevMithran:
Signed-off-by: DaevMithran <daevmithran1999@gmail.com>

Question
- instead of indyLedgers. Can we change it to ledgers : { indy: [ ], cosmos: [ ] } .etc


packages/core/src/agent/Agent.ts
```diff
@@ -98,7 +98,7 @@ export class Agent extends BaseAgent {
     // set the pools on the ledger.
     this.ledger.setPools(this.ledger.config.indyLedgers)
     // As long as value isn't false we will async connect to all genesis pools on startup
-    if (this.ledger.config.connectToIndyLedgersOnStartup) {
+    if (this.ledger.config.connectToLedgersOnStartup) {
```
TimoGlastra, line 101: Changing this config will make it a breaking change. I think we can get this into 0.2.3 and we can make the bigger changes with breaking changes (as described in #977) for the 0.3.0 version.

TimoGlastra:
Thanks for this PR @DaevMithran. Left a suggestion, as I think we can get this into 0.2.x (we would like to get it in before 0.3.0).

Also it seems the tests are failing. Could you take a look?

DaevMithran:
@TimoGlastra I beleive the failing test cases are due to the previous commits. I'll create another pull request for v0.2.3 then. I'll remove the breaking change

naonishijima:
is there any update on this?

PR #972, opened 2022-08-05
feat(proofs): present proof as nested protocol

genaris:
As exemplified in [Aries RFC 0008](https://github.com/hyperledger/aries-rfcs/tree/main/concepts/0008-message-id-and-threading#nested-example), there are certain situations where proofs must be exchanged as part of an inner flow, such as a credential issuance.

This PR attempts to support launching a Present Proof V1 protocol as a sub-protocol from another one, without impacting current Agent behaviour and API (apart from adding some fields and methods). It does not, however, add support for all threading concepts (such as `sender_order` and `receiver_order`), as it might introduce breaking changes and also needs to be more carefully designed.

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -1,0 +1,11 @@
+## Maintainers
+
+### Active Maintainers
+
+| name               | Github                                                     | Discord          |
+| ------------------ | ---------------------------------------------------------- | ---------------- |
+| Berend Sliedrecht  | [@blu3beri](https://github.com/blu3beri)                   | blu3beri#2230    |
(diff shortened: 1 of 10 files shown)


TimoGlastra (approved):

TimoGlastra:
Nice addition @genaris! I think it would be great to look at a generic way to do this (if possible, maybe we should have a base protocol implementation with handlers/records etc..), but for now this will do fine.

genaris:
Thanks @TimoGlastra! And I agree that it would be nice to find an easier way to start any protocol as a child of another one (even if some of them, such as Basic Messages and Action Menu, are not originally meant to work as sub-protocols).

TimoGlastra:
Reminder: this PR needs to be integrated with the PP v2 work that has changed basically everything about the implementation.

PR #974, opened 2022-08-06
feat: Action Menu protocol (Aries RFC 0509) implementation

genaris:
Implementation of module fully supporting [Aries RFC 0509](https://github.com/hyperledger/aries-rfcs/blob/main/features/0509-action-menu/README.md) for both `requester` and `responder` roles.

Protocol states and information are stored in records of type `ActionMenuRecord`. As this protocol currently defines a single active menu per connection, there will be at most 2 records per connection (as theoretically is possible to serve as requester and responder at the same time). While this information could be stored also in Connection Metadata, specific records were created thinking in a future version/extension of this protocol where multiple menus could be deployed (e.g. a menu by thread).

Still plenty of tests to do and fixes/support for error messages, as well as check compatibility with ACA-Py implementation.

Solves #471 

### DIFF
--- a/packages/core/src/modules/action-menu/index.ts
+++ b/packages/core/src/modules/action-menu/index.ts
@@ -1,0 +1,4 @@
+export * from './ActionMenuModule'
+export * from './handlers'
+export * from './messages'
+export * from './services'
--- a/packages/core/src/modules/action-menu/messages/MenuMessage.ts
+++ b/packages/core/src/modules/action-menu/messages/MenuMessage.ts
@@ -1,0 +1,53 @@
+import type { ActionMenuOptionOptions } from '../models'
+
+import { Type } from 'class-transformer'
+import { IsInstance, IsOptional, IsString } from 'class-validator'
+
+import { AgentMessage } from '../../../agent/AgentMessage'
+import { IsValidMessageType, parseMessageType } from '../../../utils/messageType'
+import { ActionMenuOption } from '../models'
+
+export interface MenuMessageOptions {
+  id?: string
+  title: string
+  description: string
+  errormsg?: string
+  options: ActionMenuOptionOptions[]
+  threadId?: string
+}
+
+export class MenuMessage extends AgentMessage {
+  public constructor(options: MenuMessageOptions) {
+    super()
+
+    if (options) {
+      this.id = options.id ?? this.generateId()
+      this.title = options.title
+      this.description = options.description
+      this.options = options.options.map((p) => new ActionMenuOption(p))
+      if (options.threadId) {
+        this.setThread({
+          threadId: options.threadId,
+        })
+      }
+    }
+  }
+
+  @IsValidMessageType(MenuMessage.type)
+  public readonly type = MenuMessage.type.messageTypeUri
+  public static readonly type = parseMessageType('https://didcomm.org/action-menu/1.0/menu')
+
+  @IsString()
+  public title!: string
+
+  @IsString()
+  public description!: string
+
+  @IsString()
+  @IsOptional()
+  public errormsg?: string
+
+  @IsInstance(ActionMenuOption, { each: true })
+  @Type(() => ActionMenuOption)
+  public options!: ActionMenuOption[]
+}
--- a/packages/core/src/modules/action-menu/services/ActionMenuService.ts
+++ b/packages/core/src/modules/action-menu/services/ActionMenuService.ts
@@ -1,0 +1,282 @@
+import type { InboundMessageContext } from '../../../agent/models/InboundMessageContext'
+import type { Logger } from '../../../logger'
+import type { ActionMenuStateChangedEvent } from '../ActionMenuEvents'
+import type { FindActiveMenuOptions } from '../ActionMenuModuleOptions'
+import type { CreateMenuOptions, CreatePerformOptions, CreateRequestOptions } from './ActionMenuServiceOptions'
+
+import { AgentConfig } from '../../../agent/AgentConfig'
+import { EventEmitter } from '../../../agent/EventEmitter'
+import { AriesFrameworkError } from '../../../error'
+import { injectable } from '../../../plugins'
+import { JsonTransformer } from '../../../utils'
+import { ActionMenuEventTypes } from '../ActionMenuEvents'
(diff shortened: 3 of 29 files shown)


packages/core/src/modules/action-menu/index.ts
```diff
@@ -0,0 +1,4 @@
+export * from './ActionMenuModule'
+export * from './handlers'
```
TimoGlastra, line 2: Do we need to export the handlers?

packages/core/src/modules/action-menu/messages/MenuMessage.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  @IsString()
+  public title!: string
+
+  @IsString()
+  public description!: string
+
+  @IsString()
+  @IsOptional()
+  public errormsg?: string
```
TimoGlastra, line 48: Maybe we can use an `Expose` here with different name and call it `errorMsg` or `errorMessage` in AFJ? Seems more aligned with the rest

packages/core/src/modules/action-menu/services/ActionMenuService.ts
```diff
@@ -0,0 +1,282 @@
...
+      role: ActionMenuRole.Requester,
+    })
+
+    if (actionMenuRecord) {
+      const previousState = actionMenuRecord.state
+      actionMenuRecord.state = ActionMenuState.AwaitingRootMenu
+      actionMenuRecord.threadId = menuRequestMessage.id
+
+      await this.actionMenuRepository.update(actionMenuRecord)
+      this.emitStateChangedEvent(actionMenuRecord, previousState)
```
TimoGlastra, line 50: What if the action menu is currently in progress? Can I just start a new one? 

I see you mentioned: 
> As this protocol currently defines a single active menu per connection, there will be at most 2 records per connection (as theoretically is possible to serve as requester and responder at the same time).

Is this a limitation imposed by the RFC?

packages/core/src/modules/action-menu/services/ActionMenuService.ts
```diff
@@ -0,0 +1,282 @@
...
+    })
+
+    if (actionMenuRecord) {
+      const previousState = actionMenuRecord.state
+      actionMenuRecord.state = ActionMenuState.PreparingRootMenu
+      actionMenuRecord.threadId = menuRequestMessage.id
+
+      await this.actionMenuRepository.update(actionMenuRecord)
+      this.emitStateChangedEvent(actionMenuRecord, previousState)
+    } else {
```
TimoGlastra, line 86: Do we need to assert state?

packages/core/tests/action-menu.test.ts
```diff
@@ -0,0 +1,119 @@
+import type { SubjectMessage } from '../../../tests/transport/SubjectInboundTransport'
```
TimoGlastra, line 1: Can we move this to the action menu module directory tests and call it `action-menu.e2e.test.ts`? I'd like to start co-locating tests (even e2e) with their module are more contained.

TimoGlastra (approved):
Nice and tidy 👍 . My preference would go to making this an optional add-on module. However I'm fine with merging it in core in 0.2.0 and then extracting it from core into a separate package (like module-tenants) in 0.3.0. What do you think?

genaris:
> Nice and tidy 👍 . My preference would go to making this an optional add-on module. However I'm fine with merging it in core in 0.2.0 and then extracting it from core into a separate package (like module-tenants) in 0.3.0. What do you think?
> 
> Edit: I see unit tests are missing, but that's probably still todo as you mention in the pr description?

Thanks! I think it would be a good idea as you say, we can merge it in 0.2.x and then move it to a separate package, as it is not a core protocol.

I'm writing the unit tests and actually found a few issues, so I'll fix them and submit the PR for review shortly (hopefully 😄).

packages/core/src/modules/action-menu/services/ActionMenuService.ts
genaris: From the RFC I understood that this situation would discard current flow and request back the root menu:
> At any time a requester can reset the protocol by requesting the root menu from a responder.

The RFC states also that:

> both Agents may support both requester and responder roles simultaneously. Such cases would result in two instances of the action-menu protocol operating in parrallel.

So that's why we are supporting one instance per role and connection. At first I thought to create an ActionMenuRecord per interaction (for instance, every time the requester agent selects an option, it creates a new record, which would be set to Done once a new menu is requested and then another record is created). But it could become a bit too complicated considering the simplicity of current specification.

packages/core/src/modules/action-menu/index.ts
genaris: Not really :smile:

packages/core/tests/action-menu.test.ts
genaris: Good idea, that will be nice to easily to separate modules from core.

genaris:
Well now I've added some error handling and tests (with our new friend SonarCloud complaining) I think it's not as nice and tidy as before, but let's see how it goes 😄 . I updated PR description with some more details about the work done and what's left.

TimoGlastra:
I'm a bit hesitant to merge this now as it will be quite a hassle to merge into 0.3.0-pre afterwards. When I get the time to immediately update the implementation to work with 0.3.0-pre I'll merge

genaris:
> I'm a bit hesitant to merge this now as it will be quite a hassle to merge into 0.3.0-pre afterwards. When I get the time to immediately update the implementation to work with 0.3.0-pre I'll merge

Do you mean to merge this module directly on 0.3.0-pre? Or do you want to first update 0.3.0-pre with current main branch status and then merge this one into main?

TimoGlastra:
> Do you mean to merge this module directly on 0.3.0-pre? Or do you want to first update 0.3.0-pre with current main branch status and then merge this one into main?

No it's probably good to have it in 0.2.x if possible, but after it has been merged in main we need to rebase the 0.3.0-pre branch and integrate the changes from main. I'll do that soon, and then we can make an additional PR to extract the action menu into a separate package.

TimoGlastra:
Can you update once more? Then I'll merge this PR

Issue #956, opened 2022-07-20
Pickup protocol returned in discover features protocol, but only supported on the client


TimoGlastra:
When using mediation in AFJ as the client and also using AFJ as the mediator without configuring the pickup strategy it will try to determine the pickup strategy using feature discovery.

Afj as a mediator will return it supports the pickup v2 protocol, but this is only partly true. AFJ supports the pickup v2 protocol as a client.

This means detecting the pickup protocol from an AFJ mediator has been broken since the PR was added for pickup v2: #711

We should add support for the `roles` property from the discover features protocol, so we can query for the specific role.

@KolbyRKunz @JamesKEbert is this something you may be willing to take a look at?

Discover features v1: https://github.com/hyperledger/aries-rfcs/blob/main/features/0031-discover-features/README.md

Issue #800, opened 2022-05-26
Split up the mediator pickup approaches into different services


TimoGlastra:
The mediation code is getting quite complex because we manage all pickup variants (Implicit, V1 and v2) in the same module/service. We should ook at splitting up the mediator service in different pickup services and let each manage their own routing lifecycle.

Issue #727, opened 2022-04-29
Add support for mediator role for pickup v2 protocol


TimoGlastra:
As part of PR #711 support for the recipient role was added. This issue tracks the progress of also adding support for the mediator role.

In addition when adding support for the mediator, we should also add e2e tests to test the whole flow of picking up messages, which can be complex when only supporting one side of the protocol.

RFC: https://github.com/hyperledger/aries-rfcs/tree/main/features/0685-pickup-v2

genaris:
Is anyone working on this at the moment? Otherwise I can pick it up (yeah I know, nice pun :smile:)

TimoGlastra:
> Is anyone working on this at the moment? Otherwise I can pick it up (yeah I know, nice pun :smile:)

I'm not aware of anyone working on this, so it's yours to take! 

If you're doing some refactoring, we also have this issue to untangle all of the routing protocols into separate services, maybe it helps to keep that in mind and make some first steps at it for the pickup v2 protocol: https://github.com/hyperledger/aries-framework-javascript/issues/800

Issue #475, opened 2021-09-30
Add support for RFC 0685: Pickup Protocol v2


TimoGlastra:
Add support for [RFC 0685: Pickup Protocol V2](https://github.com/hyperledger/aries-rfcs/pull/685)

PR #975, opened 2022-08-07
feat(routing): pickup v2 mediator role basic implementation

genaris:
Basic implementation of the missing Mediator role for Pick Up V2 protocol ([Aries RFC 0685](https://github.com/hyperledger/aries-rfcs/tree/main/features/0685-pickup-v2)). In order to make this PR smaller and non-breaking, not all features from protocol are supported, but only those used currently by `Recipient` implementation in AFJ 0.2.x. 

There are some refactoring here and there to start doing the split up suggested by #800, but attempting to maintain compatibility with current modules API.

Note that current implementation of client side assumes that the mediator will use a WebSocket connection, which is not actually mandatory. This behaviour is not yet updated, but might require to update `WsInboundTransporter` to manage long-term WebSockets connections (e.g. use a heartbeat interval mechanism) in order to be really useful.

Solves #956 and hopefully #475 and #727. 


packages/core/src/modules/routing/protocol/pickup/v1/MessagePickupService.ts
```diff
@@ -0,0 +1,64 @@
...
+    eventEmitter: EventEmitter
+  ) {
+    this.messageRepository = messageRepository
+    this.dispatcher = dispatcher
+    this.eventEmitter = eventEmitter
+
+    this.registerHandlers()
+  }
+
+  public async batch(messageContext: InboundMessageContext<BatchPickupMessage>) {
```
TimoGlastra, line 33:
```suggestion
  public async createBatch(messageContext: InboundMessageContext<BatchPickupMessage>) {
```

packages/core/src/modules/routing/protocol/pickup/v1/MessagePickupService.ts
```diff
@@ -0,0 +1,64 @@
...
+import { createOutboundMessage } from '../../../../../agent/helpers'
+import { InjectionSymbols } from '../../../../../constants'
+import { inject, injectable } from '../../../../../plugins'
+import { MessageRepository } from '../../../../../storage/MessageRepository'
+
+import { BatchHandler, BatchPickupHandler } from './handlers'
+import { BatchMessage, BatchMessageMessage } from './messages'
+
+@injectable()
+export class MessagePickupService {
```
TimoGlastra, line 16: Should this have a prefix to avoid clashes? We've taken this approach for issue credential / present proof. This would mean we also need to update the messages, handlers, etc...

```suggestion
export class V1MessagePickupService {
```

packages/core/src/storage/MessageRepository.ts
```diff
@@ -2,5 +2,6 @@ import type { EncryptedMessage } from '../types'
 
 export interface MessageRepository {
   takeFromQueue(connectionId: string, limit?: number): EncryptedMessage[] | Promise<EncryptedMessage[]>
+  available(connectionId: string): number | Promise<number>
```
TimoGlastra, line 5:
```suggestion
  getAvailableMessages(connectionId: string): number | Promise<number>
```

packages/core/src/modules/routing/protocol/pickup/v1/handlers/BatchPickupHandler.ts
```diff
@@ -1,7 +1,7 @@
-import type { Handler, HandlerInboundMessage } from '../../../agent/Handler'
-import type { MessagePickupService } from '../services'
+import type { MessagePickupService } from '..'
```
TimoGlastra, line 1: Shouldn't import from only dots

packages/core/src/modules/routing/protocol/pickup/v2/V2MessagePickupService.ts
```diff
@@ -0,0 +1,101 @@
...
+    // Assert ready connection
+    const connection = messageContext.assertReadyConnection()
+
+    const { message } = messageContext
+
+    // TODO: Add Queued Message ID
+    await this.messageRepository.takeFromQueue(
+      connection.id,
+      message.messageIdList ? message.messageIdList.length : undefined
+    )
```
TimoGlastra, line 91: We shouldn't take from the queue in this case I think? 

I'm assuming this is what you're talking about in the pr description. We should somehow be able to get messages from the queue (without removing them) and also be able to remove messages from the queue. Currently the acknowledgement doesn't matter and it will always just remove them right?

TimoGlastra (commented):
Nice work @genaris

packages/core/src/modules/routing/protocol/pickup/v1/MessagePickupService.ts
genaris: Actually, this is mostly a relocation of the original `MessagePickUpService`, as I wanted to keep current services interfaces as most as possible in this PR.  But if this doesn't count as a 'breaking change', for sure I'll be happy to change it as suggested!

packages/core/src/modules/routing/protocol/pickup/v1/MessagePickupService.ts
genaris: I think it should be prefixed. And in this case, protocol messages are named differently so I guess there will be no need to prefix all classes.

packages/core/src/modules/routing/protocol/pickup/v2/V2MessagePickupService.ts
genaris: Exactly. That's not yet handled properly. I was thinking on adding an optional parameter to `takeFromQueue` like `keepMessages` that defaults to current behaviour, so it does not break current V1 handler and limits changes to minimum for 0.2.x.

But if we are free to change `MessageRepository`/`InMemoryMessageRepository` and V1 service/handlers a bit (as long as `MediationModule` and `MediationRecipient` API does not break), probably it would be better to have a method to slice and another to splice the internal messages array.

packages/core/src/storage/MessageRepository.ts
TimoGlastra: Actually, getAvailableMessageCount or getMessageCount is probably even better

genaris:
TBH I'm not so happy with the resulting support of Pick Up V2 protocol, as the limitations to avoid breaking changes were more significant than I thought. 

Still, I think it will work fine for the moment for E2E use cases in a pure-AFJ environment, much in a similar way than using Pick Up V1, as client side is not using Live Mode and also handles the pick-up automatically (i.e. sends Delivery Request message as soon as it receives a count > 0 in Status message).

So with these modifications we are covering these scenarios: 

- Edge Agent that wants to retrieve queued messages from another one (e.g. a connection between a mobile and a cloud agent without mediation)
- Mobile Agent that retrieves all messages (regardless their recipient key) from its mediator. This will need some external help though, as `mediationRecipient` module does not send Status Request periodically (it seems that it relies on a sort of mediator's _implicit live mode_ after sending a Trust Ping):
  - Framework controller can manually execute `mediationRecipient.pickupMessages(connectionId, MediatorPickupStrategy.PickUpV2)` if they wish
  - Mediator uses a customized WsInboundTransporter that keeps long-lived sessions as ACA-Py does

I'm not sure if we can add anything else without introducing breaking changes in both client/server sides. However in another PR (to merge in 0.3.x branch) we can immediately think about all missing features, such as:

- Filtering messages by recipient key
- Clearly identifying messages in delivery and its reception acknowledge to safely remove them from the queue
- Supporting Live Mode
- Allowing connection to mediator using "Explicit Pick Up V2" (i.e. periodic Status Requests)
- Allowing Pick Up V2 with other transports than WebSocket (as it is actually only required for Live Mode)
- Refactoring Mediation module, which probably would result in decoupling Pick Up from Mediation protocols, as it can be used to query any queuing-capable agent (not only the one that forwards messages to another one)

Please let me know what you think so we can find the best approach to go-on with this.

packages/core/src/modules/routing/protocol/pickup/v1/MessagePickupService.ts
TimoGlastra: Well, moving it to another service already is 'breaking' if we see the service as public API, or is this just a moving of the file? It's not really clear what is public / private API. I've added it to the agenda for the AFJ WG call.

packages/core/src/modules/routing/protocol/pickup/v1/MessagePickupService.ts
TimoGlastra: Hmm probably it's good to start doing it everywhere as it is leading to issues over time? Not sure

packages/core/src/modules/routing/protocol/pickup/v2/V2MessagePickupService.ts
TimoGlastra: The message repo can already be registered with a custom version at the moment (I'm using it in a project) so we should be careful with changing it. 

I think for now this is fine and then we can go wild in the 0.3.0 branch. WDYT?

TimoGlastra:
That sounds great @genaris. I think this is already a huge improvement, and then making the other changes in 0.3.0-pre will be great!

packages/core/src/modules/routing/protocol/pickup/v1/MessagePickupService.ts
genaris: Yes, actually I meant that from outside of the framework you'd see the same MessagePickupService (i.e. when importing it from @aries-framework/core).

I agree that we should better clarify the rules for breaking changes. I don't know if I'll be able to attend today's call but my 2 cents from practical perspective is that we should care mostly on Module API and also items that are explicitly intended to be injected, such as Wallet, MessageRepository and the inbound/outbound transportes. It's desirable to do the same for services (as some custom plug-ins can make use of them), but IMHO it's acceptable if it's properly documented in the release and  it is possible to workaround any change without impacting the behaviour. For instance if we add V1 prefix to the service, compilation of the plugin using it will fail but it will not have an unpredictable behaviour.

packages/core/src/modules/routing/protocol/pickup/v2/V2MessagePickupService.ts
genaris: Absolutely!

TimoGlastra:
As we're going to make the other changes in 0.3.0-pre this PR is ready to merge then right?

genaris:
> As we're going to make the other changes in 0.3.0-pre this PR is ready to merge then right?

Yeah, I'll do the message classes renaming in that further PR.

TimoGlastra (approved):

Issue #899, opened 2022-06-22
Mediation issues with multiple agents attempting to connect concurrently


niallshaw-absa:
We are encountering multiple issues (with not an individual identifiable cause) when attempting to initiate mediation with a single mediator agent from up to 30 client agents simultaneously.
Mediator agent is running mysql, via vdr-tools (a fork of indy-sdk) binary.

Some of the errors that have been occurring include the following:

- SIGSEV (Segmentation fault)
- Heap corruption
- `DEBUG: Request was aborted due to timeout. Not throwing error due to return routing on sent message`

We believe one of the causes of this issue is that mysql enables queries to be executed considerably more concurrently than sqlite, and therefore when performing checks like below in parallel, it attempts to create the singleton record numerous times - resulting in a failures.
```
if (!this._mediatorRoutingRecord) {
            this.agentConfig.logger.debug('Mediator routing record not loaded yet, retrieving from storage');
            let routingRecord = await this.mediatorRoutingRepository.findById(this.mediatorRoutingRepository.MEDIATOR_ROUTING_RECORD_ID);
            // If we don't have a routing record yet, create it
            if (!routingRecord) {
                this.agentConfig.logger.debug('Mediator routing record does not exist yet, creating routing keys and record');
                const { verkey } = await this.wallet.createDid();
                routingRecord = new repository_1.MediatorRoutingRecord({
                    id: this.mediatorRoutingRepository.MEDIATOR_ROUTING_RECORD_ID,
                    routingKeys: [verkey],
                });
                await this.mediatorRoutingRepository.save(routingRecord);
            }
            this._mediatorRoutingRecord = routingRecord;
        }
```

This is one of these instances that I managed to identify, and I have created a temporary fix - by delaying all other calls of the function by 20ms, therefore allowing for the first saving query to finish execution, see below.

```
const thisQuery = ++this._totalWaitingQueries
        this.agentConfig.logger.debug('Retrieving mediator routing keys');
        // If the routing record is not loaded yet, retrieve it from storage
        if (thisQuery!==1) {
            await new Promise((resolve) => setTimeout(resolve, 20));
        }
```

However, this temporary fix is not optimal, as we should not add an arbitrary delay to all further operations.

TimoGlastra:
Thanks for opening this issue @niallshaw-absa! I think there's still lots to improve in running AFJ server side. 

I'll think about some things we can do to improve this. Do you have any suggestions on how we can best solve this?

[...]

PR #980, opened 2022-08-10
feat: always initialize mediator

niallshaw-absa:
Add mediationRole property to AgentConfig to ensure MediatorModule is initialized correctly. This prevents concurrency issues when multiple attempts to create a mediatorRoutingRecord fail.

Signed-off-by: Niall Shaw <niall.shaw@absa.africa>

### DIFF
--- a/packages/core/src/agent/Agent.ts
+++ b/packages/core/src/agent/Agent.ts
@@ -29,6 +29,7 @@ import { ProofsModule } from '../modules/proofs/ProofsModule'
 import { QuestionAnswerModule } from '../modules/question-answer/QuestionAnswerModule'
 import { MediatorModule } from '../modules/routing/MediatorModule'
 import { RecipientModule } from '../modules/routing/RecipientModule'
+import { MediationRole } from '../modules/routing/models'
 import { RoutingService } from '../modules/routing/services/RoutingService'
 import { DependencyManager } from '../plugins'
 import { StorageUpdateService, DidCommMessageRepository, StorageVersionRepository } from '../storage'
@@ -164,7 +165,8 @@ export class Agent {
   }
 
   public async initialize() {
-    const { connectToIndyLedgersOnStartup, publicDidSeed, walletConfig, mediatorConnectionsInvite } = this.agentConfig
+    const { connectToIndyLedgersOnStartup, publicDidSeed, walletConfig, mediatorConnectionsInvite, mediationRole } =
+      this.agentConfig
 
     if (this._isInitialized) {
       throw new AriesFrameworkError(
@@ -233,8 +235,8 @@ export class Agent {
       const mediationConnection = await this.getMediationConnection(mediatorConnectionsInvite)
       await this.mediationRecipient.provision(mediationConnection)
     }
-
-    await this.mediationRecipient.initialize()
+    if (mediationRole === MediationRole.Mediator) await this.mediator.initialize()
+    if (mediationRole === MediationRole.Recipient) await this.mediationRecipient.initialize()
 
     this._isInitialized = true
   }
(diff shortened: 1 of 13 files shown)


niallshaw-absa:
Aims to fix #899

TimoGlastra:
Technically an agent can be both a mediator and a recipient, so I'm not to keen on adding the role property to the config. I see two options, but there's probably other good options:
- Always initialize it. It adds some overhead, but it's not that much
- Add an option e.g. `initializeMediatorKeysOnStartup` that will init the mediator keys.

In addition, I'm not sure if it should be impossible to use mediation without the mediation role . E.g. with the ledger we have connectToLedgersOnStartup. If you have it enabled it will connect on startup, but otherwise when first requested. Is that something we could also do here?

Another issue is that this won't work for high availability deployments (multiple instances connecting to the same database). There's more places where this will lead to issues, so we can probably ignore this for now.

samples/mediator.ts
```diff
@@ -37,7 +37,7 @@ const port = process.env.AGENT_PORT ? Number(process.env.AGENT_PORT) : 3001
 const app = express()
 const socketServer = new Server({ noServer: true })
 
-const endpoints = process.env.AGENT_ENDPOINTS?.split(',') ?? [`http://localhost:${port}`, `ws://localhost:${port}`]
+const endpoints = process.env.AGENT_ENDPOINTS?.split(',') ?? [`https://ebff-31-30-172-153.ngrok.io`]
```
jakubkoci, line 40: I'm not sure if we want to commit this change.

packages/core/src/agent/Agent.ts
```diff
@@ -233,8 +235,8 @@ export class Agent {
       const mediationConnection = await this.getMediationConnection(mediatorConnectionsInvite)
       await this.mediationRecipient.provision(mediationConnection)
     }
-
-    await this.mediationRecipient.initialize()
+    if (mediationRole === MediationRole.Mediator) await this.mediator.initialize()
+    if (mediationRole === MediationRole.Recipient) await this.mediationRecipient.initialize()
```
jakubkoci, line 239: Should not we also check if `mediationRole === MediationRole.Recipient` before we try to use `mediatorConnectionsInvite`?

jakubkoci:
> Technically an agent can be both a mediator and a recipient

Then it might be an array of roles. I usually try to avoid these flag config attributes, but if the array is a problem I'm also ok with something like `initializeMediatorRoutingOnStartup`. The word `Keys` seem to be too low-level detail. 

> Another issue is that this won't work for high availability deployments (multiple instances connecting to the same database). There's more places where this will lead to issues, so we can probably ignore this for now.

Yes, without this change, it doesn't work even with a few edge agents trying to connect at the same time, so I would rather do this now and fix "high availability deployments" later.

genaris:
Is there any 

> Technically an agent can be both a mediator and a recipient, so I'm not to keen on adding the role property to the config. I see two options, but there's probably other good options:
> 
> * Always initialize it. It adds some overhead, but it's not that much
> * Add an option e.g. `initializeMediatorKeysOnStartup` that will init the mediator keys.
> 
> In addition, I'm not sure if it should be impossible to use mediation without the mediation role . E.g. with the ledger we have connectToLedgersOnStartup. If you have it enabled it will connect on startup, but otherwise when first requested. Is that something we could also do here?
> 
> Another issue is that this won't work for high availability deployments (multiple instances connecting to the same database). There's more places where this will lead to issues, so we can probably ignore this for now.

Probably I'm missing something, but I don't see a major problem on always initializing mediator module, as it will actually initialize the routing key only once and then just query it and store it in the internal variable.

In such case, no new config parameters would be needed. What do you think?

TimoGlastra:
Agree with @genaris here. I think avoiding the config option is worth the little bit of extra overhead.

niallshaw-absa:
Moved to #985

PR #985, opened 2022-08-18
feat: always initialize mediator

niallshaw-absa:
Always initialize the mediator during `agent.initialize()`, to prevent attempts to initialize twice due to parallelism


packages/core/src/modules/routing/services/MediatorService.ts
```diff
@@ -49,34 +49,37 @@ export class MediatorService {
...
+  }
+
+  private getRoutingKeys() {
+    if (this._mediatorRoutingRecord) {
+      // Return the routing keys
+      this.agentConfig.logger.debug(`Returning mediator routing keys ${this._mediatorRoutingRecord.routingKeys}`)
+      return this._mediatorRoutingRecord.routingKeys
+    }
+    throw new AriesFrameworkError(
+      `Mediation service has not been initialized. Agent must include the config {mediationRole: MediationRole.Mediator}`
```
genaris, line 81: I think the last sentence from this error comes from previous PR, doesn't it?

genaris (changes requested):
Thanks @niallshaw-absa! I've seen that some tests failed. Are they working locally for you? Maybe it'll be needed to update snapshots as now the mediator routing record is created at the start.

packages/core/src/modules/routing/services/MediatorService.ts
niallshaw-absa: Good catch

niallshaw-absa:
> Thanks @niallshaw-absa! I've seen that some tests failed. Are they working locally for you? Maybe it'll be needed to update snapshots as now the mediator routing record is created at the start.

@genaris - tests are failing for me locally also, if possible can you show me how to update the snapshots? Thanks!

genaris:
> > Thanks @niallshaw-absa! I've seen that some tests failed. Are they working locally for you? Maybe it'll be needed to update snapshots as now the mediator routing record is created at the start.
> 
> @genaris - tests are failing for me locally also, if possible can you show me how to update the snapshots? Thanks!

To update snapshots you need to add -u parameter to yarn test. However it was not enough, because this new behaviour implied the generation of a different record each time (as routing key changed), so I've added some mocks on DID generation for that particular test suite.

genaris (approved):

PR #989, opened 2022-08-23
feat(routing): manual mediator pickup lifecycle management

### LINKED ISSUE #753: Add ability to manage mediator pickup lifecycle manually
Currently you can initiate message pickup but it's not possible to stop the message pickup. When building a mobile wallet you often need more control over when the mediator pickup is active:

- when the device loses internet connection you may want to halt the pickup (other it will try to create a lot of ws connections but never succeed)
- when the device enters background mode you may want to halt the pickup
- when the device enters foreground mode again you may want to resume the pickup

genaris:
Small update on Mediation Recipient module to allow stopping message pickup loop, and then restarting it anytime by calling `initiateMessagePickup()` manually.

An optional parameter has been added to **initiateMessagePickup** method in order to let start the agent with a `None` messagePickupStrategy (so it will not attempt to connect to mediator at the beginning) and be able to start the pickup afterwards (using any strategy we want).

So it can be used in the following way:

1) Start agent normally (with whatever Mediation Strategy you already have). It will initiate Pickup loop (V1, V2 or Implicit)
2) Call stopMessagePickup() to pause this process (e.g. when agent is offline)
3) Call initiateMessagePickup() to restart the process (e.g. agent goes online)

In case that default pickup strategy is set to None, of course, initiateMessagePickup should be called after 1 using parameters defaultMediator and the desired strategy (and the same in 3)

And I'm not completely sure, but it might also help with some failures in test suites when shutting down agents.

Resolves #753


packages/core/src/modules/routing/RecipientModule.ts
```diff
@@ -47,6 +47,9 @@ export class RecipientModule {
   private mediationRepository: MediationRepository
   private routingService: RoutingService
 
+  // $stop is used for stop message pickup signal
+  private readonly stop$ = new Subject<boolean>()
```
TimoGlastra, line 51: can we give this a more specific name? I'm already not a huge fan I named it stop$ for the generic agent shutdown subject

packages/core/src/modules/routing/__tests__/pickup.test.ts
```diff
@@ -7,21 +7,17 @@ import { SubjectInboundTransport } from '../../../../../../tests/transport/Subje
...
 
-const logger = new ConsoleLogger(LogLevel.info)
-const recipientConfig = getBaseConfig('Mediation: Recipient', {
+const recipientConfig = getBaseConfig('Pickup: Recipient', {
   autoAcceptConnections: true,
   indyLedgers: [],
-  logger,
 })
-const mediatorConfig = getBaseConfig('Mediation: Mediator', {
+const mediatorConfig = getBaseConfig('Pickup: Mediator', {
```
TimoGlastra, line 17: Also ran into issues with this 😆

packages/core/src/transport/WsOutboundTransport.ts
```diff
@@ -35,10 +35,13 @@ export class WsOutboundTransport implements OutboundTransport {
...
-    this.transportTable.forEach((socket) => {
-      socket.removeEventListener('message', this.handleMessageEvent)
-      socket.close()
-    })
+
+    await Promise.all(
+      Array.from(this.transportTable.values()).map((socket) => {
+        socket.removeEventListener('message', this.handleMessageEvent)
+        socket.close()
+      })
```
TimoGlastra, line 43: I don't think there are any promises involved here?

packages/core/tests/connectionless-proofs.test.ts
```diff
@@ -342,11 +341,7 @@ describe('Present Proof', () => {
...
 
-    // We want to stop the mediator polling before the agent is shutdown.
-    // FIXME: add a way to stop mediator polling from the public api, and make sure this is
-    // being handled in the agent shutdown so we don't get any errors with wallets being closed.
-    faberAgent.config.stop$.next(true)
-    aliceAgent.config.stop$.next(true)
-    await sleep(2000)
+    await aliceAgent.mediationRecipient.stopMessagePickup()
+    await faberAgent.mediationRecipient.stopMessagePickup()
   })
```
TimoGlastra, line 346: much better API

TimoGlastra (commented):
This is a great improvement. One thing I'd like to see but is probably much more work is that you can stopMessagePickup for a specific mediator. that is probably something we can add later on.

another thing we should do as a change to 0.3.0 (feel free to open a pr 😄) is to make the public api take mediator id's instead of the records.

packages/core/src/modules/routing/RecipientModule.ts
genaris: I've set it to stopMessagePickup$ to be more specific. Didn't want it to collide with stopMessagePickup() method but, as it is a private field, it should not be confusing I think.

packages/core/src/transport/WsOutboundTransport.ts
genaris: True. Was part of an experiment and left it by mistake 🙈

genaris:
> This is a great improvement. One thing I'd like to see but is probably much more work is that you can stopMessagePickup for a specific mediator. that is probably something we can add later on.
> 
> another thing we should do as a change to 0.3.0 (feel free to open a pr 😄) is to make the public api take mediator id's instead of the records.

Thanks! I've been thinking about the management of individual message pickup loops but, as you said, requires some complexity and I think it will be better to develop it in a refactored MediationRecipient / MessagePickup module for 0.3.0.

I'll start working on these refactoring ideas (including the usage of mediatorId / connectionId) so we can discuss about it shortly.

TimoGlastra (approved):

TimoGlastra:
Awesome, merged! Thanks for all the contributions lately :)

Issue #495, opened 2021-10-20
Add support for RFC 0557: Discover Features v2


TimoGlastra:
https://github.com/hyperledger/aries-rfcs/tree/b3a3942ef052039e73cd23d847f42947f8287da2/features/0557-discover-features-v2

genaris:
It looks like this is will have some priority in the short term, as it is required for AIP 2.0 compatibility and also it becomes more important as framework is growing (as shown in #956). Is there any ongoing work on this issue?

It's not so trivial to implement, as currently Discover Features module is using Agent Dispatcher to tell which protocols it supports and this is limited to message types. But probably a sort of _ProtocolRegistry_ can be created in order to let each protocol module to register its capabilities (in terms of protocols, roles and goal codes), and then query it by the Discover Features module, and maybe this wouldn't introduce breaking changes and even be incorporated in 0.2.x releases.

TimoGlastra:
I don't think anyone is working on this, so feel free to pick this up! :)

Yeah I think we should have some sort of protocol registry. We currently register the handlers on the dispatcher, but if we extract that to a separate protocol registry where you can register handlers (for message types) but also decorators, roles for a protocol and other items so it can be extended beyond the current known discovery types. 

I was first thinking of extending the handlers, but a handler is just one message, so not sure we should add items to it that are on the protocol lever instead of the message level

genaris:
I kept thinking about this these days and my current idea is to define a 'FeatureRegistry' with the capabilities of registering items (features) you mentioned, that will be queried by DiscoveryFeatures module when some agent.

A good entry point for that feature registration can be the Module Registration API that will be fully available for 0.3.0, so I was thinking on going straight there. However, we were discussing briefly about this in today's call with @JamesKEbert and, as there is interest on supporting this soon, we can try to find a way of introducing in 0.2.x first.

So what I'm wondering now is that each module can register its features in `register`method, which would take not only the DependencyManager as parameter but also this new FeatureRegistry. Something like: 

```ts
public static register(featureRegistry: FeatureRegistry, dependencyManager: DependencyManager) {
    // Api
    dependencyManager.registerContextScoped(Module)

    // Services
    dependencyManager.registerSingleton(Service)

    // Features
    featureRegistry.registerFeatures([protocols, roles, goal-codes,,etc.])
  }
}
```

This way I guess it will not break anything (as this registration API is internal/experimental) and would give `register` method a wider scope than just registering dependencies.

[...]

PR #991, opened 2022-08-25
feat!: Discover Features V2 

genaris:
Discover Features V1 refactor + Initial API implementation of Discover Features V2 with a `FeatureRegistry` where arbitrary features can be registered and queried by modules, and also externally (i.e. to let any controller app setting a custom feature that is willing to disclose).

Discover Features module now has a configuration setting to define whether queries should be automatically responded, or leave that decision (and feature set to disclose) to business logic. In either case, events are emitted when receiving queries and disclosures.

Module API is very simple, as we have only two methods whose goal is self-explanatory:

* **queryFeatures**: to query features to a connection
* **discloseFeatures**: to send a disclosure of features to a connection (either proactively or as a response of a query)

Depending on options, queries can use Discover Features V1 or V2 (maybe we could add an auto-selection depending on the complexity of the query) and can be done either _asynchronously_ or _synchronously_, meaning that the promise will resolve when the response is received (or a timeout). This is not a common pattern in the rest of the framework, but in this case could be useful to quickly ask for features like it happens in `RecipientModule` to determine mediator pickup strategy.

**Feature Registry** has been exposed in `agent.features` key, in order to let the framework controller to add their own custom features (such as Goal Codes). Probably we should add some checks to prevent adding certain feature types from outside of the framework in order to provide consistent features in disclosures (e.g. we should not allow to register a protocol if there is no module that handles it).


packages/core/src/modules/discover-features/FeatureRegistry.ts
```diff
@@ -0,0 +1,30 @@
+import type { Feature, FeatureQuery } from './models'
+
+import { injectable } from '../../plugins'
+
+@injectable()
+class FeatureRegistry {
+  private features: Feature[] = []
+
+  public register<F extends Feature = Feature>(feature: F) {
```
blu3beri, line 9:
```suggestion
  public register<F extends Feature>(feature: F) {
```

This does not seem required

packages/core/src/modules/basic-messages/BasicMessagesModule.ts
```diff
@@ -46,6 +52,14 @@ export class BasicMessagesModule {
     dispatcher.registerHandler(new BasicMessageHandler(this.basicMessageService))
   }
 
+  private registerFeatures(featureRegistry: FeatureRegistry) {
+    featureRegistry.register<Protocol>(
```
blu3beri, line 56:
```suggestion
    featureRegistry.register(
```

This can be inferred by the TSEngine so we do not have to supply it all the time.

packages/core/src/modules/credentials/CredentialsModule.ts
```diff
@@ -625,6 +630,33 @@ export class CredentialsModule<
     return this.getService(credentialExchangeRecord.protocolVersion)
   }
 
+  private registerFeatures(featureRegistry: FeatureRegistry) {
+    featureRegistry.register<Protocol>(
```
blu3beri, line 634: Maybe a register multiple would be nice? Or register would always take an array? Not sure.

packages/core/src/modules/discover-features/DiscoverFeaturesModule.ts
```diff
@@ -82,18 +86,34 @@ export class DiscoverFeaturesModule {
...
     const connection = await this.connectionService.getById(connectionId)
 
-    const queryMessage = await this.discoverFeaturesService.createQuery(options)
+    const queryMessage =
+      typeof options.query === 'string'
+        ? await this.discoverFeaturesService.createQuery({
+            query: options.query,
+            comment: options.comment,
+          })
+        : await this.discoverFeaturesV2Service.createQuery({ queries: options.query })
```
blu3beri, line 98:
```suggestion
        : await this.discoverFeaturesV2Service.createQuery({ query: options.query })
```

Or is there a reason for it being plural?

packages/core/src/modules/discover-features/FeatureRegistry.ts
```diff
@@ -0,0 +1,30 @@
...
+    this.features.push(feature)
+  }
+
+  public query(queries: FeatureQuery[]) {
```
blu3beri, line 13: This might need some small documentation about the if statement and how the match * works. would save a lot of time :) when trying to understand the code.

packages/core/src/modules/discover-features/protocol/v2/V2DiscoverFeaturesService.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    this.registerHandlers(dispatcher)
+  }
+
+  private registerHandlers(dispatcher: Dispatcher) {
+    dispatcher.registerHandler(new V2DisclosuresMessageHandler())
+    dispatcher.registerHandler(new V2QueriesMessageHandler(this))
+  }
+
+  public async createQuery(options: { queries: FeatureQueryOptions[] }) {
```
blu3beri, line 25: Ah I see it should be plural here. Maybe we should call this createQueries or something? I am not too sure which one it has to be, but can we make them consistent?

packages/core/src/modules/question-answer/QuestionAnswerModule.ts
```diff
@@ -97,6 +102,15 @@ export class QuestionAnswerModule {
     dispatcher.registerHandler(new AnswerMessageHandler(this.questionAnswerService))
   }
 
+  private registerFeatures(featureRegistry: FeatureRegistry) {
```
blu3beri, line 105: More a note for @TimoGlastra but we should probably add this method to the `BaseModule` or w/e it is called. Makes implementing a custom module with feature discovery a bit easier.

blu3beri (commented):
Nice work @genaris! I had some very minor notes, not really about the specification but more just code cleanup. 

I remember from the last WG call that there would be some breaking changes with this module. Will these be introduced in a separate PR?

packages/core/src/modules/discover-features/FeatureRegistry.ts
genaris: True. At first I was playing with interfaces because these "Features" are quite arbitrary and generic, but at the moment I'm using classes. Probably not the best approach though, because currently we need to provide instances and would like to use options instead.

genaris:
> Nice work @genaris! I had some very minor notes, not really about the specification but more just code cleanup.
> 
> I remember from the last WG call that there would be some breaking changes with this module. Will these be introduced in a separate PR?

Thanks for the review @blu3beri! Actually this is an attempt to introduce Discover Features V2 before 0.3.0, present some concepts such as the Feature Registry and, as you mention in one of your comments, figure out how to improve module registration API (and module architecture in general) to leverage registration of all features they add to the framework. I would say that it's currently more a discussion than a PR :smile:.

During the coding and research over Aries RFCs/DIDComm spec I found these 'features' quite arbitrary and generic, so I'm trying to support the mandatory ones (with some nice typing) while still allowing to define other ones that any custom module may want to add. For the moment I'm registering only protocols but there is room for others such as goal codes, DIDComm version and AIP.

Resulting Discover Features module API is a bit confusing because V1 uses a single query while V2 allows multiple queries. Not simple to improve without introducing breaking changes in the API, but for sure in 0.3.0 will be handled better. Anyway I'll keep looking at it and hearing suggestions :thinking:

packages/core/src/modules/credentials/CredentialsModule.ts
genaris: Yeah I think it would be useful to support both single and array, as in other parts of the framework.

packages/core/src/modules/discover-features/DiscoverFeaturesModule.ts
genaris: Discover Features V1 has a "query" message that allows sending a single query, while V2 introduces "queries" allowing multiple queries. 

In this case I think it's more a copy-paste issue, as the method for `V2DiscoverFeaturesService` should be called `createQueries`.

packages/core/src/modules/discover-features/FeatureRegistry.ts
genaris: You are right. At least we need to add the reference to the RFC where these wildcards are introduced (https://github.com/hyperledger/aries-rfcs/tree/b3a3942ef052039e73cd23d847f42947f8287da2/features/0557-discover-features-v2#queries-message-type)

packages/core/src/modules/question-answer/QuestionAnswerModule.ts
genaris: Yes! I did some tries adding a static method for feature registration (similar to `register(dependencyManager)`) but was not convinced yet.

genaris: [Here](https://github.com/hyperledger/aries-framework-javascript/issues/495#issuecomment-1220126118) some discussion about it.

TimoGlastra: Yes agreed. However we should probably only add it in 0.3.0 with the custom modules

packages/core/src/agent/__tests__/Agent.test.ts
```diff
@@ -195,7 +196,35 @@ describe('Agent', () => {
...
     })
   })
+
+  it('all core features are properly registered', () => {
+    const agent = new Agent(config, dependencies)
+    const registry = agent.dependencyManager.resolve(FeatureRegistry)
+
+    const protocols = registry.query({ featureType: 'protocol', match: '*' }).map((p) => p.id)
+
+    expect(protocols).toEqual(
```
TimoGlastra, line 210: Can we maybe add a length check here? That way if new protocols are added the test will fail, forcing you to update it with the new protocol uri.

packages/core/src/modules/basic-messages/BasicMessagesModule.ts
```diff
@@ -46,6 +51,14 @@ export class BasicMessagesModule {
...
   }
 
+  private registerFeatures(featureRegistry: FeatureRegistry) {
+    featureRegistry.register(
+      new Protocol({
+        id: 'https://didcomm.org/basicmessage/1.0',
+        roles: [BasicMessageRole.Sender, BasicMessageRole.Receiver],
+      })
+    )
+  }
```
TimoGlastra, line 61: Nice, I really like this! 

The one downside is that you now have to manually register the protocol for discovery. Or can we still leverage the registered message types for feature discover and this allows to add extra metadata to it?

packages/core/src/modules/connections/ConnectionsModule.ts
```diff
@@ -67,6 +69,8 @@ export class ConnectionsModule {
     this.messageSender = messageSender
     this.didResolverService = didResolverService
     this.registerHandlers(dispatcher)
+
+    this.registerFeatures(featureRegistry)
```
TimoGlastra, line 73: What if we call this from within the register method of the module? That way it is more aligned with modules as we'll have them in 0.3.0. In 0.3.0 I would also like to move handler registration to the module

packages/core/src/modules/discover-features/DiscoverFeaturesModule.ts
```diff
@@ -82,18 +94,44 @@ export class DiscoverFeaturesModule {
...
     const connection = await this.connectionService.getById(connectionId)
 
-    const queryMessage = await this.discoverFeaturesService.createQuery(options)
+    const queryMessage =
+      typeof options.query === 'string'
+        ? await this.discoverFeaturesService.createQuery({
+            query: options.query,
+            comment: options.comment,
+          })
+        : await this.discoverFeaturesV2Service.createQueries({ queries: options.query })
```
TimoGlastra, line 118: Not too sure on swithcing the protocol version based on the query format.

What if we model it more like we do with the credential/proofs modules where the input is the same but you pass the protocol version? This will also help if a v3 is added that is the same but just uses didcomm v2. If you pass a query that is too advanced for v1 it will throw an error (like you can't issue jsonld with v1 issue credential).

Or we trigger the options type based on the protocol version. So it is still a different query type, but you MUST provide the protocolVersion

packages/core/src/modules/discover-features/FeatureRegistry.ts
```diff
@@ -0,0 +1,52 @@
...
+   * Register a single or set of Features on the registry
+   *
+   * @param features set of {Feature} objects
+   */
+  public register(...features: Feature[]) {
+    for (const feature of features) {
+      const index = this.features.findIndex((item) => item.type === feature.type && item.id === feature.id)
+
+      if (index > -1) {
+        this.features[index] = this.features[index].combine(feature)
```
TimoGlastra, line 19: I like the combine approach 👍 . Maybe we can add an optional override feature later that will not combine but override

packages/core/src/modules/discover-features/FeatureRegistry.ts
```diff
@@ -0,0 +1,52 @@
...
+   * @see https://github.com/hyperledger/aries-rfcs/blob/560ffd23361f16a01e34ccb7dcc908ec28c5ddb1/features/0557-discover-features-v2/README.md
+   *
+   * @param queries set of {FeatureQuery} objects to query features
+   * @returns array containing all matching features (can be empty)
+   */
+  public query(...queries: FeatureQuery[]) {
+    const output = []
+    for (const query of queries) {
+      const items = this.features.filter((item) => item.type === query.featureType)
+      if (query.match === '*') {
```
TimoGlastra, line 39: maybe some inline comments here would help to understand what the different paths mean?

packages/core/src/agent/__tests__/Agent.test.ts
genaris: Good idea. This test will force us to always double-check the modules we register.

packages/core/src/modules/basic-messages/BasicMessagesModule.ts
genaris: I was thinking about when adding the 'combine' method, so we can call FeatureRegistry from Dispatcher and make this registration "automatically". A problem I see with this approach (which is current AFJ approach of disclosing features) is that this will make the other agent think we fully support a protocol even if we just implement a single message from it (something related to #956). So in this sense I feel it will be more accurate if the module registers and states "I fully support this protocol, with these roles". Of course it can be wrong, but at least it will be a human interpretation error. 

That being said, probably it will be better to add this Dispatcher approach for current 0.2.x (will be useful also for Discover Features V1) and move exclusively to a module method in 0.3.x as mentioned in other comment. What do you think?

packages/core/src/modules/connections/ConnectionsModule.ts
genaris: Absolutely. I think it should be there in order to enforce feature registration when defining a module (and also the handlers, as you mentioned). As I said in a previous comment, I've done some tries by using the register method (had to add featureRegistry as parameter) but didn´t like so much. Maybe we can improve this module API to add separate methods such as:

- registerDependencies(dependencyManager)
- registerFeatures(featureManager)
- registerHandlers(dispatcher)

packages/core/src/modules/discover-features/DiscoverFeaturesModule.ts
genaris: I do also think that's the best way to proceed. However, I don't want to break current API yet so I guess I have to get some classes of TypeScript Magic to achieve that :smile:  Maybe we can add an optional parameter with protocol version that defaults to V1 and switches the expected parameters, which is more or less what you suggest if I'm not wrong.

genaris:
As discussed in today's WG, it will be better to focus on next 0.3.0 release so I'll move the base branch of this PR to 0.3.0-pre. This will help also to not need to worry about breaking changes that will be not so easy to handle with current API.

Some features that will be added to this module:
- Emit events when receiving queries and disclosures, so we won't need to actively wait for them or be online to respond them
- Add setting to define whether queries should be automatically responded or leave that decision (and set of features to disclose) to business logic

packages/core/src/agent/Agent.ts
```diff
@@ -156,6 +163,7 @@ export class Agent extends BaseAgent {
     dependencyManager.registerSingleton(TransportService)
     dependencyManager.registerSingleton(Dispatcher)
     dependencyManager.registerSingleton(EnvelopeService)
+    dependencyManager.registerSingleton(FeatureRegistry)
```
TimoGlastra, line 166: If feature registry is part of core, I think it should not live in the discover features module.

packages/core/src/modules/basic-messages/BasicMessagesModule.ts
```diff
@@ -1,5 +1,9 @@
 import type { DependencyManager, Module } from '../../plugins'
+import type { FeatureRegistry } from '../discover-features'
 
+import { Protocol } from '../discover-features'
```
TimoGlastra, line 4: Same here. I think the protocol / feature registry should be part of core so not all modules depend on the discover features module. This way the discover features module just exposes it as a protocol, but the internal registration is in core. What do you think?

packages/core/src/modules/basic-messages/__tests__/BasicMessagesModule.test.ts
```diff
@@ -9,9 +10,14 @@ const DependencyManagerMock = DependencyManager as jest.Mock<DependencyManager>
...
 
+jest.mock('../../discover-features/FeatureRegistry')
+const FeatureRegistryMock = FeatureRegistry as jest.Mock<FeatureRegistry>
+
+const featureRegistry = new FeatureRegistryMock()
+
 describe('BasicMessagesModule', () => {
   test('registers dependencies on the dependency manager', () => {
-    new BasicMessagesModule().register(dependencyManager)
+    new BasicMessagesModule().register(featureRegistry, dependencyManager)
```
TimoGlastra, line 20: Any reason you put featureRegistry first? It introduces a breaking change that may not be necessary?

packages/core/src/modules/discover-features/DiscoverFeaturesApi.ts
```diff
@@ -1,101 +1,144 @@
...
+
+    const { message: queryMessage } = await service.createQuery({
+      queries: options.queries,
+      comment: options.comment,
+    })
 
     const outbound = createOutboundMessage(connection, queryMessage)
+
+    const replaySubject = new ReplaySubject<Feature[]>(1)
+    if (options.awaitDisclosures) {
```
TimoGlastra, line 93: Interesting approach! I think I like it 😀

packages/core/src/modules/discover-features/DiscoverFeaturesEvents.ts
```diff
@@ -0,0 +1,28 @@
+import type { BaseEvent } from '../../agent/Events'
+import type { ConnectionRecord } from '../connections'
+import type { Feature, FeatureQueryOptions } from './models'
+
+export enum DiscoverFeaturesEventTypes {
```
TimoGlastra, line 5: 👍. This will provide a much better public API

packages/core/src/modules/discover-features/DiscoverFeaturesEvents.ts
```diff
@@ -0,0 +1,28 @@
...
+    protocolVersion: string
+    connection: ConnectionRecord
+    threadId: string
+  }
+}
+
+export interface DiscoverFeaturesDisclosureReceivedEvent extends BaseEvent {
+  type: typeof DiscoverFeaturesEventTypes.DisclosureReceived
+  payload: {
+    disclosures: Feature[]
```
TimoGlastra, line 23: Should we include the message in both events? If people have custom decorators for example it may be useful to expose those for extensibility

packages/core/src/modules/discover-features/DiscoverFeaturesApi.ts
```diff
@@ -1,101 +1,144 @@
...
+  DFSs extends DiscoverFeaturesService[] = [V1DiscoverFeaturesService, V2DiscoverFeaturesService]
+> implements DiscoverFeaturesApi<DFSs>
+{
   private connectionService: ConnectionService
   private messageSender: MessageSender
-  private discoverFeaturesService: DiscoverFeaturesService
   private eventEmitter: EventEmitter
   private stop$: Subject<boolean>
   private agentContext: AgentContext
+  private serviceMap: ServiceMap<DFSs>
```
TimoGlastra, line 36: In other modules I always expose the config as read only property on the API. This was you can access the config later on (and we may add a way to update certain properties at runtime)

packages/core/src/modules/discover-features/DiscoverFeaturesModuleConfig.ts
```diff
@@ -0,0 +1,24 @@
...
+}
+
+export class DiscoverFeaturesModuleConfig {
+  private options: DiscoverFeaturesModuleConfigOptions
+
+  public constructor(options?: DiscoverFeaturesModuleConfigOptions) {
+    this.options = options ?? {}
+  }
+
+  public get autoAcceptDiscoverFeatureQueries() {
```
TimoGlastra, line 21: Might be good to add an inheritDoc here (as used in other modules) so you get documentation when getting a property from the config

packages/core/src/modules/discover-features/DiscoverFeaturesModuleOptions.ts
```diff
@@ -0,0 +1,45 @@
+import type { FeatureQueryOptions } from './models'
+import type { DiscoverFeaturesService } from './services'
+
+/**
+ * Get the supported protocol versions based on the provided discover features services.
+ */
+export type ProtocolVersionType<DFSs extends DiscoverFeaturesService[]> = DFSs[number]['version']
```
TimoGlastra, line 7: This file should be Called DiscoverFeatureApiOptions I think?

packages/core/src/modules/discover-features/DiscoverFeaturesApi.ts
```diff
@@ -1,101 +1,144 @@
...
+   * Send a query to an existing connection for discovering supported features of any kind. If desired, do the query synchronously,
+   * meaning that it will await the response (or timeout) before resolving. Otherwise, response can be hooked by subscribing to
+   * {DiscoverFeaturesDisclosureReceivedEvent}.
+   *
+   * Note: V1 protocol only supports a single query and is limited to protocols
+   *
+   * @param options feature queries to perform, protocol version and optional comment string (only used
+   * in V1 protocol). If awaitDisclosures is set, perform the query synchronously with awaitDisclosuresTimeoutMs timeout.
+   */
+  public async queryFeatures(options: QueryFeaturesOptions<DFSs>) {
```
TimoGlastra, line 80: Maybe too much overhead/complex but do you think we can add return type definition that takes into account the answer to awaitDosclosures?

That way you don't have an optional return type, which can improve DX

TimoGlastra (approved):
This is perfect @genaris! Left some small notes, but overall looks good

packages/core/src/modules/discover-features/DiscoverFeaturesApi.ts
genaris: :muscle:

genaris:
> This is perfect @genaris! Left some small notes, but overall looks good

Thanks @TimoGlastra for your re-review :smile: . I think I've addressed all your suggestions (very appropriate BTW). There are of course some things to improve but I think it's ready to merge.

TimoGlastra:
Can you write a short message describing the breaking changes of this PR? Then I can include those in the commit for the change log

genaris:
> Can you write a short message describing the breaking changes of this PR? Then I can include those in the commit for the change log

OK I think that finally it's only about Discover Features V1 module usage so I would say:

- `queryFeatures` method parameters have been unified to a single `QueryFeaturesOptions` object that requires specification of Discover Features protocol to be used. 
- `isProtocolSupported` has been replaced by the more general synchronous mode of `queryFeatures`, which works when `awaitDisclosures` in options is set. Instead of returning a boolean, it returns an object with matching features
- Custom modules implementing protocols must register them in Feature Registry in order to let them be discovered by other agents (this can be done in module `register(dependencyManager, featureRegistry)` method)

PR #995, opened 2022-08-29
docs(demo): faber creates invitation

conanoc:
It's better if Faber creates an invitation than Alice does because we can test mobile agents with Faber more easily.
- Now, Faber creates an invitation and Alice receives it.
- I didn't change the code but just exchanged connection-related code between Alice and Faber.
- The second commit(1e6af7a) is a minor bug fix.


TimoGlastra (commented):
Thanks for the PR! This change is fine by me. However, do you think it will also be possible to modify the script so both sides can create and receive an invitation? Should be possible I think, and that way we don't have to remove the create invitation functionality from Alice

conanoc:
> However, do you think it will also be possible to modify the script so both sides can create and receive an invitation? Should be possible I think, and that way we don't have to remove the create invitation functionality from Alice

I'm not sure, but the functions like `getConnectionRecord()` will get complicated. Is it necessary for Alice has the invitation function?

TimoGlastra:
If it's too complicated it's fine to just switch it around 👍

TimoGlastra (approved):

PR #1003, opened 2022-09-01
fix: unable to resolve nodejs document loader in react native environment

### LINKED ISSUE #993: Unable to resolved https module in react-native
Unable to resolved `https` module from `/node_modules/@digitalcredentials/js onld/lib/documentLoaders/node.js `

I am trying to use PPV2 in react-native from branch `0.3.0-pre`, Once I installed the packages and trying to run the app I am getting below error.

Screenshot 2022-08-18 at 3 31 52 PM (1)

karimStekelenburg:
Fixes #993 by conditionally importing the XHR and NodeJS document loaders, based on the runtime environment.

### DIFF
--- a/packages/core/src/modules/vc/W3cCredentialService.ts
+++ b/packages/core/src/modules/vc/W3cCredentialService.ts
@@ -304,10 +304,10 @@ export class W3cCredentialService {
 
       let loader
 
-      if (isNodeJS()) {
-        loader = documentLoaderNode.apply(jsonld, [])
-      } else if (isReactNative()) {
-        loader = documentLoaderXhr.apply(jsonld, [])
+      if (isReactNative()) {
+        loader = documentLoaderXhr().apply(jsonld, [])
+      } else if (isNodeJS()) {
+        loader = documentLoaderNode().apply(jsonld, [])
       } else {
         throw new AriesFrameworkError('Unsupported environment')
       }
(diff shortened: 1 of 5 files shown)


packages/core/src/modules/vc/W3cCredentialService.ts
```diff
@@ -304,10 +304,10 @@ export class W3cCredentialService {
 
       let loader
 
-      if (isNodeJS()) {
-        loader = documentLoaderNode.apply(jsonld, [])
-      } else if (isReactNative()) {
-        loader = documentLoaderXhr.apply(jsonld, [])
+      if (isReactNative()) {
```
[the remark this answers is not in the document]
karimStekelenburg: Definitely not intending to go back to that and we should probably fix this by separate extensions in the future. This is just a quick fix for now ;)

packages/core/src/modules/vc/W3cCredentialService.ts
```diff
@@ -302,15 +303,9 @@ export class W3cCredentialService {
...
-      let loader
-
-      if (isNodeJS()) {
-        loader = documentLoaderNode.apply(jsonld, [])
-      } else if (isReactNative()) {
-        loader = documentLoaderXhr.apply(jsonld, [])
-      } else {
-        throw new AriesFrameworkError('Unsupported environment')
-      }
+      // fetches the documentLoader from documentLoader.ts or documentLoader.native.ts depending on the platform at bundle time
```
TimoGlastra, line 306: The isNodeJS and isReactNative methods are defined but never used. Can we remove those methods @karimStekelenburg ?

image

packages/core/src/utils/environment.ts
```diff
@@ -5,5 +5,5 @@ export function isNodeJS() {
 export function isReactNative() {
```
TimoGlastra, line 5: Are these still used?

packages/core/src/modules/vc/W3cCredentialService.ts
karimStekelenburg: Nope, removed

TimoGlastra (approved):

PR #1014, opened 2022-09-06
feat: problem report on decline

ryankoch13:
Adds a new problem report reason for credential / presentation decline and sends a problem report when declining an offer or request.

### DIFF
--- a/packages/core/src/modules/credentials/CredentialsModule.ts
+++ b/packages/core/src/modules/credentials/CredentialsModule.ts
@@ -42,6 +42,7 @@ import { V1CredentialService } from './protocol/v1/V1CredentialService'
 import { V2CredentialService } from './protocol/v2/V2CredentialService'
 import { CredentialRepository } from './repository/CredentialRepository'
 import { RevocationNotificationService } from './services'
+import { CredentialProblemReportReason } from './errors'
 
 export interface CredentialsModule<CFs extends CredentialFormat[], CSs extends CredentialService<CFs>[]> {
   // Proposal methods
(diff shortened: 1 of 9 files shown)


genaris (commented):
Thanks Ryan! It seems to me that this credential/proof decline is a fairly common situation to deserve an specific reason code. However, I could not find it mentioned in the RFCs for V2 protocols. Is there a PR for adding it explicitly?

packages/core/src/modules/credentials/CredentialsModule.ts
```diff
@@ -533,15 +542,15 @@ export class CredentialsModule<
       throw new AriesFrameworkError(`No connectionId found for credential record '${credentialRecord.id}'.`)
     }
     const connection = await this.connectionService.getById(credentialRecord.connectionId)
-
     const service = this.getService(credentialRecord.protocolVersion)
-    const problemReportMessage = service.createProblemReport({ message: options.message })
+
+    const problemReportMessage = options.code ? service.createProblemReport({ message: options.message, code: options.code }) : service.createProblemReport({ message: options.message })
```
genaris, line 547: If you just send an undefined `code` to createProblemReport wouldn't save the same as the ternary operator? :thinking:

packages/core/src/modules/credentials/CredentialsModule.ts
```diff
@@ -354,6 +355,14 @@ export class CredentialsModule<
     const service = this.getService(credentialRecord.protocolVersion)
     await service.updateState(credentialRecord, CredentialState.Declined)
 
+    //  if credential was offered by established connection, attempt to send problem report upon credential decline
```
TimoGlastra, line 358: There have been discussions in the past on whether sending a problem report is desired as it leaks some info you may not want to share with the other party. I think we should therefore make it something configurable by e.g. adding a `sendProblemReport` option:

```ts
await agent.credentials.declineOffer({
  credentialRecordId: 'xxx',
  sendProblemReport: true,
  problemReportMessage: 'something' # optional message
```

packages/core/src/modules/credentials/CredentialsModule.ts
```diff
@@ -354,6 +355,14 @@ export class CredentialsModule<
...
+    if (credentialRecord.connectionId) {
```
TimoGlastra, line 359: We can also respond to connectionless exchange using the same technique we do as with the other methods (service decorator). Maybe we can add that here also?

It should at least throw an error I think instead of silently not do anything without a connection id.

packages/core/src/modules/credentials/CredentialsModule.ts
```diff
@@ -354,6 +355,14 @@ export class CredentialsModule<
...
+      this.sendProblemReport({
+        message: `Credential Offer ${credentialRecordId} has been declined`,
```
TimoGlastra, line 361: The credentialRecordId is irrelevant to the other party, maybe we should use the thread id instead?

packages/core/src/modules/credentials/CredentialsModule.ts
```diff
@@ -354,6 +355,14 @@ export class CredentialsModule<
...
+        code: CredentialProblemReportReason.RequestDeclined,
```
TimoGlastra, line 362:
```suggestion
        code: CredentialProblemReportReason.OfferDeclined,
```

TimoGlastra (commented):
I'm onboard with adding this feature, but do think it should be configurable / controlable. Currently this is a breaking change, so we need to either merge it into 0.3.0-pre or change the default behaviour to not send a problem report (so it is opt-in)

TimoGlastra:
@ryankoch13 would you still like this PR to be merged? I'm going to close it for now, please create a new PR when you're ready.

PR #1020, opened 2022-09-08
feat(proofs): proofs module migration script for 0.3.0

### LINKED ISSUE #996: Write migration script for present proof in 0.3.0

TimoGlastra:
Fixes #996


packages/core/src/storage/migration/UpdateAssistant.ts
```diff
@@ -79,23 +95,47 @@ export class UpdateAssistant<Agent extends BaseAgent = BaseAgent> {
...
+          isFirstVersionEqualToSecond(currentStorageVersion, parsedToVersion))
+      ) {
+        throw new StorageUpdateError(
+          `Can't update to version ${updateToVersion} because it is lower or equal to the current agent storage version ${currentStorageVersion[0]}.${currentStorageVersion[1]}}`
+        )
+      }
 
       if (neededUpdates.length == 0) {
-        this.agent.config.logger.info('No update needed. Agent storage is up to date.')
+        this.agent.config.logger.info(`No update needed. Agent storage is up to date.`)
```
genaris, line 132:
```suggestion
        this.agent.config.logger.info('No update needed. Agent storage is up to date.')
```

packages/core/src/storage/migration/__tests__/0.2.test.ts
```diff
@@ -0,0 +1,144 @@
+import type { FileSystem } from '../../..'
```
genaris, line 1: Importing from ..

packages/core/src/storage/migration/__tests__/0.2.test.ts
```diff
@@ -0,0 +1,144 @@
...
+      },
+    ])
+
+    await updateAssistant.update()
+
+    expect(await updateAssistant.isUpToDate()).toBe(true)
+    expect(await updateAssistant.getNeededUpdates()).toEqual([])
+
+    // MEDIATOR_ROUTING_RECORD recipientKeys will be different every time, and is not what we're testing here
+    delete storageService.records.MEDIATOR_ROUTING_RECORD
```
genaris, line 78: We've recently had this issue in #985 and I've ended up mocking IndyWallet's createDid method to create a consistent routing record. 

Even if my preference would be to keep the database as loaded from file, your solution is simpler and probably better considering a future decoupling from IndyWallet, so maybe would be nice to implement it in 0.1.test.ts as well.

packages/core/src/storage/migration/__tests__/UpdateAssistant.test.ts
```diff
@@ -54,17 +54,25 @@ describe('UpdateAssistant', () => {
...
+      expect(await updateAssistant.isUpToDate('0.2')).toBe(true)
+    })
+
+    it('should return true for current agent storage version', async () => {
+      expect(await updateAssistant.isUpToDate('0.3')).toBe(true)
+    })
   })
 
   describe('UpdateAssistant.frameworkStorageVersion', () => {
     it('should return 0.2', async () => {
```
genaris, line 68:
```suggestion
    it('should return 0.3', async () => {
```
(or `CURRENT_FRAMEWORK_STORAGE_VERSION`, see my other comment)

packages/core/src/storage/migration/__tests__/UpdateAssistant.test.ts
```diff
@@ -54,17 +54,25 @@ describe('UpdateAssistant', () => {
     it('should return true when a new wallet is created', async () => {
       expect(await updateAssistant.isUpToDate()).toBe(true)
     })
+
+    it('should return true for a lower version than current storage', async () => {
+      expect(await updateAssistant.isUpToDate('0.2')).toBe(true)
+    })
+
+    it('should return true for current agent storage version', async () => {
```
genaris, line 62: Well, we can also make it return false by sending '0.4' parameter :sweat_smile:

Also I guess `CURRENT_FRAMEWORK_STORAGE_VERSION` can be used instead of hard-coded 0.3.  Could it be possible to use this constant in update the test title? :thinking:

genaris (approved):
I've left some very small comments regarding version constants and consistency with 0.1 migration test.

Interesting this new `updateToVersion` parameter for update. Do you see it useful not only for tests but also for some specific cases where the framework consumer wants to do this update process step by step? A possible scenario can be huge wallets where each step can take significant time and prefer to have a log or feedback more often.

TimoGlastra:
> Interesting this new `updateToVersion` parameter for update. Do you see it useful not only for tests but also for some specific cases where the framework consumer wants to do this update process step by step? A possible scenario can be huge wallets where each step can take significant time and prefer to have a log or feedback more often.

I initially implemented it because I wanted to be able to test a single update (0.1 to 0.2, and in future others). However I do think it can be useful in mobile wallets to show a loading screen where you update per version, and get more insights into the process. A better way to do this would be by adding a progress object that you can see currentVersion / totalVersions and a percentage or something, but that would be more work to implement.

PR #1029, opened 2022-09-16
feat: use did:key flag

genaris:
An initial step to support [Aries RFC 0360](https://github.com/hyperledger/aries-rfcs/tree/main/features/0360-use-did-key) by adding an optional configuration flag `useDidKeyInProtocols` that will define how agent should format keys in protocols where keys are exchanged: 'naked' (base58) or did:key encoded. 

This setting will be used unless the other party has already sent us keys (for instance, a Mediator sent us the routing key in a Mediation Grant message). In such case, their format will be used in subsequent messages for that protocol (e.g. KeyList Update). This is achieved by adding a specific metadata key `UseDidKeysForProtocol` to the related connection record, which  intended to track the other party support of did:key in different RFCs. 

Currently, only Coordinate Mediation (RFC 0211) uses this feature, but in further versions other protocols will use this metadata/settings key (such as Pickup V2 - RFC 0685). 

For key reception, the agent will always accept both formats and internally store in base58.

### DIFF
--- a/packages/core/src/modules/connections/repository/ConnectionMetadataTypes.ts
+++ b/packages/core/src/modules/connections/repository/ConnectionMetadataTypes.ts
@@ -1,0 +1,9 @@
+export enum ConnectionMetadataKeys {
+  UseDidKeysForProtocol = '_internal/useDidKeysForProtocol',
+}
+
+export type ConnectionMetadata = {
+  [ConnectionMetadataKeys.UseDidKeysForProtocol]: {
+    [protocolUri: string]: boolean
+  }
+}
--- a/packages/core/src/modules/routing/services/MediationRecipientService.ts
+++ b/packages/core/src/modules/routing/services/MediationRecipientService.ts
@@ -134,9 +134,9 @@ export class MediationRecipientService {
     // update keylist in mediationRecord
     for (const update of keylist) {
       if (update.action === KeylistUpdateAction.add) {
-        mediationRecord.addRecipientKey(verkeyToDidKey(update.recipientKey))
+        mediationRecord.addRecipientKey(didKeyToVerkey(update.recipientKey))
       } else if (update.action === KeylistUpdateAction.remove) {
-        mediationRecord.removeRecipientKey(verkeyToDidKey(update.recipientKey))
+        mediationRecord.removeRecipientKey(didKeyToVerkey(update.recipientKey))
       }
     }
 
--- a/packages/core/src/modules/routing/services/MediationRecipientService.ts
+++ b/packages/core/src/modules/routing/services/MediationRecipientService.ts
@@ -5,7 +5,7 @@ import type { EncryptedMessage } from '../../../types'
 import type { ConnectionRecord } from '../../connections'
 import type { Routing } from '../../connections/services/ConnectionService'
 import type { MediationStateChangedEvent, KeylistUpdatedEvent } from '../RoutingEvents'
-import type { KeylistUpdateResponseMessage, MediationDenyMessage } from '../messages'
+import type { MediationDenyMessage } from '../messages'
 import type { StatusMessage, MessageDeliveryMessage } from '../protocol'
 import type { GetRoutingOptions } from './RoutingService'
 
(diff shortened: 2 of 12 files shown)


swcurran:
@andrewwhitehead -- heads up on this.  We probably want to have ACA-Py doing the same thing.
@ianco, @shaangill025 -- heads up on this.  We need some AATH tests that use this, with the ACME agent started with this option.
@genaris -- it would good to get this added to AATH so that compatibility with other Aries Frameworks can be verified.

packages/core/src/modules/connections/repository/ConnectionMetadataTypes.ts
```diff
@@ -0,0 +1,9 @@
+export enum ConnectionMetadataKeys {
+  UseDidKeysForProtocol = '_internal/useDidKeysForProtocol',
+}
+
+export type ConnectionMetadata = {
+  [ConnectionMetadataKeys.UseDidKeysForProtocol]: {
+    [protocolUri: string]: boolean
```
TimoGlastra, line 7: Specific reason for making this a key/value object instead of an array with string uri's?

packages/core/src/modules/routing/services/MediationRecipientService.ts
```diff
@@ -104,6 +105,14 @@ export class MediationRecipientService {
     // Update record
     mediationRecord.endpoint = messageContext.message.endpoint
 
+    // Update connection metadata to use their key format in further protocol messages
+    const connectionUsesDidKey = messageContext.message.routingKeys.some(isDidKey)
+
+    const useDidKeysForProtocol = connection.metadata.get(ConnectionMetadataKeys.UseDidKeysForProtocol) ?? {}
+    useDidKeysForProtocol[MediationGrantMessage.type.protocolUri] = connectionUsesDidKey
```
TimoGlastra, line 112: How does this work with minor version updates?

packages/core/src/modules/routing/services/MediationRecipientService.ts
```diff
@@ -125,9 +134,9 @@ export class MediationRecipientService {
     // update keylist in mediationRecord
     for (const update of keylist) {
       if (update.action === KeylistUpdateAction.add) {
-        mediationRecord.addRecipientKey(update.recipientKey)
+        mediationRecord.addRecipientKey(verkeyToDidKey(update.recipientKey))
```
TimoGlastra, line 137: Do we need a migration script to update non did keys?

TimoGlastra (commented):
Nice 👍

TimoGlastra (approved):

packages/core/src/modules/connections/repository/ConnectionMetadataTypes.ts
genaris: The goal of this metadata key is to override agent settings, based on what we learn from the connection. So if our agent is set to always use did:key by default and we learn that the other side is not using it for a certain protocol, we'll set the key for that protocol to false.

And we could also have the opposite case: we don't use did:key but it happens that the other side is using them, so we set the key for that protocol to true. Also we might change our agent settings afterwards, and it will still take into account the metadata key for the given connection.

packages/core/src/modules/routing/services/MediationRecipientService.ts
genaris: This is tied to a specific protocol major.minor version. If a new interaction is started with a newer protocol version, a new key/value will be added. Not super nice in terms of storage optimization, but I guess this will be mostly limited to specific connections (such as mediators), so should not represent a big issue 🤔

packages/core/src/modules/routing/services/MediationRecipientService.ts
genaris: No, actually it was supposed to be a fix for the case where we receive keys in did:key format and it made it worst 😝.

It should use `didKeyToVerkey` instead. I'll add a test to `MediationRecipientService.test.ts` to make sure it stores all keys in base58 and it behaves exactly as current wallets.

packages/core/src/modules/routing/services/MediationRecipientService.ts
TimoGlastra: Ok, something to look out for. We maybe don't need to store it separtely per minor version (however if ACA-Py updated to mediator coordinator 1.1 and 1.0 already supported did:key I would assume the newer version also does).

packages/core/src/modules/connections/repository/ConnectionMetadataTypes.ts
TimoGlastra: Makes sense. I thought we would only use it to say a protocol does support did:key, but there is also the case where we explicitly store a protocol doesn't store it.

Is this also updated autmoatically when this changes? E.g. we have stored false so we keep sending base58 keys, but then the other agent sends a did:key, do we update it automatically? Or is it set once and kept like this forever?

TimoGlastra:
Ready to merge?

packages/core/src/modules/connections/repository/ConnectionMetadataTypes.ts
genaris: > Is this also updated autmoatically when this changes? E.g. we have stored false so we keep sending base58 keys, but then the other agent sends a did:key, do we update it automatically? Or is it set once and kept like this forever?

I think it depends on each protocol/service but generally the best would be to update it automatically. As of current AFJ support for mediator role, it does not add so much value because we only support Mediation Grant and Keylist Update request/response. However, considering a future support of Keylist queries and responses it's better to handle it right from the start.

So I'm adding automatic handling for both recipient and mediator in the messages we currently support. Hopefully in 0.3.x we'll support all of them :smile:

PR #1030, opened 2022-09-16
chore: merge branch 'main' into 0.3.0-pre

genaris:
Merging 0.2.3 and 0.2.4 features into 0.3.x branch:

- All tests pass
- Demo works
- Action Menu module was added to the core for the moment

Hope to not have missed anything. Please double check!


TimoGlastra (approved):
One comment, but otherwise looks good to me! Thanks for doing this @genaris!

TimoGlastra:
It seems DCO is failing. I think I could bypass, but worried that it will cause issues when we merge 0.3.0 into main

genaris:
> It seems DCO is failing. I think I could bypass, but worried that it will cause issues when we merge 0.3.0 into main

Yes, it's something that annoys me too. Do you know why it happens, since the original commits were correctly signed-off (or at least somehow passed it at that time)?

Probably, if DCO keeps failing, we can rebase this branch just before merging it into main. I don't want to do it now because I'm afraid that it will break some ongoing PRs, but I'm not an expert on this... I wanted to keep using SVN! 😝

genaris:
Allright, I've re-updated this branch to match current 0.3.0-pre and main and did some local tests: first, I've created a PR of this branch against my 0.3.0-pre (using squash & merge), and then a PR from my 0.3.0-pre to my main. Results of a potential merge into main can be seen in https://github.com/2060-io/aries-framework-javascript/commits/main-merged. Not so bad: just a few extra Merge commits and the commit history is mostly fine.

I couldn't test the DCO issue (maybe we could create copies of 0.3.0-pre and main in hyperledger repo to do a final test). However, as far as I can see, in #965 there was already a DCO issue that you set to pass with the magic button and currently 0.3.0-pre branch is fine.

PR #1035, opened 2022-09-22
feat(bbs): extract bbs logic into separate module

### LINKED ISSUE #948: Extract bbs dependencies into @aries-framework/bbs-signatures packages

TimoGlastra:

Fixes #948 

Extract the bbs logic into a separate `@aries-framework/module-bbs`. I had to export a lot of the vc stuff from the core bloating the core imports a bit. I think this will be solved when we extract the vc logic into it's own vc package.

For now I've disabled the bbs tests to run on node 17 and 18, so we will have passing CI again, with the knowledge that BBS is broken in node 17 and 18. I also set the package to private for now so that it won't be released with the 0.3.0 release and we have more time to fix it and get it ready for release.

What do we think of the name `@aries-framework/module-bbs`? Should we call it `@aries-framework/bbs-signatures`? It's private for now, so we have room to rename it until we remote the private: true from the package.json

There's is only really ugly hack (one of the first files in the file overview) that has to do with tsyringe injection not allowing nothing to be registered for a token. We have a signing provider that is only used for bbs, and in if you don't register the bbs module it will fail. I didn't want to spend too much time to rework it, so implemented a quick workaround for now


TimoGlastra:
Good to see all tests passing again in 0.3.0 branch :)

packages/core/src/crypto/WalletKeyPair.ts
```diff
@@ -1,15 +1,14 @@
 import type { Wallet } from '..'
```
blu3beri, line 1: No import from '..' (not your in your PR though :))

packages/core/src/modules/vc/deriveProof.ts
```diff
@@ -51,9 +52,8 @@ export const deriveProof = async (
   if (proofs.length === 0) {
     throw new Error(`There were not any proofs provided that can be used to derive a proof with this suite.`)
```
blu3beri, line 53: We should throw an AriesFramework Error or BbsSignatures Error? How do we do the error class for modules? Always AriesFramework or does the module have a specific error class?

packages/module-bbs/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "@aries-framework/module-bbs",
+  "main": "build/index",
+  "types": "build/index",
+  "version": "0.2.2",
```
blu3beri, line 5: Will all the separate modules be in version-sync?

packages/module-bbs/tests/util.ts
```diff
@@ -0,0 +1,19 @@
+export function testSkipNode17And18(...parameters: Parameters<typeof test>) {
```
blu3beri, line 1: Will this prompt an output as to why they are being skipped?

blu3beri (approved):
Just a couple of questions. Nothing really major.

packages/core/src/crypto/WalletKeyPair.ts
genaris: > No import from '..' (not your in your PR though :))

The classic one!

genaris:
> Good to see all tests passing again in 0.3.0 branch :)

So refreshing!

I have no preference with the naming but I agree with @blu3beri that we should add a discussion regarding the modularisation (once we finish this 0.3.0 though) and that it would be appropriate to indicate the reason why we are skipping node 17 and 18 tests. I know we will not forget, as it has prevented us from sleeping for months, but just in case...

jakubkoci:
Just my two cents about the naming, I would rather remove `module` or `package` from the name and just use `@aries-framework/bbs`. We have some struggles with a `module` name inside the framework, which particular meaning is still to be defined. The fact that it's some sort of package is possible to deduce from the context.

packages/module-bbs/package.json
```diff
@@ -0,0 +1,39 @@
...
+  },
+  "scripts": {
+    "build": "yarn run clean && yarn run compile",
+    "clean": "rimraf -rf ./build",
+    "compile": "tsc -p tsconfig.build.json",
+    "prepublishOnly": "yarn run build",
+    "test": "jest"
+  },
+  "dependencies": {
+    "@aries-framework/core": "0.2.2",
```
blu3beri, line 28: Shouldn't core be a peer dependency?

packages/module-bbs/src/__tests__/BbsModule.test.ts
```diff
@@ -0,0 +1,41 @@
...
+import { BbsBlsSignature2020, BbsBlsSignatureProof2020 } from '../signature-suites'
+
+const dependencyManager = {
+  registerInstance: jest.fn(),
+  registerSingleton: jest.fn(),
+} as unknown as DependencyManager
+
+describe('BbsModule', () => {
+  test('registers dependencies on the dependency manager', () => {
+    const tenantsModule = new BbsModule()
```
blu3beri, line 21:
```suggestion
    const bbsModule = new BbsModule()
```

packages/core/package.json
```diff
@@ -26,8 +26,6 @@
     "@digitalcredentials/jsonld": "^5.2.1",
```
blu3beri, line 26: Shouldn't the new module also be added in core now?

packages/core/src/modules/vc/deriveProof.ts
TimoGlastra: Create an issue for this! There's a lot of other places where we do this in the vc code, so would like to address this in a separate PR

packages/module-bbs/tests/util.ts
TimoGlastra: Nope... :)

blu3beri: Do you know if this is something we can add? Maybe just a `console.warn` giving a small explanation and linking to an issue?

blu3beri (approved):
Looks good to me!

PR #1040, opened 2022-10-02
feat(question-answer): separate logic to a new module

blu3beri:
Separated the question-answer logic to a new module.

Since it is now a separate module, if core gets released before the question-answer module it will not be included in the core anymore. This is a breaking change and might cause some headaches.

We can just merge it in after 0.3.0 is released and take our time with a strategy to add it back into the core for now. 

---

I could also keep the question-answer module inside the core and release this one separately. When we do a new 0.3.1 or 0.4.0 release the can remove it and make it dependent on this one.


genaris:
Good to see you are taking your part of the modularisation homework :-)

Just keep in mind that there were some fixes in question-answer, not yet merged into 0.3.0-pre branch so I think are not yet integrated here

blu3beri:
> Just keep in mind that there were some fixes in question-answer, not yet merged into 0.3.0-pre branch so I think are not yet integrated here

Good point, will update when 0.3.0-pre is rebased.

TimoGlastra:
Is there any reason not to merge this before 0.3.0 is released? If it's ready we can merge it right?

blu3beri:
> Is there any reason not to merge this before 0.3.0 is released? If it's ready we can merge it right?

It should be possible. However I have a suggestion and I am not sure if it makes sense. 

I would like to release this package with the 0.3.0 release and ALSO keep the original module still in core. When we release a 0.3.1 we can take it out. This makes sure that the module is inside core before it is released. 

If we release core first it will look at the registry and this module does not exist. 

Does this make sense or is there an easier solution?

TimoGlastra:
> Does this make sense or is there an easier solution?

We can't take it out in 0.3.1, as that would be a breaking change and warrant a 0.4.0 release. But I'm also not sure if it makes 100% sense. The packages will be released at the same time, and if you want to use the question answer module you can just add it as as custom module? Why would one be released before the other?

genaris:
I agree that 0.3.0 is a good timing to do this. The same with Action Menu module. I can do it directly in #1030 or create a new PR like this one once if it's merged (would make the commit history a bit cleaner I guess). Let me know your thoughts!

genaris (approved):

genaris:
@TimoGlastra @blu3beri, I don't know what you guys think but I'd rather prefer to merge first #1030 and then update this PR with the changes introduced in question-answer module. And in parallel I can extract action-menu so we can get the initial modularisation job done in a clean way before releasing 0.3.0.

blu3beri:
> @TimoGlastra @blu3beri, I don't know what you guys think but I'd rather prefer to merge first #1030 and then update this PR with the changes introduced in question-answer module. And in parallel I can extract action-menu so we can get the initial modularisation job done in a clean way before releasing 0.3.0.

Sounds good to me :).

PR #1042, opened 2022-10-03
feat: Presentation Exchange Format / PPV2

NB-MikeRichardson:
This PR puts the PEX PPV2 code into 0.3.0-pre.

Note that there is still some development to do but this is the correct time to merge


NB-MikeRichardson:
I will reopen this PR when the JsonLd ICV2 branch is merged as there are many duplicate commits across the two PRs.

PR #1043, opened 2022-10-04
feat: possibility to set masterSecretId inside of WalletConfig

an-uhryn:


jakubkoci:
Thanks, @an-uhryn. Could you please fix the linter error? I think we can disable `@typescript-eslint/no-non-null-assertion` for the whole `Wallet.test.ts` file.

genaris:
> Thanks, @an-uhryn. Could you please fix the linter error? I think we can disable `@typescript-eslint/no-non-null-assertion` for the whole `Wallet.test.ts` file.

Yes, and don't forget the DCO issue following the instructions from [here](https://github.com/hyperledger/aries-framework-javascript/pull/1043/checks?check_run_id=8703837779).

packages/core/src/types.ts
```diff
@@ -26,6 +26,7 @@ export interface WalletConfig {
     type: string
     [key: string]: unknown
   }
+  masterSecretId?: 'string'
```
genaris, line 29: Maybe I'm missing something but I think it should be string without quotes
```suggestion
  masterSecretId?: string
```

packages/core/src/wallet/IndyWallet.ts
```diff
@@ -56,6 +56,10 @@ export class IndyWallet implements Wallet {
   }
 
   public get masterSecretId() {
+    if (this.walletConfig?.masterSecretId) {
```
genaris, line 59: Considering the checks that are done in line 63, for consistency with other conditions, maybe we could avoid this if sentence and simply return `this.walletConfig.masterSecretId ?? this.walletConfig.id`

genaris (commented):
Thank you @an-uhryn!

I'm not an expert on this topic so consider some of my comments more as suggestions than changes requested.

I guess this PR tackles mainly scenarios where wallet has been created externally, but the resulting code left me a doubt regarding usage of this masterSecretId when it is created within the framework: in the method `createAndOpen` from IndyWallet.ts I see: 

```ts
      await this.createMasterSecret(this.handle, walletConfig.id)
``` 

Shouldn't be `walletConfig.masterSecretId` used in this case unless not specified (in which case `walletConfig.id` would be used instead)?

packages/core/src/types.ts
blu3beri: Yes, this only allows the string literal `string`. Any other string is not allowed.

an-uhryn:
@jakubkoci @genaris @blu3beri Thanks for your review and suggestions. I made all changes.

packages/core/src/wallet/IndyWallet.ts
```diff
@@ -56,13 +56,13 @@ export class IndyWallet implements Wallet {
...
   public get masterSecretId() {
-    if (!this.isInitialized || !this.walletConfig?.id) {
+    if (!this.isInitialized) {
       throw new AriesFrameworkError(
         'Wallet has not been initialized yet. Make sure to await agent.initialize() before using the agent.'
       )
     }
 
-    return this.walletConfig.id
+    return this.walletConfig?.masterSecretId ?? this.walletConfig?.id
```
TimoGlastra, line 65: This can lead to undefined being returned, which wasn't possible before. We should probably check if either have a value and throw an error otherwise

packages/core/src/wallet/Wallet.test.ts
```diff
@@ -1,13 +1,24 @@
...
-  const config = getAgentConfig('WalletTest')
-  const wallet = new IndyWallet(config)
+  let wallet: IndyWallet
+  let config: AgentConfig
+  let configWithMasterSecretId: AgentConfig
+
+  beforeEach(() => {
+    config = getAgentConfig('WalletTest')
+    configWithMasterSecretId = getAgentConfig('WalletTestWithMasterSecretId')
+    configWithMasterSecretId.walletConfig!.masterSecretId = 'customMasterSecretId'
```
TimoGlastra, line 16: Maybe you can extend the getAgentConfig method and pass the masterSecretId to that method?

TimoGlastra (commented):
As the masterSecret is still created when you create the wallet, what is the use case of this feature?

Fine to merge, just curious as to why you'd want this

an-uhryn:
> As the masterSecret is still created when you create the wallet, what is the use case of this feature?
> 
> Fine to merge, just curious as to why you'd want this

Hey @TimoGlastra. We are migrating from VCX Library to Aries Framework JavaScript. We have some credentials which were created with VCX Library (with masterSecretId) and after credentials migrating we can't accept proof request without setting masterSecretId.

packages/core/src/wallet/Wallet.test.ts
an-uhryn: Yeah, sure!

packages/core/src/wallet/IndyWallet.ts
an-uhryn: It's good question. I'm not sure but is it possible case that `isInitialized` is `true` but we don't have `this.walletConfig.id`?
It seems like we have to open wallet to make `isInitialized` to be `true`, but we can't open wallet without wallet id.

TimoGlastra:
Makes sense! 👍

packages/core/src/wallet/IndyWallet.ts
genaris: I think this is the reason why CI is failing (a few lines in `IndyHolderService.ts` are expecting it to be always defined, something that makes sense because otherwise it cannot work).

an-uhryn: I returned back this check and also aded for `masterSecretId`. Hope now everything will be fine 🙂

TimoGlastra:
There's some prettier errors, can you fix those?

packages/core/tests/helpers.ts
```diff
@@ -65,12 +66,17 @@ export const genesisPath = process.env.GENESIS_TXN_PATH
 export const publicDidSeed = process.env.TEST_AGENT_PUBLIC_DID_SEED ?? '000000000000000000000000Trustee9'
 export { agentDependencies }
 
-export function getBaseConfig(name: string, extraConfig: Partial<InitConfig> = {}) {
+export function getBaseConfig(
+    name: string,
+    extraConfig: Partial<InitConfig> = {},
```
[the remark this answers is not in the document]
an-uhryn: Yeah thanks, I missed that.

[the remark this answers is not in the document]
an-uhryn: Did you mean that?
`
getAgentConfig('WalletTestWithMasterSecretId', {
      walletConfig: {
        id: `Wallet: WalletTestWithMasterSecretId`,
        key: `Key: WalletTestWithMasterSecretId`,
        masterSecretId: 'customMasterSecretId',
      },
    })
`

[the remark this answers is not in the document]
an-uhryn: Did you mean `getAgentConfig`?
```
getAgentConfig('WalletTestWithMasterSecretId', {
    walletConfig: {
        id: `Wallet: WalletTestWithMasterSecretId`,
        key: `Key: WalletTestWithMasterSecretId`,
        masterSecretId: 'customMasterSecretId',
    },
})
```

genaris (approved):

TimoGlastra (approved):

Issue #1022, opened 2022-09-09
Expose findByQuery methods in public API of all modules


TimoGlastra:
We want to expose some of the repository methods to the public api to allow better querying/updating support without needing to inject internal dependneices. A lot of modules already expose some repository methods (getAll, findById, etc..), but we should look to extend this.

This issue focuses on the findByQuery method, we should add it to all modules that expose a getAll method

PR #1044, opened 2022-10-05
feat: expose findAllByQuery method in modules and services

rapaktech:
…ecessary tests for the switch

Signed-off-by: Jim Ezesinachi <ezesinachijim@gmail.com>


packages/core/src/modules/action-menu/services/ActionMenuService.ts
```diff
@@ -9,6 +9,8 @@ import type {
   CreateRequestOptions,
   FindMenuOptions,
 } from './ActionMenuServiceOptions'
+// eslint-disable-next-line no-restricted-imports
+import type { Query } from 'packages/core/src/storage/StorageService'
```
genaris, line 13: To avoid these eslint errors you should use relative paths from each file. Something like:

```suggestion
// eslint-disable-next-line no-restricted-imports
import type { Query } from '../../../storage/StorageService'
```

packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -8,6 +8,8 @@ import type { ConnectionStateChangedEvent } from '../ConnectionEvents'
 import type { ConnectionProblemReportMessage } from '../messages'
 import type { ConnectionType } from '../models'
 import type { ConnectionRecordProps } from '../repository/ConnectionRecord'
+// eslint-disable-next-line no-restricted-imports
```
genaris, line 11: Same as above regarding relative imports

packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -8,6 +8,8 @@ import type { ConnectionStateChangedEvent } from '../ConnectionEvents'
...
+import type { Query } from 'packages/core/src/storage/StorageService'
```
genaris, line 12: Same as above regarding relative imports

packages/core/src/modules/action-menu/services/ActionMenuService.ts
```diff
@@ -346,6 +348,12 @@ export class ActionMenuService {
     })
   }
 
+  public async findByQuery(options: Query<ActionMenuRecord>) {
+    return await this.actionMenuRepository.findByQuery({
```
genaris, line 352: Isn't possible to simply pass options as is, i.e. without the spread operator?

packages/core/src/modules/proofs/services/ProofService.ts
```diff
@@ -7,6 +7,8 @@ import type { ProofStateChangedEvent } from '../ProofEvents'
 import type { PresentationPreview, PresentationPreviewAttribute } from '../messages'
 import type { PresentationProblemReportMessage } from './../messages/PresentationProblemReportMessage'
 import type { CredDef, IndyProof, Schema } from 'indy-sdk'
+// eslint-disable-next-line no-restricted-imports
+import type { Query } from 'packages/core/src/storage/StorageService'
```
genaris, line 11: Same as above regarding relative imports

packages/core/src/modules/credentials/services/CredentialService.ts
```diff
@@ -24,6 +24,8 @@ import type {
 } from '../CredentialServiceOptions'
 import type { CredentialFormat, CredentialFormatService } from '../formats'
 import type { CredentialExchangeRecord, CredentialRepository } from './../repository'
+// eslint-disable-next-line no-restricted-imports
+import type { Query } from 'packages/core/src/storage/StorageService'
```
genaris, line 28: Same as above regarding relative imports

genaris (changes requested):
Thanks @rapaktech ! I think this will be very useful. I left some comments in the code but I would like to add some general suggestions that maybe others can comment on as well:

- I would prefer to use the naming 'findAllByQuery' to emphasize that multiple records could be retrieved. This is the same naming as used in other modules such as basic-messages and generic-records
- As I think this is related to #1022, it would be nice to expose these methods in the corresponding module classes. If I'm not wrong, you can even modify the parameters of `findAllByQuery` of `BasicMessagesModule` and `GenericRecordsModule` without introducing a breaking change, since `Query<T extends BaseRecord>` contains `Partial<RecordTags<T extends BaseRecord>>`

packages/core/src/modules/action-menu/services/ActionMenuService.ts
rapaktech: It should, but I was thinking it maight be safer to pass in a spread object

packages/core/src/modules/question-answer/QuestionAnswerModule.ts
```diff
@@ -92,6 +94,17 @@ export class QuestionAnswerModule {
     return this.questionAnswerService.getAll()
   }
 
+  /**
+   * Get all QuestionAnswer records by specified query params
+   *
+   * @returns list containing all QuestionAnswer records matching specified query params
+   */
+  public findAllByQuery(
+    query: Partial<{ connectionId: string; role: QuestionAnswerRole; state: QuestionAnswerState; threadId: string }>
```
TimoGlastra, line 103: Query<QuestionAnswerRecord>

packages/core/src/modules/routing/services/MediationRecipientService.ts
```diff
@@ -365,6 +366,10 @@ export class MediationRecipientService {
     return this.mediationRepository.getAll()
   }
 
+  public async getMediatorsByQuery(query: Query<MediationRecord>): Promise<MediationRecord[]> {
```
TimoGlastra, line 369:
```suggestion
  public async findAllMediatorsByQuery(query: Query<MediationRecord>): Promise<MediationRecord[]> {
```

packages/core/src/storage/StorageService.ts
```diff
@@ -10,7 +11,7 @@ interface AdvancedQuery<T extends BaseRecord> {
   $not?: Query<T>
 }
 
-export type Query<T extends BaseRecord> = AdvancedQuery<T> | SimpleQuery<T>
+export type Query<T extends BaseRecord<any, any, any>> = AdvancedQuery<T> | SimpleQuery<T>
```
genaris, line 14: This is causing the validation error in the CI. What's the reason for this change over the previous Query<T extends BaseRecord>

samples/extension-module/dummy/DummyApi.ts
```diff
@@ -1,3 +1,4 @@
+import type { Query } from '../../../../aries-framework-javascript/packages/core/src/storage/StorageService'
```
genaris, line 1: This import should not depend on the directory the repo is cloned

samples/extension-module/dummy/services/DummyService.ts
```diff
@@ -1,3 +1,4 @@
+import type { Query } from '../../../../../aries-framework-javascript/packages/core/src/storage/StorageService'
```
genaris, line 1: This import should not depend on the directory the repo is cloned. Ideally, it should be imported directly from @aries-framework/core:

```suggestion
import type { Query } from '@aries-framework-javascript/core'
```

For that purpose, you will need to expose `Query` in **packages/core/src/index.ts**: change line 20 to `export { StorageService, Query } from './storage/StorageService'`

genaris (changes requested):
Looks pretty good and I actually approved it but noticed some remaining issues afterwards.

samples/extension-module/dummy/DummyApi.ts
rapaktech: On it

samples/extension-module/dummy/services/DummyService.ts
rapaktech: On it

packages/core/src/storage/StorageService.ts
rapaktech: I was getting type error when trying to set the type of the query parameter to Query<ConnectionRecord>, Query<CredentialRecord> etc

genaris (approved):

PR #1047, opened 2022-10-06
fix(oob): set connection alias when creating invitation

jakubkoci:
It was possible to set an alias when creating an invitation but it wasn't propagated to the connection created after receiving a connection request based on the invitation.


packages/core/src/storage/migration/__tests__/__snapshots__/0.1.test.ts.snap
```diff
@@ -3132,443 +3132,3 @@ Object {
   },
 }
 `;
-
-exports[`UpdateAssistant | v0.1 - v0.2 should correctly update the role in the mediation record: allMediator 1`] = `
```
TimoGlastra, line 3136: There's some snapshots being removed

TimoGlastra (commented):
Nice! Isn't this a fix instead of a feature?

jakubkoci:
> Nice! Isn't this a fix instead of a feature?

Yeah, maybe 🤷‍♂️ I'll change it.

TimoGlastra (approved):

TimoGlastra:
Your last commit is not DCO signed off... :( Can you fix, then we can merge

PR #1049, opened 2022-10-11
feat(action-menu)!: move to separate package

genaris:

BREAKING CHANGE: action-menu module has been removed from the core and moved to a separate package. To integrate it in an Agent instance, it can be injected in constructor like this:
```ts
const agent = new Agent({
  config: { /* config */ },
  modules: {
    actionMenu: new ActionMenuModule(),
    /* other custom modules */
   }
})
```

Then, module API can be accessed in `agent.modules.actionMenu`.


TimoGlastra (approved):

packages/action-menu/src/__tests__/utils.ts
```diff
@@ -0,0 +1,72 @@
...
+  })
+}
+
+export function getAgentConfig(name: string, extraConfig: Partial<InitConfig> = {}) {
+  const { config, agentDependencies } = getBaseConfig(name, extraConfig)
+  return new AgentConfig(config, agentDependencies)
+}
+
+const publicDidSeed = process.env.TEST_AGENT_PUBLIC_DID_SEED ?? '000000000000000000000000Trustee9'
+export function getBaseConfig(name: string, extraConfig: Partial<InitConfig> = {}) {
```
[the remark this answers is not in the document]
genaris: Agree. Actually it seems that it was a copy-paste issue from question-answer, as I'm not using it. :-) Will update it in question-answer as well.

packages/action-menu/package.json
```diff
@@ -0,0 +1,40 @@
...
+    "test": "jest"
+  },
+  "dependencies": {
+    "@aries-framework/core": "0.2.4",
+    "rxjs": "^7.2.0",
+    "class-transformer": "0.5.1",
+    "class-validator": "0.13.1"
+  },
+  "devDependencies": {
+    "@aries-framework/core": "0.2.4",
```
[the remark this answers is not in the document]
genaris: Yes. And regarding the version number, should it always match the version of AFJ core/node/react-native?

For the moment it's 0.2.4 but I guess we'll need to update to 0.3.0 for all packages in a further PR before merging to main.

packages/action-menu/README.md
```diff
@@ -6,7 +6,7 @@
     height="250px"
```
blu3beri, line 6: Not as a comment on your work, but we should probably export the dummyModule so other people can get started aswell :).

TimoGlastra: What do you mean exactly by this @blu3beri?

blu3beri: We have a dummyModule in AFJ explaining to people on how to create an extension module, the old way. If we update that everyone has a clear example to start modularising other modules and their custom modules that they might use.

TimoGlastra: Ah gotcha! Yes totally agree

PR #1050, opened 2022-10-12
refactor: fix inconsistencies in issue credential and present proof api

NB-MikeRichardson:


packages/core/src/modules/credentials/protocol/v2/__tests__/v2-credentials.e2e.test.ts
```diff
@@ -215,7 +215,7 @@ describe('v2 credentials', () => {
     })
   })
 
-  test('Faber issues credential which is then deleted from Alice`s wallet', async () => {
+  xtest('Faber issues credential which is then deleted from Alice`s wallet', async () => {
```
TimoGlastra, line 218: Why are these disabled?

packages/core/src/modules/proofs/formats/indy/IndyProofFormatsServiceOptions.ts
```diff
@@ -24,12 +26,6 @@ export interface IndyVerifyProofFormat {
   proofRequest: Attachment
 }
 
-export interface IndyPresentationProofFormat {
```
TimoGlastra, line 27: In the `IndyRequestedCredentialsFormat` these are all required, while they were optional in this interface. Is this going to lead to issues?

packages/core/src/modules/proofs/protocol/v2/V2ProofService.ts
```diff
@@ -768,6 +771,82 @@ export class V2ProofService<PFs extends ProofFormat[] = ProofFormat[]> extends P
...
+    // way so we don't have to add the same operation code four times
+    const messages = {
+      proposal: [proposalMessage?.formats, proposalMessage?.proposalsAttach],
+      request: [requestMessage?.formats, requestMessage?.requestPresentationsAttach],
+      presentation: [presentationMessage?.formats, presentationMessage?.presentationsAttach],
+    } as const
+
+    const formatData: GetFormatDataReturn = {}
+
+    // // We loop through all of the message keys as defined above
```
TimoGlastra, line 792:
```suggestion
    // We loop through all of the message keys as defined above
```

packages/core/src/modules/proofs/protocol/v1/V1ProofService.ts
```diff
@@ -996,6 +997,36 @@ export class V1ProofService extends ProofService<[IndyProofFormat]> {
...
+      this.findRequestMessage(agentContext, proofRecordId),
+      this.findPresentationMessage(agentContext, proofRecordId),
+    ])
+
+    return {
+      proposalAttributes: proposalMessage?.presentationProposal?.attributes,
+      proposalPredicates: proposalMessage?.presentationProposal?.predicates,
+      proposal: proposalMessage
+        ? {
+            indy: proposalMessage,
```
TimoGlastra, line 1015: We should return the attachment data here rather than the messages itself. As you can see in the credentials implementation, we extract the attachment data: https://github.com/hyperledger/aries-framework-javascript/blob/main/packages/core/src/modules/credentials/protocol/v1/V1CredentialService.ts#L1085-L1116

Also, for the proposal message we should transform the message to an RFC0592 compliant proposal message, something like this, but for proofs: https://github.com/hyperledger/aries-framework-javascript/blob/main/packages/core/src/modules/credentials/protocol/v1/V1CredentialService.ts#L1085-L1087

packages/core/src/modules/credentials/protocol/v2/__tests__/v2-credentials.e2e.test.ts
NB-MikeRichardson: Oops. Disabled during testing, must have forgot to turn them back on

packages/core/src/modules/proofs/formats/indy/IndyProofFormatsServiceOptions.ts
NB-MikeRichardson: I saw that and checked it, my thinking was that it was ok to use the same interface.

packages/core/src/modules/proofs/protocol/v1/V1ProofService.ts
NB-MikeRichardson: yes good spot. I will do that conversion in conjunction with the comment above.

packages/core/src/modules/proofs/formats/indy/IndyProofFormat.ts
```diff
@@ -59,19 +59,10 @@ export interface IndyProofFormat extends ProofFormat {
...
-  //   offer: CredOffer
-  //   request: CredReq
-  //   credential: Cred
-  // }
+
+  formatData: {
+    proposal: IndyProposeProofFormat
+    request: IndyRequestProofFormat
+    presentation: IndyPresentationProofFormat
+  }
```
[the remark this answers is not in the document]
NB-MikeRichardson: There is a Json class in our code, ProofRequest, and one in the Indy SDK, IndyProof. These seem to do what we need. 

(They are the 0592 classes)

packages/core/src/modules/proofs/protocol/v1/V1ProofService.ts
```diff
@@ -996,6 +997,38 @@ export class V1ProofService extends ProofService<[IndyProofFormat]> {
...
+  ): Promise<GetFormatDataReturn<ProofFormat[]>> {
+    const [proposalMessage, requestMessage, presentationMessage] = await Promise.all([
+      this.findProposalMessage(agentContext, proofRecordId),
+      this.findRequestMessage(agentContext, proofRecordId),
+      this.findPresentationMessage(agentContext, proofRecordId),
+    ])
+
+    const indyProposeProof = proposalMessage?.indyProofProposal ?? undefined
+    const indyRequestProof = requestMessage?.indyProofRequest ?? undefined
+    const indyPresentProof = presentationMessage?.indyProof ?? undefined
```
TimoGlastra, line 1012: For the credentials API we actually return the raw JSON attachment data, without transforming it to a class instance. These methods do the JSON transform on it

packages/core/tests/v1-indy-proofs.test.ts
```diff
@@ -252,6 +252,54 @@ describe('Present Proof', () => {
...
+              threshold: 50,
+            },
+          ],
+        },
+      },
+      request: {
+        indy: {
+          name: 'Proof Request',
+          version: '1.0',
+          nonce: expect.any(String),
```
TimoGlastra, line 292: This doesn't include the whole proof request. I think it would be good to check that here, to make sure the casing is correct (it should be snake_case here as that's what indy uses natively)

packages/core/tests/v1-indy-proofs.test.ts
```diff
@@ -252,6 +252,54 @@ describe('Present Proof', () => {
...
+              name: 'image_0',
+            },
+          ],
+          requestedPredicates: [
+            {
+              name: 'age',
+              predicate: '>=',
+              threshold: 50,
+            },
+          ],
```
TimoGlastra, line 285: This should match the proposal format as described in RFC 0592

packages/core/tests/v2-indy-proofs.test.ts
```diff
@@ -243,6 +254,117 @@ describe('Present Proof', () => {
...
+      'requested_attributes' as keyof IndyProofProposal
+    ] as PresentationPreviewAttribute[]
+
+    const requestedPredicates: PresentationPreviewAttribute[] = format[
+      'requested_predicates' as keyof IndyProofProposal
+    ] as PresentationPreviewAttribute[]
+
+    // this key is dynamically generated
+    const key1 = Object.keys(requestedAttributes)[1]
+    const key2 = Object.keys(requestedPredicates)[0]
```
TimoGlastra, line 282: Is this an UUID? Or is it consistent across re-runs of the test?

packages/core/tests/v2-indy-proofs.test.ts
```diff
@@ -243,6 +254,117 @@ describe('Present Proof', () => {
...
+              p_type: '>=',
+              p_value: 50,
+              restrictions: [
+                {
+                  cred_def_id: credDefId,
+                },
+              ],
+            },
+          },
+        },
```
TimoGlastra, line 315: The v1 and v2 model should return the same structure (as defined in RFC 0592)

packages/core/src/modules/proofs/protocol/v1/messages/V1RequestPresentationMessage.ts
```diff
@@ -65,17 +67,25 @@ export class V1RequestPresentationMessage extends AgentMessage {
...
   public requestPresentationAttachments!: Attachment[]
 
-  public get indyProofRequest(): ProofRequest | null {
+  public get indyProofRequest(): IndyProofRequest | null {
     const attachment = this.requestPresentationAttachments.find(
       (attachment) => attachment.id === INDY_PROOF_REQUEST_ATTACHMENT_ID
     )
     // Extract proof request from attachment
-    const proofRequestJson = attachment?.getDataAsJson<ProofRequest>() ?? null
+    const proofRequestJson = attachment?.getDataAsJson<IndyProofRequest>() ?? null
```
TimoGlastra, line 75: We can now replace this with `const proofRequestJson = this.indyProofRequestJson`

packages/core/src/modules/proofs/formats/indy/models/ProofProposal.ts
```diff
@@ -0,0 +1,76 @@
+import { Expose, Type } from 'class-transformer'
```
TimoGlastra, line 1: As mentioned in other comment, this file and its contents are not needed

packages/core/src/modules/proofs/formats/indy/IndyProofFormat.ts
```diff
@@ -59,19 +63,10 @@ export interface IndyProofFormat extends ProofFormat {
...
-  //     cred_def_id?: string
-  //   }
-  //   offer: CredOffer
-  //   request: CredReq
-  //   credential: Cred
-  // }
+
+  formatData: {
+    proposal: IndyProofProposal
+    request: IndyProofRequest
```
TimoGlastra, line 69: This should both be `IndyProofRequest` imported form the `@types/indy-sdk` package (so `import type { IndyProofRequest } from 'indy-sdk'`, not the class instance (as we're returning the json)

TimoGlastra (changes requested):
There's some misalignments between what you implemented and what the RFCs describe. I've tried to add detailed comments on where I think we should make changes to align with the RFCs

packages/core/src/modules/proofs/protocol/v1/V1ProofService.ts
```diff
@@ -996,6 +999,50 @@ export class V1ProofService extends ProofService<[IndyProofFormat]> {
...
+      presentation: presentationMessage
+        ? {
+            indy: indyPresentProof,
+          }
+        : undefined,
+    }
+  }
+
+  private rfc0592ProposalFromV1ProposeMessage(proposalMessage: V1ProposePresentationMessage) {
+    const options: ProofProposalOptions = {
```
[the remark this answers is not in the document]
NB-MikeRichardson: Ok this seems to work fine, the only problem is that, for some reason, that utility class expects an options object that contains connectionId, which is not required by that method.

Can I make this optional?

packages/core/src/modules/proofs/protocol/v1/messages/V1RequestPresentationMessage.ts
```diff
@@ -12,6 +12,8 @@ import { V2_INDY_PRESENTATION_REQUEST } from '../../../formats/ProofFormatConsta
 import { ProofRequest } from '../../../formats/indy/models/ProofRequest'
 import { ProofFormatSpec } from '../../../models/ProofFormatSpec'
 
+export type IndyProofRequest = ProofRequest
```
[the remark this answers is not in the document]
NB-MikeRichardson: The problem here is a name clash, as there is a class named IndyProofRequest in the SDK. 

I'll leave it for now, pending suggestions.

packages/core/src/modules/proofs/protocol/v1/V1ProofService.ts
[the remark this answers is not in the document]
TimoGlastra: I think you can just use the `IndyProofUtils .createReferentForProofRequest` method directly

packages/core/tests/helpers.ts
```diff
@@ -818,3 +820,55 @@ export async function setupProofsTest(faberName: string, aliceName: string, auto
...
+  requestKey1 = Object.keys(attributes)[1]
+
+  const predicates: Map<string, ProofAttributeInfo> = proofRequest[
+    'requested_predicates' as keyof IndyProofRequest
+  ] as Map<string, ProofAttributeInfo>
+
+  // eslint-disable-next-line prefer-const
+  requestKey2 = Object.keys(predicates)[0]
+
+  return { proposeKey1, proposeKey2, requestKey1, requestKey2 }
```
TimoGlastra, line 873: This can be simplified quite a bit, however I'm not sure if this can go as a generic test util. You access _very_ specific indexes in the proof request, meaning you can't really reuse this in other tests -- only between the v1 and v2 get format data return test. So maybe it should be hosted somewhere in the proof module tests folder. Just want to make sure we're not cluttering this file with methods that aren't actually reusable.

```suggestion
  const proofRequest = formatData.request?.indy
  const proofProposal = formatData.proposal?.indy

  if (!proofProposal) {
    throw new AriesFrameworkError('missing indy proof proposal')
  }
  if (!proofRequest) {
    throw new AriesFrameworkError('missing indy proof request')
  }

  const proposeAttributes = proofProposal.requested_attributes
  const proposePredicates = proofProposal.requested_predicates
  const requestAttributes = proofRequest.requested_attributes
  const requestPredicates = proofRequest.requested_predicates

  const proposeKey1 = Object.keys(proposeAttributes)[1]
  const proposeKey2 = Object.keys(proposePredicates)[0]

  const requestKey1 = Object.keys(requestAttributes)[1]
  const requestKey2 = Object.keys(requestPredicates)[0]

  return { proposeKey1, proposeKey2, requestKey1, requestKey2 }
```

packages/core/src/modules/proofs/models/ProofServiceOptions.ts
```diff
@@ -19,7 +19,7 @@ interface BaseOptions {
 }
 
 export interface CreateProposalOptions<PFs extends ProofFormat[]> extends BaseOptions {
-  connectionRecord: ConnectionRecord
+  connectionRecord?: ConnectionRecord
```
TimoGlastra: Currently all proposals require a connection, so we shouldn't make it optional here. 

```suggestion
  connectionRecord: ConnectionRecord
```

packages/core/src/modules/proofs/protocol/v1/V1ProofService.ts
```diff
@@ -123,6 +122,9 @@ export class V1ProofService extends ProofService<[IndyProofFormat]> {
   ): Promise<{ proofRecord: ProofRecord; message: AgentMessage }> {
     const { connectionRecord, proofFormats } = options
 
+    if (!connectionRecord) {
+      throw new AriesFrameworkError('Missing Connection Record in CreateProposalOptions')
+    }
```
TimoGlastra: We should keep the connectionRecord required in the CreateProposalOptions

```suggestion
```

packages/core/src/modules/proofs/protocol/v2/V2ProofService.ts
```diff
@@ -109,6 +109,9 @@ export class V2ProofService<PFs extends ProofFormat[] = ProofFormat[]> extends P
       parentThreadId: options.parentThreadId,
     })
 
+    if (!options.connectionRecord) {
+      throw new AriesFrameworkError('Missing connection record')
+    }
```
TimoGlastra: Same here

```suggestion
```

packages/core/src/modules/proofs/protocol/v1/V1ProofService.ts
```diff
@@ -1034,14 +1036,30 @@ export class V1ProofService extends ProofService<[IndyProofFormat]> {
...
+
+    const preview = new PresentationPreview({
+      attributes: indyFormat.attributes,
+      predicates: indyFormat.predicates,
+    })
+
+    if (!preview) {
+      throw new AriesFrameworkError(`No preview found`)
+    }
```
TimoGlastra, line 1058: This can't be undefined

```suggestion
```

packages/core/tests/helpers.ts
```diff
@@ -15,9 +15,11 @@ import type {
 import type { AgentModulesInput } from '../src/agent/AgentModules'
 import type { IndyOfferCredentialFormat } from '../src/modules/credentials/formats/indy/IndyCredentialFormat'
 import type { RequestProofOptions } from '../src/modules/proofs/ProofsApiOptions'
+import type { IndyProofFormat } from '../src/modules/proofs/formats/indy/IndyProofFormat'
 import type { ProofAttributeInfo, ProofPredicateInfo } from '../src/modules/proofs/formats/indy/models'
 import type { AutoAcceptProof } from '../src/modules/proofs/models/ProofAutoAcceptType'
-import type { CredDef, Schema } from 'indy-sdk'
+import type { GetFormatDataReturn } from '../src/modules/proofs/models/ProofServiceOptions'
+import type { CredDef, IndyProofRequest, Schema } from 'indy-sdk'
```
TimoGlastra, line 22: I think these changes can be reverted now?

TimoGlastra (approved):

PR #1057, opened 2022-10-20
chore: merge 0.3.0-pre into main branch

TimoGlastra:
This merges the 0.3.0-pre into the main branch. 

NOTE: we should use rebase when merging so all the original commits are preserved for attribution and changelogs.

Theres still some PRs to be made, but this means we can start testing the 0.3.0-alpha release and discover if there are any issues with new/existing features


genaris (approved):
Good to see this finally going to main branch. Congrats for the clean merge!

Now I see everything in place, I have a few doubts regarding the packages versioning: I guess all packages will move from 0.2.5 to 0.3.0 as soon as **./scripts/get-next-bump.ts** does his job. However, what would happen when either minor or major changes are introduced in a module? For instance, suppose we introduce a breaking change in action-menu. That would mean that core, react-native, node, etc. will also have a 'version bump only' release? It seems to me a bit awkward but it would be more difficult to have different versioning scheme for each package within this monorepo.

An issue I've recently noticed is that, even if our 'set-DCO-to-pass and squash' theory worked pretty well, DCO is failing again... this time for other commits than the ones that caused trouble in the past. As this time we'll not squash, I'm afraid that they will still be failing once merged to main.

TimoGlastra:
> That would mean that core, react-native, node, etc. will also have a 'version bump only' release?

Yes, for now we keep the versions in sync. This way it'll be easier to keep all packages in sync and working with each other (must use the same version for all). What we can look at at some point is creating package groups, where all packages in a group will be versions togeteher. Modules can be versioned separately, but the core,node,react native  packages will still stay in sync.

> It seems to me a bit awkward but it would be more difficult to have different versioning scheme for each package within this monorepo.

Is this a question or a statement? 😬

PR #1071, opened 2022-10-25
refactor(proofs)!: rename ProofRecord to ProofExchangeRecord

NB-MikeRichardson:


packages/core/src/storage/migration/__tests__/__snapshots__/0.2.test.ts.snap
```diff
@@ -165,7 +165,7 @@ Object {
       "state": "done",
       "threadId": "7fcfc074-43ac-43cc-92b9-76afceeebe82",
     },
-    "type": "ProofRecord",
+    "type": "ProofExchangeRecord",
```
TimoGlastra, line 168: We should keep the type as 'ProofRecord' in the ProofExchangeRecord

packages/core/src/modules/proofs/repository/ProofExchangeRecord.ts
```diff
@@ -30,7 +30,7 @@ export type DefaultProofTags = {
 
 // T-TODO: rename to proof exchange record
```
TimoGlastra, line 31: Can you remove this comment?

TimoGlastra (commented):
Nice! Some small suggestions but otherwise LGTM!

genaris:
This refactoring would represent a breaking change, right? In that case I'll note it for the migration guide. Would it be merged as _refactor(proofs)!: Rename ProofRecord to ProofExchangeRecord_?

TimoGlastra:
@NB-MikeRichardson CI is failing it seemes

NB-MikeRichardson:
> @NB-MikeRichardson CI is failing it seems. I think you need to regenerate the snapshot files

I'll take a look

TimoGlastra (approved):

PR #1073, opened 2022-10-26
refactor(proofs)!: make nonce optional and add missing exports

genaris:
Some refactoring to solve minor issues found during migration to 0.3.0 on AATH:

- Expose several classes and interfaces from proofs module that were missing (especially for Present Proof V1) and are useful for API usage. This impacts a bit both credentials and discover-features, as there were a few clashes in types and interfaces which needed to be renamed.
- Make name, version and nonce optional for Indy format in Proofs API (e.g. **proposeProof** and **requestProof** methods). If not provided, the framework will generate a nonce by calling the underlying wallet's `generateNonce` method. This is the same behaviour we had in 0.2.x
- Remove `generateProofRequestNonce` method from `V1ProofService` and `ProofService` as it is Indy/AnonCreds-specific. It will call directly wallet's `generateNonce` method

After @TimoGlastra's precise and appropriate suggestion, I moved logic from `IndyProofUtils` to `IndyProofFormatService`, resulting in a cleaner `V2ProofService` (less Indy-specific stuff there).

It's marked as breaking change mainly because of the renaming of some types and interfaces. However it does not change any field besides the aforementioned 'nonce'.


blu3beri (approved):

TimoGlastra:
Discussed in AFJ WG call 27-10: make nonce optional in proof request, and generate it if not present in the indy proof format service.

genaris:
> Discussed in AFJ WG call 27-10: make nonce optional in proof request, and generate it if not present in the indy proof format service.

So I've finally converted this PR from a mere 'expose types PR' (the ones that @blu3beri likes the most :stuck_out_tongue: ) to a refactoring that impacts three modules :sweat_smile:. Not completely sure about the naming conventions I've selected, but happy to hear any feedback about it!

packages/core/src/modules/proofs/protocol/v2/V2ProofService.ts
```diff
@@ -87,21 +88,29 @@ export class V2ProofService<PFs extends ProofFormat[] = ProofFormat[]> extends P
...
+      let formats
+      if (key === PresentationRecordType.Indy) {
+        const indyFormat = (options as CreateProposalOptions<[IndyProofFormat]>).proofFormats.indy
+        if (!indyFormat) {
+          throw new AriesFrameworkError('No Indy format found.')
+        }
+        formats = await IndyProofUtils.createRequestFromPreview({
+          ...indyFormat,
+          nonce: indyFormat.nonce ?? (await this.wallet.generateNonce()),
         })
```
TimoGlastra, line 104: Maybe we can move the indy createRequestFromPreview logic to the indy proof format service? I know you didn't write this, but we should aim to leave format specific code out of the format service and it will make for more beatiful code I think

genaris: You are absolutely right! After analyzing it a bit more I think the same behaviour can be respected without that `if (key === PresentationRecordType.Indy)`, and just calling IndyProofUtils.createRequestFromPreview from `IndyProofFormatService`. 

Actually I think there is room for improvement but I didn't want to touch too much because I'm not so familiar with proofs code. At least `V2ProofService` seems cleaner now :smile: .

TimoGlastra (approved):

PR #1084, opened 2022-11-02
ci: use graph-type all for lerna publish

genaris:
CD is broken since #1082 has been merged, and it seems that it's because lerna is using alphabetical order when publishing packages (`lerna publish` command), so attempts to build action-menu before core (marked as peerDependency).

According to [Lerna Publish documentation](https://github.com/lerna/lerna/blob/main/commands/publish/README.md#--graph-type-alldependencies), this is because lerna uses by default a `dependencies` graph-type, which will take into account only packages listed in dependencies to determine the build order. As in this case it is listed in peerDependencies, an `all` graph-type must be used.

Another option of course would be to modify action-menu's package.json to set core as a direct dependency. However, this would be also needed for any other package that comes before to core in alphabetical order (such as `bbs-signatures`). I'm not sure if this would be conceptually correct (this was discussed recently [when extracting BBS module](https://github.com/hyperledger/aries-framework-javascript/pull/1035#discussion_r989228931)).

### DIFF
--- a/.github/workflows/continuous-deployment.yml
+++ b/.github/workflows/continuous-deployment.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Release Unstable
         run: |
           export NEXT_VERSION_BUMP=$(./node_modules/.bin/ts-node ./scripts/get-next-bump.ts)
-          yarn lerna publish --loglevel=verbose --canary $NEXT_VERSION_BUMP --exact --force-publish --yes --no-verify-access --dist-tag alpha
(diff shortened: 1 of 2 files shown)


TimoGlastra:
I see no harm in using dependencies over peer dependencies. For this repo all versions **must** be the same, so there's not really an advantage in using peer deps over deps. If we in the future start using separate versions for packages in this repo peer deps make sense. I'm impartial on this discussion.

If we go with the approach in this PR we should also update it for lerna publish for stable release (maybe even add it to lerna.json so we can't publish without)

TimoGlastra (approved):

PR #1092, opened 2022-11-09
feat: issue credentials v2 (W3C/JSON-LD)

NB-MikeRichardson:


TimoGlastra:
@NB-MikeRichardson seems like tests are still failing. Can't merge before those are resolved...

NB-MikeRichardson:
> 

The failing test only seems to fail in 18.x. I have run this test on my local machine for all versions of node, and it runs fine except in 18.x. It is related to the Bbs signatures it seems.

How does one go about fixing this test?

genaris:
> @NB-MikeRichardson seems like tests are still failing. Can't merge before those are resolved...

It's the problem with BBS libraries. How can we avoid running them in node 18?

NB-MikeRichardson:
> > @NB-MikeRichardson seems like tests are still failing. Can't merge before those are resolved...
> 
> It's the problem with BBS libraries. How can we avoid running them in node 18?

Strangely enough I can get it running (sometimes) on my local machine. It seems to be inconsistent, do we know what the problem is with BBS libraries in 18.x?

TimoGlastra:
@NB-MikeRichardson for all tests that include BBS code you should use `describeSkipNode17And18` instead of `describe` at the top of the test file. This will skip the tests in Node 17 and 18

TimoGlastra (approved):

PR #1093, opened 2022-11-09
fix(routing): async message pickup on init

genaris:
Currently, when initializing the agent using a `MediationPickupStrategy` different from `None` (e.g. `PickupV1` or `PickupV2`), it attempts to start message pickup in a synchronous way, meaning that it might take some more time for startup depending on the . Also, if mediator is not reachable at that moment, it will throw an error and Agent's `_isInitialized` flag will not be set to true even if can connect afterwards (with the polling of PickupV1 or backoff strategy of PickupV2).

In this PR we are starting the message pickup process in an async way and catching any possible initial error, in a similar way as Agent's initialize method for connecting to ledger pools.

### DIFF
--- a/packages/core/src/modules/routing/RecipientApi.ts
+++ b/packages/core/src/modules/routing/RecipientApi.ts
@@ -101,7 +101,9 @@ export class RecipientApi {
     // Poll for messages from mediator
     const defaultMediator = await this.findDefaultMediator()
     if (defaultMediator) {
-      await this.initiateMessagePickup(defaultMediator)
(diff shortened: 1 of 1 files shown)


TimoGlastra (approved):

genaris:
> Makes sense +1 I've currently handled it by setting it to None and starting pickup manually afterwards. I think handling this better by default means less custom code and if you need more control you can just set it to None and handle it yourself

Yes, actually I noticed this while writing the tutorial for message pickup that I owe since several weeks :sweat_smile:. There will be the suggestion of using None and handle the cycle manually depending on network status.

PR #1112, opened 2022-11-21
fix: remove sensitive information from agent config toJSON() method

jimezesinachi:
Signed-off-by: Jim Ezesinachi <jim@animo.id>

### DIFF
--- a/packages/core/src/agent/AgentConfig.ts
+++ b/packages/core/src/agent/AgentConfig.ts
@@ -193,10 +193,10 @@ export class AgentConfig {
       ...this.initConfig,
       walletConfig: {
         ...this.walletConfig,
-        key: this.walletConfig?.key ? '[omitted...]' : undefined,
+        key: this.walletConfig?.key ? '[*****]' : undefined,
         storage: {
           ...this.walletConfig?.storage,
-          credentials: this.walletConfig?.storage?.credentials ? '[omitted...]' : undefined,
+          credentials: this.walletConfig?.storage?.credentials ? '[*****]' : undefined,
         },
       },
       logger: this.logger.logLevel,
--- a/packages/core/src/wallet/WalletApi.ts
+++ b/packages/core/src/wallet/WalletApi.ts
@@ -45,10 +45,10 @@ export class WalletApi {
   public async initialize(walletConfig: WalletConfig): Promise<void> {
     this.logger.info(`Initializing wallet '${walletConfig.id}'`, {
       ...walletConfig,
-      key: walletConfig?.key ? '[omitted...]' : undefined,
+      key: walletConfig?.key ? '[*****]' : undefined,
       storage: {
         ...walletConfig?.storage,
-        credentials: walletConfig?.storage?.credentials ? '[omitted...]' : undefined,
+        credentials: walletConfig?.storage?.credentials ? '[*****]' : undefined,
       },
     })
 


packages/core/src/agent/AgentConfig.ts
```diff
@@ -191,8 +191,13 @@ export class AgentConfig {
...
     return {
       ...this.initConfig,
-      logger: this.logger !== undefined,
-      agentDependencies: this.agentDependencies != undefined,
+      walletConfig: {
+        ...this.walletConfig,
+        key: undefined,
+        storage: { ...this.walletConfig?.storage, credentials: undefined },
+      },
+      logger: Boolean(this.logger), // Maybe show the loglevel here instead of true / false?
```
blu3beri, line 199: I think we can just log the level here. `this.logger.logLevel` I think.

blu3beri (changes requested):

rapaktech:
All checks passed! @blu3beri

packages/core/src/agent/AgentConfig.ts
```diff
@@ -191,8 +191,13 @@ export class AgentConfig {
...
       ...this.initConfig,
-      logger: this.logger !== undefined,
-      agentDependencies: this.agentDependencies != undefined,
+      walletConfig: {
+        ...this.walletConfig,
+        key: undefined,
+        storage: { ...this.walletConfig?.storage, credentials: undefined },
+      },
+      logger: this.logger.logLevel,
+      agentDependencies: Boolean(this.agentDependencies),
```
blu3beri, line 200:
```suggestion
      walletConfig: {
        ...this.walletConfig,
        key: key ? "[omitted...]" : undefined,
        storage: { ...this.walletConfig?.storage, credentials: this.walletConfig?.credentials ? "[omitted...]" : undefined },
      },
      logger: this.logger.logLevel,
      agentDependencies: Boolean(this.agentDependencies),
```

blu3beri:
@rapaktech Prettier is failing now :). Could you fix that? I can merge it afterwards.

blu3beri:
@TimoGlastra would this be considered a breaking change? The return type of the function is quite different now, but I don't think it is a "public method"

TimoGlastra (commented):
I wouldn't call this a breaking change, if you need properties from the agent config you should do so using the class methods. 

This LGTM @rapaktech. I just found another place where we log keys, could you also tackle that one in this PR? It's in the `WalletApi.initialize()` method (`Initializing wallet xxx`)

BTW -- marking this as a fix over chore, as we shouldn't have logged the sensitive info in the first place

rapaktech:
You got it, Timo. On it.

packages/core/src/agent/AgentConfig.ts
```diff
@@ -191,8 +191,16 @@ export class AgentConfig {
...
     return {
       ...this.initConfig,
-      logger: this.logger !== undefined,
-      agentDependencies: this.agentDependencies != undefined,
+      walletConfig: {
+        ...this.walletConfig,
+        key: this.walletConfig?.key ? '[omitted...]' : undefined,
+        storage: {
+          ...this.walletConfig?.storage,
+          credentials: this.walletConfig?.storage?.credentials ? '[omitted...]' : undefined,
```
jakubkoci, line 199: This is great and long awaited addition 👍 

Just a naming idea... wouldn't it be better to use rather `[hidden...]` or, quite common use of `*****` (I saw it in Github Action I think)

rapaktech: 👍🏾

PR #1116, opened 2022-11-23
fix(proof): removed

blu3beri:


blu3beri:
@TimoGlastra Would it be possible to get this in 0.3.0? I assume if it goes to main now it will be in 0.3.0, right?

genaris:
> @TimoGlastra Would it be possible to get this in 0.3.0? I assume if it goes to main now it will be in 0.3.0, right?

We are working on 0.3.0.alpha.xxx, so every commit will be included to the 0.3.0 release. A new PR will be created later on (I think Timo created the current release PR just to help me on the migration guide, as it generates the changelog).

TimoGlastra:
> @TimoGlastra Would it make sense to add a rule that you must mark void/await for async calls? that would've probably avoided this, quite big, bug.

Yes I think we should add [no-floating-promises](https://typescript-eslint.io/rules/no-floating-promises/) and [no-misused-promises](https://typescript-eslint.io/rules/no-misused-promises/)

blu3beri:
Apparently the tests were written for the incorrect way of doing autoAccept. Will fix this ASAP tomorrow.

blu3beri:
Sorry that it took so long, the bug was a bit bigger than just a missing "await". Our deep equality check did not work, and I had to change it. (also added some more tests to be sure).

PR #1118, opened 2022-11-24
feat: add anoncreds package

karimStekelenburg:
This PR adds a package that contains a set of interfaces that define unified APIs for AnonCreds related logic.

The goal of adding this package is that we can have two separate implementations, one for `indy-sdk` and another one for the `anoncreds-rs` Rust library that is currently being developed.

> **Note**: the interfaces in this package contain a lot of `todo` comments and `any` statements. The idea is to update these during the implementation of the first package.


swcurran:
Just as a matter of interest, why the "-" in "anon-creds"?

TimoGlastra:
> Just as a matter of interest, why the "-" in "anon-creds"?

Agree, I think we should just call it `anoncreds`

TimoGlastra (commented):
Overall looks good, I'll make some amendments on the interfaces with the missing types. Can't comment on the file, but internal.ts is an empty file. I think we can remove it for now?


For others reviewing this PR. Some notes/todos that we would like input on:
- Do we want to follow data model terms (credDefId) or AFJ terms (credentialDefinitionId) in the interfaces. E.g. for the createSchema method it should always return the data model. However when we call createOffer we need to pass a cred def id. This is not part of the data model, so in this case should we use credentialDefinitionId?

packages/anon-creds/src/models/exchange.ts
```diff
@@ -0,0 +1,104 @@
+// TODO: Maybe we can make this a bit more specific?
+export type WalletQuery = Record<string, unknown>
+
+export interface NonRevokedInterval {
+  from?: number | undefined
+  to?: number | undefined
```
blu3beri, line 6:
```suggestion
  from?: number
  to?: number
```

packages/anon-creds/src/models/exchange.ts
```diff
@@ -0,0 +1,104 @@
...
+}
+
+export interface AnonCredsCredentialOffer {
+  schema_id: string
+  cred_def_id: string
+  nonce: string
+  key_correctness_proof: Record<string, unknown>
```
blu3beri, line 13: Would it be possible, or maybe later, to make this more explicit? If we can type the fields in here it is always a nice benefit. We can also pick this up later and just create and issue now if we want this.

blu3beri: This comment basically applies to any `Record<string, unknown>`

packages/anon-creds/src/models/exchange.ts
```diff
@@ -0,0 +1,104 @@
...
+
+export interface AnonCredsProof {
+  requested_proof: {
+    revealed_attrs: {
+      [key: string]: {
+        sub_proof_index: number
+        raw: string
+        encoded: string
+      }
+    }
```
blu3beri, line 48:
```suggestion
    revealed_attrs: Record<string, {
        sub_proof_index: number
        raw: string
        encoded: string
      }>
    
```

Take it or leave it. Might be a bit more consistent with the other `Record<string, ...>` types.

packages/anon-creds/src/models/exchange.ts
```diff
@@ -0,0 +1,104 @@
...
+    self_attested_attrs: {
+      [key: string]: string
+    }
+    requested_predicates: {
+      [key: string]: { sub_proof_index: number }
+    }
+  }
+  proof: any
```
blu3beri, line 72: would `unknown` suffice here?

packages/anon-creds/src/models/exchange.ts
```diff
@@ -0,0 +1,104 @@
...
+      non_revoked?: NonRevokedInterval | undefined
+    }
+  }
+  requested_predicates: {
+    [key: string]: {
+      name: string
+      p_type: '>=' | '>' | '<=' | '<'
+      p_value: number
+      restrictions?: WalletQuery[] | undefined
+      non_revoked?: NonRevokedInterval | undefined
```
blu3beri, line 99:
```suggestion
      restrictions?: WalletQuery[]
      non_revoked?: NonRevokedInterval
```

packages/anon-creds/src/models/exchange.ts
```diff
@@ -0,0 +1,104 @@
...
+    }
+  }
+  non_revoked?: NonRevokedInterval | undefined
+  ver?: '1.0' | '2.0' | undefined
```
blu3beri, line 103:
```suggestion
  non_revoked?: NonRevokedInterval
  ver?: '1.0' | '2.0'
```

packages/anon-creds/src/models/exchange.ts
```diff
@@ -0,0 +1,104 @@
...
+export interface AnonCredsProofRequest {
+  name: string
+  version: string
+  nonce: string
+  requested_attributes: {
+    [key: string]: {
+      name?: string | undefined
+      names?: string | undefined
+      restrictions?: WalletQuery[] | undefined
+      non_revoked?: NonRevokedInterval | undefined
```
blu3beri, line 90:
```suggestion
      name?: string
      names?: string
      restrictions?: WalletQuery[]
      non_revoked?: NonRevokedInterval
```

packages/anon-creds/src/models/exchange.ts
```diff
@@ -0,0 +1,104 @@
+// TODO: Maybe we can make this a bit more specific?
+export type WalletQuery = Record<string, unknown>
+
+export interface NonRevokedInterval {
+  from?: number | undefined
+  to?: number | undefined
+}
+
+export interface AnonCredsCredentialOffer {
+  schema_id: string
```
blu3beri, line 10: Very very small knowledge on the spec and the changes, but is it not `schemaId` now? very unsure, but I thought I remembered it from somewhere...

packages/anon-creds/src/AnonCredsApiOptions.ts
```diff
@@ -0,0 +1,16 @@
+import type { AnonCredsSchema } from './models/vdr'
+
+// FIXME: we should update the name in the AnonCreds registry interface to include extra wording in the name to prevent name clashes.
+export interface RegisterSchemaApiOptions {
+  schema: AnonCredsSchema
+
+  // Identifier of the identifier that will create the schema. In most cases a did, but can also be an URI.
+  //   TODO: do we want an issuerId? Or should we add such properties to an options object that is different per method
+  // more like the did registration interface?
+  issuerId: string
```
blu3beri, line 10: Do we want to create a specific type for a DID or URI? something like `did:${string}:${string}` or would that just be an annoyance and not very useful?

blu3beri (dismissed):
Left a couple of comments, mainly some cleanup and consistency.

blu3beri:
> Just as a matter of interest, why the "-" in "anon-creds"?

I think the reason was as it is AnonCreds, which is in pascal case. If you transform pascal case to kebab case, which is what we use for packages and folders, you get `anon-creds`.

I prefer simply `anoncreds`, but I think that this is the reason for that choice.

karimStekelenburg:
The reasoning behind the "-" is simple. AnonCreds is an acronym for Anonymous Credentials, so separating those words made sense to me.

I'm happy to change it though, consistency is more important.

@swcurran any ideas on what the rationale is behind not separating those words is? I've been wondering the same thing about the "cloudagent" in Aries Cloudagent Python for years 😅.

swcurran:
Naming is fun!

AnonCreds is the term that has been used in the community for year, and the project at Hyperledger is "AnonCreds", so I think if you are using it in code, it should be "anoncreds".  And yes, AnonCreds was originally short for "Anonymous Credentials", but the name has a life of its own, beyond just an acronym.  I've never seen it hyphenated before, hence my comment.

Cloudagent was just what was settled on, before all the rest of the "Aries things" were called Frameworks.  Had we named it a few months later, it probably would have been called "Aries Framework Python".  The cloudagent was to indicate that it was not a suitable candidate for mobile deployment.  In other words, nothing special -- just a name that someone came up with and the rest of us thought was less terrible than the other options we had come up with.

karimStekelenburg:
@TimoGlastra, agree with your points. Maybe I should have made it a draft as it's not finalised yet. I just copied what you defined and planned on updating it during the indy-sdk implementation.

I assumed you had a purpose in mind for `internals.ts`, that is why I kept it.

I thought the idea was to merge this package first, and then merge the indy-sdk implementation once finished. However, now I'm thinking about it that, I don't see a reason to really.

If you agree, I will make this PR a draft for now and start adding the indy-sdk implementation to it.

I can finalise the of the interfaces while implementing them for indy-sdk and merge them in all together. That way we keep main clean of unused, semi-implemented packages.

I guess the PPV2 situation scared me into shrinking down PRs a bit too much 😅

TimoGlastra:
> However, now I'm thinking about it that, I don't see a reason to really.

I think smaller PRs is a good reason! If we remove the API part and clean it up I'm for merging this first. It's a private package for now. That'll also allow e.g. cheqd work to already start

packages/anoncreds/src/services/AnonCredsVerifierServiceOptions.ts
```diff
@@ -0,0 +1,53 @@
...
+        [timestamp: string]: AnonCredsRevocationList
+      }
+    }
+  }
+}
+
+export type VerifyProofMetadata = {
+  schemas: {
+    [key: string]: {
+      ver: string
```
TimoGlastra, line 36: "ver" is always "1.0" for schema, revocationRegistryDefinition and credentialDefinition. Can we remove them and just hardcode them in the indy-sdk anoncreds interfaces?

packages/anoncreds/src/services/AnonCredsVerifierServiceOptions.ts
```diff
@@ -0,0 +1,53 @@
...
+    }
+  },
+  credentialDefinitions: {
+    [key: string]: {
+      ver: string
+    }
+  },
+  revocationRegistryDefinition: {
+    [key: string]: {
+      issuanceType: 'ISSUANCE_BY_DEFAULT' | 'ISSUANCE_ON_DEMAND',
```
TimoGlastra, line 47: issuanceType is removed from the spec, we don't need to pas it here. The revocation registry state will contain all indexes with 0 / 1 for whether they're revoked / unrevoked and therefore this field is not needed

packages/anoncreds/src/services/AnonCredsHolderServiceOptions.ts
```diff
@@ -0,0 +1,102 @@
...
+  credentialRequestMetadata: Record<string, unknown>
+  credential: AnonCredsCredential
+  credentialDefinition: AnonCredsCredentialDefinition
+  credentialId?: string
+  revocationRegistryDefinition?: AnonCredsRevocationRegistryDefinition
+}
+export type StoreCredentialMetadata = {
+  credentialDefinitionMetadata: {
+    id: string
+    ver: string
```
TimoGlastra, line 59:
```suggestion
```

packages/anoncreds/src/services/AnonCredsHolderServiceOptions.ts
```diff
@@ -0,0 +1,102 @@
...
+  }
+  revocationRegistryDefinition: {
+    id: string
+    issuanceType: 'ISSUANCE_BY_DEFAULT' | 'ISSUANCE_ON_DEMAND'
+    ver: string
```
TimoGlastra, line 64:
```suggestion
```

packages/anoncreds/src/services/AnonCredsHolderServiceOptions.ts
```diff
@@ -0,0 +1,102 @@
...
+  credentialId?: string
+  revocationRegistryDefinition?: AnonCredsRevocationRegistryDefinition
+}
+export type StoreCredentialMetadata = {
+  credentialDefinitionMetadata: {
+    id: string
+    ver: string
+  }
+  revocationRegistryDefinition: {
+    id: string
```
TimoGlastra, line 62: For the create credential definition method we pass the schema with id as:

```
{
  value: schema,
   id: schemaId
}
```

As we're only including the id here, maybe we use that appraoch as well. The id doesn't feel like metadata to me

packages/anoncreds/src/services/AnonCredsIssuerServiceOptions.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  // If schema doesn't include the id, we need to add it as a separate input parameter
+  schema: {
+    value: AnonCredsSchema
+    id: string
+  }
+}
+
+export interface CreateCredentialOfferOptions {
+  credentialDefinitionId: string
```
TimoGlastra, line 29: For all other methods it seems we pass the whole object (credential definition). The issue will be in that the createOffer needs the private part of the credential definition (which we can't get out of the indy-sdk).

packages/anoncreds/src/models/exchange.ts
```diff
@@ -0,0 +1,97 @@
...
+  cred_def_id: string
+  rev_reg_id?: string
+  values: Record<string, CredValue>
+  signature: unknown
+  signature_correctness_proof: unknown
+}
+
+export interface AnonCredsProof {
+  requested_proof: {
+    revealed_attrs:
```
TimoGlastra, line 42: prettier is failing here

TimoGlastra (changes requested):
Overall looks good. I think all metadata values can be removed except for the seqNo value. 

However I think that should be typed in the indy-sdk implementation, not in the generic AnonCreds interfaces? Also could we rename it to `indyTransactionSeqNo` or something to make it very focused on indy credentials?

Not sure if the metadata approach is worth it for just this one value

packages/anoncreds/src/services/AnonCredsHolderServiceOptions.ts
TimoGlastra: Otherwise we should update the createCredentialDefinition method to also put the id in the metadata for consistency

packages/anoncreds/src/services/AnonCredsHolderServiceOptions.ts
karimStekelenburg: Why did you remove the `issuanceType` here?

TimoGlastra: Because it's not part of the spec anymore. The new revocation list model will contain 1s or 0s for all credential indexes, so you can infer the value from the state of revocation list. Indy will still use it and store it on ledger, but it's not needed for the anoncreds spec

karimStekelenburg: Can we hardcode it for Indy? The reason it's here is for Indy. But I get where the confusion comes from, all these metadata interfaces should be part of the indy-sdk package, not the anoncreds package.

TimoGlastra: Even for indy it is in not needed on this layer. It will be needed to derie the revocation list model, but once we have that it doesn't matter which one we're using. If we need to go back to it we can just transform it to say ISSUANCE_BY_DEFUALT and align the deltas based on it (If all are revoked by default and we add all items to the detla that are unrevoked, or if all are unrevoked by default and we add the revoked credentials to the delta. The end result is the same -> we know which credentials are revoked)

karimStekelenburg: Hardcode to `ISSUANCE_BY_DEFUALT`

packages/anoncreds/src/services/AnonCredsHolderServiceOptions.ts
karimStekelenburg: **consistency**

packages/anoncreds/src/services/AnonCredsIssuerServiceOptions.ts
karimStekelenburg: Only use IDs where absolutely necessary

packages/anon-creds/src/AnonCredsApi.ts
```diff
@@ -0,0 +1,20 @@
+import type { RegisterSchemaApiOptions, RegisterSchemaApiReturn } from './AnonCredsApiOptions'
+import type { AnonCredsSchema } from './models/vdr'
+
+export interface AnonCredsApi {
```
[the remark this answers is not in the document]
karimStekelenburg: Just removed the API and API options. Posting the contents of those files here for future reference.

**AnonCredsApi.ts**

```typescript
import type { RegisterSchemaApiOptions, RegisterSchemaApiReturn } from './AnonCredsApiOptions'
import type { AnonCredsSchema } from './models/registry'

export interface AnonCredsApi {
  getSchema(schemaId: string): Promise<AnonCredsSchema>
  register(options: RegisterSchemaApiOptions): Promise<RegisterSchemaApiReturn>

  // TODO: getCredentialDefinition
  // TODO: registerCredentialDefinition

  // NOTE: Do we want to expose these methods? Probably, but it should also just auto trigger when   is set to true in the credential definition.
  // TODO: getRevocationRegistryDefinition
  // TODO: registerRevocationRegistryDefinition

  // NOTE: Naming TDB
  // NOTE: Do we want to expose these methods? Probably, but it should also just auto trigger when credentials are revoked
  // NOTE: How will users revoke credentials? This hasn't been supported in the past, but is an important feature going forward that we need to implement.
  // TODO: getRevocationList
  // TODO: registerRevocationList
}
```

**AnonCredsApiOptions**

```typescript
import type { AnonCredsSchema } from './models/registry'

// FIXME: we should update the name in the AnonCreds registry interface to include extra wording in the name to prevent name clashes.
export interface RegisterSchemaApiOptions {
  schema: AnonCredsSchema

  // Identifier of the identifier that will create the schema. In most cases a did, but can also be an URI.
  //   TODO: do we want an issuerId? Or should we add such properties to an options object that is different per method
  // more like the did registration interface?
  issuerId: string
}

export interface RegisterSchemaApiReturn {
  schemaId: string
  schema: AnonCredsSchema
}
```

TimoGlastra (approved):
Awesome! Let's get this merged and we can iterate over it as the spec changes

PR #1119, opened 2022-11-24
fix: use custom document loader in jsonld.frame

### LINKED ISSUE #1111: W3cCredentialService Custom Document Loader does not work when no Internet -> e2e Tests Fail
If I run the w3cCredentialService tests (in modules/vc/____tests____/W3cCredentialService.test.ts) with my internet connection turned off most of the tests fail (with the invalid URL error that you get when it cannot resolve an external uri).

See the custom document loader in packages/core/src/modules/vc/__tests__/documentLoader.ts

The custom loader has a call to jsonld.frame for **did**: URLs :-

    if (url.startsWith('did:')) {
      result = await jsonld.frame(result, {
        '@context': result['@context'],
        '@embed': '@never',
        id: url,
      })
    }
    
Inside the jsonld.frame call (deep inside the digitalcredentials library) it is attempting to resolve the context by attempting to go across the internet (which fails)

It looks like it uses its own default document loader to do this (which I believe relies on having a network connection)

karimStekelenburg:
We use a custom document loader for tests that use local files instead of downloading them from the web. Inside this custom document loader, we're calling `jsonld.frame()`, but didn't pass the custom document loader. As a result, documents were being fetched from the web during tests, which made the tests fail when there is no internet connection.

This minor fix passes the custom document loader to the `jsonld.frame()` function and makes sure only local files are used.

Fixes #1111 


TimoGlastra (approved):
That's the longest branch name I've ever seen :)

genaris:
> That's the longest branch name I've ever seen :)

Testing the boundaries of git! :stuck_out_tongue:

karimStekelenburg:
Haha, it's what happens if you let vscode create a branch based on an issue 🤣

packages/core/src/modules/vc/__tests__/contexts/bbs_v1.ts
```diff
@@ -51,6 +51,7 @@ export const BBS_V1 = {
         '@protected': true,
         id: '@id',
         type: '@type',
+
```
blu3beri, line 54:
```suggestion
```

blu3beri (approved):

genaris:
@blu3beri is it OK to merge? (I've seen you disabled auto-merge but don't know if it's for an specific reason)

blu3beri:
> @blu3beri is it OK to merge? (I've seen you disabled auto-merge but don't know if it's for an specific reason)

I think @karimStekelenburg wants to add something to an error message still. I will merge it when he adds it :).

TimoGlastra:
@karimStekelenburg is this ready?

PR #1131, opened 2022-11-29
feat(proofs): proof negotiation

Przytua:
Signed-off-by: Łukasz Przytuła <lprzytula@gmail.com>

Allow proof negotiation by responding to presentation proposal with new presentation request, and responding to presentation request with new presentation proposal

### DIFF
--- a/packages/core/src/modules/proofs/ProofsApi.ts
+++ b/packages/core/src/modules/proofs/ProofsApi.ts
@@ -11,6 +11,8 @@ import type {
   ProposeProofOptions,
   RequestProofOptions,
   ProofServiceMap,
+  NegotiateRequestOptions,
+  NegotiateProposalOptions,
 } from './ProofsApiOptions'
 import type { ProofFormat } from './formats/ProofFormat'
 import type { IndyProofFormat } from './formats/indy/IndyProofFormat'
@@ -28,6 +30,7 @@ import type {
   FormatRetrievedCredentialOptions,
   DeleteProofOptions,
   GetFormatDataReturn,
+  CreateProposalAsResponseOptions,
 } from './models/ProofServiceOptions'
 import type { ProofExchangeRecord } from './repository/ProofExchangeRecord'
 
@@ -57,11 +60,13 @@ export interface ProofsApi<PFs extends ProofFormat[], PSs extends ProofService<P
   // Proposal methods
   proposeProof(options: ProposeProofOptions<PFs, PSs>): Promise<ProofExchangeRecord>
   acceptProposal(options: AcceptProofProposalOptions): Promise<ProofExchangeRecord>
+  negotiateProposal(options: NegotiateProposalOptions<PFs>): Promise<ProofExchangeRecord>
 
   // Request methods
   requestProof(options: RequestProofOptions<PFs, PSs>): Promise<ProofExchangeRecord>
   acceptRequest(options: AcceptProofPresentationOptions<PFs>): Promise<ProofExchangeRecord>
   declineRequest(proofRecordId: string): Promise<ProofExchangeRecord>
+  negotiateRequest(options: NegotiateRequestOptions<PFs>): Promise<ProofExchangeRecord>
 
   // Present
   acceptPresentation(proofRecordId: string): Promise<ProofExchangeRecord>
@@ -204,13 +209,14 @@ export class ProofsApi<
    */
   public async acceptProposal(options: AcceptProofProposalOptions): Promise<ProofExchangeRecord> {
     const { proofRecordId } = options
+
     const proofRecord = await this.getById(proofRecordId)
 
     const service = this.getService(proofRecord.protocolVersion)
 
     if (!proofRecord.connectionId) {
       throw new AriesFrameworkError(
-        `No connectionId found for credential record '${proofRecord.id}'. Connection-less issuance does not support presentation proposal or negotiation.`
+        `No connectionId found for proof record '${proofRecord.id}'. Connection-less issuance does not support presentation proposal or negotiation.`
       )
     }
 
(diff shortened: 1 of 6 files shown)


packages/core/src/modules/proofs/ProofsApi.ts
```diff
@@ -270,6 +317,46 @@ export class ProofsApi<
...
+
+    const requestOptions: CreateRequestAsResponseOptions<PFs> = {
+      proofRecord,
+      proofFormats: options.proofFormats,
+      autoAcceptProof: options.autoAcceptProof,
+      comment: options.comment,
+    }
+    const { message } = await service.createRequestAsResponse(this.agentContext, requestOptions)
+
+    const outbound = createOutboundMessage(connection, message)
```
genaris, line 354: createOutboundMessage has been replaced by the class OutboundMessageContext

genaris (commented):
This is very interesting. But I'm curious about the usage of this negotiation in a loop that involves more than a single message of each type. I mean: is it possible for a proof exchange to follow the path: propose-presentation -> request-presentation -> propose-presentation -> request-presentation -> presentation? If so, what happens when we do `agent.proofs.findRequestMessage` or `agent.proofs.findProposalMessage`? As it is supposed to be a single message, would it be updated to the latest one?

Przytua:
> is it possible for a proof exchange to follow the path: propose-presentation -> request-presentation -> propose-presentation -> request-presentation -> presentation?

I assume (according to this diagram: https://github.com/hyperledger/aries-rfcs/blob/main/features/0037-present-proof/README.md#choreography-diagram) that it should be possible. With a small patch of AFJ and current implementation of acapy we were able to achieve `request-presentation -> propose-presentation -> request-presentation -> presentation` and it was properly updating former proof request message with the new one updated according to presentation proposal.

> As it is supposed to be a single message, would it be updated to the latest one?

Yes. I added some more tests to make sure it returns latest message.

packages/core/src/modules/proofs/ProofsApi.ts
```diff
@@ -191,6 +196,47 @@ export class ProofsApi<
...
+   */
+  public async proposeProofAsResponse(options: ProposeProofAsResponseOptions<PFs>): Promise<ProofExchangeRecord> {
+    const { proofRecordId } = options
+    const proofRecord = await this.getById(proofRecordId)
+
+    const service = this.getService(proofRecord.protocolVersion)
+
+    if (!proofRecord.connectionId) {
+      throw new AriesFrameworkError(
+        `No connectionId found for credential record '${proofRecord.id}'. Connection-less issuance does not support presentation proposal or negotiation.`
```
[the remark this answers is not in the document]
Przytua: Fixed. I copied it from the `acceptProposal` from the same file, so fixed it there as well.

packages/core/src/modules/proofs/ProofsApi.ts
```diff
@@ -270,6 +317,46 @@ export class ProofsApi<
...
+  public async requestProofAsResponse(options: RequestProofAsResponseOptions<PFs>): Promise<ProofExchangeRecord> {
+    const { proofRecordId } = options
+
+    const proofRecord = await this.getById(proofRecordId)
+
+    const service = this.getService(proofRecord.protocolVersion)
+
+    if (!proofRecord.connectionId) {
+      throw new AriesFrameworkError(
+        `No connectionId found for credential record '${proofRecord.id}'. Connection-less issuance does not support negotiation.`
```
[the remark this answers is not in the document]
Przytua: Fixed

packages/core/src/modules/proofs/ProofsApi.ts
Przytua: Fixed

packages/core/src/modules/proofs/ProofsApi.ts
```diff
@@ -56,10 +59,12 @@ import { ProofRepository } from './repository/ProofRepository'
 export interface ProofsApi<PFs extends ProofFormat[], PSs extends ProofService<PFs>[]> {
   // Proposal methods
   proposeProof(options: ProposeProofOptions<PFs, PSs>): Promise<ProofExchangeRecord>
+  proposeProofAsResponse(options: ProposeProofAsResponseOptions<PFs>): Promise<ProofExchangeRecord>
```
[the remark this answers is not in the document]
Przytua: Renamed to `negotiateRequest` and `negotiateProposal` accordingly in the `ProofsApi`. Functions in services were already implemented, and I just used them in the api. They're named `createProposalAsResponse` and `createRequestAsResponse`. Do you want me to rename them as well or is that fine?

packages/core/src/modules/proofs/protocol/v2/__tests__/indy-proof-negotiation.test.ts
```diff
@@ -0,0 +1,444 @@
+import type { Agent } from '../../../../../agent/Agent'
```
TimoGlastra, line 1: It seems there's thests both in the top level tests and also in the protocol specific directories. What's the difference between them?

packages/core/src/modules/proofs/ProofsApi.ts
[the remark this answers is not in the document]
TimoGlastra: We probably want to rename them, but it's more for a different pr I think.

TimoGlastra:
sorry forgot about this one. Code LGTM, just one qeustion about the tests

packages/core/src/modules/proofs/protocol/v2/__tests__/indy-proof-negotiation.test.ts
Przytua: Yes, I added tests in the `core/tests` first (as it seemed to me more of a e2e test), and later I noticed, that there are test cases in the `__tests__` folder of the protocol, that are covering similar scenarios, so added there as well. But true, they're almost no different

TimoGlastra: Can we remove one of them in that case to keep this tidy? I think my preference would go to keep them in the credentials module folder

Przytua: I assume you mean proofs module folder. Anyway removed those in `core/tests`.

genaris: Yeah I think this comes from the proofs module, where its root __tests__ directory has some unit tests but protocol/vx includes e2e tests.

I agree on merge these tests and leaving one of each in modules/proofs/__tests__. They can be named 'xxx.e2e.test.ts' like it happens in other modules E2E tests. Later we should open a new PR to harmonize this structure.

TimoGlastra: > I assume you mean proofs module folder

yes 😬

Przytua:
> @Przytua if you can update the branch with main I can merge

@TimoGlastra updated

Issue #1145, opened 2022-12-07
Querying for tags whose content is an array


genaris:
When merging #994 we set the possibility of adding multiple connection types in order to give more flexibility to the application to categorize the different connections. This is implemented by setting the "connectionType" tag as a string array.

My question is: is there a way to do an efficient query for a tag like that if it contains more than a single element? Let's say I have a connection whose `connectionType` tag is `[ 'Mediator', 'Issuer' ]` and I want to search for all connections whose connectionType tag contains the type 'Issuer'. Is it possible to do so with Indy and Askar WQL?

TimoGlastra:
Yes! I think you can just do the following:

```
  await faberAgent.connections.findAllByQuery({
    connectionType: ['a', 'b'],
  })
```

this will translate to:

```
    {
      'connectionType:a': '1',
      'connectionType:b': '1',
      '$and': undefined,
      '$or': undefined,
      '$not': undefined
    }
```

Indy doesn't natively support array tags, but we achieve it by having the `key:value` as the tag name.

---

Had to make two slight adjustions:
1. Add `DefaultConnectionTags & CustomConnectionTags` as return type for the `getTags` in `ConnectionRecord`, I think there's something wrong with the inference of findByQuery. 
2. Had to change the type of the connectionType tag to `Array<ConnectionType | string>` as `[ConnectionType | string]` doesn't work and means an array with one entry

genaris:
Great! Thanks for the explanation. I've added these adjustments to #1147 so we can query connection types properly.

PR #1147, opened 2022-12-08
fix(routing): add connection type on mediation grant

genaris:
Small update to the feature added in #994 to aggregate the connection type when a connection is related to a mediation grant. 

Previously, it was setting **only** `ConnectionType.Mediator`, so if the connection was previously tagged with another type, it would be lost.

Also fixes typing for `connectionType` in `ConnectionRecord` in order to be able to properly query for connections matching one or multiple types.

### DIFF
--- a/packages/core/src/modules/connections/services/ConnectionService.ts
+++ b/packages/core/src/modules/connections/services/ConnectionService.ts
@@ -642,21 +642,26 @@ export class ConnectionService {
     return connectionRecord
   }
 
-  public async addConnectionType(agentContext: AgentContext, record: ConnectionRecord, type: string) {
-    const tags = (record.getTag('connectionType') as string[]) || ([] as string[])
-    record.setTag('connectionType', [type, ...tags])
-    await this.update(agentContext, record)
+  public async addConnectionType(agentContext: AgentContext, connectionRecord: ConnectionRecord, type: string) {
+    const tags = (connectionRecord.getTag('connectionType') as string[]) || ([] as string[])
+    connectionRecord.setTag('connectionType', [type, ...tags])
+    await this.update(agentContext, connectionRecord)
   }
 
-  public async removeConnectionType(agentContext: AgentContext, record: ConnectionRecord, type: string) {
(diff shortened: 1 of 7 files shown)


TimoGlastra (approved):
Thanks @genaris! Could you add some tests to make sure it correctly adds instead of overwrites?

genaris:
> Thanks @genaris! Could you add some tests to make sure it correctly adds instead of overwrites?

Done and also fixed the typing as discussed in #1145. Could you quickly review these new changes?

packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -642,6 +642,28 @@ export class ConnectionService {
     return connectionRecord
   }
 
+  public async addConnectionType(agentContext: AgentContext, connectionRecord: ConnectionRecord, type: string) {
+    const tags = (connectionRecord.getTag('connectionType') as string[]) || ([] as string[])
```
TimoGlastra, line 646: Now that the typing is correctly set on getTags, I think the casting is not needed anymore?

packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -642,6 +642,28 @@ export class ConnectionService {
...
+    const newTags = tags.filter((value: string) => {
+      if (value != type) return value
+    })
+    connectionRecord.setTag('connectionType', [...newTags])
+
+    await this.update(agentContext, connectionRecord)
+  }
+
+  public async getConnectionTypes(connectionRecord: ConnectionRecord) {
+    const tags = connectionRecord.getTag('connectionType') as string[]
```
TimoGlastra, line 663: This cast is incorrect (as it could be null), but per my comment above, not sure if the cast is needed anymore

packages/core/src/modules/connections/services/ConnectionService.ts
genaris: If we use `getTag('connectionType')` it's still needed. However I think it's more elegant to use `getTags().connectionType`, where the typing is already there as you say.

PR #1148, opened 2022-12-08
feat: Corrected peer DIDs resolution (key IDs)

spivachuk:
This PR corrected resolution of peer DIDs of method 0 (`did:peer:0`) and method 2 (`did:peer:2`) into DID documents, specifically resulting key IDs.
For peer DIDs of method 0 and method 2, key fragment IDs are multibase representations of public keys **without** the multibase prefix (i.e. **without** `z` prefix in case of base58 encoding).

This logic is based on:
- _Example 4_ from https://identity.foundation/peer-did-method-spec/#multi-key-creation
- examples from https://github.com/sicpa-dlab/peer-did-jvm
- tests from https://github.com/sicpa-dlab/didcomm-demo/blob/main/didcomm-demo-python/tests/test_did_resolver_peer_did.py and https://github.com/sicpa-dlab/didcomm-demo/blob/main/didcomm-demo-jvm/didcomm-demo/src/test/kotlin/org/didcommx/didcomm/demo/DIDDocResolverPeerDIDTest.kt

At the same time, this PR keeps the existing resolution logic for key DIDs (`did:key`). So for key DIDs, key fragment IDs are multibase representations of public keys including the multibase prefix (as specified at https://w3c-ccg.github.io/did-method-key/)

This PR is based on https://github.com/hyperledger/aries-framework-javascript/pull/1096 which must be merged first.


TimoGlastra:
Closing as it's superseded by other PRs

Issue #497, opened 2021-10-20
Add support for RFC 0360: did:key Usage


TimoGlastra:
https://github.com/hyperledger/aries-rfcs/tree/b3a3942ef052039e73cd23d847f42947f8287da2/features/0360-use-did-key

genaris:
I'll take a look at this one. For RFC 0211 usage of did:key maybe we can do something with Discover Features V2 to determine the key format to use.

genaris:
I'm having some headaches trying to find a proper way of implementing this while maintaining compatibility with older agents. IMHO, this did:key thing has introduced a breaking change in protocols but their version number is kept (and there is no new flag to explicitly indicate RFC 0360 support). And also it has been introduced some time before we implemented RFC 0211, so we can admit that our implementation (probably biased by ACA-Py's) was not according to the spec.

It seems to me that a balanced solution would be to add a (yet another) flag in routing module that, when enabled, agent will encode keys in did:key for all related protocols (i.e Coordinate Mediation, Pick Up V1 and Pick Up V2). This flag will be default to false in 0.2.x to keep current behaviour.

For next major version, I think it should default to true, to enforce alignment with AIP 2.0 in new AFJ deployments. Yes, I know this is controversial, as some AFJ contributors like @TimoGlastra and myself are not convinced of this approach of encoding keys :stuck_out_tongue:, but it could be useful to ensure compatibility with other agents that might no be aware of this backwards compatibility (especially considering that this RFC 0360 was accepted 1,5 years ago.

BTW apart from routing protocols, is there any other protocol that we should still adapt? I'm not sure about Connections (RFC 0160) but I guess the other ones (OOB, DID Exchange, etc.) are already covered.

swcurran:
I assume the breaking issue is when to send a key in `did:key` format. Presumably it is pretty easy to support to detect/use `did:key` when you receive it.  Has that been added to AFJ? However, when sending something, you can't be sure the other side has done even that much. Is that correct?

@ianco @andrewwhitehead @shaangill025 @dbluhm -- we've got to make sure we are doing the same things in ACA-Py.  And the comment above about ACA-Py implementation not being to spec.

genaris:
> I assume the breaking issue is when to send a key in `did:key` format. Presumably it is pretty easy to support detecting/using `did:key` when you receive it. Has that been added to AFJ? However, when sending something, you can't be sure the other side has done even that much. Is that correct?

[...]

PR #1149, opened 2022-12-09
feat!: use did:key in protocols by default

genaris:
As discussed some time ago in #497, now we are having a new major release, we use did:key in protocols by default. We'll of course still accept base58-encoded keys if the other party is not yet using did:key.

BREAKING CHANGE:
`useDidKeyInProtocols` configuration parameter is now enabled by default. If your agent only interacts with modern agents (e.g. AFJ 0.2.5 and newer) this will not represent any issue. Otherwise it is safer to explicitly set it to `false`. However, keep in mind that we expect this setting to be deprecated in the future, so we encourage you to update all your agents to use did:key.

### DIFF
--- a/packages/core/src/agent/AgentConfig.ts
+++ b/packages/core/src/agent/AgentConfig.ts
@@ -125,7 +125,7 @@ export class AgentConfig {
    * in a given protocol (i.e. it does not support Aries RFC 0360).
    */
   public get useDidKeyInProtocols() {
-    return this.initConfig.useDidKeyInProtocols ?? false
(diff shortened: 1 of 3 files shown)


TimoGlastra (approved):
This is a breaking change right?

genaris:
> This is a breaking change right?

Yes. I'm updating the migration guide to reflect this change and also the new OutboundMessageContext class. Check [this PR](https://github.com/hyperledger/aries-javascript-docs/pull/81).

PR #1156, opened 2022-12-13
feat(oob): receive Invitation with timeout

Zzocker:
Signed-off-by: Pritam Singh <pkspritam16@gmail.com>

### DIFF
--- a/packages/core/src/modules/oob/OutOfBandApi.ts
+++ b/packages/core/src/modules/oob/OutOfBandApi.ts
@@ -75,6 +75,7 @@ export interface ReceiveOutOfBandInvitationConfig {
   autoAcceptConnection?: boolean
   reuseConnection?: boolean
   routing?: Routing
+  acceptInvitationTimeoutMs?: number
 }
 
 @injectable()
@@ -388,6 +389,7 @@ export class OutOfBandApi {
         autoAcceptConnection,
         reuseConnection,
         routing,
+        timeoutMs: config.acceptInvitationTimeoutMs,
       })
     }
 
@@ -417,6 +419,7 @@ export class OutOfBandApi {
       alias?: string
       imageUrl?: string
       routing?: Routing
+      timeoutMs?: number
     }
   ) {
     const outOfBandRecord = await this.outOfBandService.getById(this.agentContext, outOfBandId)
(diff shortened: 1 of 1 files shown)


packages/core/src/modules/oob/OutOfBandApi.ts
```diff
@@ -483,7 +486,7 @@ export class OutOfBandApi {
         } else {
           // Wait until the connection is ready and then pass the messages to the agent for further processing
           this.connectionsApi
-            .returnWhenIsConnected(connectionRecord.id)
+            .returnWhenIsConnected(connectionRecord.id, config.timeoutMs ? { timeoutMs: config.timeoutMs } : undefined)
```
TimoGlastra, line 489: I think we can simplify to this?

```suggestion
            .returnWhenIsConnected(connectionRecord.id, { timeoutMs: config?.timeoutMs })
```

packages/core/src/modules/oob/OutOfBandApi.ts
```diff
@@ -75,6 +75,7 @@ export interface ReceiveOutOfBandInvitationConfig {
   autoAcceptConnection?: boolean
   reuseConnection?: boolean
   routing?: Routing
+  timeoutMs?: number
```
TimoGlastra, line 78: Thinking whether we should give this another name, or whether we should set a timer for the overall action. If I set a timeoutMs I expect the whole action to have a specific timeout, but this only relates to the time we wait for the connection to be completed afterwards.

Also if autoAcceptConnection is false, this parameter won't do anything which may be a bit misleasing. 

What do you think?

Zzocker: Yes, make sense, [FORCED PUSHED] changed the name to `acceptInvitationTimeoutMs`

packages/core/src/modules/oob/OutOfBandApi.ts
```diff
@@ -483,7 +487,7 @@ export class OutOfBandApi {
         } else {
           // Wait until the connection is ready and then pass the messages to the agent for further processing
           this.connectionsApi
-            .returnWhenIsConnected(connectionRecord.id)
+            .returnWhenIsConnected(connectionRecord.id, { timeoutMs: timeoutMs })
```
genaris, line 490:
```suggestion
            .returnWhenIsConnected(connectionRecord.id, { timeoutMs })
```

genaris (approved):
Just a super small suggestion but LGTM

packages/core/src/modules/oob/OutOfBandApi.ts
Zzocker: Made the suggested changes.

PR #1157, opened 2022-12-13
fix: credential values encoding

Przytua:
Signed-off-by: Łukasz Przytuła <lprzytula@gmail.com>

String values containing round decimal numbers (for example `1.0`, `6.0`) were encoded as a number (into `1` and `6` respectively) instead of properly hashing the string value, which causes issues with accepting credentials containing such values, issued by aca-py.


TimoGlastra (approved):

TimoGlastra:
Same here, I can't update the branch if you create the PR from an organization

Przytua:
@TimoGlastra updated

PR #1158, opened 2022-12-14
feat(credentials)!: custom registration of credential protocols

TimoGlastra:
This PR allows for custom registration of credential protocols. Lot's to unpack here, so an overview of the changes. This is just for credentials for now. We can repeat this setup for Proofs

## Custom Registration of Credential Protocols

This PR makes it possible to register custom credential protocols using the module configuration. See the example below. By default, it registers the `V1CredentialProtocol` and `V2CredentialProtocol` with the `IndyCredentialFormatService`. If you provide the `credentialProtocols` array yourself it will override this default. The credentials module typing will still be correctly inferred based on the credential protocols you provide (so e.g. if you only register `V3CredentialProtocol` you'll only be able to use `v3` when offering a credential). This means a bit more type magic, but for the added benefit of a better user experience.

```ts
const indyCredentialFormat = new IndyCredentialFormatService()

const agent new Agent({
  config: {},
  modules: {
    credentials: new CredentialsModule({
       credentialProtocols: [
         new V1CredentialProtocol({ 
           indyCredentialFormat,
         }),
         new V2CredentialProtocol({ 
           credentialFormats: [indyCredentialFormat]
         })
       ]
    })
 }
})
```

An important thing to note here is that the protocols and credential formats are created when you pass them (instead of having their dependencies injected in the constructor). This is to allow for configuration of the custom protocols and formats. For the v1 protocol it meas passing the `indyCredentialFormat`, and for the v2 protocol it means passing a generic array of credential foramts. The reason for doing this is that it means less magic for the user of the framework. You can add config options in the constructor. This will become useful for other items that are specific to the custom class. E.g. the `CheqdDidResolver` will need an RPC url configured. By allowing it to be passed in the constructor, it's easy for the user to provide the needed configuration.

This does mean we can't inject dependencies in the constructor anymore. As the AgentContext is now passed to all methods, we can however just inject them in the needed methods.  (see example below). 

```ts
// example
class IndyCredentialFormatService {

  public issueCredential(agentContext: AgentContext) {
    const issuerService = agentContext.dependencyMaanger.resolve(IndyIssuerService)
  }
```

It's different than we have now, but it allows for some added benefits, namely:
- Dependencies that are registrered after the class has been constructed will now be used the next time the dependency is injected (dynamic dependencies)
- We could now start using Apis instead of Services in these classes. The benefit of this is that for custom classes we recommend people to only use the Apis of other modules, which means that you won't suddenly be surprised with a change in an internal service. This was first impossible due to dependency cycles.

```
class IndyCredentialFormatService {
  public issueCredential(agentContext: AgentContext, connectionId: string) {
    // use public api
    const conectionApi = agentContext.dependencyMaanger.resolve(ConnectionApi)
  }
```

## Handler Registry

With the registration of custom protocls we need to be able to register their handlers and dependencies. For this I've created a new `MessageHandlerRegistry`, which basically extract this functionality out of the Dispatcher. Through the `DependencyManager` you can now register a handler.

E.g. the `CredentialProtocol` now defines a `register` method that will be called to register the dependencies of the custom protocol. 

For the `V1CredentialProtocol` the implementation looks like this. It registers the handlers, and it registers the protocol in the feature registry (for discover features). I've had a discussion with @genaris to make this an object (declarative) instead of a method (imperative), but I wasn't sure on the structure yet and we already use the register method in other places (so it's consistent)

```ts
  /**
   * Registers the protocol implementation (handlers, feature registry) on the agent.
   */
  public register(dependencyManager: DependencyManager, featureRegistry: FeatureRegistry) {
    // Register message handlers for the Issue Credential V1 Protocol
    dependencyManager.registerHandlers([
      new V1ProposeCredentialHandler(this),
      new V1OfferCredentialHandler(this),
      new V1RequestCredentialHandler(this),
      new V1IssueCredentialHandler(this),
      new V1CredentialAckHandler(this),
      new V1CredentialProblemReportHandler(this),
    ])

    // Register Issue Credential V1 in feature registry, with supported roles
    featureRegistry.register(
      new Protocol({
        id: 'https://didcomm.org/issue-credential/1.0',
        roles: ['holder', 'issuer'],
      })
    )
  }
```

BREAKING CHANGE: credential service has been renamed to credential protocol. This means `CredentialService` is now `CredentialProtocol`, `V1CredentialService` is now `V1CredentialProtocol`, etc.


packages/core/src/agent/HandlerRegistry.ts
```diff
@@ -0,0 +1,59 @@
+import type { AgentMessage } from './AgentMessage'
+import type { Handler } from './Handler'
+
+import { injectable } from 'tsyringe'
+
+import { canHandleMessageType, parseMessageType } from '../utils/messageType'
+
+@injectable()
+export class MessageHandlerRegistry {
```
genaris, line 9: I think this file should be named `MessageHandlerRegistry.ts`

packages/core/src/modules/credentials/CredentialsApi.ts
```diff
@@ -41,34 +38,32 @@ import { RoutingService } from '../routing/services/RoutingService'
...
 import { CredentialRepository } from './repository/CredentialRepository'
 
-export interface CredentialsApi<CFs extends CredentialFormat[], CSs extends CredentialService<CFs>[]> {
+export interface CredentialsApi<CPs extends CredentialProtocol[]> {
   // Propose Credential methods
-  proposeCredential(options: ProposeCredentialOptions<CFs, CSs>): Promise<CredentialExchangeRecord>
-  acceptProposal(options: AcceptCredentialProposalOptions<CFs>): Promise<CredentialExchangeRecord>
-  negotiateProposal(options: NegotiateCredentialProposalOptions<CFs>): Promise<CredentialExchangeRecord>
+  proposeCredential(options: ProposeCredentialOptions<CPs>): Promise<CredentialExchangeRecord>
+  acceptProposal(options: AcceptCredentialProposalOptions<[]>): Promise<CredentialExchangeRecord>
```
TimoGlastra, line 46: Need to fix these empty arrays

packages/core/src/modules/credentials/protocol/v1/__tests__/V1CredentialServiceProposeOffer.test.ts
```diff
@@ -1,6 +1,5 @@
 import type { CredentialStateChangedEvent } from '../../../CredentialEvents'
```
genaris, line 1: I guess this file and other ex-service tests should be renamed to V1Credential**Protocol**XXX.test.ts

genaris (approved):
This is great (and thanks for the clear explanation of the proposed changes)! It will be clearly beneficial for the 'new generation modules' that support multiple protocol versions and crucial if we would like to create customized agents like a pure DIDComm V2 that only supports protocols based on it, while still having a default 'batteries-includes' implementation.

Something I really like is your point about leveraging modules to call other modules Apis instead of Services (or Protocols). This will simplify the implementation and also encourage to write better Apis!

I just left a few comments regarding the naming on handlers some of their associated methods, as the word 'Handler' is maybe too generic in software engineering. But is just a suggestion, won't be mad if we leave for another PR or just discard them 😛

packages/core/src/agent/HandlerRegistry.ts
```diff
@@ -0,0 +1,59 @@
+import type { AgentMessage } from './AgentMessage'
+import type { Handler } from './Handler'
```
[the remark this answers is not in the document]
TimoGlastra: Yes sounds good. I'll do this in a separate PR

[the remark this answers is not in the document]
TimoGlastra: See #1161

PR #1160, opened 2022-12-14
feat(indy-vdr): add indy-vdr package and indy vdr pool

Vickysomtee:
This PR implements the new IndyVdr Pool Package

**Work funded by the government of Ontario**


packages/indy-vdr/src/pool/IndyVdrPoolService.ts
```diff
@@ -0,0 +1,230 @@
...
+      this.logger.debug(`Finished connection to pool: ${pool.id}`)
+      handleArray.push(poolHandle)
+    }
+    return handleArray
+  }
+
+  /**
+   * Get the most appropriate pool for the given did. The algorithm is based on the approach as described in this document:
+   * https://docs.google.com/document/d/109C_eMsuZnTnYe2OAd02jAts1vC4axwEKIq7_4dnNVA/edit
+   */
```
TimoGlastra, line 70:
```suggestion
   * Get the most appropriate pool for the given did. If the did is a qualified indy did, the pool will be determined based on the namespace. If it is a legacy unqualified indy did, the pool will be determined based on the algorithm as described in this document:
   * https://docs.google.com/document/d/109C_eMsuZnTnYe2OAd02jAts1vC4axwEKIq7_4dnNVA/edit
   */
```

packages/indy-vdr/src/pool/IndyVdrPoolService.ts
```diff
@@ -0,0 +1,230 @@
...
+  public async getPoolForDid(agentContext: AgentContext, did: string): Promise<{ pool: IndyVdrPool }> {
+    // Check if the did starts with did:indy
+    if (did.startsWith('did:indy')) {
+      const nameSpace = did.split(':')[2]
```
TimoGlastra, line 74: The namespace can contain multiple colons. 

E.g. the following dids are all correct:

- `did:indy:sovrin:builder:xxx` -> namespace = `sovrin:builder`
- `did:indy:bcovrin:xxx` -> namespace = `bcovrin`.

We should proabbly use a regex to match the did. If you store the followign const in a separate file:

```
const DID_INDY_REGEX = /^did:indy:((?:[a-z][_a-z0-9-]*)(?::[a-z][_a-z0-9-]*)):([1-9A-HJ-NP-Za-km-z]{21,22})$/
```

Then you can use it like this:

```ts
const match = did.match(DID_INDY_REGEX)

// If we have a match it means the did starts with did:indy and is a qualified indy did
if (match) {
  const [, namespace] = match
} else {
  // legacy way
}
```

packages/indy-vdr/src/pool/IndyVdrPoolService.ts
```diff
@@ -0,0 +1,230 @@
...
+
+      const pool = this.pools.find((pool) => pool.IndyNamespace === nameSpace)
+
+      if (pool) return { pool }
+
+      throw new LedgerNotFoundError('Pool not found')
```
TimoGlastra, line 80: Can we also update this to the indy vdr error?

packages/indy-vdr/src/pool/IndyVdrPoolService.ts
```diff
@@ -0,0 +1,230 @@
...
+      value = productionOrNonProduction[0].value
+    }
+
+    await this.didCache.set(agentContext, did, {
+      nymResponse: {
+        did: value.did.nymResponse.did,
+        verkey: value.did.nymResponse.verkey,
+      },
+    })
+    return { pool: value.pool, did: value.did.nymResponse.did }
```
TimoGlastra, line 149: I think we can update this to just return the pool

packages/indy-vdr/src/pool/IndyVdrPoolService.ts
```diff
@@ -0,0 +1,230 @@
...
+      )
+    }
+
+    const cachedNymResponse = await this.didCache.get(agentContext, did)
+    const pool = this.pools.find((pool) => pool.config.indyNamespace === cachedNymResponse?.poolId)
+
+    // If we have the nym response with associated pool in the cache, we'll use that
+    if (cachedNymResponse && pool) {
+      this.logger.trace(`Found ledger id '${pool.id}' for did '${did}' in cache`)
+      return { did: cachedNymResponse.nymResponse.did, pool }
```
TimoGlastra, line 104: Same here

packages/indy-vdr/src/pool/IndyVdrPoolService.ts
```diff
@@ -0,0 +1,230 @@
...
+   * https://docs.google.com/document/d/109C_eMsuZnTnYe2OAd02jAts1vC4axwEKIq7_4dnNVA/edit
+   */
+  public async getPoolForDid(agentContext: AgentContext, did: string): Promise<{ pool: IndyVdrPool }> {
+    // Check if the did starts with did:indy
+    if (did.startsWith('did:indy')) {
+      const nameSpace = did.split(':')[2]
+
+      const pool = this.pools.find((pool) => pool.IndyNamespace === nameSpace)
+
+      if (pool) return { pool }
```
TimoGlastra, line 78: I think we can return the pool on the top level. So just return pool instead of return { pool }

packages/indy-vdr/src/pool/IndyVdrPoolService.ts
```diff
@@ -0,0 +1,230 @@
...
+        "No indy ledgers configured. Provide at least one pool configuration in the 'indyLedgers' agent configuration"
+      )
+    }
+
+    if (!indyNamespace) {
+      this.logger.warn('Not passing the indyNamespace is deprecated and will be removed in the future version.')
+      return this.pools[0]
+    }
+
+    const pool = this.pools.find((pool) => pool.config.indyNamespace === indyNamespace) // TODO check if this is corect
```
TimoGlastra, line 182: We can use `pool.indyNamespace` directly instead of using `pool.config` I think

packages/indy-vdr/src/pool/IndyVdrPoolService.ts
TimoGlastra: We can then also just use `this.getPoolForNamespace` and reuse that logic

packages/indy-vdr/src/pool/IndyVdrPoolService.ts
```diff
@@ -0,0 +1,230 @@
...
+    } catch (error) {
+      this.logger.trace(`Error retrieving did '${did}' from ledger '${pool.id}'`, {
+        error,
+        did,
+      })
+      if (isIndyError(error, 'LedgerNotFound')) {
+        throw new LedgerNotFoundError(`Did '${did}' not found on ledger ${pool.id}`)
+      } else {
+        throw isIndyError(error) ? new IndySdkError(error) : error
+      }
```
TimoGlastra, line 221: This is from how the indy-sdk worked and it would trhow a ledger not found error. Now, it will return the json and the `data` filed will be `null` (you already do that check).

So this can be simlified to:

```suggestion
 throw error
```

packages/indy-vdr/src/pool/IndyVdrPoolService.ts
```diff
@@ -0,0 +1,230 @@
...
+  private async getDidFromPool(did: string, pool: IndyVdrPool): Promise<PublicDidRequest> {
+    try {
+      this.logger.trace(`Get public did '${did}' from ledger '${pool.id}'`)
+      const request = await new GetNymRequest({ dest: did })
+
+      this.logger.trace(`Submitting get did request for did '${did}' to ledger '${pool.id}'`)
+      const response = await pool.submitReadRequest(request)
+
+      if (!response.result.data) {
+        throw new LedgerError('Not Found')
```
TimoGlastra, line 200: We should probably also create an IndyVdrNotFoundError error based on the indy vdr error and throw that here. Aslo can you make the error message a bit more descriptive?

TimoGlastra (changes requested):
Great work @Vickysomtee! Left some comments mostly on things that can be done a bit differently in the indy vdr library compared to the indy sdk

packages/core/src/index.ts
```diff
@@ -27,6 +27,7 @@ export type {
   WalletConfigRekey,
   WalletExportImportConfig,
 } from './types'
+export * from './cache'
```
TimoGlastra, line 30: The cache directory doesn't exist anymore in `main`. It's now in `packages/core/src/module/cache`. However the needed files should be exported already

packages/indy-vdr/README.md
```diff
@@ -0,0 +1,78 @@
+<p align="center">
+  <br />
+  <img
+    alt="Hyperledger Aries logo"
+    src="https://raw.githubusercontent.com/hyperledger/aries-framework-javascript/aa31131825e3331dc93694bc58414d955dcb1129/images/aries-logo.png"
+    height="250px"
+  />
+</p>
+<h1 align="center"><b>Aries Framework JavaScript Action Menu Plugin</b></h1>
```
TimoGlastra, line 9: The readme is still for the action menu pacakges. Can you update all references?

packages/indy-vdr/package.json
```diff
@@ -0,0 +1,48 @@
+{
+  "name": "@aries-framework/indy-vdr",
+  "main": "build/index",
+  "types": "build/index",
+  "version": "0.2.5",
```
TimoGlastra, line 5: Can you align the version with version of the core package?

packages/indy-vdr/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "test": "jest"
+  },
+  "dependencies": {
+    "class-transformer": "0.5.1",
+    "indy-vdr-test-shared": "^0.1.3",
+    "class-validator": "0.13.1",
+    "rxjs": "^7.2.0"
+  },
+  "peerDependencies": {
+    "@aries-framework/core": "0.2.5",
```
TimoGlastra, line 33: Can you align all `@aries-framework/*` versions with the core one?

packages/indy-vdr/src/error/IndyVdrNotConfiguredError.ts
```diff
@@ -0,0 +1,7 @@
+import {IndyVdrError} from './IndyVdrError'
+
+export class IndyVdrConfiguredError extends IndyVdrError {
```
TimoGlastra, line 3:
```suggestion
export class IndyVdrNotConfiguredError extends IndyVdrError {
```

packages/indy-vdr/src/pool/DidIdentifier.ts
```diff
@@ -0,0 +1 @@
+export const DID_INDY_REGEX = /^did:indy:((?:[a-z][_a-z0-9-]*)(?::[a-z][_a-z0-9-]*)):([1-9A-HJ-NP-Za-km-z]{21,22})$/
```
TimoGlastra, line 1: can you rename this file to `didIdentifier` instead of `DidIdentifier`? We use Pascal case for files that contains classes only. 

See here how to commit a case only name change in git (it's a pain): https://stackoverflow.com/questions/17683458/how-do-i-commit-case-sensitive-only-filename-changes-in-git

packages/indy-vdr/src/pool/IndyVdrPool.ts
```diff
@@ -0,0 +1,184 @@
...
+
+export interface IndyVdrPoolConfig {
+  genesisTransactions: string
+  isProduction: boolean
+  indyNamespace: string
+  transactionAuthorAgreement?: TransactionAuthorAgreement
+}
+
+export class IndyVdrPool {
+  // indyVdr?: typeof IndyVdr
```
TimoGlastra, line 41:
```suggestion
```

packages/indy-vdr/src/pool/IndyVdrPool.ts
```diff
@@ -0,0 +1,184 @@
...
+      },
+    })
+
+    return this.pool.handle
+  }
+
+  private get pool(): indyVdrPool {
+    if (!this._pool) {
+      // TODO: create custom IndyVdrError
+      throw new AriesFrameworkError('Pool is not connected. Make sure to call .connect() first')
```
TimoGlastra, line 73: This should use IndyVdrError

packages/indy-vdr/src/pool/IndyVdrPoolService.ts
```diff
@@ -0,0 +1,224 @@
...
+
+import { IndyVdrError, IndyVdrNotFoundError, IndyVdrConfiguredError } from '../error'
+import { isSelfCertifiedDid } from '../utils/did'
+import { allSettled, onlyFulfilled, onlyRejected } from '../utils/promises'
+
+import { DID_INDY_REGEX } from './DidIdentifier'
+import { IndyVdrPool } from './IndyVdrPool'
+
+export const INDY_VDR_LEGACY_DID_POOL_CACHE_ID = 'INDY_VDR_LEGACY_DID_POOL_CACHE'
+export const DID_POOL_CACHE_LIMIT = 500
```
TimoGlastra, line 16: These are not used anymore

packages/indy-vdr/README.md
genaris, line 9: This README will need to be updated with info about indy-vdr

packages/indy-vdr/src/index.ts
```diff
@@ -0,0 +1,5 @@
+try {
+  require('indy-vdr-test-nodejs')
```
genaris, line 2: How will this work when we want to use this package in React Native? Should we add the require('indy-vdr-test-react-native') to the catch block?

packages/indy-vdr/src/pool/IndyVdrPoolService.ts
```diff
@@ -0,0 +1,224 @@
...
+    if (this.pools.length === 0) {
+      throw new IndyVdrConfiguredError(
+        "No indy ledgers configured. Provide at least one pool configuration in the 'indyLedgers' agent configuration"
+      )
+    }
+
+    if (!indyNamespace) {
+      this.logger.warn('Not passing the indyNamespace is deprecated and will be removed in the future version.')
+      return this.pools[0]
+    }
```
TimoGlastra, line 177: This can be removed. The nameSpace is already required

packages/indy-vdr/src/pool/IndyVdrPoolService.ts
```diff
@@ -0,0 +1,224 @@
...
+
+    const pool = this.pools.find((pool) => pool.indyNamespace === indyNamespace)
+
+    if (!pool) {
+      throw new IndyVdrNotFoundError(`No ledgers found for IndyNamespace '${indyNamespace}'.`)
```
TimoGlastra, line 182: Should be `IndyVdrError` (not found is for when it's not found on the ledger)

packages/indy-vdr/src/pool/IndyVdrPool.ts
```diff
@@ -0,0 +1,184 @@
...
+        transactions: this.config.genesisTransactions,
+      },
+    })
+
+    return this.pool.handle
+  }
+
+  private get pool(): indyVdrPool {
+    if (!this._pool) {
+      // TODO: create custom IndyVdrError
```
genaris, line 72: I guess you can use IndyVdrError now it is created

packages/indy-vdr/tests/indy-vdr-pool.e2e.test.ts
```diff
@@ -0,0 +1,243 @@
...
+          name: 'test-schema',
+          ver: '1.0',
+          version: dynamicVersion,
+          attrNames: ['first_name', 'last_name', 'age'],
+        },
+      })
+
+      const response = await pool.submitWriteRequest(agentContext, schemaRequest, signerKey)
+      expect(response).toBeDefined()
+    })
```
TimoGlastra, line 241: This test doesn't test what it says it does. I think we can just remove it for now

genaris (dismissed):
Looks good! I've left a few comments and questions about the initialization.

Probably as part of another PR (when adding consumers to Indy VDR Pool like the resolvers and registry), but should we create it as a module that registers services and also initializes the binding library within its `register` method?

packages/indy-vdr/src/index.ts
TimoGlastra: I was thinking we could either do it with the catch as you say, or we detect the platform. Or would you prefer the manual approach? (so you provide it to the module). That's also fine for me.

packages/indy-vdr/src/pool/IndyVdrPool.ts
```diff
@@ -0,0 +1,184 @@
...
+    return this._pool
+  }
+
+  public close() {
+    if (!this.pool) {
+      throw new IndyVdrError("Can't close pool. Pool is not connected")
+    }
+
+    // FIXME: this method doesn't work??
+    // this.pool.close()
```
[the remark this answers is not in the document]
TimoGlastra: Yes. it hangs indefinitely. should create an issue for it in the indy-vdr repo. @blu3beri do you still know the details?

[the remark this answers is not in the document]
blu3beri: Ah yeah, apparently, almost a year ago now, I create an issue for this.
https://github.com/hyperledger/indy-vdr/issues/92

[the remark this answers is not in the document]
blu3beri: I can investigate this further if it's required. The reasoning (no write-access to lock) might be completely incorrect.

packages/indy-vdr/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "indy-vdr-test-nodejs": "^0.1.3"
+  },
+  "devDependencies": {
+    "@aries-framework/node": "0.2.5",
+    "indy-vdr-test-nodejs": "^0.1.3",
+    "reflect-metadata": "^0.1.13",
+    "rimraf": "~3.0.2",
+    "typescript": "~4.3.0"
```
genaris, line 41: You will also need to update this line to "typescript": "~4.9.4" to align with other packages. Also for the rimraf (you'll need to remove the -rf parameter in the clean test script)

packages/indy-vdr/tests/indy-vdr-pool.e2e.test.ts
```diff
@@ -100,24 +98,37 @@ describe('IndyVdrPoolService', () => {
...
   })
 
-  xdescribe('CredentialDefinition', () => {})
-
   describe('Schemas & credential Definition', () => {
     test('can write a schema using the pool', async () => {
       const pool = indyVdrPoolService.getPoolForNamespace('pool:localtest')
 
-      const dynamicVersion = `1.${Math.random() * 100}` // TODO Remove this before pushing
+      const dynamicVersion = `1.${Math.random() * 100}`
```
karimStekelenburg, line 123: @TimoGlastra this was originally added by me. I assumed we could remove this, as the CI uses a clean ledger for every run. Is my assumption correct?

karimStekelenburg (approved):
Looks like these are mostly style changes and they look good to me. I'm assuming the README will be done in a follow-up PR, so I'll approve. If not, this needs to be added before merging.

In order to do so, just have a look at the READMEs of other packages inside this repository. They'll give you an idea of what is required.

packages/indy-vdr/tests/indy-vdr-pool.e2e.test.ts
TimoGlastra: Yeah we can. But then subsequent test runs (local) will fail I think? So I thought there's not harm in keeping it in.

packages/indy-vdr/tests/setup.ts
```diff
@@ -0,0 +1,5 @@
+import 'reflect-metadata'
```
TimoGlastra, line 1: This is not needed, as we don't use `class-validator` or `class-transformer`

```suggestion
```

packages/indy-vdr/src/pool/didIdentifier.ts
```diff
@@ -0,0 +1 @@
+export const DID_INDY_REGEX = /^did:indy:((?:[a-z][_a-z0-9-]*)(?::[a-z][_a-z0-9-]*)):([1-9A-HJ-NP-Za-km-z]{21,22})$/
```
TimoGlastra, line 1: Maybe we can move this to the `utlils/did.ts` file next to the `ABBREVIATED_VERKEY_REGEX`?

packages/indy-vdr/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "class-validator": "0.13.1",
+    "eslint-plugin-import": "^2.23.4",
+    "indy-vdr-test-shared": "^0.1.3",
+    "rxjs": "^7.2.0"
+  },
+  "devDependencies": {
+    "@aries-framework/node": "0.3.3",
+    "indy-vdr-test-nodejs": "^0.1.3",
+    "reflect-metadata": "^0.1.13",
+    "rimraf": "~3.0.2",
```
TimoGlastra, line 41:
```suggestion
    "rimraf": "^4.0.7",
```

packages/indy-vdr/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "class-transformer": "0.5.1",
+    "class-validator": "0.13.1",
+    "eslint-plugin-import": "^2.23.4",
+    "indy-vdr-test-shared": "^0.1.3",
+    "rxjs": "^7.2.0"
+  },
+  "devDependencies": {
+    "@aries-framework/node": "0.3.3",
+    "indy-vdr-test-nodejs": "^0.1.3",
+    "reflect-metadata": "^0.1.13",
```
TimoGlastra, line 40:
```suggestion
```

packages/indy-vdr/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "@typescript-eslint/eslint-plugin": "^5.48.1",
+    "@typescript-eslint/parser": "^5.48.1",
+    "class-transformer": "0.5.1",
+    "class-validator": "0.13.1",
+    "eslint-plugin-import": "^2.23.4",
+    "indy-vdr-test-shared": "^0.1.3",
+    "rxjs": "^7.2.0"
+  },
+  "devDependencies": {
+    "@aries-framework/node": "0.3.3",
```
TimoGlastra, line 38:
```suggestion
```

packages/indy-vdr/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "dependencies": {
+    "@aries-framework/core": "0.3.3",
+    "@types/express": "^4.17.13",
+    "@typescript-eslint/eslint-plugin": "^5.48.1",
+    "@typescript-eslint/parser": "^5.48.1",
+    "class-transformer": "0.5.1",
+    "class-validator": "0.13.1",
+    "eslint-plugin-import": "^2.23.4",
+    "indy-vdr-test-shared": "^0.1.3",
+    "rxjs": "^7.2.0"
```
TimoGlastra, line 35:
```suggestion
```

packages/indy-vdr/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "test": "jest"
+  },
+  "dependencies": {
+    "@aries-framework/core": "0.3.3",
+    "@types/express": "^4.17.13",
+    "@typescript-eslint/eslint-plugin": "^5.48.1",
+    "@typescript-eslint/parser": "^5.48.1",
+    "class-transformer": "0.5.1",
+    "class-validator": "0.13.1",
+    "eslint-plugin-import": "^2.23.4",
```
TimoGlastra, line 33:
```suggestion
```

.eslintrc.js
```diff
@@ -22,6 +22,7 @@ module.exports = {
         alwaysTryTypes: true,
       },
     },
+    'import/core-modules': ['indy-vdr-test-nodejs'],
```
TimoGlastra, line 25: We shouldn't need this I think. What is the error you're getting?



```suggestion
```

Vickysomtee: I am getting this no-extranous-dependencies error from lint

I figured out it's from this

```
try {
  require('indy-vdr-test-nodejs')
} catch (error) {
  throw new Error('Error registering nodejs bindings for Indy VDR')
}

```

TimoGlastra: ah you can add an eslint ignore above the import, it's fine in this case

Vickysomtee: This is the full error message

`'indy-vdr-test-nodejs' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies`

TimoGlastra: This should work:

```
try {
  // eslint-disable-next-line import/no-extraneous-dependencies
  require('indy-vdr-test-nodejs')
} catch (error) {
  throw new Error('Error registering nodejs bindings for Indy VDR')
}
```

TimoGlastra (approved):

packages/indy-vdr/src/pool/IndyVdrPool.ts
```diff
@@ -0,0 +1,182 @@
...
+
+    // TAA can be null
+    if (taaData == null) {
+      this.authorAgreement = null
+      return null
+    }
+
+    // If TAA is not null, we can be sure AcceptanceMechanisms is also not null
+    const authorAgreement = taaData as Omit<AuthorAgreement, 'acceptanceMechanisms'>
+    const acceptanceMechanisms = acceptanceMechanismResponse.result.data as AcceptanceMechanisms
```
TimoGlastra, line 174: Cast to unknown first until we have a new release  for indy-vdr

```suggestion
     // FIME: remove cast when https://github.com/hyperledger/indy-vdr/pull/142 is released
    const acceptanceMechanisms = acceptanceMechanismResponse.result.data as unknown as AcceptanceMechanisms
```

packages/indy-vdr/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "@aries-framework/indy-vdr",
+  "main": "build/index",
+  "types": "build/index",
+  "version": "0.3.3",
```
TimoGlastra, line 5:
```suggestion
  "version": "0.3.3",
  "private": true,
```

genaris:
I think it would be nice to use @hyperledger/indy-vdr-* packages now that they are going to be released in NPM after the v0.4.0-dev.2 release (https://github.com/hyperledger/indy-vdr/releases/tag/v0.4.0-dev.2)

TimoGlastra:
> I think it would be nice to use @hyperledger/indy-vdr-* packages now that they are going to be released in NPM after the v0.4.0-dev.2 release ([hyperledger/indy-vdr@`v0.4.0-dev.2` (release)](https://github.com/hyperledger/indy-vdr/releases/tag/v0.4.0-dev.2))

It doesn't quite work yet, so I'm going to merge this once CI passes and once the build is working we can make a follow up PR.

Issue #1162, opened 2022-12-15
Start using @internal, @beta, @alpha, @public tag on classes, types, etc...


TimoGlastra:
- https://tsdoc.org/pages/tags/internal/
- https://tsdoc.org/pages/tags/beta/
- https://tsdoc.org/pages/tags/alpha
- https://tsdoc.org/pages/tags/public

PR #1163, opened 2022-12-15
refactor(action-menu): clearly mark public api

TimoGlastra:
clearly mark the public and internal api of the action-menu package. This is as a start and example to #1162.

Thoughts?

I've kept the repo, services and messages as internal, but the api, record, models are public as they're used in the public api in some way


packages/action-menu/src/__tests__/ActionMenuModule.test.ts
```diff
@@ -2,13 +2,10 @@ import type { DependencyManager, FeatureRegistry } from '@aries-framework/core'
...
-  ActionMenuApi,
-  ActionMenuModule,
-  ActionMenuRepository,
-  ActionMenuRole,
-  ActionMenuService,
-} from '@aries-framework/action-menu'
+import { ActionMenuRepository } from '../repository'
+import { ActionMenuService } from '../services'
+
+import { ActionMenuApi, ActionMenuModule, ActionMenuRole } from '@aries-framework/action-menu'
```
jakubkoci, line 8: Should we not import all directly from the package in a unit test inside the package? I mean, replace `@aries-framework/action-menu` with relative paths?

jakubkoci (approved):
Interesting idea. I still feel the ideal solution would be to use export/import to achieve such behavior, but I understand it might not be so easy in our case. It seems like reasonable step forward.

packages/action-menu/src/__tests__/ActionMenuModule.test.ts
TimoGlastra: Yes could do it like that!

PR #1173, opened 2022-12-16
feat: allow to connect with self

TimoGlastra:
This was previously possible in 0.1.0, but broken in 0.2.0. This allows to connect to yourelf using connections protocols and didexchange protocol


genaris (approved):
Good to have this feature back!

Issue #1155, opened 2022-12-13
Write migration script for mediator connection types


TimoGlastra:
Since 0.2.5 we store a connection type of mediator in the connection record when a mediator relationship is set-up. We should write a migration script for 0.3.0 to add this tag to all existing connection records that were created before 0.2.5

PR #1177, opened 2022-12-19
chore(migrations)!: connections 0.3.0 migration script and tests

amanji:
This PR introduces changes to how connection types are handled in connection records:
- Connection types were previously stored in tags under the `connectionType` label. Connection types are now directly part of the record.
- Methods to add, update and remove types in the connection service now apply directly to the record instead of tags.
- The `connectionType` property is now pluralized to `connectionTypes` to reflect the fact that the property accepts an `Array` of connection types.
- Methods in the connection API and service have also been pluralized accordingly.
- Relevant tests have been updated to reflect changes to the connection record, API and service and have been added for the connection migration script from AFJ 0.2 to 0.3.

## Related Issues

#1155


packages/core/src/storage/migration/updates/0.2-0.3/connection.ts
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * @param agent
+ */
+export async function migrateConnectionRecordToV0_3<Agent extends BaseAgent>(agent: Agent) {
+  agent.config.logger.info('Migrating connection records to storage version 0.3')
+  const connectionRepository = agent.dependencyManager.resolve(ConnectionRepository)
+  const mediationRepository = agent.dependencyManager.resolve(MediationRepository)
+
+  agent.config.logger.debug('Fetching all connection records from storage')
+  const allConnections = await connectionRepository.getAll(agent.context)
```
TimoGlastra, line 18: I don't think we have to resolve _all_ connection records, only the one that are for mediation records. Reason is that the number of connections _could_ be quite large. We should probably add a way to do pagination.

So we can resolve all mediaton records. Then for each of the mediation records, fetch the connection record, add the needed tags and then save the connection record

packages/core/src/storage/migration/updates/0.2-0.3/connection.ts
```diff
@@ -0,0 +1,61 @@
...
+  const mediatorConnectionIds = allMediators.map((mediator) => mediator.connectionId)
+
+  for (const connectionRecord of allConnections) {
+    agent.config.logger.debug(`Migrating connection record with id ${connectionRecord.id} to storage version 0.3`)
+
+    if (mediatorConnectionIds.includes(connectionRecord.id)) {
+      await migrateConnectionRecordMediatorTags(agent, connectionRecord)
+    }
+
+    await connectionRepository.update(agent.context, connectionRecord)
```
TimoGlastra, line 34: We don't have to update the connection record if no changes were applied, I think we can put it inside the if?

packages/core/src/storage/migration/updates/0.2-0.3/connection.ts
```diff
@@ -0,0 +1,61 @@
...
+ */
+export async function migrateConnectionRecordMediatorTags<Agent extends BaseAgent>(
+  agent: Agent,
+  connectionRecord: ConnectionRecord
+) {
+  agent.config.logger.debug(
+    `Migrating internal connection record ${connectionRecord.id} to storage version 0.3 with mediator tags`
+  )
+
+  const connectionTypeTags = (connectionRecord.getTag('connectionType') || []) as [string]
```
TimoGlastra, line 55: If we use `connectionRecord.getTags().connectionType` it will automatically infer the type.

TimoGlastra (commented):
One thing that I noticed during this PR review, and maybe as we are already writing a migration script we could update is that a) the property is `connectionType` and not `connectionTypes`. It was first singular and I think we forgot to update the naming to plural. Second is that it is only stored as tag. As tags are not always nice to work with, my preference always goes to adding it as either a value to the connection record itself (and add it as a tag in the `getTags()` method in the connection record). or setting it in the metadata, where we can still include it as a tag for the record.

Would you be willing to include those changes in this PR? So:
- Add an `_internal/connectionTypes` value to the record metadata with the `connectionTypes` property
- rename the `connectionType` to `connectionTypes` in the `DefaultConnectionTags`
- Update all methods that were using the connectionType tags to rather set the value on the record (connection service, mediator module, etc..)

And update the migration script to (i think the way you retrieve all connections, is already set up to do this):
- set the `connectionTypes` value on the connection record based on the `connectionType` tag
- remove the `connectionType` tag from the record (`connectionRecord.setTag('connectionType', undefined)`)


Otherwise I can look at making that additional PR, but I think it would make it easier to work with the connection types

genaris:
> One thing that I noticed during this PR review, and maybe as we are already writing a migration script we could update is that a) the property is `connectionType` and not `connectionTypes`. It was first singular and I think we forgot to update the naming to plural. Second is that it is only stored as tag. As tags are not always nice to work with, my preference always goes to adding it as either a value to the connection record itself (and add it as a tag in the `getTags()` method in the connection record). or setting it in the metadata, where we can still include it as a tag for the record.
> 
> Would you be willing to include those changes in this PR? So:
> 
> * Add an `_internal/connectionTypes` value to the record metadata with the `connectionTypes` property
> * rename the `connectionType` to `connectionTypes` in the `DefaultConnectionTags`
> * Update all methods that were using the connectionType tags to rather set the value on the record (connection service, mediator module, etc..)
> 
> And update the migration script to (i think the way you retrieve all connections, is already set up to do this):
> 
> * set the `connectionTypes` value on the connection record based on the `connectionType` tag
> * remove the `connectionType` tag from the record (`connectionRecord.setTag('connectionType', undefined)`)
> 
> Otherwise I can look at making that additional PR, but I think it would make it easier to work with the connection types

While I'm fine with current naming, I think it makes sense to rename it to `connectionTypes` but finally by reading your comment it was not clear to me if you are willing to add the property to the record itself (like connectionId,  threadId, imageUrl, etc) or to its metadata. In my opinion, working with metadata is as complicated as working with tags (and maybe more). The advantage of it is you can set an arbitrary object there, which is not required in this case.

Adding connectionTypes as a direct ConnectionRecord property could make it easier to work with, mostly when we are in the need of updating several ConnectionRecord properties at once (I don't see a common use case yet but it might be). For current use case I think it's not needed because usually, as an an end user, for setter we always need to call ConnectionRepository.update() afterwards so that's why the addConnectionTypes / removeConnectionTypes API/Service methods are convenient to deal with it even if the internals are not so straightforward. The getter, as you said in another comment is not so complicated anyway (just record.getTags().connectionType or the Api method getConnectionTypes(record).

amanji:
@genaris @TimoGlastra What if we employ the "crawl, walk, run strategy" here as it seems that this next release is time sensitive? How about I at least rename the metadata correctly and in a subsequent release we can discuss moving the data directly into the record?

TimoGlastra:
> @genaris @TimoGlastra What if we employ the "crawl, walk, run" strategy here as it seems that this next release is time sensitive? How about I at least rename the metadata correctly and in a subsequent release we can discuss moving the data directly into the record?

Yes makes sense. I was overly complicating things...

TimoGlastra:
> > One thing that I noticed during this PR review, and maybe as we are already writing a migration script we could update is that a) the property is `connectionType` and not `connectionTypes`. It was first singular and I think we forgot to update the naming to plural. Second is that it is only stored as tag. As tags are not always nice to work with, my preference always goes to adding it as either a value to the connection record itself (and add it as a tag in the `getTags()` method in the connection record). or setting it in the metadata, where we can still include it as a tag for the record.
> > Would you be willing to include those changes in this PR? So:
> > 
> > * Add an `_internal/connectionTypes` value to the record metadata with the `connectionTypes` property
> > * rename the `connectionType` to `connectionTypes` in the `DefaultConnectionTags`
> > * Update all methods that were using the connectionType tags to rather set the value on the record (connection service, mediator module, etc..)
> > 
> > And update the migration script to (i think the way you retrieve all connections, is already set up to do this):
> > 
> > * set the `connectionTypes` value on the connection record based on the `connectionType` tag
> > * remove the `connectionType` tag from the record (`connectionRecord.setTag('connectionType', undefined)`)
> > 
> > Otherwise I can look at making that additional PR, but I think it would make it easier to work with the connection types
> 
> While I'm fine with current naming, I think it makes sense to rename it to `connectionTypes` but finally by reading your comment it was not clear to me if you are willing to add the property to the record itself (like connectionId, threadId, imageUrl, etc) or to its metadata. In my opinion, working with metadata is as complicated as working with tags (and maybe more). The advantage of it is you can set an arbitrary object there, which is not required in this case.
> 
> Adding connectionTypes as a direct ConnectionRecord property could make it easier to work with, mostly when we are in the need of updating several ConnectionRecord properties at once (I don't see a common use case yet but it might be). For current use case I think it's not needed because usually, as an an end user, for setter we always need to call ConnectionRepository.update() afterwards so that's why the addConnectionTypes / removeConnectionTypes API/Service methods are convenient to deal with it even if the internals are not so straightforward. The getter, as you said in another comment is not so complicated anyway (just record.getTags().connectionType or the Api method getConnectionTypes(record).

@genaris you're right. This comment from me was a lot of thoughts going in all directions. Let's rename to plural naming now and maybe add it to the record in a later release. I'm fine with adding it to the record directly, and can understand the complexity of the record metadata. I think I'm not a fan of using the tags as a way to _store data_. I see it as away to _query data_, and we should call `getTags()` as little as possible. If we need to use `getTags()`, most of the times I think the data should actually be stored in the record (either directly or in metadata), and automatically set as a tag in the `getTags()` method.

genaris:
> 
> @genaris you're right. This comment from me was a lot of thoughts going in all directions. Let's rename to plural naming now and maybe add it to the record in a later release. I'm fine with adding it to the record directly, and can understand the complexity of the record metadata. I think I'm not a fan of using the tags as a way to _store data_. I see it as away to _query data_, and we should call `getTags()` as little as possible. If we need to use `getTags()`, most of the times I think the data should actually be stored in the record (either directly or in metadata), and automatically set as a tag in the `getTags()` method.

I think this is a very good point, as maybe we were too implementation-specific when deciding to use tags, as the original need solved by `connectionType` was to filter out Mediator connections and the we generalized it. I agree on handling it later in a separate PR/release.

packages/core/src/storage/migration/updates/0.2-0.3/connection.ts
amanji: Since we are adding the `connectionTypes` directly to the record, iterating through all records would be necessary

packages/core/src/storage/migration/updates/0.2-0.3/connection.ts
amanji: Since we are adding the connectionTypes directly to the record, updating each record would be necessary

packages/core/src/storage/migration/updates/0.2-0.3/connection.ts
amanji: This won't work after pluralizing the property in the `DefaultConnectionTags` type

amanji:
So I didn't see these additional comments before proceeding, but I've gone ahead and added the connectionTypes directly to the record and updated the methods accordingly. I have added a metadata type (which is not being used at this point) so I can remove that change to keep the PR clean. Let me know your thoughts

packages/core/src/modules/connections/repository/ConnectionMetadataTypes.ts
```diff
@@ -1,9 +1,13 @@
...
 export enum ConnectionMetadataKeys {
   UseDidKeysForProtocol = '_internal/useDidKeysForProtocol',
+  ConnectionTypes = '_internal/connectionTypes',
 }
 
 export type ConnectionMetadata = {
   [ConnectionMetadataKeys.UseDidKeysForProtocol]: {
     [protocolUri: string]: boolean
   }
+  [ConnectionMetadataKeys.ConnectionTypes]: Array<ConnectionType | string>
```
amanji, line 12: @TimoGlastra Should we just remove this for now?

packages/core/src/modules/connections/repository/ConnectionRecord.ts
```diff
@@ -38,6 +39,10 @@ export type DefaultConnectionTags = {
   theirDid?: string
   outOfBandId?: string
   invitationDid?: string
+  connectionTypes?: Array<ConnectionType | string>
```
amanji, line 42: While I have added this here, if we are doing away with adding connection types to tags, do we need this anymore?

packages/core/src/modules/connections/repository/ConnectionMetadataTypes.ts
TimoGlastra: Yes, let's remove it if it isn't being used

packages/core/src/modules/connections/repository/ConnectionRecord.ts
TimoGlastra: We still need to add them to the tags, it's just based on the record value now

amanji: I have changed the methods in the connection service to add and remove directly from the record now, I suppose I should update the `getTags` method in the record to also return the value for `this.connectionTypes`?

TimoGlastra (commented):
This looks great @amanji. Left some small final remarks

TimoGlastra (approved):

PR #1178, opened 2022-12-20
refactor: jsonld credential format improvements

TimoGlastra:
Some small improvements to the jsonld credential format services of things I discovered when using the service. It mostly includes:
- check the whole credential instead of the credentialSubject against the request
- check the created property against the proof (if it was defined)
- do not require a class as input for the credential detail options in the credentials api
- fix incorrect interfaces
- remove redundant checks

### DIFF
--- a/packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialDetailOptions.ts
+++ b/packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialDetailOptions.ts
@@ -1,11 +1,11 @@
 import { IsObject, IsOptional, IsString } from 'class-validator'
 
-export interface JsonLdOptionsCredentialStatusOptions {
+export interface JsonLdCredentialDetailCredentialStatusOptions {
   type: string
 }
 
-export class JsonLdOptionsCredentialStatus {
-  public constructor(options: JsonLdOptionsCredentialStatusOptions) {
+export class JsonLdCredentialDetailCredentialStatus {
+  public constructor(options: JsonLdCredentialDetailCredentialStatusOptions) {
     if (options) {
       this.type = options.type
     }
@@ -14,17 +14,17 @@ export class JsonLdOptionsCredentialStatus {
   public type!: string
 }
 
-export interface JsonLdOptionsRFC0593Options {
+export interface JsonLdCredentialDetailOptionsOptions {
   proofPurpose: string
   created?: string
   domain?: string
   challenge?: string
-  credentialStatus?: JsonLdOptionsCredentialStatus
+  credentialStatus?: JsonLdCredentialDetailCredentialStatus
   proofType: string
 }
 
-export class JsonLdOptionsRFC0593 {
-  public constructor(options: JsonLdOptionsRFC0593Options) {
+export class JsonLdCredentialDetailOptions {
+  public constructor(options: JsonLdCredentialDetailOptionsOptions) {
     if (options) {
       this.proofPurpose = options.proofPurpose
       this.created = options.created
@@ -55,5 +55,5 @@ export class JsonLdOptionsRFC0593 {
 
   @IsOptional()
   @IsObject()
-  public credentialStatus?: JsonLdOptionsCredentialStatus
+  public credentialStatus?: JsonLdCredentialDetailCredentialStatus
 }
--- a/packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormat.ts
+++ b/packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormat.ts
@@ -2,7 +2,6 @@ import type { JsonObject } from '../../../../types'
 import type { SingleOrArray } from '../../../../utils'
 import type { IssuerOptions } from '../../../vc/models/credential/Issuer'
 import type { CredentialFormat } from '../CredentialFormat'
-import type { JsonLdOptionsRFC0593 } from './JsonLdOptionsRFC0593'
 
 export interface JsonCredential {
   '@context': Array<string> | JsonObject
@@ -15,19 +14,53 @@ export interface JsonCredential {
   [key: string]: unknown
 }
 
-// this is the API interface (only)
-export interface JsonLdSignCredentialFormat {
+/**
+ * Format for creating a jsonld proposal, offer or request.
+ */
+export interface JsonLdCredentialDetailFormat {
   credential: JsonCredential
-  options: JsonLdOptionsRFC0593
+  options: {
+    proofPurpose: string
+    proofType: string
+  }
 }
 
-// use this interface internally as the above may diverge in future
-export interface SignCredentialOptionsRFC0593 {
-  credential: JsonCredential
-  options: JsonLdOptionsRFC0593
+// use empty object in the acceptXXX jsonld format interface so we indicate that
+// the jsonld format service needs to be invoked
+type EmptyObject = Record<string, never>
+
+/**
+ * Format for accepting a jsonld credential request. Optionally allows the verification
+ * method to use to sign the credential.
+ */
+export interface JsonLdAcceptRequestFormat {
+  verificationMethod?: string
 }
 
-export interface JsonVerifiableCredential extends JsonLdSignCredentialFormat {
+export interface JsonLdCredentialFormat extends CredentialFormat {
+  formatKey: 'jsonld'
+  credentialRecordType: 'w3c'
+  credentialFormats: {
+    createProposal: JsonLdCredentialDetailFormat
+    acceptProposal: EmptyObject
+    createOffer: JsonLdCredentialDetailFormat
+    acceptOffer: EmptyObject
+    createRequest: JsonLdCredentialDetailFormat
+    acceptRequest: JsonLdAcceptRequestFormat
+  }
+  formatData: {
+    proposal: JsonLdFormatDataCredentialDetail
+    offer: JsonLdFormatDataCredentialDetail
+    request: JsonLdFormatDataCredentialDetail
+    credential: JsonLdFormatDataVerifiableCredential
+  }
+}
+
+/**
+ * Represents a signed verifiable credential. Only meant to be used for credential
+ * format data interfaces.
+ */
+export interface JsonLdFormatDataVerifiableCredential extends JsonCredential {
   proof: {
     type: string
     proofPurpose: string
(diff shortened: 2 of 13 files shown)


packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialDetailOptions.ts
```diff
@@ -14,17 +14,17 @@ export class JsonLdOptionsCredentialStatus {
   public type!: string
 }
 
-export interface JsonLdOptionsRFC0593Options {
+export interface JsonLdCredentialDetailOptionsOptions {
```
blu3beri, line 17: Is the double options here required? It looks a bit funky.

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormat.ts
```diff
@@ -15,19 +14,53 @@ export interface JsonCredential {
...
+  }
 }
 
-// use this interface internally as the above may diverge in future
-export interface SignCredentialOptionsRFC0593 {
-  credential: JsonCredential
-  options: JsonLdOptionsRFC0593
+// use empty object in the acceptXXX jsonld format interface so we indicate that
+// the jsonld format service needs to be invoked
+type EmptyObject = Record<string, never>
```
blu3beri, line 30: Not sure how it would work the modularisation, but would it make sense to abstract these types so we can reuse them elsewhere? Might be a bit overengineerd, but I think we have quite a bit of object types in different files defined as well.

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
```diff
@@ -229,9 +222,9 @@ export class JsonLdCredentialFormatService implements CredentialFormatService<Js
 
     const options = credentialRequest.options
 
-    if (options.challenge || options.domain || options.credentialStatus) {
+    if (options.challenge || options.domain || options.credentialStatus || options.created) {
       throw new AriesFrameworkError(
-        'The fields challenge, domain and credentialStatus not currently supported in credential options '
+        'The fields challenge, domain, created and credentialStatus not currently supported in credential options'
```
blu3beri, line 227: Might be nice in the error, or log, which options were supplied that are not supported. Can save some time with debugging.

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
```diff
@@ -330,41 +315,49 @@ export class JsonLdCredentialFormatService implements CredentialFormatService<Js
...
   }
 
-  private compareCredentialSubject(
+  private verifyReceivedCredentialMatchesRequest(
     credential: W3cVerifiableCredential,
-    requestAsJson: JsonLdSignCredentialFormat
+    request: JsonLdFormatDataCredentialDetail
   ): void {
-    const request = JsonTransformer.fromJSON(requestAsJson.credential, W3cCredential)
+    // const request = JsonTransformer.fromJSON(requestAsJson.credential, W3cCredential)
```
blu3beri, line 322: Commented code :-).

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
```diff
@@ -330,41 +315,49 @@ export class JsonLdCredentialFormatService implements CredentialFormatService<Js
...
 
-    if (!objectEquality(credential.credentialSubject, request.credentialSubject)) {
-      throw new AriesFrameworkError('Received credential subject does not match subject from credential request')
+    const jsonCredential = JsonTransformer.toJSON(credential)
+    delete jsonCredential.proof
```
blu3beri, line 325: Do we know if `toJSON` does a shallow or deep copy? Deleting the proof field with a shallow copy would be problematic.

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
```diff
@@ -330,41 +315,49 @@ export class JsonLdCredentialFormatService implements CredentialFormatService<Js
...
+
+    // Check whether the received credential (minus the proof) matches the credential request
+    if (!areObjectsEqual(jsonCredential, request.credential)) {
```
blu3beri, line 328: Can we move this check below the others? This is the most computationally heavy. If this is the main reason for invalid credentials it might be fine, but if all cases are as common as the next, it would be good to do this check at the last validation step.

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
```diff
@@ -330,41 +315,49 @@ export class JsonLdCredentialFormatService implements CredentialFormatService<Js
...
+      throw new AriesFrameworkError('Credential proof arrays are not supported')
+    }
 
-    const requestProof = request.options
+    if (request.options.created && credential.proof.created !== request.options.created) {
+      throw new AriesFrameworkError('Received credential proof created does not match created from credential request')
+    }
 
-    if (credentialProof.domain !== requestProof.domain) {
+    if (credential.proof.domain !== request.options.domain) {
```
blu3beri, line 340: Can't we do a `areObjectsEqual` on credential.proof? Or do we want the explicit error messages?

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
```diff
@@ -406,27 +398,10 @@ export class JsonLdCredentialFormatService implements CredentialFormatService<Js
...
-        this.compareProofs(credential.proof, requestAsJson)
-        return true
-      } catch (error) {
-        return false
-      }
-    }
+  public shouldAutoRespondToCredential() {
+    // The credential is already compared in the processCredential method, for now
+    // we will always return true.
+    return true
```
blu3beri, line 404: Not the biggest fan of this. I think double verification is better here. I don't think that it is a very expensive operation and if, somehow, this is ever used separate from `processCredential` (via a custom module or something else) it will cause unexpected behaviour.

blu3beri (changes requested):
Liking the cleanup! Mainly just some notes and one blocker IMO about the auto acceptance for the credential. I might miss some context for that one, but it feels weird to me.

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
TimoGlastra: Sure, can add the verification back in. It's just that a lot of the methods are written with the knowledge the knowledge that the processXXX methods are called up front. These handle initial processing and validation of the objects. But I agree it looks weird.

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
TimoGlastra: The right side of the equation are the request.options. Although overlapping with a lot of properties from the proof, it's not the same object (and thus will fail)

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
TimoGlastra: Oh my... 😳

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
TimoGlastra: Hmm makes sense. This was the cleanest way I could come up with to show which fields were the issue, without adding 4 if statements (which may be fine??). Let me know if you have a better approach:

```ts
    // Get a list of fields found in the options that are not supported at the moment
    const unsupportedFields = ['challenge', 'domain', 'credentialStatus', 'created'] as const
    const foundFields = unsupportedFields.filter((field) => options[field])

    if (foundFields) {
        `The fields ${foundFields.join(', ')} are not currently supported in credential options`
      )
    }
```

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
TimoGlastra: Added the check here as well

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormat.ts
TimoGlastra: Yes, maybe. As we're going to abstract more things out of core I'm more leaning to having some duplication. We now have to duplicate them whenver we extract it out of core, as exporting an `EmptyObject` type doesn't make sense.

I'm not too worried about such simple types

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialDetailOptions.ts
TimoGlastra: The class is already called `JsonLdCredentialDetailOptions`, so yes it is needed. Gave it some thought but couldn't come up with a good solution and this is at least consistent

TimoGlastra:
@blu3beri addressed feedback with some questions

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
genaris: I would slightly rephrase error message to adapt a bit better for the case where is a single unsupported field: `Some fields are not currently supported in credential options: ${foundFields.join(', ')}`

TimoGlastra: updated it to your suggestion @genaris

blu3beri: I think the function you used is okay @TimoGlastra (I tried to do it with Object.keys but it did not change much) and I agree with the comment from @genaris

packages/core/src/modules/credentials/formats/jsonld/JsonLdCredentialFormatService.ts
blu3beri: Ah yeah so It would require a new function that checks if one is fully included in the other, might a bit overkill for here. Can see it as resolved.

blu3beri (approved):

PR #1182, opened 2022-12-22
feat: adding trust ping events and trust ping command

reflectivedevelopment:
Signed-off-by: Kim Ebert <kim@indicio.tech>


packages/core/src/modules/connections/ConnectionsApi.ts
```diff
@@ -227,6 +230,42 @@ export class ConnectionsApi {
...
+
+  /**
+   * Send a trust ping to an established connection
+   *
+   * @param connectionId the id of the connection for which to accept the response
+   * @param responseRequested do we want a response to our ping
+   * @param withReturnRouting do we want a response at the time of posting
+   * @returns boolean indicating success or failure
+   */
+  public async trustPing(connectionId: string, responseRequested=true, withReturnRouting=false): Promise<boolean> {
```
TimoGlastra, line 242: For other apis we usually use objects when there's multiple config parameters. Otherwise you'll just see a call with (false, true). If we make it an object, you can directly see where the options are for.

```suggestion
export interface SendPingOptions {
  responseRequested?: boolean
}

  public async sendPing(connectionId: string, { responseRequested = true }: SendPingOptions): Promise<boolean> {
```

packages/core/src/modules/connections/ConnectionsApi.ts
```diff
@@ -227,6 +230,42 @@ export class ConnectionsApi {
...
+    const connectionRecord = await this.connectionService.getById(this.agentContext, connectionId)
+
+    let outboundMessageContext
+    if (connectionRecord.state === DidExchangeState.Completed) {
```
TimoGlastra, line 246: The createTrustPing message check the state of the connection record and asserts it is ResponseReceived or Completed. We can leave this check here. 

That way we also don't have to return true / false

packages/core/src/modules/connections/ConnectionsApi.ts
```diff
@@ -227,6 +230,42 @@ export class ConnectionsApi {
...
+
+      const { message } = await this.connectionService.createTrustPing(this.agentContext, connectionRecord, {
+        responseRequested: responseRequested,
+      })
+      // Disable return routing as we don't want to receive a response for this message over the same channel
+      // This has led to long timeouts as not all clients actually close an http socket if there is no response message
+      if (!withReturnRouting) {
+        message.setReturnRouting(ReturnRouteTypes.none)
+      }
```
TimoGlastra, line 255: Where are you using the trust ping for? I'm thinking whether we can come up with a general way of doing this. No other method in the API of AFJ has this option, and it feels a bit use case specific specific. Thinking how we can generically solve this without needing to add it to all other messages. Also because this seems more of an issue of the other agents, rather than an issue of AFJ. We just instruct the other agent that they _may_ return it over the same channel. For ACA-Py this has been fixed in the main branch, but in any release yet sadly: https://github.com/hyperledger/aries-cloudagent-python/pull/1853

Could we base it off the `responseRequested` parameter maybe for now?

packages/core/src/modules/connections/ConnectionsApi.ts
TimoGlastra, line 242: The DIDComm V2 PR also adds support for the trust ping method. Can we make sure to align this implementation with that method? The method name is `sendPing`.

packages/core/src/modules/connections/TrustPingEvents.ts
```diff
@@ -0,0 +1,22 @@
...
+export enum TrustPingEventTypes {
+  TrustPingRequestEvent = 'TrustPingRequestEvent',
+  TrustPingResponseEvent = 'TrustPingResponseEvent',
+}
+
+export interface TrustPingRequestEvent extends BaseEvent {
+  type: typeof TrustPingEventTypes.TrustPingRequestEvent
+  payload: {
+    connectionRecord: ConnectionRecord
+  }
```
TimoGlastra, line 14: Can we also add the message to the event? That will make it possible to track based on the threadId

```suggestion
  payload: {
    connectionRecord: ConnectionRecord
    message: TrustPingMessage
  }
```

packages/core/src/modules/connections/TrustPingEvents.ts
```diff
@@ -0,0 +1,22 @@
+import type { BaseEvent } from '../../agent/Events'
+import type { DidExchangeState } from './models'
+import type { ConnectionRecord } from './repository/ConnectionRecord'
+
+export enum TrustPingEventTypes {
+  TrustPingRequestEvent = 'TrustPingRequestEvent',
+  TrustPingResponseEvent = 'TrustPingResponseEvent',
```
TimoGlastra, line 7: Received is more consistent with existing event names in AFJ:

```suggestion
  TrustPingReceivedEvent = 'TrustPingReceivedEvent',
  TrustPingResponseReceivedEvent = 'TrustPingResponseReceivedEvent',
```

packages/core/src/modules/connections/TrustPingEvents.ts
```diff
@@ -0,0 +1,22 @@
...
+  type: typeof TrustPingEventTypes.TrustPingRequestEvent
+  payload: {
+    connectionRecord: ConnectionRecord
+  }
+}
+
+export interface TrustPingResponseEvent extends BaseEvent {
+    type: typeof TrustPingEventTypes.TrustPingResponseEvent
+    payload: {
+      connectionRecord: ConnectionRecord
```
TimoGlastra, line 20: same here, if we can add the message to the event, we can determine to which message we got a response

packages/core/src/modules/connections/ConnectionsApi.ts
```diff
@@ -227,6 +230,42 @@ export class ConnectionsApi {
...
+      outboundMessageContext = new OutboundMessageContext(message, {
+        agentContext: this.agentContext,
+        connection: connectionRecord,
+      })
+
+      await this.messageSender.sendMessage(outboundMessageContext)
+      return true
```
TimoGlastra, line 262: If we can return the `message`, we can get the threadId and wait for a `TrustPingResponseReceived` event with the same `message.threadId`

packages/core/src/modules/connections/services/TrustPingService.ts
```diff
@@ -20,6 +39,16 @@ export class TrustPingService {
 
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   public processPingResponse(inboundMessage: InboundMessageContext<TrustPingResponseMessage>) {
-    // TODO: handle ping response message
+
+    const { agentContext, connection } = inboundMessage
```
TimoGlastra, line 43: You can do:

```
const connection = inboundMessage.assertReadyConnection()
```

The trust ping handler already checks whether there is a connection, so this should never error but if the handler ever changes makes sure an error is thrown instead of it being silently ignored

TimoGlastra (changes requested):
Thanks for the PR @reflectivedevelopment! 

Could you fix the prettier and eslint errors? Most can be fixed by running `yarn lint --fix`. Also could you add a test that sends a trust ping to another agent, and another agent responding with a response and also checking whether the events are received?

packages/core/src/modules/connections/ConnectionsApi.ts
reflectivedevelopment: The trust ping can be used as a diagnostic tool. Being able to specify the return route allows us to perform different diagnostics. For example, we can test the return route is behaving as expected.

In the particular use case, I'm trying to ensure that the websocket of the mediator is responding correctly, and test the latency of the mediator.

swcurran: @reflectivedevelopment -- are you working on figuring out a scalable mediator (Aries in general) deployment?  E.g. persistent queue/shared cache enabled with support for websockets and return-route?  We're working on that, and would like to collaborate on findings/recommendations.

TimoGlastra: If that's the case you may be able to get away with a custom plugin / module that hooks into the metal of AFJ?

Diagnostics / Latency tests seem like an advanced use case to me. 

What I'm getting at is that I want to be careful with adding every needed property to a specific method when a single person needs it. This makes the API inconsistent, and adds complexity where it may not be needed. That said, if you think it is generally useful to have this, I'm okay with adding such an option -- but as I said, it feels like an advanced use case that may deserve it's own method / api

packages/core/src/modules/connections/ConnectionsApi.ts
reflectivedevelopment: Fixed

packages/core/src/modules/connections/ConnectionsApi.ts
reflectivedevelopment: Fixed

packages/core/src/modules/connections/ConnectionsApi.ts
reflectivedevelopment: I think supporting testing frameworks is an important aspect of the AFJ framework, and seeing that there is little overhead in this change, it seems the most reasonable approach to me.

packages/core/src/modules/connections/ConnectionsApi.ts
reflectivedevelopment: I adjusted to model off of sendPing, so I think it lines up better with sendPing.

packages/core/src/modules/connections/TrustPingEvents.ts
reflectivedevelopment: Done

packages/core/src/modules/connections/TrustPingEvents.ts
reflectivedevelopment: Done

packages/core/src/modules/connections/TrustPingEvents.ts
reflectivedevelopment: Done

packages/core/src/modules/connections/ConnectionsApi.ts
reflectivedevelopment: Done

packages/core/src/modules/connections/services/TrustPingService.ts
reflectivedevelopment: I'm not actually 100% sure that we need to require the connection to be complete to send the trust ping, so perhaps this actually isn't needed.  🤔

packages/core/src/modules/connections/ConnectionsApi.ts
```diff
@@ -28,10 +28,15 @@ import {
   TrustPingMessageHandler,
   TrustPingResponseMessageHandler,
 } from './handlers'
-import { HandshakeProtocol } from './models'
+import { HandshakeProtocol, DidExchangeState } from './models'
```
TimoGlastra, line 31: image

packages/core/src/modules/connections/ConnectionsApi.ts
```diff
@@ -227,6 +232,37 @@ export class ConnectionsApi {
     return connectionRecord
   }
 
+  /**
+   * Send a trust ping to an established connection
+   *
+   * @param connectionId the id of the connection for which to accept the response
+   * @param responseRequested do we want a response to our ping
+   * @param withReturnRouting do we want a response at the time of posting
+   * @returns boolean indicating success or failure
```
TimoGlastra, line 241: This method doesn't return a boolean anymore

packages/core/src/modules/connections/ConnectionsApi.ts
```diff
@@ -227,6 +232,37 @@ export class ConnectionsApi {
...
+
+    const { message } = await this.connectionService.createTrustPing(this.agentContext, connection, {
+      responseRequested: responseRequested,
+    })
+
+    // Disable return routing as we don't want to receive a response for this message over the same channel
+    // This has led to long timeouts as not all clients actually close an http socket if there is no response message
+    if (!withReturnRouting) {
+      message.setReturnRouting(ReturnRouteTypes.none)
+    }
```
TimoGlastra, line 257: If return routing is true, we should explicitly enable It I think. Otherwise it can lead to unexpected behaviour (I pass `withReturnRouting: true`, but it won't actually use return routing because we have an endpoint.

I would make the default of `withReturnRouting` undefined, then we have three options:
- `withReturnRouting: undefined` -- we don't set the return routing
- `withReturnRouting: true` -- we set the return routing to `ReturnRouteTypes.All`
- `withReturnRouting: false` -- we set the return routing to `ReturnRoutingTypes.None`

Another option would be to make the parameter `returnRouteType` and actually allow the user to pass thread / none / all

packages/core/src/modules/connections/TrustPingEvents.ts
```diff
@@ -0,0 +1,25 @@
+import type { BaseEvent } from '../../agent/Events'
+import type { TrustPingMessage, TrustPingResponseMessage } from './messages'
+import type { DidExchangeState } from './models'
```
TimoGlastra, line 3: Did exchange state is unused

packages/core/src/modules/connections/services/TrustPingService.ts
```diff
@@ -20,6 +37,16 @@ export class TrustPingService {
 
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
TimoGlastra, line 38:
```suggestion
```

packages/core/src/modules/connections/services/TrustPingService.ts
TimoGlastra: In what cases is a connection not required? We can only send trust ping messages to completed connections, so if we were to receive a response we MUST have a connection, otherwise the ping response is not in response to a ping we sent

TimoGlastra (commented):
Some final nits

packages/core/src/modules/connections/ConnectionsApi.ts
reflectivedevelopment: Fixed

packages/core/src/modules/connections/ConnectionsApi.ts
reflectivedevelopment: Done

packages/core/src/modules/connections/TrustPingEvents.ts
reflectivedevelopment: Done

packages/core/src/modules/connections/services/TrustPingService.ts
reflectivedevelopment: Done

packages/core/src/modules/connections/services/TrustPingService.ts
reflectivedevelopment: Fixed

packages/core/src/modules/connections/ConnectionsApi.ts
reflectivedevelopment: Done

TimoGlastra (approved):

TimoGlastra:
Thanks @reflectivedevelopment!

PR #1184, opened 2022-12-26
fix: missing migration script and exports

genaris:
Unfortunately, `migrateDidRecordToV0_3` was not added to `updateV0_2ToV0_3` script, meaning that DID records were not upgraded, making DIDs non-resolvable.

This PR fixes that and also exposes a few classes that are mentioned in [the documentation](https://aries.js.org/guides/updating/update-assistant) but were no exported directly.


TimoGlastra (commented):
Dangit. We may need to add it as a separate upgrade (like 0.3-patch.1) as agents that have already upgraded won't be able to upgrade again. Thoughts?

TimoGlastra:
And time to add some tests for this!! Will add soon

genaris:
I was thinking that 

> Dangit. We may need to add it as a separate upgrade (like 0.3-patch.1) as agents that have already upgraded won't be able to upgrade again. Thoughts?

Well, at the first moment I had the problem that I was on 0.3.0.alpha.xx and when upgrading to 0.3.0 "release" it already thought that I had the last storage version while of course it wasn't. 

I think it could be a bit complicated to get version granularity up to alpha level, but maybe it could be possible to do it up to patch version. This way we could add this `migrateDidRecordToV0_3` to an `updateV0_3ToV0_3_1` that will add an additional step.

genaris:
> And time to add some tests for this!! Will add soon

I've added a test to 0.2 migration considering it will go straight from 0.2.x to 0.3.1, and also a 0.3 migration test that expects to update only DIDs. Working fine also in a 'real-world' 0.3.0 wallet so I hope this will fix the problem as long as most users skip 0.3.0 when upgrading their wallets.

What do you think on supporting patch versions in migrations? Theoretically we'll not need it and we will go now from storage version 0.3.1 to -> 0.4 directly, but if something unfortunate happens (as this issue) we can cover it easily.

TimoGlastra (approved):
This is great, thanks @genaris. I think patch version migrations should be avoided as it makes updates harder, but yeah totally agree that in such cases it's good to have patch updates

PR #1188, opened 2023-01-03
fix(credentials): typing if no modules provided

### LINKED ISSUE #1187: Types mismatch for `credentials` property on `Agent`
# Overview
After upgrading to `"@aries-framework/core": "^0.3.1"` the types on Agents `credentials` are resulting into `any`.
Looking at the migration docs from 0.2.X to 0.3.X I couldn't find any changes regarding to `credentials`.

## Reproduce
```ts
const agent = new Agent({ config, dependencies: agentDependencies })
await agent.credentials // typed as `any`
```

TimoGlastra:
Fixes the typing of the credentials module on the agent when no modules objects has been provided to the agent constructor.

Fixes #1187


morrieinmaas (commented):
LGTM

packages/core/src/agent/AgentModules.ts
```diff
@@ -101,7 +102,9 @@ export type AgentApi<Modules extends ModulesMap> = {
 export type CustomOrDefaultApi<
   CustomModuleType,
   DefaultModuleType extends ApiModule
-> = CustomModuleType extends ApiModule
+> = IsAny<CustomModuleType> extends true
+  ? InstanceType<DefaultModuleType['api']>
```
blu3beri, line 106: This confuses me a bit.

It seems like we are checking if the CustomModuleType is supplied and if so, we use the `DefaultModuleType['api']`. Should that not be `CustomModuleType['api']` and if not, maybe we should rename that type if we also use it for custom modules.

Small note, I am not very familiar with the generic module type setup so I might have missed something ;).

packages/core/src/types.ts
```diff
@@ -193,3 +193,8 @@ export type FlatArray<Arr> = Arr extends ReadonlyArray<infer InnerArr> ? FlatArr
  * Get the awaited (resolved promise) type of Promise type.
  */
 export type Awaited<T> = T extends Promise<infer U> ? U : never
+
+/**
+ * Type util that returns `true` or `false` based on whether the input type `T` is of type `any`
+ */
+export type IsAny<T> = unknown extends T ? ([keyof T] extends [never] ? false : true) : false
```
blu3beri, line 200: Looking at this makes me think that this should not work, but it does haha. Quite interesting.

blu3beri (approved):
One minor comment.

packages/core/src/types.ts
TimoGlastra: Yeah this comes straight from Stack overflow. No idea why it works. I can add a link to th so post though

PR #1197, opened 2023-01-10
feat: add minimal oidc-client package

karimStekelenburg:
This PR adds the very minimum of the new `oidc-client` package. It doesn't contain any logic so far.


packages/oidc-client/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "@aries-framework/oidc-client",
```
TimoGlastra, line 2: Thinking whether oidc-client is unique enough. OIDC is something on it's own, and OIDC4VCI and OIDC4VP are specifically for SSI. We're not building a generic oidc client in this case. Not sure what a good name would be though

packages/oidc-client/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "test": "jest"
+  },
+  "dependencies": {
+    "@sphereon/openid4vci-client": "^0.3.4",
+    "class-transformer": "0.5.1",
+    "class-validator": "0.13.1",
+    "rxjs": "^7.2.0"
+  },
+  "peerDependencies": {
+    "@aries-framework/core": "0.2.5"
```
TimoGlastra, line 33: I think this needs to be `0.3.2`?

packages/oidc-client/src/OidcClientApi.ts
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * @public
+ */
+@injectable()
+export class OidcClientApi {
+  private agentContext: AgentContext
+  private oidcClientService: OidcClientService
+
+  public constructor(agentContext: AgentContext, oidcClientService: OidcClientService) {
+    ;(this.agentContext = agentContext), (this.oidcClientService = oidcClientService)
```
TimoGlastra, line 14:
```suggestion
    this.agentContext = agentContext
    this.oidcClientService = oidcClientService
```

packages/oidc-client/src/OidcClientEvents.ts
```diff
@@ -0,0 +1,21 @@
...
+//  * @public
+//  */
+// export enum ActionMenuEventTypes {
+//   ActionMenuStateChanged = 'ActionMenuStateChanged',
+// }
+
+// /**
+//  * @public
+//  */
+// export interface ActionMenuStateChangedEvent extends BaseEvent {
```
TimoGlastra, line 15: Can we remove this for now? Would like to prevent commented code

packages/oidc-client/src/OidcClientModule.ts
```diff
@@ -0,0 +1,25 @@
...
+  /**
+   * Registers the dependencies of the question answer module on the dependency manager.
+   */
+  public register(dependencyManager: DependencyManager) {
+    // Api
+    dependencyManager.registerContextScoped(OidcClientApi)
+
+    // Services
+    dependencyManager.registerSingleton(OidcClientService)
+    dependencyManager.registerSingleton(W3cCredentialService)
```
TimoGlastra, line 23: The W3cCredentialService is already registered by core. Were you running into issues if it isn't registered?

TimoGlastra (commented):
Nice. Like the small PR approach. Makes it MUCH easier to review 👍

packages/oidc-client/package.json
blu3beri: Yeah I agree with this. I prefer to be a bit too verbose/explicit than incorrect. How would all the packages be named?

OIDC4VCI-(client|server)
OIDC4VP-(client|server)

So 4 packages?

TimoGlastra: I think it'll be two packages. One for client (issuance + proving) and one for server (issuance + verification). That feels like the right tradeoff (hard to include server components in the client), but also keep more simplicity (instead of having 4 packagers).

Would you lean towards 4 packages?

blu3beri: No my preference is for two packages (100% separate client and server). Just more for the naming as we do not want `OIDC-client` but `OIDC4VCI-client` is incorrect as it also does VP.

`OIDC4VCIVP-client` is getting a bit "odd". 

Context of `@aries-framework/` might imply enough that it is about the VCI/VP flow, but hard to say.

packages/oidc-client/src/OidcClientModule.ts
karimStekelenburg: Oh I assumed by calling `registerSingleton` it will make sure it only gets included once if it has already been added. I can remove it, just wasn't sure.

packages/oidc-client/package.json
karimStekelenburg: Yes I was thinking the same. It really depends on where the VP stuff will go. That's why I kept it generic for now. I could change it to `oidc4vc-client` to at least make it VC-specific. Thoughts?

packages/oidc-client/package.json
karimStekelenburg: Good catch!

karimStekelenburg:
> Nice. Like the small PR approach. Makes it MUCH easier to review 👍

I know right? I'm actually jealous, would have loved to review this myself 💯

packages/oidc-client/package.json
TimoGlastra: Hmmm maybe `oidc4v(c|p)`? 😆 

But oidc4vc sounds good I think.

packages/oidc-client/package.json
TimoGlastra: I've moved them to `dependencies` in all packages as `peerDependencies` don't get updated. As we're alsways using the exact same versions there's not much benefit to using peer deps

swcurran:
One question about this.  The Open ID folks at IIW were talking about "OpenID4VCs" and had deliberately dropped the "OpenID Connect" component.  AFAIU -- they are using underlying OAuth capabilities but are building new protocols that are not based on on OIDC. Is the goal of this to align with OpenID4VCs work, or is this to integrate OIDC using VCs into Aries (similar to oidc-vc-authn, and the work that IDunion did).  Depending on your answer, you might want to change the names of the modules.

packages/oidc-client/src/OidcClientModule.ts
TimoGlastra: It won't cause Issues I think, but if you register the W3cCredentialService, you should also register the W3cCredentialRepository, etc.. Otherwise it will still break if the W3c module isn't registered

TimoGlastra:
> One question about this. The Open ID folks at IIW were talking about "OpenID4VCs" and had deliberately dropped the "OpenID Connect" component. AFAIU -- they are using underlying OAuth capabilities but are building new protocols that are not based on on OIDC. Is the goal of this to align with OpenID4VCs work, or is this to integrate OIDC using VCs into Aries (similar to oidc-vc-authn, and the work that IDunion did). Depending on your answer, you might want to change the names of the modules.

Good note, the spec is also called OpenID for Verifiable Credentials (no Connect). 

This also solves our issue. There's a specific page called "OpenID For Verifiable Credentials (codenamed `openid4vc`), which lists all specs related to open id and vcs:

https://openid.net/openid4vc/

So it should maybe be `@aries-framework/openid4vc-client`?

packages/oidc-client/package.json
karimStekelenburg: Based on the comment below I have changed the package name to `@aries-framework/openid4vc-client`

packages/oidc-client/src/OidcClientModule.ts
karimStekelenburg: I have removed it for now. Will add it again if I run into issues down the line.

karimStekelenburg:
Thanks for the comments on naming @TimoGlastra and @swcurran. I have changed the package name to `@aries-framework/openid4vc-client`. If we want to be consistent, the file names should also be updated (e.g. `OidcClientService` -> `OpenId4VcClientService`). I'm in favor of this change, but it does make them quite long though. Thoughts?

packages/openid4vc-client/README.md
```diff
@@ -0,0 +1,37 @@
+<p align="center">
+  <br />
+  <img
+    alt="Hyperledger Aries logo"
+    src="https://raw.githubusercontent.com/hyperledger/aries-framework-javascript/aa31131825e3331dc93694bc58414d955dcb1129/images/aries-logo.png"
+    height="250px"
+  />
+</p>
+<h1 align="center"><b>Aries Framework JavaScript Open ID Connect Client Module</b></h1>
```
blu3beri, line 9: Still the old name, correct?

packages/openid4vc-client/README.md
```diff
@@ -0,0 +1,37 @@
...
+    <a href="https://www.npmjs.com/package/@aries-framework/openid4vc-client"
+    ><img
+      alt="@aries-framework/openid4vc-client version"
+      src="https://img.shields.io/npm/v/@aries-framework/openid4vc-client"
+  /></a>
+
+</p>
+<br />
+
+Open ID Connect Client plugin for [Aries Framework JavaScript](https://github.com/hyperledger/aries-framework-javascript.git).
```
blu3beri, line 31: Same here, also `.git` can be omitted from the url.

packages/openid4vc-client/package.json
```diff
@@ -0,0 +1,40 @@
...
+    "compile": "tsc -p tsconfig.build.json",
+    "prepublishOnly": "yarn run build",
+    "test": "jest"
+  },
+  "dependencies": {
+    "@aries-framework/core": "0.3.2",
+    "@sphereon/openid4vci-client": "^0.3.6",
+    "class-transformer": "0.5.1",
+    "class-validator": "0.13.1",
+    "rxjs": "^7.2.0"
```
blu3beri, line 31: Is this dependent on rxjs?

blu3beri (approved):
Very small points re dependency and the new name.

packages/openid4vc-client/README.md
karimStekelenburg: Nope, nice catch

PR #1200, opened 2023-01-10
feat(indy-sdk): add indy-sdk package

TimoGlastra:
This commits add the `@aries-framework/indy-sdk` package. It implements all the new interfaces based on the anonreds package. Testing is still light, but I want to add it in upcoming prs when we add the anoncreds api.

The classes aren't used yet, and all old Indy implementations are still in the core package. Once we have finished the new classes we can deprecate the old ones and point to the new ones.


packages/indy-sdk/src/IndySdkModuleConfig.ts
```diff
@@ -0,0 +1,45 @@
+import type * as IndySdk from 'indy-sdk'
```
karimStekelenburg, line 1: Any reason this is imported as `IndySDK` as opposed to `Indy` in other parts of the code?

packages/indy-sdk/src/anoncreds/services/IndySdkHolderService.ts
```diff
@@ -0,0 +1,320 @@
...
+
+import { IndySdkError, isIndyError } from '../../error'
+import { IndySdk, IndySdkSymbol } from '../../types'
+import { assertIndySdkWallet } from '../../utils/assertIndySdkWallet'
+import { getIndySeqNoFromUnqualifiedCredentialDefinitionId } from '../utils/identifiers'
+import {
+  indySdkCredentialDefinitionFromAnonCreds,
+  indySdkRevocationRegistryDefinitionFromAnonCreds,
+  indySdkSchemaFromAnonCreds,
+} from '../utils/transform'
```
karimStekelenburg, line 27: Nice, much cleaner 🚀

packages/indy-sdk/src/storage/IndySdkStorageService.ts
```diff
@@ -0,0 +1,324 @@
+import type { IndySdkWallet } from '../wallet/IndySdkWallet'
+import type {
+  BaseRecordConstructor,
+  AgentContext,
+  BaseRecord,
+  TagsBase,
+  Query,
+  StorageService,
+} from '@aries-framework/core'
+import type { default as Indy, WalletQuery, WalletRecord, WalletSearchOptions } from 'indy-sdk'
```
karimStekelenburg, line 10: Same comment as before. (`IndySdk` vs `Indy`)

karimStekelenburg (changes requested):
Left a few comments.

Apart from those I'm a bit confused on how we abbreviate things. E.g. for in some interfaces we write `revocationRegistryId` and in others we have `attrNames`. My personal preference is to write everything out completely, but I care mostly about consistency.

packages/core/src/modules/routing/__tests__/pickup.test.ts
```diff
@@ -5,7 +5,7 @@ import { Subject } from 'rxjs'
 
 import { SubjectInboundTransport } from '../../../../../../tests/transport/SubjectInboundTransport'
```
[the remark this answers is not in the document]
TimoGlastra: They don't...

I made a change to set the default key derivation to RAW to speed up wallet opening and closing. I should have probably done this in a separate PR. But funnily enough that broke some tests because they're quicker now (and thus certain events happened that shouldn't happen, and weren't happening before)

packages/indy-sdk/README.md
```diff
@@ -0,0 +1,31 @@
+<p align="center">
+  <br />
+  <img
+    alt="Hyperledger Aries logo"
+    src="https://raw.githubusercontent.com/hyperledger/aries-framework-javascript/aa31131825e3331dc93694bc58414d955dcb1129/images/aries-logo.png"
+    height="250px"
+  />
+</p>
+<h1 align="center"><b>Aries Framework JavaScript IndySDK Module</b></h1>
```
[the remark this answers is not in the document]
TimoGlastra: Made a PR: https://github.com/hyperledger/aries-framework-javascript/pull/1201

packages/indy-sdk/src/IndySdkModuleConfig.ts
TimoGlastra: I've tried to rename as much as possible to IndySDK, as Indy is a project, and IndySdk a specific implementation. Just to make it more clear it's the interface for the IndySdk

packages/indy-sdk/src/storage/IndySdkStorageService.ts
TimoGlastra: Will update, needs to be Sdk everywhere

TimoGlastra:
> Apart from those I'm a bit confused on how we abbreviate things. E.g. for in some interfaces we write revocationRegistryId and in others we have attrNames. My personal preference is to write everything out completely, but I care mostly about consistency.

The path I've gone with is that all anoncreds modules are used as defined in the AnonCreds spec (so `attrNames`). All interfaces specific to AFJ use full naming (so revocationRegistryId). 

But if we're e.g. working with the revocation list model, we'll still use the formatting as defined in that model, so we'll use `revocationList.revRegId` over `revocationList.revocationRegistryId`. Otherwise we'll have to do loads of transformations between AFJ format and AnonCreds formats, and I think it adds confusion to the end user that the models as defined in the spec aren't what is used by the AFJ API

TimoGlastra:
CI is failing until https://github.com/DefinitelyTyped/DefinitelyTyped/pull/63867 is merged

karimStekelenburg (approved):
Nice one!

PR #1201, opened 2023-01-10
chore: rename plugin to module

TimoGlastra:

### DIFF
--- a/packages/action-menu/README.md
+++ b/packages/action-menu/README.md
@@ -6,7 +6,7 @@
     height="250px"
   />
 </p>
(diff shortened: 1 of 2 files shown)


genaris (approved):
Both README were based on [Push Notifications](https://github.com/hyperledger/aries-framework-javascript-ext/tree/main/packages/push-notifications) from aries-framework-javascript-ext, so it would be appropriate to update the naming there as well. However I'm not sure how to call it there because that package includes two different modules (the one for APN and the one for FCM) 🤯

Issue #1123, opened 2022-11-28
Create AnonCreds service that can handle multiple AnonCredsRegistries


karimStekelenburg:
### Description
One of the things we're building into AFJ is the ability to talk to multiple ledgers. In the architecture we're building (as depicted in the image below), we separate the credential format-related logic (orange box) from the logic that deals with communication to the ledgers (red box). Connection management and ledger communication are handled by implementations of the `AnonCredsRegistry` interface. In order to support multiple ledgers, we need to create an `AnonCredsService` that manages these `AnonCredsRegistry`s.

<p align="center">
<img src="https://user-images.githubusercontent.com/18397186/204326005-5a786068-5374-45fd-bd33-cb03aa64c088.png" width=50% height=50%>
</p>

#### Knowing which ledger to use
With support for multiple ledgers, the `AnonCredsService` will need to figure out what `AnonCredsRegistry` instance it should use when a call is being made. This can be derived from the identifiers that are passed via the method parameters. The `AnonCredsService` needs a regex-like mechanism to match this identifier to the corresponding registry.

##### Identifiers
Indy networks can be identified in two ways: `did:indy` identifiers and legacy identifiers. More information on `did:indy` identifiers can be found [here](https://hyperledger.github.io/indy-did-method/).

The other network that we will add support for is cheqd. This network uses the `did:cheqd` identifier. More info [here](https://docs.cheqd.io/node/architecture/adr-list/adr-002-cheqd-did-method).

> 💡 The logic required here is similar to That that of the [DidRegistrarService](https://github.com/hyperledger/aries-framework-javascript/blob/main/packages/core/src/modules/dids/services/DidRegistrarService.ts) and [DidResolverService](https://github.com/hyperledger/aries-framework-javascript/blob/main/packages/core/src/modules/dids/services/DidResolverService.ts).

### Approaches
There are two possible approaches to implementing this.

#### Using `AnonCredsService` as a bridge
Here the `AnonCredsService` fetches the right `AnonCredsRegistry` and calls it.

```ts
class AnonCredsService {
  holder: AnonCredsHolder;
  issuer: AnonCredsIssuer;
  verifier: AnonCredsVerifier;

  registries: AnonCredsRegistry[];

  private getRegistryForIdentifier(identifier: string): AnonCredsRegistry { /** not important **/ }

  public async createCredentialDefinition(options: CreateCredentialOptions) {
    const registry = getRegistryForIdentifier(options.schemaId);
[...]

PR #1204, opened 2023-01-11
feat(anoncreds): add anoncreds registry service

TimoGlastra:
Fixes https://github.com/hyperledger/aries-framework-javascript/issues/1123, and adds the initial structure for the `AnonCredsModule`

### DIFF
--- a/packages/anoncreds/src/AnonCredsModuleConfig.ts
+++ b/packages/anoncreds/src/AnonCredsModuleConfig.ts
@@ -1,0 +1,28 @@
+import type { AnonCredsRegistry } from './services'
+
+/**
+ * @public
+ * AnonCredsModuleConfigOptions defines the interface for the options of the AnonCredsModuleConfig class.
+ */
+export interface AnonCredsModuleConfigOptions {
+  /**
+   * A list of AnonCreds registries to make available to the AnonCreds module.
+   */
+  registries: [AnonCredsRegistry, ...AnonCredsRegistry[]]
+}
+
+/**
+ * @public
+ */
+export class AnonCredsModuleConfig {
+  private options: AnonCredsModuleConfigOptions
+
+  public constructor(options: AnonCredsModuleConfigOptions) {
+    this.options = options
+  }
+
+  /** See {@link AnonCredsModuleConfigOptions.registries} */
+  public get registries() {
+    return this.options.registries
+  }
+}
--- a/packages/anoncreds/src/services/registry/AnonCredsRegistry.ts
+++ b/packages/anoncreds/src/services/registry/AnonCredsRegistry.ts
@@ -12,9 +12,7 @@ import type { AgentContext } from '@aries-framework/core'
  * @public
  */
 export interface AnonCredsRegistry {
-  // TODO: should this be a single regex? So just supportedMethodRegex. If you support multiple / has complex queries you can just combine them.
-  // This would prevent the need for a double loop (loop through all registries, then loop through all supportedMethods)
-  supportedMethods: RegExp[]
+  supportedIdentifier: RegExp
 
   getSchema(agentContext: AgentContext, schemaId: string): Promise<GetSchemaReturn>
   registerSchema(agentContext: AgentContext, options: RegisterSchemaOptions): Promise<RegisterSchemaReturn>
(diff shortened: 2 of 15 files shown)


TimoGlastra:
Thinking whether we should rename `supportedMethods` to `supportedIdentifiers` in the AnonCredsRegistry interface, as we're not really filtering based on method, but rather on the structure of the identifier

packages/anoncreds/src/AnonCredsModuleConfig.ts
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * @public
+ * AnonCredsModuleConfigOptions defines the interface for the options of the AnonCredsModuleConfig class.
+ */
+export interface AnonCredsModuleConfigOptions {
+  /**
+   * A list of AnonCreds registries to make available to the AnonCreds module.
+   */
+  registries: [AnonCredsRegistry, ...AnonCredsRegistry[]]
```
blu3beri, line 11:
```suggestion
  registries: Array<AnonCredsRegistry>
```

Or am I missing a sick hidden TypeScript feature?

packages/anoncreds/src/services/registry/AnonCredsRegistry.ts
```diff
@@ -8,8 +8,14 @@ import type { GetRevocationRegistryDefinitionReturn } from './RevocationRegistry
...
 import type { AgentContext } from '@aries-framework/core'
 
-// This service can be registered multiple times in a single AFJ instance.
+/**
+ * @public
+ */
 export interface AnonCredsRegistry {
+  // TODO: should this be a single regex? So just supportedMethodRegex. If you support multiple / has complex queries you can just combine them.
+  // This would prevent the need for a double loop (loop through all registries, then loop through all supportedMethods)
+  supportedMethods: RegExp[]
```
blu3beri, line 17: I think here we should assume the most common use case, which I think is one identifier. Which regex it is rather easy to match on `A` or `B` (`(A|B)`).

blu3beri: I think I'd prefer the `supportedIdentifier`, but I might prefer it if it indicates a regex. so `identifierPattern` or something? And we can provide a util function with would be something like this:

```typescript
const getSimpleMethodPattern(method:string): RegExp {
  return new RegExp(`^did:${method}:.*$`) // not a perfect pattern but you get the gist :).
}

File: packages/anoncreds/src/services/registry/AnonCredsRegistryService.ts [hunk 3]
```diff
@@ -0,0 +1,28 @@
...
+ * @internal
+ * The AnonCreds registry service manages multiple {@link AnonCredsRegistry} instances
+ * and returns the correct registry based on a given identifier
+ */
+@injectable()
+export class AnonCredsRegistryService {
+  public async getRegistryForIdentifier(agentContext: AgentContext, identifier: string): Promise<AnonCredsRegistry> {
+    const registries = agentContext.dependencyManager.resolve(AnonCredsModuleConfig).registries
+
+    // TODO: should we check if multiple are registered?
```
blu3beri, line 19: Ah yeah that's a good question. I am okay with picking the first one, but we have to mark it explicitly in the code documentation for the public function which uses this `getRegistryForIdentifier` under-the-hood.

packages/indy-sdk/src/anoncreds/services/IndySdkAnonCredsRegistry.ts
```diff
@@ -33,32 +36,39 @@ import {
...
 export class IndySdkAnonCredsRegistry implements AnonCredsRegistry {
-  private indySdk: IndySdk
-  private indySdkPoolService: IndySdkPoolService
-
-  public constructor(@inject(IndySdkSymbol) indySdk: IndySdk, indySdkPoolService: IndySdkPoolService) {
-    this.indySdk = indySdk
-    this.indySdkPoolService = indySdkPoolService
-  }
+  /**
+   * This class only supports resoling and registering objects with legacy indy identifiers.
```
blu3beri, line 40:
```suggestion
   * This class only supports resolving and registering objects with legacy indy identifiers.
```

blu3beri (changes requested):
A couple of notes, nothing major.

packages/anoncreds/src/AnonCredsModuleConfig.ts
TimoGlastra: It means the array must have at least one entry in it. As using the module without registries isn't really useful

packages/anoncreds/src/services/registry/AnonCredsRegistry.ts
TimoGlastra: I like it!

packages/anoncreds/src/AnonCredsModuleConfig.ts
blu3beri: Ah, cool!

blu3beri (approved):

PR #1206, opened 2023-01-11
docs: update readme packages

karimStekelenburg:
I noticed only the `core`, `node` and `react-native` packages were listed in the `README.md`. I'm not sure if this was done on purpose or just forgotten. I've added the `bbs-signatures`, `action-menu`, `question-answer`, and `tenants` modules and updated the `tenants` roadmap state from 🚧  to  ✅. Please comment/reject if this is incorrect.

Also, I've just appended the missing modules to the table for now. I'm not sure about the ordering of these modules/table rows. One approach is to order them alphabetically, but my personal preference is to order them by relevance.

**Example:**
1. `core` -- as you'll always need it
2. `node` & `react-native` (in random order) -- as you'll always need one of the two
3. Other modules -- as they really depend on your use case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ Some features are not yet supported, but are on our roadmap. Check [the roadmap]
 - ✅ Smart Auto Acceptance of Connections, Credentials and Proofs
 - 🚧 Receiving and Verifying revocable Indy Credentials
 - 🚧 W3C Linked Data VCs, BBS+ Signatures
-- 🚧 Multi Tenancy
+- ✅ Multi Tenancy
 - ❌ Browser
 
 ### Packages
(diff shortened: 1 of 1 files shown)


blu3beri (approved):
I think it was just forgotten. Good catch :). 

Also, I think the order you did is correct and just keep adding at the bottom.

README.md
```diff
@@ -101,6 +101,38 @@ Some features are not yet supported, but are on our roadmap. Check [the roadmap]
...
+  <tr>
+    <td>@aries-framework/bbs-signatures</td>
+    <td>
+      <a href="https://npmjs.com/package/@aries-framework/bbs-signatures">
+        <img alt="@aries-framework/bbs-signatures version" src="https://img.shields.io/npm/v/@aries-framework/bbs-signatures"/>
+      </a>
+    </td>
+  </tr>
+  <tr>
+    <td>@aries-framework/action-menu</td>
```
genaris, line 121:
```suggestion
    <td>@aries-framework/question-answer</td>
```

README.md
```diff
@@ -101,6 +101,38 @@ Some features are not yet supported, but are on our roadmap. Check [the roadmap]
...
+    <td>
+      <a href="https://npmjs.com/package/@aries-framework/question-answer">
+        <img alt="@aries-framework/question-answer version" src="https://img.shields.io/npm/v/@aries-framework/question-answer"/>
+      </a>
+    </td>
+  </tr>
+  <tr>
+    <td>@aries-framework/action-menu</td>
```
genaris, line 129:
```suggestion
    <td>@aries-framework/tenants</td>
```

README.md
```diff
@@ -101,6 +101,38 @@ Some features are not yet supported, but are on our roadmap. Check [the roadmap]
...
+      <a href="https://npmjs.com/package/@aries-framework/action-menu">
+        <img alt="@aries-framework/action-menu version" src="https://img.shields.io/npm/v/@aries-framework/action-menu"/>
+      </a>
+    </td>
+  </tr>
+  <tr>
+    <td>@aries-framework/bbs-signatures</td>
+    <td>
+      <a href="https://npmjs.com/package/@aries-framework/bbs-signatures">
+        <img alt="@aries-framework/bbs-signatures version" src="https://img.shields.io/npm/v/@aries-framework/bbs-signatures"/>
```
genaris, line 116: As it's private for the moment, in the rendered version I see 'invalid' on the badge. Not related to this PR but I'm wondering if it would be good to publish it so users can start testing AFJ with this kind of credentials (I think we already marked JSON-LD and that stuff as 'experimental' from 0.3.0, and this README file also remarks this WIP status)

genaris (changes requested):
Just a few errors I noticed when looking at the rendered version.

Would be good also to review all the RFCs we support (e.g. Present Proof V2 is not yet there) and add them to the list but I think it could be part of a separate PR.

README.md
karimStekelenburg: Thanks @genaris, good points. I removed the `bbs-signatures`. I agree that only public packages should be listed.

karimStekelenburg: Oh and I agree with publishing the `bbs-signatures` package. It's already usable without PPV2 being fully merged. The sooner people start testing the better IMO.

genaris (approved):

README.md
TimoGlastra: Yeah sure. We should have a good look at it, test it out again in react native, make sure the flaws are documented (it doesn't work in Node 17 + 18). So if someone is willing to perform these steps, we can set it to public. But otherwise it will automatically be published during the next patch release

karimStekelenburg: I'm wondering if it's worth it to document these manual react-native tests somewhere. Since we can't test it programmatically, it may be worth to track it somehow...

PR #1209, opened 2023-01-12
chore: remove husky pre-push hook

TimoGlastra:
Removes the husky pre-push hook, as the command takes way too long to run, and we have CI to make sure we aren't including broken code.

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,1 +1,0 @@
-yarn validate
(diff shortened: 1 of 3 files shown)


blu3beri:
@karimStekelenburg I assume you are quite happy right now ;). No more: `--no-verify`.

blu3beri (approved):

Issue #1198, opened 2023-01-10
Remove masterSecret from the wallet in AFJ


TimoGlastra:
As AnonCreds is fully becoming an separate module, it doesn't make as much sense to keep the masterSecret logic in the wallet class. Therefore, we should remove the masterSecret logic from the wallet, and instead, add the need methods to the AnonCreds  services and API

Blocker: the old indy services still depend on it. We need to remove those to be able to do this task

Issue #706, opened 2022-04-14
Update basePath to not use temp directory


TimoGlastra:
Update `basePath` in FileSystem to not use temp directory. Add new property `tempPath` for storing temp files

## Checklist
- [ ] Update `basePath` to non-temp file path
- [ ] Add `tempPath` property
- [ ] Add method to remove files / dirs to the file systems (ReactNativeFileSystem, NodeFileSystem)

PR #1211, opened 2023-01-12
feat: add initial askar package

genaris:
Initial implementation of Aries Askar module for AFJ, including interfaces for Storage and Wallet. 

The module registers a Storage and a Wallet instance automatically when it registers itself on the dependency manager. If there is already a module that has registered a Storage or a Wallet, it throws an error.

Features already implemented:
- Wallet creation, deletion and key rotation using both SQLite and PosgreSQL backends
- Record Get/Update/Save including WQL queries
- DIDComm V1 packing (basic Connection + BasicMessage exchange flow tested briefly in askar-aries-e2e-wallet-subject.test.ts)
- Key creation, sign and verify, using custom SigningProvider if key type is not supported by Askar (however, once https://github.com/hyperledger/aries-askar/pull/85 is merged I think all current key types are covered natively).

Features missing to match 1:1 with Indy SDK:

- Wallet Import/Export: not currently provided by Askar, so it should be developed (either in Askar as suggested by https://github.com/hyperledger/aries-askar/issues/82, or in AFJ)

There are several FIXMEs and TODOs in the code. Most of them are related to either fixes or new features to be added to aries-askar JS Wrapper, so this module will need to be adapted accordingly. Unfortunately, a problem related to JavaScript GC/memory issues (https://github.com/hyperledger/aries-askar/issues/76) prevents E2E tests using Askar wallets to properly run in CI, so they are currently skipped.

Some more TODOs for a further PR:

- [ ] Update README con information about module contents and usage
- [ ] Improve modeling of JweEnvelope and support/test different scenarios
- [ ] Add more advanced tests


genaris:
For some reason, ESLint is failing in packages/askar/src/wallet/AskarWallet.ts but I don't find a way to fix it:

> Error:    37:1   error    There should be no empty line within import group  import/order

Any advice will be welcome!

packages/askar/src/wallet/AskarWallet.ts
```diff
@@ -0,0 +1,745 @@
...
+        recreate: false,
+        uri: askarWalletConfig.uri,
+        profile: askarWalletConfig.profile,
+        keyMethod: askarWalletConfig.keyMethod,
+        passKey: askarWalletConfig.passKey,
+      })
+      this.walletConfig = walletConfig
+      this._session = await this._store.openSession()
+
+      // TODO: Master Secret creation (now part of IndyCredx/AnonCreds)
```
TimoGlastra, line 156: Yeah I want to extract this from the wallet as it is more an AnonCreds thing. See #1198

packages/askar/src/storage/AskarStorageService.ts
```diff
@@ -0,0 +1,291 @@
...
+    const tags = record.tags ? this.transformToRecordTagValues(record.tags) : {}
+    instance.replaceTags(tags)
+
+    return instance
+  }
+
+  /** @inheritDoc */
+  public async save(agentContext: AgentContext, record: T) {
+    assertAskarWallet(agentContext.wallet)
+    const session = (agentContext.wallet as AskarWallet).session
```
TimoGlastra, line 136: I don't this you need to cast here. The `assertAskarWallet` returns `asserts wallet is AskarWallet` so after calling that method typescript should know the wallet is of type `AskarWallet`

packages/askar/src/storage/AskarStorageService.ts
```diff
@@ -0,0 +1,291 @@
...
+        throw new RecordNotFoundError(`record with id ${id} not found.`, {
+          recordType: recordClass.type,
+        })
+      }
+      return this.recordToInstance(record, recordClass)
+    } catch (error) {
+      if (
+        isAskarError(error) &&
+        (error.code === askarErrors.NotFound ||
+          // FIXME: this is current output from askar wrapper but does not describe specifically a not found scenario
```
TimoGlastra, line 231: It seems Askar also just returns not found if the record wasn't found: 

```
        try:
            item = await self._session.handle.fetch(
                record_type, record_id, for_update=for_update
            )
        except AskarError as err:
            raise StorageError("Error when fetching storage record") from err
        if not item:
            raise StorageNotFoundError(f"Record not found: {record_type}/{record_id}")
```

Will it sometimes throw an error instead of not returning a record?

packages/askar/src/storage/AskarStorageService.ts
```diff
@@ -0,0 +1,291 @@
...
+    try {
+      const records = await scan.fetchAll()
+      for (const record of records) {
+        instances.push(this.recordToInstance(record, recordClass))
+      }
+      return instances
+    } catch (error) {
+      if (
+        isAskarError(error) && // FIXME: this is current output from askar wrapper but does not describe specifically a 0 length scenario
+        error.message === 'Received null pointer. The native library could not find the value.'
```
TimoGlastra, line 284: Isn't this something we can fix in the wrapper? So the wrapper should jus allow a null pointer as return value? Maybe @blu3beri knows

packages/askar/src/storage/__tests__/AskarStorageService.test.ts
```diff
@@ -0,0 +1,311 @@
...
+            $and: undefined,
+            $or: undefined,
+            $not: undefined,
+          },
+        ],
+        $not: { myTag: 'notfoobar', $and: undefined, $or: undefined, $not: undefined },
+      }
+
+      expect(
+        //@ts-ignore
```
TimoGlastra, line 291: Why the ts-ignore? Should we add an eslint disable here?

packages/askar/src/utils/askarWalletConfig.ts
```diff
@@ -0,0 +1,42 @@
...
+  const correspondanceTable = {
+    [KeyDerivationMethod.Raw]: StoreKeyMethod.Raw,
+    [KeyDerivationMethod.Argon2IInt]: `${StoreKeyMethod.Kdf}:argon2i:int`,
+    [KeyDerivationMethod.Argon2IMod]: `${StoreKeyMethod.Kdf}:argon2i:mod`,
+  }
+
+  return correspondanceTable[keyDerivationMethod] as StoreKeyMethod
+}
+
+export const uriFromWalletConfig = (walletConfig: WalletConfig, basePath: string): { uri: string; path?: string } => {
```
TimoGlastra, line 20: We may need to tweak the wallet config a bit if that leads to better results. I've even been thinking about removing the wallet config from the agent config and just making it specific to the wallet interface so there's no need for a standardized wallet config interface. However that would cause some issues for e.g. the tenant module (that needs to create wallet configs on demand).  Thoughts?

packages/askar/src/wallet/AskarWallet.ts
```diff
@@ -0,0 +1,745 @@
...
+    return {
+      plaintextMessage: JsonEncoder.fromBuffer(message),
+      senderKey,
+      recipientKey,
+    }
+  }
+
+  public async generateNonce(): Promise<string> {
+    try {
+      return new TextDecoder().decode(CryptoBox.randomNonce())
```
TimoGlastra, line 669: TextDecoder/EncoderCoder is not available in react native.

packages/askar/src/wallet/AskarWallet.ts
```diff
@@ -0,0 +1,745 @@
...
+    }
+  }
+
+  private async getAskarWalletConfig(walletConfig: WalletConfig) {
+    const { uri, path } = uriFromWalletConfig(walletConfig, this.fileSystem.basePath)
+
+    // Make sure path exists before creating the wallet
+    if (path) {
+      await this.fileSystem.createDirectory(path)
+    }
```
TimoGlastra, line 692: Will it throw an error if the path doesn't exist?

We need to take a good look at the fileSystem basePath. I think for some reason the fileSystem by default is a temp directory. See https://github.com/hyperledger/aries-framework-javascript/issues/706

packages/askar/src/wallet/AskarWallet.ts
```diff
@@ -0,0 +1,745 @@
...
+        return Key.fromPublicKey(key.publicBytes, keyType)
+      } else {
+        // Check if there is a signing key provider for the specified key type.
+        if (this.signingKeyProviderRegistry.hasProviderForKeyType(keyType)) {
+          const signingKeyProvider = this.signingKeyProviderRegistry.getProviderForKeyType(keyType)
+
+          const keyPair = await signingKeyProvider.createKeyPair({ seed })
+          await this.storeKeyPair(keyPair)
+          return Key.fromPublicKeyBase58(keyPair.publicKeyBase58, keyType)
+        }
```
TimoGlastra, line 372: Can askar store keys that it doesn't support? So we still create the key using the SigningKeyProviderRegistry, but we store it using askar?

packages/askar/src/wallet/AskarWallet.ts
```diff
@@ -0,0 +1,745 @@
...
+   * @throws {WalletError} When the key could not be created
+   */
+  public async createKey({ seed, keyType }: WalletCreateKeyOptions): Promise<Key> {
+    try {
+      if (keyTypeSupportedByAskar(keyType)) {
+        const algorithm = keyAlgFromString(keyType)
+
+        // Create key from seed
+        const key = seed
+          ? AskarKey.fromSeed({ seed: new TextEncoder().encode(seed), algorithm })
```
TimoGlastra, line 358: text encoder is not available in RN

packages/askar/src/wallet/AskarWallet.ts
```diff
@@ -0,0 +1,745 @@
...
+   * @param data Buffer The data that needs to be signed
+   * @param key Key The key that is used to sign the data
+   *
+   * @returns A signature for the data
+   */
+  public async sign({ data, key }: WalletSignOptions): Promise<Buffer> {
+    try {
+      if (!TypedArrayEncoder.isTypedArray(data)) {
+        throw new WalletError(`Currently not supporting signing of multiple messages`)
+      }
```
TimoGlastra, line 395: The BBS Signing provider does support multilpe message I think? Maybe we can place this withiin the `if (keyTypeSupportedByAskar(key.keyType)) {`?

packages/askar/src/wallet/AskarWallet.ts
```diff
@@ -0,0 +1,745 @@
...
+   *
+   * @param payload message to send
+   * @param recipientKeys array containing recipient keys in base58
+   * @param senderVerkey sender key in base58
+   * @returns JWE Envelope to send
+   */
+  public async pack(
+    payload: Record<string, unknown>,
+    recipientKeys: string[],
+    senderVerkey?: string // in base58
```
TimoGlastra, line 491: Not for now, we should update this to be key instances. The sign/verify already uses keys, and this is just leftover...

TimoGlastra: Also (not for now), but we should update the interface to support DIDComm v2 as well.

Ideally we make the interface more lower level and not `pack`. But rather to encrypt something, and we create a DIDComm packing service that handles the other parts. 

Or we keep it pack but we make the interface more generic and for indy-sdk we only support `xchacha20poly1305_ietf`.

TimoGlastra: I kinda liked how ACA-Py has extracted this into a separate didcomm/v1.py and didcomm/v2.py files to separet the didcomm implementation. Do you think we could do the same?

packages/askar/src/wallet/AskarWallet.ts
```diff
@@ -0,0 +1,745 @@
...
+      }
+      throw new WalletError('Error retrieving KeyPair record', { cause: error })
+    }
+  }
+
+  private async storeKeyPair(keyPair: KeyPair): Promise<void> {
+    try {
+      await this.session.insert({
+        category: 'KeyPairRecord',
+        name: `key-${keyPair.publicKeyBase58}`,
```
TimoGlastra, line 732: Is this the same category / id as the indy-sdk? That'll be important for the migration script

packages/askar/src/wallet/JweEnvelope.ts
```diff
@@ -0,0 +1,65 @@
...
+  aad?: string
+  header?: string[]
+  encryptedKey?: string
+}
+
+export class JweEnvelope {
+  public protected!: string
+  public unprotected?: string
+
+  @Expose()
```
TimoGlastra, line 36: Fields are exposed by default and only needed if we want a custom property name

packages/askar/src/wallet/__tests__/AskarWallet.test.ts
```diff
@@ -0,0 +1,254 @@
...
+      }
+
+      public async sign(options: SignOptions): Promise<Buffer> {
+        return new Buffer('signed')
+      }
+
+      public async verify(options: VerifyOptions): Promise<boolean> {
+        return true
+      }
+    }
```
TimoGlastra, line 134: Can we extarct this out of hte beforeEach?

packages/core/src/utils/base64.ts
```diff
@@ -1,3 +1,9 @@
+import { Buffer } from './buffer'
+
 export function base64ToBase64URL(base64: string) {
   return base64.replace(/\+/g, '-').replace(/\//g, '_').replace(/=/g, '')
 }
+
+export function uint8ArrayToBase64URL(array: Uint8Array) {
```
TimoGlastra, line 7: We have `TypedArrayEncoded.toBase64Url`. Can we use that instead? That uses `base64ToBase64URL` internally.

packages/react-native/src/ReactNativeFileSystem.ts
```diff
@@ -21,6 +21,10 @@ export class ReactNativeFileSystem implements FileSystem {
     return RNFS.exists(path)
   }
 
+  public async createDirectory(path: string): Promise<void> {
```
TimoGlastra, line 24: Annoying we need to create the directory ourselves for Askar. I would like to use the file system as little as possible, oh wel...

tests/e2e-wallet-subject.test.ts
```diff
@@ -0,0 +1,154 @@
+import type { SubjectMessage } from './transport/SubjectInboundTransport'
+
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { NodeJSAriesAskar } from 'aries-askar-test-nodejs'
+import { Subject } from 'rxjs'
```
TimoGlastra, line 5: can we rename this file to include askar and indy-sdk in the filename?

tests/e2e-wallet-subject.test.ts
```diff
@@ -0,0 +1,154 @@
+import type { SubjectMessage } from './transport/SubjectInboundTransport'
+
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { NodeJSAriesAskar } from 'aries-askar-test-nodejs'
```
TimoGlastra, line 4: We should add this package to the root if we use it here

TimoGlastra (changes requested):
This is great @genaris. Overall looks good and very clean. Like the extensive tests.

We should still look at how we're going to register the storage and wallet. Is the AskarModule going to register them, or do we need to provide them to the storage and wallet modules? Something to solve in a later PR ofc.

TimoGlastra:
Could you create tickets for the outstanding tasks so we have insights in what still needs to happen (and we can prioritize what is important for now)

packages/askar/src/storage/AskarStorageService.ts
blu3beri: Yeah if that is the intended behaviour we can fix that in the wrapper.

genaris: I've added https://github.com/hyperledger/aries-askar/issues/77 to follow this issue. I'm not absolutely sure that is a specific problem for JS, as a preliminar quick-debugging I've done showed me that the wrapper is getting an Error Code: 0 in those cases so it tries to create an EntryList from a null pointer (and that's why it shows that custom error).

packages/askar/src/storage/AskarStorageService.ts
genaris: Not 100% sure but I'd say that it always sends the error.code Custom (100). Let's analyze on Askar side to see where the problem is.

packages/askar/src/storage/AskarStorageService.ts
genaris: Good catch! I've added the assert afterwards and didn't consider that fact.

packages/askar/src/storage/__tests__/AskarStorageService.test.ts
genaris: This is because I'm calling a private method. Is there an elegant way to call a private method from jest? I know the correct way of doing things is to test it indirectly but in this case just wanted to check the internals.

packages/askar/src/utils/askarWalletConfig.ts
genaris: I agree that it may become difficult to define a single base interface for wallets. If it's getting complicated to get a common interface for two projects originated from the same codebase (sort of), I can imagine that will be very hard to adapt any kind of storage backend to AFJ. 

So I think it will be nice if we can leave the format to the nature of the wallet itself and make it free to choose how to get configured. We'll check how to allow Tenants module to support it (maybe a callback for dynamic config generation?).

packages/askar/src/wallet/AskarWallet.ts
genaris: Yes, unfortunately that's what I've found for SQLite backend. In the code I left the possibility of choosing our very own path or just leave the framework to decide (based on that `basePath` that I agree we should not put on a temporary directory). This could be greatly improved as soon as we define a better wallet config interface for Askar.

packages/askar/src/wallet/AskarWallet.ts
genaris: > Ideally we make the interface more lower level and not `pack`. But rather to encrypt something, and we create a DIDComm packing service that handles the other parts.
> 

Yeah, in my opinion this is the way to go: the Wallet provides means to encrypt and sign/verify data while the DIDCommPackingService just uses it (and the DID Resolver) to pack and unpack the message. We'll need of course to define a good interface allowing all the ciphering we are doing for packing.

We can provide in AFJ Core "DIDCommPackingService" for all DIDComm versions and, in the end, supported versions for an Agent will be given by the Wallet it uses.

packages/askar/src/wallet/AskarWallet.ts
genaris: Yes. And as Askar is supporting natively more algorithms than Indy, the migration script must make sure to write all those keys natively supported into the specific Key Storage, so it will not be a simply "copy-paste".

packages/askar/src/wallet/AskarWallet.ts
genaris: I've seen that TextEncoder is being used in [a few parts ](https://github.com/hyperledger/aries-askar/blob/697a77fa71312ee2bedc414e31cb3c9ba1124ecd/wrappers/javascript/aries-askar-shared/src/store/Session.ts#L98)of aries-askar-shared. As this code is used in both NodeJS and RN I think this should be updated, right?

packages/askar/src/wallet/AskarWallet.ts
genaris: Well, that's what we are doing now, so I guess you mean to store using Askar's specific API to do so (e.g. `session.insertKey `instead of `session.insert`). I don't think it will work because to create a `Key` object it will check `KeyAlgs` enum and it will not be there, so most likely it will throw an `Unsupported` error.

packages/askar/src/storage/__tests__/AskarStorageService.test.ts
TimoGlastra: Yeah sometimes that's fine. It sometimes add typing with a comment:

```
// we want to test the private method
type A = Original & { privateMethod: something }
const a = b as unknown as A
```

An alternative is to extract the logic into an util and testing the util.

Your approach is fine as well, but a comment would be good!

packages/askar/src/wallet/AskarWallet.ts
TimoGlastra: Hmm yes...

TimoGlastra: We've had this issue before, and it was doable to do it in other ways, however it'll probably be hard to do without a buffer polyfill in RN (or a text encoder polyfill)

Maybe the encoding/decoding needs to happen on the platform specific implementation?

Or we add a polyfill to both (afj already uses buffer polyfill). Thoughts?

packages/askar/src/storage/__tests__/AskarStorageService.test.ts
genaris: Thanks for the info! I moved them to an util to make it cleaner.

genaris:
> We should still look at how we're going to register the storage and wallet. Is the AskarModule going to register them, or do we need to provide them to the storage and wallet modules? Something to solve in a later PR ofc.

Absolutely. I created https://github.com/hyperledger/aries-framework-javascript/issues/1235 to discuss how can we implement it.

genaris:
> Could you create tickets for the outstanding tasks so we have insights in what still needs to happen (and we can prioritize what is important for now)

Sure. I've created a few in AFJ and others in aries-askar. I think for this particular PR to be merged in a meaningful way we'll need two main things:
- Solve https://github.com/hyperledger/aries-askar/issues/76, as currently it is quite hard to get test suites to pass in CI: this GC overhead affects not only Askar tests but also others like the ones for BBS Signatures
- Have an initial release of @hyperledger/aries-askar-shared / @hyperledger/aries-askar-nodejs. Not absolutely mandatory but, as we are very close to release them, it would be nice to merge into main the module linked to the actual packages and not the test ones

Once we get this, I think we can adapt the module to the fixes in aries-askar and the Wallet API changes that will be introduced when Indy is removed from the Core.

packages/askar/src/AskarModule.ts
```diff
@@ -0,0 +1,33 @@
...
+    try {
+      // eslint-disable-next-line import/no-extraneous-dependencies
+      require('@hyperledger/aries-askar-nodejs')
+    } catch (error) {
+      try {
+        require('@hyperledger/aries-askar-react-native')
+      } catch (error) {
+        throw new Error('Could not load aries-askar bindings')
+      }
+    }
```
TimoGlastra, line 19: 👍 nice. We could probably improve this at some point by detecting the env and loading packages based on that. I think it could succeed in requireing the nodejs package even though we're in a RN env?

packages/askar/src/wallet/__tests__/AskarWallet.test.ts
```diff
@@ -0,0 +1,250 @@
...
+          privateKeyBase58: 'privateKeyBase58',
+          keyType: KeyType.Bls12381g1g2,
+        }
+      }
+
+      public async sign(options: SignOptions): Promise<Buffer> {
+        return new Buffer('signed')
+      }
+
+      public async verify(options: VerifyOptions): Promise<boolean> {
```
TimoGlastra, line 128: Can you add some eslint ignores here?

TimoGlastra (approved):

packages/askar/src/AskarModule.ts
genaris: I'm not convinced about this way of importing it but I wanted to left it consistent with the way we are using other shared components. And it's nice that it attempts to make life easier to developers using the framework.. unless they import nodejs package when on react native environment :open_mouth: . Would be nice to investigate if there is a reliable way to get the JS environment nowadays.

TimoGlastra: Agreed. Also not 100% convinced yet. I think it's just another thing the user needs to provide to the agent config, while we could just infer it. Maybe we only try to import it if it's not provided yet? This way it 'works' out of the box. but you can also manually configure it

Issue #1213, opened 2023-01-16
Update packages


TimoGlastra:
There's quite some updates to dependencies of AFJ packages. We can update dev dependencies without issue, but doing breaking updates for non-dev dependencies means it'll be a breaking change for AFJ. 

Before we're intending to make a 0.4.0 we should update all dependencies to the latest version.

We could say that we will already do this, and just make a 0.4.x quite soon to keep the releases flowing. Thoughts?

PR #1214, opened 2023-01-16
fix: fix typing issues with typescript 4.9

### LINKED ISSUE #1205: CredentialsApi types does not satisfy the constraints
Hey

I tried to create module of agent initialization class for further usage it in my Node server, but when I tries compile from **.ts** to **.js** with `tsc --esModuleInterop true <file path>` I got following error:
```
node_modules/@aries-framework/core/build/modules/credentials/CredentialsApiOptions.d.ts:94:49 - error TS2344: Type 'CFsFromCPs<CPs>' does not satisfy the constraint 'CredentialFormat[]'.
  Type '[] | _CFsFromCPs<CPs>' is not assignable to type 'CredentialFormat[]'.
    Type '_CFsFromCPs<CPs>' is not assignable to type 'CredentialFormat[]'.
      Type 'FlatArray<{ [CP in keyof CPs]: ExtractCredentialFormats<ExtractCredentialFormatServices<CPs[CP]>>; }>' is not assignable to type 'CredentialFormat'.
        Type 'unknown' is not assignable to type 'CredentialFormat'.

94     credentialFormats?: CredentialFormatPayload<CFsFromCPs<CPs>, 'acceptRequest'>;
```

It is interesting because using `ts-node <file path> `everything works ok and as I understand` ts-node` uses `tsc` underneath

Tried to compile the next code: 
```
import type { InitConfig } from '@aries-framework/core'

import { Agent, AutoAcceptCredential, AutoAcceptProof, HttpOutboundTransport } from '@aries-framework/core'
import { agentDependencies, HttpInboundTransport } from '@aries-framework/node'

const bcovrin = //hardcoded genesis to Bcovrin Test Network

export class BaseAgent {
  public port: number
  public name: string
  public config: InitConfig
  public agent: Agent

  public constructor(port: number, name: string) {
    this.name = name
    this.port = port

    const config: InitConfig = {
      label: name,
      walletConfig: {
        id: name,
        key: name,
      },
      publicDidSeed: 'test0000000000000000000000000000',
      indyLedgers: [
        {
          genesisTransactions: bcovrin,
          id: 'testt' + name,
          indyNamespace: 'testt' + name,
          isProduction: false,
        },
      ],
      endpoints: [`http://localhost:${this.port}`],
      autoAcceptConnections: true,
      autoAcceptCredentials: AutoAcceptCredential.ContentApproved,
      autoAcceptProofs: AutoAcceptProof.ContentApproved,
    }

    this.config = config

    this.agent = new Agent({ config, dependencies: agentDependencies })
    this.agent.registerInboundTransport(new HttpInboundTransport({ port }))
    this.agent.registerOutboundTransport(new HttpOutboundTransport())
  }

  public async initializeAgent() {
    await this.agent.initialize()
  }
}
```

TimoGlastra:
Updates typescript to 4.9.4, and also updates some related dev dependencies. 

Haven't updated any non-dev dependencies (that should be done separately as mentioned in #1213).

Fixes #1205 

There were some issues specific to the latest TypeScript release which broke the types of AFJ. Mainly I've made the followign changes:
- update ts to 4.9.4
- Add @types/express to dependencies in @aries-framework/node. TS will complain that it can't find the types otherwise.
- Due to the update of eslint and typescript-eslint the order of imports have changed. Not sure why, but there's no functional change to the code
- Simplifies and changed the broken types in the credentials api.


TimoGlastra:
> but anyway it is handling 'type' group separately, so it does not take into account if we chose parents to be before siblings or vice versa

Yeah this has been the case since we've started using type imports. Not sure why, but it's okay I think. I don't really mind the order, as long as it's consistent so we don't have merge conflicts based on it.

TimoGlastra:
> but I'm curious about the reasons of separating type imports from the other ones.

I kinda like it because it shows it's used as a type, and thus will be removed at compilation.

morrieinmaas (commented):
LGTM! massive PR @TimoGlastra. In general, reading through this I was wondering whether it'd make sense to use interfaces instead of abstract classes for base instances. Not 100% sure, but generally might be nice if the base class(es) never get(s) instantiated. Not part of this PR anyway but just as an idea. With interfaces we'd still get the typing right but without the impl (which saves some space by getting compiled away, given we don't need an impl of the abstract base itself, but just the implementations and extensions thereof).

PR #1222, opened 2023-01-18
fix(transport)!: added docs moved connection to connectionId

blu3beri:
Was working on some documentation for myself for the transport and figured it would be to add some code comments about the transport session.

I also swapped the `connection` to `connectionId` as I we do not need a full reference to the connection, just the id which is used for `findSessionByConnectionId`.

The PR contains a breaking change, so I am not sure if I should use `main` as the base branch.

BREAKING CHANGE:
The `connection` on the `TransportSession` is now a `connectionId`.


packages/core/src/agent/TransportService.ts
```diff
@@ -36,12 +35,34 @@ interface TransportSessionTable {
...
+  // The enveloping keys that can be used during the transport. This is used so the framework
+  // does not have to look up the associated keys for sending a message.
   keys?: EnvelopeKeys
+
+  // A received message that will be used to check whether it has any return routing.
   inboundMessage?: AgentMessage
-  connection?: ConnectionRecord
+
+  // A stored connection id used to find this session via the `TransportService` for a specific connection
+  connectionId?: string
```
TimoGlastra, line 61: We should probably update the code that assigns the connection to the session right?

TimoGlastra (approved):
There's a foobarbaz file?

I'm fine with merging this, there's more breaking changes coming up anyway. Maybe hold on for a bit longer in case we want to merge some fixes?

packages/core/src/agent/TransportService.ts
blu3beri: Done.

blu3beri:
> There's a foobarbaz file?

Essential to the framework!

> I'm fine with merging this, there's more breaking changes coming up anyway. Maybe hold on for a bit longer in case we want to merge some fixes?

Yeah we can keep this PR open until everything else is ready. No problem.

karimStekelenburg:
What is that `foobarbaz` thing? It's not very obvious LOL

PR #1229, opened 2023-01-20
feat(cache): add caching interface

### LINKED ISSUE #1227: IndyPoolCache is broken when using multi-tenancy
When using multi-tenancy and interacting with the ledger the cache record will be stored in the PersistedLruCache class, however the cache itself is stored differently per tenant. 

There's two possible solutions to this, which may different dependant on the use case:
- Store the cache in the base wallet. The indy pools are shared between all tenants, so why not share the cache for ledger ready requests
- Fix the cache to work with multiple tenants

We should add an interface for the cache anyway, as the current implementation is not really designed for server use cases.

TimoGlastra:
Adds a caching interface with a default `SingleContextStorageLruCache` which mimicks the behaviour of the current cache implementation. This is enabled by default for backwards compatability, but in 0.4.0 i would like to switch the default implementation to the `InMemoryLruCache`.

For react native users it would then be recommended to use the `SingleContextStorageLruCache`.

The `SingleContextStorageLruCache` does not work with multi tenancy and an error will be thrown if it does. So this doesn't fix #1227, but it does allow to register a custom implementation. Maybe the tenants module can provide a custom one at some point, but for now it can use the in memory one, or implement a cache using e.g. redis.


packages/core/src/modules/cache/InMemoryLruCache.ts
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  public async remove(agentContext: AgentContext, key: string): Promise<void> {
+    this.removeExpiredItems()
+    this.cache.delete(key)
+  }
+
+  private removeExpiredItems() {
+    this.cache.forEach((value, key) => {
+      if (value.expiresAt && Date.now() > value.expiresAt) {
```
blu3beri, line 61: we can compare `Date.getTime()` and `Date.now()`? Could very well be the case, but the Date api is always a bit confusing.

packages/core/src/modules/cache/singleContextLruCache/CacheRecord.ts
```diff
@@ -0,0 +1,41 @@
...
+  value: unknown
+  expiresAt?: number
+}
+
+export interface CacheStorageProps {
+  id?: string
+  createdAt?: Date
+  tags?: TagsBase
+
+  entries: Array<{ key: string; item: CacheItem }>
```
blu3beri, line 16: Woudn't a `Map` make more sense here?

packages/core/src/modules/cache/singleContextLruCache/SingleContextStorageLruCache.ts
```diff
@@ -0,0 +1,161 @@
...
+    )
+  }
+
+  /**
+   * Asserts this class is not used with multiple agent context instances.
+   */
+  private assertContextCorrelationId(agentContext: AgentContext) {
+    if (!this._contextCorrelationId) {
+      this._contextCorrelationId = agentContext.contextCorrelationId
+    }
```
blu3beri, line 153: Would it make sense for the assert to also set the value? Feels a bit hacky.

blu3beri (approved):
LGTM!

packages/core/src/modules/cache/singleContextLruCache/SingleContextStorageLruCache.ts
TimoGlastra: My reasoning was that it asserts it over time, and thus it keeps internal state to make sure it can make the assertion

packages/core/src/modules/cache/singleContextLruCache/CacheRecord.ts
TimoGlastra: This is the structure as used by lru_map. Can change it to a map, however that would be a breaking change to the existing cache records (I wanted to avoid breaking changes for now)

TimoGlastra: Fixed. It's now a map

Issue #1193, opened 2023-01-06
Create a migration script for AFJ updating an SQLite Indy SDK wallet to an SQLite Aries Askar wallet


TimoGlastra:
Related to: https://github.com/hyperledger/aries-framework-javascript/issues/1164

Can be based on: [andrewwhitehead/acapy-wallet-upgrade@main/bin/askar-upgrade.py (raw)](https://raw.githubusercontent.com/andrewwhitehead/acapy-wallet-upgrade/main/bin/askar-upgrade.py)

We need to create a migration script that allows to update an Indy based wallet to an Aries Askar based wallet.

In this issue only focus on the upgrade of an SQLite based Indy SDK to an SQLite based Aries Askar wallet. We'll likely make it a separate addon (TBD) because it requires an sqlite execution environment which is different per platform. 

The steps (taken from the acapy upgrade script) are:
1. Run Pre Upgrade script that directly executes SQLite on the encrypted storage
2. Register a new Askar profile using SQLite (which uses the indy key)
3. Loop through all items stored in the indy database and re-encrypt them in the askar structure
4. remove old indy SDK tables
5. Apply post upgrade transformations directly using Askar. See "Updates" for the list of updates

## Updates

Once the wallet database has been transformed to an Askar compliant interface there's still records that use the Indy SDK style context structure. All records that aren't created by AFJ, but internal by the Indy SDK (e.g. the private parts of a credential definition) need to be fetched, transformed to an storage record and the re-saved. 

Using storage records for all those objects means that from now on the storage storage is completely separated from the rest of the codebase (i.e. you can swap out askar and it will work fine)

The updates that need to happen (may be missing things):
- Fetch all `Indy::Key` records and store them as Askar key instances
  -  We haven't used key metadata in AFJ ever. Does Indy store metadata itself? If yes, we may need to migrate it
  - After migration remove the Indy::Key and Indy::KeyMetadata records
- Fetch all `Indy::MasterSecret` records and store them in an `AnonCredsMasterSecret` record (https://github.com/hyperledger/aries-framework-javascript/issues/1164)
  - After migration remove the Indy::MasterSecret record
- Fetch all `Indy::Did` records, extract the key and store it as an Askar key object.
[...]

Issue #1164, opened 2022-12-15
Records for anoncreds related objects


TimoGlastra:
Once we switch to anoncreds rs we need to manually store all anoncreds related objects. Before, this was handled internally by the indy-sdk.

To make sure the anoncreds implementation in AFJ is not dependant on Aries Askar, we need to start defining records for these objects instead of directly storing them using Askar.

Object types that will need records:
- ~~master secret~~
- ~~credential definition (we have one, but doesn't include private part)~~
  - ~~private~~
  - ~~public~~
  - ~~key correctness proof~~
- ~~revocation registry definition~~
   - ~~public~~
   - ~~private~~
- ~~revocation registry (will probably be updated to revocation list)~~
  - ~~info~~
- credential (something like W3cCredential, but AnonCreds) (@genaris)

this list is based on the acapy upgrade script: https://raw.githubusercontent.com/andrewwhitehead/acapy-wallet-upgrade/main/bin/askar-upgrade.py

> **Note** This will complicate the migration script form Indy SDK -> Aries Askar, but that'll be handled in a separate issue.

PR #1232, opened 2023-01-23
feat(anoncreds): add anoncreds API

TimoGlastra:


TimoGlastra:
revocationList -> revocationStatusList

packages/anoncreds/src/services/AnonCredsHolderService.ts
```diff
@@ -19,8 +21,10 @@ export interface AnonCredsHolderService {
...
   ): Promise<string>
 
-  // TODO: indy has different return types for the credential
-  getCredential(agentContext: AgentContext, options: GetCredentialOptions): Promise<CredentialInfo>
+  // TODO: this doesn't actually return the credential, as the indy-sdk doesn't support that
+  // We could come up with a hack (as we've received the credential at one point), but for
+  // now I think it's not that much of an issue
+  getCredential(agentContext: AgentContext, options: GetCredentialOptions): Promise<AnonCredsCredentialInfo>
 
   createCredentialRequest(
```
genaris, line 29: What about MasterSecret creation and storage? It is part of AnonCredsApi, isn't it? If that's the case, will it be passed in `CreateCredentialRequestOptions` and `CreateProofOptions`?

genaris (commented):
I know this is still a draft but I've been checking it for an integration with anoncreds-rs and I left some comments, mostly some questions to make sure I understood correctly the way these interfaces should be used.

packages/anoncreds/src/services/AnonCredsIssuerService.ts
```diff
@@ -9,6 +9,8 @@ import type { AnonCredsCredentialOffer } from '../models/exchange'
 import type { AnonCredsCredentialDefinition, AnonCredsSchema } from '../models/registry'
 import type { AgentContext } from '@aries-framework/core'
 
+export const AnonCredsIssuerServiceSymbol = Symbol('AnonCredsIssuerService')
+
 export interface AnonCredsIssuerService {
   createSchema(agentContext: AgentContext, options: CreateSchemaOptions): Promise<AnonCredsSchema>
```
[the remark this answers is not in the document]
TimoGlastra: My idea is that we add records for this in the AnonCreds package and we instruct that the same repositories are used by implementations that can use it. However, for Indy SDK we can't use the records and therefore it's part of the specific issuer/holder/verifier services.

This way there also won't be migration scripts needed, as they'll be using the same records. The idea is to create a separate record for the private parts so it's not exposed to the end user (unless they specifically fetch the private record).

See https://github.com/hyperledger/aries-framework-javascript/issues/1193 and https://github.com/hyperledger/aries-framework-javascript/issues/1164 for extra context.

Do you think this approach will work? Happy to go other directions

packages/anoncreds/src/services/AnonCredsHolderService.ts
TimoGlastra: Yeah it should be handled by the AnonCreds Api / services (prob the holder service), and removed from the wallet interface.

TimoGlastra:
> I know this is still a draft but I've been checking it for an integration with anoncreds-rs and I left some comments, mostly some questions to make sure I understood correctly the way these interfaces should be used.

Thanks for the feedback, comments!

packages/anoncreds/src/services/AnonCredsIssuerService.ts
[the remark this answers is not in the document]
genaris: Yes, I think it will be better that way, especially because we are talking about records that are aligned with AnonCreds specification, so any implementation should be fine by using it (and if it needs something else, it can probably use an internal metadata value).

The same would apply to `AnonCredsCredentialRecord`, right? I mean there will be a common Record/Repository and each `AnonCredsHolderService` implementation will use it. In such case, wouldn't be the methods `getCredential` / `deleteCredential` / `storeCredential` nearly the same for every implementation (except of indy-sdk of course)?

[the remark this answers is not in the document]
TimoGlastra: Yes probably! I think `storeCredential` in indy-sdk verifies the credential so that will be different.

TimoGlastra:
> Not completely related with this PR but it seems that `packages/anoncreds/src/formats/__tests__/LegacyIndyCredentialFormatService.test.ts` takes quite a significant time in CI (between 20 and 30 seconds). Not sure where it is getting stuck but maybe we'll need to increase the timeout for anoncreds suite.

It's actually doing integration tests (while the indy-sdk was mocked before) so that's prob what's taking so long. I've been a bit more focused on the integration tests as unit tests take A LOT of time to write and often don't really test if something actually works. But it does mean tests take longer to run ....

packages/anoncreds/src/services/AnonCredsHolderService.ts
genaris: I figured out that it's only needed for CreateCredentialRequestOptions and can even be optional, using a default one as in 99% of the cases it will be a single master secret. I've implemented some draft classes in my PR https://github.com/TimoGlastra/aries-framework-javascript/pull/18

packages/anoncreds/src/services/AnonCredsIssuerService.ts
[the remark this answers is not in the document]
TimoGlastra:
```
interface AnonCredsCredentialInfo {
        credentialId: string
        attributes: Record<string, string>
        schemaId: string
        credentialDefinitionId: string
        revocationRegistryId?: string
        credentialRevocationId?: string
}

getCredential(credentialId: string): Promise<AnonCredsCredentialInfo>
deleteCredential(credentialId: string): Promise<void>
```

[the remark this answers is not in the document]
TimoGlastra:
```
interface AnonCredsCredentialInfo {
        credentialId: string
        attributes: Record<string, string>
        schemaId: string
        credentialDefinitionId: string
        revocationRegistryId?: string
        credentialRevocationId?: string
}

getCredential(credentialId: string): Promise<AnonCredsCredentialInfo>
deleteCredential(credentialId: string): Promise<void>
```

TimoGlastra:
This is now ready for review. I've renamed master secret to link secret and opened a new issue for adding the getCredential/getCredentials methods.

I've added end to end tests for the functionality in anoncreds.test.ts which uses an in memory anoncreds registry to register/resolve all anoncreds objects

packages/anoncreds/src/services/AnonCredsHolderServiceOptions.ts
```diff
@@ -81,9 +81,19 @@ export type GetCredentialsForProofRequestReturn = Array<{
 export interface CreateCredentialRequestOptions {
   credentialOffer: AnonCredsCredentialOffer
   credentialDefinition: AnonCredsCredentialDefinition
+  masterSecretId?: string
```
genaris, line 84:
```suggestion
  linkSecretId?: string
```

packages/anoncreds/tests/InMemoryAnonCredsRegistry.ts
```diff
@@ -26,8 +28,28 @@ export class InMemoryAnonCredsRegistry implements AnonCredsRegistry {
   // we want, but the indy-sdk is picky about the identifier format.
   public readonly supportedIdentifier = /^[a-zA-Z0-9]{21,22}/
 
-  private schemas: Record<string, AnonCredsSchema> = {}
-  private credentialDefinitions: Record<string, AnonCredsCredentialDefinition> = {}
+  private schemas: Record<string, AnonCredsSchema>
+  private credentialDefinitions: Record<string, AnonCredsCredentialDefinition>
+  private revocationRegistryDefinitions: Record<string, AnonCredsRevocationRegistryDefinition>
+  // { revocationRegistryDefinitionId: { timestamp: revocationStatusList }
```
genaris, line 34: Is this commented code left for some purpose to handle later on?

genaris (changes requested):
I left some very minor comments. There are a few more details (like defining KeyCorrectnessProof and CredentialDefinitionPrivate structure) but we can handle in another PR.

packages/anoncreds/tests/InMemoryAnonCredsRegistry.ts
TimoGlastra, line 34:
```suggestion
```

TimoGlastra: It can be removed! 👍

packages/anoncreds/src/services/AnonCredsHolderServiceOptions.ts
TimoGlastra: I've applied the suggestion. It's currently not used in the IndySdkHolderService (it's using the masterSecretId from the wallet).

However, I think we can best address this in #1198

genaris (approved):

genaris:
Seems that passing anoncreds test is still hard due to timing issues... Based on the run for node 14 it took around 23 secs to pass. Maybe would be neeeded to increase the timeout to 30000?

PR #1241, opened 2023-01-25
chore: deprecate injectionContainer on agent

TimoGlastra:
Marks the `agent.injectionContainer` as deprecated. We now have the `dependencyManager` that wraps the injection container from tsyringe.

This doesn't remove the property yet, but updates all usage of it in the codebase and marks it as deprecated so we can remove it in 0.4.0.

### DIFF
--- a/packages/bbs-signatures/tests/v2.ldproof.credentials.propose-offerBbs.test.ts
+++ b/packages/bbs-signatures/tests/v2.ldproof.credentials.propose-offerBbs.test.ts
@@ -35,7 +35,7 @@ describeSkipNode17And18('credentials, BBS+ signature', () => {
       'Faber Agent Credentials LD BBS+',
       'Alice Agent Credentials LD BBS+'
     ))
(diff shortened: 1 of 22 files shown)


jleach:
I didn't notice the deprecated API in 0.3.3. If you add the following VSCode will annotate it as deprecated. In any case I've updated Bifold to use `dependencyManager` and it seems to be working fine. Nice work!

```
/**
 * @deprecated The function should not be used
 */
 ```

karimStekelenburg:
Strange... I'm also not seeing VSCode warnings although the `@deprecated` tag seems to have been added in `packages/core/src/agent/BaseAgent.ts`...

karimStekelenburg (approved):
LGTM

TimoGlastra:
> I didn't notice the deprecated API in 0.3.3.

Version 0.3.3 didn't have the `@deprecated` tag yet! I noticed in your issue that it was still available, and it should have been removed already in 0.3.0. This will give everyone that updates to 0.3.4 (when we release it) a warning to upgrade before 0.4.0

PR #1243, opened 2023-01-25
feat(openid4vc-client): pre-authorized

karimStekelenburg:
This PR adds support for the `pre-authorized` OpenID for Verifiable Credentials issuance flow to the new `openid4vc-client` module.

Here are some highlights of the work:
- Allows the user to execute the entire `pre-authorized` flow by calling a single method.
- Adds a happy-flow test
    - HTTP(S) requests and responses are mocked using a network mocking library called [nock](https://github.com/nock/nock)
    - Because the JSON-LD credential that is received is expanded by the `W3cCredentialService`, I've added a few new contexts to our test document loader.
    - Not-so-happy-flow tests will be added later on. If you have any suggestions for edge cases that deserve testing, feel free to drop a comment. 
- Modifies the `JwsService`
    - The `JwsService` was geared towards a very specific use case. I've generalized its API so it's usable for a wider range of applications.
    - All pre-existing tests and calls to the `JwsService` have been updated.

It's worth noting that I have had to add some `@ts-ignore` statements here and there to get around some incomplete types in the `OpenID4VCI-Client` library we're using. Once these issues have been resolved in the client library, they will be removed.

**Work funded by the government of Ontario**


karimStekelenburg:
The CI seems to have the same validation issue I ran into locally. For some reason, it's validating the `OpenID4VC-Client` library. Does anyone know why this is happening? I've tried to find the issue, but without success...

packages/core/src/crypto/JwsService.ts
```diff
@@ -123,3 +159,17 @@ export interface VerifyJwsResult {
   isValid: boolean
   signerVerkeys: string[]
 }
+
+export interface Jwk {
+  kty: string
```
blu3beri, line 164: Might be fine, but do we have a list of key types we support? Would be good to only allow those strings.

packages/openid4vc-client/src/OpenId4VcClientService.ts
```diff
@@ -1,10 +1,164 @@
...
+      credentialType: accessToken.scope,
+      format: 'ldp_vc', // Allows us to override the format
+    })
+
+    this.logger.debug('Credential request response', credentialResponse)
+
+    if (!credentialResponse.successBody) {
+      throw new AriesFrameworkError('Did not receive a successful credential response')
+    }
+
```
TimoGlastra, line 152: I think it would be good to verify the received credential before storing, just so we know the signatuer is valid and we'll be able to use i in the future. We also do this in the `JsonLdCredentialFormatService`.:

```
// verify signatures of the credential
    const result = await w3cCredentialService.verifyCredential(agentContext, { credential })
    if (result && !result.verified) {
      throw new AriesFrameworkError(`Failed to validate credential, error = ${result.error}`)
    }
```

packages/openid4vc-client/src/OpenId4VcClientService.ts
```diff
@@ -1,10 +1,164 @@
...
+        throw new AriesFrameworkError('No header present on JWT')
+      }
+
+      if (!jwt.payload) {
+        throw new AriesFrameworkError('No payload present on JWT')
+      }
+
+      const did = kid.split('#')[0]
+
+      const verkey = didKeyToVerkey(did)
```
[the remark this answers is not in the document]
TimoGlastra: I had to add the `did` property as an optional propety to getCreatedDids in the DidsApi and `DidsRepository`

packages/openid4vc-client/tests/setup.ts
```diff
@@ -1,3 +1,6 @@
 import 'reflect-metadata'
+import console from 'console'
+
+global.console = console
```
TimoGlastra, line 4: can your revert this?

packages/openid4vc-client/src/OpenId4VcClientService.ts
```diff
@@ -1,10 +1,164 @@
...
+
+      const payload = JsonEncoder.toBuffer(jwt.payload)
+
+      const jws = await this.jwsService.createJwsCompact(agentContext, {
+        verkey, // FIXME null check
+        header: jwt.header as unknown as Record<string, unknown>,
+        payload,
+        protectedHeaderOptions: {
+          alg: Alg.EdDSA,
```
blu3beri, line 64: Does this always use eddsa or is that only what we support right now?

packages/openid4vc-client/src/OpenId4VcClientService.ts
TimoGlastra: Not for now, just an idea. We should probably also start exposing an `W3cCredentialApi` so people can use the public api to find/store/validate w3c credentials. Then we also wouldn't have to depend on internal Apis here.

packages/openid4vc-client/src/OpenId4VcClientService.ts
```diff
@@ -1,10 +1,164 @@
...
+          kid,
+        },
+      })
+
+      return jws
+    }
+  }
+
+  private getSignCallback(agentContext: AgentContext) {
```
blu3beri, line 73: This would later also include a verify callback right? If so, the naming might be a bit off.

TimoGlastra (changes requested):
Great work. I've left a bunch of comments to reuse some methods we have to handle all cases (more did methods, etc..).

packages/openid4vc-client/src/OpenId4VcClientService.ts
```diff
@@ -1,10 +1,164 @@
...
+  public async preAuthorized(agentContext: AgentContext, options: PreAuthorizedOptions): Promise<W3cCredentialRecord> {
+    this.logger.debug('Running pre-authorized flow with options', options)
+
+    const client = await OpenID4VCIClient.initiateFromURI({
+      issuanceInitiationURI: options.issuerUri,
+      flowType: AuthzFlowType.PRE_AUTHORIZED_CODE_FLOW,
+      kid: options.kid,
+      alg: Alg.EdDSA,
+      // The clientId is set to a dummy value for now as we don't have
+      // one in the pre-auth flow, but it's currently required by OpenID4VCIClient.
```
blu3beri, line 88: Will this be fixed in the sdk that we use?

packages/openid4vc-client/src/OpenId4VcClientService.ts
```diff
@@ -1,10 +1,164 @@
...
+      metadata: serverMetadata,
+    })
+      .withTokenFromResponse(accessToken)
+      .build()
+
+    const credentialResponse = await credentialRequestClient.acquireCredentialsUsingProof({
+      proofInput,
+      // @ts-ignore
+      credentialType: accessToken.scope,
+      format: 'ldp_vc', // Allows us to override the format
```
blu3beri, line 144: Same with the eddsa, is this the only format we will support?

Mainly asking as we might allow the user to pass these values in, but only allow eddsa and ldp_vc for now.

packages/openid4vc-client/src/OpenId4VcClientService.ts
```diff
@@ -1,10 +1,164 @@
...
+
+    if (!credentialResponse.successBody) {
+      throw new AriesFrameworkError('Did not receive a successful credential response')
+    }
+
+    const credential = JsonTransformer.fromJSON(credentialResponse.successBody.credential, W3cVerifiableCredential)
+
+    const storedCredential = await this.w3cCredentialService.storeCredential(agentContext, {
+      credential: credential,
+    })
```
blu3beri, line 157:
```suggestion
    const storedCredential = await this.w3cCredentialService.storeCredential(agentContext, {
      credential,
    })
```

😛

packages/openid4vc-client/tests/setup.ts
blu3beri, line 4: Was just used for dev or do we want to keep this in?

blu3beri (commented):
Just some small items, nothing major.

packages/core/src/modules/connections/DidExchangeProtocol.ts
```diff
@@ -468,6 +469,14 @@ export class DidExchangeProtocol {
           header: {
             kid,
           },
+          protectedHeaderOptions: {
+            alg: 'EdDSA',
+            jwk: {
+              kty: 'OKP',
+              crv: 'Ed25519',
+              x: TypedArrayEncoder.toBase64URL(TypedArrayEncoder.fromBase58(verkey)),
+            },
```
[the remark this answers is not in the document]
karimStekelenburg: I've implemented this for `Ed25519` only for now. As we're currently only creating JWKs and not really receiving them, is [your comment here](https://github.com/hyperledger/aries-framework-javascript/pull/1243#discussion_r1087638517) still relevant? The `toJwk()` method will format the JWK correctly based on the key type.

It feels like the mappings you describe are mostly for verification purposes, but I'm probably missing something.

Am I?

packages/core/src/modules/vc/repository/W3cCredentialRecord.ts
```diff
@@ -44,12 +44,16 @@ export class W3cCredentialRecord extends BaseRecord<DefaultW3cCredentialTags, Cu
   }
 
   public getTags() {
+    // Contexts are usually strings, but can sometimes be objects. We're unable to use objects as tags,
+    // so we filter out the objects before setting the tags.
+    const stringContexts = this.credential.contexts.filter((ctx) => typeof ctx === 'string') as string[]
```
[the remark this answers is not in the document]
karimStekelenburg: I thought about that as well. However, let me explain my rationale.

The reason I've put this logic in `getTags()`, is because a `@context` can have many different forms. In this case, I was dealing with a credential with the following context:
```
'@context': [
      'https://www.w3.org/2018/credentials/v1',
      {
        '@vocab': 'https://w3id.org/security/undefinedTerm#'
      },
      'https://mattr.global/contexts/vc-extensions/v1',
      'https://purl.imsglobal.org/spec/ob/v3p0/context.json',
      'https://w3c-ccg.github.io/vc-status-rl-2020/contexts/vc-revocation-list-2020/v1.jsonld'
    ],
```
The `@vocab` property here, sets the default vocabulary. In other words, any type that isn't defined by one of the other contexts will fall back to the `@vocab` value. In this case, though, the `@vocab` value is also a URL. However, if you try to resolve the URL, you'll find that it's non-existent.

Apart from that, if you look at [example 7](https://www.w3.org/TR/json-ld11/#example-7-in-line-context-definition) of the JSON-LD spec, contexts that are objects can contain nested URLs as well.

Personally, I'd expect a getter called `contextUris` or `contextUrls` to return all the URLs inside the context (so also the ones that are deeply nested). Additionally, I'd expect all those URIs to be resolvable (assuming the corresponding services/domains are up).

My idea of putting this logic inside `getTags()` is that we filter out anything that can't be used as a tag only. Putting this logic in a dedicated getter on the `credential` class would make it specific to this use case, potentially causing confusion for other callers of this getter down the road.

That all being said, I'm happy to change it if you think I'm overthinking here ;)

packages/openid4vc-client/package.json
```diff
@@ -33,6 +33,7 @@
   "peerDependencies": {},
   "devDependencies": {
     "@aries-framework/node": "0.3.3",
+    "nock": "^13.3.0",
     "reflect-metadata": "^0.1.13",
```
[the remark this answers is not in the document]
karimStekelenburg: Good catch

packages/openid4vc-client/src/OpenId4VcClientApi.ts
```diff
@@ -1,7 +1,14 @@
+import type { W3cCredentialRecord } from '@aries-framework/core'
+
 import { AgentContext, injectable } from '@aries-framework/core'
 
 import { OpenId4VcClientService } from './OpenId4VcClientService'
 
+interface PreAuthorizedOptions {
```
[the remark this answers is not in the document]
karimStekelenburg: Will do.

However, I'm wondering about this often. With all the options interfaces we have, the public API gets quite polluted. Exporting the interfaces isn't needed for autocompletion and showing the user what the options are. I think the only benefit is that users can declare typed option variables like `const options: PreAuthorizedOptions = {...}`. Maybe it's just me, but I never do that.

packages/openid4vc-client/src/OpenId4VcClientService.ts
```diff
@@ -1,10 +1,164 @@
...
+import {
+  Alg,
+  AuthzFlowType,
+  CredentialRequestClientBuilder,
+  OpenID4VCIClient,
+  ProofOfPossessionBuilder,
+} from '@sphereon/openid4vci-client'
+
+import { JwsService } from '../../core/src/crypto/JwsService'
+import { didKeyToVerkey } from '../../core/src/modules/dids/helpers'
```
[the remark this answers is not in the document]
karimStekelenburg: `didKeyToVerkey` is now not needed anymore indeed. However, why do we need to export the `JwsService`?  The `kid`s are passed as strings to the `OpenId4VcClientService` at this point. Do you want to change that to `Key` instances instead?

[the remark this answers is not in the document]
karimStekelenburg: The thing here is that we would need to convert it to a string anyway, as that is what the OpenId library expects.

packages/openid4vc-client/src/OpenId4VcClientService.ts
```diff
@@ -1,10 +1,164 @@
...
+
+      return jws
+    }
+  }
+
+  private getSignCallback(agentContext: AgentContext) {
+    return {
+      signCallback: this.signCallback(agentContext),
+    }
+  }
```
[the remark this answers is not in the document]
karimStekelenburg: Our `signCallback` needs the `agentContext` in order to access the wallet and sign the JWT. I can't pass the `signCallback` like this, as it would execute the function and set the return value rather than the function itself.

```typescript
const proofInput = await ProofOfPossessionBuilder.fromAccessTokenResponse({
      accessTokenResponse: accessToken,
      callbacks: { signCallback: this.signCallback(agentContext) },
    })
```

There is probably a prettier way of doing this. So if you have suggestions, shout!

packages/openid4vc-client/src/OpenId4VcClientService.ts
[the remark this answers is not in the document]
karimStekelenburg: Thanks! Adding this as we speak. The OpenId4VCI spec doesn't specify what the proof purpose should be. Since this is about proving control over a certain DID, I'm assuming your suggestion here is right..

packages/openid4vc-client/src/OpenId4VcClientService.ts
karimStekelenburg: The verify callback probably doesn't need the `agentContext`, so we'll probably be able to solve that differently :)

packages/openid4vc-client/src/OpenId4VcClientService.ts
karimStekelenburg: We currently only support LDP credentials here. I think letting the user specify the format will complicate the API for no good reason (at this point). I'd prefer to do this once we start supporting more formats.

Also, note that although in theory, it's possible to issue AnonCreds formatted credentials, this is currently not supported by the OpenID4VCI library.

packages/openid4vc-client/tests/setup.ts
karimStekelenburg: You're right, removed it ;)

packages/openid4vc-client/src/OpenId4VcClientService.ts
karimStekelenburg: Assuming you're talking about the comment, yes it will!

packages/core/src/modules/vc/repository/W3cCredentialRecord.ts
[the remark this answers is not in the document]
TimoGlastra: Ok makes sense I think. What I mostly want to prevent is that we get multiple implementations of this that do it slightly different. But your reasoning makes sense.

packages/openid4vc-client/src/OpenId4VcClientService.ts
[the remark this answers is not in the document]
TimoGlastra: >  why do we need to export the JwsService

The openid package calls the JwsService right? For that it needs to import the JwsService from the core package. 

>  Do you want to change that to Key instances instead?

No I was refering to update the `JwsService` to use `Key` instances over verkeys

packages/openid4vc-client/src/OpenId4VcClientService.ts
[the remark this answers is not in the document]
TimoGlastra: Makes sense!

packages/openid4vc-client/src/OpenId4VcClientService.ts
```diff
@@ -1,10 +1,201 @@
...
+}
 
 @injectable()
 export class OpenId4VcClientService {
+  private logger: Logger
   private w3cCredentialService: W3cCredentialService
+  private jwsService: JwsService
 
-  public constructor(w3cCredentialService: W3cCredentialService) {
+  public constructor(agentConfig: AgentConfig, w3cCredentialService: W3cCredentialService, jwsService: JwsService) {
```
TimoGlastra, line 36: You shouldn't inject the `AgentConfig` anymore as it could change based on the tenant. Maybe it's okay for the logger (but we may add context scoped loggers soon), but it probs leads to issues at some point.

We should maybe look at making the `AgentConfig` not injectable when most classes are instantiated, so this won't even be possible

packages/openid4vc-client/src/OpenId4VcClientService.ts
```diff
@@ -1,10 +1,201 @@
...
     this.w3cCredentialService = w3cCredentialService
+    this.jwsService = jwsService
+    this.logger = agentConfig.logger
+  }
+
+  private signCallback(agentContext: AgentContext) {
+    return async (jwt: Jwt, kid: string) => {
+      // TODO should we check if the did exists here, or just let the wallet throw?
```
TimoGlastra, line 44: This is resolved right?

packages/openid4vc-client/src/OpenId4VcClientService.ts
```diff
@@ -1,10 +1,201 @@
...
+
+      // TODO: which purposes are allowed?
+      const verificationMethod = didResult.didDocument.dereferenceKey(kid, ['authentication'])
+      const { getKeyFromVerificationMethod } = getKeyDidMappingByVerificationMethod(verificationMethod)
+      const key = getKeyFromVerificationMethod(verificationMethod)
+
+      const payload = JsonEncoder.toBuffer(jwt.payload)
+
+      const jws = await this.jwsService.createJwsCompact(agentContext, {
+        key, // FIXME null check
```
TimoGlastra, line 88: This is resolved right?

packages/openid4vc-client/src/OpenId4VcClientService.ts
```diff
@@ -1,10 +1,201 @@
...
+  public async preAuthorized(
+    agentContext: AgentContext,
+    options: PreAuthorizedOptions,
+    checkRevocationState = true
+  ): Promise<W3cCredentialRecord> {
+    this.logger.debug('Running pre-authorized flow with options', options)
+
+    // The clientId is set to a dummy value for now as we don't have
+    // one in the pre-auth flow, but it's currently required by OpenID4VCIClient.
+    const clientId = 'some-client'
```
TimoGlastra, line 115: Is there an issue/PR open in the client library? Does it matter what this value is? In that case should it be an option in the API?

packages/openid4vc-client/src/OpenId4VcClientService.ts
```diff
@@ -1,10 +1,201 @@
...
+
+    const client = await OpenID4VCIClient.initiateFromURI({
+      issuanceInitiationURI: options.issuerUri,
+      flowType: AuthzFlowType.PRE_AUTHORIZED_CODE_FLOW,
+      kid: options.kid,
+      alg: Alg.EdDSA,
```
TimoGlastra, line 121: Do we check somewhere whether the kid matches with the alg? E.g. if I pass a bls did:key, will it fail somewhere?

packages/core/src/crypto/JwsService.ts
```diff
@@ -92,28 +131,33 @@ export class JwsService {
...
+    }
+
     return {
-      alg: 'EdDSA',
-      jwk: {
-        kty: 'OKP',
-        crv: 'Ed25519',
-        x: TypedArrayEncoder.toBase64URL(TypedArrayEncoder.fromBase58(verkey)),
-      },
+      alg: options.alg,
```
TimoGlastra, line 143: We don't check whether the `alg` is valid for the provided `KeyType`. Is that up to the user?

packages/openid4vc-client/src/OpenId4VcClientService.ts
TimoGlastra: It seems the JwsService also doesn't check the alg (just the key type)

packages/openid4vc-client/src/OpenId4VcClientService.ts
```diff
@@ -1,10 +1,201 @@
...
+      clientId: clientId,
+    })
+
+    const accessToken = await client.acquireAccessToken({ clientId: clientId })
+
+    this.logger.info('Fetched server accessToken', accessToken)
+
+    // We currently need the ts-ignore because the type
+    // inside of OpenID4VCIClient needs to be updated.
+    // @ts-ignore
```
TimoGlastra, line 131: Is there an issue/PR in the client lib?

packages/openid4vc-client/src/OpenId4VcClientService.ts
```diff
@@ -1,10 +1,201 @@
...
+      metadata: serverMetadata,
+    })
+      .withTokenFromResponse(accessToken)
+      .build()
+
+    const credentialResponse = await credentialRequestClient.acquireCredentialsUsingProof({
+      proofInput,
+      // @ts-ignore
+      credentialType: accessToken.scope,
+      format: 'ldp_vc', // Allows us to override the format
```
TimoGlastra, line 174: should we check though if the `foramt` is present in the issuer metadata, or will that be handled by the sphereon lib?

TimoGlastra (commented):
Some final cosmetic comments

packages/openid4vc-client/src/OpenId4VcClientService.ts
karimStekelenburg: Fixed now, removed it.

packages/openid4vc-client/src/OpenId4VcClientService.ts
karimStekelenburg: I've fixed this, and it has been merged. The client lib doesn't have an automatic release pipeline though, so it hasn't been released yet.

TimoGlastra: 👍

packages/core/src/crypto/JwsService.ts
```diff
@@ -92,28 +131,33 @@ export class JwsService {
...
   payload: Buffer
   header: Record<string, unknown>
+  protectedHeaderOptions: ProtectedHeaderOptions
 }
 
+type CreateJwsBaseOptions = Omit<CreateJwsOptions, 'header'>
+
+type CreateCompactJwsOptions = Omit<CreateJwsOptions, 'header'>
+
 export interface VerifyJwsOptions {
```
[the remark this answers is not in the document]
karimStekelenburg: Done

packages/openid4vc-client/src/OpenId4VcClientApi.ts
```diff
@@ -14,4 +21,8 @@ export class OpenId4VcClientApi {
     this.agentContext = agentContext
     this.openId4VcClientService = openId4VcClientService
   }
+
+  public async preAuthorized(options: PreAuthorizedOptions, checkRevocationState = true): Promise<W3cCredentialRecord> {
```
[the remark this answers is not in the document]
karimStekelenburg: No, that makes sense. This name doesn't say anything useful anyway.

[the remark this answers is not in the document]
karimStekelenburg: I'm going for `requestCredentialPreAuthorized`, though. Makes more sense to me when we add `requestCredentialAuthorized`.

packages/openid4vc-client/README.md
```diff
@@ -32,6 +32,136 @@ Open ID Connect For Verifiable Credentials Client Module for [Aries Framework Ja
...
+
+#### Requesting the credential (Pre-Authorized)
+
+Now a credential issuance can be requested as follows.
+
+```ts
+const w3cCredentialRecord = await agent.modules.openId4VcClient.requestCredentialPreAuthorized({
+  issuerUri,
+  kid,
+  checkRevocationState: false,
```
TimoGlastra, line 105: Nit. Doesn't have to be added in this PR

```suggestion
   // When we receive a credential, we will verify it before storing to make sure it's valid. As we don't support
   // revocation in AFJ, we set `checkRevocation` to false, so we can still receive and store revocable credentials
  checkRevocationState: false,
```

TimoGlastra (approved):

PR #1247, opened 2023-01-27
feat(indy-vdr): did:sov resolver

genaris:
Following the strategy of 'small PRs', this one adds to indy-vdr package a DID Resolver for did:sov method. It offers the same features as the one based on indy-sdk. 

The one for did:indy will be added in a further PR.


packages/indy-vdr/src/dids/IndyVdrSovDidResolver.ts
```diff
@@ -0,0 +1,99 @@
...
+
+import { addServicesFromEndpointsAttrib, sovDidDocumentFromDid } from './didSovUtil'
+
+@injectable()
+export class IndyVdrSovDidResolver implements DidResolver {
+  private indyVdrPoolService: IndyVdrPoolService
+
+  public constructor(indyVdrPoolService: IndyVdrPoolService) {
+    this.indyVdrPoolService = indyVdrPoolService
+  }
```
TimoGlastra, line 18: With the 0.3.0 release the dids module now expectes instances of dids. This is also still incorrect in the `IndySdkSovDidResolver` from the indy-sdk package, but if you look at the `IndySdkSovDidResolver` from the `core` package you can see it doesn't use constructor injection anymore.

This allows the did resolvers to be configured with their own configuration when needed (e.g. check did resolver taking an `rpcUrl`.).

(note to self: we should remove the `@injectable` from the  `IndySdkSovDidResolver` in the core package, that is not needed anymore)

packages/indy-vdr/src/dids/__tests__/IndyVdrSovDidResolver.test.ts
```diff
@@ -0,0 +1,127 @@
+import { JsonTransformer } from '@aries-framework/core'
+
+import didSovR1xKJw17sUoXhejEpugMYJFixture from '../../../../core/src/modules/dids/__tests__/__fixtures__/didSovR1xKJw17sUoXhejEpugMYJ.json'
+import didSovWJz9mHyW9BZksioQnRsrAoFixture from '../../../../core/src/modules/dids/__tests__/__fixtures__/didSovWJz9mHyW9BZksioQnRsrAo.json'
```
TimoGlastra, line 4: I think my preference would go to duplicating the fixtures, mainly because the different packages may diverge over time and it becomes easier I think when all packages are more 'isolated'.

packages/indy-vdr/src/dids/__tests__/IndyVdrSovDidResolver.test.ts
```diff
@@ -0,0 +1,127 @@
...
+import { parseDid } from '../../../../core/src/modules/dids/domain/parse'
+import { getAgentConfig, getAgentContext, mockFunction, mockProperty } from '../../../../core/tests/helpers'
```
TimoGlastra, line 6: Not for now, but we would _really_ benefit from creating a test utils package i think

packages/indy-vdr/src/index.ts
```diff
@@ -1,3 +1,7 @@
+export * from './dids'
+export * from './error'
+export * from './pool'
```
TimoGlastra, line 3: Do we want to export all these services / methods? I want to be more careful going forward in what we see as public api. I think seeing the specific things built on top of the pool service (so anoncreds registry, did resolver, did registrar) as public api means we can more easily make non breaking changes.

What do you think?

packages/indy-vdr/src/pool/IndyVdrPoolService.ts
```diff
@@ -188,7 +188,7 @@ export class IndyVdrPoolService {
       this.logger.trace(`Retrieved did '${did}' from ledger '${pool.indyNamespace}'`, result)
 
       return {
-        did: result,
+        did: { nymResponse: { did: result.dest, verkey: result.verkey }, indyNamespace: pool.indyNamespace },
```
TimoGlastra, line 191: Why is this change needed? We can already extract the namespace from the pool (which is so returned). Should we make `nymResponse` a top-level property?

TimoGlastra (approved):
Some comments, but overall looks good. Nice!

packages/indy-vdr/src/dids/IndyVdrSovDidResolver.ts
genaris: True! I've updated the implementation to take `IndyVdrPoolService` instance from `AgentContext`. I think in practice it will make easier to create an instance and pass it into dids module config.

packages/indy-vdr/src/dids/__tests__/IndyVdrSovDidResolver.test.ts
genaris: Yeah I was hesitating about that. I copied them, as it will make more sense once did:indy resolver is added to the package (as there are more specific fixtures to add).

packages/indy-vdr/src/index.ts
genaris: Yes, maybe it's better to be more selective from the beginning. 

For the moment I only picked the `IndyDidSovResolver`, which is the only class (among the existing ones) I think it will be used externally once an IndyVdrModule or so is created.

packages/indy-vdr/src/pool/IndyVdrPoolService.ts
genaris: This was added just to match `CachedDidResponse` interface that is part of the `PublicDidRequest` supposed to be returned by `getDidFromPool()` (otherwise, it wouldn't properly process the GET_NYM response). For the moment it seems to be only used internally, so probably it would be good to refactor it a bit to match the usage within the wonderful indy-vdr world.

genaris:
> Some comments, but overall looks good. Nice!

Thanks! I think I covered all the feedback. Some other changes to be added in a further PR.

I noticed it's becoming a bit harder to make CI pass: I had to increase the timeout for some tests (create DID with endpoints and resolve takes ~30 secs) and also I've seen some OutOfMemory/Garbage Collector problems like:

```
FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
 1: 0xb6b850 node::Abort() [/opt/hostedtoolcache/node/18.13.0/x64/bin/node]
 2: 0xa806a6  [/opt/hostedtoolcache/node/18.13.0/x64/bin/node]
 3: 0xd52140 v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, bool) [/opt/hostedtoolcache/node/18.13.0/x64/bin/node]
 4: 0xd524e7 v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, bool) [/opt/hostedtoolcache/node/18.13.0/x64/bin/node]
 5: 0xf2fbe5  [/opt/hostedtoolcache/node/18.13.0/x64/bin/node]
 6: 0xf420cd v8::internal::Heap::CollectGarbage(v8::internal::AllocationSpace, v8::internal::GarbageCollectionReason, v8::GCCallbackFlags) [/opt/hostedtoolcache/node/18.13.0/x64/bin/node]
```

Not sure if this could be related to the issue found in [Askar](https://github.com/hyperledger/aries-askar/issues/76). If it continues, maybe we can test with a patched ref-napi version to see how it performs.

packages/indy-vdr/src/dids/IndyVdrSovDidResolver.ts
TimoGlastra: Awesome!

TimoGlastra:
> > Some comments, but overall looks good. Nice!
> 
> Thanks! I think I covered all the feedback. Some other changes to be added in a further PR.
> 
> I noticed it's becoming a bit harder to make CI pass: I had to increase the timeout for some tests (create DID with endpoints and resolve takes ~30 secs) and also I've seen some OutOfMemory/Garbage Collector problems like:
> 
> ```
> FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
>  1: 0xb6b850 node::Abort() [/opt/hostedtoolcache/node/18.13.0/x64/bin/node]
>  2: 0xa806a6  [/opt/hostedtoolcache/node/18.13.0/x64/bin/node]
>  3: 0xd52140 v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, bool) [/opt/hostedtoolcache/node/18.13.0/x64/bin/node]
>  4: 0xd524e7 v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, bool) [/opt/hostedtoolcache/node/18.13.0/x64/bin/node]
>  5: 0xf2fbe5  [/opt/hostedtoolcache/node/18.13.0/x64/bin/node]
>  6: 0xf420cd v8::internal::Heap::CollectGarbage(v8::internal::AllocationSpace, v8::internal::GarbageCollectionReason, v8::GCCallbackFlags) [/opt/hostedtoolcache/node/18.13.0/x64/bin/node]
> ```
> 
> Not sure if this could be related to the issue found in [Askar](https://github.com/hyperledger/aries-askar/issues/76). If it continues, maybe we can test with a patched ref-napi version to see how it performs.

Hmm but Node 18 should work fine right?

TimoGlastra:
Is this ready to merge otherwise @genaris?

genaris:
> Is this ready to merge otherwise @genaris?

Yes, and if the issue with CI persists we can investigate it further to see where it is taking so much heap memory (and what's different from local node setup, as I haven't seen this particular issue locally).

PR #1253, opened 2023-02-02
feat(indy-vdr): resolver and registrar for did:indy

genaris:
An initial implementation of a resolver and registrar compliant with [did:indy](https://hyperledger.github.io/indy-did-method).

`IndyVdrIndyDidResolver` expects a full did:indy as an input and attempts to get it from the pools configured in IndyVdrPoolService. If the NYM is found, first attempts to build the DID Document combining the did/verkey and the `diddocContent`. If no diddocContent is received, it falls back to the good-old ATTRIB endpoint and reconstructs the DID Document using it.

`IndyVdrDidRegistrar` is a bit more flexible in terms of interface, as it can receive:
- a full did and a matching verkey; or
- a seed; or
- none of them

So in the first case it's assumed that the Key pair has been already created in the wallet and/or it is not needed to do so (e.g. we are anchoring a DID for someone else). In the other cases, the verkey is created and a matching did is dynamically generated.

In addition, in the options we can specify an array of `DidDocumentService` that will be added into the `diddocContent` on the NYM request, unless the user sets the flag `useEndpointAttrib`, which can be used for legacy deployments that don't support the diddocContent.

Some notes:

- Currently, only the legacy cases were tested and expected to be properly working. This means that we can create public DIDs without services or create/resolve DIDs with services using old endpoint ATTRIB mechanism
- There will be probably some other options for setting the version in NYM request (not yet supported by indy-vdr).
- DID Update and Deactivate are not yet implemented.


packages/indy-vdr/src/dids/didIndyUtil.ts
```diff
@@ -0,0 +1,76 @@
+import {
+  AriesFrameworkError,
+  convertPublicKeyToX25519,
+  DidDocument,
+  DidDocumentBuilder,
+  JsonTransformer,
+  TypedArrayEncoder,
+} from '@aries-framework/core'
+import { mergeWith, compact, transform, isUndefined, isEqual, isObject, isArray } from 'lodash'
```
TimoGlastra, line 9: I think dependencies for the framework are already heave, so I'm not too much of a fan of adding another. Espeically lodash is quite big, and hasn't had releases for the past 2 years, which I'm not sure we should start depending on now.

There's been a lot of feature additions to JS over the years that have add native utils for what lodash provides. There's even a whole website dedicated to it: https://you-dont-need.github.io/You-Dont-Need-Lodash-Underscore/#/

If you really think we need to use lodash in this case, I think it would be good to at least not import from the main bundle as that will pull in the whole lodash library (which is really big).

packages/indy-vdr/src/dids/IndyVdrSovDidResolver.ts
```diff
@@ -69,7 +73,9 @@ export class IndyVdrSovDidResolver implements DidResolver {
       )
       const response = await pool.submitReadRequest(request)
 
-      if (!response.result.data) return {}
+      if (!response.result.data) {
+        return
```
TimoGlastra, line 77: Maybe we can return `null` here to indicate there's not value (return type of `{} | void` feels a bit weird)

packages/indy-vdr/src/dids/IndyVdrIndyDidRegistrar.ts
```diff
@@ -0,0 +1,309 @@
...
+      }
+
+      // Save the did so we know we created it and can issue with it
+      const didRecord = new DidRecord({
+        id: did,
+        did,
+        role: DidDocumentRole.Created,
+        tags: {
+          recipientKeyFingerprints: didDocument.recipientKeys.map((key: Key) => key.fingerprint),
+          did,
```
TimoGlastra, line 145: I think the did is automatically added as a tag

TimoGlastra (commented):
This is great @genaris. Did a quick glance. Haven't looked at the implmentation in-depth yet

packages/indy-vdr/src/dids/didIndyUtil.ts
genaris: Yes, I don't like to add dependencies for such specific tasks. In this case I didn't care so much because I've seen that lodash is already retrieved as an indirect dependency of @aries-framework/core.  Sorry for my ignorance here, but is there any difference in practice in this case?

Anyway I'll try to get some stand-alone implementations for merge/diff.

packages/indy-vdr/tests/indy-vdr-pool.e2e.test.ts
```diff
@@ -10,9 +11,11 @@ import { IndyVdrPoolService } from '../src/pool/IndyVdrPoolService'
 import { indyDidFromPublicKeyBase58 } from '../src/utils/did'
 
 import { indyVdrModuleConfig } from './helpers'
+import '@hyperledger/aries-askar-nodejs'
 
 const indyVdrPoolService = new IndyVdrPoolService(testLogger, indyVdrModuleConfig)
-const wallet = new IndyWallet(agentDependencies, testLogger, new SigningProviderRegistry([]))
+const wallet = new AskarWallet(testLogger, new agentDependencies.FileSystem(), new SigningProviderRegistry([]))
```
TimoGlastra, line 17: Do we want to start using askar in tests with the performance issues? CI is taking extremely long

TimoGlastra (commented):
Implementation looks good. I do want to look at removing lodash though. I can also pick that up if you want

packages/indy-vdr/src/dids/IndyVdrIndyDidRegistrar.ts
```diff
@@ -0,0 +1,329 @@
...
+          )
+        }
+      }
+
+      // Build did document
+      const didDocument = didDocumentBuilder.build()
+
+      // If there are services and we are using legacy indy endpoint attrib, make sure they are suitable before registering the DID
+      if (services && useEndpointAttrib) {
+        endpointsAttribFromServices(services)
```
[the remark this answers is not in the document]
genaris: It was actually there just to verify services before doing DID registration (because when useEndpointAttrib is set, endpoints are added in a second step), but I finally updated the logic to be more clear.

packages/indy-vdr/src/dids/IndyVdrIndyDidRegistrar.ts
```diff
@@ -0,0 +1,329 @@
...
+
+      const didRepository = agentContext.dependencyManager.resolve(DidRepository)
+      await didRepository.save(agentContext, didRecord)
+
+      return {
+        didDocumentMetadata: {
+          did,
+        },
+        didRegistrationMetadata: {
+          namespace,
```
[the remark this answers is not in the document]
genaris: Done. Not sure why I was returning different fields since I based the development in a previous IndySovRegistrar from not so long ago :thinking:

packages/indy-vdr/tests/indy-vdr-pool.e2e.test.ts
genaris: In this case, it does not affect so much because the bottleneck in these tests is the `@hyperledger/indy-vdr-nodejs` iself. But I think it might be better to just wait until we solve the ref-napi issue before replacing indy-sdk for tests.

packages/indy-vdr/src/dids/IndyVdrIndyDidResolver.ts
```diff
@@ -0,0 +1,134 @@
...
+          error: 'notFound',
+          message: `resolver_error: Unable to resolve did '${did}': ${error}`,
+        },
+      }
+    }
+  }
+
+  private async buildDidDocument(agentContext: AgentContext, getNymResponseData: GetNymResponseData, did: string) {
+    // Create base Did Document
+    // We assume that verkey from GET_NYM is always a full verkey in base58
```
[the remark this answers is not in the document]
genaris: From [the spec](https://hyperledger.github.io/indy-did-method/#backwards-compatibility) I understand that abbreviated keys are not supposed to be used in modern did:indy. 

But I think it doesn't hurt to support them so I added it using old convention (as we use in `IndyVdrSovDidResolver`).

packages/indy-vdr/tests/indy-vdr-did-resolver.e2e.test.ts
```diff
@@ -38,15 +41,18 @@ const agentContext = getAgentContext({
...
   beforeAll(async () => {
     await indyVdrPoolService.connectToPools()
 
     if (agentConfig.walletConfig) {
       await wallet.createAndOpen(agentConfig.walletConfig)
     }
 
-    signerKey = await wallet.createKey({ seed: '000000000000000000000000Trustee9', keyType: KeyType.Ed25519 })
+    signerKey = await wallet.createKey({
+      secretKey: TypedArrayEncoder.fromString('000000000000000000000000Trustee9'),
```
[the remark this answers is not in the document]
genaris: Since I reverted the usage of askar within indy-vdr tests, This update will be done in a separate PR (if actually needed)

packages/indy-vdr/src/dids/didIndyUtil.ts
```diff
@@ -0,0 +1,165 @@
...
+    return { namespace, id }
+  } else {
+    throw new AriesFrameworkError(`${did} is not a valid did:indy did`)
+  }
+}
+
+const deepMerge = (a: Record<string, unknown>, b: Record<string, unknown>) => {
+  const output: Record<string, unknown> = {}
+
+  ;[...new Set([...Object.keys(a), ...Object.keys(b)])].forEach((key) => {
```
TimoGlastra, line 50: Got what I asked for 😆

packages/indy-vdr/tests/indy-vdr-did-registrar.e2e.test.ts
```diff
@@ -0,0 +1,348 @@
...
+      method: 'indy',
+      options: {
+        submitterDid: 'did:indy:pool:localtest:TL1EaPFCZ8Si5aUrqScBDt',
+        submitterVerkey: signerKey.publicKeyBase58,
+      },
+    })
+
+    expect(JsonTransformer.toJSON(didRegistrationResult)).toMatchObject({
+      didDocumentMetadata: {
+        qualifiedIndyDid: expect.stringMatching(DID_INDY_REGEX),
```
TimoGlastra, line 87: We don't really need this anymore I guess?

TimoGlastra (approved):

packages/indy-vdr/tests/indy-vdr-did-registrar.e2e.test.ts
genaris: Yes, in this case makes no sense as full did is available within the doc

PR #1272, opened 2023-02-06
fix: set updateAt on records when updating a record

KolbyRKunz:
Just some pretty simple updates to allow accurate sorting of messages based on their created time or the time they were last updated.


TimoGlastra (commented):
Nice, this has been long overdue. Can you add some tests for this?

KolbyRKunz:
The tests should now check the value of the updatedAt property.

packages/core/src/storage/__tests__/IndyStorageService.test.ts
```diff
@@ -113,6 +114,13 @@ describe('IndyStorageService', () => {
 
       expect(record).toEqual(found)
     })
+
+    it('After a save the record should have update the updatedAt property', async () => {
+      const time = new Date().getTime()
+      await sleep(1000)
+      const record = await insertRecord({ id: 'test-updatedAt' })
+      expect(record.updatedAt?.getTime()).toBeGreaterThan(time)
+    })
```
TimoGlastra, line 123: In general it's better to mock the date, as this will make the test wait for a second without specific reason which increases the overall time it takes to run tests

https://dev.to/maxpou/how-to-mock-date-with-jest-3k4b

KolbyRKunz: Instead of mocking the dates I just got the time before all the tests had started and made sure that after an update and save the updated at time was after that initial time.

KolbyRKunz:
I made the same updates to the Askar storage service as well

TimoGlastra (approved):

TimoGlastra:
Hi @KolbyRKunz, can you update the PR with main?

KolbyRKunz:
Updated to Main @TimoGlastra

genaris:
@KolbyRKunz Sorry to ask you again but can you update the branch?

KolbyRKunz:
@genaris No problem lol

PR #1278, opened 2023-02-09
feat: added endpoint setter to agent InitConfig

rapaktech:
Signed-off-by: Jim Ezesinachi <jim@animo.id>


packages/core/src/agent/AgentConfig.ts
```diff
@@ -141,6 +141,11 @@ export class AgentConfig {
     return this.initConfig.endpoints as [string, ...string[]]
   }
 
+  public set endpoints(endpoints: string[]) {
+    // if endpoints is not set, return queue endpoint
+    this.initConfig.endpoints = endpoints
+  }
```
TimoGlastra, line 147: I think we shouldn't override the init config, as it's an object that the user provided, and we should see it as an immutable object. 

So we should probably add a private property `endpoints` that is initially seeded with the `endpoints` from the init config ,but later can be overwritten. Some generic post on mutation that may be interesting to read: https://blog.sapegin.me/all/avoid-mutation/

TimoGlastra (commented):
Can you also add a test to https://github.com/rapaktech/aries-framework-javascript/blob/feat/add-endpoint-setter-to-agent-init-config/packages/core/src/agent/__tests__/AgentConfig.test.ts?

packages/core/src/agent/AgentConfig.ts
```diff
@@ -11,12 +11,14 @@ import { DidCommMimeType } from '../types'
...
 export class AgentConfig {
   private initConfig: InitConfig
+  private initConfigEndpoints: string[]
   public label: string
   public logger: Logger
   public readonly agentDependencies: AgentDependencies
 
   public constructor(initConfig: InitConfig, agentDependencies: AgentDependencies) {
     this.initConfig = initConfig
+    this.initConfigEndpoints = initConfig.endpoints ?? [DID_COMM_TRANSPORT_QUEUE]
```
TimoGlastra, line 21: This implementation changes the behaviour, and I think we should be careful here.

Previously if you would pass an empty `endpoints` array it would still return the `DID_COMM_TRANSPORT_QUEU` endpoint. Now it will use an empty array.

If I also later set it as an empty array it will actually use that empty array, and not defualt back to the `DID_COMM_TRANSPORT_QUEU` variable. 

Maybe we can structure it like this:
- We add a new, optional, private value  _endpoints to the class
- In the constructor we assign the value of `initConfig.endpoints` to the `_endpoints` variable
- We add a `set endpoints` method that assigns the input value to `_endpoints`
- We update the `get endpoints` method that will return the value of `_endpoints` if `_endpoints` is not undefined and has at least one entry. Otherwise we return `[DID_COMM_TRANSPORT_QUEUE]` as we did before. This was we don't introduce any breaking changes. This is important, because the framework assumes there's always at least one endpoint returned (hence the type `[string, ...string[]]` instead of `string[]`).

TimoGlastra (changes requested):

packages/core/src/agent/AgentConfig.ts
rapaktech: Makes sense! On it!

TimoGlastra (approved):

PR #1279, opened 2023-02-09
refactor(proofs)!: generalize proofs api and improve consistency with credentials module

TimoGlastra:
This refactors the proofs api to be more generic and allow dynamic registration of formats (which we need for the anoncreds works) in addition with making it consistent with the credentials module. With all the refactoring over the last year, although they followed the same patterns they became somewhat inconsistent.

This is based on the DIF Presentation exchange branch from @NB-MikeRichardson, but doesn't include the Presentation Exchange work yet. I've done this to avoid the need for refactoring in the presentation exchange branch.


packages/core/src/modules/proofs/protocol/ProofProtocolOptions.ts
```diff
@@ -0,0 +1,165 @@
...
+/**
+ * Get the format data payload for a specific message from a list of ProofFormat interfaces and a message
+ *
+ * For an indy offer, this resolves to the proof request format as defined here:
+ * https://github.com/hyperledger/aries-rfcs/tree/b3a3942ef052039e73cd23d847f42947f8287da2/features/0592-indy-attachments#proof-request-format
+ *
+ * @example
+ * ```
+ *
+ * type RequestFormatData = FormatDataMessagePayload<[IndyProofFormat, PresentationExchangeProofFormat], 'createRequest'>
```
genaris, line 23:
```suggestion
 * type RequestFormatData = ProofFormatDataMessagePayload<[IndyProofFormat, PresentationExchangeProofFormat], 'createRequest'>
```

packages/core/src/modules/proofs/protocol/v1/messages/V1RequestPresentationMessage.ts
```diff
@@ -1,23 +1,16 @@
...
-import { ProofRequest } from '../../../formats/indy/models/ProofRequest'
-import { ProofFormatSpec } from '../../../models/ProofFormatSpec'
 
-export interface RequestPresentationOptions {
+export interface V1RequestPresentationMessageOptions {
   id?: string
   comment?: string
-  parentThreadId?: string
-  requestPresentationAttachments: Attachment[]
+  requestAttachments: Attachment[]
```
genaris, line 13: Same as above

packages/core/src/agent/__tests__/Agent.test.ts
```diff
@@ -263,10 +259,11 @@ describe('Agent', () => {
         'https://didcomm.org/messagepickup/2.0',
         'https://didcomm.org/out-of-band/1.1',
         'https://didcomm.org/present-proof/1.0',
+        'https://didcomm.org/present-proof/2.0',
```
genaris, line 262: PPV2 was already present, right? So we were missing the feature registration before?

genaris (approved):
Nice refactoring! Protocol classes seem to be a bit overloaded but otherwise is a clean structure to add more protocols and formats in an easy way. 

I left just some minor nitpicks for outdated comments. Looking forward to the LegacyIndyProofFormatService!

packages/core/src/agent/__tests__/Agent.test.ts
TimoGlastra: Yes

packages/core/src/modules/proofs/protocol/v1/messages/V1RequestPresentationMessage.ts
TimoGlastra: I can't find the comment this is related to?

genaris: Nevermind! It was a question about pthid but deleted because I found the answer in another change

packages/core/src/modules/proofs/protocol/v2/messages/V2RequestPresentationMessage.ts
```diff
@@ -121,5 +67,9 @@ export class V2RequestPresentationMessage extends AgentMessage {
   @IsArray()
   @ValidateNested({ each: true })
   @IsInstance(Attachment, { each: true })
-  public requestPresentationsAttach!: Attachment[]
+  public requestAttachments!: Attachment[]
+
+  public getRequestAttachmentById(id: string): Attachment | undefined {
+    return this.requestAttachments.find((attachment) => attachment.id == id)
```
blu3beri, line 73:
```suggestion
    return this.requestAttachments.find((attachment) => attachment.id === id)
```

blu3beri (approved):
Some small nits, like `==` and clones. Really nice changes here!!!

packages/core/src/modules/proofs/formats/indy/util.ts
```diff
@@ -0,0 +1,252 @@
...
+function areRestrictionsEqual(
+  restrictionsA: Indy.WalletQuery[] | undefined,
+  restrictionsB: Indy.WalletQuery[] | undefined
+) {
+  // Having an undefined restrictions property or an empty array is the same
+  if (restrictionsA === undefined) return restrictionsB === undefined || restrictionsB.length === 0
+  if (restrictionsB === undefined) return restrictionsA.length === 0
+
+  // Clone array to not modify input object
+  const bList = [...restrictionsB]
```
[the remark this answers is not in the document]
TimoGlastra: In this case we're not modifying the object. The method checks if two restrictions arrays are equal, but uses splice to loop over it, which modifies the array (but not the objects inside of it) in place.

I could also find an alternative to splice, but I'm not sure it's needed to clone the whole object here

packages/core/src/modules/proofs/formats/indy/util.ts
```diff
@@ -0,0 +1,252 @@
...
+
+/**
+ * Checks whether two `names` arrays are equal. The order of the names doesn't matter.
+ */
+function areNamesEqual(namesA: string[] | undefined, namesB: string[] | undefined) {
+  if (namesA === undefined) return namesB === undefined || namesB.length === 0
+  if (namesB === undefined) return namesA.length === 0
+
+  // Check if there are any duplicates
+  if (new Set(namesA).size !== namesA.length || new Set(namesB).size !== namesB.length) return false
```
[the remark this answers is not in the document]
TimoGlastra: Hmm, this will actually make the check a lot more complex, as then we'd need to check the number of occurences for each property name and count how many we've already processed. So in this case it makes the implementation easier

packages/core/src/modules/proofs/formats/indy/util.ts
```diff
@@ -0,0 +1,252 @@
...
+   * @see https://github.com/hyperledger/aries-rfcs/blob/master/features/0037-present-proof/README.md#referent
+   *
+   * {
+   *  "referent1": [Attribute1, Attribute2],
+   *  "referent2": [Attribute3]
+   * }
+   */
+  const attributesByReferent: Record<string, V1PresentationPreviewAttributeOptions[]> = {}
+  for (const proposedAttributes of attributes ?? []) {
+    if (!proposedAttributes.referent) proposedAttributes.referent = uuid()
```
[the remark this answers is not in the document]
TimoGlastra: I already adressed it in the legayc indy proof format PR :https://github.com/TimoGlastra/aries-framework-javascript/blob/feat/legacy-indy-proof-format-service/packages/anoncreds/src/utils/createRequestFromPreview.ts#L42

packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -447,8 +447,8 @@ export class ConnectionService {
       previousSentMessage,
       previousReceivedMessage,
     }: {
-      previousSentMessage?: AgentMessage
-      previousReceivedMessage?: AgentMessage
+      previousSentMessage?: AgentMessage | null
```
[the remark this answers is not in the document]
TimoGlastra: We can avoid it. I this case it often led to a clunky API as we would often get `null` back from a find message call and then had to do `message ?? null`. This way we can pass both. I can revert it if you want

PR #1282, opened 2023-02-10
feat: add devcontainer support

### LINKED ISSUE #1281: Add a devcontainer support to the project to simplify developer on-boarding
Running AFJ requires a few things to be setup on a developers machine before one can start contributing. Notably on macs the `python` command does not work out-of-the-box and the indy libs noted below are missing. I think adding a devcontainer support may make life easier. Also, I think it'll ensure all devs have a similar env to work in. Figure work can look at making this a [codespace](https://github.com/codespaces) to take it to the next level. 

Missing Indy files for Mac (amd64 & arm64), Windows (amd64 & arm64), Linux, etc.

1. /usr/include/indy
2. /usr/lib/libindy.a
3. /usr/lib/libindy.so

jleach:
Initial work on adding devcontainer support to make development easier for Mac, Windows and Linux. The [Dockerfile](.devcontainer/Dockerfile) located in the `.devcontainer` directory is used to build the working container.

Fixes #1281

screenshot

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,43 +1,20 @@
-FROM ubuntu:21.10
+# arm + amd compatible Dockerfile
+FROM ghcr.io/findy-network/findy-base:indy-1.16.ubuntu-18.04 AS indy-base
 
-ENV TZ=UTC
-RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && \
-    echo $TZ > /etc/timezone
+FROM ubuntu:18.04
 
-RUN apt-get update && \
-    apt-get install -y tzdata curl gnupg git \
-    software-properties-common iproute2
+# install indy deps and files from base
+RUN apt-get update && apt-get install -y libsodium23 libssl1.1 libzmq5 git zsh
 
-RUN apt-get install -y \
-    build-essential \
-    pkg-config \
-    cmake \
-    libssl-dev \
-    libsqlite3-dev \
-    libzmq3-dev \
-    libncursesw5-dev
+COPY --from=indy-base /usr/include/indy /usr/include/indy
+COPY --from=indy-base /usr/lib/libindy.a /usr/lib/libindy.a
+COPY --from=indy-base /usr/lib/libindy.so /usr/lib/libindy.so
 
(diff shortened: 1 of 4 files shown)


Vickysomtee (commented):
I had this set up and tested, looks good to me. I think this can be approved?

karimStekelenburg (approved):
Also tested it locally and it LGTM! Looks like there are some validation issues and it needs to be rebased with main :)

TimoGlastra (approved):
Maybe we can add a short section to the devreadme that this is an alternative, more simple approach to start developing on the framework?

jleach:
@TimoGlastra Updated DEVREADME with some comments.

jleach:
@TimoGlastra @karimStekelenburg @Vickysomtee I think my next step will be to use a docker-compose for the devContainer and add Postgres and Ledger setup so its a 1-stop-shop for all your contributor needs :)

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+COPY --from=indy-base /usr/lib/libindy.so /usr/lib/libindy.so
+
+RUN apt-get install -y curl python3 build-essential ca-certificates && \
+  curl https://raw.githubusercontent.com/creationix/nvm/master/install.sh | bash && \
+  export NVM_DIR="$HOME/.nvm" && \
+  [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh" && \
+  nvm install v16 && \
+  npm install yarn -g
+
+RUN sh -c "$(curl -fsSL https://raw.githubusercontent.com/ohmyzsh/ohmyzsh/master/tools/install.sh)"
```
blu3beri, line 20: Is there a specific reason to install oh my zsh? I think it would be good to keep the container unopinionated and just use the default shell environment.

jleach: When you try the container it starts the default shell 'bash'. You have only seen this when you typed `zsh` to run an alternative shell w/ ohmyzsh. In any case, I removed `ohmyzsh` to move the PR along.

TimoGlastra:
Thanks!

PR #1283, opened 2023-02-10
feat(anoncreds): legacy indy proof format service

TimoGlastra:
Adds a `LegacyIndyProofFormatService` that can be used with v1 and v2 of the proof protocol.

I've made some improvements to the tails handling and it now verifies the hash after downloading. 

There's some high level tests that use both the legacy indy credential format service and the legacy indy proof format service to do the full flow of issuance and verification. It's a bit hard to test in isolation with the indy-sdk, but we can change that once we have the AnonCredsRs backend.

Dependant on #1279


packages/anoncreds/src/models/AnonCredsProofRequest.ts
```diff
@@ -0,0 +1,83 @@
...
+import { AnonCredsRequestedAttribute } from './AnonCredsRequestedAttribute'
+import { AnonCredsRequestedPredicate } from './AnonCredsRequestedPredicate'
+import { AnonCredsRevocationInterval } from './AnonCredsRevocationInterval'
+
+export interface AnonCredsProofRequestOptions {
+  name: string
+  version: string
+  nonce: string
+  nonRevoked?: AnonCredsRevocationInterval
+  ver?: '1.0' | '2.0'
```
genaris, line 18: I know this comes from the spec but it seems to me that will be very confusing:

- ver is an optional string, specifying the presentation request version
- version is a string set by the verifier, the version of the presentation request

packages/anoncreds/src/models/__tests__/AnonCredsRestriction.test.ts
```diff
@@ -0,0 +1,80 @@
...
+          },
+        }),
+      ],
+    })
+
+    expect(JsonTransformer.toJSON(restrictions)).toMatchObject({
+      restrictions: [
+        {
+          'attr::test_prop::value': 'test_value',
+          //   'attr::test_prop2::value': 'test_value2',
```
genaris, line 73: Why these are not included?

genaris (approved):
I liked the usage of the modern satisfied operator, making worthy the recent upgrade to TypeScript 4.9 :stuck_out_tongue: 

I found a few issues when integrating it with Askar/AnonCredsRs but I think they are not related to changes on this PR, so surely will be part of a fix PR once I debug it more.

packages/anoncreds/src/models/__tests__/AnonCredsRestriction.test.ts
TimoGlastra: They should be. Also not sure :)

packages/anoncreds/src/models/AnonCredsProofRequest.ts
TimoGlastra: Yup that's very weird. We don't expose them in the AFJ API as 2.0 is the default I think when you don't provide it

packages/anoncreds/src/formats/__tests__/LegacyIndyProofFormatService.test.ts
```diff
@@ -4,6 +4,8 @@ import {
   SigningProviderRegistry,
```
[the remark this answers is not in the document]
genaris: I'm sorry, I commented in the wrong line. The class missing in the exports is `LegacyIndyProofFormatService`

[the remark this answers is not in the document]
TimoGlastra: Ah good shout. Yes, that one should be exported

genaris:
> > I liked the usage of the modern satisfied operator, making worthy the recent upgrade to TypeScript 4.9 stuck_out_tongue
> 
> Yup really like the operator. It's often exactly what you need: verify something conforms to a type

When running an e2e test using anoncreds-rs I'm experiencing an issue that maybe it's related to this. I create a present proof much in the same way than `presentProof` function in `packages/core/tests/helpers.ts` (but using v1 protocol instead of v2).

Logging the proof request object that `AnonCredsRsHolderService.createProof()` receives, I see the following:

```json
{"name":"test-proof-request",
"version":"1.0",
"nonce":"928479155009903779791520",
"requested_attributes":{
  "name":{
     "name":"name",
     "restrictions":[{"credentialDefinitionId":"TL1EaPFCZ8Si5aUrqScBDt:3:CL:15449:Employee Credential"}]}},
"requested_predicates":{}}
```

Note that camelCase in **credentialDefinitionId**, when AnonCredsRestriction model uses cred_def_id. Are we somewhere missing a JSONTransformer.toJSON()?

Issue #1261, opened 2023-02-03
Enable toggling of return route (true, false) during credential issuance and present proof flows


NB-MikeRichardson:
During a connectionless proof request, the wallet has to wait for a network timeout as the request has return route set to true. This slows the user experience as the proof is accepted while the wallet spinner continues for another 10 seconds or so.

A similar fix was put in place for connection to the mediator on wallet initiation. It may make sense to implement this as a generic solution. Will leave that up to the AFJ community to discuss.

TimoGlastra:
Relevant PR in ACA-Py for why this is happening with ACA-Py agents: https://github.com/hyperledger/aries-cloudagent-python/pull/1853

TimoGlastra:
It hasn't been merged yet, right?

NB-MikeRichardson:
Sorry I thought this was merged.

I have created a new PR as the previous one was broken by recent changes; it was quicker to redo the changes in a new branch

PR #1284, opened 2023-02-10
feat: optional return route

NB-MikeRichardson:
Provides a fix for https://github.com/hyperledger/aries-framework-javascript/issues/1261

This sets returnRoute to true by default but allows it to be overridden in the Proofs API. 

Has been tested with aca-py [verifier] agent with AFJ [Holder].  

I decided against testing for existence of "please-ack" flag and setting the value based on that, as I think that's a bit too specific to one use case, and this is really a temporary workaround pending changes made in aca-py and/or a more generic solution in AFJ.


NB-MikeRichardson:
17.x CI tests failing with error:

`FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
`

NB-MikeRichardson:
Closed by mistake

PR #1285, opened 2023-02-10
feat: indy-vdr module registration

karimStekelenburg:


packages/indy-vdr/src/IndyVdrModuleConfig.ts
```diff
@@ -0,0 +1,21 @@
+export interface IndyVdrNetworkConfig {
```
TimoGlastra, line 1: I think we already have a similar interface in the IndyVdrPoolService

TimoGlastra: This is missing the transactionAuthorAgreement, maybe we can reuse that type?

packages/indy-vdr/src/IndyVdrModuleConfig.ts
```diff
@@ -0,0 +1,21 @@
...
+  indyNamespace: string
+  genesisTransaction: string
+  isProduction: boolean
+}
+
+export interface IndyVdrModuleConfigOptions {
+  networks: IndyVdrNetworkConfig[]
```
TimoGlastra, line 8: Some documentation you can reuse from the indy sdk module:

```
  /**
   * Array of indy networks to connect to.
   *
   * @default []
   *
   * @example
   * ```
   * {
   *   isProduction: false,
   *   genesisTransactions: 'xxx',
   *   indyNamespace: 'localhost:test',
   *   transactionAuthorAgreement: {
   *     version: '1',
   *     acceptanceMechanism: 'accept'
   *   }
   * }
   * ```
   */
  networks?: IndySdkPoolConfig[]
```

TimoGlastra (changes requested):
I think we're missing some changes to the IndyVdrPoolService that leverages the networks from the config.

Currently you have to call `setPools()`, which isn't called.

In the IndySdkPoolService I inject the `IndySdkModuleConfig` and assign the pools property based on the `networks` property in the config and have removed the `setPools` property.

TimoGlastra (approved):
LGTM, we should probably look at the exports as the module etc.. are not exported from the root of the package it seems

PR #1286, opened 2023-02-10
refactor!: remove indy from core

TimoGlastra:
The long awaited change. This pr removes all dependency on the indy-sdk from the core package. The indy-sdk is still availalbe through a custom package. 

Tests in core still depend on indy-sdk, mainly because askar storage is slow atm, and we still need to do some more work to create lighter tests that only depend on e.g. an in memory wallet.

It's a REALLY BIG pr, as I've approached this in the dumbest way possible. I started out by removing `indy` from the agent dependencies, which led to an error, and then I was three days in needing to rewrite 90% of tests because they all dependend on the indy-sdk. 

I think this is the last really big PR, after which we're going to need to do a lot of small improvements before we can make the 0.4.0 release.

That said, I was able to remove over 9000 lines of code

pppoof

I also updated the demo where one side uses askar and the other indy, but it's currently broken due to the seed not deriving the correct did as it was registered on the netwrok as it should actually use fromSecretKey. Once we can create keys from the secretKey in AskarWallet we can create the correct did.


demo/src/Alice.ts
```diff
@@ -8,7 +8,7 @@ export class Alice extends BaseAgent {
   public connectionRecordFaberId?: string
 
   public constructor(port: number, name: string) {
-    super(port, name)
+    super({ port, name, useSharedComponents: false })
```
genaris, line 11: I just would like to settle also in GitHub my feelings about using this term because I know it's clear for everyone involved in this for some time that it came from 'Indy shared components'. But I don't want to commit the same mistake of carrying the legacy forever and confuse newcomers (I'm basing these words on my own experience).

For the time being I think it's nice to reflect it in this demo and be explicit that we support both the Indy SDK and the 'replacement shared components', as this is a major stepping stone for AFJ and we are super excited about it.

But in a near future I would like to simply set a demo based in Askar or the InMemoryWallet, showing the exchange of AnonCreds and JSON-LD credentials and making this modular architecture to seem natural and clearly ledger/credential format agnostic.

demo/src/BaseAgent.ts
```diff
@@ -10,46 +36,165 @@ const bcovrin = `{"reqSignature":{},"txn":{"data":{"data":{"alias":"Node1","blsk
...
 
   public async initializeAgent() {
     await this.agent.initialize()
+
+    // FIXME:
+    // We need to make sure the key to submit transactions is created. We should update this to use the dids module, and allow
+    // to add an existing did based on a seed/secretKey, and not register it on the the ledger. However for Indy SDK we currently
+    // use the deprecated publicDidSeed property (which will register the did in the wallet), and for Askar we manually create the key
+    // in the wallet. Additional issue is that when creating a key in askar using the seed, it will give different results than indy-sdk
+    // but if we create the key in askar from the seed, but use it as the secret key, it will work.
```
genaris, line 104: I think we can attempt to review/merge #1301 before this one, so we can have it clean and be able to do the demo right from the main branch.

packages/anoncreds/src/protocols/proofs/v1/V1ProofProtocol.ts
```diff
@@ -64,20 +54,17 @@ import {
...
 
-export class V1ProofProtocol extends BaseProofProtocol implements ProofProtocol<[IndyProofFormatServiceLike]> {
-  private indyProofFormat: ProofFormatService<IndyProofFormat>
+export class V1ProofProtocol extends BaseProofProtocol implements ProofProtocol<[LegacyIndyProofFormatService]> {
+  private indyProofFormat: LegacyIndyProofFormatService
 
   public constructor({ indyProofFormat }: V1ProofProtocolConfig) {
     super()
 
+    // TODO: just create a new instance of LegacyIndyProofFormatService here so it makes the setup easier
```
genaris, line 67: This is true, as we are always using a `LegacyIndyProofFormatService`. It will represent that we'll use separate instances for V1 and V2 protocols. If we foresee that most agents will support both protocols, it won't be so different in terms of simplicity.

genaris (approved):
Even if there were a massive amount of lines and files involved in this PR, it was quite straightforward to review. Maybe it's because of the excitement!

I have not commented a lot because, even if there are some TODOs and remaining improvements, we can tackle them afterwards. 

The only functional doubt I have is regarding the V1 credential and proof protocol feature registration but otherwise you have my blessing!

TimoGlastra:
> The only functional doubt I have is regarding the V1 credential and proof protocol feature registration but otherwise you have my blessing!

Can you elaborate on this part?

demo/src/Alice.ts
TimoGlastra: Good point, fully agree!

demo/src/BaseAgent.ts
TimoGlastra: Agreed! Just approved

packages/anoncreds/src/protocols/proofs/v1/V1ProofProtocol.ts
TimoGlastra: My initial approach was that you register credential formats also in the agent and the protocols could hook into it. But I changed it so you have to provide the formats yourself. Do you think we should change it?

What do you think will be the simplest way to give users flexibility?

TimoGlastra:
Thanks for the quick review! That must be around 1000 lines / hour since the Pr was ready for review!

genaris:
> > The only functional doubt I have is regarding the V1 credential and proof protocol feature registration but otherwise you have my blessing!
> 
> Can you elaborate on this part?

I don't know if it was me who forgot to push a button, or GitHub that lost my comment... But was about registering V1 credential/proof protocol in the FeatureRegistry in AnonCredsModule. I haven't seen it among the changes so I suppose they are forgotten.

packages/anoncreds/src/protocols/proofs/v1/V1ProofProtocol.ts
genaris: Oh, I think I misunderstood the comment. I thought that you would like to remove indyProofFormat from constructor parameters and simply do a `this.indyProofFormat = new LegacyIndyProofFormatService()`.

But I think it's fine the way it is, letting to provide the formats we want.

demo/src/Alice.ts
TimoGlastra: I've changed the parameter name to `useLegacyIndySdk` for now

PR #1289, opened 2023-02-12
feat: indy sdk aries askar migration script

blu3beri:
- feat: setup package
- feat: setup some base components


packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
```diff
@@ -0,0 +1,198 @@
...
+        })
+        updateCount++
+      }
+      await txn.commit()
+    }
+
+    console.log(`Updated ${updateCount} instances inside ${category}`)
+  }
+
+  private credentialTags(credentialData: Record<string, unknown>) {
```
blu3beri, line 168: @TimoGlastra You mentioned to me this would be good to have in another package. Where would that be?

packages/indy-sdk-to-askar-migration/package.json
```diff
@@ -25,28 +25,9 @@
...
-    "class-validator": "0.13.1",
-    "crypto-js": "4.1.1",
-    "react-native-fs": "2.20.0",
-    "react-native-sodium": "0.4.0",
-    "react-native-sqlite-storage": "^6.0.1",
-    "rxjs": "^7.2.0",
-    "sodium-javascript": "0.8.0",
-    "text-encoding-polyfill": "^0.6.7",
-    "tiny-msgpack": "2.0.2",
-    "uint8arrays": "^4.0.3"
```
TimoGlastra: 🚀

blu3beri: Yeah that felt good :). Also, now we directly have node.js support.

TimoGlastra: Ah didn't think about that. That is awesome

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
TimoGlastra: We have the `AnonCredsCredentialRecord`. that if you create it with all the required fields, you can call `getTags` and it will automatically have the correct format

packages/indy-sdk-to-askar-migration/package.json
```diff
@@ -0,0 +1,40 @@
...
+    "prepublishOnly": "yarn run build",
+    "test": "jest"
+  },
+  "dependencies": {
+    "@aries-framework/anoncreds": "0.4.0-alpha.71",
+    "@aries-framework/askar": "0.4.0-alpha.71",
+    "@aries-framework/core": "0.4.0-alpha.71",
+    "@aries-framework/indy-sdk": "0.4.0-alpha.71",
+    "@aries-framework/node": "0.4.0-alpha.71",
+    "@hyperledger/aries-askar-shared": "file:/Users/beri/Developer/work/hyperledger/aries-askar/wrappers/javascript/aries-askar-shared"
```
blu3beri, line 32: Will replace the local files when they are released again.

genaris: Shouldn't the reference to `@aries-framework/*` packages point to * instead of an specific version (0.4.0-alpha.71)?

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
```diff
@@ -0,0 +1,363 @@
...
+  }
+
+  private async updateKeys() {
+    const category = 'Indy::Key'
+
+    this.agent.config.logger?.trace(`[indy-sdk-to-askar-migration]: Updating category: ${category}`)
+
+    let updateCount = 0
+    const session = this.store.transaction()
+    for (;;) {
```
genaris, line 238: Nice to see this in TypeScript!

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
```diff
@@ -0,0 +1,363 @@
...
+
+import { IndySdkToAskarMigrationError } from './errors/IndySdkToAskarMigrationError'
+import { transformFromRecordTagValues } from './utils'
+
+/**
+ *
+ * Migration class to move a wallet form the indy-sdk structure to the new
+ * askar wallet structure.
+ *
+ * Right now, this is ONLY supported within React Native environments AND only sqlite.
```
genaris, line 18: I think this disclaimer is fine here but should be part of the README. 

I understand the reasons about stating that it works in React Native because it assumes that no issuer or verifier side has been supported there officially, but I don't think it is accurate as we can also have a NodeJS _holder-role_ agent that would be in the same situation.

Also, what about mediation records? Mediator configuration is not taken into account. In that case you assume only current Bifold case where a mediatorInvite is provided at startup and there is apparently no problem on creating the mediation coordination again. But that's not necessarily the case of every app using AFJ.

What about apps using generic records module? And those that have custom plug-ins that store custom records? 🤯 

The latter case is maybe extreme (don't worry for us, as we are starting production using 0.4.0 right away 😎 ), but generic records and maybe proofs are used in some mobile wallets.

packages/indy-sdk-to-askar-migration/package.json
```diff
@@ -0,0 +1,40 @@
...
+  },
+  "devDependencies": {
+    "@hyperledger/aries-askar-nodejs": "file:/Users/beri/Developer/work/hyperledger/aries-askar/wrappers/javascript/aries-askar-nodejs",
+    "indy-sdk": "1.16.0-dev-1655",
```
genaris, line 36: It's needed to fix the version to 1655 or can be aligned with other modules that have ^1655?

yarn.lock
```diff
@@ -10,6 +10,88 @@
     "@jridgewell/gen-mapping" "^0.1.0"
     "@jridgewell/trace-mapping" "^0.3.9"
 
+"@aries-framework/anoncreds@0.4.0-alpha.71", "@aries-framework/anoncreds@0.4.0-alpha.71+a4204ef2":
```
genaris, line 13: These additions here can be avoided if using * for modules in package.json

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
```diff
@@ -0,0 +1,363 @@
...
+      if (!keys || keys.length === 0) {
+        await txn.close()
+        break
+      }
+
+      for (const row of keys) {
+        await txn.remove({ category, name: row.name })
+        const signKey: string = JSON.parse(row.value as string).signkey
+        const keySk = TypedArrayEncoder.fromBase58(signKey)
+        const key = Key.fromSecretBytes({
```
genaris, line 250: Maybe not significant in most practical cases, but might be good to clear each key handle after usage

genaris (changes requested):
Didn't do a deep review yet but left a few acid remarks regarding the scope and to :clarifying it 💥

packages/indy-sdk-to-askar-migration/package.json
TimoGlastra: I think this should use 0.3.3 (will be updated when released)

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
TimoGlastra: I think all the cases you described should work in Node.JS, however it's not too much work to add credential definitions (the only one missing) after which it will fully work in Node.JS. We can do that in an upcoming release I think

packages/indy-sdk-to-askar-migration/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+```sh
+npm info "@aries-framework/indy-sdk-to-askar-migration" peerDependencies
+```
+
+Then add the indy-sdk-toaskar-migration module to your project.
+
+```sh
+yarn add @aries-framework/indy-sdk-to-askar-migration
+```
```
TimoGlastra, line 45: We don't use peer dependencies anymore

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
```diff
@@ -0,0 +1,363 @@
...
+  }: {
+    dbPath: string
+    agent: Agent
+    deleteOnFinish?: boolean
+    defaultLinkSecretId?: string
+  }) {
+    const {
+      config: { walletConfig },
+    } = agent
+    if (!walletConfig) throw new IndySdkToAskarMigrationError('Wallet config is required for updating the wallet')
```
TimoGlastra, line 66: Can we add a check that the `walletConfig.storage` is undefined (meaning default sqlite is used) or `walletConfig.storage.type` is `sqlite` and if not throw an error? That way it will fail if you try to do this with a postgres config.

TimoGlastra, line 66: Can we also add a check if we're in Node.JS and throw an error?

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
```diff
@@ -0,0 +1,363 @@
...
+
+  /**
+   * Backup the database file. This function makes sure that the the indy-sdk
+   * database file is backed up within our temporary directory path. If some
+   * error occurs, `this.revertDatbase()` will be called to revert the backup.
+   */
+  private async backupDatabase() {
+    const src = this.dbPath
+    const dest = this.backupFile
+    this.agent.config.logger?.trace(`Creating backup from '${src}' to '${dest}'`)
```
TimoGlastra, line 126:
```suggestion
    this.agent.config.logger.trace(`Creating backup from '${src}' to '${dest}'`)
```

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
```diff
@@ -0,0 +1,363 @@
...
+      throw new IndySdkToAskarMigrationError('Could not locate the new backup file')
+    }
+  }
+
+  /**
+   * Reverts backed up database file to the original path, if its missing, and
+   * deletes the backup. We do some additional, possible redundant, exists checks
+   * here to be extra sure that only a happy flow occurs.
+   */
+  private async revertDatabase() {
```
TimoGlastra, line 144:
```suggestion
  private async restoreDatabase() {
```

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
```diff
@@ -0,0 +1,363 @@
...
+    // "Impossible" state. Since we do not continue if `this.backupDatabase()`
+    // fails, this file should always be there. If this error is thrown, we
+    // cannot correctly revert the state.
+    if (!(await this.fs.exists(this.backupFile))) {
+      throw new IndySdkToAskarMigrationError('Backup file could not be found while trying to revert the state')
```
TimoGlastra, line 149:
```suggestion
      throw new IndySdkToAskarMigrationError('Backup file could not be found while trying to restore the state')
```

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
```diff
@@ -0,0 +1,363 @@
...
+    }
+
+    /**
+     * Since we used `copy` to get the file, it should still be there. We
+     * double-check here to be sure.
+     */
+    if (!(await this.fs.exists(this.dbPath))) {
+      return
+    } else {
+      this.agent.config.logger?.trace(`Moving '${this.backupFile}' back to the original path: '${this.dbPath}`)
```
TimoGlastra, line 159:
```suggestion
      this.agent.config.logger.trace(`Moving '${this.backupFile}' back to the original path: '${this.dbPath}`)
```

TimoGlastra: Seem this everywhere in the file, it's not needed right?

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
```diff
@@ -0,0 +1,363 @@
...
+        const key = Key.fromSecretBytes({
+          algorithm: KeyAlgs.Ed25519,
+          secretKey: keySk.subarray(0, 32),
+        })
+        await txn.insertKey({ name: row.name, key })
+        updateCount++
+      }
+      await txn.commit()
+    }
+
```
TimoGlastra, line 259: do we also need to close the sesion?

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
```diff
@@ -0,0 +1,363 @@
...
+    for (;;) {
+      const txn = await session.open()
+      const keys = await txn.fetchAll({ category, limit: 50 })
+      if (!keys || keys.length === 0) {
+        await txn.close()
+        break
+      }
+
+      for (const row of keys) {
+        await txn.remove({ category, name: row.name })
```
TimoGlastra, line 247: Can we add a log for each of the keys that is migrated? trace is fine, but it will be really nice to have this when doing a large migration

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
```diff
@@ -0,0 +1,363 @@
...
+          algorithm: KeyAlgs.Ed25519,
+          secretKey: keySk.subarray(0, 32),
+        })
+        await txn.insertKey({ name: row.name, key })
+        updateCount++
+      }
+      await txn.commit()
+    }
+
+    this.agent.config.logger?.trace(
```
TimoGlastra, line 260: Can you reorder the severity a bit. I think start/finished for a category can be of level info, while each key that is migrated could be of severity debug. Even smaller steps could be trace etc...

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
```diff
@@ -0,0 +1,363 @@
...
+        await txn.close()
+        break
+      }
+
+      if (!masterSecrets.some((ms: EntryObject) => ms.name === this.defaultLinkSecretId)) {
+        throw new IndySdkToAskarMigrationError('defaultLinkSecretId can not be established.')
+      }
+
+      for (const row of masterSecrets) {
+        const isDefault = masterSecrets.length === 0 ?? row.name === this.walletId
```
TimoGlastra, line 286: Can you also add logs here, and specific which is set as default?

packages/indy-sdk-to-askar-migration/src/utils.ts
```diff
@@ -0,0 +1,37 @@
+import type { TagsBase } from '@aries-framework/core'
+
+export const transformFromRecordTagValues = (tags: TagsBase): { [key: string]: string | undefined } => {
```
TimoGlastra, line 3:
```suggestion
/**
* Adopted from `AskarStorageService` implementation and should be kept in sync.
*/
export const transformFromRecordTagValues = (tags: TagsBase): { [key: string]: string | undefined } => {
```

packages/indy-sdk-to-askar-migration/tests/migrate.test.ts
```diff
@@ -0,0 +1,66 @@
...
+
+describe('Migrate', () => {
+  const config = {
+    label: 'test-agent',
+    walletConfig: {
+      id: `walletwallet.0-${utils.uuid()}`,
+      key: 'GfwU1DC7gEZNs3w41tjBiZYj7BNToDoFEqKY6wZXqs1A',
+      keyDerivationMethod: KeyDerivationMethod.Raw,
+    },
+    logger: new ConsoleLogger(LogLevel.trace),
```
TimoGlastra, line 20:
```suggestion
```

packages/indy-sdk-to-askar-migration/tests/migrate.test.ts
```diff
@@ -0,0 +1,66 @@
...
+  })
+
+  const oldDbPath = `${homedir()}/.indy_client/wallet/${oldAgent.config.walletConfig?.id}/sqlite.db`
+
+  beforeAll(() => {
+    registerAriesAskar({ askar: ariesAskar })
+  })
+
+  // TODO: update with an aca-py issued revokable credential
+  // community agent MIGHT have revocrevoc
```
TimoGlastra, line 42: community agent has a tails server set-up so that should work

packages/indy-sdk-to-askar-migration/tests/migrate.test.ts
```diff
@@ -0,0 +1,66 @@
...
+    const record = await oldAgent.genericRecords.save({ content: genericRecordContent })
+
+    await oldAgent.shutdown()
+
+    await Migration.migrate({
+      walletName: config.walletConfig.id,
+      walletKey: config.walletConfig.key,
+      kdfLevel: config.walletConfig.keyDerivationMethod,
+      specUri: oldDbPath,
+    })
```
TimoGlastra, line 57: I would expect this is handled by the `IndySdkToAskarMigrationUpdater` rather than having to call this ourselves?

packages/indy-sdk-to-askar-migration/package.json
TimoGlastra: Are there any blockers for a new release? Otherwise we can cut one today

blu3beri: Yes, the. migration addition. So just releasing the wrappers would be good.

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
blu3beri: Yeah, apparently it was a fix for aslant not allowing `while(true)` :). This might be some code-smell, but for now I just wanted to be consistent with aca-py. Also, never knew this was also possible in TS.

packages/indy-sdk-to-askar-migration/package.json
blu3beri: Yeah I am fine with that. Personally, I just always lock them to a specific version as I have encountered my times than I would like to admit that a breaking change is pushed on a patch version. So I just did it out of habit here :).

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
blu3beri: Yep good one!

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
blu3beri: Yes, good catch! I am all for more strictness on this one.

blu3beri: Yeah, would we do that with the instance type of the `dependencies` or something?

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
blu3beri: I think it's leftover from some previous code of the logger. We should add an eslint rule to not use `?` if it can't be `null`/`undefined` :).

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
blu3beri: Yeah can do! Just something like `Migrating ${row.name} to the new askar format`?

packages/indy-sdk-to-askar-migration/tests/migrate.test.ts
blu3beri: Yeah I was thinking a bit about this as well. I am okay with calling this from within the updater. Will fix.

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
```diff
@@ -0,0 +1,363 @@
...
+
+    if (agent.isInitialized)
+      throw new IndySdkToAskarMigrationError('Wallet migration can not be done on an initialized agent')
+
+    if (!(agent.dependencyManager.resolve(InjectionSymbols.Wallet) instanceof AskarWallet)) {
+      throw new IndySdkToAskarMigrationError("Wallet on the agent must be of instance 'AskarWallet'")
+    }
+
+    const keyMethod =
+      walletConfig.keyDerivationMethod == KeyDerivationMethod.Raw ? StoreKeyMethod.Raw : StoreKeyMethod.Kdf
```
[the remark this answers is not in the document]
blu3beri: For askar we just have to supply `raw`, `kdf` or `none`. Askar can detect which kdf is used.

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
TimoGlastra: ```typeof process?.versions?.node !== 'undefined'```

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
TimoGlastra: Yes!

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
```diff
@@ -63,10 +63,21 @@ export class IndySdkToAskarMigrationUpdater {
     const {
       config: { walletConfig },
     } = agent
-    if (!walletConfig) throw new IndySdkToAskarMigrationError('Wallet config is required for updating the wallet')
+    if (typeof process?.versions?.node !== 'undefined') {
+      throw new IndySdkToAskarMigrationError('Node.js is currently not officialy supported')
```
TimoGlastra, line 67: If it's throws an error, there's no way to do it unofficial right? :)

```suggestion
      throw new IndySdkToAskarMigrationError('Node.js is currently not supported')
```

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
blu3beri: I just added this and we have tests in node.js so it would be good to keep support for Node.JS.

I could add a warning log that's it not fully working and a doc comment to the function? Or an override key that it can be used with caution?

genaris: Yeah I think it can show a warning and also it will be documented so I don't see why it should be so strictly limited.

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
genaris: Ah good to see that all records are copied (it´s not always a good idea to review a PR at midnight). So if the only missing objects in migration are the Credential Definitions I think it makes more sense to allow migration in NodeJS as not all server deployments are used to issue credentials (we have a mediator for instance, or other services that verify but not issue credentials).

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
```diff
@@ -0,0 +1,411 @@
...
+  }
+
+  /**
+   * Function that updates the values from an indy-sdk structure to the new askar stucture.
+   *
+   * - Assert that the paths that will be used are free
+   * - Create a backup of the database
+   * - Migrate the database to askar structure
+   * - Update the Keys
+   * - Update the Master Secret (Link Sercet)
```
genaris, line 228:
```suggestion
   * - Update the Master Secret (Link Secret)
```

The classic @blu3beri typo!

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
```diff
@@ -0,0 +1,411 @@
...
+
+    // Copy the file from the database path to the new location
+    await this.fs.copyFile(src, dest)
+
+    // Delete the original, only if specified by the user
+    if (this.deleteOnFinish) await this.fs.delete(this.dbPath)
+  }
+
+  /**
+   * Function that updates the values from an indy-sdk structure to the new askar stucture.
```
genaris, line 222:
```suggestion
   * Function that updates the values from an indy-sdk structure to the new askar structure.
```

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
blu3beri: Commit also closes it.

blu3beri:
@TimoGlastra @genaris I think I will do a release of the askar wrappers and after that it should be ready for a rereview.

packages/indy-sdk-to-askar-migration/package.json
```diff
@@ -0,0 +1,40 @@
...
+    "compile": "tsc -p tsconfig.build.json",
+    "prepublishOnly": "yarn run build",
+    "test": "jest"
+  },
+  "dependencies": {
+    "@aries-framework/anoncreds": "^0.4.0-alpha.71",
+    "@aries-framework/askar": "^0.4.0-alpha.71",
+    "@aries-framework/core": "^0.4.0-alpha.71",
+    "@aries-framework/indy-sdk": "^0.4.0-alpha.71",
+    "@aries-framework/node": "^0.4.0-alpha.71",
```
TimoGlastra, line 31: this should be 0.3.3 (it will be automatically updated when we release)

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
```diff
@@ -0,0 +1,416 @@
...
+    deleteOnFinish?: boolean
+    defaultLinkSecretId?: string
+  }) {
+    const {
+      config: { walletConfig },
+    } = agent
+    if (typeof process?.versions?.node !== 'undefined') {
+      agent.config.logger.warn(
+        'Node.JS is not fully supported. Using this will likely leave the wallet in a half-migrated state'
+      )
```
TimoGlastra, line 67: I'm not happy with this. A half migrated wallet is not worth a thing. We should either not support Node.JS or add support for the missing properties or throw an error when we detect that the wallet has credential definition record upfront.

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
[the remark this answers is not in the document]
TimoGlastra: Ah that's good to know! We should probably update this in the askar wallet implementation then?!?

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
```diff
@@ -0,0 +1,416 @@
...
+      // Move the migrated and updated file to the expected location for afj
+      await this.moveToNewLocation()
+    } catch (cause) {
+      this.agent.config.logger.error('Migration failed. Restoring state.')
+
+      await this.restoreDatabase()
+
+      throw new IndySdkToAskarMigrationError('Migration failed. State has been restored.', { cause })
+    } finally {
+      await this.cleanBackup()
```
TimoGlastra, line 262: will finally be called if we throw in a catch?

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
```diff
@@ -0,0 +1,416 @@
...
+      const txn = await session.open()
+      const credentials = await txn.fetchAll({ category, limit: 50 })
+      if (!credentials || credentials.length === 0) {
+        await txn.close()
+        break
+      }
+
+      for (const row of credentials) {
+        this.agent.config.logger.debug(`Migrating ${row.name} to the new askar format`)
+        const data = JSON.parse(row.value as string) as {
```
TimoGlastra, line 380: how does this type differ from `AnonCredsCredential` interface? 

I see it has rev_reg and witness keys. Are those added when stored, or are those also transfered over the wire from issuer to holder?

yarn.lock
```diff
@@ -895,13 +977,33 @@
...
+  dependencies:
+    "@hyperledger/aries-askar-shared" "0.1.0-dev.5"
+    "@mapbox/node-pre-gyp" "^1.0.10"
+    ffi-napi "^4.0.3"
+    node-cache "^5.1.2"
+    ref-array-di "^1.2.2"
+    ref-napi "^3.0.3"
+    ref-struct-di "^1.1.1"
+
 "@hyperledger/aries-askar-shared@0.1.0-dev.4", "@hyperledger/aries-askar-shared@^0.1.0-dev.4":
```
TimoGlastra, line 993: can you update all places where askar is used? We now have multilple versions installed

packages/indy-sdk-to-askar-migration/tests/migrate.test.ts
```diff
@@ -0,0 +1,56 @@
...
+    dependencies: agentDependencies,
+  })
+
+  const oldDbPath = `${homedir()}/.indy_client/wallet/${oldAgent.config.walletConfig?.id}/sqlite.db`
+
+  beforeAll(() => {
+    registerAriesAskar({ askar: ariesAskar })
+  })
+
+  test('indy-sdk sqlite to aries-askar sqlite', async () => {
```
TimoGlastra, line 40: test feel light. Would like to have some of the AFJ specific transformations tests. Will you address that in a separate PR?

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
```diff
@@ -0,0 +1,416 @@
...
+    this.deleteOnFinish = deleteOnFinish
+  }
+
+  public static async initialize({
+    dbPath,
+    agent,
+    deleteOnFinish,
+    defaultLinkSecretId,
+  }: {
+    dbPath: string
```
TimoGlastra, line 56: I think most people won't really know where the database is stored. Can we make it optional and compute if not defined based on defaults?

TimoGlastra (changes requested):

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
TimoGlastra: Not needed before merging

packages/indy-sdk-to-askar-migration/tests/migrate.test.ts
TimoGlastra: Not needed before merging

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
TimoGlastra: Not needed before merging

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
blu3beri: The only reason I added this is because we have nodejs tests. I can do a hack thing by overriding the global `process.version` check we do as well. 

I agree that this is suboptimal.

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
blu3beri: I think we can, but it can be a hassle with android, iOS, *nix and Windows.

packages/indy-sdk-to-askar-migration/tests/migrate.test.ts
blu3beri: Yeah I was thinking about this and I was not super sure what the best way to get mock wallets was.

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
TimoGlastra: Then it's something for the docs.

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
blu3beri: Yes, https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch#the_finally-block

> Immediately before a control-flow statement (return, throw, break, continue) is executed in the try block or catch block.

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
blu3beri: Overriding `process.versions.node` does not seem to work as straight forward as I thought. I could check if its a Jest environment inside the `initialise` code, but that does not seem too clean.

blu3beri: Detecting credential definition might actually be the best solution. I will add that, I assume we do not have to check other fields as well.

genaris: Yeah that's a good one. Will make everybody of us half-happy!.

packages/indy-sdk-to-askar-migration/tests/migrate.test.ts
blu3beri: I added one that checks the restoring of the wallet, but I 100% agree there need to be better tests.

packages/indy-sdk-to-askar-migration/src/IndySdkToAskarMigrationUpdater.ts
TimoGlastra: I see your interface contains the `rev_reg` and `witness` properties, but these are not in the `AnonCredsCredential`. Do you know if these properties are added when storing to the credential, or is the interface we have in AFJ incorrect (and we should add those keys to that interface)

TimoGlastra (approved):

PR #1290, opened 2023-02-13
feat: add fetch indy schema method

Vickysomtee:
This PR implements the `fetchIndySchema` method which fetches the actual schema ID for the `getCredentialDefinition` method in the` IndySdkAnonCredsRegistry` and  `IndyVdrAnonCredsRegistry`


packages/indy-sdk/src/anoncreds/services/IndySdkAnonCredsRegistry.ts
```diff
@@ -509,6 +511,47 @@ export class IndySdkAnonCredsRegistry implements AnonCredsRegistry {
...
+        version: schema.txn.data.data.version,
+        issuerId: did,
+        seqNo,
+      },
+      indyNamespace: pool.didIndyNamespace,
+    }
+  }
+}
+
+export interface SchemaType {
```
TimoGlastra, line 545: I don't think there's a need to export this interface, it's only used internally

```suggestion
interface SchemaType {
```

packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
```diff
@@ -218,20 +219,26 @@ export class IndyVdrAnonCredsRegistry implements AnonCredsRegistry {
 
       const response = await pool.submitReadRequest(request)
 
+      //FIXME: Is the ledgerType 1 ok or is it meant to be a dynamic value?
+      // if yes, what the best way to get the ledgerType
+      const schema = await this.fetchIndySchemaWithSeqNo(agentContext, response.result.ref, did, 1)
```
TimoGlastra, line 224: When fetching a schema it will always be `1`, which is the `DOMAIN` ledger.  If you look at e.g. https://indyscan.io/home/SOVRIN_MAINNET, you can see there's three subledgers.

So I think you can just hardcode the `1` inside the `fetchIndySchemaWithSeqNo` method as it will always be 1

packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
```diff
@@ -218,20 +219,26 @@ export class IndyVdrAnonCredsRegistry implements AnonCredsRegistry {
...
-            schemaId: response.result.ref.toString(),
-            tag: response.result.tag,
-            type: 'CL',
-            value: response.result.data,
-          },
-          credentialDefinitionMetadata: {
-            didIndyNamespace: pool.indyNamespace,
-          },
-          resolutionMetadata: {},
+        if (schema) {
```
TimoGlastra, line 227: I think we can write it as: `if (response.result.data && schema)` so we can remove the nesting of the if statements

packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
```diff
@@ -416,6 +423,46 @@ export class IndyVdrAnonCredsRegistry implements AnonCredsRegistry {
...
+          name: schema.data.name,
+          version: schema.data.version,
+          issuerId: did,
+          seqNo,
+        },
+        indyNamespace: pool.indyNamespace,
+      }
+    }
+
+    return agentContext.config.logger.error(`Could not get schema from ledger for seq no ${seqNo}'`)
```
TimoGlastra, line 456: In general I always write the unhappy flow first, and return early in that case. This way you can often prevent big if statement blocks, and simplify the code. This article explain the principle quite nicely I think: https://medium.com/swlh/return-early-pattern-3d18a41bba8

So in this case we'd first check if the type is _not_ equal to 101, and if that's the case we return null. Then the main happy flow of this function (the biggest code chunk) will return on the top level of the method.

TimoGlastra: Also we return a logger call here, I think that should be null:

```suggestion
    agentContext.config.logger.error(`Could not get schema from ledger for seq no ${seqNo}'`)
    return null
```

packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
```diff
@@ -416,6 +423,46 @@ export class IndyVdrAnonCredsRegistry implements AnonCredsRegistry {
...
+  }
+}
+
+export interface SchemaType {
```
TimoGlastra, line 460: Same here, no need to export

```suggestion
interface SchemaType {
```

packages/indy-sdk/tests/indy-sdk-anoncreds-registry.e2e.test.ts
```diff
@@ -0,0 +1,187 @@
...
+      indySdk,
+    }),
+  },
+})
+
+agent.dependencyManager.registerInstance(IndySdkPoolService, indySdkPoolService)
+indySdkPoolService.setPools(agentConfig.indyLedgers)
+const indySdkAnonCredsRegistry = new IndySdkAnonCredsRegistry()
+
+describe('IndySdkAnonCredsRegistry', () => {
```
TimoGlastra, line 33: Perfect!

TimoGlastra (changes requested):
Some small comments, but this is great @Vickysomtee!

packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
```diff
@@ -218,12 +219,14 @@ export class IndyVdrAnonCredsRegistry implements AnonCredsRegistry {
 
       const response = await pool.submitReadRequest(request)
 
-      if (response.result.data) {
+      const schema = await this.fetchIndySchemaWithSeqNo(agentContext, response.result.ref, did, 1)
```
TimoGlastra, line 222: no need to pass type

```suggestion
      const schema = await this.fetchIndySchemaWithSeqNo(agentContext, response.result.ref, did)
```

packages/indy-sdk/src/anoncreds/services/IndySdkAnonCredsRegistry.ts
```diff
@@ -509,6 +511,47 @@ export class IndySdkAnonCredsRegistry implements AnonCredsRegistry {
...
+
+    const { pool } = await indySdkPoolService.getPoolForDid(agentContext, did)
+    agentContext.config.logger.debug(`Getting transaction with seqNo '${seqNo}' from ledger '${pool.didIndyNamespace}'`)
+
+    const request = await indySdk.buildGetTxnRequest(did, 'DOMAIN', seqNo)
+
+    agentContext.config.logger.trace(`Submitting get transaction request to ledger '${pool.didIndyNamespace}'`)
+    const response = await indySdkPoolService.submitReadRequest(pool, request)
+
+    const schema = response.result.data as SchemaType
```
TimoGlastra, line 527: We should do the same check here as we do with the `IndyVdrAnonCredsRegistry` and check if the txn type is `101`

TimoGlastra (approved):

PR #1291, opened 2023-02-13
fix(indy-vdr): export relevant packages from root

karimStekelenburg:
Adds missing exports to the `indy-vdr` package's public API.


packages/indy-vdr/src/index.ts
```diff
@@ -6,3 +6,8 @@ try {
 } catch (error) {
   throw new Error('Error registering nodejs bindings for Indy VDR')
```
genaris, line 7: I think it will be better to remove this try-catch here and put it in Module's register() method as we did for askar and anoncreds-rs. Note that here you are checking the presence of nodejs wrapper, while in the other cases we are trying to import react-native before throwing the error.

karimStekelenburg: Agreed, not sure why this is happening here.

packages/indy-vdr/src/IndyVdrModule.ts
```diff
@@ -15,6 +15,17 @@ export class IndyVdrModule implements Module {
...
 
   public register(dependencyManager: DependencyManager) {
+    try {
+      // eslint-disable-next-line import/no-extraneous-dependencies
+      require('@hyperledger/indy-vdr-nodejs')
+    } catch (error) {
+      try {
+        require('@hyperledger/indy-vdr-react-native')
+      } catch (error) {
+        throw new Error('Error registering nodejs bindings for Indy VDR')
```
TimoGlastra, line 25:
```suggestion
        throw new Error('Error registering bindings for Indy VDR')
```

TimoGlastra (changes requested):

TimoGlastra (approved):

PR #1303, opened 2023-02-14
fix: imports from core

genaris:
These are preventing `@aries-framework/anoncreds-rs` and `@aries-framework/askar` from bein properly imported. 

Also removes the outdated `AskarModuleConfig`.


packages/askar/src/wallet/AskarWallet.ts
```diff
@@ -364,7 +364,7 @@ export class AskarWallet implements Wallet {
         const key = seed ? AskarKey.fromSeed({ seed: Buffer.from(seed), algorithm }) : AskarKey.generate(algorithm)
 
         // Store key
-        await this.session.insertKey({ key, name: encodeToBase58(key.publicBytes) })
+        await this.session.insertKey({ key, name: utils.encodeToBase58(key.publicBytes) })
```
TimoGlastra, line 367: Can't we use the TypedArrayEncoder for this?

genaris: Yeah I think so, I will use it directly.

TimoGlastra (approved):

PR #1306, opened 2023-02-16
feat!: add data, cache and temp dirs to FileSystem

genaris:
Files created by the framework have different nature, and in this PR we introduce three categories:

- **data**: intended to be persistent and included in OS backups. For instance, wallet contents (used by default for Askar-based wallets)
- **cache**: files intended to be persistent, but that can be regenerated or re-downloaded. An example of this are the tails files for AnonCreds
- **temp**: temporary files that can be safely deleted by the OS after usage. For instance, the genesis file created from genesisTransactions or wallet backup generated during migration

Another change from current behaviour is that all files created by the framework are contained within `basePath/.afj` directory. This is somewhat similar to indy-sdk, which creates a `homedir/.indy_client`. 

Some remaining questions:

- Will it be necessary to write migration scripts to move files (if existant) to new directories? Probably not, considering that until now all files created by AFJ are either temporary or cache
- Is it OK to use user home directory and Document directory as a base for data in (node and react-native respectively)? This selection and categorization is based on [iOS storage guidelines](https://developer.apple.com/documentation/foundation/optimizing_your_app_s_data_for_icloud_backup/)


packages/node/src/NodeFileSystem.ts
```diff
@@ -5,21 +5,30 @@ import { createHash } from 'crypto'
...
+   * @param baseCachePath The base path to use for reading and writing cache files used within the framework.
+   * Files will be created under baseCachePath/.afj directory. If not specified, it will be set to homedir()
+   * @param baseTempPath The base path to use for reading and writing temporary files within the framework.
+   * Files will be created under baseTempPath/.afj directory. If not specified, it will be set to tmpdir()
    */
-  public constructor(basePath?: string) {
-    this.basePath = basePath ?? tmpdir()
+  public constructor(options?: { baseDataPath?: string; baseCachePath?: string; baseTempPath?: string }) {
+    this.dataPath = `${options?.baseDataPath ?? homedir()}/.afj`
+    this.cachePath = `${options?.baseCachePath ?? homedir()}/.afj`
```
TimoGlastra, line 30: I would assume the `data` and `cache` paths couldn't override each other. But if both are in `~/.afj` they could compete with each other. should we nest it under `~/.afj/cache` etc...?

packages/core/src/storage/migration/UpdateAssistant.ts
```diff
@@ -192,7 +192,7 @@ export class UpdateAssistant<Agent extends BaseAgent<any> = BaseAgent> {
   }
 
   private getBackupPath(backupIdentifier: string) {
-    return `${this.fileSystem.basePath}/afj/migration/backup/${backupIdentifier}`
+    return `${this.fileSystem.tempPath}/migration/backup/${backupIdentifier}`
```
TimoGlastra, line 195: Do we want to use the temp path for this? If you start a migration and close the app in the process, it may be nice that we could recover it and thus have to manually delete the backup (this is done I think already afterwards)

TimoGlastra (commented):
Nice, great improvement 👍

TimoGlastra:
> Will it be necessary to write migration scripts to move files (if existant) to new directories? Probably not, considering that until now all files created by AFJ are either temporary or cache

No I don't think so. We may want to do a delete all for the old folders. But as it's all in the temp folder the env should take care of that autmoatically

> Is it OK to use user home directory and Document directory as a base for data in (node and react-native respectively)? This selection and categorization is based on [iOS storage guidelines](https://developer.apple.com/documentation/foundation/optimizing_your_app_s_data_for_icloud_backup/)

Seems fine to me

TimoGlastra:
@JamesKEbert knows a lot about mobile environments. Does this look good to you?

packages/node/src/NodeFileSystem.ts
genaris: Agreed. We can put data under `.afj/data` and cache under `.afj/cache`

packages/core/src/storage/migration/UpdateAssistant.ts
genaris: Yes, that's not a very good idea as it is unspecified when the OS will clean up the files, especially large ones.

But if we are going to use data directory I think we should automatically delete the file once we successfully finish the migration or restore the backup from it. This is something not happening right now (at least on framework level), so I can add it (and some test for it) within this PR.

genaris: Just a remark regarding something I found during this update: framework produces a backup-error file when restoring a wallet. I guess the value of doing so is to be able to do some further debugging. Anyway, in this case this file will remain in the file system.

It's not supposed to be a common case, so it's fine for me but just would like to note this if you see it could represent an issue.

packages/react-native/src/ReactNativeFileSystem.ts
```diff
@@ -4,17 +4,26 @@ import { TypedArrayEncoder, AriesFrameworkError, getDirFromFilePath, Buffer } fr
...
+   * Files will be created under baseTempPath/.afj directory. If not specified, it will be set to tmpdir()
    *
    * @see https://github.com/itinance/react-native-fs#constants
    */
-  public constructor(basePath?: string) {
-    this.basePath = basePath ?? RNFS.TemporaryDirectoryPath
+  public constructor(options?: { baseDataPath?: string; baseCachePath?: string; baseTempPath?: string }) {
+    this.dataPath = `${options?.baseDataPath ?? RNFS.DocumentDirectoryPath}/.afj`
+    this.cachePath = `${options?.baseCachePath ?? RNFS.CachesDirectoryPath}/.afj`
+    this.tempPath = `${options?.baseTempPath ?? RNFS.TemporaryDirectoryPath}/.afj`
```
JamesKEbert, line 26: On Android, the TemporaryDirectoryPath will redirect to the Caching Directory, which means you should be aware you could have conflicts between cache & temp as they end up in the same space on Android. 

> The absolute path to the temporary directory (falls back to Caching-Directory on Android)

JamesKEbert (approved):
Overall looks good to me though--good changes here 👍

packages/react-native/src/ReactNativeFileSystem.ts
genaris: Good point! Added a temp/cache suffix when platform is android

PR #1311, opened 2023-02-17
fix: thread id improvements

### LINKED ISSUE #1309: Incorrect thid in response message
When responding to a DIDComm message with both an `@id` and a `~thread.thid`, it appears that AFJ will use the `@id` as the `thid` of the response message. I believe it should prefer the `thid` that was provided.

TimoGlastra:
Fixes #1309


packages/core/src/modules/credentials/protocol/v1/V1CredentialProtocol.ts
```diff
@@ -547,7 +547,7 @@ export class V1CredentialProtocol
       // No credential record exists with thread id
       credentialRecord = new CredentialExchangeRecord({
         connectionId: connection?.id,
-        threadId: offerMessage.id,
+        threadId: offerMessage.threadId,
```
TimoGlastra, line 550: Found the issue!

genaris (approved):

andrewwhitehead:
Had to check that AgentMessage.threadId defaults to the id, and it does. Looks good!

PR #1312, opened 2023-02-17
fix: loosen base64 validation

### LINKED ISSUE #1308: Unpadded base64 data is not accepted in an attachment decorator
Notably, this can happen with the attachment on a did-exchange response message. The message is rejected due to the IsBase64 check failing.

TimoGlastra:
Fixes #1308


andrewwhitehead:
Thanks!

packages/core/src/decorators/attachment/Attachment.ts
```diff
@@ -44,7 +34,7 @@ export class AttachmentData {
    * Base64-encoded data, when representing arbitrary content inline instead of via links. Optional.
    */
   @IsOptional()
-  @IsBase64()
+  @IsString()
```
genaris, line 37: I'm OK with this change if it fixes the problem but I'm curious about the reason it does not accept unpadded base64. Is an issue with class-validator?

genaris (approved):
I'm OK with this change but curious about why is it a problem for

packages/core/src/decorators/attachment/Attachment.ts
TimoGlastra: I think it only accept plain base64, not any of the variant in between base64 and base64url?

Issue #1125, opened 2022-11-28
Remove indy-sdk from core


karimStekelenburg:
### Description
One of the things we're trying to achieve is to remove the dependency on the `indy-sdk` Rust library. Instead, the `indy-sdk` related logic will be handled by a set of new components that rely on the newer `indy-vdr`, `aries-askar`, and `anoncreds-rs` Rust libraries. These components should have been finalized in the preceding tasks.

There is a difference, however. Currently, the `core` of AFJ depends on `indy-sdk`, which is provided through the `indy` property inside the [`AgentDependencies`](https://github.com/hyperledger/aries-framework-javascript/blob/c4e96799d8390225ba5aaecced19c79ec1f12fa8/packages/core/src/agent/AgentDependencies.ts#L11) object. In order to make the framework more flexible, this will not be replaced `indy-vdr`, `aries-askar`, and `anoncreds-rs` dependencies. Instead, we will have new `packages` (or modules) that depend on these libraries, and users can provide these modules to the agent upon initialization.

Assuming the other components are in place, as a final step we need to remove the `indy-sdk` from the [`AgentDependencies`](https://github.com/hyperledger/aries-framework-javascript/blob/c4e96799d8390225ba5aaecced19c79ec1f12fa8/packages/core/src/agent/AgentDependencies.ts#L11). This will likely cause some issues that need to be addressed.

### Todos

- [ ] Remove the `indy` property from  [`AgentDependencies`](https://github.com/hyperledger/aries-framework-javascript/blob/c4e96799d8390225ba5aaecced19c79ec1f12fa8/packages/core/src/agent/AgentDependencies.ts#L11)  and see where the code breaks.
- [ ] If this results in serious issues, create new tasks for solving those.
- [ ] Solve all issues that came up.
- [ ] Remove the `indy` property from [`AgentDependencies`](https://github.com/hyperledger/aries-framework-javascript/blob/c4e96799d8390225ba5aaecced19c79ec1f12fa8/packages/core/src/agent/AgentDependencies.ts#L11) (if not already done).
- [ ] Remove the `indy` property from the [`node` implementation](https://github.com/hyperledger/aries-framework-javascript/blob/c4e96799d8390225ba5aaecced19c79ec1f12fa8/packages/node/src/index.ts#L18).
[...]

PR #1320, opened 2023-02-18
refactor: remove master secret id from wallet

TimoGlastra:
Removes the `masterSecretId` from the wallet and now requires it to be handled through the anoncreds api (as link secret).

we will need to incorporate this into the migration script for 0.4.0, this is tracked in https://github.com/hyperledger/aries-framework-javascript/discussions/1316

Dependant on #1125


tests/e2e-test.ts
```diff
@@ -50,6 +50,12 @@ export async function e2eTest({
   const [recipientSenderConnection, senderRecipientConnection] = await makeConnection(recipientAgent, senderAgent)
   expect(recipientSenderConnection).toBeConnectedWith(senderRecipientConnection)
 
+  // Create default link secret
+  await recipientAgent.modules.anoncreds.createLinkSecret({
```
genaris, line 54: I think in this high-level e2e test it could be nice to call createLinkSecret() with default parameters just to check that it is set as default and everything works in a transparent way (I guess most users will only register a single link secret and won't care about its particular id).

genaris (approved):
The repo must be merged with main but, based on changes over #1286, I think is clean and good to go!

tests/e2e-test.ts
TimoGlastra: Updated!

PR #1321, opened 2023-02-18
fix(samples): dummy module response message type

genaris:
Remove suspicious reference to 2060.io in `DummyResponseMessage`


TimoGlastra (approved):
I've noticed before, but just thought you wanted to leave your mark 😄

PR #1324, opened 2023-02-19
fix: seed and private key validation and return type in registrars

genaris:
Continuation of #1301 where there were some missing points:

- DidRegistrars secret object return type includes buffers for privateKey and seed instead of converting them to string
- Seed and private key validation logic in DidRegistrars is moved to Wallet
- Some utility functions were added in core to check validity of a given seed and private key depending on key type

Currently, all supported key types have the same private key length (32 bytes) and the minimum length of seed for all of them is also set to 32 bytes, so it is a bit silly to use a mapper when all values map to the same number but I think it's better to be explicit to easily change things if needed, especially considering that I'm not too sure about the minimum length for seeds (both bbs-signatures and askar actually allow passing very short seed values, although it's not recommended).


packages/core/src/wallet/IndyWallet.ts
```diff
@@ -22,11 +22,11 @@ import { inject, injectable } from 'tsyringe'
 
 import { AgentDependencies } from '../agent/AgentDependencies'
 import { InjectionSymbols } from '../constants'
-import { KeyType } from '../crypto'
+import { isValidPrivateKey, isValidSeed, KeyType } from '../crypto'
 import { Key } from '../crypto/Key'
 import { SigningProviderRegistry } from '../crypto/signing-provider/SigningProviderRegistry'
 import { AriesFrameworkError, IndySdkError, RecordDuplicateError, RecordNotFoundError } from '../error'
-import { Logger } from '../logger'
+import { ConsoleLogger, Logger } from '../logger'
```
TimoGlastra, line 29:
```suggestion
import { Logger } from '../logger'
```

TimoGlastra (approved):

PR #1327, opened 2023-02-20
feat: default return route

NB-MikeRichardson:
Created new PR as previous one was broken by recent changes


packages/core/src/modules/proofs/ProofsApiOptions.ts
```diff
@@ -94,6 +94,7 @@ export interface RequestProofOptions<PPs extends ProofProtocol[] = ProofProtocol
  */
 export interface AcceptProofRequestOptions<PPs extends ProofProtocol[] = ProofProtocol[]> extends BaseOptions {
   proofRecordId: string
+  hasReturnRoute?: boolean
```
TimoGlastra, line 97: hasReturnRoute gives me the idea of something that is already decided (so the name of a boolean), but the user can make the choise when calling this method.

Maybe we can call it `enableReturnRoute` or `useReturnRoute`?

packages/core/src/modules/proofs/ProofsApi.ts
```diff
@@ -495,7 +495,7 @@ export class ProofsApi<PPs extends ProofProtocol[]> implements ProofsApi<PPs> {
           serviceParams: {
             service: recipientService.resolvedDidCommService,
             senderKey: ourService.resolvedDidCommService.recipientKeys[0],
-            returnRoute: true,
+            returnRoute: options.hasReturnRoute !== false, // defaults to true if missing
```
TimoGlastra, line 498: I think this may be a bit easier to read. `??` will only take the second value if the first one is undefined (not if it's false like `||` does)

```suggestion
            returnRoute: options.hasReturnRoute ?? true
```

packages/core/src/modules/proofs/ProofsApi.ts
```diff
@@ -356,7 +356,7 @@ export class ProofsApi<PPs extends ProofProtocol[]> implements ProofsApi<PPs> {
...
```
TimoGlastra, line 359:
```suggestion
            returnRoute: options.hasReturnRoute ?? true
```

packages/anoncreds/src/protocols/proofs/v1/__tests__/v1-connectionless-proofs.e2e.test.ts
```diff
@@ -134,6 +134,7 @@ describe('V1 Proofs - Connectionless - Indy', () => {
 
     await aliceAgent.proofs.acceptRequest({
       proofRecordId: aliceProofExchangeRecord.id,
+      hasReturnRoute: returnRoute,
```
TimoGlastra, line 137: We're only passing `hasReturnRoute` when we're calling `acceptRequest`. What's the scenario for adding this when calling `acceptPresentation`?

packages/anoncreds/src/protocols/proofs/v1/__tests__/v1-connectionless-proofs.e2e.test.ts
```diff
@@ -154,6 +155,18 @@ describe('V1 Proofs - Connectionless - Indy', () => {
...
+    await connectionlessTest()
+  })
+
+  test('Faber starts with connection-less proof requests to Alice return route set to false', async () => {
+    await connectionlessTest(false)
+  })
+
+  test('Faber starts with connection-less proof requests to Alice return route set to true', async () => {
+    await connectionlessTest(false)
   })
```
TimoGlastra, line 170: We're not really testing whether the parameter is acutally used and influences the flow. It's the same test, ran three times, and it will behave the same as AFJ will already close the session. Maybe a unit tests is more in place here, or we check whether the sent message contains a `~transport` decorator with the correct value set?

packages/core/src/modules/proofs/ProofsApi.ts
TimoGlastra, line 498: Thinking of it more. What if we just disable return routing for the acceptPresentation call? It's the end of the protocol and we never expect a return. For the accept request it makes sense as we expect a presentation back. This way we don't have to add any new parameters but we set returnRoute to:
- `true` in acceptRequest (ProofsApi)
- `false` in `acceptPresentation (ProofsApi)
- `true` in acceptOffer (CredentialsApi)
- `true` in acceptRequest (CredentialsApi)
- `false` in acceptCredential (CredentialsApi)

What do you think?

packages/anoncreds/src/protocols/proofs/v1/__tests__/v1-connectionless-proofs.e2e.test.ts
NB-MikeRichardson: I deliberately only put it in acceptRequest as a quick fix for the use case causing our problem (which was when we were accepting a request and then timing out because no ack was received from the verifier).

My understanding (from James) was that there would be a fuller solution.

That said, no reason why we can't add it to acceptPresentation as well.

TimoGlastra: No reason to add it I think, was just asking because you added it: https://github.com/NB-MikeRichardson/aries-framework-javascript/blob/feat/return-route/packages/core/src/modules/proofs/ProofsApiOptions.ts#L121

packages/anoncreds/src/protocols/proofs/v1/__tests__/v1-connectionless-proofs.e2e.test.ts
NB-MikeRichardson: Added a check on the transport decortator to ensure that it has the correct value

packages/core/src/modules/proofs/ProofsApi.ts
NB-MikeRichardson: The `acceptOffer `and `acceptRequest `in credentials are already set to true so we are ok there. I have set `acceptPresentation `and `acceptCredential `to false as it is quite true, since it's the end of the protocol no need for it to be anything else.

I would like to leave `acceptRequest `in ProofsAPI with the option to pass in a value as I think for AFJ <-> AFJ we might have an ack and could need this set to true, but for aca-py <-> AFJ we would need it set to false as no ack (at least according to my colleagues who have been doing the testing).

There was a suggestion from @JamesKEbert to set the `returnRoute `flag according to the `pleaseAck `decorator value but I thought this was overkill so I didn't do it that way.

packages/core/src/modules/proofs/ProofsApiOptions.ts
```diff
@@ -117,6 +118,7 @@ export interface NegotiateProofRequestOptions<PPs extends ProofProtocol[] = Proo
  */
 export interface AcceptProofOptions {
   proofRecordId: string
+  useReturnRoute?: boolean
```
TimoGlastra, line 121: This is not used

```suggestion
```

packages/core/src/modules/proofs/ProofsApiOptions.ts
```diff
@@ -94,6 +94,7 @@ export interface RequestProofOptions<PPs extends ProofProtocol[] = ProofProtocol
  */
 export interface AcceptProofRequestOptions<PPs extends ProofProtocol[] = ProofProtocol[]> extends BaseOptions {
   proofRecordId: string
+  useReturnRoute?: boolean
```
TimoGlastra, line 97:
```suggestion

  /**
  * Whether to enable return routing on the sent presentation message. This value only has an effect for connectionlness exchanges.
  * @default true
  */
  useReturnRoute?: boolean
```

packages/core/src/modules/proofs/protocol/v2/__tests__/v2-indy-connectionless-proofs.e2e.test.ts
```diff
@@ -139,6 +144,11 @@ describe('V2 Connectionless Proofs - Indy', () => {
       state: ProofState.PresentationReceived,
     })
 
+    const didCommMessageRepository = aliceAgent.dependencyManager.resolve(DidCommMessageRepository)
+    sentPresentationMessage = await didCommMessageRepository.findAgentMessage(aliceAgent.context, {
+      associatedRecordId: aliceProofExchangeRecord.id,
+      messageClass: V2PresentationMessage,
+    })
```
TimoGlastra, line 151: You can use:

```suggestion
    sentPresentationMessage= aliceAgent.proofs.findPresentationMessage( aliceProofExchangeRecord.id)
```

packages/anoncreds/src/protocols/proofs/v1/__tests__/v1-connectionless-proofs.e2e.test.ts
TimoGlastra: I only see it added in the v2 test. Can you also add it here?

TimoGlastra:
CI is failing, can you address the validation issues?

NB-MikeRichardson:
CI Validation failure now fixed, can we merge this?

packages/core/src/modules/proofs/ProofsApi.ts
```diff
@@ -341,26 +341,23 @@ export class ProofsApi<PPs extends ProofProtocol[]> implements ProofsApi<PPs> {
...
+            returnRoute: options.useReturnRoute ?? true, // defaults to true if missing
           },
         })
       )
-
+      await this.didCommMessageRepository.saveOrUpdateAgentMessage(this.agentContext, {
+        agentMessage: message,
+        role: DidCommMessageRole.Sender,
+        associatedRecordId: proofRecord.id,
+      })
```
TimoGlastra, line 360: Why did you change the order of this? We save the message before sending in all other paths, so this can lead to unexpected behvaiour.

TimoGlastra:
> CI Validation failure now fixed, can we merge this?

Yes, one final comment for clarification

packages/core/src/modules/proofs/ProofsApi.ts
NB-MikeRichardson: The problem is that the `sendMessageToService `method appends the transport decorator to the message. If I save the proof exchange record before sending, that decorator does not get saved which seems wrong to me.

Is there any other way of doing this?

TimoGlastra: I think it's not needed for the outbound message... Otherwise, how did the test pass before this PR?

NB-MikeRichardson: The existence of the decorator is new (something I added) - we never tested for the existence of the decorator before as far as I can see

NB-MikeRichardson: Certainly what I have found is that if I put the ordering back the tests fail (return route undefined, since it can't find the decorator)

TimoGlastra: I think we should remove it for now, as this is the only place where we'd do this, and consistency is more important IMO. In a follow up PR we can change the behaviour, but we should it for all messages.

That said, I think we don't really need to store the decorator. It's nice, but we may refactor the outbound sender quite a bit (and support queues) so we won't know when the message will be actually sent

NB-MikeRichardson:
Can we merge this now?

TimoGlastra (approved):

PR #1328, opened 2023-02-20
feat: IndyVdrAnonCredsRegistry revocation methods

Vickysomtee:
This PR implements the Reveocation methods for `IndyVdrAnonCredsRegistry`

**Work funded by the Government of Ontario**


packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
```diff
@@ -390,31 +396,196 @@ export class IndyVdrAnonCredsRegistry implements AnonCredsRegistry {
     }
   }
 
+  public async getRevocationRegistryDefinition(
+    agentContext: AgentContext,
+    revocationRegistryDefinitionId: string
+  ): Promise<GetRevocationRegistryDefinitionReturn> {
+    try {
+      const indySdkPoolService = agentContext.dependencyManager.resolve(IndyVdrPoolService)
+      const did = didFromRevocationRegistryDefinitionId(revocationRegistryDefinitionId)
```
TimoGlastra, line 405: You're calling `didFromRevocationRegistryDefinitionId`. I think you can reuse the value

packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
```diff
@@ -390,31 +396,196 @@ export class IndyVdrAnonCredsRegistry implements AnonCredsRegistry {
...
+      }
+
+      const revocationRegistryDelta = {
+        value: {
+          accum: response.result.data?.value.accum_to.value.accum,
+          issued: response.result.data?.value.issued,
+          revoked: response.result.data?.value.revoked,
+        },
+        ver: '1.0',
+      }
```
TimoGlastra, line 558: Althoug this is fine, and it works, it would feel  a bit cleaner to me to only make the interface contain values that would actually be needed by the `anonCredsRevocationStatusListFromIndyVdr` method. 

E.g. I think the following value would work just as well:

```suggestion
      const revocationRegistryDelta = {
        accum: response.result.data?.value.accum_to.value.accum,
        issued: response.result.data?.value.issued,
        revoked: response.result.data?.value.revoked
      }
```

TimoGlastra (changes requested):
Awesome, great work @Vickysomtee! Left some small remarks

packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
Vickysomtee: @TimoGlastra This is not clear. What do I reuse the value for?

TimoGlastra: I meant to say you're calling `didFromRevocationRegistryDefinitionId` twice

TimoGlastra (approved):

PR #1332, opened 2023-02-21
fix(askar): anoncrypt messages unpacking

genaris:
AskarWallet unpack method was not properly handling the case of incoming anoncrypt messages, which is the case of forward messages. As a result, mediation was not working.

In this PR there are also some updates in the E2E tests to do the full flow (including mediation), interacting with an Indy SDK issuer. These tests are still skipped due to performance issues in CI/CD.


TimoGlastra (approved):

swcurran:
Was this an Askar issue, or an issue in AFJ’s use of Askar?

genaris:
> Was this an Askar issue, or an issue in AFJ’s use of Askar?

It's on the AFJ side. Both Askar and ACA-Py implementation based on it seems OK.

PR #1334, opened 2023-02-22
feat: Add cheqd-sdk module

DaevMithran:
This PR adds cheqd-sdk module in AFJ with the following interfaces

- CheqdDidRegistrar
- CheqdDidResolver
- CheqdSdkAnonCredsRegistry

Added tests for all the classes


packages/cheqd-sdk/package.json
```diff
@@ -0,0 +1,48 @@
+{
+  "name": "@aries-framework/cheqd-sdk",
+  "main": "build/index",
+  "types": "build/index",
+  "version": "0.1.0",
```
TimoGlastra, line 5: Version should be in sync with the other packages (we keep them in sync to make releases and version compatiblity easier for now)

packages/cheqd-sdk/package.json
```diff
@@ -0,0 +1,48 @@
...
+  "private": true,
```
TimoGlastra, line 6: If we want to publish the package, we should remove this line. do you want to make adiitional prs before an alpha release is created?

packages/cheqd-sdk/src/anoncreds/services/CheqdSdkAnonCredsRegistry.ts
```diff
@@ -0,0 +1,300 @@
...
+      const response = await cheqdDidRegistrar.createResource(agentContext, schema.issuerId, schemaResource)
+      if (response.resourceState.state !== 'finished') {
+        throw new Error(response.resourceState.reason)
+      }
+
+      return {
+        schemaState: {
+          state: 'finished',
+          schema,
+          schemaId: `${schema.issuerId}/resources/${schemaResource.id!}`,
```
TimoGlastra, line 100: Nit, but if we generate the id outside of the resource it can never be undefined and we don't have to use `!`

packages/cheqd-sdk/src/anoncreds/utils/transform.ts
```diff
@@ -0,0 +1,255 @@
...
+}
+
+export interface AccumKey {
+  z: string
+}
+
+// Converts JSON strings to/from your types
+// and asserts the results of JSON.parse at runtime
+export class Convert {
+  public static toCheqdSchema(json: string): CheqdSchema {
```
TimoGlastra, line 55: I see you call `json.stringify` before passing it to this method, and then calling `json.parse` here again. Maybe we can save some resource by passing in the json directly and not calling `json.parse` here?

packages/cheqd-sdk/src/anoncreds/utils/transform.ts
```diff
@@ -0,0 +1,255 @@
+// To parse this data:
```
TimoGlastra, line 1: I have little understanding of what's going on in this file. Is this needed because the resource is returned in a different structure by the cheqd sdk? When we did the integration a while back I didn't seem to need these?

Or is this for validation of the objects? in that case we have `class-transformer` and `class-validator` which would be a bit more consinstent with the rest of the framework and hides a lot of the complexity.

packages/cheqd-sdk/src/anoncreds/utils/transform.ts
```diff
@@ -0,0 +1,255 @@
...
+  CheqdRevocationRegistryDefinitionValue: o(
+    [
+      { json: 'publicKeys', js: 'publicKeys', typ: r('PublicKeys') },
+      { json: 'maxCredNum', js: 'maxCredNum', typ: 0 },
+      { json: 'tailsLocation', js: 'tailsLocation', typ: '' },
+      { json: 'tailsHash', js: 'tailsHash', typ: '' },
+    ],
+    false
+  ),
+  PublicKeys: o([{ json: 'accumKey', js: 'accumKey', typ: r('AccumKey') }], false),
```
TimoGlastra, line 253: The `utils.ts` file seems to be empty

packages/cheqd-sdk/src/dids/CheqdDidRegistrar.ts
```diff
@@ -0,0 +1,423 @@
...
+  DidRepository,
+  KeyType,
+  Key,
+  Buffer,
+  isValidPrivateKey,
+} from '@aries-framework/core'
+import { MethodSpecificIdAlgo, createDidVerificationMethod } from '@cheqd/sdk'
+import { MsgCreateResourcePayload } from '@cheqd/ts-proto/cheqd/resource/v2'
+import { fromString, toString } from 'uint8arrays'
+import { v4 } from 'uuid'
```
TimoGlastra, line 27: You can import `utils` from '@aries-framework/core` which exposes an `uuid` method. This will generate an v4 uuid

packages/cheqd-sdk/src/dids/CheqdDidRegistrar.ts
```diff
@@ -0,0 +1,423 @@
...
+        didState: {
+          state: 'failed',
+          reason: 'Provide a didDocument or atleast one verificationMethod with seed in secret',
+        },
+      }
+    }
+
+    try {
+      const payloadToSign = await MsgCreateDidDocPayloadToSign(didDocument as DIDDocument, versionId)
+      const signInputs = await this.signPayload(agentContext, payloadToSign, didDocument.verificationMethod!)
```
TimoGlastra, line 90: Even though we know the did document will have the verification method,  I think it would be good to handle the case of it not existing. If the `generateDidDoc` method changes we won't get any errors here of it not existing.

You can also make the return type of `generateDidDoc` to return `DidDocument & { verificationMethod: VerificadtionMethod[] }` so it assures the vMethod is there

packages/cheqd-sdk/src/dids/CheqdDidResolver.ts
```diff
@@ -0,0 +1,202 @@
...
+      }
+    } catch (error) {
+      return {
+        error: 'notFound',
+        message: `resolver_error: Unable to resolve resource '${did}': ${error}`,
+      }
+    }
+  }
+
+  private async ResolveAllDidDocVersions(agentContext: AgentContext, parsedDid: ParsedCheqdDid) {
```
TimoGlastra, line 143: We usually use camelCase for function names

packages/cheqd-sdk/src/dids/CheqdDidResolver.ts
```diff
@@ -0,0 +1,202 @@
...
+    await cheqdSdkLedgerService.connect({ network })
+    const { didDocumentVersionsMetadata } = await cheqdSdkLedgerService.resolveMetadata(did)
+    return {
+      didDocument: { id: did } as DidDocument,
+      didDocumentMetadata: didDocumentVersionsMetadata,
+      didResolutionMetadata: {},
+    }
+  }
+
+  private async DereferenceCollectionResources(agentContext: AgentContext, parsedDid: ParsedCheqdDid) {
```
TimoGlastra, line 156: Curious: does this mean you can resolve did-linked resources using this did resolver?

packages/cheqd-sdk/src/dids/CheqdDidResolver.ts
```diff
@@ -0,0 +1,202 @@
...
+        } else {
+          const date = params.time ? new Date(params.time) : new Date()
+          // find the resourceId to the created time
+          resources.sort(function (a, b) {
+            const distancea = Math.abs(date.getTime() - a.created!.getTime())
+            const distanceb = Math.abs(date.getTime() - b.created!.getTime())
+            return distancea - distanceb
+          })
+          resource = resources[0]
+        }
```
TimoGlastra, line 101: Maybe we can extract some of these things into util methods so it'll become easier to understand the flow of what's happening here, without needing to understand the implementation.

E.g.:

```ts
resources = filterResourcesByNameAndType(resources)

if (params.version) {
  resource = findResourceByVersion(resources)
} else {
  const date = params.time ? new Date(params.time) : new Date()
  resources = sortResourcesByDateDescending(date)

  resource = resources[0]
}
```

packages/cheqd-sdk/src/dids/CheqdDidResolver.ts
```diff
@@ -0,0 +1,202 @@
...
+      }
+
+      const mimeType = metadata.mediaType
+      let result: any
+      if (mimeType == 'application/json') {
+        result = await JsonEncoder.fromBuffer(resource.data)
+      } else if (mimeType == 'text/plain') {
+        result = toString(resource.data)
+      } else {
+        result = toString(resource.data, 'base64')
```
TimoGlastra, line 127: same here, I think we could extract this to a util like (not too happy with the name yet) `resourceDataFromMimeType(resource.data, mimeType)`

packages/cheqd-sdk/src/dids/didCheqdUtil.ts
```diff
@@ -0,0 +1,110 @@
...
+
+import {
+  createDidPayload,
+  createDidVerificationMethod,
+  createVerificationKeys,
+  DIDModule,
+  VerificationMethods,
+} from '@cheqd/sdk'
+import { MsgCreateDidDocPayload, MsgDeactivateDidDocPayload } from '@cheqd/ts-proto/cheqd/did/v2'
+import { bases } from 'multiformats/basics'
```
TimoGlastra, line 12: The `multiformats` package has caused issues in react native in the past due to it's dependency on `TextEncoder`. Do we need this?  We have a base58 encoder implementation availalbe in AFJ (see TypedArrayEncoder). We also have a way to parse multibase keys in the `Key` class (`Key.fromFingerprint()`)

packages/cheqd-sdk/src/dids/didCheqdUtil.ts
```diff
@@ -0,0 +1,110 @@
...
+        return Array.isArray(s?.serviceEndpoint) && s?.id && s?.type
+      })
+    : true
+
+  if (!isValidService) return { valid: false, error: 'Service is Invalid' }
+  return { valid: true } as SpecValidationResult
+}
+
+// Create helpers in sdk like MsgCreateDidDocPayload.fromDIDDocument to replace the below
+export async function MsgCreateDidDocPayloadToSign(didPayload: DIDDocument, versionId: string) {
```
TimoGlastra, line 55: we usually use camelCame methods. Maybe this can be called `createMsgCreateDidDocPayloadToSign` or `getMsgCreateDidDocPayloadToSign`?

packages/cheqd-sdk/src/index.ts
```diff
@@ -0,0 +1,17 @@
+// Dids
+export {
+  CheqdDidRegistrar,
```
TimoGlastra, line 3: Some are called `CheqdXXX` while others are called `CheqdSdkXXX`. I'm fine with `CheqdXX`, but we should make them consistent I think

packages/cheqd-sdk/src/ledger/CheqdSdkLedgerService.ts
```diff
@@ -0,0 +1,107 @@
+import type { AbstractCheqdSDKModule, CheqdSDK, ICheqdSDKOptions, DidStdFee, DIDDocument } from '@cheqd/sdk'
+import type { SignInfo } from '@cheqd/ts-proto/cheqd/did/v2'
+import type { MsgCreateResourcePayload } from '@cheqd/ts-proto/cheqd/resource/v2'
+
+import { injectable } from '@aries-framework/core'
+import { createCheqdSDK, DIDModule, ResourceModule, CheqdNetwork } from '@cheqd/sdk'
+import { DirectSecp256k1HdWallet, DirectSecp256k1Wallet } from '@cosmjs/proto-signing'
+import { EnglishMnemonic as _ } from '@cosmjs/crypto'
+import { fromString } from 'uint8arrays'
```
TimoGlastra, line 9: We have a `TypedArrayEncoder`  that generally handles transformign from/to UInt8arrays and buffers. Can we leverage that you think?

packages/cheqd-sdk/src/ledger/CheqdSdkLedgerService.ts
```diff
@@ -0,0 +1,107 @@
...
+        ? DefaultRPCUrl.Mainnet
+        : DefaultRPCUrl.Testnet,
+      wallet: await this.cosmosPayerWallet.catch(() => {
+        throw new Error(`[did-provider-cheqd]: valid cosmosPayerSeed is required`)
+      }),
+    }
+
+    this.sdk = await createCheqdSDK(sdkOptions)
+    this.fee = options.fee
+  }
```
TimoGlastra, line 63: Generally the approach we've taken with the indy pool service is that we create pool implemetnation that handle the connection to a specific network. This way you can connect to multiple ledgers simultaneously. E.g. the resolver now calls connect, but in the meantime another resolve action could also call connect and we'd be opening the gate for race conditions. 

Do you think we coudl add that layer where we'd be manaing multiple networks? In the indy implementation we then get the netwrok based on the indyNamespace, which in cheqd's case would be `mainnet` vs `testnet`.

packages/cheqd-sdk/src/ledger/CheqdSdkLedgerService.ts
```diff
@@ -0,0 +1,107 @@
...
+    const { cosmosPayerSeed, rpcUrl } = config
+    if (!cosmosPayerSeed || cosmosPayerSeed === '') {
+      this.cosmosPayerWallet = DirectSecp256k1HdWallet.generate()
+      return
+    }
+    this.cosmosPayerWallet = EnglishMnemonic._mnemonicMatcher.test(cosmosPayerSeed)
+      ? DirectSecp256k1HdWallet.fromMnemonic(cosmosPayerSeed, { prefix: 'cheqd' })
+      : DirectSecp256k1Wallet.fromKey(fromString(cosmosPayerSeed.replace(/^0x/, ''), 'hex'), 'cheqd')
+
+    this.rpcUrl = rpcUrl
```
TimoGlastra, line 45: How feasible would it be to integrate this with the wallet from AFJ? Generally we try to avoid doing crypto not using the wallet as it makes it harder to integrate with e.g. a KMS.

Askar supports secp256k keys I believe, but otherwise we can also provide a `SigningProvider`. This way people could use more performant hardware based implemetnations when they want to.

We could then just send the bytes that should be signed to the wallet if possible?

packages/cheqd-sdk/src/ledger/CheqdSdkLedgerService.ts
```diff
@@ -0,0 +1,107 @@
...
+  Testnet = 'https://rpc.cheqd.network',
+}
+
+export class EnglishMnemonic extends _ {
+  public static readonly _mnemonicMatcher = /^[a-z]+( [a-z]+)*$/
+}
+
+@injectable()
+export class CheqdSdkLedgerService {
+  private readonly cosmosPayerWallet: Promise<DirectSecp256k1HdWallet | DirectSecp256k1Wallet>
```
TimoGlastra, line 30: Would it be possible to dynamically set the payer wallet for each transaction? We've just removed the global `publicDidSeed` from the agent for indy-sdk and now require you to provide a `submitterDid` when writing to the indy ledger, just so it's more flexible in what the options are. 

Maybe you control multiple dids, or for example in multi-tenancy mode this could switch based on the agent context.

For indy we require you to add the key for the submitter to the wallet first, and then reference that in the input parameters, after which the wallet signs the request.

packages/cheqd-sdk/tests/cheqd-did-registrar.e2e.test.ts
```diff
@@ -0,0 +1,49 @@
...
+      secret: {
+        verificationMethod: {
+          id: 'key-1',
+          type: VerificationMethods.Ed255192018,
+          privateKey,
+        },
+      },
+      options: {
+        network: CheqdNetwork.Testnet,
+        methodSpecificIdAlgo: MethodSpecificIdAlgo.Uuid,
```
TimoGlastra, line 44: I'm thinking whether we can hide the usage of the `@cheqd/sdk`, and whether that's a desired feature or not?

I think most of these values are just strings, and requiring the user to add the `@cheqd/sdk` package just for that seems unnessary. 

WDYT?

packages/cheqd-sdk/tests/cheqd-did-registrar.e2e.test.ts
```diff
@@ -0,0 +1,49 @@
...
+      },
+    })
+    expect(did.didState).toMatchObject({ state: 'finished' })
```
TimoGlastra, line 47: Maybe we can check the full return type and see if it returns the did document as expected?

packages/cheqd-sdk/tests/setup.ts
```diff
@@ -0,0 +1 @@
+jest.setTimeout(250000)
```
TimoGlastra, line 1: This is more than 4 minutes. Do we need this long?

TimoGlastra (changes requested):
Left an initial review. Really like the addition, this is great, but have some questions/concers about integrating with existing AFJ infrastructure and making sure it works in multitenancy/paralel processing cases

packages/cheqd-sdk/src/anoncreds/utils/transform.ts
DaevMithran: Yes this is just for json schema validation, great will have a look at them

packages/cheqd-sdk/src/dids/CheqdDidResolver.ts
DaevMithran: Yep. added a separate function resolveResource in the same class which can used to resolve the actual resource. A separate function was needed as the return type is different
Generally while resolving a didDocument the list of resource metadatas are added the didDocument metadata as well.

packages/cheqd-sdk/src/ledger/CheqdSdkLedgerService.ts
DaevMithran: Not sure how this is possible, it will lead to the race conditions again you mentioned in this [comment](https://github.com/hyperledger/aries-framework-javascript/pull/1334#discussion_r1114200987)

A cosmos account is not binded to a DID/wallet like publicDIDSeed of indy-sdk

If it's just to manage different identity wallets owned by the same entity, there is one cosmos account, which is used just to pay the network fees in all the wallets. There can be two separate account's mainnet/testnet.

If it's entirely a different entity, then it's better for them to use a different agent rather than using multitenancy. As the mnemonic/seed of the cosmos account is exposed in the config unless we have a config for each multitenant wallet.

TimoGlastra: thanks for the info. I'm less familiar with how the cheqd sdk works. But couldn't we use one sdk instantiation per network. They then each get a `CheqdPool` or just manage them in the CheqdSdkLedgerService, and that doesn't have a payer key and we sign the transactions with the wallet?

Ideally we sign everything with the wallet, as that makes it more aligned with how we manage keys in AFJ (they never leave the wallet).

If our wallet would support Secpk256K would that be possible?

DaevMithran: yea I'm taking this up. It's better to have one instantiation.

Will discuss with the team on your second point, I agree it's better to manage the keys within AFJ.

packages/cheqd-sdk/src/ledger/CheqdSdkLedgerService.ts
DaevMithran: There are differences in all these outputs. The first output is not compatible with our sdk

TypedArrayEncoder.toBase64(key.publicKey)
ZZ8zBCapsnUeEXMfD+xpDNeMQQ49asVY069I466pmnQ= 

TypedArrayEncoder.toBase64URL(key.publicKey)
ZZ8zBCapsnUeEXMfD-xpDNeMQQ49asVY069I466pmnQ

toString(key.publicKey, 'base64')
ZZ8zBCapsnUeEXMfD+xpDNeMQQ49asVY069I466pmnQ

TimoGlastra: I think the `TypedArrayEncoder` supports both Buffer and Uint8Arrays, and this shouldn't happen. There's probably a bug in either of the implementation then? Seems most of it is the same. I'll look into the `toString` implementation. 

Do you know if `uint8array` works in react native?

packages/cheqd-sdk/src/dids/didCheqdUtil.ts
DaevMithran: This worked

packages/cheqd-sdk/src/ledger/CheqdSdkLedgerService.ts
DaevMithran: Sorry It's a bug in our sdk. Fixed it

packages/cheqd-sdk/tests/cheqd-did-registrar.e2e.test.ts
DaevMithran: Yea can be done

packages/cheqd-sdk/tests/cheqd-did-registrar.e2e.test.ts
DaevMithran: The uuid generated is going to be random, so can't really predict the outcome here

packages/cheqd-sdk/src/dids/CheqdDidRegistrar.ts
```diff
@@ -0,0 +1,425 @@
...
+        let key: Key
+        if (method.publicKeyBase58) {
+          key = Key.fromPublicKeyBase58(method.publicKeyBase58, KeyType.Ed25519)
+        } else if (method.publicKeyJwk) {
+          key = Key.fromJwk(method.publicKeyJwk as unknown as Jwk)
+        } else if (method.publicKeyMultibase) {
+          key = Key.fromFingerprint(method.publicKeyMultibase)
+        } else {
+          throw new Error(`Invalid verificationMethod`)
+        }
```
TimoGlastra, line 367: we have a `getKeyFromVerificationMethod` method (you need to update with latest main). I think we could let this be handled by this method. Then it will also automatically support multiple key types.

TimoGlastra (changes requested):
Some final changes. Have you tested these changes in React Native? We had to apply a lot of workarounds to get it to work, and we should probalby add documentation for this.

packages/cheqd-sdk/src/dids/CheqdDidRegistrar.ts
```diff
@@ -0,0 +1,425 @@
...
+  options: {
+    network: CheqdNetwork
+    rpcUrl?: string
+    fee?: DidStdFee
+    versionId?: string
+    methodSpecificIdAlgo?: MethodSpecificIdAlgo
+  }
+  secret: {
+    verificationMethod?: IVerificationMethod
+  }
```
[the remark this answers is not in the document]
DaevMithran: It is possible, don't think services are passed as a part of secret according to the spec.
Current way to add service is to update the didDocument by appending a service endpoint to the entire didDocument after creation.

Could add that feature if that's what aries is following.

packages/core/src/modules/dids/domain/DidDocumentRole.ts
```diff
@@ -1,4 +1,5 @@
 export enum DidDocumentRole {
   Created = 'created',
   Received = 'received',
+  Deactivated = 'deactivated',
```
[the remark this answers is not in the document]
DaevMithran: Ah got it, removing this

packages/cheqd-sdk/src/dids/CheqdDidRegistrar.ts
```diff
@@ -0,0 +1,425 @@
...
+          key = Key.fromJwk(method.publicKeyJwk as unknown as Jwk)
+        } else if (method.publicKeyMultibase) {
+          key = Key.fromFingerprint(method.publicKeyMultibase)
+        } else {
+          throw new Error(`Invalid verificationMethod`)
+        }
+        return {
+          verificationMethodId: method.id,
+          signature: await agentContext.wallet.sign({ data: Buffer.from(payload), key }),
+        } as SignInfo
```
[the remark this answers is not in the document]
DaevMithran: The payload needs to be signed by all the verification methods that belong to the controllers, there can be keys which do not belong to the controller.

There could be cases where 
1. controller is not defined in either create or update operations
2. key rotation requires all the signatures

So in afj I'm signing with all the available keys in the wallet, the cheqd ledger will extract only the required signatures to process the request

packages/cheqd-sdk/src/anoncreds/services/CheqdAnonCredsRegistry.ts
```diff
@@ -0,0 +1,335 @@
...
+      const cheqdDidRegistrar = agentContext.dependencyManager.resolve(CheqdDidRegistrar)
+
+      const schema = options.schema
+      const schemaResource = {
+        id: utils.uuid(),
+        name: schema.name,
+        resourceType: 'anonCredsSchema',
+        data: {
+          name: schema.name,
+          version: schema.version,
```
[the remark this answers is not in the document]
DaevMithran: Yep, we want everything expect issuerId as the part of resource data, 
It is recommended to have the same resource name and version, but not necessary. As the schema is queried using just the resourceId, name and version don't matter here

packages/cheqd/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "clean": "rimraf ./build",
+    "compile": "tsc -p tsconfig.build.json",
+    "prepublishOnly": "yarn run build",
+    "test": "jest"
+  },
+  "dependencies": {
+    "@aries-framework/anoncreds": "0.3.3",
+    "@aries-framework/core": "0.3.3",
+    "@cheqd/sdk": "cjs",
+    "@cheqd/ts-proto": "^2.0.1",
```
ankurdotb, line 30: This requires a new release on ts-proto with a similar CommonJS release as SDK above

packages/cheqd/src/anoncreds/utils/identifiers.ts
```diff
@@ -0,0 +1,50 @@
+import type { ParsedDid } from '@aries-framework/core'
+
+const UUID = '([a-z,0-9,-]{36,36})'
```
anikitinDSR, line 3: I think, we can add some static checks for DID or DID URL and raise an error first.
- For UUID validation let's use some specific packages or maybe regexp from UUID RFC cause current UUID regexp is not making the validation.
- If you want to get `method` or `network` or other parts of DID I think it worse to make just split(":") and handle each parts one by one.
- ID in cheqd DID it's not just string from 21 or 22 bytes. Firstly it should be base58 string and the next rule - it should 16 byte after encoding string to array of bytes. 
- About `fragment` or `query` I think it worse to use some `url` parser package and get those values from it or maybe also use official regexps from corresponded RFC.
- The main reason why it would be better - regexps are bad for support in future and for debugging from my point of view.

anikitinDSR: Also, small thing here. Fragment is not placed in URL as `#` it should be encoded to particular code, like `%23`. It's not being sent to the server directly. So, if you want to get the fragment you need to search for `%23` symbol.

packages/cheqd/src/dids/CheqdDidResolver.ts
```diff
@@ -0,0 +1,189 @@
...
+  parseCheqdDid,
+} from '../anoncreds/utils/identifiers'
+import { CheqdLedgerService } from '../ledger'
+
+import { filterResourcesByNameAndType, getClosestResourceVersion, renderResourceData } from './didCheqdUtil'
+
+export class CheqdDidResolver implements DidResolver {
+  public readonly supportedMethods = ['cheqd']
+
+  public async resolve(agentContext: AgentContext, did: string, parsed: ParsedDid): Promise<DidResolutionResult> {
```
anikitinDSR, line 23: We are gonna to support a lot of queries for resolving DID and resources BTW. Should we maybe create a base for future adds? I mean in case of supporting more then 10 queries it will be really big switch, no?

packages/cheqd/src/dids/didCheqdUtil.ts
```diff
@@ -0,0 +1,152 @@
...
+  createDidVerificationMethod,
+  createVerificationKeys,
+  DIDModule,
+  VerificationMethods,
+} from '@cheqd/sdk'
+import { MsgCreateDidDocPayload, MsgDeactivateDidDocPayload } from '@cheqd/ts-proto/cheqd/did/v2'
+import { EnglishMnemonic as _ } from '@cosmjs/crypto'
+import { DirectSecp256k1HdWallet, DirectSecp256k1Wallet } from '@cosmjs/proto-signing'
+
+export function validateSpecCompliantPayload(didDocument: DidDocument): SpecValidationResult {
```
anikitinDSR, line 17: Is it supposed to be the same static validation process as on ledger side? If yes. then I think we need to add here a static validation for publicKey for each verification method and not only check for empty.
BTW, do we have the validation on `cheqd/sdk`? I think if it's needed it should on sdk side, not here.

packages/cheqd/src/dids/didCheqdUtil.ts
```diff
@@ -0,0 +1,152 @@
...
+
+export interface IDidDocOptions {
+  verificationMethod: VerificationMethods
+  verificationMethodId: TVerificationKey<string, number>
+  methodSpecificIdAlgo: MethodSpecificIdAlgo
+  network: CheqdNetwork
+  publicKey: string
+}
+
+export function getClosestResourceVersion(resources: Metadata[], date: Date) {
```
anikitinDSR, line 117: Let's add some comments here that this methods should be changed to particular queries when they will be ready on did-resolver side?

packages/cheqd/src/ledger/CheqdLedgerService.ts
```diff
@@ -0,0 +1,123 @@
...
+  public async deactivate(
+    didPayload: DIDDocument,
+    signInputs: SignInfo[],
+    versionId?: string | undefined,
+    fee?: DidStdFee
+  ) {
+    return await this.getSdk(didPayload.id).deactivateDidDocTx(signInputs, didPayload, '', fee, undefined, versionId)
+  }
+
+  public async resolve(did: string, version?: string) {
```
anikitinDSR, line 95: Is It LedgerService part? Maybe let's create CheqdDidResolverService?

anikitinDSR (commented):
Great Job! A lot of stuff are done, let's maybe chat about possible changes if you want.

packages/cheqd/src/ledger/CheqdLedgerService.ts
DaevMithran: Its querying from the ledger using the sdk, don't need a specfic service just for this.

packages/cheqd/src/dids/didCheqdUtil.ts
DaevMithran: No it's not ledger validation that's happening here, it validates a did doc according to the did-core spec. 

Ledger validations are handled in @cheqd/sdk

packages/cheqd/src/dids/CheqdDidResolver.ts
DaevMithran: For the use case of anoncreds, only these four queries are relevant and used, Mainly to query a resource using name, type and the resourceTime. So not sure if we need to complicate it here

TimoGlastra:
@DaevMithran Is the PR ready to be merged, or do you still need to incoroporate any of the outstanding comments? 

Also, it seems DCO is failing.

packages/cheqd/src/anoncreds/utils/identifiers.ts
DaevMithran: Resolved, replaced regex validations with proper library validations

packages/cheqd-sdk/src/ledger/CheqdSdkLedgerService.ts
DaevMithran: Will be taking this up in the next release

TimoGlastra:
I think we can add caching in a follow-up PR

DaevMithran:
@TimoGlastra  The tests failing are not specific to this PR, its the same one's failing in main as well

TimoGlastra (approved):

PR #1335, opened 2023-02-22
feat: outbound message send via session

morrieinmaas:
This adds required functionality for using AFJ for BLE (Bluetooth Low Energy).

Why:
* The verifier has created and sent a proof request  to a holder OOB using create legacy connection invite. 
* Holder receive the invite OOB by using receive message (eg with BLE transport)
* In cases like BLE the verifier has no outbound transport and wants to eg send an ack message for a proof exchange. This fails currently because of the missing outbound - Verifier cannot reach the holder.

How it is addressed:
* In the proof exchange dance for example (works generally). when receiving an encrypted message we store a session and the id from the inbound message context/session. 
* That way we can later retrieve the connection related to the exchange and send the ack message to the session instead of using outbound transport.
* Further a method was added to reset the outbound transport. It stops the current outbound transports and replaces the array of outbound transports with an empty array. The outbound transports attribute of the agent was changed from read-only to read-write.


packages/core/src/agent/MessageSender.ts
```diff
@@ -408,13 +419,61 @@ export class MessageSender {
...
-      } else if (transport.supportedSchemes.includes(protocolScheme)) {
-        await transport.sendMessage(outboundPackage)
-        return
+
+    // Outbound transport is present
+    if (this.outboundTransports.length !== 0) {
+      for (const transport of this.outboundTransports) {
+        const protocolScheme = getProtocolScheme(service.serviceEndpoint)
+        if (!protocolScheme) {
+          this.logger.warn('Service does not have valid protocolScheme.')
```
blu3beri, line 428:
```suggestion
          this.logger.warn('Service does not have a protocol scheme.')
```

packages/core/src/agent/MessageSender.ts
```diff
@@ -408,13 +419,61 @@ export class MessageSender {
...
+        )
+        try {
+          await this.sendMessageToService(outboundMessageContext)
+          this.emitMessageSentEvent(outboundMessageContext, OutboundMessageSendStatus.SentToSession)
+          return
+        } catch (error) {
+          this.logger.debug(`Sending an outbound message via session failed with error: ${error.message}.`, error)
+          throw new MessageSendingError(
+            `Unable to send message to service: ${service.serviceEndpoint} ${JSON.stringify(
+              message.threadId
```
blu3beri, line 454: Why is the threadId stringified?

packages/core/src/agent/models/OutboundMessageContext.ts
```diff
@@ -26,15 +28,17 @@ export interface OutboundMessageContextParams {
...
   public serviceParams?: ServiceMessageParams
   public outOfBand?: OutOfBandRecord
   public associatedRecord?: BaseRecord<any, any, any>
   public sessionId?: string
   public readonly agentContext: AgentContext
 
-  public constructor(message: T, context: OutboundMessageContextParams) {
+  public constructor(message: T, context: OutboundMessageContextParams, inboundContext?: InboundMessageContext) {
     this.message = message
+    this.sessionIdFromInbound = inboundContext?.sessionId ? inboundContext.sessionId : undefined
```
blu3beri, line 41:
```suggestion
    this.sessionIdFromInbound = inboundContext?.sessionId
```

packages/core/src/agent/models/OutboundMessageContext.ts
```diff
@@ -26,15 +28,17 @@ export interface OutboundMessageContextParams {
 
 export class OutboundMessageContext<T extends AgentMessage = AgentMessage> {
   public message: T
+  public sessionIdFromInbound?: string | undefined
```
blu3beri, line 31:
```suggestion
  public sessionIdFromInbound?: string
```

packages/core/src/agent/MessageSender.ts
```diff
@@ -408,13 +419,61 @@ export class MessageSender {
...
+        try {
+          await this.sendMessageToSession(agentContext, session, message)
+          this.emitMessageSentEvent(outboundMessageContext, OutboundMessageSendStatus.SentToSession)
+          return
+        } catch (error) {
+          this.logger.debug(`Sending an outbound message via session failed with error: ${error.message}.`, error)
+          throw new MessageSendingError(
+            `Unable to send message to service: ${
+              outboundMessageContext.serviceParams?.service.serviceEndpoint
+            } with threadId ${JSON.stringify(message.threadId)} `,
```
blu3beri, line 471: Same here about stringifying a threadId

packages/core/src/agent/MessageSender.ts
```diff
@@ -408,13 +419,61 @@ export class MessageSender {
...
+            }
+          )
+        }
+      } else if (outboundMessageContext.serviceParams?.service.serviceEndpoint && session) {
+        try {
+          await this.sendMessageToSession(agentContext, session, message)
+          this.emitMessageSentEvent(outboundMessageContext, OutboundMessageSendStatus.SentToSession)
+          return
+        } catch (error) {
+          this.logger.debug(`Sending an outbound message via session failed with error: ${error.message}.`, error)
```
blu3beri, line 467: I am not sure what the exact pattern is in AFJ here, but is it common that we log a `debug` message when we also throw an error? I would say either no log or an error log.

blu3beri: Also, we already pass the `error` object to the log function, so why is the `error.message` also added to the `message`?

packages/core/src/modules/proofs/protocol/v1/__tests__/v1-indy-connectionless-proofs.e2e.test.ts
```diff
@@ -198,6 +201,82 @@ describe('Present Proof', () => {
     await faberProofExchangeRecordPromise
   })
 
+  test('Faber starts with connection-less proof requests to Alice with auto-accept enabled and without outbound transport', async () => {
```
blu3beri, line 204:
```suggestion
  test('Faber starts with connection-less proof requests to Alice with auto-accept enabled and without an outbound transport', async () => {
```

blu3beri (dismissed):
I am still don't fully understand the transport structure in AFJ, so I mainly commented on the code itself.

morrieinmaas:
thx for all the feedback @blu3beri 🙏. I'll have a go at it today (as well as the merge conflicts)

packages/core/src/agent/MessageSender.ts
morrieinmaas: no good reason. i'll change it

packages/core/src/agent/models/OutboundMessageContext.ts
morrieinmaas: 😅 - obviously. thanks

packages/core/src/agent/models/OutboundMessageContext.ts
blu3beri: The question mark handles undefined. It's just marked twice now.

packages/core/src/agent/MessageSender.ts
```diff
@@ -203,8 +212,10 @@ export class MessageSender {
 
     let session: TransportSession | undefined
 
-    if (sessionId) {
-      session = this.transportService.findSessionById(sessionId)
+    if (sessionIdFromInbound) {
+      session = this.transportService.findSessionById(sessionIdFromInbound)
+    } else if (sessionId) {
+      this.transportService.findSessionById(sessionId)
```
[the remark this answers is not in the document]
morrieinmaas: ah, my bad. This should assining it to session. Actually, funny that this works. maybe I can remove using the sessionId one altogether then

packages/core/src/agent/MessageSender.ts
```diff
@@ -68,6 +68,15 @@ export class MessageSender {
     this.outboundTransports.push(outboundTransport)
   }
 
+  public async resetOutboundTransport() {
+    // NOTE: we could also make this void instead of an async fct.
```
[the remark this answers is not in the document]
morrieinmaas: sure. good idea

packages/core/src/agent/MessageSender.ts
morrieinmaas: I remember now why this is. I was logging/throwing the entire object when debugging. Regardless, removing now

packages/core/src/agent/MessageReceiver.ts
```diff
@@ -85,7 +85,7 @@ export class MessageReceiver {
       if (this.isEncryptedMessage(inboundMessage)) {
         await this.receiveEncryptedMessage(agentContext, inboundMessage as EncryptedMessage, session)
       } else if (this.isPlaintextMessage(inboundMessage)) {
-        await this.receivePlaintextMessage(agentContext, inboundMessage, connection)
+        await this.receivePlaintextMessage(agentContext, inboundMessage, connection, session)
```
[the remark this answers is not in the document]
morrieinmaas: hmm good idea, Timo might know. I'm not sure about this tbh.

morrieinmaas:
@TimoGlastra @blu3beri happy for more feedback?

packages/core/src/agent/MessageSender.ts
```diff
@@ -408,13 +419,59 @@ export class MessageSender {
...
     outboundPackage.endpoint = service.serviceEndpoint
     outboundPackage.connectionId = connection?.id
-    for (const transport of this.outboundTransports) {
-      const protocolScheme = getProtocolScheme(service.serviceEndpoint)
-      if (!protocolScheme) {
-        this.logger.warn('Service does not have valid protocolScheme.')
-      } else if (transport.supportedSchemes.includes(protocolScheme)) {
-        await transport.sendMessage(outboundPackage)
-        return
+
```
TimoGlastra, line 422: There's some duplicate flows now, so I think we may have to clean the `sendMessage` method up a bit. 

In the sendMessage method we do the following:
- check if there's a session to be found and if the message has return routing, if so return the message over the session
- if not returned over the session, loop through the services and call `this.sendToService` for each of the services in the did document.
- if the message could not be delivered, and the did document has a queue endpoint, we queu the message


The in the `sendToService` method we do the following:
- Check if there's outbound transports, and if so try to use the outbound transports
- If there's no outbound transport check if there's a session and try to send it over the sessionn.


There's a few improvements to this flow I think we can make. Especially since the message sender is already  one of the most complex classes.
- If you follow the flow above, `sendMessage` now has logic for sending to a session, but then the `sendToService` message as well. This means we now have two flows for sending a message to a session on a single path (sendMessage calls sendToService as well)
- The way sendToService handles session vs transports is that it will prefer outbound transports if they're available and otherwise fall back to the session sending. If we look at the flow for sendMessage above, it's the opposite: we first try to send it to the session, and then we start using the transports. I think we should keep the flows as similar as possible, and reduce duplciated code/flows as much as possible.

My suggestion would be to:
- Remove the session related code you added to `sendToService`. This way the `sendToService` keeps focused on one task: sending a message to a service, not dealing with session and inbound transports.
- Create a new private method `findSessionForOutboundContext` that will receive the outbound context, it will look for a session (based on the sessionId, the inboundSessionId and the connection in that order, based on what's availalbe). Then if a session has been found and if the message has return routing,  we return the session. This means if this method returns something we're certain we can use it to send a response.
- Update the `sendMessage` method to use this new `findSessionForOutboundContext` method, and if it returns a session, send the message to a session (this is alraedy implemented in `sendMessageToSession` and can be called). This logic is mostly already present in the `sendMessage` method and we can extract that into a new method.
- Extend the publcic `sendMessageToService` method to also include the `findSessionForOutboundContext` and `sendMessageToSession`. If no session is avaialble, or sending to the session failed we call `sendToService`.

We may want to rename the public `sendMessageToService` at some point to indicate we also support sending to a session (or combine it with sendMessage as the input api is 99% the same), but for now I think the above changes are enough.

packages/core/src/agent/MessageSender.ts
```diff
@@ -408,13 +419,59 @@ export class MessageSender {
...
+          return
+        } catch (error) {
+          this.logger.error(`Sending an outbound message via session failed with error: `, error)
+          throw new MessageSendingError(
+            `Unable to send message to service: ${outboundMessageContext.serviceParams?.service.serviceEndpoint} with threadId ${message.threadId} `,
+            {
+              outboundMessageContext,
+            }
+          )
+        }
```
TimoGlastra, line 474: We shouldn't send it over the session if a serviceEndpoint was present. It's only allowed to use the session if return routing is enabled, and this is something that should be fixed from the other side (so in this case the presentation MUST have return route).

We should add it to the presentation handlers (for both v1 and v2):
https://github.com/morrieinmaas/aries-framework-javascript/blob/feat/outbound-message-send-via-session/packages/core/src/modules/proofs/protocol/v2/handlers/V2PresentationHandler.ts#L56-L62

If you see the difference with the proofs Api, we explicitly pass that returnRoute should be used: https://github.com/morrieinmaas/aries-framework-javascript/blob/feat/outbound-message-send-via-session/packages/core/src/modules/proofs/ProofsApi.ts#L498

TimoGlastra (changes requested):
This is great addition @morrieinmaas and aligns the connectionless flow a lot more with the connection flow. However I think we're not 100% there yet. I've left some suggestions on how we may be able to improve the flow a bit

morrieinmaas:
@TimoGlastra thanks for the feedback and insights. I'll have a go at it asap

packages/core/src/agent/MessageReceiver.ts
```diff
@@ -135,7 +137,7 @@ export class MessageReceiver {
     // We want to save a session if there is a chance of returning outbound message via inbound transport.
     // That can happen when inbound message has `return_route` set to `all` or `thread`.
     // If `return_route` defines just `thread`, we decide later whether to use session according to outbound message `threadId`.
-    if (senderKey && recipientKey && message.hasAnyReturnRoute() && session) {
+    if (senderKey && recipientKey && session && (message.hasAnyReturnRoute() || message.service?.serviceEndpoint)) {
```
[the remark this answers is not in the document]
morrieinmaas: strangely yes removing breaks the flow. I'll do some digging.

packages/core/src/modules/proofs/protocol/BaseProofProtocol.ts
```diff
@@ -1,3 +1,13 @@
+import type { AgentMessage } from '../../../agent/AgentMessage'
```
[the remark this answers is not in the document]
morrieinmaas: it's my format on save I think

packages/core/src/agent/Agent.ts
```diff
@@ -141,6 +141,10 @@ export class Agent<AgentModules extends AgentModulesInput = any> extends BaseAge
     this.messageSender.registerOutboundTransport(outboundTransport)
   }
 
+  public async resetOutboundTransport() {
+    await this.messageSender.resetOutboundTransport()
+  }
```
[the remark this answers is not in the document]
morrieinmaas: hmm good point. Maybe that method could then also remove all as a special case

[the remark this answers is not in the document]
TimoGlastra: I'd keep it simple. The need for unregistering it special and needing to remove them one by one is fine in this case (otherwise the naming of the method will be harder again)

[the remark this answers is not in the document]
morrieinmaas: fair enough

packages/core/src/agent/MessageReceiver.ts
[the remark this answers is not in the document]
TimoGlastra: See my last comment. I thinks it'll be solved by that

rapaktech:
@TimoGlastra side note, we still have the issue with the Proofs API automatically setting up mediation for the presentation message, as shown here: https://github.com/hyperledger/aries-framework-javascript/blob/main/packages/core/src/modules/proofs/ProofsApi.ts#L379-L384. How best do we solve this?

packages/anoncreds/src/protocols/proofs/v1/__tests__/v1-indy-proof-negotiation.test.ts
```diff
@@ -0,0 +1,293 @@
...
+
+  afterAll(async () => {
+    testLogger.test('Shutting down both agents')
+    await faberAgent.shutdown()
+    await faberAgent.wallet.delete()
+    await aliceAgent.shutdown()
+    await aliceAgent.wallet.delete()
+  })
+
+  test(`Proof negotiation between Alice and Faber`, async () => {
```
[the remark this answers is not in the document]
morrieinmaas: nope. weird. is this from main? lemme see

packages/core/src/agent/models/OutboundMessageContext.ts
```diff
@@ -26,15 +28,17 @@ export interface OutboundMessageContextParams {
 
 export class OutboundMessageContext<T extends AgentMessage = AgentMessage> {
   public message: T
+  public sessionIdFromInbound?: string
```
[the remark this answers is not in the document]
morrieinmaas: shmort

packages/core/src/agent/MessageReceiver.ts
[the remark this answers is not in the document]
morrieinmaas: could you specify which one you're referring to? Can't follow.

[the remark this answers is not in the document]
TimoGlastra: https://github.com/hyperledger/aries-framework-javascript/pull/1335#discussion_r1115668827

morrieinmaas:
Thanks @TimoGlastra you're a legend. 🙏

morrieinmaas:
@TimoGlastra can this be merged?

TimoGlastra (approved):

TimoGlastra:
> @TimoGlastra can this be merged? 

Yes! We're having some issues with CI so it's been hard to merge PRs

Issue #1322, opened 2023-02-19
w3cCredentialService throws "Invalid key type" for valid key?


simonwh:
Hi!

I've been testing the OpenId4VC client using the Mattr Launchpad credentials as demoed by @karimStekelenburg. The flow works great for me, until the validation of the credential happens (just before save).

The error is: `VerificationError: Invalid key type. Key type must be "Ed25519VerificationKey2018` and it seems to be originating from here:

https://github.com/hyperledger/aries-framework-javascript/blob/64a5da937059d25e693e2491af329548b2975ef6/packages/core/src/modules/vc/signature-suites/ed25519/Ed25519Signature2018.ts#L213

Where `hasEd2018` is `true`, but `jsonld.hasValue(verificationMethod, 'type', 'Ed25519VerificationKey2018')` is false.

If I log the value of `verificationMethod`, it is:

`{"@context": ["https://w3.org/ns/did/v1", "https://w3id.org/security/suites/ed25519-2018/v1"], "id": "did:web:launchpad.vii.electron.mattrlabs.io#6BhFMCGTJg"}`

If I understand correctly, [looking up the DID](https://dev.uniresolver.io/#did:web:launchpad.vii.electron.mattrlabs.io), the `verificationMethod` of the VC is referencing the `assertMethod` in the DID, which maps to the following public key, which is indeed `Ed25519VerificationKey2018`.

```json
"publicKey": [
    {
      "id": "did:web:launchpad.vii.electron.mattrlabs.io#6BhFMCGTJg",
      "type": "Ed25519VerificationKey2018",
      "controller": "did:web:launchpad.vii.electron.mattrlabs.io",
      "publicKeyBase58": "6BhFMCGTJg9DnpXZe7zbiTrtuwion5FVV6Z2NUpwDMVT"
    }
  ]
```

I extracted [the VC from the response](https://gist.github.com/simonwh/48dd198357a05c8ff5ee781c32615b31) to make it easier to replicate.

What am I missing here? :-)

**Appreciate your help!**

TimoGlastra:
I think it may be because the mattr did document is using publicKey rather than using verificationMethod.

In the tests we have a modified version of that did document. Not sure if this is the case, but I know @karimStekelenburg also didn't get it to work with the mattr did document.

What we could do is update the web did resolver to transform the document to the current spec rules, but ideally the did document itself is updated

simonwh:
[...]

PR #1336, opened 2023-02-22
fix(openid): remove unnecessary scope check

karimStekelenburg:
The OpenID4VC-client module was throwing errors due to the absence of the optional scope property in some access_token responses, leading to valid tokens being treated as invalid.

Related to #1322 


blu3beri (approved):
So how do we deal with getting the scope now? It might be somewhere else defined but we should check if its supplied, use that one, otherwise fallback to the other one, right?

Issue #1025, opened 2022-09-10
'Socket is closed' when agent has no endpoint and uses WS


genaris:
After the fix #946, I noticed a problem when an agent does not have an endpoint (e.g. mobile agent not bound to a mediator yet) and uses WebSocket as outbound transport. 
 
As expected, the socket is being closed when DID Exchange is completed, but if another message (e.g. request a credential or request mediation) is immediately sent to the newly created connection, message sender will fail with an 'Socket is closed'  error. It seems like the socket is still opened when it checks if a session is opened, but it is effectively closed when the transporter calls ws.send() method.

If I wait some time before sending the next message (1 second or so) it works fine consistently, so that makes me think that it's just a synchronization issue in the WebSocket session handling.

PR #1337, opened 2023-02-23
fix: create new socket if socket state is 'closing'

niall-shaw:
Found issue where sending message fails with `Socket is not open.` when only using websocket transports (no http transports).

WS is used to create initial connection, and then starts before mediation request, causing socket state to update to 'CLOSING'. Since this is done before the `close` event is emitted, the relevant socket map value  of `WsOutboundTransport` hasn't been deleted yet, so when we try to send the next message too quickly this error case can happen.

### DIFF
--- a/packages/core/src/transport/WsOutboundTransport.ts
+++ b/packages/core/src/transport/WsOutboundTransport.ts
@@ -75,7 +75,7 @@ export class WsOutboundTransport implements OutboundTransport {
     // If we already have a socket connection use it
     let socket = this.transportTable.get(socketId)
 
-    if (!socket) {
+    if (!socket || socket.readyState === this.WebSocketClass.CLOSING) {
       if (!endpoint) {
         throw new AriesFrameworkError(`Missing endpoint. I don't know how and where to send the message.`)
       }


TimoGlastra (approved):

genaris:
I think this is related to the problem found in #1025

niall-shaw:
Any ideas why the test failed?

swcurran:
@jleach — could this be related to the “first credential delay” issue?  

Does this manifest in delays or in errors?

niall-shaw:
> Does this manifest in delays or in errors?

@swcurran in the use cases I was experimenting with, it always resulted in an error.

genaris (approved):
This is a very nice and elegant fix for this issue. 

I think in a near future we have to re-think a bit the way we handle WebSockets in inbound and outbound transporters, as IMHO one of the key purposes of them is to have avoid the need of re-opening and closing HTTP connections when there are several exchanges within a short time frame. So, even if a message does not have a return_route set, I think the socket should not be immediately closed but rather kept opened for some seconds (configurable if needed), so it can be re-used in case more messages are going to be sent.

niall-shaw:
@genaris @TimoGlastra - do you know why the tests are now failing? It doesn't make any sense to me why this change would cause the tests to fail

TimoGlastra:
The issues are not related to your PR. @genaris just made a PR to fix the CI issues, let's see if they pass now

PR #1346, opened 2023-02-25
fix(indy-sdk): import from core

genaris:
A relative import from core was preventing indy-sdk from being built correctly.

Here we add an eslint plugin to check these relative imports and other stuff that can be useful now we have plenty of packages in this repo.


TimoGlastra (approved):
Nice, i was hoping there would be a plugin for this.

This will help a lot with preventing accidentally breaking the released packages

PR #1348, opened 2023-02-25
feat(oob): implicit invitations

genaris:
Add support for Out-of-Band protocol 'implicit invitations', meaning by that to connect to a public DID given it has a DID Communication service in its DidDocument. 

The API consists on the method `oob.receiveImplicitInvitation`, similar to regular `oob.receiveInvitation` but it of course receiving a resolvable did instead of a Connection Invitation.

It can use both Connection and DID Exchange protocols and it's possible to create multiple connections from the same public DID. By querying `connections.findByInvitationDid(publicDid)` it's possible for an application to know if it's already connected to a public entity.

Using AFJ agents it's working, but didn't test it against ACA-Py yet. Hopefully with these changes we can pass some DID Exchange tests in the AATH.


packages/core/src/modules/oob/OutOfBandApi.ts
```diff
@@ -76,6 +76,13 @@ export interface ReceiveOutOfBandInvitationConfig {
   reuseConnection?: boolean
   routing?: Routing
   acceptInvitationTimeoutMs?: number
+  isImplicit?: boolean
```
TimoGlastra, line 79: Do you think we could split up the `receiveInvitation` method into a public and private one. The private one will allow for an `isImplicit`, but we don't add it to the public api. I don't think it's needed and you can already call the other method.

packages/core/src/modules/oob/OutOfBandApi.ts
```diff
@@ -430,12 +465,13 @@ export class OutOfBandApi {
...
-    const services = outOfBandInvitation.getServices()
-    const messages = outOfBandInvitation.getRequests()
+    const services = outOfBandInvitation?.getServices()
+    const messages = outOfBandInvitation?.getRequests()
     const timeoutMs = config.timeoutMs ?? 20000
 
-    const existingConnection = await this.findExistingConnection(outOfBandInvitation)
+    const { handshakeProtocols } = outOfBandInvitation
+
+    const existingConnection = outOfBandInvitation ? await this.findExistingConnection(outOfBandInvitation) : undefined
```
TimoGlastra, line 474: In which case is outOfBandInvitation undefined? It looks as if you an destructure it above,

packages/core/src/modules/oob/repository/OutOfBandRecord.ts
```diff
@@ -44,6 +45,7 @@ export class OutOfBandRecord extends BaseRecord<DefaultOutOfBandRecordTags, Cust
   public autoAcceptConnection?: boolean
   public mediatorId?: string
   public reuseConnectionId?: string
+  public isImplicitInvitation?: boolean
```
TimoGlastra, line 48: Curious whether we'd need to store this as a separate property, or whether we may be able to derive it based on the implicit out of band invitation? Should we make the out of band invitation optoinal? Or will that break a lot of existing code?

packages/core/src/modules/connections/handlers/DidExchangeRequestHandler.ts
```diff
@@ -39,16 +44,39 @@ export class DidExchangeRequestHandler implements MessageHandler {
...
+        reusable: true,
+        autoAcceptConnection: this.connectionsModuleConfig.autoAcceptConnections,
+        isImplicitInvitation: true,
+        outOfBandInvitation,
+      })
+
+      await this.outOfBandService.save(messageContext.agentContext, outOfBandRecord)
+      this.outOfBandService.emitStateChangedEvent(messageContext.agentContext, outOfBandRecord, null)
+      return outOfBandRecord
+    }
```
TimoGlastra, line 73: Creating oob records is more the domain of the oob module. Maybe we can add this to the out of band service and reuse it for both the connection and did exchange protocols? I think that will make it easier if we need to change the oob implementation

packages/core/src/modules/connections/handlers/ConnectionRequestHandler.ts
```diff
@@ -32,16 +37,54 @@ export class ConnectionRequestHandler implements MessageHandler {
...
+
+      const outOfBandRecord = new OutOfBandRecord({
+        role: OutOfBandRole.Sender,
+        state: OutOfBandState.AwaitResponse,
+        reusable: true,
+        autoAcceptConnection: this.connectionsModuleConfig.autoAcceptConnections,
+        isImplicitInvitation: true,
+        outOfBandInvitation,
+        tags: {
+          recipientKeyFingerprints: [recipientKey.fingerprint],
```
TimoGlastra, line 75: Why do we store the recipientKeyFingerprints here, but not in the did exchange request handler?

packages/core/src/modules/oob/__tests__/implicit.e2e.test.ts
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    faberAgent = new Agent(faberAgentOptions)
+    faberAgent.registerInboundTransport(new SubjectInboundTransport(faberMessages))
+    faberAgent.registerOutboundTransport(new SubjectOutboundTransport(subjectMap))
+    await faberAgent.initialize()
+
+    aliceAgent = new Agent(aliceAgentOptions)
+    aliceAgent.registerInboundTransport(new SubjectInboundTransport(aliceMessages))
+    aliceAgent.registerOutboundTransport(new SubjectOutboundTransport(subjectMap))
```
TimoGlastra, line 48: I think we can use the new `setupSubjectTransports` for this

packages/core/src/modules/oob/__tests__/implicit.e2e.test.ts
```diff
@@ -0,0 +1,215 @@
...
+    await faberAgent.wallet.delete()
+    await aliceAgent.shutdown()
+    await aliceAgent.wallet.delete()
+  })
+
+  afterEach(async () => {
+    const credentials = await aliceAgent.credentials.getAll()
+    for (const credential of credentials) {
+      await aliceAgent.credentials.deleteById(credential.id)
+    }
```
TimoGlastra, line 63: Is this needed?

packages/core/src/modules/connections/handlers/ConnectionRequestHandler.ts
```diff
@@ -32,16 +37,54 @@ export class ConnectionRequestHandler implements MessageHandler {
...
+        label: '',
+        services: [did],
+        handshakeProtocols: [HandshakeProtocol.Connections],
+      })
+
+      const outOfBandRecord = new OutOfBandRecord({
+        role: OutOfBandRole.Sender,
+        state: OutOfBandState.AwaitResponse,
+        reusable: true,
+        autoAcceptConnection: this.connectionsModuleConfig.autoAcceptConnections,
```
TimoGlastra, line 71: I see we take over autoAcceptConnections from the agent config. One thing we should be careful of here is that before you always had to take a manual step for people to be able to connect to you (receiveInvitation, or createInvitation). With implicit invitations it means that if someone gets hold of my did they'll be able to create connections with me.

I'm thinking whether we should add a new `autoAcceptImplicitInvitations` property that determines whether you auto accept these invitations. What do you think?

TimoGlastra (commented):
This is really great @genaris!!!

I have some code in a local branch somewhere that has most of the work done for creating invitations using public dids, if we can add that soon as well we fully support public dids!

One thing I'm curious about, when someone connects to our public did, will we always rotate to a new did? Or will it use the keys from the invitation to create a peer did (in that case we should be careful as it'll use the same key but a diferent did). If i remember correctly a key will be created when an invitation is created, and then later a peer did will be created from that key.  Not 100% if this is working as intended, but as it's so little code I'm curious whether it jus works (I know creating invitations with public dids required quite some chagnes...)

TimoGlastra:
With the new `import` method in the dids module it should now be possible to host a mediator, create a did:web document (creating keys with AFJ, but publishing yourself outside of afj), importing the did document, and then you could use the `did:web` as a mediatorInvitationUrl.

packages/core/src/modules/connections/handlers/ConnectionRequestHandler.ts
TimoGlastra: Maybe we should even allow to create an out of band record for a specific did that specifies if the did can be connected to. But then we need creating oob invitations for public dids of course.

Just thinking how we can make it easy to connect, but also don't open the gates that allows every did in our wallet to be connected to (even peer:dids ?!?!?)

packages/core/src/modules/oob/OutOfBandApi.ts
genaris: At the beginning, I was going to make outOfBandInvitation optional, but later on I switched to this more conservative approach to not break so many things... and forgot to revert this line.

packages/core/src/modules/connections/handlers/DidExchangeRequestHandler.ts
genaris: Agree, I'll move this logic to oob and call it from the handlers.

packages/core/src/modules/connections/handlers/ConnectionRequestHandler.ts
genaris: Very good point regarding adding `autoAcceptImplicitInvitations`. 

I think we can add a flag somewhere in `DidCreateOptions` (and dids.import I think) to define whether this is a public did applicable to implicit invitations. Something like `isPublicDid` or `supportsImplicitInvitations` (always complicated to get a good naming...).

packages/core/src/modules/oob/repository/OutOfBandRecord.ts
genaris: As stated in another comment, at first I wanted to make it optional and use its absence to determine if OOB Record belongs to an implicit invitation.

But besides the breaking code (which is not a lot), I found it useful for each OOB Record to have an unique invitation, especially to store certain parameters like the threadId for the flow (used for instance in `DidExchangeCompleteHandler`). 

We can of course do the same by adding the `threadId` and the `publicDid` as optional properties of OOB Record. This would probably feel more 'natural' and less 'hacky'.

genaris:
> With the new `import` method in the dids module it should now be possible to host a mediator, create a did:web document (creating keys with AFJ, but publishing yourself outside of afj), importing the did document, and then you could use the `did:web` as a mediatorInvitationUrl.

This is exactly one of the main goals I want to achieve with this and the import! Using did:web will simplify a lot the deployments and discovery of public agents.

genaris:
> This is really great @genaris!!!
> 
> I have some code in a local branch somewhere that has most of the work done for creating invitations using public dids, if we can add that soon as well we fully support public dids!

Thanks for the fast but quite in-depth review! It was actually meant to be marked as 'draft' but fortunately I put it for review and you gave some good pointers about the approach to take.
Thanks for the fast but quite in-depth review! It was actually meant to be marked as 'draft' but fortunately I put it for review and you gave some good pointers about the approach to take.
 
> One thing I'm curious about, when someone connects to our public did, will we always rotate to a new did? Or will it use the keys from the invitation to create a peer did (in that case we should be careful as it'll use the same key but a diferent did). If i remember correctly a key will be created when an invitation is created, and then later a peer did will be created from that key. Not 100% if this is working as intended, but as it's so little code I'm curious whether it jus works (I know creating invitations with public dids required quite some chagnes...)

I'm not completely sure if I understood your question but the idea is that for both DID Exchange Request and Connection Request a new "reusable" Out of Band record will be created, so it will rotate the keys as it would be managing the requests for a reusable OOB Invitation (in the code, it will call `getRouting()` and will effectively create a new Ed25519 key pair).  Of course all DID Exchange and Connection Responses will be signed by using the same key (the one published in the public DidDoc) before rotation, which could maybe lead to some security issues (replay attacks or things like that) but I guess it's the way it was designed.

TimoGlastra:
> I'm not completely sure if I understood your question but the idea is that for both DID Exchange Request and Connection Request a new "reusable" Out of Band record will be created, so it will rotate the keys as it would be managing the requests for a reusable OOB Invitation (in the code, it will call `getRouting()` and will effectively create a new Ed25519 key pair). Of course all DID Exchange and Connection Responses will be signed by using the same key (the one published in the public DidDoc) before rotation, which could maybe lead to some security issues (replay attacks or things like that) but I guess it's the way it was designed.

This answers my question. I was mostly curious whether we were always rotating to a new did/key in the response message, but if it works the same as reusable out of band invitations then I understand. 👍

packages/core/src/modules/connections/handlers/ConnectionRequestHandler.ts
TimoGlastra: Yes that makes sense 👍

We may be able to handle this is in the did resolver service so methods don't have to explicitly support it.

We should probably call it something related to didcomm though. Maybe `autoAcceptImplicitDidCommInvitations`, but that's getting too long... Or maybe we should never autoAccept for now, and require the user to handle the request (so there's at least still one manual step). Then later if we understand a bit better what the common use cases are we can add a config flag to automate it?

genaris: I agree... let's make it manual acceptance for now and then we can automate it if we find it useful. The good thing about this manual approach is that when receiving a request, the event handler can tell what DID this relates to and it can decide based on that if it should accept it or simply ignore it.

packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -119,9 +119,11 @@ export class ConnectionService {
       imageUrl: imageUrl ?? agentContext.config.connectionImageUrl,
     })
 
+    const parentThreadId = outOfBandRecord.isImplicitInvitation ? invitationDid : outOfBandInvitation.id
```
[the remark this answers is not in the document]
genaris: The best thing about that is that it will sweep that ugly `isImplicitInvitation` property under the carpet :stuck_out_tongue: 

But take a look at the new approach, where we have this convenient getter but also removes `isImplicitInvitation` and instead adds the threadId tag to properly identify the OobRecord (only in those cases where implicit invitations are being used). I think it's more compact an clean.

packages/core/src/modules/connections/handlers/ConnectionRequestHandler.ts
```diff
@@ -32,16 +34,23 @@ export class ConnectionRequestHandler implements MessageHandler {
...
 
-    const outOfBandRecord = await this.outOfBandService.findCreatedByRecipientKey(
-      messageContext.agentContext,
-      recipientKey
-    )
+    const parentThreadId = message.thread?.parentThreadId
+
+    const outOfBandRecord =
+      parentThreadId && tryParseDid(parentThreadId)
+        ? await this.outOfBandService.createImplicitInvitation(agentContext, {
```
TimoGlastra, line 47: I'm not 100% on the naming of this method. We've just received a request based on an implicit invitation (the service in the did document is the implicit invitation). It feels like `createFromImplicitInvitation` would be more suitable? Also not 100% happy with that, but it should denote that we're creating an out of band record based on a request that was sent based on an implicit invitation.

packages/core/src/modules/oob/OutOfBandApi.ts
```diff
@@ -76,6 +76,15 @@ export interface ReceiveOutOfBandInvitationConfig {
   reuseConnection?: boolean
   routing?: Routing
   acceptInvitationTimeoutMs?: number
+  isImplicit?: boolean
+}
+
+export type ReceiveOutOfBandInvitationConfig = Omit<BaseReceiveOutOfBandInvitationConfig, 'isImplicit'>
+
+export interface ReceiveOutOfBandImplicitInvitationConfig
+  extends Omit<BaseReceiveOutOfBandInvitationConfig, 'isImplicit' | 'reuseConnection'> {
```
TimoGlastra, line 85: Why don't we allow `reuseConnection`? If I'm already connected I may or may not reuse the connection right?

packages/core/src/modules/oob/OutOfBandService.ts
```diff
@@ -28,6 +38,43 @@ export class OutOfBandService {
...
 
+  public async createImplicitInvitation(
+    agentContext: AgentContext,
+    config: CreateImplicitInvitationConfig
+  ): Promise<OutOfBandRecord> {
+    const { did, threadId, handshakeProtocols, autoAcceptConnection, recipientKey } = config
+
+    // Verify it is a valid did and it is present in the wallet
+    const publicDid = parseDid(did)
+    if (!(await agentContext.dependencyManager.resolve(DidsApi).getCreatedDids({ did: publicDid.did }))) {
```
TimoGlastra, line 49: getCreatedDids will always return an array I think?

```suggestion
    const didsApi = agentContext.dependencyManager.resolve(DidsApi)
    const [createdDid] = await didsApi.getCreatedDids({ did: publicDid.did })
    if (!createdDid) {
```

packages/core/src/modules/oob/OutOfBandService.ts
```diff
@@ -28,6 +38,43 @@ export class OutOfBandService {
...
+      throw new AriesFrameworkError(`Referenced public did ${did} not found.`)
+    }
+
+    const outOfBandInvitation = new OutOfBandInvitation({
+      id: did,
```
TimoGlastra, line 54: Is the did the id so I can't make multiple?

packages/core/src/modules/oob/OutOfBandService.ts
```diff
@@ -28,6 +38,43 @@ export class OutOfBandService {
     this.eventEmitter = eventEmitter
   }
 
+  public async createImplicitInvitation(
```
TimoGlastra, line 41: What's implicit if I explicilty create an out of band invitaiton? 🤔

packages/core/src/modules/oob/OutOfBandApi.ts
```diff
@@ -430,11 +481,12 @@ export class OutOfBandApi {
 
     const { outOfBandInvitation } = outOfBandRecord
     const { label, alias, imageUrl, autoAcceptConnection, reuseConnection, routing } = config
-    const { handshakeProtocols } = outOfBandInvitation
-    const services = outOfBandInvitation.getServices()
-    const messages = outOfBandInvitation.getRequests()
+    const services = outOfBandInvitation?.getServices()
+    const messages = outOfBandInvitation?.getRequests()
```
TimoGlastra, line 485: outOfBandInvitation is still required right?

packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -121,7 +121,7 @@ export class ConnectionService {
 
     connectionRequest.setThread({
       threadId: connectionRequest.threadId,
-      parentThreadId: outOfBandInvitation.id,
+      parentThreadId: outOfBandRecord.outOfBandInvitation.id,
```
TimoGlastra, line 124: Ah we make it the `id` so the parentThreadId is always correct. Smart

packages/core/src/modules/connections/handlers/DidExchangeCompleteHandler.ts
```diff
@@ -32,12 +33,14 @@ export class DidExchangeCompleteHandler implements MessageHandler {
...
-    if (!message.thread?.parentThreadId) {
+    const parentThreadId = message.thread?.parentThreadId
+    if (!parentThreadId) {
       throw new AriesFrameworkError(`Message does not contain pthid attribute`)
     }
     const outOfBandRecord = await this.outOfBandService.findByCreatedInvitationId(
       messageContext.agentContext,
-      message.thread?.parentThreadId
+      parentThreadId,
+      tryParseDid(parentThreadId) ? message.threadId : undefined
```
TimoGlastra, line 43: Why do we now need to store the threadId? I'm trying to wrap my head around

TimoGlastra (approved):
I'm okay with merging, but left some additional comments/questions

packages/core/src/modules/connections/handlers/ConnectionRequestHandler.ts
genaris: Well, at least I removed that `isImplicitInvitation` that didn't convince you neither 😆. The naming was somehow borrowed from the OutOfBandApi method called `createInvitation` that returned an `OutOfBandRecord`. But you are completely right: `createFromImplicitInvitation` is a bit more accurate. Could be also `createFromConnectionRequest` or `createFromImplicitInvitationResponse`? I still prefer your suggestion.

packages/core/src/modules/oob/OutOfBandApi.ts
genaris: I set this before moving to the idea of using `id = did `for all OOBInvitations related to the same did. I think it can be re-added, but it can add some complexity: suppose you have a connection related to a public did and call this method with `reuseConnection = false`, meaning that it will create a new connection related to the same public did. As a result, we will have two connections related to the same did. If you call this method again with `reuseConnection = true`, which of those two connections should it reuse?

I think it´s valuable to have this feature enabled in an automatic way (otherwise it's the consumer app who should check if there is an existing connection before calling this method). Maybe we can support it and, if two or more connections are found, we throw an error?

packages/core/src/modules/connections/handlers/DidExchangeCompleteHandler.ts
genaris: This is to relate a DidExchange Complete message to the flow where the OOBRecord comes from. Mainly because pthid will be the public did instead of an unique identifer.

packages/core/src/modules/connections/services/ConnectionService.ts
genaris: I thought this approach would be more appropriate because we would not be 'inventing` an OOB Invitation, but rather using a template based on the public DID. If someone looks at the OOBRecord, they will know that it comes from an implicit invitation because its OOB Invitation will have a did as id.

packages/core/src/modules/oob/OutOfBandService.ts
genaris: I guess this is answered in other comments. It has some drawbacks but I think it's a bit more clear than previous approach.

packages/core/src/modules/oob/OutOfBandApi.ts
genaris: The problem I see here is that `reuseConnection` parameter is related to the reuse messages as specified in [RFC 434](https://github.com/hyperledger/aries-rfcs/blob/main/features/0434-outofband/README.md#reuse-messages) (`handshake-reuse` and `handshake-reuse-accepted`). This one uses OOB Invitation `pthid` as the way of identifying the invitee... but in an implicit invitation, the pthid will be the public did of the inviter :thinking: .

This makes me think that this flow is not possible or intended to work with implicit invitations. Any clarification will be certainly welcome (my brains are as fried as yours nowadays :stuck_out_tongue:)

TimoGlastra: Yes makes sense. I think we may also be able to just return an existing connection based on `reuseConnection`. if you pass `reuseConnection=true` it will just look if we already have a connection with this did, but if you pass false it will always create a new connection. With an OOB invitation you MUST use the handshake-reuse. But this approach is perfect for now

TimoGlastra (approved):

TimoGlastra:
LGTM

PR #1349, opened 2023-02-27
feat: configure indy-vdr optional peer dependencies

Vickysomtee:


packages/indy-vdr/src/IndyVdrModuleConfig.ts
```diff
@@ -1,6 +1,45 @@
...
+import type { IndyVdr } from '@hyperledger/indy-vdr-shared'
 
 export interface IndyVdrModuleConfigOptions {
+
+    /**
+   *
+   * ## Node.JS
+   *
+   * ```ts
+   * import * as indyVdr from '@hyperledger/indy-vdr-nodejs';
```
TimoGlastra, line 11: I think we can import `indyVdr` like this:

```
import { indyVdr } from '@hyperledger/indy-vdr-nodejs';

File: packages/indy-vdr/src/IndyVdrModuleConfig.ts [hunk 2]
```diff
@@ -1,6 +1,45 @@
...
+   *  modules: {
+   *   indyVdr: new IndyVdrModule({
+   *      indyVdr,
+   *   }) 
+   *  }
+   * })     
+   * ```
+   * ```
+   */
+    indyVdr: IndyVdr
```
TimoGlastra, line 41: I think you need to update all places that create an instance of the module (`new IndyVdrModule`) to provide this parameter

Vickysomtee: @TimoGlastra the indyVdr arguement will come from the @hyperledger/indy-vdr-shared or @hyperledger/indy-vdr-nodejs?

packages/indy-vdr/src/IndyVdrModuleConfig.ts
```diff
@@ -1,6 +1,44 @@
...
+   *  dependencies: agentDependencies,
+   *  modules: {
+   *   indyVdr: new IndyVdrModule({
+   *      indyVdr,
+   *   })
+   *  }
+   * })
+   * ```
+   *
+   * ## React Nativeß
```
genaris, line 23:
```suggestion
   * ## React Native
```

packages/indy-vdr/src/IndyVdrModuleConfig.ts
```diff
@@ -1,6 +1,44 @@
...
+   *
+   * ```ts
+   * import { indyVdr } from '@hyperledger/indy-vdr-nodejs';
```
genaris, line 26:
```suggestion
   * import { indyVdr } from '@hyperledger/indy-vdr-react-native';
```

packages/indy-vdr/src/IndyVdrModuleConfig.ts
```diff
@@ -1,6 +1,44 @@
...
+   * const agent = new Agent({
+   *  config: {},
+   *  dependencies: agentDependencies,
+   *  modules: {
+   *   indyVdr: new IndyVdrModule({
+   *      indyVdr,
+   *   })
+   *  }
+   * })
+   * ```
```
genaris, line 37:
```suggestion
   *
```

genaris (commented):
Just a few typos in the IndyVdrModuleConfig comments.

packages/indy-vdr/tests/helpers.ts
```diff
@@ -2,13 +2,14 @@ import type { IndyVdrPoolService } from '../src/pool/IndyVdrPoolService'
 import type { AgentContext, Key } from '@aries-framework/core'
 
 import { KeyType } from '@aries-framework/core'
-import { AttribRequest, NymRequest } from '@hyperledger/indy-vdr-shared'
+import { AttribRequest, NymRequest, indyVdr } from '@hyperledger/indy-vdr-shared'
```
TimoGlastra, line 5: This should also import `indyVdr` from `indy-vdr-nodejs` instead of `shared`

TimoGlastra: General strategy on where to import from:
- non-test files should always be platform agnostic (so `-shared`).
- test files need a specific implementation, and we're running in Node.JS (so `-nodejs`).
- We only need to have documentation for the `-react-native` dependency in the module config

TimoGlastra (approved):

demo/package.json
```diff
@@ -13,6 +13,10 @@
     "faber": "ts-node src/FaberInquirer.ts",
     "refresh": "rm -rf ./node_modules ./yarn.lock && yarn"
   },
+  "dependencies": {
+    "@hyperledger/indy-vdr-shared": "^0.1.0-dev.6",
+    "@hyperledger/anoncreds-shared": "^0.1.0-dev.6"
```
TimoGlastra, line 18: This should depend on the nodejs packages

demo/src/BaseAgent.ts
```diff
@@ -25,6 +25,8 @@ import {
 import { IndySdkAnonCredsRegistry, IndySdkModule, IndySdkSovDidResolver } from '@aries-framework/indy-sdk'
 import { IndyVdrAnonCredsRegistry, IndyVdrModule, IndyVdrSovDidResolver } from '@aries-framework/indy-vdr'
 import { agentDependencies, HttpInboundTransport } from '@aries-framework/node'
+import { anoncreds } from '@hyperledger/anoncreds-shared'
+import { indyVdr } from '@hyperledger/indy-vdr-shared'
```
TimoGlastra, line 29: this should also be the nodejs packages. Have you tried to run the demo and made sure it worked?

packages/anoncreds/src/AnonCredsModuleConfig.ts
```diff
@@ -1,10 +1,47 @@
 import type { AnonCredsRegistry } from './services'
+import type { Anoncreds } from '@hyperledger/anoncreds-shared'
 
 /**
  * @public
  * AnonCredsModuleConfigOptions defines the interface for the options of the AnonCredsModuleConfig class.
  */
 export interface AnonCredsModuleConfigOptions {
+  /**
```
TimoGlastra, line 9: The anoncreds dependency should actually be added to the AnonCredsRs module (I can understand this is confusing). The `@aries-framework/anoncreds` (so no `-rs`) can be used with multiple anoncreds implementations (either `anoncreds-rs` or `indy-sdk`)

packages/askar/src/AskarModule.ts
```diff
@@ -1,33 +1,18 @@
...
+  }
+  public register(dependencyManager: DependencyManager) {
+    dependencyManager.registerInstance(AskarModuleConfig, this.config)
 
-    if (dependencyManager.isRegistered(InjectionSymbols.StorageService)) {
-      throw new AriesFrameworkError('There is an instance of StorageService already registered')
-    } else {
-      dependencyManager.registerSingleton(InjectionSymbols.StorageService, AskarStorageService)
-    }
+    dependencyManager.registerSingleton(AskarStorageService)
```
TimoGlastra, line 16: Would be good to keep this check

packages/askar/src/AskarModule.ts
```diff
@@ -1,33 +1,18 @@
...
-        throw new Error('Could not load aries-askar bindings')
-      }
-    }
+  public readonly config: AskarModuleConfig
 
-    if (dependencyManager.isRegistered(InjectionSymbols.Wallet)) {
-      throw new AriesFrameworkError('There is an instance of Wallet already registered')
-    } else {
-      dependencyManager.registerContextScoped(InjectionSymbols.Wallet, AskarWallet)
-    }
```
TimoGlastra, line 25: Same for this

packages/askar/src/AskarModuleConfig.ts
```diff
@@ -0,0 +1,65 @@
...
+   *  }
+   * })
+   * ```
+   */
+  ariesAskar: AriesAskar
+
+  /**
+   * A list of Askar wallets to connect to.
+   */
+  wallets: [AskarWalletPostgresConfig, ...AskarWalletPostgresConfig[]]
```
TimoGlastra, line 48: We don't configure the wallet through the askar module (for now at least). only ariesAskar is fine for now

packages/core/package.json
```diff
@@ -51,6 +51,7 @@
     "web-did-resolver": "^2.0.8"
   },
   "devDependencies": {
+    "@hyperledger/anoncreds-nodejs": "^0.1.0-dev.6",
```
TimoGlastra, line 54: I don't think there's a need to add this?

packages/indy-vdr/src/__tests__/IndyVdrModule.test.ts
```diff
@@ -1,5 +1,7 @@
 import type { DependencyManager } from '@aries-framework/core'
 
+import { indyVdr } from '@hyperledger/indy-vdr-shared'
```
TimoGlastra, line 3: Please see my other comment about which of the packages should be used where and carefully review this pr yourself to make sure we're importing from the right pacakges

packages/anoncreds/package.json
```diff
@@ -31,6 +31,8 @@
   },
   "devDependencies": {
     "@aries-framework/node": "0.3.3",
+    "@hyperledger/indy-vdr-nodejs": "^0.1.0-dev.6",
```
TimoGlastra, line 34: anoncreds doesn't need to depend on indyVdr I think? I think you can import `indyVdrModuleConfig` from the indy-vdr package test folder

packages/askar/src/AskarModule.ts
Vickysomtee: @TimoGlastra  keep this check along with my new implementation?
Ex.
```ts
    dependencyManager.registerInstance(AskarModuleConfig, this.config)

    dependencyManager.registerSingleton(AskarStorageService)

    if (dependencyManager.isRegistered(InjectionSymbols.Wallet)) {
      throw new AriesFrameworkError('There is an instance of Wallet already registered')
    } else {
      dependencyManager.registerContextScoped(InjectionSymbols.Wallet, AskarWallet)
    }

    if (dependencyManager.isRegistered(InjectionSymbols.StorageService)) {
      throw new AriesFrameworkError('There is an instance of StorageService already registered')
    } else {
      dependencyManager.registerSingleton(InjectionSymbols.StorageService, AskarStorageService)
    }
  }
```

genaris: In this particular case is good to keep it because if there is already a Wallet and a StorageService registered (which means that there is other module serving to the same functionality) and we simply register the symbols, it will override the classes registered by the other module, which is for sure not a desired behaviour. So this will force the user to 'do the right thing' and not use two equivalent modules (it's like using all these new components and indy-sdk at the same time).

packages/anoncreds/src/AnonCredsModuleConfig.ts
TimoGlastra: You added the `anoncreds` property to the `@aries-framework/anoncreds` package (`AnonCredsModule`), while it needs to be added to the `@aries-framework/anoncreds-rs` (`AnonCredsRsModule`)

genaris (approved):

packages/anoncreds/tests/legacyAnonCredsSetup.ts
```diff
@@ -158,14 +161,14 @@ export const getAskarAnonCredsIndyModules = ({
     anoncreds: new AnonCredsModule({
       registries: [new IndyVdrAnonCredsRegistry()],
     }),
-    anoncredsRs: new AnonCredsRsModule(),
-    indyVdr: new IndyVdrModule({
```
TimoGlastra, line 162: `indyNetworkConfig` is now unused. I think we can remove it if we're using `indyVdrModuleConfig`

packages/askar/src/AskarModule.ts
```diff
@@ -1,22 +1,22 @@
...
-    } catch (error) {
-      try {
-        require('@hyperledger/aries-askar-react-native')
-      } catch (error) {
-        throw new Error('Could not load aries-askar bindings')
-      }
-    }
+    dependencyManager.registerInstance(AskarModuleConfig, this.config)
+
+    dependencyManager.registerSingleton(AskarStorageService)
```
TimoGlastra, line 19: This one is registered conditionally a few lines below

```suggestion
```

packages/askar/src/AskarModuleConfig.ts
```diff
@@ -0,0 +1,58 @@
...
+   *   })
+   *  }
+   * })
+   * ```
+   */
+  ariesAskar: AriesAskar
+
+  /**
+   * A list of Askar wallets to connect to.
+   */
```
TimoGlastra, line 42:
```suggestion
```

packages/indy-vdr/src/__tests__/IndyVdrModuleConfig.test.ts
```diff
@@ -1,12 +1,15 @@
 import type { IndyVdrPoolConfig } from '../pool'
 
+import { indyVdr } from '@hyperledger/indy-vdr-shared'
```
TimoGlastra, line 3: This should import from the nodejs package

genaris:
It became almost impossible to get past CI in node 18 due to memory issues. The recent Askar update can be related to this, however not sure why, as we are taking more measures for object handle clearing than before.

I'm wondering if merging first https://github.com/hyperledger/aries-framework-javascript/pull/1377 can help, as in that PR there are some important additions to handle clearing, notably regarding keys, which are used intensively in E2E tests when packing/unpacking DIDComm messages.

TimoGlastra:
> It became almost impossible to get past CI in node 18 due to memory issues. The recent Askar update can be related to this, however not sure why, as we are taking more measures for object handle clearing than before.
> 
> I'm wondering if merging first #1377 can help, as in that PR there are some important additions to handle clearing, notably regarding keys, which are used intensively in E2E tests when packing/unpacking DIDComm messages.

Curious why it seems this PR is hit by it way worse than others?

TimoGlastra:
Yeah it has to be something with this branch. All other PRs have no issues with CI

genaris:
> Yeah it has to be something with this branch. All other PRs have no issues with CI
> 
> We switched indy-vdr from v9 to v10 in this branch. Let's see if it resolves the issue if we downgrade.

D'oh haven't seen this and I upgraded again to v10 in #1373 😞. Anyway, there it worked fine at the first try.

TimoGlastra:
Yeah seems that downgrading didn't do the trick. So it's something else, which is weird!?!? 

Or maybe it's because we're now importing the node.js modules more often, which results in the memory issues?

packages/indy-vdr/tests/setup.ts
```diff
@@ -1,6 +1,3 @@
-// Needed to register indy-vdr node bindings
-import '../src/index'
-
-require('@hyperledger/indy-vdr-nodejs')
+import '@hyperledger/indy-vdr-nodejs'
```
genaris, line 1: These imports in tests' setup.ts are not needed anymore, are they?

TimoGlastra: They are! Some tests don't use the module (e.g. `IndyVdrSovDidResolver.test.ts`), but do use the binary. It will result in the binary not being registered on the shared package if you run the test (mainly in isolation).

I first started importing the nodejs bindings in these specific files, but I thought why make it more complex than needs to be and just imported it in the setup

genaris:
> Yeah seems that downgrading didn't do the trick. So it's something else, which is weird!?!?
> 
> Or maybe it's because we're now importing the node.js modules more often, which results in the memory issues?

This is absolutely possible, as in the wrappers we only provide a method to load libraries but not free resources used by them (which is certainly a bug, unless the runtime is freeing them automatically... which is not usually the case from our recent experience). `ffi-napi` provides ways to unload libraries, so I'll dig a bit more on it to see how it works and if we can solve it in a relatively easy way.

TimoGlastra:
> > Yeah seems that downgrading didn't do the trick. So it's something else, which is weird!?!?
> > Or maybe it's because we're now importing the node.js modules more often, which results in the memory issues?
> 
> This is absolutely possible, as in the wrappers we only provide a method to load libraries but not free resources used by them (which is certainly a bug, unless the runtime is freeing them automatically... which is not usually the case from our recent experience). `ffi-napi` provides ways to unload libraries, so I'll dig a bit more on it to see how it works and if we can solve it in a relatively easy way.

We should maybe add a check and see if it's already loaded, and if so, don't load it again. I wouldn't expect importing a package multiple times has implications. If you're working in Node.JS you could only work with the Node.JS package, and never add / import the shared package...

TimoGlastra:
100th time's the charm 🤞

TimoGlastra:
Never thought this day would come

genaris:
> Never thought this day would come

The only thing we had to do was to rename it! Let's keep this in mind for the next cursed PR.

PR #1354, opened 2023-02-28
refactor!: remove Dispatcher.registerMessageHandler

genaris:
It was marked as deprecated since 0.3.0

BREAKING CHANGE:

`Dispatcher.registerMessageHandler` has been removed in favour of `MessageHandlerRegistry.registerMessageHandler`

### DIFF
--- a/packages/action-menu/src/ActionMenuApi.ts
+++ b/packages/action-menu/src/ActionMenuApi.ts
@@ -10,7 +10,6 @@ import {
   AgentContext,
   AriesFrameworkError,
   ConnectionService,
-  MessageHandlerRegistry,
   MessageSender,
   OutboundMessageContext,
   injectable,
@@ -36,7 +35,6 @@ export class ActionMenuApi {
   private agentContext: AgentContext
 
   public constructor(
-    messageHandlerRegistry: MessageHandlerRegistry,
     connectionService: ConnectionService,
     messageSender: MessageSender,
     actionMenuService: ActionMenuService,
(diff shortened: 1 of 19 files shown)


packages/action-menu/src/ActionMenuApi.ts
```diff
@@ -36,7 +36,7 @@ export class ActionMenuApi {
   private agentContext: AgentContext
 
   public constructor(
-    dispatcher: Dispatcher,
+    messageHandlerRegistry: MessageHandlerRegistry,
```
TimoGlastra, line 39: Do you think we should inject the  `DependencyManager` and register the handlers on that class?

genaris: Right, for extension modules `DependencyManager` is more convenient. In such case I think there is no need to inject it, as it can be retrieved from `AgentContext`.

Actually I think it can be used in all the case. Or do you think that conceptually would be better to only use it out of the core?

packages/core/src/modules/basic-messages/BasicMessagesApi.ts
```diff
@@ -91,7 +91,7 @@ export class BasicMessagesApi {
     await this.basicMessageService.deleteById(this.agentContext, basicMessageRecordId)
   }
 
-  private registerMessageHandlers(dispatcher: Dispatcher) {
-    dispatcher.registerMessageHandler(new BasicMessageHandler(this.basicMessageService))
+  private registerMessageHandlers(messageHandlerRegistry: MessageHandlerRegistry) {
+    messageHandlerRegistry.registerMessageHandler(new BasicMessageHandler(this.basicMessageService))
   }
```
TimoGlastra, line 96: I'm thinking now, if we register these in the Api, wouldn't it mean they're registered every time a new agent is initialized? (in multi-tenancy). They share the same message handler registry

packages/action-menu/src/ActionMenuApi.ts
TimoGlastra: > Actually I think it can be used in all the case. Or do you think that conceptually would be better to only use it out of the core?

I think if we see the `DependencyManager` as the public api for extension module is not a bad idea. That can wrap the message handler registry and we'll be free to make changes to that class, but changes to the the `DependencyManager` will be seen as breaking

TimoGlastra (approved):
Approving as my last comment is not related to the changes in this PR

PR #1356, opened 2023-02-28
fix(core): repository event when calling deleteById

genaris:
When calling Repository method `deleteById`, no `RecordEvent` event is emitted. This PR fixes this by internally getting the record by id and then calling `delete` method, which does emit the event.

Even if it could be more efficient to avoid the need of getById, we need to get the whole record in order to be able to emit the event with the record object (otherwise we would only have its id), being consistent with other repository events.

### DIFF
--- a/packages/core/src/storage/Repository.ts
+++ b/packages/core/src/storage/Repository.ts
@@ -68,12 +68,13 @@ export class Repository<T extends BaseRecord<any, any, any>> {
   }
 
   /**
-   * Delete record by id. Returns null if no record is found
+   * Delete record by id. Throws {RecordNotFoundError} if no record is found
    * @param id the id of the record to delete
    * @returns
    */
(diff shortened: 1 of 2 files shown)


genaris:
> The reason for the deleteById was to prevent the need to call getById.
> 
> I'm thinking whether it may be better to update the delete even to either include an id or record?

True...🤦. I'm biased because we are calling the Repository method directly using the id and we got into this problem as we are registered to repository events using hooks similar to the ones from aries-framework-javascript-ext.

Anyway I think an event must be emitted. I don't like so much the idea of having either the id or the record but I guess it's a good trade-off to allow both ways of deleting a record.

TimoGlastra (approved):
This makes sense to me. We may want to tweak it in the future, but at least this doesn't impact the deleteById method

PR #1361, opened 2023-03-02
test: various improvements

TimoGlastra:
This PR adds a few improvements to the test setup that should hopefully prevent timeouts, but also make sure we test the askar setup:
- only run askar related tests in node18
- add ref-napi resolution in node18
- increase timeout for some packages that often timeout

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -89,9 +89,13 @@ module.exports = {
       },
     },
     {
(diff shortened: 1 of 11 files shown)


genaris (approved):
Looks super hacky but effective! I wish there was an official support in yarn for selective resolutions according to engine version...

PR #1362, opened 2023-03-02
feat(indy-vdr)!: extend did:indy support

TimoGlastra:
Same changes as in #1347, but now for indy-vdr.

It updates the anoncreds registry to support both legacy and did:indy identifiers.

There's one issue that needs to be resolved, but I'd like to do that in a separate PR, is that currently the code isn't set up in a way yet that we support using both did:indy and legacy identifiers in anoncreds for _issuance_. As the holder and verifier only do resolving, it works with both now. 

What still needs to be done:
- when a credential with legacy identifiers or vice versa is received, allow it to be shared as the other format. I want to store all credentials using the new did:indy identifiers, and we can optionally transform the identifiers to legacy when needed (in the credential/proof format services)
- creating an offer for legacy format as the record always stores using the did:indy format. I'd also like to handle this in the format services, so the storage / issuer/holder/verifier services always use the new formats.


packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
```diff
@@ -41,12 +45,14 @@ export class IndyVdrAnonCredsRegistry implements AnonCredsRegistry {
     try {
       const indyVdrPoolService = agentContext.dependencyManager.resolve(IndyVdrPoolService)
 
-      const did = didFromSchemaId(schemaId)
-
-      const pool = await indyVdrPoolService.getPoolForDid(agentContext, did)
-
+      // parse schema id (supports did:indy and legacy)
+      const { did, didIdentifier, schemaName, schemaVersion } = parseSchemaId(schemaId)
```
genaris, line 49: This new `didIdentifier` term (used in various new RegExp) comes to replace legacy `unqualifiedDid`, right? As it corresponds to the namespace identifier (as defined in did:indy method spec for instance) maybe could it be called `namespaceIdentifier`?

packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
```diff
@@ -106,57 +110,41 @@ export class IndyVdrAnonCredsRegistry implements AnonCredsRegistry {
...
-          schema: options.schema,
-          state: 'failed',
-        },
-      }
-    }
-
     try {
+      // This will throw an error if trying to register a schema with a legacy indy identifier. We only support did:indy identifiers
+      // for registering, that will allow us to extract the namespace and means all stored records will use did:indy identifiers.
+      const { id: unqualifiedDid, namespace } = parseIndyDid(options.schema.issuerId)
```
genaris, line 118: This leaves me the same question about difference between `unqualifiedDid` and `didIdentifier` (or my suggested `namespaceIdentifier`)

packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
```diff
@@ -203,53 +190,58 @@ export class IndyVdrAnonCredsRegistry implements AnonCredsRegistry {
...
 
       agentContext.config.logger.trace(
         `Submitting get credential definition request for credential definition '${credentialDefinitionId}' to ledger '${pool.indyNamespace}'`
       )
-
       const response = await pool.submitReadRequest(request)
 
-      const schema = await this.fetchIndySchemaWithSeqNo(agentContext, response.result.ref, did)
+      // We need to fetch the schema to determine the schemaId (we only have the seqNo)
+      const schema = await this.fetchIndySchemaWithSeqNo(agentContext, response.result.ref, didIdentifier)
```
genaris, line 212: Would it make sense for AnonCredsRegistrars to first check in the AnonCredsSchemaRepository so no extra transaction is required? And also store it after retrieving? Or this is something that should be done in AnonCredsApi level? 

This is not strictly related to this PR but was something that felt a bit awkward when doing integration with AATH.

packages/indy-vdr/src/dids/IndyVdrIndyDidRegistrar.ts
```diff
@@ -76,26 +78,62 @@ export class IndyVdrIndyDidRegistrar implements DidRegistrar {
             },
           }
         }
-        if (!isSelfCertifiedIndyDid(did, verkey)) {
-          throw new Error(`Initial verkey ${verkey} does not match did ˇ${did}`)
+
+        const { namespace, id } = parseIndyDid(did)
```
genaris, line 82: I see it's used afterwards, but maybe would be nice to use directly here the same naming than Anoncreds Registry class, as it is more explicit.

packages/indy-vdr/src/dids/IndyVdrIndyDidRegistrar.ts
```diff
@@ -62,11 +63,12 @@ export class IndyVdrIndyDidRegistrar implements DidRegistrar {
     }
 
     try {
-      const { namespace, id: submitterId } = parseIndyDid(submitterDid)
+      // Parse submitterDid and extract namespace based on the submitter did
+      const { namespace: submitterNamespace, id: submitterDidIdentifier } = parseIndyDid(submitterDid)
+      const submitterSigningKey = await verificationKeyForIndyDid(agentContext, submitterDid)
```
genaris, line 68: Nice! I didn't want to call so many APIs from this class but it is certainly more clear for this Registrar API

packages/indy-vdr/src/dids/IndyVdrIndyDidRegistrar.ts
```diff
@@ -222,41 +268,44 @@ export class IndyVdrIndyDidRegistrar implements DidRegistrar {
...
+        dest: unqualifiedDid,
+        verkey: signingKey.publicKeyBase58,
+        alias,
+      })
 
-      const response = await pool.submitWriteRequest(agentContext, request, signingKey)
+      const response = await pool.submitWriteRequest(agentContext, request, submitterSigningKey)
 
-      agentContext.config.logger.debug(`Registered public did '${targetDid}' on ledger '${pool.indyNamespace}'`, {
+      agentContext.config.logger.debug(`Registered public did '${unqualifiedDid}' on ledger '${pool.indyNamespace}'`, {
```
genaris, line 296: Do you think it could be convenient to log using the full did format? Aligned with the way we are logging ledger objects in AnonCredsRegistry

genaris (approved):
Left some minor comments / questions but LGTM!!!

packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
TimoGlastra: We could check in the schema repository. I'm not sure about storing it as the records are for things this agent created, not for just any schema record.

I think not having to fetch the schema is something we could solve using the cache.

packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
TimoGlastra: Yeah `namespaceIdentifier` is great!

packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
TimoGlastra: We can't actually look at the schemaRecord as we don't store the seqNo. We should probably add it to the metadata, but I wanted to have as little indy specific logic as possible. 

I'll address this in a follow up PR

packages/indy-vdr/src/dids/IndyVdrIndyDidRegistrar.ts
TimoGlastra: I though I understood this comment, but I don't understand. Do you mean to also use `namespaceIdentifier` here?

packages/indy-vdr/src/dids/IndyVdrIndyDidRegistrar.ts
TimoGlastra: Yeah that would be convenient, but it means we need to pass either both the qualified and unqualified did or parse it again (which as overhead). 

What's your preference?

packages/indy-vdr/src/dids/IndyVdrIndyDidRegistrar.ts
genaris: Yes, basically I was referring to that. But just a bit nitpicking there, I think it's still clear.

packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
genaris: Good point! Of course not needed for this PR, just some thinking that came up to my head while reviewing this one.

packages/indy-vdr/src/dids/IndyVdrIndyDidRegistrar.ts
genaris: Oh I see.. I think it's OK to use the unqualified in that case. Maybe not calling it a 'public did' but simply id or namespace id. But again, it's just some nitpicking here.

genaris (approved):

PR #1363, opened 2023-03-02
fix(askar): expose tag utilities

blu3beri:
Signed-off-by: blu3beri <blu3beri@proton.me>


TimoGlastra (approved):
I would rather not expose these, but not so much we can do about it rigth?

blu3beri:
> I would rather not expose these, but not so much we can do about it rigth?

I am also okay with reimplementing that one function I need. 

Its a tiny amount of overhead so I will do that.

PR #1367, opened 2023-03-04
fix(anoncreds): Buffer not imported from core

genaris:
Subtle omission that was causing `encodeCredentialValue` to fail under React Native.

### DIFF
--- a/packages/anoncreds/src/utils/credential.ts
+++ b/packages/anoncreds/src/utils/credential.ts
@@ -1,7 +1,7 @@
 import type { AnonCredsSchema, AnonCredsCredentialValues } from '../models'
 import type { CredentialPreviewAttributeOptions, LinkedAttachment } from '@aries-framework/core'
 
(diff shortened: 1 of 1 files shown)


packages/anoncreds/src/utils/credential.ts
```diff
@@ -1,7 +1,7 @@
 import type { AnonCredsSchema, AnonCredsCredentialValues } from '../models'
 import type { CredentialPreviewAttributeOptions, LinkedAttachment } from '@aries-framework/core'
 
-import { AriesFrameworkError, Hasher, encodeAttachment } from '@aries-framework/core'
+import { AriesFrameworkError, Hasher, encodeAttachment, Buffer } from '@aries-framework/core'
```
TimoGlastra, line 4: Hmm I added an eslint rule to disallow global Buffer. Not sure why it doesn't work?...

TimoGlastra (approved):

PR #1370, opened 2023-03-07
feat: add oob proof proposal

Zzocker:


Zzocker:
Yes, I wanted to add test for the changes but I wasn't sure whether to create a new test file or add in an existing test file.

When verifier receives the proposal message, the proposal handler function will be invoked and proposal will be proceeded, and if auto accept proof is true, verifier will send the proof request. Do we need to have aditional logic for reciving the proposal message ?

Zzocker:
I have added oob proof proposal for both (`v1` and `v2`) versions of the proof request. Also added test cases. @TimoGlastra can you again review this PR?

packages/anoncreds/src/protocols/proofs/v1/V1ProofProtocol.ts
```diff
@@ -416,6 +416,12 @@ export class V1ProofProtocol extends BaseProofProtocol implements ProofProtocol<
 
     let proofRecord = await this.findByThreadAndConnectionId(agentContext, proofRequestMessage.threadId, connection?.id)
 
+    if (!proofRecord) {
+      // Proof request not bound to any connection: requests_attach in OOB msg
```
TimoGlastra, line 420: This doesn't mean the proof record is not bound to any connection, and it means if i know the thread id i respond to this exchange even though not involved. 

We should update this logic to be fully integrated with the oob module:
- we look for the oob record
- if oob is connectionless/now has a connection afterwards we allow the proof record to not have a connection


I'd like to solve this generically for both connectionless/non-connectionless exchanges with attached messages. Is this needed for this PR?

packages/anoncreds/src/protocols/proofs/v1/__tests__/v1-indy-proof-proposal.e2e.test.ts
```diff
@@ -103,4 +111,114 @@ describe('Present Proof', () => {
...
+      },
+      comment: 'V1 propose proof test',
+    })
+    const { outOfBandInvitation } = await aliceAgent.oob.createInvitation({
+      messages: [message],
+      autoAcceptConnection: true,
+    })
+    await faberAgent.oob.receiveInvitation(outOfBandInvitation)
+    testLogger.test('Faber waits for proof proposal message from Alice')
+    let faberProofExchangeRecord = await waitForProofExchangeRecord(faberAgent, {
```
TimoGlastra, line 169: Can you use waitForProofExchangeRecordSubject? This method can have timing issues

packages/core/src/modules/proofs/ProofsApiOptions.ts
```diff
@@ -38,6 +38,17 @@ export interface ProposeProofOptions<PPs extends ProofProtocol[] = ProofProtocol
   parentThreadId?: string
 }
 
+/**
+ * Interface for ProofsApi.createProofProposal. Will create an out of band request.
```
TimoGlastra, line 42:
```suggestion
 * Interface for ProofsApi.createProofProposal. Will create an out of band proposal.
```

packages/core/src/modules/proofs/protocol/v2/V2ProofProtocol.ts
```diff
@@ -396,6 +396,12 @@ export class V2ProofProtocol<PFs extends ProofFormatService[] = ProofFormatServi
       connection?.id
     )
 
+    if (!proofRecord) {
+      // Proof request not bound to any connection: requests_attach in OOB msg
+      proofRecord = await this.findByThreadAndConnectionId(messageContext.agentContext, requestMessage.threadId)
```
TimoGlastra, line 401: Same comments here

packages/core/src/modules/proofs/protocol/v2/__tests__/v2-indy-oob-proof-proposal.e2e.test.ts
```diff
@@ -0,0 +1,418 @@
...
+
+    await waitForProofExchangeRecordSubject(aliceReplay, {
+      state: ProofState.Done,
+    })
+    await waitForProofExchangeRecordSubject(faberReplay, {
+      state: ProofState.Done,
+    })
+  })
+
+  test('Alice start with oob proof proposal for Faber with auto-accept enabled and both agents having a mediator', async () => {
```
TimoGlastra, line 250: Specific reason you added this test? We have a lot of tests that do almost the same and it's making updates harder. 

So if there's a specific reason for this test great, but otherwise we can maybe remove it?

TimoGlastra (changes requested):
Grest changes @Zzocker. I have some concerns on the connection between oob and proofs. This is something that needs to be fixes across the framework.

packages/core/src/modules/proofs/protocol/v2/__tests__/v2-indy-oob-proof-proposal.e2e.test.ts
Zzocker: Right, For the oob proof proposal, we want to test
- After the proof exchange flow is completed, 
   - The holder should have all three proof messages `proposalMessage`, `requestMessage` , and `presentationMessage`
   - proof exchange record should have `nonNull` connection id.

I will remove the first and last tests and add the above assertion in 2nd test.

TimoGlastra:
Discussed during WG call. @TimoGlastra will take another look

genaris:
Superseded by #2471

PR #1373, opened 2023-03-07
feat(anoncreds-rs): use new API methods for json conversion

genaris:
This PR updates @hyperledger/anoncreds-shared and @hyperledger/anoncreds-nodejs to the version published today. 

At the moment, only those changes referred to fromJson/toJson conversion are done. 

In `AnonCredsRsHolderService` entropy for credential request is currently fixed to a nonce.


packages/anoncreds-rs/src/services/AnonCredsRsHolderService.ts
```diff
@@ -48,31 +51,35 @@ export class AnonCredsRsHolderService implements AnonCredsHolderService {
     agentContext: AgentContext,
     options?: CreateLinkSecretOptions
   ): Promise<CreateLinkSecretReturn> {
+    let masterSecret: MasterSecret | undefined
     try {
+      masterSecret = MasterSecret.create()
+
+      // FIXME: This is a very specific format of anoncreds-rs. I think it should be simply a string
+      const linkSecretJson = masterSecret.toJson() as { value: { ms: string } }
```
blu3beri, line 59: Yeah I think it was done like that for Indy-sdk reasons. Can you create an issue for that if we should fix this in anoncreds-rs?

packages/anoncreds-rs/src/services/AnonCredsRsHolderService.ts
```diff
@@ -48,31 +51,35 @@ export class AnonCredsRsHolderService implements AnonCredsHolderService {
...
-      throw new AnonCredsRsError('Error creating Link Secret', { cause: error })
+    } finally {
+      masterSecret?.handle.clear()
     }
   }
 
   public async createProof(agentContext: AgentContext, options: CreateProofOptions): Promise<AnonCredsProof> {
     const { credentialDefinitions, proofRequest, selectedCredentials, schemas } = options
 
+    let presentation: Presentation | undefined
```
blu3beri, line 73:
```suggestion
    let presentation: Presentation
```

Should be fine I believe.

packages/anoncreds-rs/src/services/AnonCredsRsHolderService.ts
```diff
@@ -48,31 +51,35 @@ export class AnonCredsRsHolderService implements AnonCredsHolderService {
...
     try {
-      const rsCredentialDefinitions: Record<string, CredentialDefinition> = {}
+      const rsCredentialDefinitions: Record<string, JsonObject> = {}
```
blu3beri, line 75: Is there a specific reason to not cast to specifically anymore?

packages/anoncreds-rs/src/services/AnonCredsRsHolderService.ts
```diff
@@ -89,45 +96,50 @@ export class AnonCredsRsHolderService implements AnonCredsHolderService {
...
-        if (timestamp) {
-          if (revocationRegistryIndex) {
-            if (!options.revocationRegistries[revocationRegistryDefinitionId]) {
-              throw new AnonCredsRsError(`Revocation Registry ${revocationRegistryDefinitionId} not found`)
-            }
+        let revocationState: CredentialRevocationState | undefined
+        let revocationRegistryDefinition: RevocationRegistryDefinition | undefined
+        try {
+          if (timestamp) {
+            if (revocationRegistryIndex && revocationRegistryDefinitionId) {
```
blu3beri, line 110: It's a bit difficult to read, but can we combine these? It's getting a bit deep here with the nesting.

blu3beri (changes requested):
Have not finished reviewing it, but here are some initial comments.

packages/anoncreds-rs/src/services/AnonCredsRsHolderService.ts
genaris: Yes, we are passing to anoncreds library the JSON object (as it comes from the options argument) instead of instantiating an AnonCredsObject and then passing its handle to anoncreds library (so it will be the one responsible for object instantiation and handle clearing, as we don't need them for anything else).

packages/anoncreds-rs/src/services/AnonCredsRsHolderService.ts
genaris: I had to add it because I'm clearing it in the finally block (without it, it gives a `Variable 'presentation' is used before being assigned`)

packages/anoncreds-rs/src/services/AnonCredsRsHolderService.ts
genaris: Done! https://github.com/hyperledger/anoncreds-rs/issues/138

packages/anoncreds-rs/src/services/AnonCredsRsHolderService.ts
blu3beri: Ahh yeah makes sense.

TimoGlastra (approved):

PR #1374, opened 2023-03-08
feat(anoncreds): use legacy prover did

genaris:
Adds `useLegacyProverDid` flag for credential request creation, which is always set when dealing with legacy indy identifiers. Some checks were added in `LegacyIndyCredentialFormatService` to make sure we are using legacy ones.

When using IndySdk, only legacy prover_did can be used, while in AnonCredsRs through this flag we can select between prover_did and entropy, unless we are using new identifiers where only entropy can be generated.


TimoGlastra:
What's the blocker for this PR @genaris?

genaris:
> What's the blocker for this PR @genaris?

It was https://github.com/hyperledger/anoncreds-rs/pull/137

Now it's merged but waiting for the release in npm integrating it.

TimoGlastra:
Cool! Release is cooking now: https://github.com/hyperledger/anoncreds-rs/releases/tag/v0.1.0-dev.10

packages/anoncreds-rs/src/services/AnonCredsRsHolderService.ts
```diff
@@ -212,8 +214,14 @@ export class AnonCredsRsHolderService implements AnonCredsHolderService {
...
       }
 
+      const isLegacyIdentifier = credentialOffer.cred_def_id.match(legacyIndyCredentialDefinitionIdRegex)
+      if (!isLegacyIdentifier && useLegacyProverDid) {
+        throw new AriesFrameworkError('Cannot use legacy prover_did with non-legacy identifiers')
+      }
+
       createReturnObj = CredentialRequest.create({
-        entropy: anoncreds.generateNonce(), // FIXME: find a better source of entropy
+        entropy: !useLegacyProverDid || !isLegacyIdentifier ? anoncreds.generateNonce() : undefined, // FIXME: find a better source of entropy
```
genaris, line 223: Another option for generating entropy can be calling `agentContext.wallet.generateWalletKey`, which in AskarWallet creates an ephemeral key and converts its private key to a base58 string (I guess IndySdkWallet does more or less the same thing). However I'm not sure if it's good to rely on wallet implementation for that (unless we strictly specify the requirements for such `generateWalletKey` in API).

TimoGlastra: It feels like we're misusing the generateWalletKey in that case. I think a nonce should provide enough randomness right?

packages/anoncreds-rs/src/services/AnonCredsRsHolderService.ts
```diff
@@ -212,8 +214,14 @@ export class AnonCredsRsHolderService implements AnonCredsHolderService {
...
+        proverDid: useLegacyProverDid ? generateLegacyProverDidLikeString() : undefined,
```
TimoGlastra, line 224: As it should be random, do wr need to make this a secure ranom generation? E.g. nonce and then encode it as base58?

We could also add a getRandomBytes method to the wallet?

packages/anoncreds/tests/InMemoryAnonCredsRegistry.ts
```diff
@@ -94,21 +98,23 @@ export class InMemoryAnonCredsRegistry implements AnonCredsRegistry {
...
+    if (this.useLegacyIdentifiers) {
+      legacyIssuerId = options.schema.issuerId
+    } else {
+      const { namespace, namespaceIdentifier } = parseIndyDid(options.schema.issuerId)
+      legacyIssuerId = namespaceIdentifier
+      didIndySchemaId = getDidIndySchemaId(namespace, namespaceIdentifier, options.schema.name, options.schema.version)
+      this.schemas[didIndySchemaId] = options.schema
+    }
 
+    const legacySchemaId = getLegacySchemaId(legacyIssuerId, options.schema.name, options.schema.version)
```
TimoGlastra, line 112: The old implementation supported both for convenience, why is this needed?

TimoGlastra (approved):

packages/anoncreds/tests/InMemoryAnonCredsRegistry.ts
genaris: If schema issuerId is an unqualified did, `parseIndyDid` threw an error. Maybe I can do it 'automatic' (i.e. without config flag and a try-catch?)

TimoGlastra: It's for tests, so whatever works
Is fine. Just curious

packages/anoncreds-rs/src/services/AnonCredsRsHolderService.ts
genaris: IMHO the nonce generated by anoncreds, apart from being convenient for us, is enough for the purposes of giving some entropy in credential generation. But just trying to find an alternative that provides a true 'alphanumeric' random string as stated in the spec.

packages/anoncreds-rs/src/services/AnonCredsRsHolderService.ts
genaris: I copied it from anoncreds/indy-sdk package implementation, but also thought that could be a good idea to rely in anoncreds native library instead of uuid as it's being done currently in `generateLegacyProverDidLikeString`.

Any 22-bytes long base58 string will be good for that purpose, right?

TimoGlastra: As it's used for entropy, it may be needed for the string to be generated with actual secure random bytes instead of a random string?

TimoGlastra (approved):

TimoGlastra:
LGTM!!

PR #1377, opened 2023-03-08
feat(askar): import/export wallet support for SQLite

genaris:
Add support for wallet export and import by copying sqlite file and executing a key rotation to the one specified in the `WalletExportImportConfig` object. Similarly, when importing the wallet it will copy the origin file to the expected directory according to wallet id and do a rekey to the key specified.

In addition, there are a few tweaks and fixes for things I found during testing such as: 

- Key object handles clearing in pack/unpack and key creation/signing/verification methods
- Check wallet database file is created before (not sure why, but Askar is not throwing Duplicate error in `Store.provision` when wallet exists). This only works for SQLite but it's better than nothing
- Use key derivation based in argon2 Int if nothing specified, as it's the default in Indy SDK AFAIK

### DIFF
--- a/packages/askar/src/wallet/AskarWallet.ts
+++ b/packages/askar/src/wallet/AskarWallet.ts
@@ -9,8 +9,9 @@ import type {
   WalletConfigRekey,
   KeyPair,
   KeyDerivationMethod,
+  WalletExportImportConfig,
 } from '@aries-framework/core'
-import type { Session } from '@hyperledger/aries-askar-shared'
+import type { KeyEntryObject, Session } from '@hyperledger/aries-askar-shared'
 
 import {
   WalletKeyExistsError,
@@ -131,6 +132,14 @@ export class AskarWallet implements Wallet {
     this.logger.debug(`Creating wallet '${walletConfig.id}`)
 
     const askarWalletConfig = await this.getAskarWalletConfig(walletConfig)
+
+    // Check if database exists
+    const { path: filePath } = uriFromWalletConfig(walletConfig, this.fileSystem.dataPath)
+    if (filePath && (await this.fileSystem.exists(filePath))) {
+      throw new WalletDuplicateError(`Wallet '${walletConfig.id}' already exists.`, {
+        walletType: 'AskarWallet',
+      })
+    }
     try {
       this._store = await Store.provision({
         recreate: false,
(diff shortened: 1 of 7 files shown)


packages/askar/src/wallet/AskarWallet.ts
```diff
@@ -131,6 +132,14 @@ export class AskarWallet implements Wallet {
     this.logger.debug(`Creating wallet '${walletConfig.id}`)
 
     const askarWalletConfig = await this.getAskarWalletConfig(walletConfig)
+
+    // Check if database exists
+    const { path: filePath } = uriFromWalletConfig(walletConfig, this.fileSystem.dataPath)
+    if (filePath && (await this.fileSystem.exists(filePath))) {
```
TimoGlastra, line 138: Should we check whether the storage_type is postgres?

packages/askar/src/wallet/AskarWallet.ts
```diff
@@ -393,19 +479,22 @@ export class AskarWallet implements Wallet {
...
-        const keyEntry = await this.session.fetchKey({ name: key.publicKeyBase58 })
+        keyEntry = await this.session.fetchKey({ name: key.publicKeyBase58 })
 
         if (!keyEntry) {
           throw new WalletError('Key entry not found')
         }
 
         const signed = keyEntry.key.signMessage({ message: data as Buffer })
 
+        keyEntry.key.handle.free()
```
TimoGlastra, line 496: :( we should be careful where we'll allow this free handling. Ideally we split this from the normal AFJ code, or we e.g. look at something like sesssions for a key and will close all keys in a session at once. I just don't think handle logic should be exposed from the askar library

TimoGlastra: Not for now ofc.

packages/askar/src/wallet/AskarWallet.ts
```diff
@@ -699,7 +813,9 @@ export class AskarWallet implements Wallet {
       uri,
       profile: walletConfig.id,
       // FIXME: Default derivation method should be set somewhere in either agent config or some constants
-      keyMethod: keyDerivationMethodToStoreKeyMethod(walletConfig.keyDerivationMethod) ?? StoreKeyMethod.None,
+      keyMethod:
+        keyDerivationMethodToStoreKeyMethod(walletConfig.keyDerivationMethod) ??
```
TimoGlastra, line 817: Should we set the default in this method? So we have consistent default behavior?

TimoGlastra (approved):

packages/askar/src/wallet/AskarWallet.ts
genaris: Yeah I really don't like having anything related to handles in AFJ or any other consumer of askar/anoncreds/indy-vdr, except maybe for more 'global' objects like sessions or connections. We already are having some fun discussions about this with @blu3beri and the _opinionating_ thing 😛

packages/askar/src/wallet/AskarWallet.ts
genaris: For the moment it was not needed because we only return a path in sqlite, but I think this can be part of the improvements in Wallet API and configuration interfaces that we were discussing some weeks ago and probably we introduce in a near future (hopefully for 0.5.x).

PR #1378, opened 2023-03-11
fix: remove named capture groups

TimoGlastra:
named capture groups are only supported in more recent versions of hermes

It was of course stupid from me to assume regexes would be fully supported by hermes 🙃


genaris (approved):
Hermes never stops surprising us! BTW it's a good name for a pet 😆

PR #1381, opened 2023-03-14
feat: basic message pthid/thid support

genaris:
This is something we are using to keep track of the actual id of messages related to `BasicMessageRecord` and to provide a way to relate a message to another (or another ones), giving the possibility of doing a sort of `message reply` without actually defining a new protocol but using an intrinsic feature of DIDComm messaging.

I'd like to refactor `basic-messages` module to align with others (such as `proofs` and `discover-features`) and to make it adaptable to V2 protocol, but didn't want to introduce more breaking changes before releasing 0.4.0, so I simply added `parentThreadId` as an optional parameter for now.


packages/core/src/modules/basic-messages/BasicMessagesApi.ts
```diff
@@ -41,13 +41,14 @@ export class BasicMessagesApi {
    * @throws {MessageSendingError} If message is undeliverable
    * @returns the created record
    */
-  public async sendMessage(connectionId: string, message: string) {
+  public async sendMessage(connectionId: string, message: string, parentThreadId?: string) {
```
TimoGlastra, line 44: Would be good to overhaul this API, but as you said, we can do that after 0.4.0 is released

packages/core/src/modules/basic-messages/repository/BasicMessageRecord.ts
```diff
@@ -43,6 +48,8 @@ export class BasicMessageRecord extends BaseRecord<DefaultBasicMessageTags, Cust
       this.connectionId = props.connectionId
       this._tags = props.tags ?? {}
       this.role = props.role
+      this.threadId = props.threadId
```
TimoGlastra, line 51: Is there a problem in old exchanges not having a threadId?

TimoGlastra (approved):

packages/core/src/modules/basic-messages/repository/BasicMessageRecord.ts
genaris: I don't think so, besides the fact that it will not be possible for query for them.

PR #1383, opened 2023-03-14
fix: connection id in sessions for new connections

genaris:
When a message for a connection request (either for DID Exchange or Connection protocol) is received, a session is created without any connection id associated to it (as it is not created yet).

If the party requesting a connection does not have a public endpoint (e.g. a edge agent connecting to a mediator) and the responder does not have `autoAcceptConnections` enabled, it will not associate this session to the newly created connection, forcing any call to `agent.connections.acceptRequest` to queue the messages, even if the transport used allows a persistent connection (typical case for WebSockets).

So here what we do is simply to associate the connection to the session as soon as it is created. At the moment I don't see any drawback of this strategy but will be happy to hear any situation where this could represent a problem.

### DIFF
--- a/packages/core/src/agent/TransportService.ts
+++ b/packages/core/src/agent/TransportService.ts
@@ -4,6 +4,7 @@ import type { DidDocument } from '../modules/dids'
 import type { EncryptedMessage } from '../types'
 
 import { DID_COMM_TRANSPORT_QUEUE } from '../constants'
+import { AriesFrameworkError } from '../error'
 import { injectable } from '../plugins'
 
 @injectable()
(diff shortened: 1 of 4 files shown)


packages/core/src/agent/TransportService.ts
```diff
@@ -18,6 +19,15 @@ export class TransportService {
     return Object.values(this.transportSessionTable).find((session) => session?.connectionId === connectionId)
   }
 
+  public setConnectionIdForSession(sessionId: string, connectionId: string) {
+    const session = this.findSessionById(sessionId)
+    if (!session) {
+      throw new AriesFrameworkError(`Session not found with id ${sessionId}`)
+    }
+    session.connectionId = connectionId
+    this.saveSession(session)
```
TimoGlastra, line 28: We directly mutate the session so this doesn't really do anything? But for completeness it makes sense

packages/core/src/modules/connections/handlers/ConnectionRequestHandler.ts
```diff
@@ -62,30 +63,31 @@ export class ConnectionRequestHandler implements MessageHandler {
...
-      senderKey
-    )
+    const receivedDidRecord = await this.didRepository.findReceivedDidByRecipientKey(agentContext, senderKey)
     if (receivedDidRecord) {
       throw new AriesFrameworkError(`A received did record for sender key ${senderKey.fingerprint} already exists.`)
     }
 
     const connectionRecord = await this.connectionService.processRequest(messageContext, outOfBandRecord)
 
+    // Associate the new connection with the session created for the inbound message
```
TimoGlastra, line 73: Ideally we don't do transport stuff in the connection handler, but it seems this is an exception and makes sense

TimoGlastra (approved):

PR #1384, opened 2023-03-17
feat(openid4vc-client): openid authorization flow

karimStekelenburg:
This PR contains the logic for the OpenID For Verifiable Credentials Issuance authorization flow.

Funded by the Ontario Government

### DIFF
--- a/packages/core/src/modules/vc/W3cCredentialService.ts
+++ b/packages/core/src/modules/vc/W3cCredentialService.ts
@@ -105,7 +105,7 @@ export class W3cCredentialService {
     agentContext: AgentContext,
     options: VerifyCredentialOptions
   ): Promise<W3cVerifyCredentialResult> {
-    options.verifyRevocationState = options.verifyRevocationState ?? true
+    const verifyRevocationState = options.verifyRevocationState ?? true
 
     const suites = this.getSignatureSuitesForCredential(agentContext, options.credential)
 
@@ -114,7 +114,7 @@ export class W3cCredentialService {
       suite: suites,
       documentLoader: this.w3cVcModuleConfig.documentLoader(agentContext),
       checkStatus: () => {
-        if (options.verifyRevocationState) {
+        if (verifyRevocationState) {
           throw new AriesFrameworkError('Revocation for W3C credentials is currently not supported')
         }
         return {
--- a/packages/openid4vc-client/src/OpenId4VcClientApi.ts
+++ b/packages/openid4vc-client/src/OpenId4VcClientApi.ts
@@ -24,23 +24,23 @@ export class OpenId4VcClientApi {
 
   public async requestCredentialUsingPreAuthorizedCode(options: PreAuthCodeFlowOptions): Promise<W3cCredentialRecord> {
     // set defaults
-    const checkRevocationState = options.checkRevocationState ?? true
+    const verifyRevocationState = options.verifyRevocationState ?? true
 
     return this.openId4VcClientService.requestCredential(this.agentContext, {
       ...options,
-      checkRevocationState: checkRevocationState,
-      flowType: AuthFlowType.PRE_AUTHORIZED_CODE_FLOW,
+      verifyRevocationState: verifyRevocationState,
+      flowType: AuthFlowType.PreAuthorizedCodeFlow,
     })
   }
 
   public async requestCredentialUsingAuthorizationCode(options: AuthCodeFlowOptions): Promise<W3cCredentialRecord> {
     // set defaults
-    const checkRevocationState = options.checkRevocationState ?? true
+    const checkRevocationState = options.verifyRevocationState ?? true
 
     return this.openId4VcClientService.requestCredential(this.agentContext, {
       ...options,
-      checkRevocationState: checkRevocationState,
-      flowType: AuthFlowType.AUTHORIZATION_CODE_FLOW,
+      verifyRevocationState: checkRevocationState,
+      flowType: AuthFlowType.AuthorizationCodeFlow,
     })
   }
 
--- a/packages/openid4vc-client/src/OpenId4VcClientService.ts
+++ b/packages/openid4vc-client/src/OpenId4VcClientService.ts
@@ -32,7 +32,7 @@ import { randomStringForEntropy } from '@stablelib/random'
 export interface PreAuthCodeFlowOptions {
   issuerUri: string
   kid: string
-  checkRevocationState: boolean
+  verifyRevocationState: boolean
 }
 
 export interface AuthCodeFlowOptions extends PreAuthCodeFlowOptions {
@@ -43,8 +43,8 @@ export interface AuthCodeFlowOptions extends PreAuthCodeFlowOptions {
 }
 
 export enum AuthFlowType {
-  AUTHORIZATION_CODE_FLOW = 'Authorization Code Flow',
-  PRE_AUTHORIZED_CODE_FLOW = 'Pre-Authorized Code Flow',
+  AuthorizationCodeFlow,
+  PreAuthorizedCodeFlow,
 }
 
 export type RequestCredentialOptions = { flowType: AuthFlowType } & PreAuthCodeFlowOptions &
@@ -185,7 +185,7 @@ export class OpenId4VcClientService {
     const base64Url = TypedArrayEncoder.toBase64URL(codeVerifierSha256)
 
     this.logger.debug('Converted code_verifier to code_challenge', {
-      codeVerifier: options.codeVerifier,
+      codeVerifier: codeVerifier,
       sha256: codeVerifierSha256.toString(),
       base64Url: base64Url,
     })
@@ -208,9 +208,9 @@ export class OpenId4VcClientService {
     const credentialFormat = 'ldp_vc'
 
     let flowType: AuthzFlowType
-    if (options.flowType === AuthFlowType.AUTHORIZATION_CODE_FLOW) {
+    if (options.flowType === AuthFlowType.AuthorizationCodeFlow) {
       flowType = AuthzFlowType.AUTHORIZATION_CODE_FLOW
-    } else if (options.flowType === AuthFlowType.PRE_AUTHORIZED_CODE_FLOW) {
+    } else if (options.flowType === AuthFlowType.PreAuthorizedCodeFlow) {
       flowType = AuthzFlowType.PRE_AUTHORIZED_CODE_FLOW
     } else {
       throw new AriesFrameworkError(
@@ -227,7 +227,7 @@ export class OpenId4VcClientService {
 
     let accessToken: AccessTokenResponse
 
-    if (options.flowType === AuthFlowType.AUTHORIZATION_CODE_FLOW) {
+    if (options.flowType === AuthFlowType.AuthorizationCodeFlow) {
       if (!options.authorizationCode)
         throw new AriesFrameworkError(
           `The 'authorizationCode' parameter is required when 'flowType' is ${options.flowType}`
@@ -302,7 +302,7 @@ export class OpenId4VcClientService {
     // verify the signature
     const result = await this.w3cCredentialService.verifyCredential(agentContext, {
       credential,
-      verifyRevocationState: options.checkRevocationState,
+      verifyRevocationState: options.verifyRevocationState,
     })
 
     if (result && !result.verified) {
--- a/packages/openid4vc-client/tests/openid4vc-client.e2e.test.ts
+++ b/packages/openid4vc-client/tests/openid4vc-client.e2e.test.ts
@@ -86,7 +86,7 @@ describe('OpenId4VcClient', () => {
       const w3cCredentialRecord = await agent.modules.openId4VcClient.requestCredentialUsingPreAuthorizedCode({
         issuerUri,
         kid,
-        checkRevocationState: false,
+        verifyRevocationState: false,
       })
 
       expect(w3cCredentialRecord).toBeInstanceOf(W3cCredentialRecord)
@@ -220,7 +220,7 @@ describe('OpenId4VcClient', () => {
         clientId: clientId,
         authorizationCode: 'test-code',
         codeVerifier: codeVerifier,
-        checkRevocationState: false,
+        verifyRevocationState: false,
         kid: kid,
         issuerUri: initiationUri,
         redirectUri: redirectUri,


packages/openid4vc-client/src/OpenId4VcClientApi.ts
```diff
@@ -32,4 +33,16 @@ export class OpenId4VcClientApi {
       checkRevocationState: checkRevocationState,
     })
   }
+
+  public async generateCodeVerifier(): Promise<string> {
```
TimoGlastra, line 37: This method gives me the impression I can make a code for a verifier. What's purpose of this method?

packages/openid4vc-client/src/OpenId4VcClientService.ts
```diff
@@ -3,34 +3,60 @@ import type { EndpointMetadata, Jwt } from '@sphereon/openid4vci-client'
...
 } from '@aries-framework/core'
 import {
   Alg,
   AuthzFlowType,
+  CodeChallengeMethod,
   CredentialRequestClientBuilder,
   OpenID4VCIClient,
   ProofOfPossessionBuilder,
 } from '@sphereon/openid4vci-client'
+import { randomStringForEntropy } from '@stablelib/random'
```
TimoGlastra, line 30: We should probably add a getRandomBytes method to the wallet so we can replace this. It's fine for now thiugh

packages/openid4vc-client/src/OpenId4VcClientService.ts
```diff
@@ -134,6 +160,42 @@ export class OpenId4VcClientService {
...
+    const client = await OpenID4VCIClient.initiateFromURI({
+      issuanceInitiationURI: options.initiationUri,
+      flowType: AuthzFlowType.AUTHORIZATION_CODE_FLOW,
+    })
+
+    const hashed = sha256(TypedArrayEncoder.fromString(options.codeVerifier))
+    const base64Url = TypedArrayEncoder.toBase64URL(Buffer.from(hashed))
+
+    this.logger.debug('Converted code_verifier to code_challenge', {
+      codeVerifier: options.codeVerifier.toString(),
```
TimoGlastra, line 185: codeVerifier already is string?

packages/openid4vc-client/src/OpenId4VcClientService.ts
```diff
@@ -134,6 +160,42 @@ export class OpenId4VcClientService {
...
+      sha256: hashed.toString(),
```
TimoGlastra, line 186: What encoding must the hash have? Utf8?

packages/openid4vc-client/src/OpenId4VcClientService.ts
```diff
@@ -134,6 +160,42 @@ export class OpenId4VcClientService {
...
+      base64Url: base64Url,
+    })
+
+    return client.createAuthorizationRequestUrl({
+      clientId: options.clientId,
+      codeChallengeMethod: CodeChallengeMethod.SHA256,
```
TimoGlastra, line 192: Could the authorization server support other algorithms than sha256? Or is this generally the one to support?

packages/openid4vc-client/src/OpenId4VcClientService.ts
```diff
@@ -232,4 +294,89 @@ export class OpenId4VcClientService {
...
+
+    const accessToken = await client.acquireAccessToken({
+      clientId: options.clientId,
+      code: options.code,
+      codeVerifier: options.codeVerifier,
+      redirectUri: options.redirectUri,
+    })
+
+    const serverMetadata = await client.retrieveServerMetadata()
+    // @ts-ignore
```
TimoGlastra, line 315: We should avoid ts ignore as much as possible. I think it would be better to add a cast one time and cast it to the right type once as this disables all type checks for yhay line.

Same for the scope having the wrong type

packages/openid4vc-client/package.json
```diff
@@ -25,7 +25,9 @@
   },
   "dependencies": {
     "@aries-framework/core": "0.3.3",
-    "@sphereon/openid4vci-client": "^0.3.6"
+    "@sphereon/openid4vci-client": "^0.4.0",
+    "@stablelib/random": "^1.0.2",
+    "@stablelib/sha256": "^1.0.1"
```
TimoGlastra, line 30:
```suggestion
```

See comment about hasher

packages/openid4vc-client/src/OpenId4VcClientService.ts
```diff
@@ -232,4 +294,89 @@ export class OpenId4VcClientService {
 
     return storedCredential
   }
+
+  public async requestCredential(agentContext: AgentContext, options: RequestCredentialOptions) {
```
TimoGlastra, line 298: This method is quite similar to the request preAuthorized method. Do you think we could create a single method to request a credential that supports both flows (one has authorization code, other pre-authorized code), or do you think it would be good to keep them separate? If so it would be good to at least reduce duplication a bit to prevent the implementations from getting out of sync

packages/openid4vc-client/src/OpenId4VcClientService.ts
```diff
@@ -232,4 +294,89 @@ export class OpenId4VcClientService {
...
+      metadata: serverMetadata,
+    })
+      .withTokenFromResponse(accessToken)
+      .build()
+
+    const credentialResponse = await credentialRequestClient.acquireCredentialsUsingProof({
+      proofInput,
+      // @ts-ignore
+      credentialType: accessToken.scope,
+      format: 'ldp_vc', // Allows us to override the format
```
TimoGlastra, line 351: We define credential format at the top if this method. Should we reuse it here?

packages/openid4vc-client/src/OpenId4VcClientService.ts
```diff
@@ -232,4 +294,89 @@ export class OpenId4VcClientService {
...
+    const credentialRequestClient = CredentialRequestClientBuilder.fromIssuanceInitiationURI({
+      uri: options.issuerUri,
+      metadata: serverMetadata,
+    })
+      .withTokenFromResponse(accessToken)
+      .build()
+
+    const credentialResponse = await credentialRequestClient.acquireCredentialsUsingProof({
+      proofInput,
+      // @ts-ignore
```
TimoGlastra, line 349: Wasn't this fixed with the new release?

packages/openid4vc-client/tests/openid4vc-client.e2e.test.ts
```diff
@@ -98,6 +98,137 @@ describe('OpenId4VcClient', () => {
...
+      expect(typeof codeVerifier).toBe('string')
+
+      // length restrictions are specified here https://www.rfc-editor.org/rfc/rfc7636#section-4.1
+      expect(codeVerifier.length).toBeGreaterThanOrEqual(43)
+      expect(codeVerifier.length).toBeLessThanOrEqual(128)
+    })
+
+    it('should generate a valid authorization url', async () => {
+      const clientId = 'test-client'
+      const codeVerifier = await agent.modules.openId4VcClient.generateCodeVerifier()
```
TimoGlastra, line 146: Do we need to expose this as a separate method? Do you think we could generate it in the generateAurhorizationUrl and just return it as part of the method?

For all other methods in AFJ we also just generate the nonce

packages/openid4vc-client/tests/openid4vc-client.e2e.test.ts
```diff
@@ -98,6 +98,137 @@ describe('OpenId4VcClient', () => {
...
+        method: 'key',
+        options: {
+          keyType: KeyType.Ed25519,
+        },
+        secret: {
+          privateKey: TypedArrayEncoder.fromString('96213c3d7fc8d4d6754c7a0fd969598e'),
+        },
+      })
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      const keyInstance = didKeyToInstanceOfKey(did.didState.did!)
```
TimoGlastra, line 203:
```suggestion
      const keyInstance = didKeyToInstanceOfKey(did.didState.did as string)
```

packages/openid4vc-client/tests/openid4vc-client.e2e.test.ts
```diff
@@ -98,6 +98,137 @@ describe('OpenId4VcClient', () => {
...
+
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      const kid = `${did.didState.did!}#${keyInstance.fingerprint}`
```
TimoGlastra, line 206:
```suggestion
      const kid = `${did.didState.did as string}#${keyInstance.fingerprint}`
```

packages/openid4vc-client/src/OpenId4VcClientService.ts
```diff
@@ -232,4 +294,89 @@ export class OpenId4VcClientService {
...
+      throw new AriesFrameworkError('Did not receive a successful credential response')
+    }
+
+    const credential = JsonTransformer.fromJSON(credentialResponse.successBody.credential, W3cVerifiableCredential)
+
+    // verify the signature
+    const result = await this.w3cCredentialService.verifyCredential(
+      agentContext,
+      { credential },
+      options.checkRevocationState
```
TimoGlastra, line 366: You'd move this to the options object right?

TimoGlastra (changes requested):
Nice! Mostly some small stylistic suggestions

TimoGlastra:
Seems DCO is failing BTW

packages/openid4vc-client/src/OpenId4VcClientService.ts
karimStekelenburg: Yes! Good catch

packages/openid4vc-client/src/OpenId4VcClientService.ts
karimStekelenburg: In theory the code challenge method can also be set to TEXT where a plain text string is used instead. This basically defeats the purpose and according to my understanding it's not recommended or common. So far I've not come across auth servers that support it.

packages/openid4vc-client/src/OpenId4VcClientService.ts
karimStekelenburg: The hash is base64 URL encoded (see `codeVerifierSha256`) variable. Renamed some variables for clarity.

packages/openid4vc-client/src/OpenId4VcClientService.ts
```diff
@@ -3,34 +3,60 @@ import type { EndpointMetadata, Jwt } from '@sphereon/openid4vci-client'
...
+  codeVerifier: string
+  issuerUri: string
+  kid: string
+  redirectUri: string
+  checkRevocationState: boolean
+}
+
+// The code_challenge_method is omitted here
+// because we assume it will always be SHA256
+// as clear text code_challenges are unsafe
```
karimStekelenburg, line 51: Ah lol, there was a comment about it here @TimoGlastra

packages/openid4vc-client/tests/openid4vc-client.e2e.test.ts
karimStekelenburg: Wow, much better!

packages/openid4vc-client/tests/openid4vc-client.e2e.test.ts
karimStekelenburg: The caller needs to pass the same `codeVerifier` value when generating the url (`generateAuthorizationUrl`) as when requesting the credential (`requestCredential`) for the flow to succeed. This is to prevent others from intercepting the authorization code and pretending they're us. If we don't expose this method, we'd have to store and manage it internally. That seemed a bit extra to me :)

packages/openid4vc-client/src/OpenId4VcClientService.ts
karimStekelenburg: I've tried that method that Indy provides (genrateRandomSeed or something), but that one didn't meet the requirements for this use case (I forgot what the exact issue was).

packages/openid4vc-client/src/OpenId4VcClientService.ts
TimoGlastra: Does to string automatically encode it as base64?

karimStekelenburg:
Thanks, @TimoGlastra. I've processed your comments.

You were right, the two implementations had huge overlaps, and I've merged them together in the service. I did create two separate methods in the API. I figured that is a clearer API than optional values depending on the flow type.

Regarding the `code_verifier` stuff, does the following description resolve your questions?

> In OpenID Connect (OIDC) with Proof Key for Code Exchange (PKCE), the code_verifier is a random string used as a cryptographic challenge when requesting an authorization code from the authorization server.
> 
> PKCE is a security mechanism designed to prevent certain types of attacks that can occur when using the Authorization Code flow in OAuth 2.0. Specifically, PKCE is intended to protect against code injection attacks where an attacker intercepts the authorization code and uses it to obtain access tokens without the user's knowledge.
> 
> The PKCE flow involves creating a random code_verifier string, calculating a hash of this string, and then sending both the code_verifier and the hash (called the code_challenge) to the authorization server when requesting an authorization code. Later, when the client exchanges the authorization code for an access token, it must again send the code_verifier to the authorization server, which will use it to verify that the code_challenge matches the original code_verifier.
> 
> By requiring the code_verifier in addition to the code_challenge, PKCE ensures that only the client that initiated the authorization request can obtain an access token, even if an attacker is able to intercept the authorization code.

packages/openid4vc-client/src/OpenId4VcClientService.ts
TimoGlastra: I'm always a bit worried with the toString method as it's a method available on all objects and will ofyen give you `[Object object]` or something

packages/openid4vc-client/tests/openid4vc-client.e2e.test.ts
TimoGlastra: We could just generate it and return it in the generateAuthorizationUrl?

packages/core/src/modules/vc/W3cCredentialService.ts
```diff
@@ -103,17 +103,18 @@ export class W3cCredentialService {
    */
   public async verifyCredential(
     agentContext: AgentContext,
-    options: VerifyCredentialOptions,
-    verifyRevocationState = true
+    options: VerifyCredentialOptions
   ): Promise<W3cVerifyCredentialResult> {
+    options.verifyRevocationState = options.verifyRevocationState ?? true
```
TimoGlastra, line 108: It's usually bad pratice to mutate input objects. Maybe we assign it as separate variable instead of mutating the input object?

packages/openid4vc-client/src/OpenId4VcClientApi.ts
```diff
@@ -23,13 +22,29 @@ export class OpenId4VcClientApi {
     this.openId4VcClientService = openId4VcClientService
   }
 
-  public async requestCredentialPreAuthorized(options: PreAuthorizedOptions): Promise<W3cCredentialRecord> {
+  public async requestCredentialUsingPreAuthorizedCode(options: PreAuthCodeFlowOptions): Promise<W3cCredentialRecord> {
+    // set defaults
+    const checkRevocationState = options.checkRevocationState ?? true
```
TimoGlastra, line 27: Is there a reason it's called checkRevocationState here and verifyRevocationState in the w3c service?

packages/openid4vc-client/src/OpenId4VcClientService.ts
```diff
@@ -1,36 +1,65 @@
...
 
+export interface AuthCodeFlowOptions extends PreAuthCodeFlowOptions {
+  clientId: string
+  authorizationCode: string
+  codeVerifier: string
+  redirectUri: string
+}
+
+export enum AuthFlowType {
+  AUTHORIZATION_CODE_FLOW = 'Authorization Code Flow',
```
TimoGlastra, line 46: Nit: but we usually use pascal case for enum keys and values, and don't use spaces etc in the values. 

Would be good to keep this consistent with AFJ more than with the sphereon lib

TimoGlastra (approved):
Some final nits for consistency, but LGTM!!

TimoGlastra:
Nice !!

PR #1385, opened 2023-03-17
feat(anoncreds): add AnonCreds format services

genaris:
Add Credential and Proof format services according to [Aries RFC 0771](https://github.com/hyperledger/aries-rfcs/tree/main/features/0771-anoncreds-attachments).

Essentially the same as Legacy Indy counterparts but using the new identifiers, using entropy instead of prover_did for Credential Request and not restricting identifiers to legacy Indy format, so any AnonCreds method can be used with them.


TimoGlastra (approved):
Nice, great work!!! 

Would be good if we could add one e2e test using the public api leveraging the anoncreds formats, but that can be added in a follow up pr as well!

PR #1387, opened 2023-03-18
feat(anoncreds): store method name in records

### LINKED ISSUE #1265: Store anoncreds method identifier in the AnonCreds records
This will allow us to find e.g. all schemas created for a specific ledger. Currently we only can query by the full issuerId

TimoGlastra:
Fixes #1265 

Adds a `methodName` property to the schema, cred def and anoncreds credential records so we can query all those by the anoncreds method it is linked to.

Currently I've used `indy` for indy anoncreds method (both legacy and non-legacy).

Need to incorporate this with the `getCredentials` options from #1386 (can be in a follow up pr or this PR after that one is merged)


packages/anoncreds/src/AnonCredsApi.ts
```diff
@@ -179,6 +181,10 @@ export class AnonCredsApi {
     }
   }
 
+  public async getCreatedSchemas(query: SimpleQuery<AnonCredsSchemaRecord>) {
```
genaris, line 184: I think this (and the credential definition counterpart) are some nice hidden gems of this PR :laughing:

genaris (approved):
I'm OK with the changes but would like to do a warning about that there is a distinction in [AnonCreds Method Registry](https://hyperledger.github.io/anoncreds-methods-registry/) between did:indy and 'Indy Legacy' methods.

I know there has been a big effort within AFJ to do a migration from all unqualified to qualified but I'm not completely sure if there could be a case where this is not desirable and could be preferable to keep things separated.

PR #1388, opened 2023-03-18
chore!: update dependencies

### LINKED ISSUE #1213: Update packages

### LINKED ISSUE #1267: Update jest to latest version

TimoGlastra:
Fixes #1213 and Fixes #1267 

Also removed support for node 17 as it's reached EOL (not a LTS release), and it's not compatible with all pacakges. In addition I've removed support for node 14 as not all packages support it, and it'll reach EOL this month. I think it's fine to remove it for 0.4.0, and focus on Node 16 and 18

BREAKING CHANGE:
Node.JS 14 and 17 have been dropped. Node.JS 17 has reached it's end of life, and Node.JS 14 will do so within a month. Support for Node.JS 20 will be added once it's released.


demo/package.json
```diff
@@ -28,10 +28,12 @@
...
     "@types/figlet": "^1.5.4",
     "@types/indy-sdk": "^1.16.26",
-    "@types/inquirer": "^8.1.3",
+    "@types/inquirer": "^8.2.6",
     "clear": "^0.1.0",
-    "commander": "^8.3.0",
     "figlet": "^1.5.2",
     "ts-node": "^10.4.0"
+  },
+  "dependencies": {
```
genaris, line 36: This is duplicating `dependencies` key

genaris (approved):
Just a few details, but good to see these updates. Dropping support for Node 14 and 17 will also make CI easier to pass :laughing: .

Do you think we should start adding Node 19 to the suite?

PR #1389, opened 2023-03-18
refactor(cache)!: change default cache to in memory

TimoGlastra:
Changes default cache implementation to an in memory cache. This simplifies the caching, and makes sure it works for all cases (nodejs/react-native, but also mulit-tenancy).

BREAKING CHANGE:
The default cache implementation has been changed from `SingleContextStorageLruCache` to `InMemoryLruCache`. In React Native environments it is still recommended to use the `SingleContextStorageLruCache` cache implementation, but due to it's dependence on the wallet storage, it is not recommend for server side or multi-tenancy deployments. A custom caching interface can be implemented in those cases.

### DIFF
--- a/packages/core/src/modules/cache/CacheModule.ts
+++ b/packages/core/src/modules/cache/CacheModule.ts
@@ -3,6 +3,7 @@ import type { DependencyManager, Module } from '../../plugins'
 import type { Optional } from '../../utils'
 
 import { CacheModuleConfig } from './CacheModuleConfig'
+import { InMemoryLruCache } from './InMemoryLruCache'
 import { SingleContextLruCacheRepository } from './singleContextLruCache/SingleContextLruCacheRepository'
 import { SingleContextStorageLruCache } from './singleContextLruCache/SingleContextStorageLruCache'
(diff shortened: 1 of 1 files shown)


genaris:
I don't know the details about it but an in memory cache makes me think that, unless a server has 100% of uptime, it will be lost. Would that be something that we should leverage using by default? 

I totally agree that it's still better than `SingleContextStorageLruCache` for multi-tenancy and probably also for server-side single-tenancy agents, but given that AFJ is mostly used in Mobile Agents where, as you said, it's preferable to use the other one, I'm not convinced that in memory should be set to default in 0.4.0.

The main concern about `SingleContextStorageLruCache` is that it does not work in multi-tenancy, right? In such case couldn't be an error thrown at agent initialization when attempting to use it for multi-tenancy?

As I said previously, I don't know so much about the context so it's very likely that my reasoning is wrong, but that's what I'm thinking based on some things that we recently defaulted to the 'right way' and caused some troubles to current AFJ audience (e.g. setting the did:key flag to true in 0.3.0).

Issue #1391, opened 2023-03-19
Always store qualified identifiers for AnonCreds objects, still supporting legacy indy identifiers in exchange


TimoGlastra:
To make the transition to qualified identifiers as easy as possible, I think from now on we should start storing the anoncreds credential records with qualified identifiers. This means we know which network it is associated with. We should add tags that allow us to query it based on the legacy/unqualified identifiers (only if the qualified identifiers are `did:indy` identifiers).

- [ ] Add the following tags to the records if the identifiers for the object are `did:indy` identifiers
  - Credential: `legacySchemaIssuerId`, `legacyIssuerId`, `legacyCredentialDefinitionId`, `legacyRevocationRegistryId`, `legacySchemaId`
  - Schema: `legacySchemaId`, `legacyIssuerId`
   - Credential Definition: `legacyIssuerId`, `legacySchemaId`, `legacyCredentialDefinitionId`
  - This allows us to query the records using legacy identifiers
- [ ] In the AnonCredsRsHolderService store the credential with qualified schema and credential definition id. We may have to tweaks the interfaces a bit to know the qualified identifiers, or we just always require it to be passed with qualified identifiers (and do the transformation in the credential/proof format as we've fetched the objects from the registry there).
- [ ] When using `did:indy` identifiers for schemas and credential definitions, we should also store the unqualified/legacy tags.
- [ ] In the AnonCredsRsHolder service when querying credentials using legacy identifiers, make sure to query using the `legacyXXX` tags instead of the normal tags.
  - e.g. `getCredential`, `getCredentials`, `getCredentialsForProofRequest`
  - Should we also return the credentials using the legacy identifiers? probably yes
- [ ] In the AnonCredsRS services when creating proofs, creating credentials, creating requests, creating offers, etc.. make sure to correctly handle the legacy vs qualified identifiers taking into account input could be qualified / unqualified, but the storage is always qualified (need to query using the legacy tags), transform identifiers before returning etc...

We could also say the AnonCreds RS services only support qualified identifiers, and the mapping needs to be done on a higher level, but that increases complexity I think.

Currently issuing using unqualified identifiers is broken using AnonCreds RS and Aries Askar because we store the schema and credential definition using qualified identifiers.

## Unresolved questions

- Indy SDK will still store all credentials still using unqualified identifiers. For schemas and credential definitions we can use our custom schema / credential definition records, but we don't have the qualified identifiers for the credentials stored using indy-sdk.
[...]

PR #1392, opened 2023-03-19
feat: 0.4.0 migration script

TimoGlastra:
Adds migration script for 0.4.0 covering most of the changes between 0.3.1 and 0.4.0. There's some issues to resolve with how we are going to handle legacy identifiers vs qualified identifiers in AnonCreds (see #1391). `did:sov` dids are all transformed to `did:indy` dids already as we store the qualified indy did. Same for AnonCreds schema and cred def records, so it's mainly about anoncreds credential records that need some work.

Need to extend the migration docs a bit more, to say you need to add the anoncreds module to update the anoncreds related records, and that any dids created using `agent.ledger.` should be imported manually into the agent as we don't have the `DidRecord` for them.


packages/anoncreds/src/updates/0.3.1-0.4/credentialExchangeRecord.ts
```diff
@@ -0,0 +1,148 @@
...
+    agent.config.logger.debug(
+      `Successfully migrated credential exchange record with id ${credentialRecord.id} to storage version 0.4`
+    )
+  }
+}
+
+/**
+ * Migrates the indy credential record binding to anoncreds credential record binding.
+ *
+ * The following 0.3.1 creedential record structure (unrelated keys omitted):
```
blu3beri, line 42:
```suggestion
 * The following 0.3.1 credential record structure (unrelated keys omitted):
```
```suggestion
 * The following 0.3.1 creedential record structure (unrelated keys omitted):
```

packages/anoncreds/src/updates/0.3.1-0.4/credentialExchangeRecord.ts
```diff
@@ -0,0 +1,148 @@
...
+ */
+export function migrateIndyCredentialMetadataToAnonCredsMetadata<Agent extends BaseAgent>(
+  agent: Agent,
+  credentialRecord: CredentialExchangeRecord
+) {
+  agent.config.logger.debug(
+    `Migrating credential record with id ${credentialRecord.id} to anoncreds metadata for version 0.4`
+  )
+
+  const INDY_CREDENTIAL_REQUEST_METADATA = '_internal/indyRequest'
```
blu3beri, line 126: non-global constants can be camelCase here, right?

blu3beri (approved):
Awesome! Looks extensively tested, very nice!

Do you think we should create some tests that do all the migrations in different order? Maybe they change some state in a weird way which breaks it (and seemingly unrevertable).

packages/anoncreds/src/updates/0.3.1-0.4/credentialExchangeRecord.ts
```diff
@@ -0,0 +1,148 @@
...
+  const INDY_CREDENTIAL_METADATA = '_internal/indyCredential'
+
+  const ANONCREDS_CREDENTIAL_REQUEST_METADATA = '_anoncreds/credentialRequest'
+  const ANONCREDS_CREDENTIAL_METADATA = '_anoncreds/credential'
+
+  const indyCredentialRequestMetadata = credentialRecord.metadata.get(INDY_CREDENTIAL_REQUEST_METADATA)
+  if (indyCredentialRequestMetadata) {
+    // TODO: we if we choose to rename master secret to link secret in anoncreds-rs we should also rename it in the request
```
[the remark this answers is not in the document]
TimoGlastra: When you call create credential request in indy-sdk / anoncreds rs) it will return the following metadata:

```
{
          "master_secret_blinding_data": {
            "v_prime": "6088566065720309491695644944398283228337587174153857313170975821102428665682789111613194763354086540665993822078019981371868225077833338619179176775427438467982451441607103798898879602785159234518625137830139620180247716943526165654371269235270542103763086097868993123576876140373079243750364373248313759006451117374448224809216784667062369066076812328680472952148248732117690061334364498707450807760707599232005951883007442927332478453073050250159545354197772368724822531644722135760544102661829321297308144745035201971564171469931191452967102169235498946760810509797149446495254099095221645804379785022515460071863075055785600423275733199",
            "vr_prime": null
          },
          "nonce": "131502096406868204437821",
          "master_secret_name": "walletId28c602347-3f6e-429f-93cd-d5aa7856ef3f"
        }
```

If we move to using link secret in AnonCreds RS, we should update all metadata and transform the master_secret keys to link_secret keys in the Indy SDK holder service in AFJ.

Did you also update these fields in your link secret pr in anoncreds rs?

TimoGlastra:
> Do you think we should create some tests that do all the migrations in different order? Maybe they change some state in a weird way which breaks it (and seemingly unrevertable).

Do you mean the indy sdk to askar update script and the normal update assistant? If so, yes that'd probably be a good idea!

PR #1398, opened 2023-03-22
feat(core): add W3cCredentialApi

karimStekelenburg:
This PR adds a W3cCredentialApi class with methods for storing, removing, getting, and finding credentials and credential records.

Funded by the Ontario Government


packages/core/src/modules/vc/W3cCredentialApi.ts
```diff
@@ -0,0 +1,46 @@
...
+import type { StoreCredentialOptions } from './models/W3cCredentialServiceOptions'
+import type { W3cCredentialRecord } from './repository'
+import type { Query } from '../../storage/StorageService'
+
+import { AgentContext } from '../../agent'
+import { injectable } from '../../plugins'
+
+import { W3cCredentialService } from './W3cCredentialService'
+
+@injectable()
```
TimoGlastra, line 11: do you think some light tests can be added for the public api? Maybe just one test that stores a credential, fetches it by id, fetches it using the find by id, deletes it and then expects getAllRecords to return an empty array or something.

TimoGlastra:
Also, should we call it `W3cCredentialsApi` (plural). as we have `CredentialsApi`, `ProofsApi`, `AnonCredsApi`, etc..

packages/core/src/modules/vc/W3cCredentialApi.ts
karimStekelenburg: Done!

karimStekelenburg:
> Also, should we call it `W3cCredentialsApi` (plural). as we have `CredentialsApi`, `ProofsApi`, `AnonCreds`

I noticed the service was renamed to `W3cVcService`, so I did the same for the API.

karimStekelenburg:
> In main it's still W3cCredentialService? https://github.com/hyperledger/aries-framework-javascript/blob/main/packages/core/src/modules/vc/W3cCredentialService.ts
> 
> I think I like W3cCredentialsApi more I think, and it aligns with the credentials module naming more maybe?

Yes but it looks like you renamed the `module` to `W3cVcModule`? I don't really mind, but I do think it's nice to stay consistent across the services and modules :)

karimStekelenburg:
> Ah yes, the module is called `W3cVcModule`. What do you prefer? I think we can change the name of the module if we want

Haha, this is an interesting one. I like `W3cVc` because it's nice and short. However, the current `CredentialsModule` implies that the `CredentialsModule` deals with non-verifiable credentials.

I'd suggest we rename everything to `W3cCredentialsFoo` for now and change it later if needed. That way we at least stay consistent.

> **Note:** I do think this rename should make it plural (e.g. W3cCredential**s**Module), as this is also the case for the `ProofsModule` and `CredentialsModule`.

packages/bbs-signatures/tests/bbs-signatures.e2e.test.ts
```diff
@@ -1,4 +1,4 @@
-import type { W3cCredentialRepository } from '../../core/src/modules/vc/repository'
+import type { W3cCredentialsRepository } from '../../core/src/modules/vc/repository'
```
TimoGlastra, line 1: We keep the services and repositories singular (It's also CredentialRepository). It's a bit inconsistent, but we've had discussions at the time and that's what we landed on. This would be a bigger thing to change instead of `VC` vc `Credentials`

TimoGlastra (approved):
It seems CI is failing.

PR #1406, opened 2023-03-25
test(indy-vdr): add delay after DID creation

genaris:
In Indy VDR tests, after DID creation we are attempting to resolve the recently created DID immediately, which does not always work properly in CI. Here we add the 1000 ms delay that we also do in other Indy registrar tests.

### DIFF
--- a/packages/indy-vdr/tests/helpers.ts
+++ b/packages/indy-vdr/tests/helpers.ts
@@ -4,6 +4,7 @@ import type { Agent } from '@aries-framework/core'
 import { DidCommV1Service, DidCommV2Service, DidDocumentService, KeyType } from '@aries-framework/core'
 import { indyVdr } from '@hyperledger/indy-vdr-nodejs'
 
+import { sleep } from '../../core/src/utils/sleep'
 import { genesisTransactions } from '../../core/tests/helpers'
 import { IndyVdrModuleConfig } from '../src/IndyVdrModuleConfig'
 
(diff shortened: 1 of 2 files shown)


packages/indy-vdr/tests/helpers.ts
```diff
@@ -60,5 +61,8 @@ export async function createDidOnLedger(agent: Agent, submitterDid: string) {
     )
   }
 
+  // Wait some time pass to let ledger settle the object
+  await sleep(1000)
```
blu3beri, line 65: Wouldn't it be better to see if we can resolve it? This would still make it non-deterministic albeit less.

packages/indy-vdr/tests/indy-vdr-did-registrar.e2e.test.ts
```diff
@@ -149,6 +153,8 @@ describe('Indy VDR Indy Did Registrar', () => {
       },
     })
 
+    await new Promise((res) => setTimeout(res, 1000))
```
blu3beri, line 156: Can this be `sleep` if we decide to keep the sleep?

packages/indy-vdr/tests/helpers.ts
genaris: Could you elaborate a bit more on this suggestion? Do you mean to do a call to dids.resolve() before leaving this method?

TimoGlastra (approved):

PR #1408, opened 2023-03-28
feat: allow sending problem report when declining a proof request

Artemkaaas:
Updated `declineRequest` method to acces optional `sendProblemReport` parameter indicating that Problem Report message also should be sent to Verifier.
Also updated `sendProblemReport` method to support connectionless proof request case.


packages/core/src/modules/proofs/ProofsApi.ts
```diff
@@ -368,13 +368,17 @@ export class ProofsApi<PPs extends ProofProtocol[]> implements ProofsApi<PPs> {
     }
   }
 
-  public async declineRequest(proofRecordId: string): Promise<ProofExchangeRecord> {
+  public async declineRequest(proofRecordId: string, sendProblemReport?: boolean): Promise<ProofExchangeRecord> {
```
blu3beri, line 371: In general we tend to create objects for functionality like named parameters. Here for example, calling `declineProofRequest('abc-id', true)` would be quite vague. 

If we include this into 0.4.0 we can make the breaking change and the function would accept 1 parameter `DeclineRequestOptions` which is `{proofRecordId: string, sendProblemRequest?: boolean}`

packages/core/src/modules/proofs/ProofsApi.ts
```diff
@@ -555,29 +559,59 @@ export class ProofsApi<PPs extends ProofProtocol[]> implements ProofsApi<PPs> {
...
-    })
+    if (proofRecord.connectionId) {
+      const connectionRecord = await this.connectionService.getById(this.agentContext, proofRecord.connectionId)
 
-    await this.messageSender.sendMessage(outboundMessageContext)
-    return proofRecord
+      // Assert
+      connectionRecord.assertReady()
+
+      const outboundMessageContext = new OutboundMessageContext(problemReport, {
```
blu3beri, line 578: @TimoGlastra is this the correct place to do this logic? If possbile I think it would be good to defer this to another method (creating the service decorator and adding it as `serviceParams`).

packages/core/src/modules/proofs/ProofsApi.ts
```diff
@@ -555,29 +559,59 @@ export class ProofsApi<PPs extends ProofProtocol[]> implements ProofsApi<PPs> {
...
+      })
+      const recipientService = requestMessage.service
+
+      await this.messageSender.sendMessageToService(
+        new OutboundMessageContext(problemReport, {
+          agentContext: this.agentContext,
+          serviceParams: {
+            service: recipientService.resolvedDidCommService,
+            senderKey: ourService.resolvedDidCommService.recipientKeys[0],
+            returnRoute: options.useReturnRoute ?? true, // defaults to true if missing
```
blu3beri, line 602: I think it would be good to define the default value either inside the function signature or at the first line. This is a bit hidden.

blu3beri (approved):
Couple small points, but non-blocking for me.

packages/core/src/modules/proofs/ProofsApi.ts
Artemkaaas: The logical changes for `send Problem Report` function (handle connectionless request) repeats logic from `accept Request` method.

packages/core/src/modules/proofs/ProofsApi.ts
Artemkaaas: I agree with you that accepting single parameter is better but I afraid to make breaking change.

TimoGlastra: We're just about to release 0.4.0 with a lot of breaking changes to the proofs api. Now is the time!

packages/core/src/modules/proofs/ProofsApi.ts
TimoGlastra: Nit sure if we need it. We don't expect a responsw to the problem report right? I'd just set it to false / leave it out

packages/core/src/modules/proofs/ProofsApi.ts
Artemkaaas: I updated the method parameters to accept single `DeclineProofRequestOptions` object

packages/core/src/modules/proofs/ProofsApi.ts
Artemkaaas: Removed setting of `returnRoute` and `useReturnRoute` field in the parameters

packages/core/src/modules/proofs/ProofsApi.ts
```diff
@@ -555,29 +560,58 @@ export class ProofsApi<PPs extends ProofProtocol[]> implements ProofsApi<PPs> {
...
+
+      const outboundMessageContext = new OutboundMessageContext(problemReport, {
+        agentContext: this.agentContext,
+        connection: connectionRecord,
+        associatedRecord: proofRecord,
+      })
+
+      await this.messageSender.sendMessage(outboundMessageContext)
+      return proofRecord
+    } else if (requestMessage?.service) {
```
TimoGlastra, line 587: What if we want to send a problem report as the verifier? We'd send the message to ourselves in that case (as we're the ones that created the request). I think we shouldn't assume the role here, and instead infer it based on the state

TimoGlastra (changes requested):
There's some issues with the sendProblemReport method still

packages/core/src/modules/proofs/ProofsApi.ts
Artemkaaas: Added check that state is `ProofState.RequestReceived`

TimoGlastra (approved):

PR #1413, opened 2023-03-30
feat: add message pickup module

genaris:
This PR moves Message Pickup specific features into a separate module, rather than putting all together into Mediator and Recipient modules.

The idea behind this (besides making mediator/routing module more lightweight) is to start adding more advanced features to Message Pickup, for both recipient (e.g. be able to manage multiple message pickup loops) and message holder roles (e.g. support message pickup sessions in order to determine what to do with forwarded messages depending on Live mode status).

For the moment, it attempts to not introduce any breaking change at API level (other than renaming some message and handler classes that were previously exposed so somebody could have been referencing them externally).

`MessageRepository` API remains the same, although now it can be injected by using the `MessagePickupModuleConfig`, which would be the most revolutionary (?) addition from this PR from API perspective.

```typescript
new Agent({
      config,
      modules: {
         /* ... */
         messagePickup: new MessagePickupModule({ messageRepository: new CustomMessageRepository() }),
      }
})
```


packages/core/src/modules/message-pìckup/MessagePickupModuleConfig.ts
```diff
@@ -0,0 +1,57 @@
...
+   * [V1MessagePickupProtocol, V2MessagePickupProtocol]
+   * ```
+   */
+  protocols: MessagePickupProtocols
+
+  /**
+   * Allows to specify a custom pickup message queue. It defaults to an in-memory repository
+   *
+   */
+  messageRepository?: MessageRepository
```
TimoGlastra, line 33: Curious what the line is between mediator vs message pickup. Is the mediator just about routing, and is message pickup module about queuing messages? As you mentioned before, allowing messages to be queued even without mediator (so just e.g. issuer that queues messages for the holder directly) should be possible, in which case this approach makes sense.

Does the message pickup module still support `implicit` pickup? And if so, what's the protocolVersion in that case?!

packages/core/src/modules/message-pìckup/MessagePickupApi.ts
```diff
@@ -0,0 +1,96 @@
...
+    await messageRepository.add(connectionRecord.id, options.message)
+  }
+
+  /**
+   * Pickup queued messages from a message holder. It attempts to retrieve all current messages from the
+   * queue, receiving up to `batchSize` messages per batch retrieval.
+   *
+   * @param options connectionId, protocol version to use and batch size
+   */
+  public async pickupMessages(options: PickupMessagesOptions<MPPs>): Promise<PickupMessagesReturnType> {
```
TimoGlastra, line 79: This doesn't actually wait for the pickup to be returned. Should we add this to function documentationion or something? 

If i understand correctly, mediation recipient will call the pickup in an interval, and you still need to handle the lifecycle in the mediation recipient module? Shouldn't we move all pickup logic, to the message pickup module? So picking up messages, whether implicit, explicit, live module, on an interval is managed by this module. The mediation recipient module is just a way to request mediation, register new keys, etc... (so really about the routing part)

packages/core/src/agent/AgentModules.ts
```diff
@@ -145,6 +146,7 @@ function getDefaultAgentModules(agentConfig: AgentConfig) {
         maximumMediatorReconnectionIntervalMs: agentConfig.maximumMediatorReconnectionIntervalMs,
         mediatorPollingInterval: agentConfig.mediatorPollingInterval,
       }),
+    messagePickup: () => new MessagePickupModule({ maximumBatchSize: agentConfig.maximumMessagePickup }),
```
TimoGlastra, line 149: the maximumMessagePickup is now removed from the agent config I think after my PR on agent config was merged

TimoGlastra (approved):
In general looks good, nice! Just some questions on what's the role of the mediation recipient vs message pickup module.

now that the recipient will become much simpler, should we combine the mediator and mediation recipient modules?

packages/core/src/modules/message-pìckup/MessagePickupModuleConfig.ts
genaris: We have to dig a little bit more about it but I initially I see "Message Pickup" mainly dedicated to the _explicit_ pickup and setting up the pickup mode in general. Although the "live mode" can be also seen as an implicit mode, it can probably also handled by this module (maybe using a specific queue as suggested by @rodolfomiranda in https://github.com/hyperledger/aries-rfcs/issues/760).

The _implicit_ mode is tricky because to me it's like a particular case that works for ACA-Py but don't see it specified in an RFC. Maybe it can still be triggered from the Mediation Recipient module (as it is being done right now) or left it to be handled from outside of the framework: for instance, we can add `MediatorWebSocketConnected` and `MediatorWebSocketDisconnected` events so the consumer application can manage the reconnections as needed.

packages/core/src/modules/message-pìckup/MessagePickupApi.ts
genaris: That's true: this `pickupMessages` will perform a single loop that can be called once or within an interval, and will return before the actual messages were pickup (e.g. Batch or Status/Delivery messages are received). Do you think it would be good to make it sync in order to block until the loop is completed or fail in the case of a timeout?

In this PR I left the lifecycle management in mediation module mostly to avoid introducing more breaking changes, but what I would like to do in a future is `MessagePickupApi` to create pickup loops that can be initiated/queried/stopped either by `MediationRecipientApi` or externally.

So I would say that MediationRecipientModule will be dedicated to the things you say and also holding the configuration (the `MediatorPickupStrategy`) and call `MessagePickupApi` to start/stop pickup loops as needed at the initialization/shutdown/mediator switching/etc.

genaris:
> now that the recipient will become much simpler, should we combine the mediator and mediation recipient modules?

Yes! I would like to do so. I think it could make it easier when adding more protocols like the ones for DIDComm V2

PR #1415, opened 2023-03-31
refactor(anoncreds): master secret to link secret

genaris:
Updates `AnonCredsRsHolderService` according to the recent changes in `@hyperledger/anoncreds` API and applies transformations to credential request metadata (including migration script).

There is a test that is not passing, because anoncreds-rs expects the credential definition to have a `link_secret` field instead of the former/legacy `master_secret` field (this is according to this [test](https://github.com/hyperledger/anoncreds-rs/blob/1dd3f069de74b82569b0f782cfc071db8d856ace/src/data_types/cred_def.rs#L166) added recently by @blu3beri). 

From my understanding, we should still be tied to `master_secret` in this case but AnonCreds spec is not consistent about this: in [section 7.2.2](https://hyperledger.github.io/anoncreds-spec/#generating-a-credential-definition-without-revocation-support)., it first shows an example where link_secret is used (supposed to be based in an example in Sovrin MainNet which actually uses master_secret). And then, when describing all the fields within a Credential Definition, it states:

> master_secret (also known as [link secret](https://hyperledger.github.io/anoncreds-spec/#term:link-secret), but kept as master_secret for backwards compatibility)


TimoGlastra:
Yeah the credential definition should still use `master_secret` and we should fix that in AnonCreds RS and the specification

packages/anoncreds-rs/src/services/AnonCredsRsHolderService.ts
```diff
@@ -55,19 +55,9 @@ export class AnonCredsRsHolderService implements AnonCredsHolderService {
...
-
-      return {
-        linkSecretId: options?.linkSecretId ?? utils.uuid(),
-        linkSecretValue: linkSecretJson.value.ms,
-      }
-    } finally {
-      masterSecret?.handle.clear()
+    return {
+      linkSecretId: options?.linkSecretId ?? utils.uuid(),
+      linkSecretValue: LinkSecret.create(),
```
TimoGlastra, line 60: much better :)

TimoGlastra (approved):
Fabulous!

packages/anoncreds-rs/src/services/AnonCredsRsHolderService.ts
genaris: Yeah I was a really happy moment for me when I did that update :-D

TimoGlastra:
@genaris made https://github.com/hyperledger/anoncreds-rs/pull/173

TimoGlastra:
@genaris https://github.com/hyperledger/anoncreds-rs/releases/tag/v0.1.0-dev.13 is cooking, after which we can merge this PR I think?

genaris:
Yes, I think we'll need to update some package.json files first so yarn.lock takes this new version.

PR #1416, opened 2023-04-01
fix: various anoncreds revocation fixes

TimoGlastra:
This PR addresses several issues I've encountered while testing receiving and proving revocable anoncreds credentials with both the indy-sdk and the anoncreds rs libraries. 

I've moved some methods from the specific format classes anoncreds and legacy indy to util methods as the logic is quite complex and can easily be shared between the implementations. I think we should extract more logic into util methods that can be called by both format services.

There was an inconsistency with how the holder classes used timestamps. I've now updated the code to work the same and it's the responsiblity of the caller of `HolderService.createProof` to make sure it knows which timestamp of the revocation status list to use. 

Haven't tested verification of revocable credentials yes, and also haven't tested the final changes in React Native as there were enough issues to first solve in the shared AFJ code.

~~These changes are dependant on AnonCreds RS 0.1.0-dev.13, but for that we first need #1415 to be merged.~~


swcurran:
@dbluhm and team — FYI about the narrative in this change.  Might be helpful in the ACA-Py implementation.

genaris (approved):

PR #1417, opened 2023-04-03
feat(indy-vdr)!: did transaction endorsement

2mau:


.github/actions/setup-indy-pool/action.yml
```diff
@@ -6,6 +6,9 @@ inputs:
   seed:
     description: Seed to register on the ledger
     required: true
+  endorserSeed:
```
TimoGlastra, line 9: we already have a seed registered with endorser role right? Why do we need an extra one?

DEVREADME.md
```diff
@@ -79,7 +85,7 @@ yarn test
 If you're not using the ledger setup from above, make sure you pass the correct environment variables from [Setting environment variables](#setting-environment-variables) for connecting to the indy **ledger** pool.
 
 ```sh
-GENESIS_TXN_PATH=network/genesis/local-genesis.txn TEST_AGENT_PUBLIC_DID_SEED=000000000000000000000000Trustee9 yarn test
+GENESIS_TXN_PATH=network/genesis/local-genesis.txn TEST_AGENT_PUBLIC_DID_SEED=000000000000000000000000Trustee9 yarn test ENDORSER_AGENT_PUBLIC_DID_SEED=00000000000000000000000Endorser9 yarn test
```
TimoGlastra, line 88:
```suggestion
GENESIS_TXN_PATH=network/genesis/local-genesis.txn TEST_AGENT_PUBLIC_DID_SEED=000000000000000000000000Trustee9 ENDORSER_AGENT_PUBLIC_DID_SEED=00000000000000000000000Endorser9 yarn test
```

.github/actions/setup-indy-pool/action.yml
2mau: "unknownError: Error: Request failed: client request invalid: UnauthorizedClientRequest(\"Endorser must have one of the following roles: {'ENDORSER'}\",)"

I get this error when I try to endorse transactions with the trustee role. Maybe I am doing something wrong?

TimoGlastra: Ah okay. And if we switch trustee to endorser for the main seed?

packages/indy-vdr/src/IndyVdrApi.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  public async endorseTransaction<Request extends IndyVdrRequest>(transaction: Request, endorserDid: string) {
+    const endorserSigningKey = await verificationKeyForIndyDid(this.agentContext, endorserDid)
+    const { namespaceIdentifier } = parseIndyDid(endorserDid)
+
+    const endorsedTransaction = await this.multiSignRequest(transaction, endorserSigningKey, namespaceIdentifier)
+    return endorsedTransaction
+  }
+
+  public async endorseCreateDidRequest(didAction: EndorseDidTxAction): Promise<EndorseDidTxAction> {
```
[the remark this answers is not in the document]
2mau: It works we either have 

```ts
public async signRequest<Request extends IndyVdrRequest>(request: Request, submitterDid: string) {
    const signingKey = await verificationKeyForIndyDid(this.agentContext, submitterDid)

    const signature = await this.agentContext.wallet.sign({
      data: TypedArrayEncoder.fromString(request.signatureInput),
      key: signingKey,
    })

    request.setSignature({
      signature,
    })

    return request
  }

  public async endorseTransaction<Request extends IndyVdrRequest>(transaction: Request, endorserDid: string) {
    const endorserSigningKey = await verificationKeyForIndyDid(this.agentContext, endorserDid)
    const { namespaceIdentifier } = parseIndyDid(endorserDid)

    const endorsedTransaction = await this.multiSignRequest(transaction, endorserSigningKey, namespaceIdentifier)
    return endorsedTransaction
  }
```
  
 or 
 
 
 ``` ts
public async endorseTransaction2<Request extends IndyVdrRequest>(transaction: Request, endorserDid: string) {
    const endorserSigningKey = await verificationKeyForIndyDid(this.agentContext, endorserDid)
    const { namespaceIdentifier } = parseIndyDid(endorserDid)

    let endorsedTransaction: Request
    if (transaction.body.signature) {
      endorsedTransaction = await this.multiSignRequest(transaction, endorserSigningKey, namespaceIdentifier)
    } else {
      endorsedTransaction = await this.signRequest(transaction, endorserDid)
    }
    return endorsedTransaction
}
```

File: packages/indy-vdr/src/IndyVdrApi.ts [hunk 4]
```diff
@@ -0,0 +1,71 @@
...
+    })
+
+    request.setSignature({
+      signature,
+    })
+
+    return request
+  }
+
+  public async endorseTransaction<Request extends IndyVdrRequest>(transaction: Request, endorserDid: string) {
```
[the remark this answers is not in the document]
2mau: I am not sure about this. How can we later create a Nym / AttribRequest from the JSON representation?

packages/indy-vdr/src/IndyVdrApi.ts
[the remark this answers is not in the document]
TimoGlastra: Nice! I think there's something to say for both approaches. The second one hides the real working a bit more, so maybe signRequest is better?  However, personally I prefer to have a single method that just endorses the transaction. Whether the signature is applied different doesn't really matter from a caller's perspective.

What do you prefer?


For both approaches we probably need to make sure that the submitterDid  (i don't know if we can extract that) in case of the single signature is the same as the did we're going to sign with. So we don't have the case where the submitterDid is X while the did used to sign the request is Y

packages/indy-vdr/src/IndyVdrApi.ts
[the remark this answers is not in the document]
TimoGlastra: I think you can just do `request.body` to get a json/string representation of the request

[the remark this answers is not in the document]
TimoGlastra: Then on the other side you can do `new CustomRequest({ customRequest: theRequestJsonOrString })`. 

It may be that the types aren't fully happy yet. In that case if a cast works, that's fine for now and we can update indy-vdr to allow more loose types (just any json / string)

TimoGlastra:
It seems type check if failing @2mau

packages/indy-vdr/src/dids/IndyVdrIndyDidRegistrar.ts
```diff
@@ -386,3 +563,32 @@ interface IndyVdrDidCreateOptionsWithoutDid extends IndyVdrDidCreateOptionsBase
...
+export interface EndorseDidTxAction extends DidOperationStateActionBase {
+  name: 'signNymTx'
+  did: string
+  submitterDid: string
+  nymRequest: NymRequest | CustomRequest
+  attribRequest?: AttribRequest | CustomRequest
+  didDocument: DidDocument
+  verificationKey: Key
+  role?: string
+  alias?: string
```
TimoGlastra, line 591: Are all these fields required for the endorsement? It feels a bit bloated, and I'm thinking we could just skip some of the values.

Also, should we rename `submitterDid` to `endorserDid` here? That feels more aligned with the endorsement model.

From looking at the indy vdr endorser API, we just need: `nymRequest`, optional `attribRequest`, `endorserDid`. 

The other fields seems not needed? Is that because we want to continue the create action? Shouldn't those values be inputted from the other fields of the create action?

TimoGlastra (changes requested):
Great work on the endorsement stuff. I think we may be able to simplify the implementation a bit more and streamline the flows a bit more. I left quite some suggestions on the indy vdr pool and api on what I see we can do, for the indy vdr did registrar I'm not fully sure. 

I think we can maybe split up the flows a bit more, as it's a bit of a tangled if / else flow now (which it already quite was), and that makes it quite hard to follow what's happening, and makes it more complex than needed I think.

The flows I see we have are:
1. Create did and directly sign / submit it with a did already in the wallet.
2. Create did and return it to be signed and endorsed by an endorser. When this is done we continue the submission.

I think the flow where we directly sign and submit it with a did already in the wallet is a bit tricky, because we can easily submit it with that did, but once that's done, the new did either needs to have the endorser / trustee role themselves, or all subsequent request (such as the attrib request) need to be endorsed by a did that does have that role.

Looking at what's common, it's very uncommon to create new dids with endorser role (and it's often paid and needs to be done manually). So I'm thinking we should maybe remove that flow, and also just leverage endorsement for that feature. Which means there's two modes: internal endorsement, and external endorsement. If you're using internal endorsement the initial did will be signed and written directly by the endorserDid, and the attrib request will be authored and signed by the author, after which the internal endorser will put it's signature on it.

For external endorsement the process is kinda the same, but we return the transaction to be signed before we continue.

This means we can simplify the API into the following:
- endorserDid: `` -- this is always required (because we need an endorser to write the did to the network, whether it's internal or external)
- endorsmentMode: `internal` | `external` -- for internal endorsement the endorser did MUST be in the wallet. For external endorsement mode this is not needed.

This way the implementation is kind of the same for both, except that the endorsement signinig etc.. is skipped in case of `external` endorserMode.

packages/indy-vdr/src/dids/IndyVdrIndyDidRegistrar.ts
```diff
@@ -41,205 +45,315 @@ import { endpointsAttribFromServices } from './didSovUtil'
...
+          // approach
+          seed: seed,
+          privateKey: privateKey,
+        },
+      },
+    }
+  }
 
-    const { alias, role, submitterDid, services, useEndpointAttrib } = options.options
+  public async parseInput(agentContext: AgentContext, options: IndyVdrDidCreateOptions): Promise<ParseInputResult> {
```
TimoGlastra, line 113:
```suggestion
  private async parseInput(agentContext: AgentContext, options: IndyVdrDidCreateOptions): Promise<ParseInputResult> {
```

TimoGlastra (approved):

PR #1421, opened 2023-04-05
fix: make revocation status list inline with the spec

blu3beri:
Signed-off-by: blu3beri <blu3beri@proton.me>


genaris (changes requested):
We are in the same boat: I noticed this recently and was just about to submit a PR like this one :smile: 

It seems there are some tests not yet updated to expect this field name. Can you check them?

blu3beri:
> It seems there are some tests not yet updated to expect this field name. Can you check them?

Should all be picked up now, but we shall see from the tests :).

genaris (approved):

PR #1422, opened 2023-04-05
fix(anoncreds-rs): revocation status list as JSON

genaris:
Now that anoncreds-rs exposes revocation status list from JSON, there should not need in `AnonCredsRsVerifierService` to recreate an object, so here we pass it directly as JSON.


packages/anoncreds-rs/src/services/AnonCredsRsVerifierService.ts
```diff
@@ -24,27 +24,15 @@ export class AnonCredsRsVerifierService implements AnonCredsVerifierService {
...
-              issuanceByDefault: true,
-              revocationRegistryDefinition: revocationRegistryDefinitions[revocationRegistryDefinitionId],
-              revocationRegistryDefinitionId,
-              timestamp: Number(timestamp),
-            })
-          )
-        }
+        revocationRegistryDefinitions[revocationRegistryDefinitionId] = definition as unknown as JsonObject
+
+        lists.push(...Object.values(revocationStatusLists).map((item) => item as unknown as JsonObject))
```
blu3beri, line 35:
```suggestion
        lists.push((...Object.values(revocationStatusLists)) as unknown as Array<JsonObject>)
```

packages/anoncreds-rs/src/services/AnonCredsRsVerifierService.ts
```diff
@@ -24,27 +24,15 @@ export class AnonCredsRsVerifierService implements AnonCredsVerifierService {
...
-            RevocationStatusList.create({
-              issuerId: definition.issuerId,
-              issuanceByDefault: true,
-              revocationRegistryDefinition: revocationRegistryDefinitions[revocationRegistryDefinitionId],
-              revocationRegistryDefinitionId,
-              timestamp: Number(timestamp),
-            })
-          )
-        }
+        revocationRegistryDefinitions[revocationRegistryDefinitionId] = definition as unknown as JsonObject
```
blu3beri, line 33: Any chance we can improve the casting? If we do `as unknown as...` it is very close to just using any and should be avoided whenever.

blu3beri (dismissed):
Some small comments about the casting.

packages/anoncreds-rs/src/services/AnonCredsRsVerifierService.ts
genaris: Yes, I agree, it's something we are still missing in anoncreds-rs JS API. I think we'll need to complete AnonCreds object modeling in AFJ and replicate it in anoncreds-rs javascript wrapper so they match.

packages/anoncreds-rs/src/services/AnonCredsRsVerifierService.ts
genaris: Done!

TimoGlastra (approved):

PR #1427, opened 2023-04-11
feat(anoncreds): issue revocable credentials

genaris:
This PR includes some initial work done for supporting the issuance of revocable AnonCreds credentials in AFJ. 

**Brief summary of the changes:**

- Update of `AnonCredsRegistry` and `AnonCredsIssuerService` interfaces to add registration of revocation registry definitions and revocation status lists 
- AnonCreds module, through its `AnonCredsApi`, now allows to register these objects and update revocation status lists
- `AnonCredsCredentialFormatService` has been updated to handle accept revocable credential requests, retrieving a Revocation Registry Definition and assigning an index to the credential that is going to be issued. Both the revocation registry definition Id and index are specified by agent controller when creating the credential offer.
- Credentials module allows to send revocation notification for both `indy-anoncreds` and a potentially new `anoncreds` revocation format. This is very basic right now and needs some more work to support future formats and get information directly from the credential record to be more aligned with other modules API.
- AnonCreds module now has some new configuration items for object creation. In particular it adds the new `TailsFileService`, which is an interface for downloading and uploading tails files. The default implementation works as right now (only allows to download files), while a custom one can be used to upload files to any chosen tails server. In `samples` directory there is an example of tails server and service interface (not sure if they belong there but for the moment didn't find a better place). For the tests there is a dummy implementation that allows us to run them without the need of having an actual HTTP server

**Flow:**

For the moment, as seen in the few tests added in anoncreds-rs package, all VDR objects are managed exclusively from AnonCreds API and created/registered separately in order to have a simpler state management. For instance, if we want to create a revocable credential definition, we'd need to call:
- registerSchema
- registerCredentialDefinition
- registerRevocationRegistryDefinition
- registerRevocationStatusList

The only step that does two actions atomically is `registerRevocationRegistryDefinition`, as it will attempt to successfully uploading the tails file before registering the object in the VDR (this is mandatory because we may not know upfront what's the publicly available tails location).

To revoke a credential (or a number of credentials), `AnonCredsApi.updateRevocationStatusList()` must be called using the revocation registry definition id and credential indexes as an input.  If we want to notify the holder/s about this revocation, we must use Credentials API afterwards.

**Some notes/missing pieces**

- In this PR there is only a generic implementation using `InMemoryAnonCredsRegistry`. There is not yet any implementation for Indy VDR or Indy SDK
- It could feel more 'natural ' if we add a method called `revokeCredentials` in `CredentialsApi` that does both the revocation state update and send notification. However, this integration would need us to deal with the different credential formats and, after all, I'm not sure if it will be really useful in real world scenarios, because I think that usually an issuer would prefer to revoke multiple credentials at once (to avoid creating lots of ledger updates)
- In terms of agent storage, there are no specific changes to mark credentials as revoked and to store more information about credential revocation identification from the issuer side. I think it could be useful to find a good place to do so within AFJ model
- Sorry for the monster PR!


TimoGlastra:
This is great @genaris, I need some more time to look at this, but really looking forward to merge this 🚀

packages/anoncreds/src/models/registry.ts
```diff
@@ -23,6 +23,7 @@ export interface AnonCredsRevocationRegistryDefinition {
   credDefId: string
   tag: string
   value: {
+    issuanceType: 'ISSUANCE_BY_DEFAULT' | 'ISSUANCE_ON_DEMAND'
```
TimoGlastra, line 26: Revocation registry definition doesn't have `issuanceType` anymore. This is something specific to indy.

```suggestion
```

packages/anoncreds/src/repository/AnonCredsRevocationRegistryDefinitionPrivateRecord.ts
```diff
@@ -0,0 +1,62 @@
+import type { TagsBase } from '@aries-framework/core'
+
+import { BaseRecord, utils } from '@aries-framework/core'
+
+export enum RevocationRegistryState {
```
TimoGlastra, line 5:
```suggestion
export enum AnonCredsRevocationRegistryState {
```

packages/anoncreds/src/repository/AnonCredsRevocationRegistryDefinitionPrivateRecord.ts
```diff
@@ -0,0 +1,62 @@
...
+  Active = 'active',
+  Full = 'full',
+}
+
+export interface AnonCredsRevocationRegistryDefinitionPrivateRecordProps {
+  id?: string
+  revocationRegistryDefinitionId: string
+  credentialDefinitionId: string
+  value: Record<string, unknown>
+  index?: number
```
TimoGlastra, line 16: Is this the current index? Maybe we can call it `currentRevocationIndex` or something to make it a bit more descriptive?

TimoGlastra, line 16: I'm thinking whether we should add an option to manage the index that will be used for issuance outside of the framework. As mentioned before, we can have issues when multiple issuance processed are in process as they will overrwrite the value (can't do index += 1).

For us we would thus like to manage this outside of AFJ and store the index in a service where it is possible to prevent overwriting of the indexes.

packages/anoncreds/src/repository/AnonCredsRevocationStatusListRecord.ts
```diff
@@ -0,0 +1,46 @@
...
+  revocationStatusList: AnonCredsRevocationStatusList
+}
+
+export type DefaultAnonCredsRevocationStatusListTags = {
+  revocationRegistryDefinitionId: string
+  credentialDefinitionId: string
+  timestamp: string
+}
+
+export class AnonCredsRevocationStatusListRecord extends BaseRecord<
```
TimoGlastra, line 18: Why do we store this object? Do we take this as the base when calling `updateRevocationStatusList`?

packages/anoncreds/src/services/AnonCredsIssuerServiceOptions.ts
```diff
@@ -17,11 +22,36 @@ export interface CreateCredentialDefinitionOptions {
...
+  tag: string
+  credentialDefinitionId: string
+  credentialDefinition: AnonCredsCredentialDefinition
+  maximumCredentialNumber: number
+  tailsDirectoryPath: string
+}
+
+export interface CreateRevocationStatusListOptions {
+  issuerId: string
+  issuanceByDefault: boolean
```
TimoGlastra, line 40: I think we should remove this property and always use ISSUANCE_BY_DEFAULT. It's not recommended to use ISSUANCE_ON_DEMAND, and e.g. ACA-Py doesn't allow you to create it with ISSUANCE_ON_DEMAND at all.

packages/anoncreds/src/services/tails/BasicTailsFileService.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  public async getTailsFilePath(agentContext: AgentContext, tailsHash: string) {
+    return `${await this.getTailsBasePath(agentContext)}/${tailsHash}`
+  }
+
+  public async tailsFileExists(agentContext: AgentContext, tailsHash: string): Promise<boolean> {
+    const fileSystem = agentContext.dependencyManager.resolve<FileSystem>(InjectionSymbols.FileSystem)
+    const tailsFilePath = await this.getTailsFilePath(agentContext, tailsHash)
+    return await fileSystem.exists(tailsFilePath)
+  }
```
TimoGlastra, line 31: I think we shouldn't assume the tails file is only accessed using it's hash. Do you think we could also require the revocation registry id etc.. for the methods?

packages/anoncreds/src/services/tails/BasicTailsFileService.ts
```diff
@@ -0,0 +1,92 @@
...
+        agentContext.config.logger.debug(`Retrieving tails file from URL ${tailsLocation}`)
+
+        // download file and verify hash
+        await fileSystem.downloadToFile(tailsLocation, tailsFilePath, {
+          verifyHash: {
+            algorithm: 'sha256',
+            hash: TypedArrayEncoder.fromBase58(tailsHash),
+          },
+        })
+        agentContext.config.logger.debug(`Saved tails file to FileSystem at path ${tailsFilePath}`)
```
TimoGlastra, line 79: Thinking whether we should split the downloading and verification of the tails file? Or should the implementer of a tails file service take this into account when implementing?

packages/anoncreds/src/utils/getRevocationRegistries.ts
```diff
@@ -88,8 +88,10 @@ export async function getRevocationRegistriesForRequest(
             )
           }
 
-          const { tailsLocation, tailsHash } = revocationRegistryDefinition.value
-          const { tailsFilePath } = await downloadTailsFile(agentContext, tailsLocation, tailsHash)
+          const tailsFileService = agentContext.dependencyManager.resolve(AnonCredsModuleConfig).tailsFileService
+          const { tailsFilePath } = await tailsFileService.downloadTailsFile(agentContext, {
+            revocationRegistryDefinition,
```
TimoGlastra, line 93: Is it clear that this will check if it's already downloaded? Or do you think we should do a check if it exists first?

packages/anoncreds/tests/anoncreds.test.ts
```diff
@@ -46,6 +46,7 @@ const existingRevocationRegistryDefinitions = {
     issuerId: 'VsKV7grR1BUE29mG2Fm2kX',
     revocDefType: 'CL_ACCUM',
     value: {
+      issuanceType: 'ISSUANCE_BY_DEFAULT',
```
TimoGlastra, line 49:
```suggestion
```

packages/cheqd/src/anoncreds/utils/transform.ts
```diff
@@ -102,6 +102,9 @@ export class CheqdRevocationRegistryDefinitionValue {
 
   @IsString()
   public tailsHash!: string
+
+  @IsString()
+  public issuanceType!: 'ISSUANCE_BY_DEFAULT' | 'ISSUANCE_ON_DEMAND'
```
TimoGlastra, line 107: This isn't part of the spec anymore

```suggestion
```

packages/core/src/modules/credentials/CredentialsApi.ts
```diff
@@ -565,6 +569,68 @@ export class CredentialsApi<CPs extends CredentialProtocol[]> implements Credent
...
+   * Send a revocation notification for a credential exchange record. Currently Revocation Notification V2 protocol is supported
+   *
+   * @param credentialRecordId The id of the credential record for which to send revocation notification
+   */
+  public async sendRevocationNotification(options: SendRevocationNotificationOptions): Promise<void> {
+    const { credentialRecordId, revocationId, revocationFormat, comment, requestAck } = options
+
+    const credentialRecord = await this.getById(credentialRecordId)
+
+    const { message } = await this.revocationNotificationService.v2CreateRevocationNotification({
```
TimoGlastra, line 582: We should look at a way to remove the `v2` prefix here. But it's fine for now 👍

packages/core/src/modules/credentials/CredentialsApi.ts
```diff
@@ -565,6 +569,68 @@ export class CredentialsApi<CPs extends CredentialProtocol[]> implements Credent
     }
   }
 
+  /**
+   * Send a revocation notification for a credential exchange record. Currently Revocation Notification V2 protocol is supported
+   *
+   * @param credentialRecordId The id of the credential record for which to send revocation notification
+   */
+  public async sendRevocationNotification(options: SendRevocationNotificationOptions): Promise<void> {
```
TimoGlastra, line 577: Do you think we can make it possible to send this notification without requiring the credential record id? I think in a lot of cases you'd want to remove the credential record after you've issued so there's no PII in storage

packages/core/src/modules/credentials/protocol/revocation-notification/services/RevocationNotificationService.ts
```diff
@@ -116,14 +143,15 @@ export class RevocationNotificationService {
 
     const credentialId = messageContext.message.credentialId
 
-    if (messageContext.message.revocationFormat !== v2IndyRevocationFormat) {
+    if (![v2IndyRevocationFormat, v2AnonCredsRevocationFormat].includes(messageContext.message.revocationFormat)) {
       throw new AriesFrameworkError(
-        `Unknown revocation format: ${messageContext.message.revocationFormat}. Supported formats are indy-anoncreds`
+        `Unknown revocation format: ${messageContext.message.revocationFormat}. Supported formats are indy-anoncreds and anoncreds`
```
TimoGlastra, line 148: Is the RFC extended with this new format?

packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
```diff
@@ -443,6 +446,7 @@ export class IndyVdrAnonCredsRegistry implements AnonCredsRegistry {
         issuerId: did,
         revocDefType: response.result.data.revocDefType,
         value: {
+          issuanceType: response.result.data.value.issuanceType,
```
TimoGlastra, line 449:
```suggestion
```

samples/tails/README.md
```diff
@@ -0,0 +1,5 @@
+<h1 align="center"><b>Test tails file server</b></h1>
+
+This is a very simple server that is used to host tails files during for tests suites where revocable credentials are issued and verified.
```
TimoGlastra, line 3: Couldn't we use file paths in the revocation registries for tests?

TimoGlastra (changes requested):
This is amazing @genaris. I've left quite some comments, a lot are just questions and thoughts, not sure we want to address them all (or now)

packages/anoncreds-rs/src/services/AnonCredsRsIssuerService.ts
```diff
@@ -83,6 +101,102 @@ export class AnonCredsRsIssuerService implements AnonCredsIssuerService {
...
+      options
+
+    let revocationStatusList: RevocationStatusList | undefined
+    try {
+      revocationStatusList = RevocationStatusList.create({
+        issuanceByDefault,
+        revocationRegistryDefinitionId,
+        revocationRegistryDefinition: {
+          ...revocationRegistryDefinition,
+          value: { ...revocationRegistryDefinition.value, tailsLocation: tailsFilePath },
```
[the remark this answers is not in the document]
genaris: anoncreds-rs expects to read the contents of the tails file from a local file system path (otherwise it throws a 'Can't read tails bytes from file' error). So that's the reason behind this 'trick'.

packages/anoncreds-rs/src/services/AnonCredsRsIssuerService.ts
```diff
@@ -83,6 +101,102 @@ export class AnonCredsRsIssuerService implements AnonCredsIssuerService {
...
+      updatedRevocationStatusList = RevocationStatusList.fromJson(revocationStatusList as unknown as JsonObject)
+
+      if (timestamp && !issued && !revoked) {
+        updatedRevocationStatusList.updateTimestamp({
+          timestamp,
+        })
+      } else {
+        revocationRegistryDefinitionObj = RevocationRegistryDefinition.fromJson({
+          ...revocationRegistryDefinition,
+          value: { ...revocationRegistryDefinition.value, tailsLocation: tailsFilePath },
```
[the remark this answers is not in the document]
genaris: As noted above, this is only for the purpose of processing the update of the revocation status list.

packages/anoncreds-rs/src/services/AnonCredsRsVerifierService.ts
```diff
@@ -41,9 +58,89 @@ export class AnonCredsRsVerifierService implements AnonCredsVerifierService {
...
+      schemaId?: string
+      credentialDefinitionId?: string
+      revocationRegistryDefinitionId?: string
+    }[] = []
+
+    for (const value of [
+      ...Object.values(proofRequest.requested_attributes),
+      ...Object.values(proofRequest.requested_predicates),
+    ]) {
+      const nonRevokedInterval = globalNonRevokedInterval ?? value.non_revoked
```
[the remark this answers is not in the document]
genaris: Right!

packages/anoncreds-rs/tests/InMemoryTailsFileService.ts
```diff
@@ -0,0 +1,60 @@
+import type { AnonCredsRevocationRegistryDefinition } from '@aries-framework/anoncreds'
+import type { AgentContext } from '@aries-framework/core'
+
+import { BasicTailsFileService } from '@aries-framework/anoncreds'
+
+export class InMemoryTailsFileService extends BasicTailsFileService {
+  private tailsFilePaths: Record<string, string> = {}
+
+  public async uploadTailsFile(
```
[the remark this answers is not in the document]
genaris: This `tailsLocation` thing is a bit confusing, as the output from the library is a RevocationRegistryDefinition whose tailsLocation is actually the local file path, and then we should update it with the actual URL path before storing in the wallet and publishing to the VDR.

Besides this explanation, I think you are right and it will be better to not use the revocation registry definition object but its ID and either the local tails file path or the file contents itself. At first sight I'd prefer to pass the path, as usually libraries like fetch or axios do work well with file streams and that could be more efficient if handled by lower level modules.

packages/anoncreds/src/formats/AnonCredsCredentialFormatService.ts
```diff
@@ -303,10 +311,67 @@ export class AnonCredsCredentialFormatService implements CredentialFormatService
...
+      if (!revocationRegistryDefinitionPrivateRecord) {
+        throw new AriesFrameworkError(`No available revocation registry found for ${credentialRequest.cred_def_id}`)
+      }
+
+      revocationRegistryDefinitionId = revocationRegistryDefinitionPrivateRecord.revocationRegistryDefinitionId
+
+      // get current revocation status list
+      const registryService = agentContext.dependencyManager.resolve(AnonCredsRegistryService)
+      const registry = registryService.getRegistryForIdentifier(agentContext, credentialRequest.cred_def_id)
+      const result = await registry.getRevocationStatusList(
```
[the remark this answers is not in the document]
genaris: Very good point. It seems that revocation status list is required by `anoncreds-rs` to determine [if the index uses issuance by default or not](https://github.com/hyperledger/anoncreds-rs/blob/82e50f5e8215c4dbf4bb60ba65533716f5e1ae08/src/services/issuer.rs#L741).

If we are only supporting issuance by default, of course this processing does not make sense. I feel this should be addressed in anoncreds-rs, but as a workaround for this, might we send to the library a dummy revocation status list in this case?

packages/anoncreds/src/repository/AnonCredsRevocationStatusListRecord.ts
genaris: Yes, we do, but it's true that we don't need a specific record for it, unless we are going to use it as a cache. However I guess such cache will be responsibility of the corresponding AnonCredsRegistry.

packages/anoncreds/src/AnonCredsModuleConfig.ts
```diff
@@ -9,6 +12,18 @@ export interface AnonCredsModuleConfigOptions {
    * A list of AnonCreds registries to make available to the AnonCreds module.
    */
   registries: [AnonCredsRegistry, ...AnonCredsRegistry[]]
+
+  /**
+   * Maximum credential number per revocation registry
+   * @default 1000
+   */
+  maximumCredentialNumberPerRevocationRegistry?: number
```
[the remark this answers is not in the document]
genaris: I think I've added this because initially I wanted to do an 'automatic approach' like ACA-Py and then switched to a manual one. So for the current code this setting does not make any sense.

samples/tails/README.md
genaris: Actually, in tests we are using the `InMemoryTailsFileService`, so this sample tails server is only kept as a reference (not sure if it's the best place to store it though). The README is a bit misleading: I'll update it

packages/core/src/modules/credentials/CredentialsApi.ts
genaris: We are using the credential exchange record Id in order to get either the connection or services (though the corresponding request/offer messages) to send the message to. 

I think some alternatives to this approach would be to either tag connections with the `revocationId` of every credential issued to them, or create a "CredentialRevocationRecord" every time a credential is issued. The second approach is a bit more flexible because such record can contain the connectionId, the service used, etc. so it works for all cases.

packages/anoncreds/src/formats/AnonCredsCredentialFormatService.ts
```diff
@@ -303,10 +311,67 @@ export class AnonCredsCredentialFormatService implements CredentialFormatService
...
+        agentContext,
+        revocationRegistryDefinitionId,
+        dateToTimestamp(new Date())
```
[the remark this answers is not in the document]
genaris: Timestamps are expressed in seconds, while Date.now() returns milliseconds.

packages/anoncreds/src/repository/AnonCredsRevocationRegistryDefinitionPrivateRecord.ts
genaris: Yeah, this can be problematic in several cases, especially in batch issuance (like I think it is also for batch revocation). 

I think for now the best would be to not store any currentRevocationIndex in the record and just leave that responsibility and logic to the controller.

genaris: The challenge for that approach is to see where to add that binding: a simple way would be setting it at `CredentialsApi.proposeCredential` step (add revocationIndex to `AnonCredsProposeCredentialFormat` and store it in credential exchange metadata). However, this will assign an index even if a credential has not actually been issued.

So we can also do it at `CredentialsApi.acceptRequest` (adding it to `AnonCredsAcceptRequestFormat`), although this would not work if ` shouldAutoRespondToRequest` happens to be true.

Another more complex option would be creating an interface for a "RevocationService" that feeds to AFJ's `AnonCredsCredentialFormatService.acceptRequest()` the proper index for a given revocationRegistryDefinitionId. That service should also decide which revocation registry is the active one instead of querying it into AFJ storage.

Now I think about it, a controller can also provide the `revocationRegistryDefinitionId` to proposeCredental/acceptRequest options, so it can manage revocation registry definition generation in a complete separate way (could be useful considering the processing time it takes to generate a new one).

packages/anoncreds/src/utils/getRevocationRegistries.ts
genaris: I renamed this to `getTailsFile`, so it will retrieve from the public URL or use a cached version if the TailsFileManager implementation does so.

packages/anoncreds/src/services/tails/BasicTailsFileService.ts
genaris: IMHO it's fine to leave this responsibility to the implementer (in the same way, Indy Tails server does its own checks regarding revocation registry existence before actually uploading a file).

packages/anoncreds/src/services/tails/BasicTailsFileService.ts
genaris: This is true but it is hard to reference local tails files in a different way given that we don't know the revocation registry id upfront.

That being said, I think some of those methods are not at the interface level, so for the moment I left only three public methods:
- uploadTailsFile
- getTailsFile
- getTailsBasePath 

The latter is used to indicate anoncreds library where to store the local tails file it will generate. I'll need to re-think this a little bit because ideally it could store in any temporary directory and then the `TailsFileService` can store it in its internal cache once it uploads it.

genaris:
> This is amazing @genaris. I've left quite some comments, a lot are just questions and thoughts, not sure we want to address them all (or now)

Thank you very much for your deep review @TimoGlastra ! I think I addressed most of your feedback: not everything implemented, but at least analyzed and left some comments for further discussion. Would be good to do another "pass" to discuss about the API and implementation. Especially regarding the revocation index and revocation registry management from outside the framework and the TailsFileService interface.

packages/anoncreds/src/formats/AnonCredsCredentialFormatService.ts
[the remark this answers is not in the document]
TimoGlastra: Workaround works for me! Maybe you can open an issue so we can track it

packages/anoncreds/src/repository/AnonCredsRevocationRegistryDefinitionPrivateRecord.ts
TimoGlastra: I think both upfront and callback/interface based would work.

If auto issuance is enabled we would need callback/interface based. So maybe that's the way to go? Auto issuance shouldn't be encouraged though i think.

I think if we have auto acceptance and no explicit offer we should jist throw an error if the vred def needs revocation?

Lot's of complexities -- i would like the simplest most extendable way (whatever that is 😬)

TimoGlastra:
Discussed in AFJ WG call. Changes have been reviewed multiple times. Some things left to improve / change, but good to get this merged. Will be available as an experimental feature. The breaking changes are to the AnonCreds registry and services, but those are marked as experimental for implementing your own (besides the ones provided by the framework), so this can be released as 0.4.1. @genaris to update with latest main

packages/core/src/modules/credentials/protocol/revocation-notification/services/RevocationNotificationService.ts
genaris: RFC has been updated

packages/anoncreds/src/repository/AnonCredsRevocationRegistryDefinitionPrivateRecord.ts
genaris: > I think if we have auto acceptance and no explicit offer we should jist throw an error if the vred def needs revocation?

For the moment I left it like that: only manual flow is supported (which will be the most common anyway) and offerCredential/acceptProposal API receive both revocationRegistryDefinitionId and revocation index

packages/anoncreds/src/formats/AnonCredsCredentialFormatService.ts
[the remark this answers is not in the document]
genaris: Unfortunately, anoncreds-rs does an strict check of revocation status list object, so the 'dummy' object is not as dummy as we would like (it's needed to generate it through the library). This generation seems to be fast but I'm not completely sure how it will perform in large use cases. 

I opened https://github.com/hyperledger/anoncreds-rs/issues/227 to discuss this and confirm that our reasoning is correct. If we are right, it would be a nice update to the library to simplify both its API and processing time.

genaris:
@TimoGlastra I did a few updates based on the last feedback from last month. Would be good to have a brief discussion of https://github.com/hyperledger/anoncreds-rs/issues/227 with other AnonCreds maintainers just to be sure that our reasoning is correct.

swcurran:
@genaris — wondering about the approach used here.  We’ve found in with ACA-Py that it is very difficult for the controller (application code) to manage RevRegs, and have put almost all the work on the Framework. Are you doing that here?  It will also make it easier to use other methods (e.g., StatusList2021) since the interface stays much the same.

Notably:

- Issuer just flags that revocation will be used when creating the CredDef and the size of the RevReg. The library takes case of creating the RevReg(s) needed without further action by the controller.
- For AnonCreds current Rev approach, two RevRegs are always active, so there is no delay when one fills up, and another is needed. Whenever one fills up, a new one is created. On creation, tails file handling is within Framework.
- On Issuance, controller receives the “revocation ID" of the issued credential to track. “Revocation ID” is the combination of RevRegId+Index.
- On Revocation, controller passes in “revocation ID" of the revoked credential with “notify holder” flag. Framework tracks revocations.
- To publish an update to the revocation state, Issuer passes in CredDefId, and Framework figures out what RevRegs need to be published and does so.

This limits the tracking to be done by controller to a minimum, and relieves it of ever dealing with RevRegs.

From experience, we’ve added to two features to deal with unexpected conditions:

- If the local copy of a RevReg gets out of sync of the Indy ledger copy (due to a failed ledger write), the Framework can create an update to “fix” the issue. This is needed for a deltas-based solution, but wouldn’t be needed for a full state version.
- From time to time there may be a need to “rotate” a RevReg — mark existing, still useful RevRegs as “decommissioned” and create new ones in their place. We’ve added an endpoint (api call) for that.

TimoGlastra:
@genaris is this ready to be merged? Any things you're waiting on?

genaris:
> @genaris is this ready to be merged? Any things you're waiting on?

Yes, I think so!

TimoGlastra (approved):

TimoGlastra:
Thanks @genaris ❤️

PR #1431, opened 2023-04-13
fix: issuance with unqualified identifiers

TimoGlastra:
Bigger PR than anticipated, but this fixes the issuance of credentials with unqualified identifiers. You can now issue with both unqulaiifed and did indy identifiers interchangeably, but only for the issuance side. the other sides don't work interchangeably yet. 

I hope this is one of the last big tasks before releasing 0.4.0?

Lot of code is moving around stuff and removing some of the duplicated code. We now have indy specific identifier code in the anoncreds package, but it's needed for the unqualified handling. As it's part of the spec, and it can provide the easiest migration path, that is okay by me. (it's not much, just some identifier parsing etc..)

Hope this doesn't clash too much with your Revocation pr @genaris


packages/anoncreds/src/repository/AnonCredsSchemaRecord.ts
```diff
@@ -48,13 +55,20 @@ export class AnonCredsSchemaRecord extends BaseRecord<
   }
 
   public getTags() {
+    let unqualifiedSchemaId: string | undefined = undefined
+    if (this.schemaId.startsWith('did:indy:')) {
```
genaris, line 59: I guess it's not needed for this particular case (and probably less performant), but could be more elegant to match with the regex that now has been added to anoncreds package? The same with other "startsWith('did:indy:')"

packages/askar/src/storage/AskarStorageService.ts
```diff
@@ -159,7 +159,7 @@ export class AskarStorageService<T extends BaseRecord> implements StorageService
       }
       return instances
     } catch (error) {
-      throw new WalletError(`Error executing query`, { cause: error })
+      throw new WalletError(`Error executing query. ${error.message}`, { cause: error })
```
genaris, line 162: Not against this change but just want to check if it's intended

packages/core/src/modules/vc/W3cCredentialsModule.ts
```diff
@@ -1,4 +1,4 @@
-import type { W3cVcModuleConfigOptions } from './W3cCredentialsModuleConfig'
+import type { W3cCredentialsModuleConfigOptions } from './W3cCredentialsModuleConfig'
```
genaris, line 1: Again, as this is unrelated with the qualified/unqualified dids thing just want to remark it to check if it's intended to add in this PR (no problem from my side ofc!).

genaris (approved):
I don't think it clashes so much with the revocation PR so LGTM 😆 

Just a few comments to double check a few things.

packages/core/src/modules/vc/W3cCredentialsModule.ts
TimoGlastra: Yes, intended. Just came across it and thought let's fix it directly

PR #1435, opened 2023-04-17
feat: Add cheqd demo and localnet for tests

DaevMithran:
This PR adds the required Readme files and integrates cheqd modules in the demo

Also included a setup-cheqd in ./.github/actions which starts a localnet for the tests to be run

### DIFF
--- a/packages/cheqd/src/anoncreds/services/CheqdAnonCredsRegistry.ts
+++ b/packages/cheqd/src/anoncreds/services/CheqdAnonCredsRegistry.ts
@@ -106,7 +106,11 @@ export class CheqdAnonCredsRegistry implements AnonCredsRegistry {
           schemaId: `${schema.issuerId}/resources/${schemaResource.id}`,
         },
         registrationMetadata: {},
-        schemaMetadata: {},
+        schemaMetadata: {
+          // NOTE: the seqNo is required by the indy-sdk even though not present in AnonCreds v1.
+          // For this reason we return a default value in metadata.
+          indyLedgerSeqNo: 1,
+        },
       }
     } catch (error) {
       agentContext.config.logger.debug(`Error registering schema for did '${options.schema.issuerId}'`, {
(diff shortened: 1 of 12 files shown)


packages/cheqd/src/anoncreds/services/CheqdAnonCredsRegistry.ts
```diff
@@ -106,7 +106,11 @@ export class CheqdAnonCredsRegistry implements AnonCredsRegistry {
           schemaId: `${schema.issuerId}/resources/${schemaResource.id}`,
         },
         registrationMetadata: {},
-        schemaMetadata: {},
+        schemaMetadata: {
```
TimoGlastra, line 109: We don't support cheqd with indy sdk, so this isn't needed i guess?

DaevMithran: Anoncreds api is throwing an error if I don't include it
 https://github.com/hyperledger/aries-framework-javascript/blob/main/packages/anoncreds/src/AnonCredsApi.ts#L264

TimoGlastra: What is the error it throws? Are you using indy sdk?

DaevMithran: Yea I tried using anoncreds-rs module for anoncreds and indysdk for wallet. It panics at that line and returns invalid did:indy did even though I'm using did:cheqd

TimoGlastra: In that case it's probably an issue with using AnonCreds RS and Indy SDK together. It's not intended to be used like that, and thus I don't think we should add code to support the flow

DaevMithran: Removed that change

TimoGlastra:
@genaris any idea what's going on here?

@ryjones is this related to the different runners?

PR #1439, opened 2023-04-25
fix: small updates to cheqd module and demo

TimoGlastra:

PR #1440, opened 2023-04-26
feat: demo now runs under a temp HOME directory.

### LINKED ISSUE #1430: CLI demo broken, espeically when running multiple times
Reported by multiple people

daidoji:
Adjusted demo so that the Alice and Faber wallets are written to a temp directory created at demo runtime (via setting the HOME env variable in the script) rather than using the home directory of the user running the demo.

Fix #1430


TimoGlastra:
Hey @daidoji, thanks for the PR! 

Is the reason to put the demo under a temp directory because of the issues with re-running the demo? I have a PR that should fix this: #1439. 

Or do you want to keep the data separate? For Indy SDK you can change the path using the `storage_config`: https://github.com/hyperledger/indy-sdk/tree/main/wrappers/nodejs#createwallet--config-credentials----void (not sure if this is exposed in the interface), and for Askar you can change the baseDataPath in the file system: https://github.com/hyperledger/aries-framework-javascript/blob/main/packages/node/src/NodeFileSystem.ts#L28-L29

daidoji:
@TimoGlastra cool, yeah if your PR covers it I can just withdraw this one.  The idea was just to keep whatever wallet state from the demo out of any indy wallet credentials I might have in my home directory.  Using HOME was the only way I could figure out how to do that because I didn't find what you linked.  

How would I check that these are available in the interface and maybe I could update the documentation for future generations.  Does that #1439 PR cover changes to get the demo working again?

TimoGlastra:
@daidoji now that #1439 is merged, could you check if it's fully working again in the main branch? Otherwise we can add the changes to this PR to fix it

PR #1442, opened 2023-04-28
ci: downgrade ubuntu docker image

genaris:
Update anoncreds-rs to 0.1.0-dev.15 which is compatible with an older GLIBC. This allows to revert recent changes in CI/CD to workaround this issue, which were using ubuntu 22.04 and a downgraded version of libssl (for indy compatibility).


genaris:
Looks like no runner is picking up jobs with the label `ubuntu-2004-paid`. @ryjones could you confirm what's the proper label we must use for the large Ubuntu 20.04 runner?

TimoGlastra (approved):

PR #1445, opened 2023-05-01
fix: Emit RoutingCreated event for mediator routing record

AlexanderShenshin:
Added `RoutingCreatedEvent` emit after creation of mediator routing record in `MediatorService`.

This change is required to enable mediator functionality for tenant agents as `registerRecipientKeyForTenant` will not be called for mediator routing key otherwise (done in `RoutingCreatedEvent` listener).

### DIFF
--- a/packages/core/src/modules/routing/services/MediatorService.ts
+++ b/packages/core/src/modules/routing/services/MediatorService.ts
@@ -212,6 +212,17 @@ export class MediatorService {
 
     await this.mediatorRoutingRepository.save(agentContext, routingRecord)
 
+    this.eventEmitter.emit(agentContext, {
+      type: RoutingEventTypes.RoutingCreatedEvent,
+      payload: {
+        routing: {
+          endpoints: agentContext.config.endpoints,
+          routingKeys: [],
+          recipientKey: routingKey,
+        },
+      },
+    })
+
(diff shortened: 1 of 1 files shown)


TimoGlastra (commented):
Maybe the event should get a different name, but this event exists mainly for mulit-tenancy, where the wallet creates a routing record, and the base wallet should add a mapping for routing the message to the correct tenant.

In this case you could listen to the record created event with type `MediatorRoutingRecord`

TimoGlastra (approved):

TimoGlastra:
Do we want tenants to act as mediators? Do you have a specific use case?

AlexanderShenshin:
Yes, we have a use case where we want to use mediator functionality in multi-tenancy mode.

It seems that in multi-tenancy mode it's not possible to normally use root agent as a mediator because of how `TenantAgentContextProvider` resolves context for inbound message (it will not find agent context for root agent as it does not have tenant routing record). So we decided to create a separate tenant agent for mediator functionality and it also seems to be more appropriate with current API.

We actually use this change in our project with `patch-package` as a quick fix for now. It works fine so far, so we would like to keep this approach for now. But thanks for clarification, we will keep possible issues in mind.

PR #1446, opened 2023-05-01
feat: support for did:jwk and p-256, p-384, p-512

TimoGlastra:


packages/core/src/crypto/keyUtils.ts
```diff
@@ -9,6 +9,9 @@ export function isValidSeed(seed: Buffer, keyType: KeyType): boolean {
     [KeyType.Bls12381g1]: 32,
     [KeyType.Bls12381g2]: 32,
     [KeyType.Bls12381g1g2]: 32,
+    [KeyType.P256]: 32,
+    [KeyType.P384]: 48,
+    [KeyType.P521]: 66,
```
TimoGlastra, line 14: @andrewwhitehead I'm not 100% on what the seed and private key sizes should be for `P-256`, `P-384` and `P-521`. Do you know if this is correct?

packages/core/src/crypto/JwkTypes.ts
```diff
@@ -1,6 +1,157 @@
...
+  x: string
+  y: string
+  use?: 'sig' | 'enc'
+}
+
+export function isEd25519Jwk(jwk: Jwk): jwk is Ed25519Jwk {
+  return jwk.kty === 'OKP' && jwk.crv === 'Ed25519' && jwk.x !== undefined && (!jwk.use || jwk.use === 'sig')
+}
+
+export function isX25519Jwk(jwk: Jwk): jwk is X25519Jwk {
```
blu3beri, line 67: Do we want to add the `x` check here? Without the X it can still be a valid x25519, just not one that contains a public key. Maybe a rename if that behaviour is desired?

packages/core/src/crypto/__tests__/Jwk.test.ts
```diff
@@ -0,0 +1,124 @@
+import type { Bls12381g1Jwk, Bls12381g2Jwk, Ed25519Jwk, P256Jwk, P384Jwk, P521Jwk, X25519Jwk } from '../JwkTypes'
+
+import { getJwkFromKey, getKeyDataFromJwk } from '../Jwk'
+import { Key } from '../Key'
+import { KeyType } from '../KeyType'
+
+describe('jwk', () => {
```
blu3beri, line 7: Are these some fixtures that are provided by the JWK specification or maybe generated by openssl or something? If so, ignore this :) otherwise I think it would be very good to add those as using the public from a jwk to verify a signature, but the key is incorrect would lead to very difficult debugging.

blu3beri (changes requested):
Really nice addition! Left some remarks, nothing that should be difficult to resolve.

packages/askar/src/utils/askarKeyTypes.ts
```diff
@@ -1,6 +1,13 @@
...
 
-export const keyTypeSupportedByAskar = (keyType: KeyType) =>
-  Object.entries(KeyAlgs).find(([, value]) => value === keyType.toString()) !== undefined
+const keyTypeToAskarAlg = {
+  [KeyType.Ed25519]: KeyAlgs.Ed25519,
+  [KeyType.X25519]: KeyAlgs.X25519,
+  [KeyType.Bls12381g1]: KeyAlgs.Bls12381G1,
+  [KeyType.Bls12381g2]: KeyAlgs.Bls12381G2,
+  [KeyType.Bls12381g1g2]: KeyAlgs.Bls12381G1G2,
+  [KeyType.P256]: KeyAlgs.EcSecp256r1,
```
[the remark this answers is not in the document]
TimoGlastra: That's `Secp256K1`. If you look at the values of `KeyAlgs` from askar, it's as follows:

```
    EcSecp256k1 = "k256",
    EcSecp256r1 = "p256"
```

packages/core/src/crypto/EcCompression.ts
```diff
@@ -0,0 +1,104 @@
...
+  const xOctet = Uint8Array.from(Buffer.from(xHex, 'hex'))
+  const yOctet = Uint8Array.from(Buffer.from(yHex, 'hex'))
+  return compressECPoint(xOctet, yOctet)
+}
+
+export function expand(publicKey: Uint8Array, curve: 'P-256' | 'P-384' | 'P-521'): Uint8Array {
+  const publicKeyComponent = Buffer.from(publicKey).toString('hex')
+  const { prime, b, pIdent } = getConstantsForCurve(curve)
+  // eslint-disable-next-line
+  var signY: any = (new Number(publicKeyComponent[1]) as any) - 2
```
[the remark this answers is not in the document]
TimoGlastra: You can't substract a `number` from an actual Number instance. I copied it from transmute's library, but will see if I can improve it a bit

packages/core/src/crypto/keyUtils.ts
TimoGlastra: I think it's p521, not p512 (so it's 65.125)

packages/core/src/crypto/Jwk.ts
```diff
@@ -0,0 +1,84 @@
...
+    return {
+      publicKey: TypedArrayEncoder.fromBase64(jwk.x),
+      keyType: jwkCurveToKeyTypeMapping[jwk.crv],
+    }
+  }
+
+  // p-256, p-384, p-521
+  if (isP256Jwk(jwk) || isP384Jwk(jwk) || isP521Jwk(jwk)) {
+    // TODO: do we want to use the compressed key in the Key instance?
+    const publicKeyBuffer = Buffer.concat([TypedArrayEncoder.fromBase64(jwk.x), TypedArrayEncoder.fromBase64(jwk.y)])
```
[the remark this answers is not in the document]
TimoGlastra: I think the did:key spec only supports the compressed variant (that's what I understand from the thread), but you're saying the JWK itself could contain both compressed and uncompressed variants? I think we could just support uncompressed as well (as it's just skipping the compression/uncompression and thus simpler). 

We'll always encode to compressed variant (because why not), but for decompression I should just look if the key is prefixed with `0x3` or `0x2` and if so, remove that previx and not do the decompression?

packages/core/src/crypto/EcCompression.ts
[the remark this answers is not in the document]
karimStekelenburg: Why the use of `var` instead of `let`? Just curious.

packages/core/src/crypto/Jwk.ts
```diff
@@ -0,0 +1,84 @@
...
+      crv: keyTypeToJwkCurveMapping[key.keyType],
+      x: TypedArrayEncoder.toBase64URL(key.publicKey),
+    } satisfies Ed25519Jwk | X25519Jwk
+  }
+
+  if (key.keyType === KeyType.Bls12381g1 || key.keyType === KeyType.Bls12381g2) {
+    return {
+      kty: 'EC',
+      crv: keyTypeToJwkCurveMapping[key.keyType],
+      x: TypedArrayEncoder.toBase64URL(key.publicKey),
```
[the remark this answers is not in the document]
TimoGlastra: Hmm I took this from the mattr lib: https://github.com/mattrglobal/bls12381-key-pair/blob/master/src/Bls12381G1KeyPair.ts#L332. But maybe it has changed now? Did:key has the same :https://github.com/w3c-ccg/did-method-key/blob/main/test-vectors/bls12381.json#L187, https://github.com/mattrglobal/bls12381-key-pair/blob/master/src/Bls12381G2KeyPair.ts#L374

packages/core/src/crypto/Jwk.ts
```diff
@@ -0,0 +1,84 @@
...
+    }
+  }
+
+  if (key.keyType === KeyType.P256 || key.keyType === KeyType.P384 || key.keyType === KeyType.P521) {
```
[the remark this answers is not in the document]
TimoGlastra: Not really right? We could just always output the compressed variant, but support both variants the other way around

packages/core/src/crypto/JwkTypes.ts
```diff
@@ -1,6 +1,157 @@
...
 export interface Jwk {
   kty: 'EC' | 'OKP'
-  crv: 'Ed25519' | 'X25519' | 'P-256' | 'P-384' | 'secp256k1'
+  crv: 'Ed25519' | 'X25519' | 'P-256' | 'P-384' | 'P-521' | 'Bls12381G1' | 'Bls12381G2'
   x: string
   y?: string
+  use?: 'sig' | 'enc'
+}
+
+export interface Ed25519Jwk {
```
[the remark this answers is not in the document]
TimoGlastra: Hmmm. Lemme check

packages/core/src/crypto/JwkTypes.ts
TimoGlastra: Good point. So you mean call it `isX25519PublicKeyJwk`?

packages/core/src/crypto/JwkTypes.ts
```diff
@@ -1,6 +1,157 @@
+import { KeyType } from './KeyType'
+
 export interface Jwk {
   kty: 'EC' | 'OKP'
-  crv: 'Ed25519' | 'X25519' | 'P-256' | 'P-384' | 'secp256k1'
+  crv: 'Ed25519' | 'X25519' | 'P-256' | 'P-384' | 'P-521' | 'Bls12381G1' | 'Bls12381G2'
```
karimStekelenburg, line 5: Personal preference: I'd instead define a separate union type with this amount of options. If possible, derived from the `KeyType` enum.

packages/core/src/crypto/JwkTypes.ts
```diff
@@ -1,6 +1,157 @@
...
   x: string
   y?: string
+  use?: 'sig' | 'enc'
```
karimStekelenburg, line 8: What do we do if the `use` param isn't provided? According to the spec we can enforce it:


[4.2](https://www.rfc-editor.org/rfc/rfc7517#section-4.2).  "use" (Public Key Use) Parameter

   The "use" (public key use) parameter identifies the intended use of
   the public key.  The "use" parameter is employed to indicate whether
   a public key is used for encrypting data or verifying the signature
   on data.

   Values defined by this specification are:

   o  "sig" (signature)
   o  "enc" (encryption)

   Other values MAY be used.  The "use" value is a case-sensitive
   string.  Use of the "use" member is OPTIONAL, **unless the application
   requires its presence.**

karimStekelenburg: I think this depends on how purely we want to follow the spec. If we go 100%, we need to allow arbitrary string values as well. If not, I think we should maybe enforce the user to specify its value?

packages/core/src/crypto/KeyType.ts
```diff
@@ -4,4 +4,7 @@ export enum KeyType {
   Bls12381g1 = 'bls12381g1',
   Bls12381g2 = 'bls12381g2',
   X25519 = 'x25519',
+  P256 = 'p256',
+  P384 = 'p384',
+  P521 = 'p521',
```
karimStekelenburg, line 9: Why are the values here different from the spec (e.g. `p256` vs `P-256`)?

packages/core/src/modules/dids/methods/jwk/DidJwk.ts
```diff
@@ -0,0 +1,55 @@
...
+import { getDidJwkDocument } from './didJwkDidDocument'
+
+export class DidJwk {
+  public readonly did: string
+
+  private constructor(did: string) {
+    this.did = did
+  }
+
+  public get forEncrypting() {
```
karimStekelenburg, line 16: Naming seems a bit vague to me here. Maybe `canEncrypt()` and `canSign()` or something similar?

packages/core/src/modules/dids/domain/key-type/keyDidJsonWebKey.ts
```diff
@@ -0,0 +1,20 @@
+import type { KeyDidMapping } from './keyDidMapping'
+import type { VerificationMethod } from '../verificationMethod'
+
+import { Key } from '../../../../crypto'
+import { AriesFrameworkError } from '../../../../error'
+import { getJsonWebKey2020VerificationMethod } from '../verificationMethod'
+import { VERIFICATION_METHOD_TYPE_JSON_WEB_KEY_2020, isJsonWebKey2020 } from '../verificationMethod/JsonWebKey2020'
+
+export const keyDidJsonWebKey: KeyDidMapping = {
```
karimStekelenburg, line 9: Not something for now, but I personally find these kinds of objects that contains various attributes and functions confusing. I know this is normal in the world of JS, but I don't think it's in line with the rest of the code base.

packages/core/src/crypto/EcCompression.ts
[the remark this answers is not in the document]
karimStekelenburg: I'm not sure why you need the `Number` class here, so I may be wrong, but can't you use the `.valueOf()` method instead of casting it to `any`?

packages/core/src/crypto/EcCompression.ts
```diff
@@ -0,0 +1,104 @@
+/**
```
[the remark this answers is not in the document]
TimoGlastra: copy pasta

packages/core/src/crypto/JwkTypes.ts
TimoGlastra: I can extract into an union type. Can't derive it from KeyType, as the values are different. (everyone defines it a just a little bit dfiferent :) )

packages/core/src/crypto/Jwk.ts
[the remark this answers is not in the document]
karimStekelenburg: AFAIK the JWK spec does not allow for compressed values.

packages/core/src/crypto/JwkTypes.ts
TimoGlastra: For now, we just use it in parsing. It's required to be understood (not required to be present) for did:jwk. So we use when handling jwks, but only when we parse a JWK from someone else. When we create a JWK, for now I just always use the full capabilities of a key, and we don't include the `use` key. That's to to keep it 'simpler' (support the spec, but don't expose all functionality on our side).

I can update it to be a 'enc' | 'sig' | string value

packages/core/src/crypto/KeyType.ts
TimoGlastra: Because we started with defining key types as lowercase in AFJ, and that's what we're now doing for consistency I guess 🤷

packages/core/src/crypto/EcCompression.ts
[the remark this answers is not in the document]
karimStekelenburg: > copy pasta

Not sure what you mean here. My point was that creating and maintaining a separate package doesn't make sense if it's AFJ specific anyway.

packages/core/src/modules/dids/methods/jwk/DidJwk.ts
TimoGlastra: Yeah maybe `allowsEncrypting` and `allowsSigning`? Think we should differentiate between can and allowed to

packages/core/src/modules/dids/domain/key-type/keyDidJsonWebKey.ts
TimoGlastra: In this particular case, I'm also not a fan of how it's implemented. But I do think it's quite a normal pattern to use, and don't see anything wrong with it. If you're only interested in static methods, an object that implements an interface is fine I think. 

Would you prefer to refactor this to a class?

packages/core/src/crypto/Jwk.ts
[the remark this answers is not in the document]
TimoGlastra: Hmm in that case, there's no reason to support the uncompressed variant I think? did:key only supports compressed, and the jwk will have separate values (x and y), not a single concatenated public key

packages/core/src/crypto/JwkTypes.ts
karimStekelenburg: If you mean that we're using different string values in AFJ, I'd say we need to make that consistent at some point. If you mean that other libs are using different values, I'd say we need to create transformer functions for those. Having multiple types that have the same semantic meaning, but are syntactically different just causes confusion, I think.

packages/core/src/crypto/KeyType.ts
karimStekelenburg: This is probably related to the discussion [here](https://github.com/hyperledger/aries-framework-javascript/pull/1446/files/2104c0da82db470390399c8e3777d5d1c60132ab#r1181893677)

packages/core/src/crypto/Jwk.ts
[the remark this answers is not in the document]
karimStekelenburg: Agreed. We can always add uncompressed later on, if needed.

packages/core/src/crypto/JwkTypes.ts
karimStekelenburg: What do you mean by "required to be understood"?

I'm fine with keeping it as is as long as we actually use it. Adding the `string` type without adding logic to handle that doesn't make sense to me.

I was just wondering why we include it, as it's optional, only used for incoming JWKs, and only used by the `forEncrypting` and `forSigning` getters. Can we actually use those getters (as the `use` value is optional)?

packages/core/src/modules/dids/domain/key-type/keyDidJsonWebKey.ts
karimStekelenburg: I agree with your point, but I just think it's very different from the rest of the code base. I would prefer refactoring to a class, purely out of consistency and readability reasons.

This is a personal preference, though. I'm not a fan of the JS way of things ;)

packages/core/src/modules/dids/methods/jwk/DidJwk.ts
karimStekelenburg: YES! Even better!

packages/core/src/crypto/KeyType.ts
karimStekelenburg: > Because we started with defining key types as lowercase in AFJ, and that's what we're now doing for consistency I guess 🤷

If we aim for consistency, I'd say we follow the specs 😉. That way, people who are new to AFJ but are familiar with the specs will recognize things faster.

packages/core/src/crypto/Jwk.ts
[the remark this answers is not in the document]
karimStekelenburg: The spec has been updated quite recently (2 months ago). Both the MATTR libs you've mentioned haven't been updated in 2 years, and the W3C one had its last update more than a year ago. I'd stick to the spec.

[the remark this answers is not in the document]
karimStekelenburg: Maybe @tplooker knows the right approach here?

packages/core/src/crypto/EcCompression.ts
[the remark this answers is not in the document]
blu3beri: Yeah I don't think extracting ourselves and maintaining it would be the way to go.

More just checking if there is a small library that does something similar as this is normally not handled in AFJ itself and currently I don't we have the expertise to back this implementation up.

packages/core/src/crypto/Jwk.ts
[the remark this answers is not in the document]
blu3beri: > AFAIK the JWK spec does not allow for compressed values.

The JWK spec does very much support compressed and uncompressed. We are talking specifically about key representation and not the compressed format of a jwt, etc.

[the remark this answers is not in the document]
blu3beri: > I think the did:key spec only supports the compressed variant (that's what I understand from the thread), but you're saying the JWK itself could contain both compressed and uncompressed variants? 

No, I mean that we should check whether it is compressed or uncompressed, based on the prefix, and handle accordingly. We will break keys if this mechanism is implemented incorrectly which will lead to a lot of errors. 

> I think we could just support uncompressed as well (as it's just skipping the compression/uncompression and thus simpler). 
 
Yes! I don't think it's too much effort. 


> We'll always encode to compressed variant (because why not), but for decompression I should just look if the key is prefixed with `0x3` or `0x2` and if so, remove that previx and not do the decompression?

Yes, I would have to look at the exact algo for this, but I think that's it.

packages/core/src/crypto/Jwk.ts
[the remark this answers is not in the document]
blu3beri: The compressed variant, using only the X coordinate for the representation needs a prefix in the jwk of 0x03 or 0x02 I think depending on wether the y coordinate is odd or even (but that's from memory, would need a source but I am fairly sure). If we encode it without a prefix it will be interpreted incorrectly by the receiver.

packages/core/src/crypto/keyUtils.ts
blu3beri: They have to be rounded up, as a side note.

packages/core/src/crypto/Jwk.ts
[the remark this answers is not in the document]
TimoGlastra: I've removed BLS key representation for now. We can add it in a follow-up PR when we know more

packages/core/src/crypto/JwkTypes.ts
TimoGlastra: Yes maybe. I wouldn't jump to use JWK as leading, but probably when we refactor the wallet interface and make it more generic, we can have better definitions of key types, algorithms etc..

packages/core/src/crypto/JwkTypes.ts
TimoGlastra: > I was just wondering why we include it

Because it's required by the `did:jwk` spec:

> If the JWK contains a use property with the value "sig" then the keyAgreement property is not included in the DID Document. If the use value is "enc" then only the keyAgreement property is included in the DID Document.
> The JWK should have the appropriate use value set to match the capabilities of the specified crv. For example, the 
curve ed25519 is only valid for "sig" use and X25519 is only valid for "enc" (see [RFC 8037](https://datatracker.ietf.org/doc/html/rfc8037) and the second example below).
> The JWK may contain additional custom properties and values which will be accessible only in the verificationMethod. Any additional properties other than use (as documented above) are not referenced or used in the generation of the DID Document.


So we use it in the minimal way possible. When we parse a `did:jwk` we do the following:
1. If no `use` property is present, we check the default capabilities of a key type (enc, sig or both)
2. If a `use` property is present, we check if the value is valid for the key type (ed25519 can't have `enc`). If it's valid we only inlcude the verificadtionMethod types that link to sig vs enc. So if it's enc, only the keyAgreement will be added to the resolved did:jwk document

packages/core/src/modules/dids/domain/key-type/keyDidJsonWebKey.ts
TimoGlastra: Sure we can refactor to a class. Just need to think about better naming for this paradigm

packages/core/src/crypto/Jwk.ts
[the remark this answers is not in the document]
TimoGlastra: @blu3beri I've addressed all comments except the ones related to compression of P keys. Can you help me out a bit in pointing out what I need to exactly?

packages/core/src/crypto/EcCompression.ts
```diff
@@ -0,0 +1,101 @@
...
+  const yHex = publicKeyHex.slice(publicKeyHex.length / 2, publicKeyHex.length)
+  const xOctet = Uint8Array.from(Buffer.from(xHex, 'hex'))
+  const yOctet = Uint8Array.from(Buffer.from(yHex, 'hex'))
+  return compressECPoint(xOctet, yOctet)
+}
+
+export function expand(publicKey: Uint8Array, curve: 'P-256' | 'P-384' | 'P-521'): Uint8Array {
+  const publicKeyComponent = Buffer.from(publicKey).toString('hex')
+  const { prime, b, pIdent } = getConstantsForCurve(curve)
+  const signY = new Number(publicKeyComponent[1]).valueOf() - 2
```
blu3beri, line 86: Also happens here.

blu3beri (approved):
Great work @TimoGlastra! I see the prefixes of compressed and uncompressed are dealt with by code I did not see :). LGTM!

PR #1451, opened 2023-05-10
feat(indy-vdr): schema + creddef endorsement

2mau:


packages/indy-vdr/tests/indy-vdr-anoncreds-registry.e2e.test.ts
```diff
@@ -33,43 +57,55 @@ const agent = new Agent({
...
-    const didIndySchemaId = `did:indy:pool:localtest:TL1EaPFCZ8Si5aUrqScBDt/anoncreds/v0/SCHEMA/test/${dynamicVersion}`
+    const legacySchemaId = `DJKobikPAaYWAu9vfhEEo5:2:test:${dynamicVersion}`
+    const didIndySchemaId = `did:indy:pool:localtest:DJKobikPAaYWAu9vfhEEo5/anoncreds/v0/SCHEMA/test/${dynamicVersion}`
 
-    const schemaResult = await indyVdrAnonCredsRegistry.registerSchema(agent.context, {
-      options: {},
+    const schemaResult = await indyVdrAnonCredsRegistry.registerSchema(endorser.context, {
+      options: {
+        endorserMode: 'internal',
+        endorserDid,
```
TimoGlastra, line 107: I think we should add a test case where endorserMode is `internal`, but where the endorserDid is different from the issuerId. Can't see if that's the case here?

TimoGlastra (changes requested):
This is great @2mau . Most comments I've left are straightforward to solve. I think main issue (and I'm not sure yet how to solve) is:

> This is indy specific code, and we should thus not add it here. Mainly because it means for other methods it's not possible to make a registrar that has an action of wait that does not require changes to the internals of the anoncres packages (the Api), instead of just needing to implement the AnonCredsRegistry interface.

PR #1453, opened 2023-05-10
feat: support more key types in jws service

TimoGlastra:
~~NOTE: this pr is dependant on #1446 and thus should be merged after that PR is merged.~~

This PR refactors the JWK and JWS implementation to have better support for JWK / JWA, and also support more algs/key types in the JWS service. Basically the JWS service now supports all JWKs supported in AFJ, although it's limited by the algs and key types supported by the wallet (for askar we support EdDsa and ES256)


packages/core/src/crypto/JwsService.ts
```diff
@@ -128,10 +137,10 @@ export class JwsService {
...
 
-    return { isValid: true, signerKeys: signerKeys }
+    return { isValid: true, signerKeys }
   }
 
-  private buildProtected(options: ProtectedHeaderOptions) {
+  private buildProtected(options: JwsProtectedHeaderOptions) {
     if (!options.jwk && !options.kid) {
       throw new AriesFrameworkError('Both JWK and kid are undefined. Please provide one or the other.')
     }
```
blu3beri, line 146: Is it fine if they are both supplied?

packages/core/src/crypto/JwsService.ts
```diff
@@ -140,22 +149,58 @@ export class JwsService {
...
+    }
+
+    // Jwk
+    if (protectedHeader.jwk) {
+      return getJwkFromJson(protectedHeader.jwk)
+    }
+
+    // Kid
+    if (protectedHeader.kid) {
+      if (!protectedHeader.kid.startsWith('did:')) {
```
blu3beri, line 173: Might be good to have a util in AFJ, if it does not already exist, like `isDid(potentialDid: string): bool`.

packages/core/src/crypto/JwsService.ts
```diff
@@ -140,22 +149,58 @@ export class JwsService {
...
+      const didResolver = agentContext.dependencyManager.resolve(DidResolverService)
+      const didDocument = await didResolver.resolveDidDocument(agentContext, protectedHeader.kid)
+
+      // This is a special case for Aries RFC 0017 signed attachments. It allows a did:key without a keyId to be used kid
+      // https://github.com/hyperledger/aries-rfcs/blob/main/concepts/0017-attachments/README.md#signing-attachments
+      if (protectedHeader.kid.startsWith('did:key:') && !protectedHeader.kid.includes('#')) {
+        return getJwkFromKey(DidKey.fromDid(protectedHeader.kid).key)
+      }
+
+      // TODO: allowedPurposes
```
blu3beri, line 188: What does this TODO mean? Would be quite difficult I think for someone else to pick this up.

packages/core/src/crypto/__tests__/JwsService.test.ts
```diff
@@ -1,28 +1,35 @@
 import type { AgentContext } from '../../agent'
 import type { Key, Wallet } from '@aries-framework/core'
 
-import { IndySdkWallet } from '../../../../indy-sdk/src'
-import { indySdk } from '../../../../indy-sdk/tests/setupIndySdkModule'
-import { getAgentConfig, getAgentContext } from '../../../tests/helpers'
+// eslint-disable-next-line import/no-extraneous-dependencies
+import '@hyperledger/aries-askar-nodejs'
```
blu3beri, line 5: Why is this required? Can we not just register it or let the `index.js` of that package deal with it?

packages/core/src/crypto/__tests__/JwsService.test.ts
```diff
@@ -1,28 +1,35 @@
...
+import { JwaSignatureAlgorithm } from '../jose/jwa'
+import { getJwkFromKey } from '../jose/jwk'
 import { SigningProviderRegistry } from '../signing-provider'
 
 import * as didJwsz6Mkf from './__fixtures__/didJwsz6Mkf'
 import * as didJwsz6Mkv from './__fixtures__/didJwsz6Mkv'
+import * as didJwszDnaey from './__fixtures__/didJwszDnaey'
 
-describe('JwsService', () => {
+describeRunInNodeVersion([18], 'JwsService', () => {
```
blu3beri, line 22: Should we add a comment why only node 18 is checked?

packages/core/src/crypto/jose/jwa/alg.ts
```diff
@@ -0,0 +1,38 @@
+export enum JwaSignatureAlgorithm {
```
blu3beri, line 1: Do we want to have all the IANA / IETF defined algorithms here or just that we support?

packages/core/src/crypto/jose/jwk/Ed25519Jwk.ts
```diff
@@ -0,0 +1,87 @@
...
+
+import { TypedArrayEncoder } from '../../../utils'
+import { KeyType } from '../../KeyType'
+import { JwaCurve, JwaKeyType } from '../jwa'
+import { JwaSignatureAlgorithm } from '../jwa/alg'
+
+import { Jwk } from './Jwk'
+import { hasKty, hasCrv, hasX, hasValidUse } from './validate'
+
+export class Ed25519Jwk extends Jwk {
```
blu3beri, line 12: Not too sure how, but it might be good to note somewhere that we only use public values inside the JWK. 

Some JWK libraries have a `toPublicMaterial` function which strips all the private values, like `d` here for instance. 

Just some docs would suffice I think.

packages/core/src/crypto/jose/jwk/Ed25519Jwk.ts
```diff
@@ -0,0 +1,87 @@
...
+  public readonly x: string
+
+  public constructor({ x }: { x: string }) {
+    super()
+
+    this.x = x
+  }
+
+  public get kty() {
```
blu3beri, line 21: we have a getter for the `kty` (key type) and `keyType`. I can imagine this causing some confusion, when to use which one. Do you think we should fix this or is it clear enough with the difference in name.

packages/core/src/crypto/jose/jwk/Jwk.ts
```diff
@@ -0,0 +1,40 @@
...
+export abstract class Jwk {
+  public abstract publicKey: Buffer
+  public abstract keyType: KeyType
+  public abstract supportedSignatureAlgorithms: JwaSignatureAlgorithm[]
+  public abstract supportedEncryptionAlgorithms: JwaEncryptionAlgorithm[]
+
+  public abstract kty: JwaKeyType
+  public use?: string
+
+  public toJson(): JwkJson {
```
blu3beri, line 22: Would it not make more sense to make this abstract?

packages/core/src/crypto/jose/jwk/P_256Jwk.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  public get crv() {
+    return JwaCurve.P_256 as const
+  }
+
+  public get keyType() {
+    return KeyType.P256
+  }
+
+  // NOTE: this is the compressed variant. We should add support for the uncompressed variant.
```
blu3beri, line 35: Can you make this a doctstring? I think this is important to know for the user.

packages/core/src/crypto/jose/jwk/P_256Jwk.ts
```diff
@@ -0,0 +1,102 @@
...
+  public get publicKey() {
+    const publicKeyBuffer = Buffer.concat([TypedArrayEncoder.fromBase64(this.x), TypedArrayEncoder.fromBase64(this.y)])
```
blu3beri, line 37: Shouldn't this be base64url?

packages/core/src/crypto/jose/jwk/P_521Jwk.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    return Buffer.from(compressedPublicKey)
+  }
+
+  public get supportedEncryptionAlgorithms() {
+    return []
+  }
+
+  public get supportedSignatureAlgorithms() {
+    return [JwaSignatureAlgorithm.ES512]
```
blu3beri, line 48: EsXXX also has a matching hashing algorithm. Should we also add that as a `supportedHashingAlgorithm?

ES256 -> Sha2-256
ES384 -> Sha2-384
ES521 -> Sha2-512

packages/core/src/crypto/keyUtils.ts
```diff
@@ -27,7 +27,37 @@ export function isValidPrivateKey(privateKey: Buffer, keyType: KeyType): boolean
...
     [KeyType.P521]: 66,
-  }
+  } as const
 
   return Buffer.isBuffer(privateKey) && privateKey.length === privateKeyLength[keyType]
 }
+
+export function isSigningSupportedForKeyType(keyType: KeyType): boolean {
+  const keyTypeSigningSupportedMapping = {
+    [KeyType.Ed25519]: true,
```
blu3beri, line 37: Do you think it wouldn't be better to define this on the Jwk class as a static method? We can keep this function but just check whether the supportedSigningAlgos has length > 1 and then we support signing. This would avoid having some duplicate code.

blu3beri (changes requested):
Left a couple of small comments again that should be relatively easy to resolve. Great work Timo!

packages/core/src/crypto/keyUtils.ts
TimoGlastra: Yes, I was thinking about it as well. Two limitations I see:
- This makes JWKs the source of truth, while our lowest layer is Key/KeyType
- That we don't support the alg, doesn't neccesarily mean it doesn't support signing / encrypting. So it can give incorrect results

But I do agree it would be cleaner

packages/core/src/crypto/jose/jwk/P_521Jwk.ts
TimoGlastra: Can you point me to the spec where it defines ES as a hashing algorithm by itself?

packages/core/src/crypto/jose/jwk/P_256Jwk.ts
TimoGlastra: the fromBase64 method supports both... A bit confusing I know.

packages/core/src/crypto/jose/jwk/P_256Jwk.ts
```diff
@@ -0,0 +1,102 @@
...
+import { TypedArrayEncoder, Buffer } from '../../../utils'
+import { KeyType } from '../../KeyType'
+import { JwaCurve, JwaKeyType } from '../jwa'
+import { JwaSignatureAlgorithm } from '../jwa/alg'
+
+import { Jwk } from './Jwk'
+import { compress, expand } from './ecCompression'
+import { hasKty, hasCrv, hasX, hasY, hasValidUse } from './validate'
+
+export class P_256Jwk extends Jwk {
```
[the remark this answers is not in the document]
TimoGlastra: Sure!

packages/core/src/crypto/jose/jwk/Ed25519Jwk.ts
TimoGlastra: Yup found it quite confusing as well, but didn't really know what to do. Any suggestions?

packages/core/src/crypto/jose/jwk/Jwk.ts
TimoGlastra: Well this covers the base of the JWK. For now we only support two fields, but we can extend this with more over time (and even add `additionalFields` that will be added). So I think having a base is nice, to not require all extensions to do the same

packages/core/src/crypto/jose/jwa/alg.ts
TimoGlastra: I would say we add them all, as that makes it easier to include them in the future. If no JWK / JWE / JWS refers to it, it wouldn't be able to be used.

but if you want, I can remove the unused ones

packages/core/src/crypto/__tests__/JwsService.test.ts
TimoGlastra: We need to import it, as we use askar here. But I can add this to the `setup.ts` file

packages/core/src/crypto/JwsService.ts
TimoGlastra: I'll extend it. It means that we should look at specific specific purposes from the did document (`authentication` vs `capabilityDelegation`, etc..).

But I think it's just authentication that we need here, so I'll pick that for now (better to be strict)

packages/core/src/crypto/JwsService.ts
TimoGlastra: No. this is handled in a different place. i'll see to refactor, as it's a bit weird. But i did it like this for a specific reason

TimoGlastra: Ah it seems they're already next to each other:

```
    if (!options.jwk && !options.kid) {
      throw new AriesFrameworkError('Both JWK and kid are undefined. Please provide one or the other.')
    }
    if (options.jwk && options.kid) {
      throw new AriesFrameworkError('Both JWK and kid are provided. Please only provide one of the two.')
    }
```

packages/core/src/crypto/jose/jwk/P_256Jwk.ts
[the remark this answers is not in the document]
TimoGlastra: Would you also want to rename the keys for all these algorithsm then?

```
export enum JwaEncryptionAlgorithm {
  RSA1_5 = 'RSA1_5',
  RSA_OAEP = 'RSA-OAEP',
  RSA_OAEP_256 = 'RSA-OAEP-256',
  A128KW = 'A128KW',
  A192KW = 'A192KW',
  A256KW = 'A256KW',
  dir = 'dir',
  ECDH_ES = 'ECDH-ES',
  ECDH_ES_A128KW = 'ECDH-ES+A128KW',
  ECDH_ES_A192KW = 'ECDH-ES+A192KW',
  ECDH_ES_A256KW = 'ECDH-ES+A256KW',
  A128GCMKW = 'A128GCMKW',
  A192GCMKW = 'A192GCMKW',
  A256GCMKW = 'A256GCMKW',
  PBES2_HS256_A128KW = 'PBES2-HS256+A128KW',
  PBES2_HS384_A192KW = 'PBES2-HS384+A192KW',
  PBES2_HS512_A256KW = 'PBES2-HS512+A256KW',
}
```

Becaues I've now changed the JwaCurve to also remove `_` to make it consistent, but then we should do it everywhere (consistency is important here I think, because otherwise we're going to have inconsistencies down the line)
```
export enum JwaCurve {
  P256 = 'P-256',
  P384 = 'P-384',
  P521 = 'P-521',
  Ed25519 = 'Ed25519',
  X25519 = 'X25519',
}
```

packages/core/src/crypto/jose/jwk/Ed25519Jwk.ts
TimoGlastra: I think my preference would be to include a `includePrivateKey: boolean` or somethign like that which by default has the value false.

However I'd like to not add it now, as we don't have modelled the private properties yet and then I'd need to implement that for all key types, although we never word with private keys in AFJ (at least not on this layer)

packages/core/src/crypto/jose/jwa/alg.ts
blu3beri: No, it's okay. Main point was to avoid confusion about what we support, but it should be fine.

packages/core/src/crypto/jose/jwk/Ed25519Jwk.ts
blu3beri: Not too sure. The only thing I could come up with is to rename `kty` to `keyRepresentation`, but I would like `kty` to stay the same as its a direct mapping to the `JWA` defined key type.

I think we can leave this for now and maybe just add a docstring above `kty` that says that this is the key type for as defined in the `JWA` spec or IANA

packages/core/src/crypto/jose/jwk/P_256Jwk.ts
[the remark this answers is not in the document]
blu3beri: Yeah I think renaming the algorithms is also fine just to stay inline with TypeScript enum naming conventions.

packages/core/src/crypto/jose/jwk/P_521Jwk.ts
blu3beri: Ah no you are correct, ignore my comment :).

blu3beri (approved):

PR #1459, opened 2023-05-17
test: tests for DidComm V2 message packing

Artemkaaas:
- Added tests for DidComm V2 message packing
- Fixed issue to pass test vectors from the spec:  https://identity.foundation/didcomm-messaging/spec/#appendix

**Issue with the current implementation**: we need to use DID Resolver to unpack messages properly:
- resolve sender key from `skid`
- resolve recipient key from `kid`

### DIFF
--- a/packages/askar/src/wallet/AskarWallet.ts
+++ b/packages/askar/src/wallet/AskarWallet.ts
@@ -1424,7 +1424,9 @@ export class AskarWallet implements Wallet {
   }
 
   private findCommonSupportedEncryptionKeys(recipientDidDocuments: DidDocument[], senderDidDocument?: DidDocument) {
-    const recipients = recipientDidDocuments.map((recipientDidDocument) => recipientDidDocument.agreementKeys)
+    const recipients = recipientDidDocuments.map(
+      (recipientDidDocument) => recipientDidDocument.dereferencedKeyAgreement
+    )
 
     if (!senderDidDocument) {
       return {
@@ -1433,7 +1435,7 @@ export class AskarWallet implements Wallet {
       }
     }
 
-    const senderAgreementKeys = senderDidDocument.agreementKeys
+    const senderAgreementKeys = senderDidDocument.dereferencedKeyAgreement
 
     let senderVerificationMethod: VerificationMethod | undefined
     const recipientsVerificationMethods: VerificationMethod[] = []
--- a/packages/askar/src/wallet/AskarWallet.ts
+++ b/packages/askar/src/wallet/AskarWallet.ts
@@ -1296,7 +1296,7 @@ export class AskarWallet implements Wallet {
     throw new AriesFrameworkError('Unable to open jwe: recipient key not found in the wallet')
   }
 
-  public async resolveRecipientKey({
+  private async resolveRecipientKey({
     kid,
     recipientDidDocuments,
   }: {
@@ -1317,7 +1317,7 @@ export class AskarWallet implements Wallet {
     }
   }
 
-  public resolveSenderKeys({
+  private resolveSenderKeys({
     skid,
     didDocument,
     apu,
--- a/packages/askar/src/wallet/AskarWallet.ts
+++ b/packages/askar/src/wallet/AskarWallet.ts
@@ -52,8 +52,8 @@ import {
   keyAlgFromString,
   KeyAlgs,
   Store,
+  Jwk,
 } from '@hyperledger/aries-askar-shared'
-import { Jwk } from '@hyperledger/aries-askar-shared/build/crypto/Jwk'
 import BigNumber from 'bn.js'
 import { inject, injectable } from 'tsyringe'
 
@@ -454,7 +454,8 @@ export class AskarWallet implements Wallet {
 
           const keyPublicBytes = key.publicBytes
           // Store key
-          await this.session.insertKey({ key, name: TypedArrayEncoder.toBase58(keyPublicBytes) })
+          const id = TypedArrayEncoder.toBase58(keyPublicBytes)
+          await this.session.insertKey({ key, name: id })
           key.handle.free()
           return Key.fromPublicKey(keyPublicBytes, keyType)
         } catch (error) {
@@ -751,7 +752,7 @@ export class AskarWallet implements Wallet {
         enc: encId,
         alg: wrapId,
       })
-        .setEpk(JsonEncoder.toString(epk.jwkPublic))
+        .setEpk({ ...epk.jwkPublic })
         .setApv([...recipientKeys].map((recipientKey) => recipientKey.fingerprint))
 
       // As per spec we firstly need to encrypt the payload and then use tag as part of the key derivation process
@@ -772,12 +773,9 @@ export class AskarWallet implements Wallet {
 
           // According to the spec `kid` MUST be a DID URI
           // https://identity.foundation/didcomm-messaging/spec/#construction
-          const recipientDidKey = new DidKey(recipientKey).did
-          const recipientKid = `${recipientDidKey}#${recipientKey.fingerprint}`
+          const recipientKid = new DidKey(recipientKey).didDocument.agreementKeys[0]?.id
 
           // Wrap the recipient key using ECDH-ES
-          // FIXME: according to the spec `tag` must be used for the wrapping but there is not such parameter
-          // https://identity.foundation/didcomm-messaging/spec/#ecdh-es-key-wrapping-and-common-protected-headers
           const encryptedKey = new EcdhEs({
             algId: jweBuilder.alg(),
             apu: jweBuilder.apu(),
@@ -832,8 +830,7 @@ export class AskarWallet implements Wallet {
       }
 
       // According to the spec `skid` MUST be a DID URI
-      const senderDidKey = new DidKey(senderKey).did
-      const senderKid = `${senderDidKey}#${senderKey.fingerprint}`
+      const senderKid = new DidKey(senderKey).didDocument.agreementKeys[0]?.id
 
       // Generated once for all recipients
       // https://identity.foundation/didcomm-messaging/spec/#ecdh-1pu-key-wrapping-and-common-protected-headers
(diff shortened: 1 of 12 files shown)


packages/askar/src/wallet/AskarWallet.ts
```diff
@@ -454,7 +454,8 @@ export class AskarWallet implements Wallet {
 
           const keyPublicBytes = key.publicBytes
           // Store key
-          await this.session.insertKey({ key, name: TypedArrayEncoder.toBase58(keyPublicBytes) })
+          const id = kid ?? TypedArrayEncoder.toBase58(keyPublicBytes)
```
TimoGlastra, line 457: Ideally we don't store keys by kid as that means we have different formats to store keys (which will get messy).

Maybe we can add the kid as metadata and tag to the key?

Otherwise we won't be able to find some keys by it's public key anymore, because we use a kid

Artemkaaas: I think this can be reverted once we get `DID Resolver` available inside of unpack function.
I added an ability to store keys by `kid` in order to write tests against test vectors from the specification.
Test data there uses `did:example:alice#key-p521-1` key references.

TimoGlastra: We can store it as a tag on the key right? 

Just want to make sure we're not changing the way ids are created / managed (we should probably change the publicKeyBase58 to be a tag as well)

But even then, if we create a key, then later add it to a did document, it wouldn't necessarily have a kid. 

Maybe we need to add something to the did documents we have stored locally so we can do the following:
- query DidRecord by kid
- extract public key from did document
- query for key

I think we should avoid using the did resolver in the wallet itself

Artemkaaas: Yeah. we can find a way to avoid using DidResolver for getting `recipient key` but we cannot for `sender key` as it is unknown (at least when we receive the first message). 

The current assumption is that the sender uses either `did:key` or `did:peer` methods. So the public key can be resolved from the `skid` fingerprint.  
But if some other method is used we need to resolve the DidDocument for the `skid`.`

Artemkaaas: As for `createKey` method changes: I can revert the change and mock values in the tests (similarly to the senderKey).  

`did:peer` method is used in the updated Alice/Faber demo workflow (sending ping message).  Keys (both ED25519 and X25519) are stored by their base58 representation.

TimoGlastra: > skid is inside of base64 encoded protected JWE field. In my opinion, it is not a good idea to parse JWE -> fetch skid -> resolve sender DidDoc and key before calling wallet.unpack function.
Currently, JWE is constructed and parsed in wallet.pack/upack methods.

I think that is fine to do. We should really extract JWE out of the wallet in the coming months (we've been adding a `crypto/jose` directory with `jwk`, `jwt`, `jwa`, `jws`. Adding JWE is the missing piece.

For now I think we should extract the kid on a higher layer and pass it in. OR something I've recently added to the jws code is a `jwkResolver`, which you can optionally pass and will be called when the key used to sign a message can't be determined (the JWS service only supports if the `jwk` properyt is used in the header, for all other purposes you need to pass a `jwkResolver`).

Maybe something like that would work for now. Then when we extract it, we could use the same approach, but as a separate JWE service along side the JWS service

packages/core/src/modules/dids/domain/DidDocument.ts
```diff
@@ -200,6 +200,14 @@ export class DidDocument {
     return recipientKeys
   }
 
+  public get agreementKeys(): Array<VerificationMethod> {
```
TimoGlastra, line 203: This duplicates the separate method `getAgreementKeys ` below. If we add this for the keyAgreements, we should add it for all types.

Maybe we can do something like this. This is similar to `derefernceKey`, and is more general purpose. We can extract `allPurposes` as it's also used by `derefenceKey`

```ts
public dereferenceVerificationMethods(purposes?: DidPurpose[] | DidPurpose) {
  const allPurposes: DidPurpose[] = [
      'authentication',
      'keyAgreement',
      'assertionMethod',
      'capabilityInvocation',
      'capabilityDelegation',
    ]

    const verificationMethods: VerificationMethod[] = []
    for (const purpose of purposes ?? allPurposes) {
      for (const verificationMethod of this[purpose] ?? []) {
        if (typeof key === 'string') {
          verificationMethods.push(this.dereferenceVerificationMethod(verificationMethod))
        } else {
          verificationMethods.push(verificationMethod)
        }
      }
    }

return verificationMethods
}
```

TimoGlastra (changes requested):
I think the resolving is something we really want to address. Currently it relies on a very specific format of the skid. I think extract the extraction of the skid to a higher layer is the best for now

packages/askar/src/wallet/AskarWallet.ts
```diff
@@ -772,12 +773,9 @@ export class AskarWallet implements Wallet {
 
           // According to the spec `kid` MUST be a DID URI
           // https://identity.foundation/didcomm-messaging/spec/#construction
-          const recipientDidKey = new DidKey(recipientKey).did
-          const recipientKid = `${recipientDidKey}#${recipientKey.fingerprint}`
+          const recipientKid = new DidKey(recipientKey).didDocument.agreementKeys[0]?.id
```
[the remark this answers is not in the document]
Artemkaaas: Fixed. Updated wallet pack parameters to accept sender and recipient Did Documents in case of did comm v2. So that we can fetch the correct id of the used key.

packages/core/src/modules/dids/domain/DidDocument.ts
Artemkaaas: Removed duplicate method.

Artemkaaas:
https://github.com/hyperledger/aries-framework-javascript/pull/1459#pullrequestreview-1442640974

I updated the interfaces of wallet pack/unpack methods. Now it is mandatory to pass sender and recipient Did Documents when calling these methods in case of did comm v2 usage. So that we can fetch the correct key id and support all DID methods.

packages/core/src/modules/dids/domain/DidDocument.ts
TimoGlastra: My comment is still valid. We only have this method for keyAgreement keys now, not for the other purposes. I think we should make the methods on the did document generic (so it works with all purposes)

yarn.lock
```diff
@@ -10371,6 +10371,16 @@ ref-napi@3.0.3, "ref-napi@^2.0.1 || ^3.0.2", ref-napi@^3.0.3:
...
+"ref-napi@npm:@2060.io/ref-napi@^3.0.4":
+  version "3.0.4"
+  resolved "https://registry.yarnpkg.com/@2060.io/ref-napi/-/ref-napi-3.0.4.tgz#6a78093b36e8f4ffeb750f706433869382e73eb1"
+  integrity sha512-Aqf699E+DKs2xANx8bSkuzXyG9gcZ2iFkAk1kUTA8KbV5BSPQtIcBJexzohSRi9QWDhP4X54NLm7xwGA0UNCdQ==
+  dependencies:
+    debug "^4.1.1"
+    get-symbol-from-current-process-h "^1.0.2"
+    node-addon-api "^3.0.0"
+    node-gyp-build "^4.2.1"
+
```
TimoGlastra, line 10383:
```suggestion
```

packages/askar/src/wallet/AskarWallet.ts
```diff
@@ -1195,14 +1286,69 @@ export class AskarWallet implements Wallet {
...
+    const recipientAskarKey = await this.session.fetchKey({
+      name: recipientKey.publicKeyBase58,
+    })
+    return {
+      recipientKey,
+      recipientAskarKey: recipientAskarKey?.key,
+    }
+  }
+
+  public resolveSenderKeys({
```
TimoGlastra, line 1320: these can be private right?

TimoGlastra (approved):
Some final comments, but changes look good to me @Artemkaaas! Thanks for taking the time to address all comments

packages/core/src/wallet/Wallet.ts
```diff
@@ -69,9 +88,23 @@ export interface UnpackedMessageContext {
...
   senderKey?: Key | null
   envelopeType?: EnvelopeType
 }
+
+export type WalletPackV2Options = {
+  didCommVersion: DidCommMessageVersion
+  recipientDidDocuments: DidDocument[]
+  senderDidDocument?: DidDocument | null
+  envelopeType?: EnvelopeType
+}
```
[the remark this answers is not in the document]
Artemkaaas: Agree. This also was my initial thought. But currently, the `Key` structure is mainly used without binding to `kid`across the code. And it provides multiple constructors untied to `kid`.
So the `kid` will be an optional field that can be confusing. 
This is why I decided to pass the whole document for now.

packages/askar/src/wallet/AskarWallet.ts
Artemkaaas: I initially made it public for mocking in tests, but now tests can be written without using mock.

packages/core/src/modules/dids/domain/DidDocument.ts
Artemkaaas: Added suggested method. This method is almost the duplication of `dereferenceKey`...

TimoGlastra: agreed. maybe we can reuse them at some point

packages/core/src/modules/dids/domain/DidDocument.ts
```diff
@@ -201,11 +227,23 @@ export class DidDocument {
...
+  public get assertionMethods(): Array<VerificationMethod> {
+    return this.dereferenceVerificationMethods(['assertionMethod'])
+  }
+
+  public get capabilityInvocations(): Array<VerificationMethod> {
+    return this.dereferenceVerificationMethods(['capabilityInvocation'])
+  }
+
+  public get capabilityDelegations(): Array<VerificationMethod> {
+    return this.dereferenceVerificationMethods(['capabilityDelegation'])
```
TimoGlastra, line 246: This is a bit confusing to me. adding the `s` to make it derefenced is not really clear. Mabe we can just keep it at the single `dereferenceVerificationMethods` method and use it like that? 
Otherwise I think we should call it `getDereferencedAuthentication`, `getDerferencedAssertionMethod`, etc..

TimoGlastra:
Ready to merge, timo will give a final review

TimoGlastra (approved):

PR #1467, opened 2023-05-24
feat: openid4vci draft 11 support

berendsliedrecht:
draft PR to update the openid4vci-client implementation to support Draft 11 of the openid4vci specification.

Currently there are some issues with the authorized code flow. Mainly the query parameters are not formatted correctly which should be fixed in the library we are using.

edit: fix for the authorization parameter issue: https://github.com/Sphereon-Opensource/OID4VCI/pull/50

Everything passes now (with my local build of the OIDC4VC library) so we just have to wait for an alpha release or stable.

### DIFF
--- a/packages/openid4vc-client/package.json
+++ b/packages/openid4vc-client/package.json
@@ -3,9 +3,7 @@
   "main": "build/index",
   "types": "build/index",
   "version": "0.3.3",
-  "files": [
-    "build"
-  ],
+  "files": ["build"],
   "license": "Apache-2.0",
   "publishConfig": {
     "access": "public"
@@ -25,7 +23,8 @@
   },
   "dependencies": {
     "@aries-framework/core": "0.3.3",
-    "@sphereon/openid4vci-client": "^0.4.0",
+    "@sphereon/oid4vci-client": "file:../../../OID4VCI/packages/client",
+    "@sphereon/oid4vci-common": "file:../../../OID4VCI/packages/common",
     "@stablelib/random": "^1.0.2"
   },
   "devDependencies": {
--- a/packages/openid4vc-client/src/OpenId4VcClientService.ts
+++ b/packages/openid4vc-client/src/OpenId4VcClientService.ts
@@ -1,5 +1,5 @@
 import type { AgentContext } from '@aries-framework/core'
-import type { AccessTokenResponse, EndpointMetadata, Jwt } from '@sphereon/openid4vci-client'
+import type { AccessTokenResponse, EndpointMetadata, Jwt } from '@sphereon/oid4vci-common'
 
 import {
   AriesFrameworkError,
@@ -19,18 +19,13 @@ import {
   W3cCredentialService,
   W3cVerifiableCredential,
 } from '@aries-framework/core'
-import {
-  Alg,
-  AuthzFlowType,
-  CodeChallengeMethod,
-  CredentialRequestClientBuilder,
-  OpenID4VCIClient,
-  ProofOfPossessionBuilder,
-} from '@sphereon/openid4vci-client'
+import { Alg, AuthzFlowType, CodeChallengeMethod } from '@sphereon/oid4vci-common'
+import { CredentialRequestClientBuilder, OpenID4VCIClient, ProofOfPossessionBuilder } from '@sphereon/oid4vci-client'
 import { randomStringForEntropy } from '@stablelib/random'
 
 export interface PreAuthCodeFlowOptions {
-  issuerUri: string
+  clientId: string
+  openidCredentialOffer: string
   kid: string
   verifyRevocationState: boolean
 }
@@ -43,8 +38,8 @@ export interface AuthCodeFlowOptions extends PreAuthCodeFlowOptions {
 }
 
 export enum AuthFlowType {
-  AuthorizationCodeFlow,
-  PreAuthorizedCodeFlow,
+  AuthorizationCodeFlow = 'authorized code flow',
+  PreAuthorizedCodeFlow = 'pre-authorized code flow',
 }
 
 export type RequestCredentialOptions = { flowType: AuthFlowType } & PreAuthCodeFlowOptions &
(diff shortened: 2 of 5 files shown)


packages/openid4vc-client/src/OpenId4VcClientService.ts
```diff
@@ -54,7 +49,7 @@ export type RequestCredentialOptions = { flowType: AuthFlowType } & PreAuthCodeF
 // because we assume it will always be SHA256
 // as clear text code_challenges are unsafe
 export interface GenerateAuthorizationUrlOptions {
-  initiationUri: string
+  openidCredentialOffer: string
```
blu3beri, line 52: Not too sure on this name. I did not want to call it `credentialOfferUri` as that is also a property inside the credential offer parameters. I think that would cause confusion, but I don't think that this name is the best we can do.

packages/openid4vc-client/src/OpenId4VcClientService.ts
```diff
@@ -147,18 +147,26 @@ export class OpenId4VcClientService {
     }
   }
 
-  private assertCredentialHasFormat(format: string, scope: string, metadata: EndpointMetadata) {
-    if (!metadata.openid4vci_metadata) {
+  private assertCredentialHasFormat(supportedFormats: Array<string>, scope: string, metadata: EndpointMetadata) {
```
blu3beri, line 150: Not needed for draft 11 support, but with the JWT credentials coming up, I assume we want to support them as well.

packages/openid4vc-client/src/OpenId4VcClientService.ts
```diff
@@ -205,7 +214,7 @@ export class OpenId4VcClientService {
   }
 
   public async requestCredential(agentContext: AgentContext, options: RequestCredentialOptions) {
-    const credentialFormat = 'ldp_vc'
+    const supportedCredentialFormats = ['ldp_vc']
```
blu3beri, line 217: Should this be configured by the user of this module? Might be useful to specify when doing a call of in the module configuration. Open for suggestions.

packages/openid4vc-client/tests/fixtures.ts
```diff
@@ -15,63 +22,50 @@ export const getMetadataResponse = {
...
-      formats: {
-        ldp_vc: {
-          name: 'JFF x vc-edu PlugFest 2',
-          description: "MATTR's submission for JFF Plugfest 2",
-          types: ['OpenBadgeCredential'],
-          binding_methods_supported: ['did'],
-          cryptographic_suites_supported: ['Ed25519Signature2018'],
-        },
-      },
+  credentials_supported: [
```
blu3beri, line 25: Restructured based on the specification. Would be nice if we can get some more official fixtures from other implementations using draft 11.

packages/openid4vc-client/tests/openid4vc-client.e2e.test.ts
```diff
@@ -98,99 +109,57 @@ describe('OpenId4VcClient', () => {
...
-        .get('/.well-known/openid-credential-issuer')
-        .reply(200, getMetadataResponse)
+      expect(url.origin).toStrictEqual('https://launchpad.vii.electron.mattrlabs.io')
+      expect(url.pathname).toStrictEqual('/oidc/v1/auth/authorize')
 
-      const clientId = 'test-client'
-      const redirectUri = 'https://example.com/cb'
-      const initiationUri =
-        'openid-initiate-issuance://?issuer=https://launchpad.mattrlabs.com&credential_type=OpenBadgeCredential'
+      expect(params.get('response_type')).toBe('code')
```
blu3beri, line 140: This errors now because the params of the authorizationUrl are incorrect.

Right now the entire JSON body is the key and the value is an empty string. It should be fixed by making every key value in the JSON object a separate query parameter key-value pair.

packages/openid4vc-client/src/OpenId4VcClientService.ts
TimoGlastra: What does the sphereon library use?

packages/openid4vc-client/src/OpenId4VcClientService.ts
TimoGlastra: I think we should set the default to all supported, and the user can overwrite it.

TimoGlastra: I'm probably going to take over some of the things in this PR, as I'm going to work on adding JWT VC support to the openid client now

packages/openid4vc-client/src/OpenId4VcClientService.ts
blu3beri: Just `uri` which I think it not very descriptive. But it might be fine as I think it is done for backwards compatibility with 8/9 which uses initiateUri instead of credentialOfferUri.

nklomp: That is indeed the reason for it

blu3beri: Thank you for the confirmation @nklomp. I will update the field to be `uri` as well.

nklomp: Also the context of the object it is a property of should be helpful enough to be clear about what the URI is about

packages/openid4vc-client/tests/openid4vc-client.e2e.test.ts
nklomp: should be fixed in latest feature/issuer commit

packages/openid4vc-client/package.json
```diff
@@ -25,7 +23,8 @@
   },
   "dependencies": {
     "@aries-framework/core": "0.3.3",
-    "@sphereon/openid4vci-client": "^0.4.0",
+    "@sphereon/oid4vci-client": "file:../../../OID4VCI/packages/client",
+    "@sphereon/oid4vci-common": "file:../../../OID4VCI/packages/common",
```
TimoGlastra, line 27: Have you made changes to the openid4vci packages? There's an unstable version that has all the latest changes: [0.4.1-unstable.267](https://www.npmjs.com/package/@sphereon/oid4vci-client/v/0.4.1-unstable.267),

and I think sphereon is going to release a 0.5 version tomorrow-ish

nklomp: Yeah, 0.5 probably gonna be Thursday. The above JSON encoding (and some other v8 fixes) have been resolved last few days

blu3beri:
Closing in favor of #1474

Issue #1466, opened 2023-05-23
Improve JSON-LD VC integration in Aries Framework JavaScript


TimoGlastra:
The JSON-LD VC integration in AFJ is suboptimal, mainly because the dependency on forks from digitalbazaar for the various vc.js related logic that doesn't work in React Native.

There is an issue open in the VC library about react native support: https://github.com/digitalbazaar/vc/issues/109. The general answer is: PRs to improve RN support are welcome, official support for React Native is not on the roadmap however.

As the implementation of jsonld and jsonld signatures can be complex, using the libraries is desired, but improving the exeprience in AFJ is also desired. I think the following middle ground can achieve the best of boths worlds:
- Remove the usage of vc.js from AFJ as the library is only a few hunderd lines of code, and with the JWT VC implementation in AFJ we already have most of it covered. This allows us to provide a more streamlined API and validation interaface that is consistent.
- Make PRs to the rdf-canonize, http-client, jsonld-signatures and jsonld.js libraries that make it work in a react native environment without having to rely on forks.
  - There's not too much code needed to make it work in react native, and it probably mostly comes down to making things more pluggable where it depends on Node.JS or browser specific APIs.
- Implement signature suites directly in AFJ (as we already have with Ed25519Siganture2018 and BbsBlsSignature2020).

This way the API surface we use from digitalbazaars library is minimal (just signing and verification of jsonld signatures), and we have the VC layer implemented in AFJ itself.

Removing the need for forks means we stay up to date with latest changes, while still being able to run it in vc-js

PR #1468, opened 2023-05-24
feat(w3cCredentials)!: add jwt vc support

TimoGlastra:
Adds support for JWT VCs in addition to the already supported JSON-LD VCs. 

It became a bigger PR than anticipated, mainly because I moved some files around and made some changes to e.g. how verification methods are handled so they can be used without JSON-LD as well.

The PR has extracted the JSON-LD specific functionality from the `W3cCredentialService` into a new `W3cJsonLdCredentialService`, and we've got a new `W3cJwtCredentialService`. The general w3c credential service exposes a generic api to work with both, while the specific classes implement the logic. Records are only handled through the general w3c credential service. The idea is that users will mostly use the W3cCredentialService, and not the underlying imlementations, that is more to split up logic based on format.

The API has changed to now require a format when signing credentials / presentations:

```ts
w3cCredentialService.signCredential({
  format: 'jwt_vc',
  credential: /* the credential */,
  alg: 'ES256'
})
```

For the verification the API has also changed to support a more rich API. This rich API has been fully implemented for JWT VCs, but could not be fully supported for JSON-LD library due to the way errors are thrown in that library (see #1466, this will be addressed in a follow up pr).

The idea for the results is that now everything is handled by a custom validation object that has the following structure. The specific validations differ a bit between a VC and VP, but the idea is the same.

```ts
{
  isValid: true,
  validations: {
    // credential has no credentialStatus, so always valid
    credentialStatus: {
      isValid: true,
    },
    // This both validates whether the credential matches the
    // data model, as well as whether the credential is expired etc..
    dataModel: {
      isValid: true,
    },
    // This validates whether the signature is valid
    signature: {
      isValid: true,
    },
    // This validates whether the issuer is also the signer of the credential
    issuerIsSigner: {
      isValid: true,
    },
  },
};
```

For now, for JSON-LD verifications the result looks like this. This means the same data is available before this PR (the VC.js structure was followed), but a bit more nested. Over time we want to remove the use of vc.js and use the same structure as above for both credential formats. This allows for fine grained control over what you want to validate. 

```ts
{
  isValid: true,
  validations: {
    vcJs: {
      isValid: true,
      // results from the VC.JS library
      results: []
     }
  }
}
```


packages/core/src/crypto/JwsService.ts
```diff
@@ -87,30 +91,63 @@
...
    */
-  public async verifyJws(agentContext: AgentContext, { jws, payload }: VerifyJwsOptions): Promise<VerifyJwsResult> {
-    const base64Payload = TypedArrayEncoder.toBase64URL(payload)
-    const signatures = 'signatures' in jws ? jws.signatures : [jws]
+  public async verifyJws(agentContext: AgentContext, { jws, jwkResolver }: VerifyJwsOptions): Promise<VerifyJwsResult> {
+    let signatures: JwsDetachedFormat[] = []
+    let payload: string
+
+    if (typeof jws === 'string') {
+      if (!JWS_COMPACT_FORMAT.test(jws)) throw new AriesFrameworkError(`Invalid JWS compact format for value '${jws}'.`)
```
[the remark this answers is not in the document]
blu3beri: Might be good to fix this somehow. if it is really that much slower sending something like `a.b.-.--------------------------------------------------------------...----` would be bad for the hosting server / mobile device.

swcurran:
Awesome stuff, Timo. Any thought to how AnonCreds in W3C format could use this?  Would it be a matter of adding an “algo” for AnonCreds and a corresponding handler?  My expectation is that anoncreds-rs could detect W3C format AnonCreds objects as needed and convert internally, and on creation, be passed a flag to indicate the generated objects be converted to W3C form when needed.  Do you see that as possible?

packages/core/src/modules/dids/methods/jwk/didJwkDidDocument.ts
```diff
@@ -14,7 +14,7 @@ export function getDidJwkDocument(didJwk: DidJwk) {
   const parsed = parseDid(didJwk.did)
   const jwkJson = JsonEncoder.fromBase64(parsed.id)
 
-  const verificationMethod = getJsonWebKey2020VerificationMethod({
+  const verificationMethod = getJsonWebKey2020({
     did: didJwk.did,
     jwk: jwkJson,
     verificationMethodId: `${didJwk.did}#0`,
```
blu3beri, line 20: here we can use the `keyReference` method right?

packages/core/src/modules/vc/data-integrity/models/W3cJsonLdVerifiableCredential.ts
```diff
@@ -0,0 +1,42 @@
...
+  proof: SingleOrArray<LinkedDataProofOptions>
+}
+
+export class W3cJsonLdVerifiableCredential extends W3cCredential {
+  public constructor(options: W3cJsonLdVerifiableCredentialOptions) {
+    super(options)
+    if (options) {
+      this.proof = Array.isArray(options.proof)
+        ? options.proof.map((proof) => new LinkedDataProof(proof))
+        : new LinkedDataProof(options.proof)
```
blu3beri, line 22: If we don't we should have a utility method for this.

```typescript
function mapSingleOrArray<T>(v: SingleOrArray<t>, map: (i: T) => SingleOrArray<T>): SingleOrArray<T>
```

and maybe also a `mapSingleOrArrayToArray` which just does the same and always returns an array.

packages/core/src/modules/vc/data-integrity/models/W3cJsonLdVerifiableCredential.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  @LinkedDataProofTransformer()
+  @IsInstanceOrArrayOfInstances({ classType: LinkedDataProof })
+  @ValidateNested()
+  public proof!: SingleOrArray<LinkedDataProof>
+
+  public get proofTypes(): Array<string> {
+    const proofArray = asArray<LinkedDataProof>(this.proof)
+    return proofArray?.map((x) => x.type) ?? []
```
blu3beri, line 33:
```suggestion
    return proofArray?.map((proof) => proof.type) ?? []
```

packages/core/src/modules/vc/jwt-vc/W3cJwtCredentialService.ts
```diff
@@ -0,0 +1,530 @@
...
+      }
+    } else {
+      if (!signerId) {
+        throw new AriesFrameworkError(`JWT 'iss' MUST be present in payload when no 'kid' is specified`)
+      }
+
+      // Find the verificationMethod in the did document based on the alg and proofPurpose
+      const jwkClass = getJwkClassFromJwaSignatureAlgorithm(credential.jwt.header.alg)
+      if (!jwkClass) throw new AriesFrameworkError(`Unsupported JWT alg '${credential.jwt.header.alg}'`)
+      const { supportedVerificationMethodTypes } = getKeyDidMappingByKeyType(jwkClass.keyType)
```
blu3beri, line 500:
```suggestion
      if (!jwkClass) throw new AriesFrameworkError(`Unsupported JWT alg '${credential.jwt.header.alg}'`)
      
      const { supportedVerificationMethodTypes } = getKeyDidMappingByKeyType(jwkClass.keyType)
```

packages/openid4vc-client/src/OpenId4VcClientService.ts
blu3beri, line 1: Thanks for keeping the changes to a minimum here :D

blu3beri (approved):
Great PR! I see a lot of manual validation is happening. Is this something we might be able to extract into a non-AFJ library which can be used by anyone, also non AFJ-projects.

packages/core/src/modules/vc/data-integrity/models/W3cJsonLdVerifiableCredential.ts
TimoGlastra: Yeah that's a good one, this is messsssy

packages/core/src/crypto/JwsTypes.ts
```diff
@@ -11,21 +10,55 @@ export interface JwsProtectedHeaderOptions {
...
+
+  /**
+   * Base64url encoded signature
+   */
   signature: string
+
+  /**
+   * Base64url encoded protected header
+   */
   protected: string
```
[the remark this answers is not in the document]
TimoGlastra: This is a direct mapping of the format as used by the specification. In other places we call it `protectedHeader`, but if we change this, the JWS service will create invalid JWSes.

packages/core/src/crypto/jose/jwt/JwtPayload.ts
```diff
@@ -0,0 +1,217 @@
+import { AriesFrameworkError } from '../../../error'
+
+/**
+ * The maximum allowed clock skew time in seconds. If an time based validation
+ * is performed against current time (`now`), the validation can be of by the skew
+ * time.
+ */
+const DEFAULT_SKEW_TIME = 300
```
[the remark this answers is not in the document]
TimoGlastra: They can't change the value of this constant. But everywhere this constant is used, the user can provide a custom skew time value. (so this is only used if the user didn't provide a value)

packages/core/src/crypto/jose/jwt/JwtPayload.ts
```diff
@@ -0,0 +1,217 @@
...
+   *
+   * The following validations are performed:
+   *  - if `nbf` is present, it must be greater than now
+   *  - if `iat` is present, it must be less than now
+   *  - if `exp` is present, it must be greater than now
+   */
+  public validate(options?: { skewTime?: number; now?: number }) {
+    const { nowSkewedFuture, nowSkewedPast } = getNowSkewed(options?.now, options?.skewTime)
+
+    if (typeof this.nbf === 'number' && this.nbf > nowSkewedFuture) {
```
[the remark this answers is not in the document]
TimoGlastra: Yeah we already validate this in the `fromJson`, and we have type guards for when you create it in TS. But I can change it

packages/core/src/modules/vc/jwt-vc/W3cJwtCredentialService.ts
TimoGlastra: It seems there's not changed here except for the newline? Is that what you meant?

packages/core/src/modules/vc/models/credential/W3cVerifiableCredential.ts
```diff
@@ -1,52 +1,45 @@
...
+
+import { Transform, TransformationType } from 'class-transformer'
+import { ValidationError } from 'class-validator'
+
+import { AriesFrameworkError, ClassValidationError } from '../../../../error'
+import { JsonTransformer } from '../../../../utils'
+import { W3cJsonLdVerifiableCredential } from '../../data-integrity/models/W3cJsonLdVerifiableCredential'
+import { W3cJwtVerifiableCredential } from '../../jwt-vc/W3cJwtVerifiableCredential'
+
+const getCredential = (v: unknown) => {
```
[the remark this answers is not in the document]
TimoGlastra: Not really. The type we get from class-transformer is unknown

TimoGlastra:
> I see a lot of manual validation is happening. Is this something we might be able to extract into a non-AFJ library which can be used by anyone, also non AFJ-projects.

Yeah I think we can definitely extract the JWT-VC part to a separate library. Only thing is that we would need to replace the class-transformer/class-validator parts with something else

TimoGlastra:
> Any thought to how AnonCreds in W3C format could use this? Would it be a matter of adding an “algo” for AnonCreds and a corresponding handler?

@swcurran, for this specific addition that wouldn't be possible as this just focuses on JWT credentials using Json Web Signatures. For the AnonCreds mapping we need a custom `proof` object and type, which we don't support yet.

But for JSON-LD credentials that would exactly how it would work yes, We can add a signature suite called `AnonCredsProof2023`, or however it is called that will handle the signing and verification of the AnonCreds credential.

PR #1470, opened 2023-05-26
feat: Split AFJ demo for two cases: Main and DidCommV2

Artemkaaas:
Started work on splitting the Alice/Faber demo into smaller cases. Using this we can add demos for various optional modules.

- `Main` - general demo of issuing credentials and presenting the proof
- `DidComm V2` - demo reflecting supported didcomm v2 protocols


TimoGlastra:
I think overall LGTM. I'm not sure though on splitting it up. What if we add an option to create a v1 or v2 invitation? And on the other side we just do receive invitation and it will accept both? 

I think an agent should be able to run both at the same time, so I don't see a specific reason to split it up

Artemkaaas:
> I think overall LGTM. I'm not sure though on splitting it up. What if we add an option to create a v1 or v2 invitation? And on the other side we just do receive invitation and it will accept both?
> 
> I think an agent should be able to run both at the same time, so I don't see a specific reason to split it up

```
What if we add an option to create a v1 or v2 invitation? And on the other side we just do receive invitation and it will accept both?
```
This is how it's currently done in `hyperledger:feat/didcomm-v2`. 
I thought of splitting the main demo into independent cases so that we can create demos for conflicting cases such as: indy/askar wallets, indy/cheqd ledgers, cl/w3c credentials etc...

TimoGlastra (approved):

PR #1479, opened 2023-06-05
fix: encode tails url

### LINKED ISSUE #1478: Unable to fetch tails registry if the schema name contains spaces
When fetching the tails URL is the schema name has spaces in it iOS (not confirmed on Android) will fails because the URL is not recognized:

```console
Card:CL_ACCUM:41e28904-4584-4250-a274-3723f628256b
 ERROR  ERROR: Error while retrieving tails file from URL https://tails-test.vonx.io/2K2h7kf8VGTLtfoxJgWazf:4:2K2h7kf8VGTLtfoxJgWazf:3:CL:814555:Member Card:CL_ACCUM:41e28904-4584-4250-a274-3723f628256b {
....
    "code": "ENSURLERRORDOMAIN-1002",
    "message": "unsupported URL",
    "domain": "NSURLErrorDomain",
```

I think we should add `encodeURI()` just before line 69 [here](https://github.com/hyperledger/aries-framework-javascript/blob/efc0ddfd23290f99bfde6a3e27c4f1e682630b2e/packages/react-native/src/ReactNativeFileSystem.ts#L65). It can be added earlier in the calls stack to effect all other platforms but since nobody else is complaining it may be on RN only.

jleach:
Some schema names can have characters that are not supported by URL loaders without being encoded. This PR fixes this issue by url encoding before passing the URL off to the loading mechanics. I tested this change agains the issue that reports the problem and it does fix the problem.

Fixes #1478

### DIFF
--- a/packages/react-native/src/ReactNativeFileSystem.ts
+++ b/packages/react-native/src/ReactNativeFileSystem.ts
@@ -66,8 +66,11 @@ export class ReactNativeFileSystem implements FileSystem {
     // Make sure parent directories exist
     await RNFS.mkdir(getDirFromFilePath(path))
 
+    // Some characters in the URL might be invalid for
+    // the native os to handle. We need to encode the URL.
+    const encodedFromUrl = encodeURI(url)
     const { promise } = RNFS.downloadFile({
-      fromUrl: url,
+      fromUrl: encodedFromUrl,
       toFile: path,
     })
 


TimoGlastra (approved):
Do you know / have you verified the `encodeURI` method works in both JSC / Hermes?

jleach:
> encodeURI

@TimoGlastra Yes, I patched this change into my local module and tested that it fixed my problem with Bifold. Once confirmed I turned it into a PR.

PR #1491, opened 2023-06-26
feat(anoncreds): allow automatic create link secret

### LINKED ISSUE #1490: Add link secret to agent initalization
I propose we add the following code:

```JavaScript
  // If we don't have any link secrets yet, we will create a
  // default link secret that will be used for all anoncreds
  // credential requests.
  const linkSecretIds = await agent.modules.anoncreds.getLinkSecretIds()
  if (linkSecretIds.length === 0) {
    await agent.modules.anoncreds.createLinkSecret({
      setAsDefault: true,
    })
  }
```

to `Agent.initalize()` if `agent.modules.anoncreds` is not `null`. It's an important step and hard to debug why credentials cannot be accepted if this step was missed.

jleach:
Automatically create a default link secret for AnonCreds to make it more intuitive how setup and Agent to receive credential offers.

I don't know how to write a test for this so I created [a script](https://gist.github.com/jleach/8a956605216ee197ed9b1dbb1fd8d85a) that can be run to validate it creates a link secret and calls initialize. 

Fixed #1490

### DIFF
--- a/packages/anoncreds/src/AnonCredsModule.ts
+++ b/packages/anoncreds/src/AnonCredsModule.ts
@@ -1,5 +1,5 @@
 import type { AnonCredsModuleConfigOptions } from './AnonCredsModuleConfig'
-import type { DependencyManager, Module, Update } from '@aries-framework/core'
+import type { DependencyManager, AgentContext, Module, Update } from '@aries-framework/core'
 
 import { AnonCredsApi } from './AnonCredsApi'
 import { AnonCredsModuleConfig } from './AnonCredsModuleConfig'
@@ -36,6 +36,20 @@ export class AnonCredsModule implements Module {
     dependencyManager.registerSingleton(AnonCredsCredentialDefinitionPrivateRepository)
     dependencyManager.registerSingleton(AnonCredsKeyCorrectnessProofRepository)
     dependencyManager.registerSingleton(AnonCredsLinkSecretRepository)
+  }
+
+  public async initalize(agentContext: AgentContext) {
+    if (this.config.createDefaultLinkSecret === false) {
+      return
+    }
+
+    const api = agentContext.dependencyManager.resolve(AnonCredsApi)
+    const linkSecretIds = await api.getLinkSecretIds()
+    if (linkSecretIds.length === 0) {
+      await api.createLinkSecret({
+        setAsDefault: true,
+      })
+    }
   }
 
   public updates = [
--- a/packages/anoncreds/src/AnonCredsModule.ts
+++ b/packages/anoncreds/src/AnonCredsModule.ts
@@ -39,9 +39,7 @@ export class AnonCredsModule implements Module {
   }
 
   public async initialize(agentContext: AgentContext): Promise<void> {
-    if (this.config.autoCreateLinkSecret === false) {
-      return
-    }
+    if (!this.config.autoCreateLinkSecret) return
 
     const anoncredsApi = agentContext.dependencyManager.resolve(AnonCredsApi)
     const linkSecretIds = await anoncredsApi.getLinkSecretIds()
--- a/packages/anoncreds/src/AnonCredsModule.ts
+++ b/packages/anoncreds/src/AnonCredsModule.ts
@@ -39,7 +39,7 @@ export class AnonCredsModule implements Module {
   }
 
   public async initalize(agentContext: AgentContext) {
-    if (this.config.createDefaultLinkSecret === false) {
+    if (this.config.autoCreateLinkSecret === false) {
       return
     }
 
--- a/packages/anoncreds/src/AnonCredsModule.ts
+++ b/packages/anoncreds/src/AnonCredsModule.ts
@@ -38,15 +38,15 @@ export class AnonCredsModule implements Module {
     dependencyManager.registerSingleton(AnonCredsLinkSecretRepository)
   }
 
-  public async initalize(agentContext: AgentContext) {
+  public async initialize(agentContext: AgentContext): Promise<void> {
     if (this.config.autoCreateLinkSecret === false) {
       return
     }
 
-    const api = agentContext.dependencyManager.resolve(AnonCredsApi)
-    const linkSecretIds = await api.getLinkSecretIds()
+    const anoncredsApi = agentContext.dependencyManager.resolve(AnonCredsApi)
+    const linkSecretIds = await anoncredsApi.getLinkSecretIds()
     if (linkSecretIds.length === 0) {
-      await api.createLinkSecret({
+      await anoncredsApi.createLinkSecret({
         setAsDefault: true,
       })
     }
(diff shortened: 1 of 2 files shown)


packages/anoncreds/src/AnonCredsModule.ts
```diff
@@ -38,6 +38,20 @@ export class AnonCredsModule implements Module {
     dependencyManager.registerSingleton(AnonCredsLinkSecretRepository)
   }
 
+  public async initalize(agentContext: AgentContext) {
```
genaris, line 41:
```suggestion
  public async initialize(agentContext: AgentContext) {
```

Agent core checks for `initialize` method existence to call it during initialization process.

packages/anoncreds/src/AnonCredsModule.ts
```diff
@@ -38,6 +38,20 @@ export class AnonCredsModule implements Module {
...
+    if (this.config.autoCreateLinkSecret === false) {
```
genaris, line 42: Maybe could be the other way around? I.e. doing all the link secret stuff only if autoCreateLinkSecret is true?

packages/anoncreds/src/AnonCredsModule.ts
```diff
@@ -38,6 +38,20 @@ export class AnonCredsModule implements Module {
...
+      return
+    }
+
+    const api = agentContext.dependencyManager.resolve(AnonCredsApi)
```
genaris, line 46: I'd call this variable `anoncredsApi` just to be consistent with other parts of the framework where apis are resolved.

genaris (commented):
Some minor comments for the moment. It would be nice to add at least an unit test in anoncreds-rs (or update some tests there) to check everything goes fine, but It is looking good!

packages/anoncreds/src/AnonCredsModule.ts
jleach: I can change it if you prefer but I think its better as is. Usually in an early return, its more common to handle edge cases, errors, or invalid input at the start of a function,  which lets the main function logic be less indented and easier to read.

LMK.

packages/anoncreds/src/AnonCredsModule.ts
jleach: Fixed

packages/anoncreds/src/AnonCredsModule.ts
jleach: Fixed

packages/anoncreds/src/AnonCredsModule.ts
```diff
@@ -38,6 +38,20 @@ export class AnonCredsModule implements Module {
     dependencyManager.registerSingleton(AnonCredsLinkSecretRepository)
   }
 
+  public async initialize(agentContext: AgentContext): Promise<void> {
+    if (this.config.autoCreateLinkSecret === false) {
+      return
+    }
```
TimoGlastra, line 44:
```suggestion
    if (!this.config.autoCreateLinkSecret) return
```

TimoGlastra (changes requested):
This won't work as intended for multi-tenancy, as all tenants share the same modules. Module initialization is meant to set everything up for a module, not do any actions that have impact on a specific agent/tenant/database user. 

I'd also like to slim down on the usage of auto created objects on agent creation. We also had this for mediation and it's a bit weird to do all these setup things if you're not using it.

In this case, It'd then rather just create a link secret by default when you request a credential for the first time. This way we create things when needed

genaris:
> This won't work as intended for multi-tenancy, as all tenants share the same modules. Module initialization is meant to set everything up for a module, not do any actions that have impact on a specific agent/tenant/database user.

Oh I didn't think about multi-tenancy cases. In such case I'm not sure where it should be a 'clean' way of adding this logic. I'm thinking on two possibilities:

- Add it to `AnonCredsCredentialFormatService` and `LegacyIndyCredentialFormatService` [just before calling createCredentialRequest](https://github.com/hyperledger/aries-framework-javascript/blob/38a0578011896cfcf217713d34f285cd381ad72c/packages/anoncreds/src/formats/AnonCredsCredentialFormatService.ts#L241). It feels a bit weird to me to call `AnonCredsApi` from there, but I don't see how to do it in a different way (considering that the API it is the only place where link secrets are being stored in the repository)
- Add it to each `AnonCredsHolderService` implementation, for instance in `AnonCredsRsHolderService.createCredentialRequest()` ([here](https://github.com/hyperledger/aries-framework-javascript/blob/38a0578011896cfcf217713d34f285cd381ad72c/packages/anoncreds-rs/src/services/AnonCredsRsHolderService.ts#L201), where it determines that no default link secret is defined). The problem of this approach is that we'd need to do it for each anoncreds holder service implementation (`indy-sdk`, `anoncreds-rs`), which can lead to confusions. And we'd be also calling `AnonCredsApi` (which will call the Holder under the hood to create the link secret), but can also call link secret repo methods directly if we want to avoid it. 

Thoughts about any cleaner way to do it?

jleach:
> everything up for a module, not do any actions that have impact on a specific agent/tenant/database user.

@TimoGlastra I think we should at least throw in an option to log a warning if this is a more complicated issue. Right now, if this step is missed, then the logic just hangs fetching the transaction with seqNo nnnnnn. It's not obvs why. My original thinking was, since every (almost every?) anoncreds user will want this logic run why not just automate it for them.

TimoGlastra:
- current flow is not good for DX
- Hard to determine best place to auto-create
- We can implement in holder service (chance on race condition)
	- https://github.com/hyperledger/aries-framework-javascript/blob/38a0578011896cfcf217713d34f285cd381ad72c/packages/anoncreds-rs/src/services/AnonCredsRsHolderService.ts#L205-L210
- Or we need an initialization method for initialization a module for an agent context (tenant).
- If we implement in the holder service, we need to do it twice.

Decision: Do it in the holder service, as it's the more straightforward way to do it now. But we should look at a way to make this easier. We also have initialization issues with mediation.

TimoGlastra:
@jleach are you able to pick up the changes as discussed in the WG call?

genaris:
> @jleach are you able to pick up the changes as discussed in the WG call?

I did a sort of follow-up of this PR [here](https://github.com/hyperledger/aries-framework-javascript/compare/main...genaris:aries-framework-javascript:feat/auto-create-link-secret). I think this approach will work fine, but I'm not convinced on the fact of making `AnonCredsApi`/`AnonCredsModuleConfig` a direct dependency of `AnonCredsRsHolderService` and `IndySdkHolderService`. For instance, in the unit tests it might be needed to add them to the context in order to properly call their methods.

From this perspective, I think it would be cleaner if the auto link secret creation feature would be part of `anoncreds-rs` and/or `indy-sdk` packages. Of course this might result in a different behaviour if choosing one or another AnonCreds implementation, but the other approach will not prevent implementations from ignoring `AnonCredsModuleConfig.autoCreateLinkSecret` neither. 

What do you guys think about this?

TimoGlastra:
Discussed on AFJ WG call. Mostly needed to get this handled by default. Not that important which specific approach we take

genaris:
Superseeded by #1521

PR #1496, opened 2023-07-05
feat: Create Connection record when receive DidCommV2

Artemkaaas:
* Added optional `autoCreateConnectionOnPing` field into Agent config and implemented corresponding logic - indicating whether we should create connection state records when receiving a trust ping message targeted to the DID in the wallet, but with a non-existing pairwise record 
* Updated MessageSender: Unified `sendDidCommV1message` and `sendDidCommV2message` functions into single method `sendMessage`.


packages/core/src/modules/connections/protocols/trust-ping/v2/handlers/TrustPingMessageHandler.ts
```diff
@@ -16,11 +18,27 @@ export class TrustPingMessageHandler implements MessageHandler {
   }
 
   public async handle(messageContext: MessageHandlerInboundMessage<TrustPingMessageHandler>) {
+    let connection: ConnectionRecord | undefined = messageContext.connection
```
Artemkaaas, line 21: If we want to create a connection on receiving any type of DidComm V2 message we can move this logic to `MessageReceiver processUnpackedMessage / findConnection` method. 
Making a connection on Ping is not a standard but seems to be more natural for me..

packages/core/src/modules/connections/protocols/trust-ping/v2/handlers/TrustPingMessageHandler.ts
```diff
@@ -16,11 +18,27 @@ export class TrustPingMessageHandler implements MessageHandler {
...
+
+    // If we received a thrust ping for nonexisting connection record,
+    // create a connection record when the corresponding option is set in the config
+    const recipient = messageContext.message.firstRecipient
```
TimoGlastra, line 25: Are we always the first recipient? Should we update the messageContext to include which recipent did we are? I think this could lead to issues...

If we don't want to add this complexity now, I think we should throw an error if there's more than one recipient, so it couldn't lead to issues

packages/core/src/types.ts
```diff
@@ -50,6 +50,7 @@ export interface InitConfig {
   useDidSovPrefixWhereAllowed?: boolean
   connectionImageUrl?: string
   autoUpdateStorageOnStartup?: boolean
+  autoCreateConnectionOnPing?: boolean
```
TimoGlastra, line 53: I think we should add this to the connections module config

packages/core/src/agent/MessageSender.ts
```diff
@@ -345,6 +337,9 @@ export class MessageSender {
         recipientKeys: queueService.recipientKeys,
         routingKeys: queueService.routingKeys,
         senderKey: firstOurAuthenticationKey,
+        recipientDidDoc: outboundMessageContext.recipientDidDoc,
+        senderDidDoc: outboundMessageContext.senderDidDoc,
+        service: queueService,
```
TimoGlastra, line 342: If we pass the service, there's no need to pass the routingKeys and recipientKeys.

Also we now pass the recipientDidDoc and the senderDidDoc. I think we should try to prevent allowing a ton of conflicting options, as it could lead to unexpected behaviour (what If I provide diferent `recipientKeys` and the `service.recipientKeys`

TimoGlastra (changes requested):
Thanks for this PR @Artemkaaas. Sorry for not getting to it sooner (I was away for a while).

packages/core/src/modules/connections/protocols/trust-ping/v2/handlers/TrustPingMessageHandler.ts
genaris: I agree on creating the connection upon the reception of the first message addressed to the DID. Especially if we want to have the concept of 'Connection' when connecting to agents not based on AFJ, as they might not send any trust ping at all.

packages/core/src/types.ts
genaris: In case we don't use trust ping to create connection, this will apply as well? I guess so, but should have a different naming.

genaris (commented):
Very nice. Really looking forward to start playing with this!

packages/core/src/modules/connections/protocols/trust-ping/v2/handlers/TrustPingMessageHandler.ts
Artemkaaas: Moved

packages/core/src/modules/connections/protocols/trust-ping/v2/handlers/TrustPingMessageHandler.ts
```diff
@@ -16,11 +18,27 @@ export class TrustPingMessageHandler implements MessageHandler {
...
+    if (!messageContext.connection && messageContext.agentContext.config.autoCreateConnectionOnPing && recipient) {
+      connection = await this.connectionService.createConnection(messageContext.agentContext, {
+        protocol: HandshakeProtocol.V2DidExchange,
```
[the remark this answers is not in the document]
Artemkaaas: Added `HandshakeProtocol.None` option

packages/core/src/modules/connections/protocols/trust-ping/v2/handlers/TrustPingMessageHandler.ts
Artemkaaas: Added check that there is no more than one recipient for now.

Artemkaaas:
@TimoGlastra  @genaris  I worked on the comments. Please review the new version.

packages/core/src/modules/connections/models/HandshakeProtocol.ts
```diff
@@ -1,5 +1,5 @@
 export enum HandshakeProtocol {
   Connections = 'https://didcomm.org/connections/1.0',
   DidExchange = 'https://didcomm.org/didexchange/1.0',
-  V2DidExchange = 'https://didcomm.org/didexchange/2.0',
+  None = '', // Used as an explicit marker for DidCommV2 connections which do not require handshake
```
TimoGlastra, line 4:
```suggestion
  None = 'None', // Used as an explicit marker for DidCommV2 connections which do not require handshake
```

packages/core/src/modules/connections/protocols/trust-ping/v2/handlers/V2TrustPingMessageHandler.ts
```diff
@@ -3,24 +3,30 @@ import type { ConnectionService } from '../../../../services/ConnectionService'
...
   public constructor(trustPingService: V2TrustPingService, connectionService: ConnectionService) {
     this.v2TrustPingService = trustPingService
     this.connectionService = connectionService
   }
 
-  public async handle(messageContext: MessageHandlerInboundMessage<TrustPingMessageHandler>) {
-    const message = await this.v2TrustPingService.processPing(messageContext)
+  public async handle(messageContext: MessageHandlerInboundMessage<V2TrustPingMessageHandler>) {
+    const { connection, message } = messageContext
+    if (!connection) {
```
TimoGlastra, line 21: We can use messageContext.assertReadyConnection for this

packages/core/src/modules/oob/OutOfBandApi.ts
```diff
@@ -40,13 +43,20 @@ import { HandshakeReuseHandler } from './protocols/v1/handlers'
 import { HandshakeReuseAcceptedHandler } from './protocols/v1/handlers/HandshakeReuseAcceptedHandler'
 import { OutOfBandInvitation } from './protocols/v1/messages'
 import { V2OutOfBandService } from './protocols/v2/V2OutOfBandService'
-import { OutOfBandInvitation as V2OutOfBandInvitation } from './protocols/v2/messages'
+import { V2OutOfBandInvitation as V2OutOfBandInvitation } from './protocols/v2/messages'
```
TimoGlastra, line 46:
```suggestion
import { V2OutOfBandInvitation } from './protocols/v2/messages'
```

packages/core/src/modules/oob/OutOfBandApi.ts
```diff
@@ -40,13 +43,20 @@ import { HandshakeReuseHandler } from './protocols/v1/handlers'
...
 import { OutOfBandRecord } from './repository/OutOfBandRecord'
 
 const didCommProfiles = ['didcomm/aip1', 'didcomm/aip2;env=rfc19']
 
-export interface CreateOutOfBandInvitationConfig {
-  version?: 'v1' | 'v2'
+export enum OutOfBandVersion {
+  V11 = '1.1',
+  V20 = '2.0',
```
TimoGlastra, line 53:
```suggestion
  V1 = 'v1',
  V2 = 'v2',
```


Can we use the same syntax for this as the issue credential protocol? Not sure if it uses V2 or v2, but would like to keep this consistent across the framework

packages/core/src/modules/oob/OutOfBandApi.ts
```diff
@@ -61,6 +71,15 @@ export interface CreateOutOfBandInvitationConfig {
   appendedAttachments?: V1Attachment[]
 }
 
+export interface CreateV20OutOfBandInvitationConfig {
```
TimoGlastra, line 74: Can we create a base type that both extend from?

packages/core/src/modules/oob/protocols/v2/V2OutOfBandService.ts
```diff
@@ -52,28 +53,32 @@ export class V2OutOfBandService {
...
+  ): Promise<V2OutOfBandInvitation> {
     const didResult = await this.createDid(agentContext, {
       routing: { endpoint: agentContext.config.endpoints[0], mediator: undefined },
     })
-    const invitation = new OutOfBandInvitation({
+    const invitation = new V2OutOfBandInvitation({
       from: didResult.didState.did,
       body: {
-        goalCode: OutOfBandGoalCode.DidExchange,
+        goalCode: V2OutOfBandGoalCode.DidExchange,
```
TimoGlastra, line 66: Should this be hardcoded?

packages/core/src/utils/parseInvitation.ts
```diff
@@ -8,7 +8,7 @@ import { AriesFrameworkError } from '../error'
 import { ConnectionInvitationMessage } from '../modules/connections'
 import { convertToNewInvitation } from '../modules/oob/helpers'
 import { OutOfBandInvitation } from '../modules/oob/protocols/v1/messages'
-import { OutOfBandInvitation as V2OutOfBandInvitation } from '../modules/oob/protocols/v2/messages'
+import { V2OutOfBandInvitation as V2OutOfBandInvitation } from '../modules/oob/protocols/v2/messages'
```
TimoGlastra, line 11:
```suggestion
import { V2OutOfBandInvitation } from '../modules/oob/protocols/v2/messages'
```

TimoGlastra (approved):

PR #1500, opened 2023-07-06
build: Update Github workflows

ryjones:
Create cleanup-cache.yml

Switch build to Yarn

Update continuous-deployment.yml

- Bump actions/checkout
- Switch to `--immutable` with yarn (this is the new command)
- Fix deprecated set-output command

Add heap memory management

Update continuous-integration.yml

- Use a consistent runner
- Fix set-output issues
- Switch to setup-node

Update lint-pr.yml

Update repolinter.yml

Update cleanup-cache.yml

Update codeql-analysis.yml

Create dependabot.yml

Update dependabot.yml

Update .dockerignore

Update Dockerfile

Remove accidental --immutable

build: test large builders

Add setup functions to CodeQL

Update codeql-analysis.yml

Delete local setup-node

Update continuous-integration.yml

Change to larger runners


ryjones:
@TimoGlastra please see this as an attempt to fix the merge conflicts in #1428

.github/workflows/codeql-analysis.yml
```diff
@@ -18,18 +25,36 @@ jobs:
...
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
+        with:
+          # Fetch history for CodeQL analysis
+          fetch-depth: 0
+
+      # setup dependencies
+      - name: Setup Libssl
```
genaris, line 39: This action has been deleted. I think it was not needed to run it here anyway.

Dockerfile
```diff
@@ -1,34 +1,41 @@
-FROM ubuntu:20.04 as base
+## Stage 1: Build indy-sdk and postgres plugin
 
-ENV DEBIAN_FRONTEND noninteractive
+FROM ubuntu:22.04 as base
```
genaris, line 3: I'm in favour of using the newer ubuntu version. However, due to indy-sdk we need to rely on the downgraded libssl 1.1 packages.  Do you guys think this could represent an issue?

genaris (commented):
Thanks @ryjones ! Could you fix the prettier errors?

.github/workflows/codeql-analysis.yml
ryjones: ok deleted

ryjones:
> Thanks @ryjones ! Could you fix the prettier errors?

done - @hyperledger/aries-framework-javascript-committers thoughts?

Dockerfile
TimoGlastra: I think it is fine for now. We should probably update this Dockerfile to use Askar instead of indy-sdk by default

TimoGlastra (commented):
Thanks @ryjones, this is great :)

TimoGlastra (approved):

PR #1508, opened 2023-07-12
fix(samples): mediator wallet and http transport

genaris:
Adds `IndySdkModule` to `Agent` modules in order to allow it to have a default Wallet and Storage implementation (it could be Askar once we solve the performance issue for node <18 and/or drop support for it).

In `HttpInboundTransport`, It also moves content-type validation to the specific endpoint rather than the whole HTTP server. This is needed to allow using an invitation endpoint.

### DIFF
--- a/packages/node/src/transport/HttpInboundTransport.ts
+++ b/packages/node/src/transport/HttpInboundTransport.ts
@@ -24,18 +24,6 @@ export class HttpInboundTransport implements InboundTransport {
     this.app = app ?? express()
     this.path = path ?? '/'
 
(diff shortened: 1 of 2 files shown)


TimoGlastra:
Should we:
- make the wrappers only support node 18 (they basically already do)
- update afj to only support 18 (and maybe 20) in 0.5
- add the patched ref napi to all the wrappers by default

Would solve quite some performance and setup issues.

Node 16 will be EOL in september, we can be ahead of EOL by a few months for once as we have a good reason

TimoGlastra (approved):

PR #1511, opened 2023-07-21
feat: oob without handhsake improvements and routing

### LINKED ISSUE #1250: Connectionless exchange is broken using new oob invitations
When doing a connectionless exchange using OOB invitations, the `~service` decorator is not added to the message after creating, meaning  the `assertConnectionOrServiceDecorator` will fail. We should extend this method to also support out of band exchanges.

Or, even better. Just like we did with legacy connection invitations, a  legacy connectionless invitation should create and OOB record under the hood, and the `assertConnectionOrServiceDecorator` should be updated to `assertConnectionOrOobExchange` (or something like that). This way we don't have to handle all different oob cases. 

It could probably also simplify the connectionless code as it's quite messy now.

```
"error": {
    "name": "AriesFrameworkError",
    "message": "Previously sent message ~service recipientKeys does not include current received message recipient key",
    "stack": "AriesFrameworkError: Previously sent message ~service recipientKeys does not include current received message recipient key\n    at ConnectionService.assertConnectionOrServiceDecorator (/Users/timo/Developer/Playground/mock-credential/node_modules/@aries-framework/core/src/modules/connections/services/ConnectionService.ts:484:10)\n    at V1CredentialProtocol.processRequest (/Users/timo/Developer/Playground/mock-credential/node_modules/@aries-framework/core/src/modules/credentials/protocol/v1/V1CredentialProtocol.ts:769:23)\n    at async V1RequestCredentialHandler.handle (/Users/timo/Developer/Playground/mock-credential/node_modules/@aries-framework/core/src/modules/credentials/protocol/v1/handlers/V1RequestCredentialHandler.ts:18:30)\n    at async Dispatcher.dispatch (/Users/timo/Developer/Playground/mock-credential/node_modules/@aries-framework/core/src/agent/Dispatcher.ts:55:25)\n    at async MessageReceiver.receiveEncryptedMessage (/Users/timo/Developer/Playground/mock-credential/node_modules/@aries-framework/core/src/agent/MessageReceiver.ts:158:5)\n    at async MessageReceiver.receiveMessage (/Users/timo/Developer/Playground/mock-credential/node_modules/@aries-framework/core/src/agent/MessageReceiver.ts:86:9)\n    at async /Users/timo/Developer/Playground/mock-credential/node_modules/@aries-framework/node/src/transport/HttpInboundTransport.ts:46:9"
  },
  "senderKey": "z6Mkndc59jX1UefxogphqFzDLPxUNcSXrJeEcL4YGAZJ9sED",
  "recipientKey": "z6MktrqiMeBmAoXTA2RDBuV8KikpBr71cMSR9PZDskdrAFqa"
}
```

### LINKED ISSUE #1376: Add ability to pass `routing` prop to `agent.oob.receiveInvitation()`
When an agent accepts a proof request, the `ProofsApi` automatically sets up mediation for the presentation message it sends back to the verifier (requester), due to this code in the Proofs API: https://github.com/hyperledger/aries-framework-javascript/blob/main/packages/core/src/modules/proofs/ProofsApi.ts. It doesn't call getRouting with `useDefaultMediator: false`. This means it will fail if no internet is present, and it will also fail with the return routing response.

When receiving an OOB invitation, we want to be able to pass an optional `routing` prop along with the invitation. This routing prop is only used when creating a connection. If we can somehow store this routing, we could use it instead of creating new routing in the ProofsApi `acceptRequest` method.

This issue is related to #1250.

TimoGlastra:
Several improvements to the OOB and connection-less exchange flows.

Fixes #1376
Fixes #1250

Mainly it introduces the following changes:
- Integrate the legacy connection-less exchange with out of band, just like we do with the legacy connection invitations. This means that an out of band record will now be created when you call `createLegacyConnectionlessInvitation`. In addition, you can now also pass a legacy connectionless invitation url to `agent.oob.recieveInvitationFromUrl`. This will transform the legacy connectionless invitation to an out of band invitation. The method to receive the message using `agent.receiveMessage` works to not introduce breaking changes, but I've marked that method as deprecated and we can remove it in 0.5.0. This leaves multiple flows now that work, and we probably want to do some cleanup of the connectionless code once we remove the old flow in 0.5.0 (not using the oob api).
- The receiveInvitation already supported a `routing` prop, but this was not used for connection-less exchanges. This means you could not control which routing config would be used for a connection-less exchange you received an invitation for. Now the routing is used when you respond to a connectionless out of band invitation.
- Several bugs in the oob flow when using connectionless (this was broken).
- Added a `getOutboundMessageContext` method (again ;), we just removed it lol). But this takes into account everything that is needed for responding in connection-full and connection-less exchanges and removes a lot of the duplicated complex logic for handling connection-less. Once we remove the old flow (not using oob api, but still can use legacy connection-less), we could even fully abstract connection-less vs connection, and thus any protocol that uses the `DidCommMessageRepository` could support connection-less out of the box.

Our main use case for these changes is DIDComm over Bluetooth, where we use a mediator for normal HTTP exchagnes, but if we respond to an invitation over BLE, we want to use custom routing and explicitly set `useDefaultMediator` to false, so we can directly communicate over bluetooth instead of through a mediator.


packages/core/src/agent/getOutboundMessageContext.ts
```diff
@@ -0,0 +1,247 @@
...
+  {
+    message,
+    connectionRecord,
+    associatedRecord,
+    lastReceivedMessage,
+    lastSentMessage,
+  }: {
+    connectionRecord?: ConnectionRecord
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    associatedRecord?: BaseRecord<any, any, any>
```
blu3beri, line 38: Could we have these `any`s in the `BaseRecord` as defaults for the generic? Usage would be a bit cleaner and we only have to disable `any` at that point.

blu3beri: We could also take a generic for `getOutboundMessageContext` as when we call this function we always know the associatedRecord type, right?

packages/core/src/agent/getOutboundMessageContext.ts
```diff
@@ -0,0 +1,247 @@
...
+  // Connectionless
+  return getConnectionlessOutboundMessageContext(agentContext, {
+    message,
+    associatedRecord,
+    lastReceivedMessage,
+    lastSentMessage,
+  })
+}
+
+export async function getConnectionlessOutboundMessageContext(
```
blu3beri, line 78: Not for reusability but more for readability, it would be good to extract a lot of logic from this into separate functions. I am trying to understand the whole flow and got confused at certain parts because I did not know in which if statement / level I was reading. We could also switch to `tabWidth: 8` :).

packages/core/src/agent/getOutboundMessageContext.ts
```diff
@@ -0,0 +1,247 @@
...
+    if (!recipientService) {
+      throw new AriesFrameworkError(
+        `Could not find a service to send the message to. Please make sure the connection has a service or provide a service to send the message to.`
+      )
+    }
+
+    // We have created the oob record with a message, that message should be provided here as well
+    if (!lastSentMessage) {
+      throw new AriesFrameworkError('Must have lastSentMessage when out of band record has role Sender')
+    }
```
blu3beri, line 128: Sections like this would be quite nice to have as a separate function even if we would only use it once.

packages/core/src/agent/getOutboundMessageContext.ts
```diff
@@ -0,0 +1,247 @@
...
+      )
+    }
+
+    if (!recipientService) {
+      agentContext.config.logger.error(
+        `No out of band record associated and missing recipient service for connection-less exchange for message ${message.id}.`
+      )
+      throw new AriesFrameworkError(
+        `No out of band record associated and missing recipient service for connection-less exchange for message ${message.id}.`
+      )
```
blu3beri, line 162: Should the logger.error get an additional option to also throw when logged? I think I have seem this pattern before and it seems a bit redundant.

packages/core/src/agent/getOutboundMessageContext.ts
```diff
@@ -0,0 +1,247 @@
...
+    role: DidCommMessageRole.Sender,
+    associatedRecordId: associatedRecord?.id,
+  })
+
+  return new OutboundMessageContext(message, {
+    agentContext: agentContext,
+    associatedRecord,
+    serviceParams: {
+      service: recipientService,
+      senderKey: ourService.recipientKeys[0],
```
blu3beri, line 243: Is the sender key the first `recipientKey`?

packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -472,43 +471,71 @@ export class ConnectionService {
...
-        // If we have previously sent a message, it is not allowed to receive an OOB/unpacked message
-        if (!recipientKey) {
-          throw new AriesFrameworkError(
-            'Cannot verify service without recipientKey on incoming message (received unpacked message)'
-          )
-        }
+      // set theirService to the value of lastReceivedMessage.service
+      let theirService =
+        messageContext.message?.service?.resolvedDidCommService ??
+        previousReceivedMessage?.service?.resolvedDidCommService
```
blu3beri, line 477: nit: Before it was referred to `lastReceivedMessage` might be good to name them consistently

packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -472,43 +471,71 @@ export class ConnectionService {
...
+      }
 
-        // Check if the inbound message sender key is present
-        // in the recipientKeys of previously received message ~service decorator
-        if (!previousReceivedMessage.service || !previousReceivedMessage.service.recipientKeys.includes(senderKey)) {
-          throw new AriesFrameworkError(
-            'Previously received message ~service recipientKeys does not include current received message sender key'
-          )
+      // ourService can be null when we receive an oob invitation or legacy connectionless message and process the message.
+      // In this case lastSentMessage and lastReceivedMessage MUST be null, because there shouldn't be any previous exchange
```
blu3beri, line 510: nit: same here the comment mismatches the var name. `last` vs `previous`.

packages/core/src/modules/oob/domain/OutOfBandDidCommService.ts
```diff
@@ -35,6 +36,24 @@ export class OutOfBandDidCommService extends DidDocumentService {
...
+  }
+
+  public static fromResolvedDidCommService(service: ResolvedDidCommService) {
+    return new OutOfBandDidCommService({
+      id: service.id,
+      recipientKeys: service.recipientKeys.map((key) => new DidKey(key).did),
+      routingKeys: service.routingKeys.map((key) => new DidKey(key).did),
+      serviceEndpoint: service.serviceEndpoint,
+    })
+  }
```
blu3beri, line 56: Don't think it is incorrect, but it seems a bit "misleading" that the recipientKeys in the first object references keys and the second one dids. Would it make sense to rename something? If not then we shouldn't change it to make sense, but it felt a bit strange to me.

packages/core/src/utils/parseInvitation.ts
```diff
@@ -102,9 +105,33 @@ export const parseInvitationShortUrl = async (
...
     const outOfBandInvitation = OutOfBandInvitation.fromUrl(invitationUrl)
     return outOfBandInvitation
-  } else if (parsedUrl['c_i'] || parsedUrl['d_m']) {
+  } else if (parsedUrl['c_i']) {
     const invitation = ConnectionInvitationMessage.fromUrl(invitationUrl)
     return convertToNewInvitation(invitation)
+  }
+  // Legacy connectionless invitation
+  else if (parsedUrl['d_m']) {
+    const messageJson = JsonEncoder.fromBase64(parsedUrl['d_m'] as string)
```
blu3beri, line 114: It uses base64url right? Maybe it does it properly under the hood, with `Buffer`, but it would be good to explicitly call it here.

packages/core/src/utils/parseInvitation.ts
```diff
@@ -102,9 +105,33 @@ export const parseInvitationShortUrl = async (
...
+    const agentMessage = JsonTransformer.fromJSON(messageJson, AgentMessage)
+
+    // ~service is required for legacy connectionless invitations
+    if (!agentMessage.service) {
+      throw new AriesFrameworkError('Invalid legacy connectionless invitation url. Missing ~service decorator.')
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
blu3beri, line 122: Maybe add a small comment that this is done to strip the `~service` from the object.

packages/core/src/utils/parseInvitation.ts
```diff
@@ -102,9 +105,33 @@ export const parseInvitationShortUrl = async (
...
+    const { '~service': service, ...messageWithoutService } = messageJson
+
+    // transform into out of band invitation
+    const invitation = new OutOfBandInvitation({
```
blu3beri, line 126: If we can have out of band invitations without a label it should be moved to optional in the 0.5.0 as it would be breaking, right?

Maybe it can also be a param provided by the user as they sometimes might know the label when scanning?

blu3beri (approved):
Woooooo! Awesome work :). I think an additional review would be good here but if no one has the capacity right now to review I think we can merge it when my minor comments are resolved.

packages/core/src/utils/parseInvitation.ts
TimoGlastra: fromBase64 handles both base64 and base64url

packages/core/src/agent/getOutboundMessageContext.ts
TimoGlastra: Yes we always use the first key. Not really a case where we have multiple sender keys

packages/core/src/agent/getOutboundMessageContext.ts
TimoGlastra: Yeah maybe, but then you can't throw different error types. Agree that this is not ideal. 

Something like this?

```ts
agentContext.config.logger.error(`message`, undefined /* extra data to log */, { throwError: true })
```

packages/core/src/agent/getOutboundMessageContext.ts
TimoGlastra: having any on the record messes with the types when you extend the `BaseRecord` (so then you always have to provide types when extending). 

We can use the generic, but then you still need to do `BaseRecord<any, any, any>` on the generic

packages/core/src/utils/parseInvitation.ts
TimoGlastra: Out of band invitations do require a label I think. It's just that legacy connectionless invitations do not have one. Do you think we should make it optional for that? I see it as a legacy behaviour where we convert it to the new api (which has different requirements)

packages/core/src/agent/getOutboundMessageContext.ts
blu3beri: Yeah something like that. Maybe something worth considering if we ever rework the logger.

packages/core/src/utils/parseInvitation.ts
TimoGlastra: Hmm so it seems the label is optional. So I think we should make it optinoal in AFJ as well, but that would be a breaking change. I'm going to leave it for now as an empty string if that's okay (or a default value), and before 0.5.0 we can make the property optional.

It would be a really simple change. Opened https://github.com/hyperledger/aries-framework-javascript/issues/1524

PR #1517, opened 2023-07-21
fix(anoncreds): wrong key name for predicates in proof object

2mau:


TimoGlastra (approved):

TimoGlastra:
There are some errors with the type checking

.github/workflows/continuous-deployment.yml
```diff
@@ -19,6 +19,7 @@ jobs:
         with:
           # pulls all commits (needed for lerna to correctly version)
           fetch-depth: 0
+          persist-credentials: false
```
genaris, line 22: I think this change should be reverted from this PR

genaris (changes requested):

PR #1521, opened 2023-07-22
feat(anoncreds): auto create link secret

### LINKED ISSUE #1490: Add link secret to agent initalization

genaris:
As discussed in July 6th WG call, here we add an optional configuration parameter for  anoncreds-rs and indy-sdk modules that makes holder services to auto create a default link secret if no one has been created yet. This check is performed only when it is actually needed (i.e. when creating a credential request) rather than at agent initialization.

It defaults to true, meaning that normally it will not be needed to the consumer to do any specific logic to use AnonCreds as a holder.

Superseeds #1491 and fixes #1490


TimoGlastra (approved):
Great! :)

Issue #1510, opened 2023-07-19
Drop support for node 16


genaris:
Should we:
- make the wrappers only support node 18 (they basically already do)
- update afj to only support 18 (and maybe 20) in 0.5
- add the patched ref napi to all the wrappers by default

Would solve quite some performance and setup issues.

Node 16 will be EOL in september, we can be ahead of EOL by a few months for once as we have a good reason

_Originally posted by @TimoGlastra in https://github.com/hyperledger/aries-framework-javascript/issues/1508#issuecomment-1639811295_

genaris:
I agree that this would help us on simplifying the setup and the issues that users have found. It will be also a good moment to switch the test suites to use askar-based wallets instead of indy-sdk.

Have you tested the patched ref-napi in node 20? We have extensively used it under node 18.x (including last releases) but not the chance to use it under 20 yet.

blu3beri:
I think dropping support for 16 is fine. Users can still override it if we set it with the `engine` field inside the `package.json` but it shows clear intent we do not support 16 officially.

TimoGlastra:
Discussed on AFJ WG. There are some issues in Node 18 when using Windows Subshell for Linux 2, which are not an issue in Node 16.

Will address that before we drop support for Node 16

@dnaicker86 will open an issue to describe issues encountered.

No further concerns for dropping support for Node 16

dnaicker86:
> Discussed on AFJ WG. There are some issues in Node 18 when using Windows Subshell for Linux 2, which are not an issue in Node 16.
> 
> Will address that before we drop support for Node 16
> 
> @dnaicker86 will open an issue to describe issues encountered.
> 
> No further concerns for dropping support for Node 16

Logged issue https://github.com/hyperledger/aries-framework-javascript/issues/1522 and published code for testing install with different versions of node 16 or 18 for windows/linux https://github.com/dnaicker86/aries-js-project-1

genaris:
Just an update: based on what we found in https://github.com/hyperledger/aries-framework-javascript/issues/1522#issuecomment-1646925186, it was not a problem of node version but command line processor. So it seems that everything is fine about dropping node 16.

PR #1538, opened 2023-08-03
feat: support askar profiles for multi-tenancy

TimoGlastra:
Adds support for using askar profiles in multi-tenancy, so you don't have to create a new database for each tenant.

As commented on #1486, I've made some 'hacks' to make it work with the currrent API. I think extracting the wallet management from the wallet implementation can clean some things up, as well as having a more sophisticated wallet provider that is a bit more aware of having possible multilple wallets active at the same time within an agent.

The askar module now has a new `multiWalletDatabaseScheme` with two options: `ProfilePerWallet` or `DatabasePerWallet`. the database per wallet is the default to not make breaking changes, and profiel per wallet will reuse the same store for all wallets.

This is the only thing you have to do to benefit from this new feature. I've used the `initialize` method from the askar module to set everything up and register the askar `Store` from the root wallet to be used by the profile wallets. Not the cleanest, but I'm quite happy that it was at least possible to do this without significant changes.

The `AskarBaseWallet` exposes all wallet functionalitiy (such as signin, packing etc..) that is then extended by the `AskarWallet` (root / main wallet that also handles the connection to the database using a store) and `AskarProfileWallet` (uses an existing store and just creates / opens a session for a specific profile)

There's one issue I've discovered with profiles and sessions that requires a 'hack', maybe it can be fixed in Askar: https://github.com/hyperledger/aries-askar/issues/163

### DIFF
--- a/packages/askar/src/wallet/AskarBaseWallet.ts
+++ b/packages/askar/src/wallet/AskarBaseWallet.ts
@@ -445,9 +445,9 @@ export abstract class AskarBaseWallet implements Wallet {
     try {
       cek = AskarKey.fromSecretBytes({ algorithm: KeyAlgs.Chacha20C20P, secretKey: payloadKey })
       const message = cek.aeadDecrypt({
-        ciphertext: TypedArrayEncoder.fromBase64(messagePackage.ciphertext as any),
-        nonce: TypedArrayEncoder.fromBase64(messagePackage.iv as any),
-        tag: TypedArrayEncoder.fromBase64(messagePackage.tag as any),
+        ciphertext: TypedArrayEncoder.fromBase64(messagePackage.ciphertext),
+        nonce: TypedArrayEncoder.fromBase64(messagePackage.iv),
+        tag: TypedArrayEncoder.fromBase64(messagePackage.tag),
         aad: TypedArrayEncoder.fromString(messagePackage.protected),
       })
       return {
--- a/packages/askar/src/wallet/AskarBaseWallet.ts
+++ b/packages/askar/src/wallet/AskarBaseWallet.ts
@@ -1,0 +1,515 @@
+import type {
+  EncryptedMessage,
+  WalletConfig,
+  WalletCreateKeyOptions,
+  WalletSignOptions,
+  UnpackedMessageContext,
+  WalletVerifyOptions,
+  Wallet,
+  WalletConfigRekey,
+  KeyPair,
+  WalletExportImportConfig,
+  Logger,
+  SigningProviderRegistry,
+} from '@aries-framework/core'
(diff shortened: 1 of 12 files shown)


packages/askar/src/wallet/AskarBaseWallet.ts
```diff
@@ -0,0 +1,515 @@
...
+
+    if (!senderKey && alg === 'Authcrypt') {
+      throw new WalletError('Sender public key not provided for Authcrypt')
+    }
+
+    let cek: AskarKey | undefined
+    try {
+      cek = AskarKey.fromSecretBytes({ algorithm: KeyAlgs.Chacha20C20P, secretKey: payloadKey })
+      const message = cek.aeadDecrypt({
+        ciphertext: TypedArrayEncoder.fromBase64(messagePackage.ciphertext as any),
```
berendsliedrecht, line 448: Do we not know the types of these? I assume the ciphertext is a base64(url) encoded str

packages/tenants/tests/tenants-askar-profiles.e2e.test.ts
```diff
@@ -0,0 +1,128 @@
...
+import { agentDependencies } from '@aries-framework/node'
+
+import { describeRunInNodeVersion } from '../../../tests/runInVersion'
+import { AskarModule, AskarMultiWalletDatabaseScheme, AskarProfileWallet, AskarWallet } from '../../askar/src'
+import { askarModuleConfig } from '../../askar/tests/helpers'
+import { testLogger } from '../../core/tests'
+
+import { TenantsModule } from '@aries-framework/tenants'
+
+// FIXME: Re-include in tests when Askar NodeJS wrapper performance is improved
```
berendsliedrecht, line 13: This can be removed, right?

berendsliedrecht (approved):

packages/tenants/tests/tenants-askar-profiles.e2e.test.ts
TimoGlastra: Not really. The tests still run only in Node 18. It can be removed when #1510 is resolved

packages/askar/src/wallet/AskarBaseWallet.ts
TimoGlastra: Yes. I improved the typing for `EncryptedMessage`

packages/tenants/tests/tenants-askar-profiles.e2e.test.ts
genaris: Maybe the comment is a bit outdated because it was done when tests were not running in CI at all. Some time after that, the "describeRunIf" trick has been added in order to run them only on node 18.

But we can live with these comments for now... #1510 will be resolved very soon! 😄

PR #1540, opened 2023-08-04
fix: DidCommV2 - avoid breaking changes 

Artemkaaas:
This PR aims to avoid breaking changes compared to the main branches.

- Preserved export of `AgentMessage` class which actually is `DidCommV1Message`. 
  - `AgentBaseMessage` is a common interface for V1/V2 - used internally
- Preserved name `Attachment` for DidCommV1 message attachments       
- Preserved name `SigningProvider`

Differences with the `main`
- `outOfBandInvitation` field of the OutOfBand record is optional
  - Reason: for DidCommV2 added separate field `v2OutOfBandInvitation`   
- `TransportSession` - `keys` and `inboundMessage`refers to different types but it should not break implementations.

TransportSession


txn.txt
genaris, line 1: I'm glad this one disappeared 😄 

What about `.npmrc`? It has a registry entry for sicpa that I think it's not needed anymore, right?

genaris (approved):

PR #1542, opened 2023-08-08
fix: listen to incoming messages on agent initialize not constructor

niall-shaw:


genaris (approved):
Thanks @niall-shaw ! I don't think it was done on purpose. `Agent` objects are usually initialized only once during their lifecycle. With this fix were you able to make several `shutdown` / `initialize` sequences?

My only concern to merge it right now is about the fact that `messageSubscription` is for some reason a public field, so making it optional could be considered a breaking change. However, I don't know in which use-cases this field is actually accessed from outside. What do you think @TimoGlastra (Git [blames](https://github.com/genaris/aries-framework-javascript/commit/93893b7ab6afd1b4d4f3be4c6b807bab970dd63a#diff-73153d22d2f93cb75bd8f86da06162bd1a9ff870a1990ada37bab0c98dbb3260) you 😛)?

niall-shaw:
@genaris 
> With this fix were you able to make several shutdown / initialize sequences?

Yes I was :)

niall-shaw:
@TimoGlastra @genaris - any updates?

packages/core/src/agent/Agent.ts
```diff
@@ -41,7 +41,7 @@ interface AgentOptions<AgentModules extends AgentModulesInput> {
 // Any makes sure you can use Agent as a type without always needing to specify the exact generics for the agent
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export class Agent<AgentModules extends AgentModulesInput = any> extends BaseAgent<AgentModules> {
-  public messageSubscription: Subscription
+  public messageSubscription?: Subscription
```
TimoGlastra, line 44: let's make it private

```suggestion
  private messageSubscription?: Subscription
```

TimoGlastra (approved):

niall-shaw:
FYI @TimoGlastra - on AFJ WG call, it was mentioned by both @genaris and @berendsliedrecht that this may be considered breaking change. I was going to look into a different solution. What do you think?

TimoGlastra:
Yeah it's definitely a breaking change, but I'm not sure why anyone would use this property, and thus I'm okay with merging it as is, as it should never have been exposed.

But other solutions also fine for me

genaris:
> Yeah it's definitely a breaking change, but I'm not sure why anyone would use this property, and thus I'm okay with merging it as is, as it should never have been exposed.
> 
> But other solutions also fine for me

Yeah @niall-shaw I think it'd be better to make this field private and treat it as a fix to not overcomplicate things. Then we merge it for 0.4.1.

PR #1547, opened 2023-08-18
fix(cheqd): make cosmos payer seed optional

TimoGlastra:
Makes the cosmos payer seed optional, as you can also used it as a holder without payer seed.


berendsliedrecht (approved):

TimoGlastra:
@berendsliedrecht can you re-review. I also added an overwrite of the `Buffer` global. Not sure though if we should do this?

berendsliedrecht (approved):

berendsliedrecht:
> @berendsliedrecht can you re-review. I also added an overwrite of the `Buffer` global. Not sure though if we should do this?

Ideally only for the cheqd specific dependencies. Since you would do this in your app anyways to make it work, I think it is okay.

PR #1549, opened 2023-08-18
fix: create message subscription first

TimoGlastra:
mediation initialization was happening before the message subscription was registered, thus meaning the received messages to provision the mediator wouldn't be processed. This moves it to the start of the initialize method, before mediation initialization


genaris (approved):
Oops!

You were right about avoiding as much as possible doing things at module initialization.

PR #1550, opened 2023-08-18
feat: did:peer:2 and did:peer:4 support in DID Exchange

genaris:
Add support to `did:peer:2` and `did:peer:4` in DID Exchange protocol, in order to start aligning with [RFC 0793](https://github.com/hyperledger/aries-rfcs/blob/fcb2cafce6abea0995804c32a6da473c8894e470/features/0793-unqualfied-dids-transition/README.md).

This includes the recent changes in signatures for DID Exchange 1.1, as specified in [RFC 0023](https://github.com/hyperledger/aries-rfcs/blob/main/features/0023-did-exchange/README.md#request-message-attributes).

Currently, AFJ uses qualified DIDs for DID Exchange protocol, in `did:peer:1` format. The community is going towards `did:peer:2` and the new `did:peer:4` (more meaningful in DIDComm v2 world). So in this PR we are trying to do the transition from `did:peer:1` to `did:peer:4` in order to be compatible with ACA-Py when it adds support for the reception of qualified DIDs in DID Exchange protocol.

The logic is the following:
- There is a new setting in `ConnectionsModuleConfig` to choose the default DID Peer num algo to use in DID Exchange Requests. If not defined, did:peer:1 will be used (this is current behaviour)
- There is a new optional setting in both OOB Reception and Invitation acceptance methods to override default setting. The use case would be: we know our mediator supports only `did:peer:4` (ACA-Py) so we connect with it using that algo, while still using `did:peer:1` for other agents (AFJ, AFGo)
- In DID Exchange Responses, we use the num algo the requester has used. So if they used `did:peer:2`, we'll respond with a `did:peer:2` as well, regardless of our default setting

In addition to this, we add here a new setting called `ourDid` to specify a did when receiving an invitation. This can be used with peer dids but also public dids (as long as their private keys are held by the agent).

### DIFF
--- a/packages/core/src/modules/connections/ConnectionsApi.ts
+++ b/packages/core/src/modules/connections/ConnectionsApi.ts
@@ -91,9 +91,14 @@ export class ConnectionsApi {
       imageUrl?: string
       protocol: HandshakeProtocol
       routing?: Routing
+      ourDid?: string
     }
   ) {
-    const { protocol, label, alias, imageUrl, autoAcceptConnection } = config
+    const { protocol, label, alias, imageUrl, autoAcceptConnection, ourDid } = config
+
+    if (ourDid && !config.routing) {
+      throw new AriesFrameworkError('If an external did is specified, routing configuration must be defined as well')
+    }
 
     const routing =
       config.routing ||
@@ -106,8 +111,13 @@ export class ConnectionsApi {
         alias,
         routing,
         autoAcceptConnection,
+        ourDid,
       })
     } else if (protocol === HandshakeProtocol.Connections) {
+      if (ourDid) {
+        throw new AriesFrameworkError('Using an externally defined did for connections protocol is unsupported')
+      }
+
       result = await this.connectionService.createRequest(this.agentContext, outOfBandRecord, {
         label,
         alias,
--- a/packages/core/src/modules/connections/ConnectionsModuleConfig.ts
+++ b/packages/core/src/modules/connections/ConnectionsModuleConfig.ts
@@ -1,3 +1,5 @@
+import { PeerDidNumAlgo } from '../dids'
+
 /**
  * ConnectionsModuleConfigOptions defines the interface for the options of the ConnectionsModuleConfig class.
  * This can contain optional parameters that have default values in the config class itself.
@@ -13,15 +15,26 @@ export interface ConnectionsModuleConfigOptions {
    * @default false
    */
   autoAcceptConnections?: boolean
+
+  /**
+   * Peer did num algo to use in requests for DID exchange protocol (RFC 0023). It will be also used by default
+   * in responses in case that the request does not use a peer did.
+   *
+   * @default PeerDidNumAlgo.GenesisDoc
+   */
+  peerNumAlgoForDidExchangeRequests?: PeerDidNumAlgo
 }
 
 export class ConnectionsModuleConfig {
   #autoAcceptConnections?: boolean
+  #peerNumAlgoForDidExchangeRequests?: PeerDidNumAlgo
+
   private options: ConnectionsModuleConfigOptions
 
   public constructor(options?: ConnectionsModuleConfigOptions) {
     this.options = options ?? {}
     this.#autoAcceptConnections = this.options.autoAcceptConnections
+    this.#peerNumAlgoForDidExchangeRequests = this.options.peerNumAlgoForDidExchangeRequests
   }
 
   /** See {@link ConnectionsModuleConfigOptions.autoAcceptConnections} */
@@ -32,5 +45,15 @@ export class ConnectionsModuleConfig {
   /** See {@link ConnectionsModuleConfigOptions.autoAcceptConnections} */
   public set autoAcceptConnections(autoAcceptConnections: boolean) {
     this.#autoAcceptConnections = autoAcceptConnections
+  }
+
+  /** See {@link ConnectionsModuleConfigOptions.peerNumAlgoForDidExchangeRequests} */
+  public get peerNumAlgoForDidExchangeRequests() {
+    return this.#peerNumAlgoForDidExchangeRequests ?? PeerDidNumAlgo.GenesisDoc
+  }
+
+  /** See {@link ConnectionsModuleConfigOptions.peerNumAlgoForDidExchangeRequests} */
+  public set peerNumAlgoForDidExchangeRequests(peerNumAlgoForDidExchangeRequests: PeerDidNumAlgo) {
+    this.#peerNumAlgoForDidExchangeRequests = peerNumAlgoForDidExchangeRequests
   }
 }
--- a/packages/core/src/modules/connections/DidExchangeProtocol.ts
+++ b/packages/core/src/modules/connections/DidExchangeProtocol.ts
@@ -21,18 +21,18 @@ import { JsonTransformer } from '../../utils/JsonTransformer'
 import { base64ToBase64URL } from '../../utils/base64'
 import {
   DidDocument,
-  DidDocumentRole,
   createPeerDidDocumentFromServices,
   DidKey,
   getNumAlgoFromPeerDid,
   PeerDidNumAlgo,
   DidsApi,
   isValidPeerDid,
+  getAlternativeDidsForPeerDid,
 } from '../dids'
 import { getKeyFromVerificationMethod } from '../dids/domain/key-type'
 import { tryParseDid } from '../dids/domain/parse'
 import { didKeyToInstanceOfKey } from '../dids/helpers'
-import { DidRecord, DidRepository } from '../dids/repository'
+import { DidRepository } from '../dids/repository'
 import { OutOfBandRole } from '../oob/domain/OutOfBandRole'
 import { OutOfBandState } from '../oob/domain/OutOfBandState'
 import { MediationRecipientService } from '../routing/services/MediationRecipientService'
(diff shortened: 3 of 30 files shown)


swcurran:
@Jsyro @dbluhm — FYI.

@Patrik-Stas — is Aries VCX moving to DID peer 2 support?

It would be really good to get some AATH tests that enable starting the Aries components to initiate connections with did:peer:2 and ensure interop.  We’ll see about what that would take.

TimoGlastra (commented):
I'm not too big a fan of adding the peer did numalgo as an option everywhere, as it will clash when we add support for public dids in didexchange. 

Do you think we could set a default numalgo, and if you don't want to use the default numalgo, you can also just provide a did to the method. This means it has the same interface whether you're using peer dids or not

TimoGlastra:
Left some thoughts on the signed attachments in did exchange: https://github.com/hyperledger/aries-rfcs/issues/717#issuecomment-1684297399

genaris:
> I'm not too big a fan of adding the peer did numalgo as an option everywhere, as it will clash when we add support for public dids in didexchange.
> 
> Do you think we could set a default numalgo, and if you don't want to use the default numalgo, you can also just provide a did to the method. This means it has the same interface whether you're using peer dids or not

So you mean to add an optional `did` parameter to both `receiveInvitation` and `acceptInvitation` in a way that we'll attempt to resolve it if defined, or use default numAlgo to generate a peer did otherwise? i.e. in `DidExchangeProtocol.createRequest()`:

```ts
const didDocument = did
      ? await didsApi.resolveDidDocument(did)
      : await this.createPeerDidDoc(
          agentContext,
          this.routingToServices(routing),
          config.peerNumAlgoForDidExchangeRequests
        )
```

In the use-case where we want to use `did:peer:2` to connect to our mediator while we use `did:peer:1` by default, we would need to do something like:

```ts
    const result = await agent.dids.create({
      method: 'peer',
      didDocument,
      options: {
        numAlgo: PeerDidNumAlgo.MultipleInceptionKeyWithoutDoc,
      },
    })
    await agent.oob.receiveInvitationFromUrl('invitationUrl', {
       ...
      did: result.didState.did,
    })
```

That seems convenient and it will be definitely a new step towards public DID support in DID Exchange Requests/Responses.

packages/core/src/modules/dids/repository/DidRecord.ts
```diff
@@ -23,6 +23,7 @@ export interface DidRecordProps {
 
 export interface CustomDidTags extends TagsBase {
   recipientKeyFingerprints?: string[]
+  alsoKnownAs?: string[]
```
TimoGlastra, line 26: We can extract this from the did document itself right? Or are they set manually?

packages/core/src/modules/dids/methods/peer/peerDidNumAlgo4.ts
```diff
@@ -0,0 +1,126 @@
...
+  delete didDocumentJson.id
+  delete didDocumentJson.alsoKnownAs
+  deleteControllerIfPresent(didDocumentJson.verificationMethod)
+  deleteControllerIfPresent(didDocumentJson.authentication)
+  deleteControllerIfPresent(didDocumentJson.assertionMethod)
+  deleteControllerIfPresent(didDocumentJson.keyAgreement)
+  deleteControllerIfPresent(didDocumentJson.capabilityDelegation)
+  deleteControllerIfPresent(didDocumentJson.capabilityInvocation)
+
+  const buffer = Buffer.concat([VarintEncoder.encode(0x0200), Buffer.from(JSON.stringify(didDocumentJson))])
```
TimoGlastra, line 91: What does 0x0200 stand for? I think if we can assign it to a variable, it makes the code a bit easier to read

TimoGlastra: Does this rely on a consistent json stringify?

TimoGlastra (commented):
One security concern about alsoKnownAs

packages/core/src/modules/connections/ConnectionsApi.ts
```diff
@@ -91,9 +91,14 @@ export class ConnectionsApi {
...
       protocol: HandshakeProtocol
       routing?: Routing
+      ourDid?: string
     }
   ) {
-    const { protocol, label, alias, imageUrl, autoAcceptConnection } = config
+    const { protocol, label, alias, imageUrl, autoAcceptConnection, ourDid } = config
+
+    if (ourDid && !config.routing) {
+      throw new AriesFrameworkError('If an external did is specified, routing configuration must be defined as well')
```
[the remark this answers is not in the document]
genaris: This was set like that because, besides service information, routing object does also specify `mediatorId`, which is passed to `ConnectionService` when creating a connection. 

But I realized that we can infer the `mediatorId` by extracting recipient keys from the did document and querying mediation records by recipient keys, removing the need of using anything from `routing` parameter.

packages/core/src/modules/connections/ConnectionsModuleConfig.ts
```diff
@@ -13,15 +15,26 @@ export interface ConnectionsModuleConfigOptions {
    * @default false
    */
   autoAcceptConnections?: boolean
+
+  /**
+   * Peer did num algo to use in requests for DID exchange protocol (RFC 0023). It will be also used by default
+   * in responses in case that the request does not use a peer did.
+   *
+   * @default PeerDidNumAlgo.GenesisDoc
```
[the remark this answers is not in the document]
genaris: I think originally the idea for the Step 1 of [RFC 0793](https://github.com/hyperledger/aries-rfcs/tree/main/features/0793-unqualfied-dids-transition) was to simply support receiving requests using the new did:peer:4, and keep behaving as before for requests, so I was thinking on keeping the default to 1 in 0.5 and update to 4 in 0.6.

But I'm fine to defaulting it to 4 right now if you prefer so. It will be a breaking change, but there will be plenty in this release.

packages/core/src/modules/connections/DidExchangeProtocol.ts
```diff
@@ -483,11 +503,114 @@ export class DidExchangeProtocol {
...
+          {
+            problemCode: DidExchangeProblemReportReason.ResponseNotAccepted,
+          }
+        )
+      }
+    }
+
+    // Now resolve the document related to the did (which can be either a public did or an inline did)
+    try {
+      return agentContext.dependencyManager.resolve(DidsApi).resolveDidDocument(message.did)
```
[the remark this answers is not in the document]
genaris: Good catch!

packages/core/src/modules/connections/ConnectionsModuleConfig.ts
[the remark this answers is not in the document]
TimoGlastra: No that makes sense, let's keep it at 1 👍

packages/core/src/modules/connections/DidExchangeProtocol.ts
```diff
@@ -306,39 +321,24 @@ export class DidExchangeProtocol {
...
     const didRecord = new DidRecord({
-      did: message.did,
+      did: didDocument.id,
       role: DidDocumentRole.Received,
       didDocument,
       tags: {
         // We need to save the recipientKeys, so we can find the associated did
         // of a key when we receive a message from another connection.
         recipientKeyFingerprints: didDocument.recipientKeys.map((key) => key.fingerprint),
+        alsoKnownAs: didDocument.alsoKnownAs,
```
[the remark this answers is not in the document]
genaris: I need to think a bit more about that, because it should not be a problem when dealing with peer dids (as it is our agent who will be generating the didDoc based on the spec, and the `alsoKnownAs` is only populated where applicable) but it might be an issue in case we get a DID Exchange Request using a public/custom did.

packages/core/src/modules/dids/repository/DidRecord.ts
genaris: At the moment they are set manually, much in the same way that `recipientKeyFingerprints` (and actually will be used similarly once we support DIDComm V2, where we'll use something like `did.findByDid` instead of `did.findByRecipientKey`).

They can be also extracted from the did document, but in cases like did:peer:2 and did:peer:4 we are not storing the whole DID Document in `didDocument` property.

I'm thinking that maybe this general approach of using `alsoKnownAs` is not a good idea, and that we can call it differently (something like `alternateDids`) to make it clear that it is a particular case of certain dids (currently did:peer:4).

packages/core/src/modules/dids/methods/peer/peerDidNumAlgo4.ts
genaris: 0x0200 is the multicodec prefix for JSON (see the procedure we follow in https://identity.foundation/peer-did-method-spec/#creating-a-did).

Regarding the JSON stringify, it's something I was wondering when I've first read the spec but I don't find a problem as long as the process for a given did is only called once, and then stored for further interactions. Do you foresee issues due to this?

TimoGlastra: Well this was a problem with did:peer:1 as well, and the whole reason json canonicalization for example exists.

If you send over the data already encoded though it's fine (e.g. with a jwt the content is included beteen the first and second dot). I need to look a bit better at did:peer:4 spec

dbluhm: I've been lurking on this PR lol; the data is sent already encoded. The verification of the hash over the doc is performed by taking the hash over the still encoded doc, much like the way jwts are verified.

packages/core/src/modules/connections/DidExchangeProtocol.ts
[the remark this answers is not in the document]
TimoGlastra: Thanks for popping in here @dbluhm! 

So we need to make sure we've verified the relation of alsoKnownAs both ways before we allow it to be queried.

Maybe we can rename it to verifiedAlsoKnownAs or something different like relatedDids you mentioned ariel.

TimoGlastra:
Is this ready to be merged @genaris, or are there some fixes related to alsoKnownAs that need to be addressed first?

genaris:
> Is this ready to be merged @genaris, or are there some fixes related to alsoKnownAs that need to be addressed first?

From my side I think it is fine to be merged. For this PR I took a more conservative approach for the `alternativeDids`so it will only consider peer dids. Later on, if we need to do use `alsoKnownAs` for public DIDs we can do some updates to do the reciprocal relationship check that @dbluhm mentioned.

TimoGlastra (approved):

PR #1555, opened 2023-08-28
fix: priority sorting for didcomm services

### LINKED ISSUE #1553: Agent endpoints and DidComm services priority
Currently, AFJ agent with 2 endpoints specified in config creates a Did doc where second service endpoint has higher priority by default.

The reasons are:
- `DidDoc.didCommServices` sorts result by highest priority value (I think it's a perfectly fine approach)
- Did doc creation is using endpoint index in array as a value for service priority (see [implementation](https://github.com/hyperledger/aries-framework-javascript/blob/main/packages/core/src/modules/connections/services/ConnectionService.ts#L754))

It seems not very obvious and in my case I figured this out when I saw that all messages for agent are sent to WS endpoint instead of HTTP.

I also recall that in older AFJ version we had transport priority implementation that was explicit, but worked in some opposite way.

@TimoGlastra, what do you think? Should this be changed or it makes sense to keep it as it is?

TimoGlastra:
Fixes #1553


berendsliedrecht (approved):

genaris (approved):
This is fine but I'd like to add a note to endpoints in AgentConfig just to clarify the priorities for the setting. Something like: 'Agent public endpoints, sorted by priority (higher priority first)'

PR #1558, opened 2023-08-29
fix(oob): support oob with connection and messages

### LINKED ISSUE #1129: Cannot process RequestCredentialMessage for the OOB attached credential-offer
Test scenario:
1. Faber creates an credential-offer-message and creates an oob with it.
```js
const issueCredentialConfig = {
    label: 'Faber College',
    handshake: true,
}
const { message } = await faber.credentials.createOffer({...})
const { outOfBandInvitation } = await faber.oob.createInvitation({
    ...issueCredentialConfig,
    messages: [message],
})
```
2. Alice receives the invitation.
2.1 Alice makes a connection with Faber
2.2 After the connection, Alice sends request-credential-message to Faber
3. Faber fails to process the request-credential-message because:
    - Faber tries to find credentialRecord with connection id [here](https://github.com/hyperledger/aries-framework-javascript/blob/v0.2.5/packages/core/src/modules/credentials/protocol/v1/V1CredentialService.ts#L691)
    - But, Faber created the credentialRecord without connection id [here](https://github.com/hyperledger/aries-framework-javascript/blob/v0.2.5/packages/core/src/modules/credentials/protocol/v1/V1CredentialService.ts#L403) in Step1. So, Faber cannot find the credentialRecord and throws an error.

Solution:
In `processRequest()`, find the credentialRecord without connection id like this:
```js
    const credentialRecord = await this.getByThreadAndConnectionId(requestMessage.threadId, null)
```

TimoGlastra:
Fixes #1129

We now correctly check whether an oob exchange is associated with an exchange if there's no connectionId, and assign the connection if this is the case. 

I think we now really support al flows and combinations of messages / handhsake / connectionless from the oob protocol.

### DIFF
--- a/packages/anoncreds/src/protocols/credentials/v1/V1CredentialProtocol.ts
+++ b/packages/anoncreds/src/protocols/credentials/v1/V1CredentialProtocol.ts
@@ -327,7 +327,7 @@ export class V1CredentialProtocol
       attachments: credentialRecord.linkedAttachments,
     })
 
-    message.setThread({ threadId: credentialRecord.threadId })
+    message.setThread({ threadId: credentialRecord.threadId, parentThreadId: credentialRecord.parentThreadId })
 
     credentialRecord.credentialAttributes = message.credentialPreview.attributes
     credentialRecord.autoAcceptCredential = autoAcceptCredential ?? credentialRecord.autoAcceptCredential
@@ -384,7 +384,7 @@ export class V1CredentialProtocol
       }),
       attachments: credentialRecord.linkedAttachments,
     })
-    message.setThread({ threadId: credentialRecord.threadId })
+    message.setThread({ threadId: credentialRecord.threadId, parentThreadId: credentialRecord.parentThreadId })
 
     credentialRecord.credentialAttributes = message.credentialPreview.attributes
     credentialRecord.autoAcceptCredential = autoAcceptCredential ?? credentialRecord.autoAcceptCredential
@@ -541,6 +541,7 @@ export class V1CredentialProtocol
       credentialRecord = new CredentialExchangeRecord({
         connectionId: connection?.id,
         threadId: offerMessage.threadId,
+        parentThreadId: offerMessage.thread?.parentThreadId,
         state: CredentialState.OfferReceived,
         protocolVersion: 'v1',
       })
@@ -612,7 +613,7 @@ export class V1CredentialProtocol
       requestAttachments: [attachment],
       attachments: offerMessage.appendedAttachments?.filter((attachment) => isLinkedAttachment(attachment)),
     })
-    requestMessage.setThread({ threadId: credentialRecord.threadId })
+    requestMessage.setThread({ threadId: credentialRecord.threadId, parentThreadId: credentialRecord.parentThreadId })
 
     credentialRecord.credentialAttributes = offerMessage.credentialPreview.attributes
     credentialRecord.autoAcceptCredential = autoAcceptCredential ?? credentialRecord.autoAcceptCredential
@@ -691,7 +692,7 @@ export class V1CredentialProtocol
       comment,
     })
 
-    message.setThread({ threadId: credentialRecord.threadId })
+    message.setThread({ threadId: credentialRecord.threadId, parentThreadId: credentialRecord.parentThreadId })
 
     await didCommMessageRepository.saveOrUpdateAgentMessage(agentContext, {
       agentMessage: message,
@@ -731,11 +732,7 @@ export class V1CredentialProtocol
 
     agentContext.config.logger.debug(`Processing credential request with id ${requestMessage.id}`)
 
-    const credentialRecord = await this.getByThreadAndConnectionId(
-      messageContext.agentContext,
-      requestMessage.threadId,
-      connection?.id
-    )
+    const credentialRecord = await this.getByThreadAndConnectionId(messageContext.agentContext, requestMessage.threadId)
     agentContext.config.logger.trace('Credential record found when processing credential request', credentialRecord)
 
     const proposalMessage = await didCommMessageRepository.findAgentMessage(messageContext.agentContext, {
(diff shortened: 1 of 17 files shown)


packages/anoncreds/src/protocols/credentials/v1/V1CredentialProtocol.ts
```diff
@@ -731,11 +732,7 @@ export class V1CredentialProtocol
 
     agentContext.config.logger.debug(`Processing credential request with id ${requestMessage.id}`)
 
-    const credentialRecord = await this.getByThreadAndConnectionId(
-      messageContext.agentContext,
-      requestMessage.threadId,
-      connection?.id
-    )
+    const credentialRecord = await this.getByThreadAndConnectionId(messageContext.agentContext, requestMessage.threadId)
```
berendsliedrecht, line 735: Any specific reason why the `connection?.id` is removed?

packages/core/src/agent/getOutboundMessageContext.ts
```diff
@@ -297,8 +297,11 @@ async function addExchangeDataToMessage(
     associatedRecord: BaseRecordAny
   }
 ) {
+  const legacyInvitationMetadata = outOfBandRecord?.metadata.get(OutOfBandRecordMetadataKeys.LegacyInvitation)
```
berendsliedrecht, line 300: this file has an unused import from `http` which is a mistake, but not from your PR. Can you remove it? 

GitHub does not allow me to add a comment there...

packages/core/src/agent/getOutboundMessageContext.ts
```diff
@@ -297,8 +297,11 @@ async function addExchangeDataToMessage(
...
+
   // Set the parentThreadId on the message from the oob invitation
-  if (outOfBandRecord) {
+  // If connectionless is used, we should not add the parentThreadId
+  if (outOfBandRecord && legacyInvitationMetadata?.legacyInvitationType !== 'connectionless') {
```
berendsliedrecht, line 304: Would it make more sense to have `legacyInvitationType` be an enum?

packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -538,6 +539,86 @@ export class ConnectionService {
...
+  /**
+   * If knownConnectionId is passed, it will compare the incoming connection id with the knownConnectionId, and skip the other validation.
+   *
+   * If no known connection id is passed, it asserts that the incoming message is in response to an attached request message to an out of band invitation.
+   * If is the case, and the state of the out of band record is still await response, the state will be updated to done
+   *
+   */
+  public async matchIncomingMessageToRequestMessageInOutOfBandExchange(
+    messageContext: InboundMessageContext,
+    { knownConnectionId }: { knownConnectionId?: string }
```
berendsliedrecht, line 551: Maybe it should be 

```suggestion
    { expectedConnectionId }: { expectedConnectionId?: string }
```

packages/core/src/modules/oob/repository/outOfBandRecordMetadataTypes.ts
```diff
@@ -1,5 +1,8 @@
+import type { InvitationType } from '../messages'
+
 export enum OutOfBandRecordMetadataKeys {
   RecipientRouting = '_internal/recipientRouting',
+  LegacyInvitation = '_internal/legacyInvitation',
```
berendsliedrecht, line 5: What is our policy on new internal metadata tags? It can be seen as a breaking change if someone, weirdly, uses the internal tags themselves. I think it is fine to keep as non-breaking, just pointing it out.

berendsliedrecht (approved):
Couple of very minor points, otherwise LGTM.

packages/anoncreds/src/protocols/credentials/v1/V1CredentialProtocol.ts
TimoGlastra: Because if we create an offer (not send) it is not linked to a connection. Then if we use it in an oob message, it can create a connection, before the user send a credential request. Then the incoming message has a connection associated, but the credential record doesn't have.

So we now fetch the credential record without connectionId, and later we check whether the connectionId matches, or if the credential record has no connectionID, and then we need a matching out of band record.

packages/core/src/agent/getOutboundMessageContext.ts
TimoGlastra: Yeah probably :)

packages/core/src/modules/connections/services/ConnectionService.ts
TimoGlastra: Yeah that makes sense. I was thinking if it is undefined it doesn't make sense. But after thining about it again, if it's undefined we don't really expect a specific connection Id

packages/core/src/modules/oob/repository/outOfBandRecordMetadataTypes.ts
TimoGlastra: I think metadata should be seen as there can be a lot of keys. anything that starts with `_internal` is internal and we can change them however we like.

packages/core/src/agent/getOutboundMessageContext.ts
genaris: In `ConnectionRecord` we have the `protocol` field where we define the handshake protocol for the DID Exchange, and in the case of DIDComm V2, where no handshake protocol is involved, we set it to `None`. If this `legacyInvitationType` refers to the message type, would it make sense to call it 'none'. in the case of a connection-less exchange?

packages/core/src/modules/connections/services/ConnectionService.ts
```diff
@@ -538,6 +539,86 @@ export class ConnectionService {
...
   }
 
+  /**
+   * If knownConnectionId is passed, it will compare the incoming connection id with the knownConnectionId, and skip the other validation.
+   *
+   * If no known connection id is passed, it asserts that the incoming message is in response to an attached request message to an out of band invitation.
+   * If is the case, and the state of the out of band record is still await response, the state will be updated to done
+   *
+   */
+  public async matchIncomingMessageToRequestMessageInOutOfBandExchange(
```
genaris, line 549: Wow this is a huge method with a huge name!

genaris (approved):

packages/core/src/agent/getOutboundMessageContext.ts
TimoGlastra: > would it make sense to call it 'none'. in the case of a connection-less exchange?

Well in that case we don't know explicitly if it's a connectionless message, and we have custom behaviour for that as you can see here. I'd still call a message with a `~service` an invitation to do something.

So invitation type here refers less to connections, and more to an invitation to an exchange that is sent not over didcomm. (connectionless, connection v1, oob v1, etc..)

packages/core/src/modules/connections/services/ConnectionService.ts
TimoGlastra: Yeah, i like explicitness over being brief. But suggestions welcome! :)

PR #1559, opened 2023-08-29
fix: do not send package via outdated session

niall-shaw:
We found an edge-case where a websocket transport session would be closed/unavailable on the client agent side, but mediator still thinks it is available (due to client being offline).
When client returns back online and initiates a flow via mediator, the mediator will attempt to send the package to the now outdated websocket, instead of the newest / current session.

This PR fixes the problem, but is there any reason why a single connection may need multiple open WebSocket transport sessions?

### DIFF
--- a/packages/core/src/agent/TransportService.ts
+++ b/packages/core/src/agent/TransportService.ts
@@ -13,8 +13,8 @@ export class TransportService {
   public transportSessionTable: TransportSessionTable = {}
 
   public saveSession(session: TransportSession) {
-    if (session.type === 'WebSocket' && session.connectionId) {
-      const oldSessions = this.getWebSocketSessionsForConnectionId(session.connectionId)
+    if (session.connectionId) {
+      const oldSessions = this.getExistingSessionsForConnectionIdAndType(session.connectionId, session.type)
       oldSessions.forEach((oldSession) => {
         if (oldSession) {
           this.removeSession(oldSession)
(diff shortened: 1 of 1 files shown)


packages/core/src/agent/TransportService.ts
```diff
@@ -40,6 +48,12 @@ export class TransportService {
   public removeSession(session: TransportSession) {
     delete this.transportSessionTable[session.id]
   }
+
+  private getWebSocketSessionsForConnectionId(connectionId: string) {
+    return Object.values(this.transportSessionTable).filter(
+      (session) => session?.connectionId === connectionId && session.type === 'WebSocket'
+    )
+  }
 }
```
TimoGlastra, line 57: Maybe we can limit it to one session per type and connection id? That way we don't have to add web socket specific logic?

niall-shaw:
FYI - this bug was found using AFJ 0.3.3, so perhaps it is not relevant anymore, but I can't see why it wouldn't be.

packages/core/src/agent/TransportService.ts
niall-shaw: Yeah, I'm happy with that also. Bug was only found in WebSocket, that's why I made it this way, but happy to change :)

niall-shaw:
@TimoGlastra - adjusted for any type

genaris (approved):

PR #1560, opened 2023-08-29
feat: credential protocols v3

genaris:
Implementation of [Issue Credential V3](https://github.com/decentralized-identity/waci-didcomm/tree/main/issue_credential) and [Present Proof V3](https://github.com/decentralized-identity/waci-didcomm/blob/main/present_proof/present-proof-v3.md), based on V2 protocols and updating them to use attachments for DIDComm V2.

This is dependent on https://github.com/hyperledger/aries-framework-javascript/pull/1546 (as there are a few general corrections and tweaks for DIDComm V2 there) and there is still some work to do, such as:
- Add tests
- Add OOB support (if needed)
- Verify its correct alignment with WACI-DIDComm definitions in terms of message definitions and flow

At the current status, it allows to run all possible use-cases for AFJ demo (connect, send message, offer credential, request proof) in both DIDComm V1 and DIDComm V2


TimoGlastra:
Nice work @genaris! :)

PR #1573, opened 2023-09-13
fix: improve logging

ericvergnaud:


packages/node/src/transport/WsInboundTransport.ts
```diff
@@ -83,18 +82,26 @@ export class WebSocketTransportSession implements TransportSession {
...
     }
-
-    this.socket.send(JSON.stringify(encryptedMessage))
+    this.socket.send(JSON.stringify(encryptedMessage), (error?) => {
+      if (error != undefined) {
+        this.logger.error('Error sending message: ' + error)
+        throw new AriesFrameworkError(`${this.type} send message failed.`)
+      } else {
+        this.logger.debug(`${this.type} sent message successfully.`)
+      }
```
jakubkoci, line 103: Wouldn't be better to just throw an error here and evaluate it where we call `await session.send(agentContext, encryptedMessage)`?

jakubkoci (commented):
Hi @ericvergnaud . Thanks for those log additions, to me, it's always hard to evaluate right place to log and not overlog. :)

packages/node/src/transport/WsInboundTransport.ts
ericvergnaud: Not sure what 'better' means, personally I prefer having debug logs as close as possible to where things happen, and from the existing logs it was impossible to decide whether a message had been sent successfully or not...

packages/core/src/agent/MessageSender.ts
```diff
@@ -99,11 +99,12 @@ export class MessageSender {
   }
 
   private async sendMessageToSession(agentContext: AgentContext, session: TransportSession, message: AgentMessage) {
-    this.logger.debug(`Existing ${session.type} transport session has been found.`)
+    this.logger.debug(`Existing ${session.type} transport session has been found, sending message.`)
```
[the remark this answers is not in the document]
ericvergnaud: Agreed

packages/node/src/transport/WsInboundTransport.ts
ericvergnaud: Have in mind that this is a debug log, typically activated only when troubleshooting, so you'd want to know where exactly things go wrong

packages/core/src/agent/MessageSender.ts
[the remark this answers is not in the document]
ericvergnaud: Done

ericvergnaud:
Hi,
I see that some tests are failing but I very much doubt that has anything to do with the log messages...
Is there anything I need to do ?

packages/node/src/transport/WsInboundTransport.ts
ericvergnaud: Slightly changed to bubble up the error

ericvergnaud:
@jakubkoci hey, any chance we can finalize this ? Looks like 'main' is evolving rapidly making it more difficult to merge every day ?

packages/node/src/transport/WsInboundTransport.ts
genaris: Do you know if `socket.send()` blocks until the callback is actually called? 

I think it will be nice to return from `session.send()` only if it could actually write successfully to the socket. Not sure how it works, but probably we'll need to wait until the callback is called and check the error code before returning.

packages/core/src/agent/MessageSender.ts
```diff
@@ -99,11 +99,12 @@ export class MessageSender {
   }
 
   private async sendMessageToSession(agentContext: AgentContext, session: TransportSession, message: AgentMessage) {
-    this.logger.debug(`Existing ${session.type} transport session has been found.`)
+    this.logger.debug(`Packing message and sending it via existing session ${session.type}...`)
     if (!session.keys) {
       throw new AriesFrameworkError(`There are no keys for the given ${session.type} transport session.`)
     }
     const encryptedMessage = await this.envelopeService.packMessage(agentContext, message, session.keys)
+    this.logger.debug(`Sending message`)
```
genaris, line 107:
```suggestion
    this.logger.debug('Sending message')
```

packages/core/src/agent/TransportService.ts
genaris, line 1: This change is not related to this PR. I think there was an issue when you merged main into your branch locally.

I usually use the  'Update branch' button in GitHub UI or do a `git fetch upstream` + `git merge upstream/main` (I have to previously set up upstream by executing `git remote add upstream https://github.com/hyperledger/aries-framework-javascript`)

packages/core/src/modules/message-pìckup/MessagePickupApi.ts
```diff
@@ -32,17 +33,20 @@ export class MessagePickupApi<MPPs extends MessagePickupProtocol[] = [V1MessageP
   private messageSender: MessageSender
   private agentContext: AgentContext
   private connectionService: ConnectionService
+  private logger!: Logger
```
genaris, line 36:
```suggestion
  private logger: Logger
```

packages/core/src/modules/message-pìckup/MessagePickupApi.ts
```diff
@@ -61,6 +65,7 @@ export class MessagePickupApi<MPPs extends MessagePickupProtocol[] = [V1MessageP
    * @param options: connectionId associated to the message and the encrypted message itself
    */
   public async queueMessage(options: QueueMessageOptions): Promise<QueueMessageReturnType> {
+    this.logger.debug(`Queuing message...`)
```
genaris, line 68:
```suggestion
    this.logger.debug('Queuing message...')
```

packages/indy-sdk/package.json
genaris, line 1: Another unrelated change

packages/node/src/transport/WsInboundTransport.ts
```diff
@@ -73,7 +72,7 @@ export class WsInboundTransport implements InboundTransport {
       try {
         await messageReceiver.receiveMessage(JSON.parse(event.data), { session })
       } catch (error) {
-        this.logger.error('Error processing message')
+        this.logger.error('Error processing message: ' + error)
```
genaris, line 75:
```suggestion
        this.logger.error(`Error processing message: ${error}`)
```

packages/node/src/transport/WsInboundTransport.ts
```diff
@@ -83,18 +82,26 @@ export class WebSocketTransportSession implements TransportSession {
   public id: string
   public readonly type = 'WebSocket'
   public socket: WebSocket
+  private logger!: Logger
```
genaris, line 85:
```suggestion
  private logger: Logger
```

yarn.lock
genaris, line 1: Unrelated change

genaris (changes requested):
Thanks @ericvergnaud ! I think it is more than just an improvement of logging, as it now handles a possible error when sending a message through a WebSocket. In regards to that I'd like to make sure that the method won't return unless it is sure it could write on the socket. Maybe it is already doing so, but there are also other possibilities (like using `rxjs` magic as we do in other parts of the framework) to wait for the callback execution and return.

ericvergnaud:
@genaris re https://github.com/hyperledger/aries-framework-javascript/pull/1573#discussion_r1328658767 yes the callback is only called after send completes, see https://github.com/websockets/ws/blob/master/doc/ws.md#websocketsenddata-options-callback

genaris:
> @genaris re [#1573 (comment)](https://github.com/hyperledger/aries-framework-javascript/pull/1573#discussion_r1328658767) yes the callback is only called after send completes, see https://github.com/websockets/ws/blob/master/doc/ws.md#websocketsenddata-options-callback

It is clear to me that the callback is called after socket is written, but my question is if this procedure is done synchronously: will the function will block until it gets a successful or failed status on the write operation?

`ws.send()` is calling `socket.write()` under the hood and simply passing the provided callback. In its [documentation](https://nodejs.org/api/net.html#socketwritedata-encoding-callback) it states _The optional callback parameter will be executed when the data is finally written out, which may not be immediately_. Also it says that its (boolean) return value depends on the data flushed to kernel buffer. 

So I'm wondering if `socket.write()` (and therefore `ws.send()`) will return immediately and executing the callback asynchronously. If that's the case, it might happen that the application continues the execution after `send()` and the error is thrown some time later, which maybe is not handled as we would like?

Anyway, apart from this curiosity, I still see the unrelated changes in this PR. You'll need to rebase/merge your branch proerly, otherwise it will not be possible to merge it into main once approved.

ericvergnaud:
I'll try rebasing

ericvergnaud:
Rebasing done

genaris (approved):

genaris:
> In my understanding and experience, `ws.send()`is synchronous, even if it's not immediate, and the callback is executed in the context of the caller, so the exception bubbles up properly.

As long as the error is thrown on the same context of the caller, I'm OK with it.

packages/node/src/transport/WsInboundTransport.ts
jakubkoci: I'm sorry, I should've clarified it in my first comment. The idea is that the `WebSocketSession` is more an entity or value object than a service object. Therefore I would rather avoid passing the logger there. In the same way, we don't pass (hopefully) the logger into records.

genaris: Oh my bad, I should have asked before merging. But don't worry, there will be more changes in these classes shortly, so we can take the opportunity to analyze the flows where session.send is called and call the logger from an upper layer if needed.

PR #1575, opened 2023-09-15
fix: service validation in OOB invitation objects

genaris:
I'm not an expert (neither a fan 😝) of `class-validator` but I recently found an issue when creating an `OutOfBandInvitation` instance whose `service` field contains a plain string. It works and allows to do a `toUrl()` and so on, but when it passes through the MessageValidator it threw an error saying:

```
OutOfBandInvitation: Failed to validate class.
    An instance of OutOfBandInvitation has failed the validation:
     - property services[0] has failed the following constraints: each value in nested property services must be either object or array·
```

I found that services field has both `@IsStringOrInstance` and `@ValidateNested` decorators. The first one seems to be doing the correct check on every element of the array, while the latter (which is the one  throwing the error) seems to me redundant in this case. If I remove it, class validation passes (as expected). 

So here I'm removing `@ValidateNested` decorator. Not sure if it's the best approach, but of course happy to change it if there is a more appropriate way.

### DIFF
--- a/packages/core/src/modules/oob/__tests__/OutOfBandInvitation.test.ts
+++ b/packages/core/src/modules/oob/__tests__/OutOfBandInvitation.test.ts
@@ -1,8 +1,11 @@
 import type { ClassValidationError } from '../../../error/ClassValidationError'
 
 import { Attachment } from '../../../decorators/attachment/Attachment'
+import { MessageValidator } from '../../../utils'
 import { JsonEncoder } from '../../../utils/JsonEncoder'
 import { JsonTransformer } from '../../../utils/JsonTransformer'
+import { HandshakeProtocol } from '../../connections'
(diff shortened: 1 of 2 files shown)


genaris:
Another interesting thing I found related to `class-validator` is that `@IsUrl` decorator (without any parameters) requires the top-level domain (TLD) to consider an URL as valid. So for instance if I set `imageUrl` to `http://localhost/image.jpg` it will fail.

I guess this won't represent any issue in production environments, but could be inconvenient when doing local development or accessing resources in a local area network.

TimoGlastra (approved):

TimoGlastra:
> I guess this won't represent any issue in production environments, but could be inconvenient when doing local development or accessing resources in a local area network.

Hmm yeah that's a bit weird. But not seeing any immediate issues with it.

PR #1576, opened 2023-09-16
feat: Discover Features V2 over DIDComm V2

genaris:
Discover Features 2.0 is one of those protocols that are suitable for both DIDComm V1 and V2. However, this becomes a bit complicated because current AFJ model considers that a given message type is defined for either V1 or V2.

So after an interesting discussion in a previous AFJ WG Call, we found out that we can leave each protocol determine which versions it supports and register both models into the `MessageHandlerRegistry`, which will now allow to filter not only by message type (PIURI) but also by DIDComm version.

In this PR we have:
- Different fixes and adaptations for DIDComm V2 (like attachment transformers, message handling, etc.) found during the work of other PRs (#1546 and #1560). This would make this PR to be the basis of those, which will hopefully only add new protocols and make their review easier
- New handling of received messages in `MessageReceiver` and `Dispatcher` to figure out whether it is a DIDComm V1 or V2 message and find the right class and handler
- Updates in Discover Features module to add DIDComm V2 modeling for Query and Disclosure messages, and register/process both in a single handler and service
- There are some slight API changes, like the return types and events, due to messages can be now either be DIDComm V1 or V2. I think in the next major release (before merging DIDComm V2 to main) we can remove the complete message object from the event, as the relevant fields are sent as separate properties (and in case we want to go low level and get the complete message we can subscribe  to `AgentMessageProcessedEvent`)
- Discover Features V2 E2E test has been updated to work in both DIDComm V1 and V2. This makes it to require Askar. As we'll switch soon our tests to node 18 + shared components I guess there is no problem on that

### DIFF
--- a/packages/core/src/agent/MessageReceiver.ts
+++ b/packages/core/src/agent/MessageReceiver.ts
@@ -153,7 +153,9 @@ export class MessageReceiver {
     const { plaintextMessage, senderKey, recipientKey } = unpackedMessage
 
     this.logger.info(
-      `Received message with type '${plaintextMessage['@type']}', recipient key ${recipientKey?.fingerprint} and sender key ${senderKey?.fingerprint}`,
+      `Received message with type '${plaintextMessage['@type'] ?? plaintextMessage['type']}', recipient key ${
+        recipientKey?.fingerprint
+      } and sender key ${senderKey?.fingerprint}`,
       plaintextMessage
     )
 
--- a/packages/core/src/agent/MessageReceiver.ts
+++ b/packages/core/src/agent/MessageReceiver.ts
@@ -300,7 +300,9 @@ export class MessageReceiver {
       throw new AriesFrameworkError(`No type found in the message: ${message}`)
     }
 
-    const MessageClass = this.messageHandlerRegistry.getMessageClassForMessageType(messageType)
+    const didcommVersion = isPlaintextMessageV1(message) ? DidCommMessageVersion.V1 : DidCommMessageVersion.V2
+
+    const MessageClass = this.messageHandlerRegistry.getMessageClassForMessageType(messageType, didcommVersion)
 
     if (!MessageClass) {
       throw new ProblemReportError(`No message class found for message type "${messageType}"`, {
--- a/packages/core/src/agent/MessageReceiver.ts
+++ b/packages/core/src/agent/MessageReceiver.ts
@@ -7,7 +7,7 @@ import type { ConnectionRecord } from '../modules/connections'
 import type { InboundTransport } from '../transport'
 
 import { InjectionSymbols } from '../constants'
-import { isPlaintextMessageV1, isPlaintextMessageV2 } from '../didcomm'
+import { DidCommMessageVersion, isPlaintextMessageV1, isPlaintextMessageV2 } from '../didcomm'
 import { getPlaintextMessageType, isEncryptedMessage, isPlaintextMessage } from '../didcomm/helpers'
 import { AriesFrameworkError } from '../error'
 import { Logger } from '../logger'
--- a/packages/core/src/agent/getOutboundMessageContext.ts
+++ b/packages/core/src/agent/getOutboundMessageContext.ts
@@ -81,8 +81,8 @@ export async function getOutboundMessageContext(
 
   if (
     !(message instanceof DidCommV1Message) ||
-    !(lastReceivedMessage instanceof DidCommV1Message) ||
-    !(lastSentMessage instanceof DidCommV1Message)
+    (lastReceivedMessage !== undefined && !(lastReceivedMessage instanceof DidCommV1Message)) ||
+    (lastSentMessage !== undefined && !(lastSentMessage instanceof DidCommV1Message))
   ) {
     throw new AriesFrameworkError('No connection record associated with DIDComm V2 messages exchange')
   }
--- a/packages/core/src/agent/getOutboundMessageContext.ts
+++ b/packages/core/src/agent/getOutboundMessageContext.ts
@@ -7,12 +7,14 @@ import type { BaseRecordAny } from '../storage/BaseRecord'
 
 import { Key } from '../crypto'
 import { ServiceDecorator } from '../decorators/service/ServiceDecorator'
-import { DidCommV1Message, DidCommV2Message } from '../didcomm'
+import { DidCommMessageVersion, DidCommV1Message, DidCommV2Message } from '../didcomm'
 import { AriesFrameworkError } from '../error'
-import { OutOfBandService, OutOfBandRole, OutOfBandRepository } from '../modules/oob'
+import { OutOfBandRole } from '../modules/oob/domain'
+import { OutOfBandService } from '../modules/oob/protocols'
+import { OutOfBandRepository } from '../modules/oob/repository'
 import { OutOfBandRecordMetadataKeys } from '../modules/oob/repository/outOfBandRecordMetadataTypes'
-import { RoutingService } from '../modules/routing'
-import { DidCommMessageRepository, DidCommMessageRole } from '../storage'
+import { RoutingService } from '../modules/routing/services'
+import { DidCommMessageRepository, DidCommMessageRole } from '../storage/didcomm'
 import { uuid } from '../utils/uuid'
 
 import { OutboundMessageContext } from './models'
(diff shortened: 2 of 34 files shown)


packages/core/src/agent/MessageReceiver.ts
```diff
@@ -300,7 +302,9 @@ export class MessageReceiver {
       throw new AriesFrameworkError(`No type found in the message: ${message}`)
     }
 
-    const MessageClass = this.messageHandlerRegistry.getMessageClassForMessageType(messageType)
+    const didcommVersion = isPlaintextMessageV1(message) ? DidCommMessageVersion.V1 : DidCommMessageVersion.V2
```
TimoGlastra, line 305: Should we make this future proof and have an if / else if / else throw error?

packages/core/src/agent/getOutboundMessageContext.ts
```diff
@@ -48,6 +51,27 @@ export async function getOutboundMessageContext(
...
+
+    // Check that message DIDComm version matches connection
+    if (
+      (connectionRecord.isDidCommV1Connection && message.didCommVersion !== DidCommMessageVersion.V1) ||
+      (connectionRecord.isDidCommV2Connection && message.didCommVersion !== DidCommMessageVersion.V2)
+    ) {
+      throw new AriesFrameworkError(
+        `Message DIDComm version ${message.didCommVersion} does not match connection ${connectionRecord.id}`
+      )
+    }
```
TimoGlastra, line 63: Interesting. I think we may have to tweak this at some point as the services in a did document dictate which didcomm versions are supported. We only have did:peer connections now, but if we want to make it future proof already, we could also handle this later on, after we've resolved the services for a connection.

Thoughts?

packages/core/src/agent/getOutboundMessageContext.ts
```diff
@@ -48,6 +51,27 @@ export async function getOutboundMessageContext(
...
+
+    // Attach 'from' and 'to' fields according to connection record (unless they are previously defined)
+    if (message instanceof DidCommV2Message) {
+      message.from = message.from ?? connectionRecord.did
+      const recipients = message.to ?? (connectionRecord.theirDid ? [connectionRecord.theirDid] : undefined)
+      if (!recipients) {
+        throw new AriesFrameworkError('Cannot find recipient did for message')
+      }
+      message.to = recipients
+    }
```
TimoGlastra, line 73: Can there be any danger in having different from/to than the connectionRecord? What would be the use case? I think if you want custom from/to you should probably not pass a connection record to the `getOutboundMessageContext` message.

packages/core/src/agent/getOutboundMessageContext.ts
```diff
@@ -67,6 +91,14 @@ export async function getOutboundMessageContext(
     )
   }
 
+  if (
+    !(message instanceof DidCommV1Message) ||
```
TimoGlastra, line 95: I think we can invert all these statements?

```suggestion
    message instanceof DidCommV2Message
```

packages/core/src/didcomm/transformers.ts
```diff
@@ -0,0 +1,39 @@
...
+      base64: data.base64,
+      json: data.json,
+      jws: data.jws,
+      links: data.links,
+      hash: data.sha256,
+    }),
+  })
+}
+
+export function toV1Attachment(v2Attachment: V2Attachment): Attachment {
```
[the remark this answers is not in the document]
TimoGlastra: Oh and finally, do you thin we need a generic didcomm attachment class / interface that we can map to v1 / v2?

packages/core/src/modules/discover-features/protocol/v2/V2DiscoverFeaturesService.ts
```diff
@@ -47,15 +49,17 @@ export class V2DiscoverFeaturesService extends DiscoverFeaturesService {
 
   public async createQuery(
     options: CreateQueryOptions
-  ): Promise<DiscoverFeaturesProtocolMsgReturnType<V2QueriesMessage>> {
-    const queryMessage = new V2QueriesMessage({ queries: options.queries })
+  ): Promise<DiscoverFeaturesProtocolMsgReturnType<V2QueriesMessage | V2QueriesDidCommV2Message>> {
+    const queryMessage = options.connectionRecord?.isDidCommV2Connection
+      ? new V2QueriesDidCommV2Message({ queries: options.queries })
+      : new V2QueriesMessage({ queries: options.queries })
```
TimoGlastra, line 55: 🤯

packages/core/src/modules/discover-features/protocol/v2/V2DiscoverFeaturesService.ts
```diff
@@ -71,10 +75,11 @@ export class V2DiscoverFeaturesService extends DiscoverFeaturesService {
       },
     })
 
-    // Process query and send responde automatically if configured to do so, otherwise
+    // Process query and send respose automatically if configured to do so, otherwise
```
TimoGlastra, line 78:
```suggestion
    // Process query and send response automatically if configured to do so, otherwise
```

packages/core/src/modules/discover-features/protocol/v2/V2DiscoverFeaturesService.ts
```diff
@@ -83,18 +88,25 @@ export class V2DiscoverFeaturesService extends DiscoverFeaturesService {
...
     options: CreateDisclosureOptions
-  ): Promise<DiscoverFeaturesProtocolMsgReturnType<V2DisclosuresMessage>> {
+  ): Promise<DiscoverFeaturesProtocolMsgReturnType<V2DisclosuresMessage | V2DisclosuresDidCommV2Message>> {
     const matches = this.featureRegistry.query(...options.disclosureQueries)
 
-    const discloseMessage = new V2DisclosuresMessage({
-      threadId: options.threadId,
-      features: matches,
-    })
+    const discloseMessage = options.connectionRecord?.isDidCommV2Connection
```
TimoGlastra, line 94: To extend on my proposal to infer which didomm version to use based on the services. Maybe there can be a `toDidCommV1Message` on the v2 base class and a `toDidCommV2Message` on the v1 base class and that this is optionally implemented when you support 2 didcomm versions for hte same protocol version. Then we could automatically transform it into the other type when sending the message if the method is available, and the didcomm version does not match.

TimoGlastra: But it doens't make fully sense, because we do need to know upfront what didcomm version a connection supports. And we should be able to update that in the connection record when we re-fetch a did and resolve their servides (so it's not something static). But that can be done later ofc.

packages/core/src/storage/didcomm/DidCommMessageRecord.ts
```diff
@@ -58,13 +58,13 @@ export class DidCommMessageRecord extends BaseRecord<DefaultDidCommMessageTags>
...
-    const messageId = this.message['@id'] as string
-    const messageType = this.message['@type'] as string
+    const messageId = (this.message['@id'] ?? this.message['id']) as string
+    const messageType = (this.message['@type'] ?? this.message['type']) as string
 
     const { protocolName, protocolMajorVersion, protocolMinorVersion, messageName } = parseMessageType(messageType)
 
     const thread = this.message['~thread']
-    let threadId = messageId
+    let threadId = (this.message['thid'] ?? messageId) as string
```
TimoGlastra, line 67: Should we split this in if (didcommv1) {} else if (didcommv2) {} ?

to not make it so intertwined. As it could lead to some issues if for some magic reason a didcomm v2 message has a `~thread` header 😆

packages/core/tests/helpers.ts
```diff
@@ -217,7 +218,7 @@ const isCredentialStateChangedEvent = (e: BaseEvent): e is CredentialStateChange
 const isConnectionStateChangedEvent = (e: BaseEvent): e is ConnectionStateChangedEvent =>
   e.type === ConnectionEventTypes.ConnectionStateChanged
 const isTrustPingReceivedEvent = (e: BaseEvent): e is TrustPingReceivedEvent =>
-  e.type === TrustPingEventTypes.TrustPingReceivedEvent
+  e.type === TrustPingEventTypes.TrustPingReceivedEvent || e.type === TrustPingEventTypes.V2TrustPingReceivedEvent
```
TimoGlastra, line 221: I think we can unify these again as we're okay with introducing breaking changes?

So just one trust ping received event (as we have with all other multi-verion protocols)

TimoGlastra (approved):

TimoGlastra:
Apologies for the long wait @genaris 😉

genaris:
Superseded by #2704

PR #1578, opened 2023-09-18
fix(core): remove node-fetch dependency

### LINKED ISSUE #1492: Dependency on @types/node-fetch required for core package
Should either add `@types/node-fetch` as dep to core, or we should remove `node-fetch` from core. Ideally the latter

genaris:
Our good friend @dependabot with its #1565 made me remember we had this dependency on core that shouldn't be there. Didn't go further in order to not deal with breaking changes. I think we can move to node-fetch v3 (and see how to deal `AgentDependencies` types) in a next major release.

Fixes #1492

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -30,6 +30,7 @@
     "@stablelib/ed25519": "^1.0.2",
     "@stablelib/random": "^1.0.1",
     "@stablelib/sha256": "^1.0.1",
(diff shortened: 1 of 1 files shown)


TimoGlastra (approved):

TimoGlastra:
We should check if the released alpha from this doesn't give issues. There was some issues with it not being in the dependencies I think

PR #1583, opened 2023-09-22
feat: allow connection invitation encoded in oob url param

### LINKED ISSUE #1581: Incompatibility with Lissi OOB URLs
I've recently trying out [Lissi demo](https://try.lissi.id/demo) to see how far an AFJ-based app can go and I noticed that in their invitation URLs they are using `oob` query parameter and encoding a `connections/1.0/invitation` message into it. AFJ expects it to be an `out-of-band/1.x/invitation` message, so as a result it throws a validation error when scanning the code.

Whilst I think AFJ interpretation is correct, based on the wording from [Aries RFC 0434](https://github.com/hyperledger/aries-rfcs/blob/b3a3942ef052039e73cd23d847f42947f8287da2/features/0434-outofband/README.md#standard-out-of-band-message-encoding) where it says _The oob query parameter is required and is reserved to contain the out-of-band message string_ one could also interpret that 'out-of-band' refers to any DIDComm message that will be send out of band (i.e not strictly from the `out-of-band` protocol). 

So in some way I don't see that it would be _harmful_ to allow processing a former `connections` protocol invitation encoded in `oob` query parameter. What do you think?

genaris:
Closes #1581

### DIFF
--- a/packages/core/src/utils/__tests__/shortenedUrl.test.ts
+++ b/packages/core/src/utils/__tests__/shortenedUrl.test.ts
@@ -3,7 +3,7 @@ import type { Response } from 'node-fetch'
 import { Headers } from 'node-fetch'
 
 import { ConnectionInvitationMessage } from '../../modules/connections'
-import { OutOfBandInvitation } from '../../modules/oob'
+import { InvitationType, OutOfBandInvitation } from '../../modules/oob'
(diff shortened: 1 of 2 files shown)


TimoGlastra (approved):
Nice!

PR #1585, opened 2023-09-25
fix: DID not registering on the indy ledger through external mode bec…

dinkar-jain:
…ause of missing signatures

### DIFF
--- a/packages/indy-vdr/src/dids/IndyVdrIndyDidRegistrar.ts
+++ b/packages/indy-vdr/src/dids/IndyVdrIndyDidRegistrar.ts
@@ -306,9 +306,7 @@ export class IndyVdrIndyDidRegistrar implements DidRegistrar {
         )
         didDocument = _didDocument
 
-        let didRegisterSigningKey: Key | undefined = undefined
-        if (options.options.endorserMode === 'internal')
-          didRegisterSigningKey = await verificationKeyForIndyDid(agentContext, options.options.endorserDid)
+        let didRegisterSigningKey: Key = didRegisterSigningKey = await verificationKeyForIndyDid(agentContext, options.options.endorserDid)
 
         nymRequest = await this.createRegisterDidWriteRequest({
           agentContext,


packages/indy-vdr/src/dids/IndyVdrIndyDidRegistrar.ts
```diff
@@ -306,9 +306,7 @@ export class IndyVdrIndyDidRegistrar implements DidRegistrar {
         )
         didDocument = _didDocument
 
-        let didRegisterSigningKey: Key | undefined = undefined
-        if (options.options.endorserMode === 'internal')
-          didRegisterSigningKey = await verificationKeyForIndyDid(agentContext, options.options.endorserDid)
+        let didRegisterSigningKey: Key = didRegisterSigningKey = await verificationKeyForIndyDid(agentContext, options.options.endorserDid)
```
TimoGlastra, line 309: Hey @dinkar-jain. I think this is working as intended, as with external mode, the agent does not sign the transaciton, but rather you have to send it to another agent to sign/endorse the transaction.

If you'd like to have the transaction endorsed by the agent itself, I think you want to use 'internal' mode

dinkar-jain: Yeah, but the problem is that when we try to register DID by external mode by endorsing the transaction to authorized agent, the authorized agent throws an Indy error of "Invalid signature" while registering the DID on public ledger.

Error File - aries-framework-javascript/packages/indy-vdr/src/dids/IndyVdrIndyDidRegistrar.ts
Error Line - const response = await pool.submitRequest(writeRequest)
Error Line No - 428

The most probable reason is that the didRegisterSigningKey is undefined according to the code flow. I was able to get it to work by applying the commit changes.

TimoGlastra: Is your code similar to the code as is implemented here: https://github.com/hyperledger/aries-framework-javascript/blob/main/packages/indy-vdr/tests/indy-vdr-did-registrar.e2e.test.ts#L171-L268

We have an `agent` and an `endorser` in this script. the endorser endorses the transaction. Please verify your implementation against this test, and see if there's any differences. otherwise we can look further, but by changing the code like you did, we are basically seeing external mode as internal and will try to sign the request internally. That is not what the purpose of the `external` flow is.

dinkar-jain: @TimoGlastra You are right. My implementation was a little off from that of the test cases. The code is working as intended. Sorry for the confusion.

PR #1603, opened 2023-10-09
fix: save AnonCredsCredentialRecord createdAt

genaris:
For some unfortunate reason, we are not saving `createdAt` at AnonCredsCredentialRecord creation. This is useful to do things like sorting credentials by their issuance date (something available with W3C Credential Reacords)


TimoGlastra (approved):
oops

PR #1604, opened 2023-10-10
feat: OpenId4VC support

2mau:


packages/openid4vc-client/src/utils/metadata.ts
```diff
@@ -0,0 +1,70 @@
...
+    credentialSubject: IssuerCredentialSubject
+  }
+  issuer: {
+    display?: MetadataDisplay[]
+    id: string
+  }
+}
+
+// what does this mean
+const openId4VcCredentialMetadataKey = '_paradym/openId4VcCredentialMetadata'
```
berendsliedrecht, line 26: This is a storage key that can be used to retrieve metadata about the credential. Likely for ease of access for displaying. @TimoGlastra likely knows whether we have to keep an equivalent or if this is up to the user to define.

2mau: The PR is not yet finished. Just pushed some progress

TimoGlastra: This is the display / issuer metadata. I chose to not add it as PR in AFJ as the spec is continuously evolving and thus it will contain different structures in the metadata. Not sure if that's a problem. It is very useful to have this though, but it may be better if we can generalize it and create a universal credential display format (not sure what OCA uses for this) with something like https://identity.foundation/wallet-rendering/

packages/openid4vc-holder/src/issuance/utils/IssuerMetadataUtils.ts
```diff
@@ -0,0 +1,219 @@
...
+): CredentialSupported[] {
+  const v8FormatEntries = Object.entries(supportedV8.formats)
+
+  return v8FormatEntries.map((entry) => {
+    const format = entry[0]
+    const credentialSupportBrief = entry[1]
+    if (typeof format !== 'string') {
+      throw Error(`Unknown format received ${JSON.stringify(format)}`)
+    }
+    let credentialSupport: Partial<CredentialSupported> = {}
```
berendsliedrecht, line 164: We assign a partial type here and later on cast to be fully typed. Can we get rid of the casting?

packages/openid4vc-holder/src/issuance/utils/IssuerMetadataUtils.ts
```diff
@@ -0,0 +1,219 @@
...
+export function handleLocations(authorizationDetails: AuthDetails, metadata: EndpointMetadataResult) {
+  if (metadata.credentialIssuerMetadata?.authorization_server || metadata.authorization_endpoint) {
+    if (!authorizationDetails.locations) authorizationDetails.locations = metadata.issuer
+    else if (Array.isArray(authorizationDetails.locations)) authorizationDetails.locations.push(metadata.issuer)
+    else authorizationDetails.locations = [authorizationDetails.locations as string, metadata.issuer]
+  }
+  return authorizationDetails
+}
+
+// TODO
```
berendsliedrecht, line 205: Is this still to do? If so, can you add some context as to why?

packages/openid4vc-holder/src/presentations/OpenId4VpHolderService.ts
```diff
@@ -0,0 +1,236 @@
...
+  Buffer,
+  DidsApi,
+  getJwkClassFromKeyType,
+  getKeyFromVerificationMethod,
+  injectable,
+  TypedArrayEncoder,
+  W3cJsonLdVerifiablePresentation,
+  asArray,
+} from '@aries-framework/core'
+import { CheckLinkedDomain, OP, ResponseMode, SupportedVersion, VerificationMode } from '@sphereon/did-auth-siop'
```
berendsliedrecht, line 26:
```suggestion
import { CheckLinkedDomain, OP as OpenIdProvider, ResponseMode, SupportedVersion, VerificationMode } from '@sphereon/did-auth-siop'
```

That is what it means, correct? OP seems quite vague.

packages/openid4vc-holder/src/presentations/OpenId4VpHolderService.ts
```diff
@@ -0,0 +1,236 @@
...
+    request.presentationDefinitions.length === 1 &&
+    request.presentationDefinitions?.[0]?.definition !== undefined
+  )
+}
+
+@injectable()
+export class OpenId4VpHolderService {
+  public constructor(private presentationExchangeService: PresentationExchangeService) {}
+
+  private getOp(agentContext: AgentContext) {
```
berendsliedrecht, line 51: Same comment about `OP` being vague.

```suggestion
  private getOpenIdProvider(agentContext: AgentContext) {
```

packages/openid4vc-holder/src/presentations/OpenId4VpHolderService.ts
```diff
@@ -0,0 +1,236 @@
...
+
+        const inputDescriptor = credentialsForInputDescriptor[submission.inputDescriptorId]
+        if (!inputDescriptor) {
+          credentialsForInputDescriptor[submission.inputDescriptorId] = [verifiableCredential.credential]
+        } else {
+          inputDescriptor.push(verifiableCredential.credential)
+        }
+      })
+
+    const vps = await this.presentationExchangeService.createPresentation(agentContext, {
```
berendsliedrecht, line 140: Does `vps` mean verifiablePresentations?

packages/openid4vc-holder/src/presentations/OpenId4VpHolderService.ts
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    const response = await op.createAuthorizationResponse(options.verifiedAuthorizationRequest, {
+      issuer: verificationMethod.controller,
+      presentationExchange: {
+        verifiablePresentations: vps.verifiablePresentations.map((vp) => vp.encoded as W3CVerifiablePresentation),
+        presentationSubmission: vps.presentationSubmission,
+      },
+      signature: {
+        signature: async (data) => {
```
berendsliedrecht, line 165: Can't we type data here? Do we know the encoding of the data?

packages/openid4vc-holder/src/presentations/OpenId4VpHolderService.ts
```diff
@@ -0,0 +1,236 @@
...
+
+  private getSupportedDidMethods(agentContext: AgentContext) {
+    const didsApi = agentContext.dependencyManager.resolve(DidsApi)
+    const supportedDidMethods: string[] = []
+
+    for (const resolver of didsApi.config.resolvers) {
+      supportedDidMethods.push(...resolver.supportedMethods)
+    }
+
+    return supportedDidMethods
```
berendsliedrecht, line 195: Might be good to filter out duplicates here. Very minor comment though.

packages/openid4vc-holder/src/presentations/OpenId4VpHolderService.ts
```diff
@@ -0,0 +1,236 @@
...
+  }
+
+  // TODO: we can do this in a simpler way, as we're now resolving it multiple times
+  private async getVerificationMethodFromVerifiablePresentation(
+    agentContext: AgentContext,
+    verifiablePresentation: W3cVerifiablePresentation
+  ) {
+    const didsApi = agentContext.dependencyManager.resolve(DidsApi)
+
+    let verificationMethod: string
```
berendsliedrecht, line 219:
```suggestion
    let verificationMethodId: string
```

packages/openid4vc-holder/src/presentations/PresentationExchangeService.ts
```diff
@@ -0,0 +1,360 @@
...
+    }: {
+      credentialsForInputDescriptor: CredentialsForInputDescriptor
+      presentationDefinition: PresentationDefinitionV1 | PresentationDefinitionV2
+      challenge?: string
+      domain?: string
+      nonce?: string
+      includePresentationSubmissionInVp?: boolean
+    }
+  ) {
+    // if (selectedCredentials.length === 0) {
```
berendsliedrecht, line 100: Remove commented code.

packages/openid4vc-holder/src/presentations/PresentationExchangeService.ts
```diff
@@ -0,0 +1,360 @@
...
+      const credentialsForVp = Object.values(inputDescriptors)
+        .flatMap((inputDescriptors) => inputDescriptors)
+        .map(getSphereonW3cVerifiableCredential)
+
+      const presentationDefinitionForVp = {
+        ...presentationDefinition,
+        input_descriptors: inputDescriptorsForVp,
+
+        // We remove the submission requirements, as it will otherwise fail to create the VP
+        // FIXME: Will this cause issue for creating the credential? Need to run tests
```
berendsliedrecht, line 152: Has this been tested?

packages/openid4vc-holder/src/presentations/PresentationExchangeService.ts
```diff
@@ -0,0 +1,360 @@
...
+      descriptor_map: [],
+    }
+
+    for (const vp of verifiablePresentationResults) {
+      presentationSubmission.descriptor_map.push(
+        ...vp.presentationSubmission.descriptor_map.map((descriptor): Descriptor => {
+          const index = verifiablePresentationResults.indexOf(vp)
+          const prefix = verifiablePresentationResults.length > 1 ? `$[${index}]` : '$'
+          return {
+            format: 'jwt_vp',
```
berendsliedrecht, line 200: Would be good to use an enum for this format and the one below in the `path_nested` object.

packages/openid4vc-holder/src/presentations/PresentationExchangeService.ts
```diff
@@ -0,0 +1,360 @@
...
+      if (!alg) {
+        throw new AriesFrameworkError(`No supported algs for key type: ${key.keyType}`)
+      }
+
+      const signedPresentation = await w3cCredentialService.signPresentation(agentContext, {
+        format: ClaimFormat.JwtVp,
+        verificationMethod: verificationMethod.id,
+        presentation: w3cPresentation,
+        alg,
+        challenge: challenge ?? nonce ?? utils.uuid(),
```
berendsliedrecht, line 272: we should not use a `uuid` as a challenge. `agenContext.wallet.generateNonce` can be used.

packages/openid4vc-holder/src/presentations/selection/PexCredentialSelection.ts
```diff
@@ -0,0 +1,334 @@
...
+      // FIXME: this is not collision proof as the name doesn't have to be unique
+      m.name === inputDescriptor.name
+  )
+
+  let name = inputDescriptor.name
+  // If there's no name on the input descriptor, but the id does not contain
+  // any special characters or numbers (so only letters and spaces),
+  // we will use a sanitized version of the id as the name
+  if (!name && inputDescriptor.id.match(/^[a-zA-Z ]+$/)) {
+    name = sanitizeString(inputDescriptor.id)
```
berendsliedrecht, line 287: I think this is seomthing that should be done by the user and not the framework. Might miss some details as to why we do this though.

packages/openid4vc-holder/src/presentations/selection/PexCredentialSelection.ts
```diff
@@ -0,0 +1,334 @@
+import type { PresentationSubmission, PresentationSubmissionRequirement, SubmissionEntry } from './types'
```
berendsliedrecht, line 1: I don't think there is anythinng we can do about this, but this file contains quite some complex code (and quite some TODO's). Would be nice if we can clean this up later on.

packages/openid4vc-holder/src/presentations/transform.ts
```diff
@@ -0,0 +1,65 @@
...
+  ClaimFormat,
+  JsonTransformer,
+  AriesFrameworkError,
+  W3cJsonLdVerifiablePresentation,
+  W3cJwtVerifiablePresentation,
+  W3cJwtVerifiableCredential,
+  W3cJsonLdVerifiableCredential,
+} from '@aries-framework/core'
+
+export type { SphereonW3cVerifiableCredential, SphereonW3cVerifiablePresentation }
```
berendsliedrecht, line 17: Are these exposed to the AFJ users?

packages/openid4vc-holder/tests/openid4vp-holder.e2e.test.ts
```diff
@@ -0,0 +1,83 @@
...
+
+import { AskarModule } from '@aries-framework/askar'
+import { KeyType, W3cJwtVerifiableCredential, Agent, Buffer } from '@aries-framework/core'
+import { agentDependencies } from '@aries-framework/node'
+import { ariesAskar } from '@hyperledger/aries-askar-nodejs'
+
+import { OpenId4VcHolderModule, OpenId4VpHolderService } from '../src'
+
+const modules = {
+  openId4VcClient: new OpenId4VcHolderModule(),
```
berendsliedrecht, line 11:
```suggestion
  openId4VcHolder: new OpenId4VcHolderModule(),
```

packages/openid4vc-holder/tests/openid4vp-holder.e2e.test.ts
```diff
@@ -0,0 +1,83 @@
...
+  askar: new AskarModule({
+    ariesAskar,
+  }),
+}
+
+describe('OpenId4VcClient | OpenID4VP', () => {
```
berendsliedrecht, line 17:
```suggestion
describe('OpenId4VcHolder | OpenID4VP', () => {
```

packages/openid4vc-holder/tests/openid4vp-holder.e2e.test.ts
```diff
@@ -0,0 +1,83 @@
...
+  })
+
+  afterEach(async () => {
+    await agent.shutdown()
+    await agent.wallet.delete()
+  })
+
+  describe('Mattr interop', () => {
+    // Not working yet. Once it works, we can mock the requests/responses
+    xit('Should succesfuly share a proof with MATTR launchpad', async () => {
```
berendsliedrecht, line 43: Can we enable this?

packages/openid4vc-holder/tests/openid4vp-holder.e2e.test.ts
```diff
@@ -0,0 +1,83 @@
...
+        ),
+      })
+
+      await agent.wallet.createKey({
+        keyType: KeyType.Ed25519,
+        seed: Buffer.from('00000000000000000000000000000000'),
+      })
+
+      await agent.dids.import({
+        did: 'did:key:z6MktqtXNG8CDUY9PrrtoStFzeCnhpMmgxYL1gikcW3BzvNW',
```
berendsliedrecht, line 57: If this is the did associated with the generated key form above, you can do the following:

```typescript
const key = await agent.wallet.createKey({
  keyType: KeyType.Ed25519,
  seed: Buffer.from('00000000000000000000000000000000'),
})

const did = new DidKey(key)
```

packages/openid4vc-issuer/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "prepublishOnly": "yarn run build",
+    "test": "jest"
+  },
+  "dependencies": {
+    "@aries-framework/core": "0.4.2",
+    "@sphereon/oid4vci-issuer": "^0.7.3",
+    "@sphereon/oid4vci-common": "^0.7.3",
+    "@sphereon/ssi-types": "^0.17.5",
+    "@stablelib/random": "^1.0.2",
+    "fast-text-encoding": "^1.0.6"
```
berendsliedrecht, line 32: Same comment regarding the fast-text-encoding and random.

packages/openid4vc-issuer/src/OpenId4VcIssuerApi.ts
```diff
@@ -0,0 +1,27 @@
+import type { IssueCredentialOptions, SendCredentialOfferOptions } from './OpenId4VcIssuerServiceOptions'
```
berendsliedrecht, line 1: Might be good to include this in a separate PR. Would not want to confuse the users as the functionality is not here, yet.

packages/openid4vc-issuer/tests/openid4vc-issuer.e2e.test.ts
```diff
@@ -0,0 +1,59 @@
...
+  DidKey,
+} from '@aries-framework/core'
+import { agentDependencies } from '@aries-framework/node'
+import { ariesAskar } from '@hyperledger/aries-askar-nodejs'
+import nock, { cleanAll, enableNetConnect } from 'nock'
+
+import { OpenId4VcIssuerModule } from '../src'
+
+const modules = {
+  openId4VcHolder: new OpenId4VcIssuerModule(),
```
berendsliedrecht, line 19:
```suggestion
  openId4VcIssuer: new OpenId4VcIssuerModule(),
```

packages/openid4vc-verifier/src/OpenId4VcVerifierApi.ts
```diff
@@ -0,0 +1,27 @@
+import type { IssueCredentialOptions, SendCredentialOfferOptions } from './OpenId4VcVerifierServiceOptions'
```
berendsliedrecht, line 1: Same goes for the verifier.

berendsliedrecht (changes requested):
Great work @2mau!!! I think there is some cleanup to do, but thats always expected with such a big PR.

packages/openid4vc-holder/src/issuance/OpenId4VciHolderService.ts
```diff
@@ -0,0 +1,703 @@
...
+  // Scope and authorization_details can be used in the same authorization request
+  // https://datatracker.ietf.org/doc/html/draft-ietf-oauth-rar-23#name-relationship-to-scope-param
+  if (!nonEmptyScope && !nonEmptyAuthDetails) {
+    throw new AriesFrameworkError('Please provide a scope or authorization_details')
+  }
+
+  // Authorization servers supporting PAR SHOULD include the URL of their pushed authorization request endpoint in their authorization server metadata document
+  // Note that the presence of pushed_authorization_request_endpoint is sufficient for a client to determine that it may use the PAR flow.
+  let parEndpoint = metadata.credentialIssuerMetadata?.pushed_authorization_request_endpoint
+  if (typeof parEndpoint !== 'string') parEndpoint = undefined
```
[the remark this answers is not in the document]
2mau: it is not required right?

packages/openid4vc-holder/src/issuance/OpenId4VciHolderService.ts
```diff
@@ -0,0 +1,703 @@
...
+    resolvedCredentialOffer: ResolvedCredentialOffer,
+    authCodeFlowOptions: AuthCodeFlowOptions
+  ): Promise<ResolvedAuthorizationRequest> {
+    const { credentialOfferPayload, metadata: _metadata } = resolvedCredentialOffer
+    const { metadata } = await getMetadataFromCredentialOffer(credentialOfferPayload, _metadata)
+    const codeVerifier = randomStringForEntropy(256)
+    const codeVerifierSha256 = Hasher.hash(TypedArrayEncoder.fromString(codeVerifier), 'sha2-256')
+    const codeChallenge = TypedArrayEncoder.toBase64URL(codeVerifierSha256)
+
+    // TODO: authdetails
```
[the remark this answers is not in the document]
2mau: One last thing I am unsure there is about https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-authorization-request
Basically users can request credentials with authorization_details or scopes
The mapping between scopes and credentials is out of scope.
In theory authorization details and scopes can be used in the same request,
and the authorization details param can be constructed automatically based on the requested credentials.
Since as far as I understood it all credentials can be requested via auth details it makes sense to only support auth details not scopes. But I am unsure if there are use cases for scopes. Should we just accept  both and send the provided scopes along the automatically generated authorization_details?
Or what is your opinion on this?

packages/openid4vc-holder/src/issuance/OpenId4VciHolderService.ts
```diff
@@ -0,0 +1,703 @@
...
+      let credentialTypes: string | string[]
+      if (version < OpenId4VCIVersion.VER_1_0_11) {
+        if (isInlineOffer) throw new AriesFrameworkError(`Inline credential offers not supported for version < 11`)
+        if (!credentialWithMetadata.credentialSupported.id) {
+          throw new AriesFrameworkError( // This should not happen
+            `No id provided for a credential supported entry in combination with the OpenId4VCI v8 draft`
+          )
+        }
+        credentialTypes = credentialWithMetadata.credentialSupported.id.split(`-${originalFormat}`)[0]
+      } else {
```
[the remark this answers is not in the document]
2mau: There is a check ensuring that the version is smaller then 11 right?
If the version is 11 this should not be required as the metadata is structured differently.

packages/openid4vc-holder/src/issuance/utils/Formats.ts
```diff
@@ -0,0 +1,39 @@
+import type { OID4VCICredentialFormat } from '@sphereon/oid4vci-common'
+import type { CredentialFormat } from '@sphereon/ssi-types'
+
+import { OpenId4VCIVersion } from '@sphereon/oid4vci-common'
+
+// Base on https://github.com/Sphereon-Opensource/OID4VCI/pull/54/files
```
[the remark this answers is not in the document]
2mau: nope but if they merge the PR at some point, we can remove this code as it would be contained in the lib.

packages/openid4vc-holder/src/index.ts
```diff
@@ -0,0 +1,16 @@
+import 'fast-text-encoding'
```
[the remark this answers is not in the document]
2mau: Okay, so I discussed this with Timo. No one is really sure why this is here, so I am removing it for now.
Tested issuance on mobile and node, and it works without it.
But we may need to add this back in at a later point if required.

packages/openid4vc-holder/src/issuance/OpenId4VciHolderService.ts
```diff
@@ -0,0 +1,703 @@
...
+      credentialOfferPayload,
+      issuerMetadata,
+      version
+    )
+
+    const credentialsToRequestWithMetadata = credentialsToRequest?.map((ctr) => {
+      const credentialToRequest = offeredCredentialsWithMetadata.find((offeredCredentialWithMetadata) => {
+        const { format, types } = offeredCredentialWithMetadata
+        // only requests credentials with the exact same set of types and format
+        return ctr.format === format && ctr.types.sort().join(',') === types.sort().join(',')
```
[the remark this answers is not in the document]
2mau: all I found is 'areObjectsEqual' and 'deepEquality'. Both do not work 👎  for this use case. Do you have any idea what the function is called? That helps me find it?

packages/openid4vc-holder/src/issuance/OpenId4VciHolderService.ts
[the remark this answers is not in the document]
berendsliedrecht: > Should we just accept both and send the provided scopes along the automatically generated authorization_details?

This seems like a good solution to me. `scopes` can still be used, but we let the user define this. 

Later on, we could create some kind of mapping a user can supply and go from there, but this should suffice for now.

It would be good to comment this code quite in detail as the behavior is a bit different.

packages/openid4vc-holder/src/issuance/OpenId4VciHolderService.ts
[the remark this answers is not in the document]
berendsliedrecht: For only primitives you can export the `equalsIgnoreOrder` function, from the `deepEquality` file.

packages/openid4vc-holder/src/issuance/OpenId4VciHolderService.ts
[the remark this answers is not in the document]
berendsliedrecht: I mean the typeof check.

the signature of `let parEndpoint = metadata.credentialIssuerMetadata?.pushed_authorization_request_endpoint` should be `string | undefined`, correct? If that is the case, the check below won't do anything unless someone intentionally tries to break it.

packages/openid4vc-holder/src/issuance/OpenId4VciHolderService.ts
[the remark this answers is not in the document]
berendsliedrecht: Yes there is a check for smaller than 11. But if it is larger than 11 we will enter the `else` statement. If we have a newer version, let's say `13`, we will also enter the else statement, even if that only has worked for `11`.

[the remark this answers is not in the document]
2mau: thanks

packages/openid4vc-holder/src/issuance/OpenId4VciHolderService.ts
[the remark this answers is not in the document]
2mau: Ah, yes. I changed it. It was coded that way in the Sphereons libs. Not sure why. It seems they added the correct typings later. But it shouldn't break anything to change it.

packages/openid4vc-holder/src/presentations/OpenId4VpHolderService.ts
2mau: I am using a set instead of an array now

packages/openid4vc-holder/src/presentations/selection/PexCredentialSelection.ts
2mau: Everything related to vp/siop is not yet finished. That explains the TODO's and some issues you mentioned in this review. Just the vci holder side is 'done'. So I will clean this up in the future.

packages/openid4vc-holder/src/issuance/OpenId4VciHolderService.ts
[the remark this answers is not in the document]
2mau: Okay, now I understand. In my opinion, this is the best behavior for now. Since in the case of v > 11, let's say 13, I want to set the behavior of v11 as the default, as this is more likely to be the case right?

packages/openid4vc-holder/tests/openid4vp-holder.e2e.test.ts
2mau: I will add it it the future

packages/openid4vc-holder/src/presentations/OpenId4VpHolderService.ts
2mau: i distructured the object for readability. Seemed also to me like verifiable presentations, but it contains multiple objects

packages/openid4vc-holder/src/presentations/selection/PexCredentialSelection.ts
2mau: I ll leave a comment, and look at it later

packages/openid4vc-holder/src/presentations/transform.ts
2mau: no

packages/openid4vc-holder/src/issuance/OpenId4VciHolderService.ts
```diff
@@ -0,0 +1,703 @@
...
+
+  private signCallback(agentContext: AgentContext, verificationMethod: VerificationMethod) {
+    return async (jwt: Jwt, kid?: string) => {
+      if (!jwt.header) throw new AriesFrameworkError('No header present on JWT')
+      if (!jwt.payload) throw new AriesFrameworkError('No payload present on JWT')
+      if (!kid) throw new AriesFrameworkError('No KID is present in the callback')
+
+      // We have determined the verification method before and already passed that when creating the callback,
+      // however we just want to make sure that the kid matches the verification method id
+      if (verificationMethod.id !== kid) {
```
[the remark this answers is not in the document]
2mau: We also set the kid to the verification method id. So this should always work.

packages/openid4vc-holder/src/issuance/OpenId4VciHolderService.ts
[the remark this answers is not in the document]
2mau: This is one way to implement this.
The main problem at the moment are inline credential offers.
With inline credential offers, we don't have a context, we can't provide one.
However, we also may not know the scope, which I think is possible, since not every provider will create a custom mapping of scopes to credentials.
That would mean we can't request that credential at all :( 

So in my opinion just send the context as undefined in that cases for now?
Or what are your thoughts on this?

```ts
  private getAuthDetailsFromOfferedCredential(
    credentialWithMetadata: OfferedCredentialWithMetadata,
    authDetailsLocation: string | undefined,
    version: OpenId4VCIVersion
  ): AuthDetails {
    const { format, types } = credentialWithMetadata
    const type = 'openid_credential'

    if (version < OpenId4VCIVersion.VER_1_0_11) {
      const credential_type = getV8CredentialType(credentialWithMetadata, format, version)
      return { type, credential_type, format }
    }

    const locations = authDetailsLocation ? [authDetailsLocation] : undefined
    if (format === OpenIdCredentialFormatProfile.JwtVcJson || format === OpenIdCredentialFormatProfile.JwtVcJsonLd) {
      return {
        type,
        format,
        types,
        locations,
      }
    } else if (format === OpenIdCredentialFormatProfile.LdpVc) {
      let context: string | undefined = undefined

      if (credentialWithMetadata.offerType === OfferedCredentialType.InlineCredentialOffer) {
        // Inline Credential Offers come with no context
      } else {
        if ('@context' in credentialWithMetadata.credentialSupported) {
          context = credentialWithMetadata.credentialSupported['@context'] as unknown as string
        } else {
          throw new AriesFrameworkError('Could not find @context in credentialSupported.')
        }
      }

      return {
        type,
        format,
        types,
        locations,
        '@context': context,
      }
    } else {
      throw new AriesFrameworkError(`Cannot create authorization_details. Unsupported credential format ${format}.`)
    }
  }
```

packages/openid4vc-holder/src/issuance/OpenId4VciHolderService.ts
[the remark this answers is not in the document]
berendsliedrecht: IMO it would be better to error out as a lot of undefined behavior can happen because of this. It is exactly "likely" that it is correct, but not 100%.

packages/openid4vc-holder/src/issuance/OpenId4VciHolderService.ts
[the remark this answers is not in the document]
berendsliedrecht: What are the implications of returning an undefined context? Would not just make more sense for that unique scenario to just error out for now? If there is a need from the users we can check it again, but it seems quite niche that it might be a waste of time to get it perfect.

packages/openid4vc-holder/src/issuance/OpenId4VciHolderService.ts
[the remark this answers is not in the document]
berendsliedrecht: I think almost all code has a fallback to v11 so it does not really matter if we change this here, you can leave it as-is.

packages/openid4vc-holder/src/OpenId4VcHolderApi.ts
```diff
@@ -0,0 +1,155 @@
...
+    redirect_url: string
+    clientMetadata?: ClientMetadataOpts
+    issuer?: string
+  }) {
+    const relyingParty = await this.openId4VpHolderService.getRelyingParty(this.agentContext, options)
+
+    // TODO: generate nonce, state, correlationId
+    const nonce = 'qBrR7mqnY3Qr49dAZycPF8FzgE83m6H0c2l0bzP4xSg'
+    const state = 'b32f0087fc9816eb813fd11f'
+    const correlationId = '1'
```
berendsliedrecht, line 46: This should be fixed before merging, correct?

packages/openid4vc-holder/src/OpenId4VcHolderApi.ts
```diff
@@ -0,0 +1,155 @@
...
+  ): Promise<W3cCredentialRecord[]> {
+    return this.openId4VcHolderService.acceptCredentialOffer(this.agentContext, {
+      resolvedCredentialOffer,
+      acceptCredentialOfferOptions,
+    })
+  }
+
+  /**
+   * Accepts a credential offer using the authorization code flow.
+   * @param resolvedCredentialOffer Obtained through @function resolveCredentialOffer
```
berendsliedrecht, line 137: `@function` does not seem like valid ts-doc. Maybe check the `@link`?

packages/openid4vc-holder/src/issuance/OpenId4VciHolderService.ts
```diff
@@ -0,0 +1,782 @@
...
+      authorizationRequestUri,
+    }
+  }
+
+  public async acceptCredentialOffer(
+    agentContext: AgentContext,
+    options: {
+      resolvedCredentialOffer: ResolvedCredentialOffer
+      acceptCredentialOfferOptions: AcceptCredentialOfferOptions
+      resolvedAuthorizationRequest?: ResolvedAuthorizationRequest & { code: string }
```
berendsliedrecht, line 361: Can't we add the `{code:string}` to the `ResolvedAuthorizationRequest`? Or create a new type?

packages/openid4vc-holder/src/presentations/OpenId4VpHolderService.ts
```diff
@@ -0,0 +1,524 @@
+import type { PresentationSubmission } from './selection'
```
berendsliedrecht, line 1: Ah I wasn't aware the this implementation of vp. I think if there is some code already there it would be good to add this in a separate PR.

packages/openid4vc-issuer/src/index.ts
```diff
@@ -0,0 +1,6 @@
+export * from './OpenId4VcIssuerApi'
+export * from './OpenId4VcIssuerModule'
+export * from './OpenId4VcIssuerService'
+
+// Contains internal types, so we don't export everything
+export {} from './OpenId4VcIssuerServiceOptions'
```
berendsliedrecht, line 6: We can just omit this line, right?

packages/openid4vc-holder/src/OpenId4VcHolderApi.ts
2mau: Yes

PR #1607, opened 2023-10-17
feat(sd-jwt-vc): Module for Issuer, Holder and verifier

berendsliedrecht:
Module for interacting with SD-JWTs as an issuer, holder and verifier.

The module is called `SdJwtModule`, however since we only deal with `SdJwtVc` I think it should be renamed to `SdJwtVcModule`.

As an issuer you get the `SdJwtModule.create` function which allows you to create an sd-jwt-vc where the framework sets all the required properties (like `iat`). It returns a record and a compact form of the sd-jwt-vc. Since key binding is required in sd-jwt-vc, a `holderDidUrl` is required for creation, which will be turned into a `JWK` and embedded in the payload as a `cnf` claim.

The holder gets the `SdJwtModule.receive` and `SdJwtModule.present`. The receive function simply validates that the signature is correct and the correct holder key is used. The present method allows a user to remove certain items from the sd-jwt-vc. The present method returns a compact format of the sd-jwt-vc which can be send to the verifier.

The verifier gets the `SdJwtModule.verify` method. This checks the keybinding, issuer signature and whether all the required fields are included.

Note: The module also exposes some repository functionality as we do not have a generic repository module, yet.


packages/core/src/crypto/jose/jwk/transform.ts
```diff
@@ -47,3 +47,17 @@ export function getJwkClassFromJwaSignatureAlgorithm(alg: JwaSignatureAlgorithm
 export function getJwkClassFromKeyType(keyType: KeyType) {
   return JwkClasses.find((jwkClass) => jwkClass.keyType === keyType)
 }
+
+/**
```
TimoGlastra, line 51: Key types can support multiple JWAs. For that reason I choose to only support key -> jwk, and then on a JWK class you can see the supported algorithms.

This is going to be an issue down the line

packages/sd-jwt/src/SdJwtService.ts
```diff
@@ -0,0 +1,179 @@
...
+    if (!isValid) {
+      throw new SdJwtError(`sd-jwt is not valid.`)
+    }
+
+    // TODO: append holder key here
+    const compact = await sdJwt.toCompact()
+
+    return new SdJwtRecord({
+      sdJwt: compact,
+    })
```
TimoGlastra, line 156: Are we interested in supporting generic SD-JWTs? Wouldn't we rather want to support the VC variant of SD-JWT and be able to re-use the W3C Credential record for this with a new format?

packages/sd-jwt/src/repository/SdJwtRecord.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    if (props) {
+      this.id = props.id ?? utils.uuid()
+      this.createdAt = props.createdAt ?? new Date()
+      this.sdJwt = props.sdJwt
+      this._tags = props.tags ?? {}
+    }
+  }
+
+  public getTags() {
```
TimoGlastra, line 40: No tags at all will make it especially hard to query the credentials for input in a VP...

yarn.lock
```diff
@@ -10,6 +10,39 @@
     "@jridgewell/gen-mapping" "^0.3.0"
     "@jridgewell/trace-mapping" "^0.3.9"
 
+"@aries-framework/core@0.4.1":
```
TimoGlastra, line 13: this is a mistake I think

packages/sd-jwt/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "scripts": {
+    "build": "yarn run clean && yarn run compile",
+    "clean": "rimraf ./build",
+    "compile": "tsc -p tsconfig.build.json",
+    "prepublishOnly": "yarn run build",
+    "test": "jest"
+  },
+  "dependencies": {
+    "@aries-framework/askar": "^0.4.2",
+    "@aries-framework/core": "0.4.1",
```
TimoGlastra, line 28: This should be updated

packages/sd-jwt/src/SdJwtService.ts
berendsliedrecht: Tbh I would not like to restrict the user here. Might need to have a short discussion on the relation between the w3c service and the sdjwt service.

packages/core/src/crypto/jose/jwk/transform.ts
berendsliedrecht: Yeah that is true. I would like to create the header for the user as we can then just accept a VC, which would be the payload, and nothing else. Do you have a suggestion on how we can create a header, which IIRC only contains the `alg` and `typ` as required?

TimoGlastra: Yeah I think we can do that, but that we need to pick the default JWA differently than is done with this method.

See this code: https://github.com/hyperledger/aries-framework-javascript/blob/b2ba7c7197139e780cbb95eed77dc0a2ad3b3210/packages/core/src/crypto/JwsService.ts#L28-L36

From keyType we can go to JWK and from a JWK we can then pick the first supported signature algorithm. But we should allow the user to specify the alg, so they can override it.

You can get the JWK class using this method: https://github.com/hyperledger/aries-framework-javascript/blob/b2ba7c7197139e780cbb95eed77dc0a2ad3b3210/packages/core/src/crypto/jose/jwk/transform.ts

And then statically extract the signature algorithms from the class: https://github.com/hyperledger/aries-framework-javascript/blob/b2ba7c7197139e780cbb95eed77dc0a2ad3b3210/packages/core/src/crypto/jose/jwk/Ed25519Jwk.ts#L15

packages/sd-jwt/src/SdJwtService.ts
berendsliedrecht: We discussed this a bit, and right now (as a start) we will only support sd-jwt-vc. Now the question is whether we make the module 'sd-jwt-vc' / 'SdJwtVcModule' or leave it as sd-jwt.

packages/sd-jwt/src/repository/SdJwtRecord.ts
berendsliedrecht: yeah I was wondering what the best tag structure would be for this. Right now I added the disclosure keys as an array, but I am not sure whether that is supported within askar.


Will likely also include the claimKeys, combined with the selectively disclosable keys.

packages/sd-jwt/src/SdJwtOptions.ts
```diff
@@ -0,0 +1,28 @@
+import type { HashName, Key } from '@aries-framework/core'
+import type { DisclosureFrame } from 'jwt-sd'
+
+export type SdJwtCreateOptions<Payload extends Record<string, unknown> = Record<string, unknown>> = {
```
berendsliedrecht, line 4: @TimoGlastra this is the input now for the creation of a sd-jwt-vc. Holder binding is required and the issuer has to provide a did, for the `iss` field and a key. 

Very curious to see if there are any alternatives for this as I think the options are a bit of a mess...

packages/sd-jwt/src/repository/SdJwtRecord.ts
TimoGlastra: You can add arrays. It will currently be added as an `<attr>:<value>: 1/0` I think. So `disclosureKeys:age: 1`. Maybe askar also supports something for this natively, which we maybe should look into

packages/sd-jwt/src/SdJwtService.ts
TimoGlastra: I think it should become a service in the W3cCredentialsModule then. We already have services for JsonLd and Jwt, we can add the SD part to it? Or maybe it becomes a completely separate module and we call it SdJwtCredentialsModule (to make it consistent with W3cCredentials?)

packages/sd-jwt/src/SdJwtOptions.ts
```diff
@@ -0,0 +1,28 @@
...
+  issuerDid: string
+  issuerKey: Key
+  disclosureFrame?: DisclosureFrame<Payload>
+  holderBinding: string | Key
+  hashingAlgorithm?: HashName
+}
+
+export type SdJwtReceiveOptions = {
+  issuerDid: Key
+  holderKey?: Key
```
TimoGlastra: These will both already be in the JWT right?

packages/sd-jwt/src/SdJwtOptions.ts
TimoGlastra: Why does the issuer need to provide both an issuerDid and an issuerKey? I think with other places we've required to pass a key reference within the did then (`did:key:123#123`) and based on that we can get the issuerDid and issuerKey.

Usually I try to avoid two params that can clash (as you can provide a key that doesn't belong to the did). We can call the parameter the `keyId` (or something like that, forgot what it's called) and also add it to the header of the JWT

packages/sd-jwt/src/SdJwtOptions.ts
berendsliedrecht: I changed it to `issuerKey` and `holderkey` which are used to confirm the signature and check whether the `cnf` claim is correct. If this info is invalid the sd-jwt cannot be used.

berendsliedrecht:
Somehow I can't reply to your comment...

> Why does the issuer need to provide both an issuerDid and an issuerKey? I think with other places we've required to pass a key reference within the did then (did:key:123#123) and based on that we can get the issuerDid and issuerKey.

> Usually I try to avoid two params that can clash (as you can provide a key that doesn't belong to the did). We can call the parameter the keyId (or something like that, forgot what it's called) and also add it to the header of the JWT

Yeah I was thinking about this for a bit and a did with key reference would be good. I did not do it like that for now as you would need to resolve your own. I will add it where you have to pass a did with key reference. It will make the method slower though.

packages/sd-jwt/src/SdJwtService.ts
TimoGlastra: sd-jwt-vc is compatible with the W3c spec though right? You're saying we will only support `sd-jwt-vc`

berendsliedrecht: sd-jwt-vc is not compatible with the w3c data model specs.

TimoGlastra: Ok, makes sense then 👍

packages/sd-jwt/src/SdJwtOptions.ts
berendsliedrecht: now using `issuerDidUrl` and `holderDidUrl`.

TimoGlastra:
> Somehow I can't reply to your comment...
> 
> > Why does the issuer need to provide both an issuerDid and an issuerKey? I think with other places we've required to pass a key reference within the did then (did🔑123#123) and based on that we can get the issuerDid and issuerKey.
> 
> > Usually I try to avoid two params that can clash (as you can provide a key that doesn't belong to the did). We can call the parameter the keyId (or something like that, forgot what it's called) and also add it to the header of the JWT
> 
> Yeah I was thinking about this for a bit and a did with key reference would be good. I did not do it like that for now as you would need to resolve your own did. I will add it where you have to pass a did with key reference. It will make the method slower though.

I think that is fine for now. We have this in a few other places, where we resolve our own did. But I thin we should solve this generically by allowing the did to be retrieved from local storage / cache.

packages/sd-jwt/src/repository/SdJwtRecord.ts
```diff
@@ -0,0 +1,98 @@
...
+export type SdJwt<
+  Header extends Record<string, unknown> = Record<string, unknown>,
+  Payload extends Record<string, unknown> = Record<string, unknown>
+> = {
+  disclosures?: Array<DisclosureItem>
+  header: Header
+  payload: Payload
+  signature: Uint8Array
+
+  // TODO: include the holder key for key binding in here, for ease of use
```
berendsliedrecht, line 20: @TimoGlastra I would like to include the holder key that is bound to this SD-JWT. We can store the public key of the pair and derive it with Askar from there. Or we include the didurl and handle it like we do in the services now by resolving it and using askar to get the private key. 

Do you have a preference / other idea for this?

berendsliedrecht: Storing the `holderDidUrl` for now.

packages/sd-jwt/src/SdJwtService.ts
```diff
@@ -0,0 +1,340 @@
...
+        `issuer did url '${issuerDidUrl}' does not contain a '#'. Unable to derive key from did document`
+      )
+    }
+
+    const { verificationMethod: issuerVerificationMethod, didDocument: issuerDidDocument } = await this.resolveDidUrl(
+      agentContext,
+      issuerDidUrl
+    )
+    const issuerKey = getKeyFromVerificationMethod(issuerVerificationMethod)
+    const alg = getJwaFromKey(issuerKey, issuerOverrideJsonWebAlgorithm)
```
TimoGlastra, line 124:
```suggestion
    const issuerKey = getKeyFromVerificationMethod(issuerVerificationMethod)
    const issuerJwa = jwa /* function input parameter */ ?? getJwkFromKey(issuerKey).supportedSigningAlgorithms[0]
    const alg = getJwaFromKey(issuerKey, issuerOverrideJsonWebAlgorithm)
```

TimoGlastra: I really would like to avoid defining which Jwa alg is associated with which key type. I think it should be key -> jwk -> jwa's associated with jwk

TimoGlastra (approved):
Nice work @berendsliedrecht. Mostly some small improvements

packages/core/src/crypto/jose/jwk/Jwk.ts
```diff
@@ -28,10 +28,7 @@ export abstract class Jwk {
   public use?: string
 
   public toJson(): JwkJson {
```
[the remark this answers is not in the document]
berendsliedrecht: I mainly did it for tests as I wanted to match on the entire object without specifying `undefined` fields. I will revert this change and fix the tests.

packages/sd-jwt/src/SdJwtApi.ts
```diff
@@ -0,0 +1,91 @@
...
+  ): Promise<{ sdJwtRecord: SdJwtRecord; compact: string }> {
+    return await this.sdJwtService.create<Payload>(this.agentContext, payload, options)
+  }
+
+  /**
+   *
+   * Validates and stores an sd-jwt from the perspective of an holder
+   *
+   */
+  public async receive(sdJwtCompact: string, options: SdJwtReceiveOptions): Promise<SdJwtRecord> {
```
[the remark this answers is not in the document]
berendsliedrecht: I think we can leave it as is for now. How we deal with storage will be reworked in the future and I believe that that would be a good time to get consistent APIs in place. After that I'd likely call it just `validate` or something.

packages/sd-jwt/src/SdJwtApi.ts
```diff
@@ -0,0 +1,91 @@
...
+    Header extends Record<string, unknown> = Record<string, unknown>,
+    Payload extends Record<string, unknown> = Record<string, unknown>
+  >(
+    sdJwtCompact: string,
+    options: SdJwtVerifyOptions
+  ): Promise<{ sdJwtRecord: SdJwtRecord<Header, Payload>; validation: SdJwtVcVerificationResult }> {
+    return await this.sdJwtService.verify<Header, Payload>(this.agentContext, sdJwtCompact, options)
+  }
+
+  public async getCredentialRecordByIdM<
```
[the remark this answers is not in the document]
berendsliedrecht: Not too opposed to keeping or removing it, so I will remove it for now.

packages/sd-jwt/src/SdJwtOptions.ts
```diff
@@ -0,0 +1,42 @@
+import type { HashName, JwaSignatureAlgorithm } from '@aries-framework/core'
+import type { DisclosureFrame } from 'jwt-sd'
+
+export type SdJwtCreateOptions<Payload extends Record<string, unknown> = Record<string, unknown>> = {
+  holderDidUrl: string
+  issuerDidUrl: string
+  issuerOverrideJsonWebAlgorithm?: JwaSignatureAlgorithm
```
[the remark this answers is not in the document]
berendsliedrecht: `override` I added because it overrides the auto-detected jwa from the issuer did.
`issuer` because there is also a holder.

Will remove both, likely a bit too verbose here.

packages/sd-jwt/src/SdJwtOptions.ts
```diff
@@ -0,0 +1,42 @@
...
+    issuedAt: number
+  }
+}
+
+/**
+ * `requiredClaimKeys` is not the best API, but it is the best alternative until something like `PEX` is supported
+ */
+export type SdJwtVerifyOptions = {
+  holderDidUrl: string
+  verifierDid: string
```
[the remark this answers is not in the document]
berendsliedrecht: The verifierDid is included in the JWT challenge. SO the verifier sends some metadata to the holder, which they then sign with their `cnf` claim key. The verifier needs to check whether the `verifierDid` was actually the same.

packages/sd-jwt/src/SdJwtService.ts
berendsliedrecht: I can do it inline, but `getJwaFromKey` just does the same thing.

berendsliedrecht:
@karimStekelenburg @TimoGlastra Quickly before merging, should the module be called `SdJwtVc` or `SdJwt`? It is `SdJwt` right now but we only really handle `SdJwtVc`.

packages/sd-jwt/src/repository/SdJwtRecord.ts
berendsliedrecht: Tags will be last as-is for now. Once we do the integration with openid they can be added as at that point we will know exactly how to query.

packages/sd-jwt/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "prepublishOnly": "yarn run build",
+    "test": "jest"
+  },
+  "dependencies": {
+    "@aries-framework/askar": "^0.4.2",
+    "@aries-framework/core": "^0.4.2",
+    "class-transformer": "0.5.1",
+    "class-validator": "0.14.0",
+    "jwt-sd": "^0.1.0"
+  },
```
karimStekelenburg, line 32: We can probably push this to `v0.1.0` right?

karimStekelenburg (commented):
Looking good overall. I left a few comments here and there, but they're mostly suggestions. I only see happy flow tests though. Doesn't have to bne noew, but I think we should add sad flow tests when we combine it with OID4VC.

packages/sd-jwt/src/SdJwtApi.ts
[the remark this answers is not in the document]
karimStekelenburg: But the `storeCredential` and this method both validate and store right? If that is the case I don't see why we should name them differently. Even if it's just for the time being.

karimStekelenburg (approved):

packages/sd-jwt/src/__tests__/SdJwtService.test.ts
```diff
@@ -0,0 +1,527 @@
...
+          ['salt', 'email', 'johndoe@example.com'],
+          ['salt', 'region', 'Anystate'],
+          ['salt', 'country', 'US'],
+          ['salt', 'given_name', 'John'],
+        ])
+      )
+    })
+  })
+
+  describe('SdJwtService.present', () => {
```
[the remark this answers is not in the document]
berendsliedrecht: Yeah good point regarding the non-happy flow tests. I would like to add them in a separate PR to keep this one "small".

packages/sd-jwt/src/SdJwtOptions.ts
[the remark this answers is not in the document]
TimoGlastra: > SO the verifier sends some metadata to the holder,

Is it 'some metadata' or is it specifically the verifiedDid? Because otherwise it may be more compatible for future api changes to make it more generic (such as `verifierChallenge`). A challenge can be anything.

packages/sd-jwt/src/SdJwtService.ts
TimoGlastra: Not when I checked this afternoon? It was still the harcoded if keyType === Ed25519 return 'Eddsa' kinda method

packages/sd-jwt/src/SdJwtApi.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  public async getCredentialRecordById(id: string): Promise<SdJwtRecord> {
+    return await this.sdJwtService.getCredentialRecordById(this.agentContext, id)
+  }
+
+  public async getAllCredentialRecords(): Promise<Array<SdJwtRecord>> {
+    return await this.sdJwtService.getAllCredentialRecords(this.agentContext)
+  }
+
+  public async findCredentialRecordsByQuery(query: Query<SdJwtRecord>): Promise<Array<SdJwtRecord>> {
```
genaris, line 77: Following the convention used in other module APIs (and services in general) we'd call it `findAllCredentialRecordsByQuery`, to make more explicit about the fact that we expect multiple records to be retrieved. Actually, in most modules we make naming simpler (e.g. `findAllByQuery`), but I see it's not the case here and `W3cCredentialsApi` (the API this module is based on I guess).

Of course if we change this here and leave `W3cCredentialsApi` as is, it would not be consistent. And if we change it in `W3cCredentialsApi` as well we'll introduce a breaking change. But I just wanted to point out this to bother you 😄

packages/sd-jwt/tests/sdJwtVc.e2e.test.ts
```diff
@@ -0,0 +1,145 @@
...
+
+import { agentDependencies } from '../../core/tests'
+import { SdJwtModule } from '../src/SdJwtModule'
+
+const getAgent = (label: string) =>
+  new Agent({
+    config: { label, walletConfig: { id: utils.uuid(), key: utils.uuid() } },
+    modules: {
+      sdJwt: new SdJwtModule(),
+      // @ts-ignore
```
genaris, line 24: What's the problem here that required a ts-ignore? It's about not using askar-nodejs as direct dependency?

packages/sd-jwt/tests/setup.ts
```diff
@@ -0,0 +1,3 @@
+import 'reflect-metadata'
+
+jest.setTimeout(120000)
```
genaris, line 3: Do you expect SD-JWT to take that long?

genaris (approved):
I like the simplicity of the API, seems to be very fun to use!

Just left a comment regarding an old discussion that we never completely materialized in code (maybe for a further breaking change PR now we are introducing a new major version).

packages/sd-jwt/src/SdJwtApi.ts
berendsliedrecht: Ah good point! I think we can change this in a separate PR which changes both the w3c and this one.

packages/sd-jwt/tests/setup.ts
berendsliedrecht: Copied from another test. It is quite quick, but askar and agent initialization makes some steps a bit slow. Better overestimate here I guess ;).

packages/sd-jwt/tests/sdJwtVc.e2e.test.ts
berendsliedrecht: The expected ariesAskar type did not match the one from the module as I think it used 0.2.0-dev.1 and I used 0.1.0. I think I can remove it again

packages/sd-jwt/src/SdJwtApi.ts
berendsliedrecht: I am actually not sure what the best name would be here. I think `findAllByQuery` is the best as it is already scoped under `sdJwt` within the `agent.modules` api.

packages/sd-jwt/src/SdJwtApi.ts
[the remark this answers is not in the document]
berendsliedrecht: I changed it to `storeCredential`.

PR #1618, opened 2023-10-30
feat: refresh the pool when connecting

wadeking98:
Previously AFJ wouldn't call refresh on genesis files, this meant that the pool could be out of date and contain dead nodes. This change makes AFJ refresh the pool when connecting to it, to ensure the pool stays up to date


TimoGlastra (commented):
I'm curious how this improves performance. As it won't store the updated genesis files, it will need to do this on every startup right?

wadeking98:
> I'm curious how this improves performance. As it won't store the updated genesis files, it will need to do this on every startup right?

Yes, it will need to do this on every startup. I was thinking I could create some method `pool.update()` or something that would return the transactions but I couldn't figure out how to expose a method from the pool service to something using the agent

wadeking98:
I've actually made an improvement on this here: https://github.com/hyperledger/aries-framework-javascript/pull/1623

PR #1623, opened 2023-10-31
feat: added ability to refresh the pool manually

wadeking98:
Previously AFJ wouldn't call refresh on genesis files, this meant that the pool could be out of date and contain dead nodes. This change adds a function to the pool service that allows an afj user to manually refresh the pool as needed. Example:
```typescript
const poolService = newAgent.dependencyManager.resolve(IndyVdrPoolService)
await poolService.refreshPoolConnections()
```
all the pools are refreshed asynchronously so it doesn't take more than 1 - 2 seconds on BC Wallet.

Note: I tried returning the transactions too so that we could cache them, however when I was playing around with the yarn patch I made, whenever I called `this._pool.transactions` in the `IndyCdrPool.js` file I would get a JSON parse error. I was unclear what was causing the error, it seems to be coming from the indy-vdr-shared library


TimoGlastra:
Thanks @wadeking98! We can probably expose this method as well on the IndyVdrApi, so you can acces it on agent.modules.indyVdr.xxx


Re the parse error, the transactios is not a valid json array/object. As each line is a separate json object, but it's not constructed as an array. Maybe we do some incorrect parsing in the indy vdr lib..

TimoGlastra:
It seems the react native wrappers calls json.parse in poolGetTransactions, while the Node.JS wrapper doesn't: https://github.com/hyperledger/indy-vdr/blob/main/wrappers/javascript/indy-vdr-react-native/src/ReactNativeIndyVdr.ts

@berendsliedrecht could you make a quick fix for this in Indy VDR?

berendsliedrecht:
> poolGetTransactions

https://github.com/hyperledger/indy-vdr/pull/234

berendsliedrecht:
> It seems the react native wrappers calls json.parse in poolGetTransactions, while the Node.JS wrapper doesn't: https://github.com/hyperledger/indy-vdr/blob/main/wrappers/javascript/indy-vdr-react-native/src/ReactNativeIndyVdr.ts
> 
> @berendsliedrecht could you make a quick fix for this in Indy VDR?

https://github.com/hyperledger/indy-vdr/pull/234

packages/indy-vdr/src/pool/IndyVdrPoolService.ts
```diff
@@ -156,6 +156,13 @@ export class IndyVdrPoolService {
     return Promise.allSettled(this.pools.map((pool) => pool.refreshConnection()))
   }
 
+  /**
+   * Get all pool transactions
+   */
+  public getAllPoolTransactions() {
+    return Promise.allSettled(this.pools.map((pool) => pool.transactions))
```
TimoGlastra, line 163: This is not a promise right? So then allSettled is not needed I think?

packages/indy-vdr/src/IndyVdrApi.ts
```diff
@@ -13,7 +13,7 @@ import { multiSignRequest, signRequest } from './utils/sign'
 @injectable()
 export class IndyVdrApi {
   private agentContext: AgentContext
-  private indyVdrPoolService: IndyVdrPoolService
+  public indyVdrPoolService: IndyVdrPoolService
```
TimoGlastra, line 16: We shouldn't make indyVdrPoolService public, but rather add the methods we want to expose to this class as well

packages/indy-vdr/src/pool/IndyVdrPoolService.ts
wadeking98: `pool.transactions` returns a promise

packages/indy-vdr/src/IndyVdrApi.ts
wadeking98: okay got it

genaris (approved):
Thanks @wadeking98! Do you think it could be useful to add some JSDoc comments in the new methods for `IndyVdrApi`? They seem to be self-explanatory (and do not get any input parameter but probably can help others to notice about their use cases (e.g. the performance issue that raised this PR).

wadeking98:
@genaris @TimoGlastra I don't have permissions to merge this so one of you will have to do it when you're ready

genaris:
> @genaris @TimoGlastra I don't have permissions to merge this so one of you will have to do it when you're ready

Sure, can you fix the [Validate](https://github.com/hyperledger/aries-framework-javascript/actions/runs/6750897503/job/18354084292?pr=1623) error? A `yarn prettier -w` should do the trick I think

Issue #1625, opened 2023-11-02
Mediator instance not consuming messages received by another mediator instance


ericvergnaud:
Hi,

following standard SRE architecture, we're running 2 mediator instances behind an ALB (application load balancer).

To ensure that queued messages are accessible from both instances (and also survive a restart), we've implemented a PostgreSQLMessageRepository which we use instead of the default InMemoryMessageRepository.

This works well when the mediated wallet connects (i.e. it picks all messages stored since last connection), but not _during_ a mediation session. 

The failing scenario, as we understand it so far, is as follows:
 - we have 2 mediator instances, named A and B
 - mobile connects to mediator A, initiates pickup (v2), finds all messages 
 - whilst mobile is connected, ledger sends _n_ messages to mediator cluster(i.e. ALB)
 - roughly half of the messages are sent to mediator A, which forwards them to mobile via WebSocket
 - the other half are sent to B, which queues them (via PostgreSQLMessageRepository)
 - after pickup initiation (processing of status-request message), mediator A never looks for messages queued by other instances.
 
We have a few questions:
 - is the above analysis correct ?
 - if it is, is there a config option (other than use pickup v1) that forces the mediator to regularly check for queued messages ?
 - if there isn't, a modern event-driven architecture would consist in plugging a CDC (Change Data Capture) in Postgres that would then notify each mediator instance about newly queued messages. Is there a mediator endpoint that can be called for that ?
 - an ugly alternative is to have the mediator regularly poll the MessageRepository for new messages and process them.

What do you think ?

TimoGlastra:
Hey @ericvergnaud. I think the behaviour you're describing is correct. 

I know @genaris has been working on improvements to the mediator implementation, not sure if there's any improvements being made to this specific functionality though.

There's currently no config option to make it check regularly for queued messages. As you describe, I think we should have some sort of push mechanism that notifies all the mediators of a new message, and if there's a mediator session open, it can directly push that message to the client.

We're also currently using Postgres for storing mediated messages, but also exploring redis for this (which supports notifications using pub/sub). There's no endpoint in the mediator, but this is something you could set up as part of your custom PostgreSQLMessageRepository. (either set up a new server, or add it to the HttpInboundTransport, but make sure it's secured).

We could explore some good options to be added to the framework by default (or as an extension package) such as Redis and PostgresQL, but I think it would be good to first implement something and then look at maybe integrating it into the framework.

[...]

Issue #1199, opened 2023-01-10
Only queue messages that have a mediator record associated


TimoGlastra:

genaris:
In my opinion, it's not strictly wrong for an agent to queue messages for an agent that is not among its 'mediatees', as it might be useful for certain services to temporarily hold messages for agents that don't have a public endpoint and don't support a stable or long-lived transport protocol. That was my first understanding when I saw that Route Coordination (RFC 211) and Pickup (RFC 212) protocols were decoupled and the latter didn't mention that the _message holder_ should be a mediator. Pickup V2, however, does call it explicitly "mediator", which I'm not sure if it's intentional or it's because it was the most common use case (maybe it's worth to open an issue there to clarify).

PR #1638, opened 2023-11-15
feat!: message pickup live mode support

genaris:
As promised, some progress in supporting missing features from [RFC 0685](https://github.com/hyperledger/aries-rfcs/tree/main/features/0685-pickup-v2), most notably the "Live Mode", in such a way that might be compatible with a multi-instance architecture as discussed in #1625.

Conceptually, it consists mostly in changes in Mediator role, where we can choose our strategy when a Forward message is received:  
- We can simply queue it to the `MessagePickupRepository`, so it will be in charge of manually trigger a delivery of queued messages: this can be possible because `MessagePickupModule` now emits events when Live Sessions are opened and closed (so any consumer of an AFJ instance can subscribe to a central Message Pickup Repository and know if a newly queued message belongs to any of its connected clients), and `MessagePickupApi` exposes a method to deliver any queued message using V2 protocol.
- We can try to deliver it immediately using V2 protocol, previously adding it to the queue and . This is mostly useful in single instance AFJ scenarios, yet taking advantage of the protocol-level acknowledges Pickup V2 provides
- We can try to deliver the encrypted message immediately, without any protocol-level encapsulation. This is what AFJ does now, in a kind of 'implicit mode' (not recommended as it is more unreliable). I left it as the default value for now, just to make it safer.

On `mediatee` side, there is a new `MediatorPickupStrategy` called `PickUpV2LiveMode` where, once it connects to Mediator WebSocket, it will set Live Delivery mode on (retrieving, as usual, any pending message). Current `PickUpV2` mode will work in polling mode (sending a `status-request` message every `mediatorPollingInterval` milliseconds. This represents a breaking change in terms of settings, as previously it was just sending a Trust Ping and expected the mediator to send messages implicitly.

Some changes done:

- `MediatorModule` now has a config parameter to select its message forwarding strategy
- `MessagePickupApi` includes Live Mode session service that keeps track of all connected clients using V2 protocol in Live Mode and exposes a method to deliver any queued message to them. When new sessions are added or removed, it emits events
- `MessageRepository` was renamed to `MessagePickupRepository` (any better naming is more than welcome!) and methods were changed to allow removal of messages by their id and filter by connectionId/recipientKey
- `MessagePickupRepository` is now completely responsibility of `MessagePickupModule`: an in-memory default implementation will be instantiated if not specified in its config or defined externally before instantiating the Agent
- `TransportService` now emits events when sessions are created and removed. This was done mainly to be consumed by `MessagePickupSessionService`, but they can also be useful externally

There are several TODOs that I would like to handle in further PRs and probably move to 0.5.x or 0.6.0 so we can have more time to fix and improve the architecture:

- [ ] Add tests for all the different settings, as currently we only have some simple E2E tests
- [ ] Remove `MessageSender` dependency on messagePickupRepository, e.g. by emitting an event when a packed message should be added to the pickup queue (which will be caught by MessagePickupApi). This will allow to remove InjectionSymbol 'MessagePickupRepository'.
- [ ] Add somewhere a parameter to use only Queue Transport for mediatees or specific connections (see #1199). A simple way can be to add a field to connection records but this would certainly introduce an storage change so it will be better left to a 0.6 (where we expect other storage changes).

Some issues related to the spec that we'll need to double check:

- Filtering by recipientKey doesn't sound that hard to implement, but Pickup V2 protocol is so open that it can allow a client to create a dedicated Live Mode session for a single recipient key
- It does not explicitly tells about that, but the protocol assumes that `delivery` message always come as a response to a `delivery-request`. Otherwise, how can I know what's the limit of messages I can batch on it (or if I should set recipient_key parameter)?


packages/core/src/transport/TransportEventTypes.ts
```diff
@@ -1,8 +1,11 @@
 import type { BaseEvent } from '../agent/Events'
+import type { TransportSession } from '../agent/TransportService'
 
 export enum TransportEventTypes {
   OutboundWebSocketClosedEvent = 'OutboundWebSocketClosedEvent',
   OutboundWebSocketOpenedEvent = 'OutboundWebSocketOpenedEvent',
+  TransportSessionSaved = 'TransportSessionSaved ',
+  TransportSessionRemoved = 'TransportSessionSaved ',
```
TimoGlastra, line 8:
```suggestion
  TransportSessionRemoved = 'TransportSessionRemoved ',
```

packages/core/src/modules/message-pìckup/protocol/v2/V2MessagePickupProtocol.ts
```diff
@@ -4,26 +4,36 @@ import type { AgentMessageReceivedEvent } from '../../../../agent/Events'
...
 import { EventEmitter } from '../../../../agent/EventEmitter'
 import { AgentEventTypes } from '../../../../agent/Events'
 import { MessageSender } from '../../../../agent/MessageSender'
 import { OutboundMessageContext, Protocol } from '../../../../agent/models'
 import { InjectionSymbols } from '../../../../constants'
 import { Attachment } from '../../../../decorators/attachment/Attachment'
-import { AriesFrameworkError } from '../../../../error'
 import { injectable } from '../../../../plugins'
 import { ConnectionService } from '../../../connections'
+import { verkeyToDidKey } from '../../../dids/helpers'
```
TimoGlastra, line 26: Not sure if this util already existed, but we should probably rename it to `publicKeyBase58ToDidKey`

packages/core/src/modules/message-pìckup/protocol/v2/V2MessagePickupProtocol.ts
```diff
@@ -82,21 +90,73 @@ export class V2MessagePickupProtocol extends BaseMessagePickupProtocol {
...
+      InjectionSymbols.MessagePickupRepository
+    )
+
+    // Get available messages from queue, but don't delete them
+    const messages = await messagePickupRepository.takeFromQueue({
+      connectionId: connectionRecord.id,
+      recipientKey: recipientKey,
+      limit: 10, // TODO: Define as config parameter
+      keepMessages: true,
+    })
```
TimoGlastra, line 110: Should we somehow add an option to mark them as delivered? but not remove them from the queue yet (so they won't be re-delivered directly, but they also aren't removed). Maybe something for future updates

packages/core/src/modules/message-pìckup/protocol/v2/V2MessagePickupProtocol.ts
```diff
@@ -82,21 +90,73 @@ export class V2MessagePickupProtocol extends BaseMessagePickupProtocol {
...
+        })
+    )
+
+    if (messages.length > 0) {
+      return {
+        message: new V2MessageDeliveryMessage({
+          attachments,
+        }),
+      }
+    }
```
TimoGlastra, line 128: Personal preference

```suggestion
    if (messages.length === 0) {
    	return { message: null }
    }
    
   
      return {
        message: new V2MessageDeliveryMessage({
          attachments,
        }),
      }
```

packages/core/src/modules/message-pìckup/protocol/v2/V2MessagePickupProtocol.ts
```diff
@@ -82,21 +90,73 @@ export class V2MessagePickupProtocol extends BaseMessagePickupProtocol {
     return { message }
   }
 
+  public async deliverMessages(
```
TimoGlastra, line 93: This should be called `createDeliveryMessage` or something I think. It does not deliver them

packages/core/src/modules/message-pìckup/MessagePickupEvents.ts
```diff
@@ -0,0 +1,22 @@
+import type { MessagePickupSession } from './MessagePickupSession'
+import type { QueuedMessage } from './storage'
```
TimoGlastra, line 2: unused

packages/core/src/modules/message-pìckup/MessagePickupApi.ts
```diff
@@ -66,13 +77,52 @@ export class MessagePickupApi<MPPs extends MessagePickupProtocol[] = [V1MessageP
...
+        await this.messageSender.sendMessage(
+          new OutboundMessageContext(deliverMessagesReturn.message, {
+            agentContext: this.agentContext,
+            connection: connectionRecord,
+          })
+        )
+      }
+    } else {
+      this.logger.debug('No live mode session')
+    }
```
TimoGlastra, line 125: nit

```suggestion
    if (activePickupSession) {
      this.logger.debug('No live mode session')
      return
    }
 const protocol = this.getProtocol(activePickupSession.protocolVersion)

  const deliverMessagesReturn = await protocol.deliverMessages(this.agentContext, {
    connectionRecord,
    recipientKey,
  })

  if (deliverMessagesReturn) {
    await this.messageSender.sendMessage(
      new OutboundMessageContext(deliverMessagesReturn.message, {
        agentContext: this.agentContext,
        connection: connectionRecord,
      })
    )
  }
```

packages/core/src/modules/message-pìckup/MessagePickupApi.ts
```diff
@@ -66,13 +77,52 @@ export class MessagePickupApi<MPPs extends MessagePickupProtocol[] = [V1MessageP
...
+  }
+
+  /**
+   * Deliver messages in the Message Pickup Queue for a given connection and key (if specified).
+   *
+   * Note that this is only available when there is an active session with the recipient. Message
+   * Pickup protocol messages themselves are not added to pickup queue
+   *
+   */
+  public async deliverQueuedMessages(options: DeliverQueuedMessagesOptions) {
```
TimoGlastra, line 97: Should this take a sessionId instead, so you wont have the case where there's no session? Maybe that'll complicate it without benefits

I think this should at least return an object, where it e.g. optionally returns the message, and a state which indicates how the delivery went (e.g. for now no open session, or delivered). Also will this deliver all messages? Or should you call this in a loop?

TimoGlastra (approved):
Very nice @genaris  :)

packages/core/src/agent/MessageSender.ts
```diff
@@ -48,15 +48,15 @@ export class MessageSender {
   public constructor(
     envelopeService: EnvelopeService,
     transportService: TransportService,
-    @inject(InjectionSymbols.MessageRepository) messageRepository: MessageRepository,
+    @inject(InjectionSymbols.MessagePickupRepository) messagePickupRepository: MessagePickupRepository,
```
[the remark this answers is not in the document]
genaris: At the beginning I named it `MessagePickupQueue` but discussing about it with my colleagues we noticed that we are not using it as a _queue_, in the sense that we are enqueuing data to it but not exactly dequeuing data when retrieving it (as we must store until it is actually acknowledged by the mediatee). Also, the data on it is somewhat organized (i.e. by connectionId, recipientKey, state, etc.) so the term 'Repository' feels to me more appropriate than 'Storage', which is usually used to generically save data.

Having said this, I'm not happy because the unfortunate name clash with `Repository` interface, so I'm more than open to opinions about how to improve the naming!

packages/core/src/agent/MessageSender.ts
```diff
@@ -176,7 +176,7 @@ export class MessageSender {
     // If the other party shared a queue service endpoint in their did doc we queue the message
     if (queueService) {
```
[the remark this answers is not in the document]
genaris: Ah yes, I remember this from https://github.com/hyperledger/aries-framework-javascript/issues/1199. I think it will be important to find a way to configure this. I'm trying to find the best place to update the code to support this, as I don't want to make `MessageSender` dependent on Mediator or Pickup modules API (we'll have some circular dependencies).

packages/core/src/agent/MessageSender.ts
```diff
@@ -327,7 +327,13 @@ export class MessageSender {
       }
 
       const encryptedMessage = await this.envelopeService.packMessage(agentContext, message, keys)
-      await this.messageRepository.add(connection.id, encryptedMessage)
+      for (const recipientKey of keys.recipientKeys) {
```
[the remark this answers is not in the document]
genaris: Yes, very good point. I'll use a recipientKey array in `add` method to avoid the need of calling it multiple times.

packages/core/src/constants.ts
```diff
@@ -1,5 +1,5 @@
 export const InjectionSymbols = {
-  MessageRepository: Symbol('MessageRepository'),
+  MessagePickupRepository: Symbol('MessagePickupRepository'),
```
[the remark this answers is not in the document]
genaris: I agree. However I think I'll need to update a bit the logic in order to avoid dependency cycles between `MessageSender` and `MessagePickupApi`. Probably adding some parameter to `sendMessage` or even triggering an internal event that can be listened in Message Pickup.

packages/core/src/agent/MessageSender.ts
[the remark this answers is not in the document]
ericvergnaud: I think `MessageRepository` is best i.e. it's _the place where you store and fetch messages_. I think `xxxQueue` would tend to incorrectly enforce the underlying technology, when any persistent storage may actually work. `Pickup` is just a fragment of the interface (we're not 'picking up' messages received that cannot be delivered, rather we're queuing them for later pickup), so I would just drop it to avoid narrowing the exhibited responsibility.

ericvergnaud:
Hi, as per the live-mode, I think the proposed approach will fix the problems we observed with websockets, where messages were irremediably lost, being enqueued without being acknowledged. So thanks for that, looking forward to it!

ericvergnaud:
I'm less convinced that the "central Message Pickup Repository" will address https://github.com/openwallet-foundation/agent-framework-javascript/issues/1625.

The "central" wording suggests that this repo will run in a separate instance. 
**If that's the case**, then what happens when that instance is down ? Am I correct in suspecting that having 2 pickup instances will exhibit the exact same problem hence we would have just pushed the problem to a separate component ?
**If that's _not_ the case**, then the problem is that each mediator instance will be required to know every other mediator instance, which is not viable because in an auto-scaling scenario, the number of mediator instances may vary dynamically from 1 to 3 or even 100. 
I may misunderstand what is being proposed, but I suspect it's trying to solve the problem completely internally, which I'm not sure can address disaster scenarios. 

IMHO, the only robust solution is to rely on a notification mechanism that exists _outside_ the mediator itself. Then each mediator instance subscribes to that notification platform (for example AWS EventBridge), and the responsibility of sending notifications lies entirely with the storage layer for queued messages. 

In short, the API should simply make it possible to configure a notification listener that will invoke the appropriate AFJ message pickup API. Also worth noting, sending those notifications should only happen _after_ storage, not before, and should also not be a responsibility of AFJ because that's easily achievable using standard middleware  (Redis or a CDC can do that).

In even shorter, AFJ should handle notifications, not provide an internalised notification platform. All I need is an entry point that I can call from my app to tell it that a message for recipient x has been stored. That entry point will check whether the corresponding mediatee is connected and if so forward the message to them (as described in the original PR comment)(if not it does nothing).

Sorry if that sounds a bit negative, I might simply be misunderstanding what is being proposed. Happy to help with testing.

TimoGlastra:
Hey @ericvergnaud,

I think you're correct this PR doesn't directly address the issue of multiple mediator instances, but it does add methods to build w plugin on top of AFJ that does this.


There's new events added that allow you to know when a message has been queued. The idea is that a) you hook into this event and emit some kind of notification that will be received by other mediators (we use redis pub/sub) and b) in the other mediator you call deliver messages method to deliver the messages to the recipient when a notification is received.

A potential next step would be to make some default implementations plugins for notification, so it's easier to set up a mediator. 

Does that resolve your uncertainties with this PR, or are there things we should do differently?

ericvergnaud:
> Hey @ericvergnaud,
> 
> I think you're correct this PR doesn't directly address the issue of multiple mediator instances, but it does add methods to build w plugin on top of AFJ that does this.
> 
> There's new events added that allow you to know when a message has been queued. The idea is that a) you hook into this event and emit some kind of notification that will be received by other mediators (we use redis pub/sub) and b) in the other mediator you call deliver messages method to deliver the messages to the recipient when a notification is received.
> 
> A potential next step would be to make some default implementations plugins for notification, so it's easier to set up a mediator.
> 
> Does that resolve your uncertainties with this PR, or are there things we should do differently?

@TimoGlastra thanks for the clarification. I guess the current wording is slightly misleading then...

One thing I remain concerned with is putting the responsibility of sending notifications with the mediator. Although it seems convenient, it won't be 100% reliable unless the mediator implements 2-phase commit:
 - If the event is broadcasted _before_ the message is stored, then it could be processed before the message is available from the store (not speaking of scenarios where it is not stored at all due to some failure)
 - if the event is broadcasted _after_ the message is stored, then anything can happen between the moment is is stored and the moment it gets broadcasted, including a crash of the mediator, in which case no notification will happen.
I don't think the mediator should implement 2-phase commit. Rather I recommend delegating that task to the storage itself. PostgreSQL comes with a CDC that is guaranteed to run for every data change, even after a crash, because it relies on the journal. Most database engines provide the same, including [sqlite](https://www.sqlite.org/c3ref/update_hook.html). 

That said, the proposed mechanism can be very useful for testing, so I would keep it. But not advertise it as the standard way of syncing instances in a mediator cluster, rather provide examples of a correct end-to-end implementation.

TimoGlastra:
That makes sense, thank you for the explanation.

I think CDC is a great solution to this. Is there anything we can do in AFJ to make this more easy to set up?

CC @swcurran @dbluhm I think ACA-Py kafka / redis plugin would be susceptible to the same problem that @ericvergnaud describes here.

ericvergnaud:
> That makes sense, thank you for the explanation.
> 
> I think CDC is a great solution to this. Is there anything we can do in AFJ to make this more easy to set up?
> 
> CC @swcurran @dbluhm I think ACA-Py kafka / redis plugin would be susceptible to the same problem that @ericvergnaud describes here.

I guess a useful thing that can be done:
 - provide a sample plugin that listens to Redis pub/sub
 - provide a sample pgsql cdc that writes to Redis pub/sub
 (since sqlite cannot act as a server, not sure it makes sense to provide a cdc sample sample for it)
Adapting these for Kafka and AWS EventBridge would also be useful (we'll be using EventBridge so I can help with that).

Another useful thing is updating the documentation for describing a 'supported' cluster setup.

For NR testing:
 - provide a sample http plugin that listens to http notifications
 - provide a sample event listener that invokes the above http endpoint
 - write a test that:
    - runs 2 mediators with the above plugins and a _mediatee_ listening to instance A
    - sends a forward message M to instance B
    - ensures that the _mediatee_ receives message M

packages/core/src/agent/MessageSender.ts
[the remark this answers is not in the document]
genaris: The problem with `MessageRepository` is that it collides with `Repository` classes used within the framework, most notably `DidCommMessageRepository`, so leaving such generic name would be more confusing. The 'Pickup' word tries to specify that it is used for message pickup protocol.

It seems it is still not clear, as this `MessagePickupRepository` is an interface rather than an implementation, like the other `Repository`.

genaris:
> The "central" wording suggests that this repo will run in a separate instance. **If that's the case**, then what happens when that instance is down ? Am I correct in suspecting that having 2 pickup instances will exhibit the exact same problem hence we would have just pushed the problem to a separate component ? **If that's _not_ the case**, then the problem is that each mediator instance will be required to know every other mediator instance, which is not viable because in an auto-scaling scenario, the number of mediator instances may vary dynamically from 1 to 3 or even 100. I may misunderstand what is being proposed, but I suspect it's trying to solve the problem completely internally, which I'm not sure can address disaster scenarios.

In this PR description, when I referred to a a'central message pickup repository' I was meaning a common storage shared among all AFJ instances (e.g. a PostgreSQL or Mongo database). That does not mean that it will reside in one of them.

Through `MessagePickupRepository` interface, each mediator instance can enqueue and retrieve messages from it, just like it does right now.

What is different from previous implementation is that now the framework gives a way for each mediator instance to push messages for a certain mediatee on-demand, so if it receives a notification that a new message has been added for it, it can send it immediately (by calling `agent.messagePickup.deliverQueuedMessages()`).

Also, there are new events triggered when Live Mode Sessions are opened and closed. This allows each mediator instance to keep track of all mediatees are connected to it, so it can e.g. subscribe to receive notifications only when there are new messages for them instead of listening to every change in message pickup database.

Whether you want to use a CDC to notify about changes, or create a pub/sub mechanism where each AFJ instance publish a message to a channel and other instances listen to it, it is up to each implementation. It is not something enforced by thir PR.

> IMHO, the only robust solution is to rely on a notification mechanism that exists _outside_ the mediator itself. Then each mediator instance subscribes to that notification platform (for example AWS EventBridge), and the responsibility of sending notifications lies entirely with the storage layer for queued messages.
> 
> In short, the API should simply make it possible to configure a notification listener that will invoke the appropriate AFJ message pickup API. Also worth noting, sending those notifications should only happen _after_ storage, not before, and should also not be a responsibility of AFJ because that's easily achievable using standard middleware (Redis or a CDC can do that).
> 

I agree on this. I hope to have been clear that the goal of this PR is to make it possible to build an architecture like the one you are describing here.

genaris:
> Hey @ericvergnaud,
> 
> I think you're correct this PR doesn't directly address the issue of multiple mediator instances, but it does add methods to build w plugin on top of AFJ that does this.
> 
> There's new events added that allow you to know when a message has been queued. The idea is that a) you hook into this event and emit some kind of notification that will be received by other mediators (we use redis pub/sub) and b) in the other mediator you call deliver messages method to deliver the messages to the recipient when a notification is received.
> 
> A potential next step would be to make some default implementations plugins for notification, so it's easier to set up a mediator.
> 
> Does that resolve your uncertainties with this PR, or are there things we should do differently?

Something I want to clarify is that the new events are more related to session opening/closing than individual message queuing. You can of course hook into either `AgentMessageSent` event or `MessagePIckupRepository.addMessage()` to implement a pub/sub mechanism as @TimoGlastra is describing, but this is not a practice being enforced by this PR.

I agree that this should be clearly documented, and add some reference plug-ins to show a full implementation. 

I'll do some more changes to this PR based on these comments and some experience I had recently while testing it, and submit it for review.

ericvergnaud:
@genaris I've provided more samples, as a PR to your branch, see https://github.com/genaris/aries-framework-javascript/pull/2

packages/core/src/modules/message-pìckup/MessagePickupApi.ts
genaris: In order to make it more flexible, I've added an API method to get live sessions and make this one to use sessionId instead. This way, the only possibility is that it will proceed with the delivery (e.g. the delivery message has been properly submitted) or an error happened (MessageSendingError or so), so it will throw and can be handled by the caller.

This method, now called `deliverMessagesFromQueue`, will retrieve any pending messages from the queue and send the initial batch of `batchSize` messages, starting a pickup loop that will be finished once all messages are delivered (the client will automatically request more messages when the status shows that there are pending messages). This initial `batchSize` number is sent as an input parameter or taken from configuration.

There is also another method called `deliverMessages` which is similar but intended to send specific messages. It will not retrieve other messages from the queue. This can be more appropriate when several messages for a given mediatee are received within a short time frame, as there will not be need of querying the queue multiple times.

packages/core/src/agent/MessageSender.ts
```diff
@@ -327,7 +333,11 @@ export class MessageSender {
       }
 
       const encryptedMessage = await this.envelopeService.packMessage(agentContext, message, keys)
-      await this.messageRepository.add(connection.id, encryptedMessage)
+      await this.messagePickupRepository.addMessage({
+        connectionId: connection.id,
+        recipientKeys: keys.recipientKeys.map((item) => verkeyToDidKey(item.publicKeyBase58)),
```
TimoGlastra, line 338: I'd prefer:

```suggestion
        recipientKeys: keys.recipientKeys.map((item) => new DidKey(item).did),
```

However, is there a need to transform the key from a key instance to a `did:key` instance? The longer we can keep it a `Key` instance the better TS can help us and it's easier to understand. Because string can be `did:key:xxx` or just a plain base58 encoded public key.

So my preference would be to pass an instance of `Key` to the `messagePickupRepository`, or rename `recipientKeys` to something more unique so it's clear how this `recipientKey` is encoded

packages/core/src/modules/message-pickup/MessagePickupApi.ts
```diff
@@ -66,13 +85,100 @@ export class MessagePickupApi<MPPs extends MessagePickupProtocol[] = [V1MessageP
...
+    this.logger.debug('Deliverying queued messages')
+
+    const { pickupSessionId, recipientKey, batchSize } = options
+
+    const session = this.messagePickupSessionService.getLiveSession(this.agentContext, pickupSessionId)
+
+    if (!session) {
+      this.logger.debug(`No active live mode session found with id ${pickupSessionId}`)
+      return
+    }
```
TimoGlastra, line 163: This seems like an error case to me?

packages/core/src/modules/message-pickup/protocol/MessagePickupProtocolOptions.ts
```diff
@@ -1,12 +1,33 @@
...
 export interface PickupMessagesProtocolOptions {
   connectionRecord: ConnectionRecord
   recipientKey?: string
   batchSize?: number
 }
 
+export interface DeliverMessagesProtocolOptions {
+  connectionRecord: ConnectionRecord
+  messages?: QueuedMessage[]
+  recipientKey?: string
```
TimoGlastra, line 14: I think here the same, even though the RFC describes what should be used as the value for `recipientKey`, I'd prefer to make the variable more explicit, or make it a class instance and transform it last minute when adding it to the message

packages/core/src/modules/message-pickup/protocol/MessagePickupProtocolOptions.ts
```diff
@@ -1,12 +1,33 @@
 import type { AgentMessage } from '../../../agent/AgentMessage'
 import type { ConnectionRecord } from '../../connections'
+import type { QueuedMessage } from '../storage'
 
 export interface PickupMessagesProtocolOptions {
   connectionRecord: ConnectionRecord
   recipientKey?: string
```
TimoGlastra, line 7: Also here

packages/core/src/modules/message-pickup/protocol/v1/V1MessagePickupProtocol.ts
```diff
@@ -55,24 +59,69 @@ export class V1MessagePickupProtocol extends BaseMessagePickupProtocol {
...
+      const message = new V1BatchMessage({
+        messages: batchMessages,
+      })
+
+      return { message }
+    }
+  }
+
+  public async setLiveDeliveryMode(
+    agentContext: AgentContext,
```
TimoGlastra, line 98: Eslint warnings. I think you can remove the parameters? Or eslint ignore

packages/core/src/modules/routing/MediatorModuleConfig.ts
```diff
@@ -9,6 +11,19 @@ export interface MediatorModuleConfigOptions {
...
+  /**
+   * Strategy to use when a Forward message is received.
+   *
+   *
+   * - `MessageForwardingStrategy.QueueOnly` - simply queue encrypted message into MessagePickupRepository. It will be in charge of manually trigering MessagePickupApi.deliver() afterwards.
+   * - `MessageForwardingStrategy.QueueAndLiveModeDelivery` - Queue message into MessagePickupRepository and deliver it (along any other queued message).
+   * - `MessageForwardingStrategy.DirectDelivery` - Deliver message directly. Do not add into queue (it might be manually added after, e.g. in case of failure)
+   *
+   * @default MessageForwardingStrategy.DirectDelivery
+   * @todo Update default to QueueAndLiveModeDelivery
```
TimoGlastra, line 24: When do we want to change the defualt?

Isn't this something that can be determined automatically? E.g. first try direct delivery (in case the mediatee has an endpoint), if it fails, try live delivery (if an open session), if it fails add to queue.


I think you'd often want a combination of 2 and 3 right? Or is direct delivery only direct **live** delivery (I'm assuming it is when you can send it directly to an endpoint, but maybe that's wrong. If so, maybe we should call it `DirectLiveDelivery`)

TimoGlastra (approved):
Nice work @genaris! I have some nits, but those can also be addressed later.

packages/core/src/modules/routing/MediatorModuleConfig.ts
genaris: This is more about selecting if we want to encapsulate forwarded messages in Message Pickup's `Delivery` message (so we can have explicit acknowledge) or we want to simply send them to either the mediatee endpoint or an existing session with them. Do you have a better wording to reflect what this setting does?

I think this way of sending forwarded messages should be deprecated at some point, but the encapsulated one should be used only in case the other party supports Message Pickup V2, and that's something that probably can be determined automatically if we do some smarter logic on the agent (e.g. we can query the other party for protocol support or mark it as supported as soon as we receive the first Pickup V2 protocol loop from them).

PR #1639, opened 2023-11-15
OpenId4Vc Support

2mau:


dbluhm:
We're eager to use these changes to test against an ACA-Py plugin we're building for OpenID4VCI.

TimoGlastra:
> We're eager to use these changes to test against an ACA-Py plugin we're building for OpenID4VCI.

Cool @dbluhm! Which format profiles are you supporting? We're focussing on JWT / SD-JWT for now

dbluhm:
We're definitely taking an iterative approach but we're focusing on JWT first. SD-JWTs will likely come next. We've been testing against the sphereon wallet and I believe this is using the Sphereon SDK so hopefully that bodes well for compatibility! 

Are you basing the SD-JWT format off of the HAIP? That's the only place I've seen it defined so far but there are some restrictions in that profile we're not thrilled about.

TimoGlastra:
> Are you basing the SD-JWT format off of the HAIP? That's the only place I've seen it defined so far but there are some restrictions in that profile we're not thrilled about.

Yes we're basing it off haip

Which restrictions are you not so thrilled about? I think our main concern was did support, but the spec recommends to use `jwk`, but you can use other ways to bind the credential to e.g. a did

PR #1647, opened 2023-11-23
feat!: allow to query tenant records using label

TimoGlastra:
Requested by someone in the discord


packages/core/src/storage/migration/updates.ts
```diff
@@ -46,6 +46,16 @@ export const supportedUpdates = [
     toVersion: '0.4',
     doUpdate: updateV0_3_1ToV0_4,
   },
+  // FIXME: this entry needs to be here to allow extension modules to add their own updates
+  // We should allow extension modules to add their own updates without having to add an entry here
+  // There will be other migrations in 0.5, so it's okay for this version upgrade
+  {
```
genaris, line 52: I think we'll need to update migration tests and/or snapshots (CI is failing)

genaris (approved):

TimoGlastra:
Actually discovered quite a serious bug with how tags were parsed with Askar.

Created a fix for it for `expandedTypes` as that's the only non-computed tag that I can think of that may include URLs. But this can be any record (also custom records) that use array tag values that contain a `:`.

The bug:
```
/**
 * Up until 0.5.0 the AskarStorageService contained a bug where a non-computed (so manually set on record) array tag values that contained a : in the value
 * would be incorrectly parsed back from an askar tag to a tag on a record. This would only cause problems for the storage if the record was re-saved and not
 * computed. The following would happen:
 * - Create record with non-computed tag, e.g. expandedTypes that contains a value with a : in it
 * - Save record. The tag is correctly set in Askar as `expandedTypes:https://example.com/'
 * - Read record. The tag is correctly read from Askar as `expandedTypes:https://example.com/'. However the transformation would result in the tag value on the record being set to `https'.
 * - Save record. The non-computed (important, as otherwise the correct computed value would overwrite the incorrect value before storing) tag value is now set to `https' instead of `https://example.com/'
 *
 * This function checks if any of the values for expandedTypes is `https` and if so, re-calculates the correct value and sets it on the record.
 *
 * NOTE: This function needs to resolve the context of a W3cCredentialRecord to be able to correctly calculate the expanded types.
 * To not brick a wallet that has no internet when updating, the storage update will allow the resolving of the expanded types to fail.
 * If this is the case, at a later point the expanded types will need to be recalculated and set on the record.
 *
 * If w3c credential records are never re-saved this shouldn't be a problem though. By default w3c credential records are not re-saved,
 * and so it only applies if you have implemented a custom flow that re-saves w3c credential records (e.g. if you add metadata).
 */
```

genaris:
> Actually discovered quite a serious bug with how tags were parsed with Askar.
> 
> Created a fix for it for `expandedTypes` as that's the only non-computed tag that I can think of that may include URLs. But this can be any record (also custom records) that use array tag values that contain a `:`.
> 

So this is just a workaround for W3C  stuff, or the root bug in Askar was already solved somewhere else? I think I didn't find any relevant change recently in `AskarStorageService` in main.

TimoGlastra:
It's here @genaris https://github.com/openwallet-foundation/credo-ts/pull/1647/files#diff-9bb6478f2f3748c0c5a22d1702d1835be0bc83a33869cb4314204d1b6c410f54

The issue is resolved in main, but it needs to be addressed for all records that use tags. 

However because it only occurs with non-computed array tags containing a `:`, the actual impact is quite small I think, and I'm not sure if other migrations are needed besides the `expandedTypes`

PR #1648, opened 2023-11-27
refactor(anoncreds)!: move supportRevocation to options

TimoGlastra:
This moves the `supportRevocation` option in the registerCredentialDefinition method to the `options` object. It feels a bit weird to me that we have an `options` object, but then not place an option like `supportRevocation` in that options object

@genaris you ok with this change?


TimoGlastra:
If we don't want to add it to the `options` object, I think we should rename `options` to `registryOptions` or something, to indicate it's only for the registry, but I think we can use it as a general purpose options object?

genaris (approved):
I think both are good improvements, especially the return type for `TailsFileService` interface methods (given the confusion we usually have regarding local paths vs URLs).

PR #1650, opened 2023-11-27
refactor: move message-pickup directory

genaris:
For some strange reason, the 'i' in `message-pickup` module directory is an special character, which makes it difficult to create patches to properly test updates in that module (like the ones in #1638), so this small PR is just fixing it using a regular 'i'.


TimoGlastra (approved):
that's funny :)

PR #1667, opened 2023-12-08
feat: openid4vc support

2mau:

PR #1672, opened 2023-12-13
feat(presentation-exchange): added PresentationExchangeService

berendsliedrecht:
- Copied the `PresentationExchangeService` from #1667 to `core` as it will also be needed by the [Present Proof Protocol V2.0](https://github.com/hyperledger/aries-rfcs/blob/main/features/0454-present-proof-v2/README.md)
- In the end this should not live in core but in a ?pex? package. When DIDComm will be moved out of core, this can also be picked up.


packages/core/src/modules/proofs/index.ts
```diff
@@ -12,3 +12,6 @@ export * from './ProofsApiOptions'
 // Module
 export * from './ProofsModule'
 export * from './ProofsModuleConfig'
+
+// Services
+export * from './services'
```
TimoGlastra, line 17: I was thinking we create a separate module for this already in AFJ, as the current proofs module is mainly focused on didcomm, and I think it's good to keep that separated from functionality that is not related to didcomm.

berendsliedrecht: Yeah, happy to do it as a module already.

berendsliedrecht: Would that just be a module without a public API then?

berendsliedrecht: It is a new module now that just contains the service, no public API.

packages/core/src/modules/presentation-exchange/PresentationExchangeModule.ts
```diff
@@ -0,0 +1,16 @@
+import type { DependencyManager, Module } from '../../plugins'
+
+import { PresentationExchangeService } from './PresentationExchangeService'
+
+/**
+ * @public
+ */
+export class PresentationExchangeModule implements Module {
```
TimoGlastra, line 8: maybe we can add the experimental warning to this module as well? As we'll need to make several changes to this over the coming weeks

packages/core/src/modules/presentation-exchange/PresentationExchangeService.ts
```diff
@@ -0,0 +1,477 @@
...
+        for (const schema of inputDescriptor.schema) {
+          query.push({
+            $or: [{ expandedType: [schema.uri] }, { contexts: [schema.uri] }, { type: [schema.uri] }],
+          })
+        }
+      }
+    } else if (presentationDefinitionVersion.version === PEVersion.v2) {
+      // FIXME: As PE version 2 does not have the `schema` anymore, we can't query by schema anymore.
+      // For now we retrieve ALL credentials, as we did the same for V1 with JWT credentials. We probably need
+      // to find some way to do initial filtering, hopefully if there's a filter on the `type` field or something.
```
TimoGlastra, line 94: We probably need to look at this in the near future. This will cause issues for sure

packages/presentation-exchange/src/index.ts
```diff
@@ -0,0 +1,4 @@
+export * from './PresentationExchangeError'
```
TimoGlastra, line 1: there now seems to be logic in core, as well as in a separate presentation exchange package

TimoGlastra (approved):

PR #1676, opened 2023-12-19
feat(present-proof): add support for aries RFC 510

berendsliedrecht:
- Adds basic support for dif presentation exchange over DIDComm
    - According to RFC 0454 and 510

- Would be good to include some more tests, but code is ready to review
- There is still an open issue I have with the credential selection. Right now if we use PEXv2, we basically call `getAll` on the w3c credentials which is not really great... For V1 we could use the schema attribute, but that is not included anymore. For now, it might be fine to just call the `getAll` function and use those credentials and resolve this later when we want to revamp the storage.


packages/core/src/modules/presentation-exchange/PresentationExchangeService.ts
```diff
@@ -57,6 +60,50 @@ export class PresentationExchangeService {
     return selectCredentialsForRequest(presentationDefinition, credentialRecords, holderDids)
   }
 
+  public validatePresentationDefinition(presentationDefinition: PresentationDefinition) {
+    const validation = PEX.validateDefinition(presentationDefinition)
+    const errorMessages = this.formatValidated(validation)
+    if (errorMessages.length > 0) {
+      throw new PresentationExchangeError(
```
TimoGlastra, line 67: Would it be possible to add these error messages to the error class as a custom property? Then they could be extracted on a higher level for more control

packages/core/src/modules/presentation-exchange/PresentationExchangeService.ts
```diff
@@ -416,13 +461,14 @@ export class PresentationExchangeService {
...
-      const presentationToSign = { ...presentationJson, presentation_submission: undefined }
+      const presentationToSign = { ...presentationJson }
+      delete presentationToSign['presentation_submission']
 
       let signedPresentation: W3cVerifiablePresentation<ClaimFormat.JwtVp | ClaimFormat.LdpVp>
       if (vpFormat === 'jwt_vp') {
         signedPresentation = await w3cCredentialService.signPresentation(agentContext, {
           format: ClaimFormat.JwtVp,
-          alg: this.getSigningAlgorithmForJwtVc(presentationDefinition, verificationMethod),
+          alg: this.getSigningAlgorithmForJwtVc(presentationDefinition as PresentationDefinition, verificationMethod),
```
TimoGlastra, line 471: Is presentationDefinition possibly undefined? Can we do a check in that case? Then we don't have to use type casting

packages/core/src/modules/presentation-exchange/PresentationExchangeService.ts
```diff
@@ -431,7 +477,11 @@ export class PresentationExchangeService {
       } else if (vpFormat === 'ldp_vp') {
         signedPresentation = await w3cCredentialService.signPresentation(agentContext, {
           format: ClaimFormat.LdpVp,
-          proofType: this.getProofTypeForLdpVc(agentContext, presentationDefinition, verificationMethod),
+          proofType: this.getProofTypeForLdpVc(
+            agentContext,
+            presentationDefinition as PresentationDefinition,
```
TimoGlastra, line 482: same here

packages/core/src/modules/presentation-exchange/utils/credentialSelection.ts
```diff
@@ -36,8 +37,12 @@ export async function selectCredentialsForRequest(
     ...selectResultsRaw,
     // Map the encoded credential to their respective w3c credential record
     verifiableCredential: selectResultsRaw.verifiableCredential?.map((encoded) => {
-      const credentialIndex = encodedCredentials.indexOf(encoded)
-      const credentialRecord = credentialRecords[credentialIndex]
+      const credentialRecord = credentialRecords.find((record) => {
+        const originalVc = getSphereonOriginalVerifiableCredential(record.credential)
+
+        return deepEquality(originalVc, encoded)
+      })
```
TimoGlastra, line 44: It feels a bit expensive to do a deepEquality on every credential to find the original record that was associated with VC. Is there no other way to do the matching?

Because we have the encodedCredenitals, we can do a `===` check on it right, as the object references are the same?

So in that case `const credentialIndex = encodedCredentials.indexOf(encoded)` should still work

But maybe I'm missing something

packages/core/src/modules/proofs/formats/presentation-exchange/PresentationExchangeProofFormat.ts
```diff
@@ -0,0 +1,40 @@
...
+    selectCredentialsForRequest: {
+      input: never
+      output: Array<W3cCredentialRecord>
+    }
+  }
+
+  formatData: {
+    proposal: unknown
+    request: unknown
+    presentation: unknown
```
TimoGlastra, line 38: This should not be unknown

packages/core/src/modules/proofs/formats/presentation-exchange/PresentationExchangeProofFormat.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    getCredentialsForRequest: {
+      input: never
+      output: Array<W3cCredentialRecord>
+    }
+
+    selectCredentialsForRequest: {
+      input: never
+      output: Array<W3cCredentialRecord>
+    }
```
TimoGlastra, line 32: Can we update these interfaces to align with the more advanced structure we have implemented in the openid4vp and pex services? We have quite an advanced structure to make it easier to detect which groups of credentials you can use to satisfy a request. Especially `getCredentialsForRequest` should not just return a w3c record array.

In addition, I think it can help to return an object with `credentials` property even for select method so if we need to add a property in the future it won't be a breaking change

packages/core/src/modules/proofs/formats/presentation-exchange/PresentationExchangeProofFormatService.ts
```diff
@@ -0,0 +1,309 @@
...
+    const ps = this.presentationExchangeService(agentContext)
+
+    const pexFormat = proofFormats.presentationExchange
+    if (!pexFormat) {
+      throw new AriesFrameworkError('Missing Presentation Exchange format in create proposal attachment format')
+    }
+
+    const { presentationDefinition } = pexFormat
+
+    ps?.validatePresentationDefinition(presentationDefinition)
```
TimoGlastra, line 70: ps is always defined right (it should be as otherwise it could lead to unexpected behaviour cause validate is not called in that case)?

```suggestion
    ps.validatePresentationDefinition(presentationDefinition)
```

packages/core/src/modules/proofs/formats/presentation-exchange/PresentationExchangeProofFormatService.ts
```diff
@@ -0,0 +1,309 @@
...
+    const { presentationDefinition } = presentationExchangeFormat
+
+    ps.validatePresentationDefinition(presentationDefinition)
+
+    const format = new ProofFormatSpec({
+      format: PRESENTATION_EXCHANGE_PRESENTATION_REQUEST,
+      attachmentId,
+    })
+
+    const options = { challenge: 'TODO' }
```
TimoGlastra, line 126: TODO

packages/core/src/modules/proofs/formats/presentation-exchange/PresentationExchangeProofFormatService.ts
```diff
@@ -0,0 +1,309 @@
...
+
+    const attachment = this.getFormatData(
+      { options, presentation_definition: presentationDefinition },
```
TimoGlastra, line 129: It might be nice to construct this object according to an interface, so it's harder to accidentally create an incorrect message. 

```suggestion
      { options, presentation_definition: presentationDefinition } satisfies XXX,
```

packages/core/src/modules/proofs/formats/presentation-exchange/PresentationExchangeProofFormatService.ts
```diff
@@ -0,0 +1,309 @@
...
+      format.attachmentId
+    )
+
+    return { attachment, format }
+  }
+
+  public async processRequest(agentContext: AgentContext, { attachment }: ProofFormatProcessOptions): Promise<void> {
+    const ps = this.presentationExchangeService(agentContext)
+    const { presentation_definition: presentationDefinition } = attachment.getDataAsJson<{
```
TimoGlastra, line 138: Same here, we should define an interface for all the attachments, and then use the interface everywhere. We need these interfaces anyway for the current `unknown` types in the format interface for proposal/request/presentation.

packages/core/src/modules/proofs/formats/presentation-exchange/PresentationExchangeProofFormatService.ts
```diff
@@ -0,0 +1,309 @@
...
+    const ps = this.presentationExchangeService(agentContext)
+
+    const format = new ProofFormatSpec({
+      format: PRESENTATION_EXCHANGE_PRESENTATION,
+      attachmentId,
+    })
+
+    const { presentation_definition: presentationDefinition, options } = requestAttachment.getDataAsJson<{
+      presentation_definition: PresentationDefinition
+      options?: { challenge?: string; domain?: string }
```
TimoGlastra, line 157: same here (won't comment on it again)

packages/core/src/modules/proofs/formats/presentation-exchange/PresentationExchangeProofFormatService.ts
```diff
@@ -0,0 +1,309 @@
...
+    const { presentation_definition: presentationDefinition } = attachment.getDataAsJson<{
+      presentation_definition: PresentationDefinition
+    }>()
+    ps.validatePresentationDefinition(presentationDefinition)
+  }
+
+  public async acceptRequest(
+    agentContext: AgentContext,
+    { attachmentId, requestAttachment }: ProofFormatAcceptRequestOptions<PresentationExchangeProofFormat>
+  ): Promise<ProofFormatCreateReturn> {
```
TimoGlastra, line 147: acceptRequest should allow user to pass which credentials they want to use for the presentation.

TimoGlastra: We have a submission type, that would be a good input to the acceptRequest method (optional, otherwise we auto-select)

So:
- getCredentialForRequest -> returns advanced objects with all the credentials you can use for different input descriptors
- selectCredentialsForRequest -> returend submission like object that maps input descriptors to which credential(s) to use to satisfy it
- acceptRequest -> optoinally take the same structure as returned by selectCredentialsForRequest, but user can construct it themselves based on return type from getCredentialsForRequest. If not passed, call select method within the acceptRequest method

packages/core/src/modules/proofs/formats/presentation-exchange/PresentationExchangeProofFormatService.ts
```diff
@@ -0,0 +1,309 @@
...
+    }>()
+    const presentation = attachment.getDataAsJson<
+      W3cVerifiablePresentation & { presentation_submission: PresentationSubmission }
+    >()
+
+    try {
+      ps.validatePresentationDefinition(presentationDefinition)
+      if (presentation.presentation_submission) {
+        ps.validatePresentationSubmission(presentation.presentation_submission)
+      }
```
TimoGlastra, line 214: I think a presentation_submission is required right, and we need to validate validate it?

packages/core/src/modules/proofs/formats/presentation-exchange/PresentationExchangeProofFormatService.ts
```diff
@@ -0,0 +1,309 @@
...
+
+      ps.validatePresentation(presentationDefinition, presentation as unknown as VerifiablePresentation)
```
TimoGlastra, line 216: Is the presentation submission not used to verify a presentation? Does it just check if it matches the presentationDefinition, but not the presentationSubmission?

Also -- should we make `validatePresentation` take a W3CVerifiablePresentation instead of a sphereon interface?

packages/core/src/modules/proofs/formats/presentation-exchange/PresentationExchangeProofFormatService.ts
```diff
@@ -0,0 +1,309 @@
...
+  public async processPresentation(
+    agentContext: AgentContext,
+    { requestAttachment, attachment }: ProofFormatProcessPresentationOptions
+  ): Promise<boolean> {
+    const ps = this.presentationExchangeService(agentContext)
+    const { presentation_definition: presentationDefinition } = requestAttachment.getDataAsJson<{
+      presentation_definition: PresentationDefinition
+    }>()
+    const presentation = attachment.getDataAsJson<
+      W3cVerifiablePresentation & { presentation_submission: PresentationSubmission }
```
TimoGlastra, line 207: W3cVerifiablePresentation is incorrect I think, as it's a class, while we just take the object here. So the type will think it has `context` property, but it will actually have `@context` property. We need to use json transformer on it as well.

packages/core/src/modules/proofs/formats/presentation-exchange/PresentationExchangeProofFormatService.ts
```diff
@@ -0,0 +1,309 @@
...
+    try {
+      ps.validatePresentationDefinition(presentationDefinition)
+      if (presentation.presentation_submission) {
+        ps.validatePresentationSubmission(presentation.presentation_submission)
+      }
+
+      ps.validatePresentation(presentationDefinition, presentation as unknown as VerifiablePresentation)
+      return true
+    } catch (e) {
+      agentContext.config.logger.error(e)
```
TimoGlastra, line 219:
```suggestion
      agentContext.config.logger.error(`Failed to verify presentation in PEX proof format service: ${e.message}`, { cause: e })
```

packages/core/src/modules/proofs/formats/presentation-exchange/PresentationExchangeProofFormatService.ts
```diff
@@ -0,0 +1,309 @@
...
+      presentation_definition: PresentationDefinition
+    }>()
+
+    ps.validatePresentationDefinition(presentationDefinition)
+
+    const presentationSubmission = await ps.selectCredentialsForRequest(agentContext, presentationDefinition)
+
+    const credentials = presentationSubmission.requirements.flatMap((r) =>
+      r.submissionEntry.flatMap((e) => e.verifiableCredentials)
+    )
```
TimoGlastra, line 259: As mentioned in other comments, the structures returned by the service adds a lot of needed context. I think we should return them as is, rather than flatmapping

packages/core/tests/jsonld.ts
```diff
@@ -38,13 +41,15 @@ export const getJsonLdModules = ({
...
       autoAcceptProofs,
+      proofProtocols: [new V2ProofProtocol({ proofFormats: [new PresentationExchangeProofFormatService()] })],
     }),
     cache: new CacheModule({
       cache: new InMemoryLruCache({ limit: 100 }),
     }),
     indySdk: new IndySdkModule({
       indySdk,
     }),
+    pex: new PresentationExchangeModule(),
```
TimoGlastra, line 52: Is this registered by default? As it's in core, I think we should register it by default

TimoGlastra (changes requested):
Overall looks good, but it's missing some things

packages/core/src/modules/presentation-exchange/PresentationExchangeService.ts
berendsliedrecht: Yes, good idea

berendsliedrecht:
@TimoGlastra I resolved your feedback, but I am now encountering an error that the `PresentationExchangeProofFormatData.formatData.presentation` does not contain the `proof` object. Do you know if we have this already defined somewhere as a type that I could reuse? I could not find it myself.

TimoGlastra:
> @TimoGlastra I resolved your feedback, but I am now encountering an error that the `PresentationExchangeProofFormatData.formatData.presentation` does not contain the `proof` object. Do you know if we have this already defined somewhere as a type that I could reuse? I could not find it myself.

I was able to fix this by using `presentation.encoded`. This return either the JSON variant of the w3c presentation or the serialized JWT as string (in case of JWT VP). You were directly using the class, which is not suitable for directly sending over the wire (it first needs to be serialized to JSON).

TimoGlastra:
@berendsliedrecht I have pushed a fix for the above + several other fixes (please take a look at the commit so you can see which changes I made).

Could you maybe to wrap this up, update the tests in `packages/core/src/modules/proofs/protocol/v2/__tests__/v2-presentation-exchange-presentation.e2e.test.ts` to make it either all standalone tests or a single test?

Currently it's multiple tests but they depend on each other. It will result in that you can't run the tests separately, and if one fails the others will fail too. So in that case it's better to make it just one test or properly split them up.

Otherwise I think we're good here, really nice work and happy that it's finally ready to be integrated into AFJ :)

Once merged I'll update it in the openid4vc branch

BTW -- added a non-spec compliant way to also support JWT VPs. Haven't added tests for it yet, but I think it is really nice to also support JWT VPs (although it only supports once and you can't add multiple proofs, see suggestion in this issue on some changes I want to propose for V2 of the attachment format: https://github.com/hyperledger/aries-rfcs/issues/807)

TimoGlastra (approved):

TimoGlastra:
Merging, would be great if you can still do a small follow up PR @berendsliedrecht

berendsliedrecht:
Should be fixed in #1696

PR #1680, opened 2023-12-22
refactor(oob)!: make label optional

### LINKED ISSUE #1524: Make `label` optional in `OutOfBandInvitation`
It is not required according to the specification.

Also need to update the `parseInvitation` method to not add an empty string as the label when parsing legacy connectionless invitations (which don't have labels).

TimoGlastra:
Fixes #1524


packages/core/src/modules/oob/helpers.ts
```diff
@@ -45,15 +45,17 @@ export function convertToOldInvitation(newInvitation: OutOfBandInvitation) {
   if (typeof service === 'string') {
     options = {
       id: newInvitation.id,
-      label: newInvitation.label,
+      // label is optional
+      label: newInvitation.label ?? '',
```
berendsliedrecht, line 49: If it is optional, why do we still fallback to an empty string?

berendsliedrecht (approved):

packages/core/src/modules/oob/helpers.ts
TimoGlastra: Because it's not optional in an conection protocol invitation, only the oob invitation.

genaris:
@TimoGlastra what about merging this before #1705 so we can create an OOB Invitation without label?

TimoGlastra:
Sounds good, it's ready to merge if CI passes :)

PR #1684, opened 2023-12-27
fix: properly print key class

TimoGlastra:
This makes sure that at least in the ConsoleLogger the `Key` class is printend as the `fingerprint` property, so we avoid this:

```
{
    "publicKey": {
      "type": "Buffer",
      "data": [
        67,
        227,
        94,
        17,
        202,
        44,
        168,
        217,
        210,
        62,
        120,
        65,
        119,
        12,
        186,
        150,
        210,
        1,
        153,
        142,
        54,
        158,
        119,
        223,
        138,
        189,
        1,
        4,
        72,
        145,
        192,
        23
      ]
    },
    "keyType": "ed25519"
  }
```

### DIFF
--- a/packages/core/src/crypto/Key.ts
+++ b/packages/core/src/crypto/Key.ts
@@ -59,4 +59,10 @@ export class Key {
   public get supportsSigning() {
     return isSigningSupportedForKeyType(this.keyType)
   }
+
+  // We return the fingerprint when this object is serialized to JSON
+  // This means it's nicely formatted when printing to the console
+  private toJSON() {
+    return this.fingerprint
+  }
 }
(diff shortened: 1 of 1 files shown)


berendsliedrecht:
Won't this cause any issues with the `JsonEncoder` or any `JSON.stringify` function?

TimoGlastra:
Yes it will, and I've been thinking about that as well, but you shouldn't really put a `Key` class into a `JsonEncoder` or `JSON.stringify`, so I thought how bad is it in that case if it encodes it to a multibase string (contains both the key type + encoding scheme)? But maybe the toJSON should rather be an object with keytype + publicKeyBase58?

genaris:
> But maybe the toJSON should rather be an object with keytype + publicKeyBase58?

I would say this might be more appropriate, as it would return a JSON object as it does today but in a more meaningful way than serializing a buffer instance.

TimoGlastra:
> I would say this might be more appropriate, as it would return a JSON object as it does today but in a more meaningful way than serializing a buffer instance.

Done ✅

genaris (approved):

PR #1693, opened 2024-01-08
feat(indy-vdr): register revocation registry definitions and status list

berendsliedrecht:
- Adds support for the `IndyVdrAnonCredsRegistry` to register:
  - Revocation Registry Definition
  - Revocation Status List
- Implementation is mostly copied from the `registerCredentialDefinition` as there is a 90% overlap.


packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
```diff
@@ -552,8 +562,180 @@ export class IndyVdrAnonCredsRegistry implements AnonCredsRegistry {
...
+      const { endorsedTransaction, endorserDid, endorserMode } = options
+      if (endorsedTransaction) {
+        agentContext.config.logger.debug(
+          `Preparing endorsed tx '${endorsedTransaction}' for submission on ledger '${namespace}' with did '${revocationRegistryDefinition.issuerId}'`,
+          revocationRegistryDefinition
+        )
+        writeRequest = new CustomRequest({ customRequest: endorsedTransaction })
+        const operation = JSON.parse(endorsedTransaction)?.operation
+        // extract the seqNo from the endorsed transaction, which is contained in the ref field of the operation
+        seqNo = Number(operation?.ref)
```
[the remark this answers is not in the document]
TimoGlastra: I think `ref` is used in the cred def transaction to reference the schema

packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
```diff
@@ -660,8 +842,108 @@ export class IndyVdrAnonCredsRegistry implements AnonCredsRegistry {
...
+          submitterDid: namespaceIdentifier,
+          revocationRegistryEntry: {
+            ver: '1.0',
+            value: {
+              accum: revocationStatusList.currentAccumulator,
+            },
+          },
+          revocationRegistryDefinitionType: 'CL_ACCUM',
+          revocationRegistryDefinitionId: revocationStatusList.revRegDefId,
+        })
```
TimoGlastra, line 883: We also need to pass the issued/revoked indices here. These are missing in the JS wrapper: https://github.com/hyperledger/indy-vdr/issues/165

Not sure if you can pass them with a ts-ignore for now and they will be passed on automatically to indy-vdr native lib or whether actual changes need to be published for this transaction can work: https://github.com/hyperledger/indy-vdr/issues/165

TimoGlastra: It seems to be passing the entry value as a whole to indy-vdr, so I think it should just work by passing the values: https://github.com/hyperledger/indy-vdr/blob/main/wrappers/javascript/indy-vdr-nodejs/src/NodeJSIndyVdr.ts#L405

See: https://github.com/hyperledger/indy-node/blob/main/design/anoncreds.md#revoc_reg_entry-txn

TimoGlastra (commented):
Nice :)

packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
berendsliedrecht: It is still a bit confusing, but the public API now also accepts the issued/revoked arrays as I assumed that that was the point. The changes between indy-sdk, indy-vdr, non-indy ledgers and anoncreds objects vs indy objects is causing quite some confusion ;).

packages/anoncreds/src/models/registry.ts
```diff
@@ -37,7 +37,9 @@ export interface AnonCredsRevocationRegistryDefinition {
 export interface AnonCredsRevocationStatusList {
   issuerId: string
   revRegDefId: string
-  revocationList: number[]
+  revocationList: Array<number>
   currentAccumulator: string
+  issued?: Array<number>
+  revoked?: Array<number>
```
TimoGlastra, line 43:
```suggestion
```

These should not be added here. The revocation status list is the model as used by AnonCreds now, and defined in the AnonCreds specification. Indy uses deltas and in those deltas are `issued` and `revoked` keys. 

So we need to do the following:
- Whenever we want to issue/revoke we create or update a status list, then in the indy anoncreds registry implementation we take the current status list and compare it to the new status list. Based on that we can get the issued/revoked indices
- Whenever we want to get a status list (this is already implemented before your PR) we get the detlas, and based on all deltas we construct the revocation list.

So it's important to know: Indy ALWAYS uses detlas on ledger, but everywhere else we use revocation status list as much as possible. So the indy way of doing things should only be used in the Indy anonCreds registry and be transformed to the anoncreds public models  (which is revocation status list)

packages/anoncreds/src/utils/indyIdentifiers.ts
```diff
@@ -32,18 +32,31 @@ export function getUnqualifiedSchemaId(unqualifiedDid: string, name: string, ver
...
+export function getUnqualifiedRevocationRegistryEntryId(
   unqualifiedDid: string,
-  seqNo: string | number,
+  schemaSeqNo: string | number,
   credentialDefinitionTag: string,
   revocationRegistryTag: string
 ) {
-  return `${unqualifiedDid}:4:${unqualifiedDid}:3:CL:${seqNo}:${credentialDefinitionTag}:CL_ACCUM:${revocationRegistryTag}`
+  return `${unqualifiedDid}:5:${unqualifiedDid}:3:CL:${schemaSeqNo}:${credentialDefinitionTag}:CL_ACCUM:${revocationRegistryTag}`
 }
```
TimoGlastra, line 60: When do we need to retrieve revocation registry entries? We only need to get the deltas right?

packages/indy-sdk/src/anoncreds/utils/transform.ts
```diff
@@ -103,6 +103,8 @@ export function anonCredsRevocationStatusListFromIndySdk(
   return {
     issuerId: revocationRegistryDefinition.issuerId,
     currentAccumulator: delta.value.accum,
+    revoked: delta.value.revoked,
+    issued: delta.value.issued,
```
TimoGlastra, line 107: We should use the deltas only to create the status list

```suggestion
```

packages/indy-vdr/src/anoncreds/utils/transform.ts
```diff
@@ -28,12 +34,8 @@ export function anonCredsRevocationStatusListFromIndyVdr(
     currentAccumulator: delta.accum,
     revRegDefId: revocationRegistryDefinitionId,
     revocationList,
+    revoked: delta.revoked,
+    issued: delta.issued,
```
TimoGlastra, line 38:
```suggestion
```

packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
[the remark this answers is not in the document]
TimoGlastra: I think this is not fixed yet (not sure if you were still looking at it)

packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
```diff
@@ -660,8 +854,171 @@ export class IndyVdrAnonCredsRegistry implements AnonCredsRegistry {
...
+
+      const schemaSeqNo = Number(schemaMetadata.indyLegderSeqNo)
+      const didIndyRevocationRegistryEntryId = getDidIndyRevocationRegistryEntryId(
+        namespace,
+        namespaceIdentifier,
+        schemaSeqNo,
+        credentialDefinition.tag,
+        revocationRegistryDefinition.tag
+      )
+
```
TimoGlastra, line 931: The revocation registry definition id is already passed as part of the input to this method, so I don't think it's needed to fetch the schema + cred def + revoc reg to construct the id.

Rather we can take it from `revocationStatusList .revRegDefId`

packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
```diff
@@ -660,8 +854,171 @@ export class IndyVdrAnonCredsRegistry implements AnonCredsRegistry {
...
+        const revocationRegistryDefinitionRequest = new RevocationRegistryEntryRequest({
+          submitterDid: namespaceIdentifier,
+          revocationRegistryEntry: {
+            ver: '1.0',
+            value: {
+              accum: revocationStatusList.currentAccumulator,
+              // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+              // @ts-expect-error
+              issued: revocationStatusList.issued ?? [],
+              revoked: revocationStatusList.revoked ?? [],
```
TimoGlastra, line 950: So here we need to get the current state, and based on the current statusList and the new statusList we need to calculate the `issued` and `revoked` indices.

packages/indy-vdr/tests/indy-vdr-anoncreds-registry.e2e.test.ts
```diff
@@ -953,177 +955,187 @@ describe('IndyVdrAnonCredsRegistry', () => {
...
+        endorserDid,
       },
-      resolutionMetadata: {},
     })
 
-    const legacyRevocationStatusList = await indyVdrAnonCredsRegistry.getRevocationStatusList(
-      agent.context,
-      legacyRevocationRegistryId,
-      entryResponse.result.txnMetadata.txnTime
+    console.log(JSON.stringify(revRegDefResp))
```
TimoGlastra, line 984:
```suggestion
```

packages/anoncreds/src/models/registry.ts
berendsliedrecht: How do we construct a status list from the deltas? Is that something that indy-vdr provides?

TimoGlastra: There's an util for that in AFJ :  https://github.com/openwallet-foundation/agent-framework-javascript/blob/1d333770dcc9e261446b43b5f4cd5626fa7ac4a7/packages/indy-vdr/src/anoncreds/utils/transform.ts#L3C7-L3C7

berendsliedrecht:
@TimoGlastra I am getting very close, but somehow it can't fetch the delta when I try to make the request (logs here https://github.com/openwallet-foundation/agent-framework-javascript/actions/runs/7531063980/job/20498795265?pr=1693#step:10:93)

Do you have any idea? Registration seems to respond with all the correct data.

TimoGlastra:
Did you add a sleep? Indy takes a while to process a write. I'll take a good look at it tomorrow

berendsliedrecht:
> Did you add a sleep? Indy takes a while to process a write. I'll take a good look at it tomorrow 

Yes! But it is just a second, so maybe increasing it will help

packages/indy-vdr/tests/indy-vdr-anoncreds-registry.e2e.test.ts
```diff
@@ -90,7 +90,7 @@ describe('IndyVdrAnonCredsRegistry', () => {
     await agent.wallet.delete()
   })
 
-  test('register and resolve a schema and credential definition (internal, issuerDid != endorserDid)', async () => {
+  xtest('register and resolve a schema and credential definition (internal, issuerDid != endorserDid)', async () => {
```
TimoGlastra, line 93: These are disabled and also are not updated with the new register methods.

packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
```diff
@@ -552,48 +582,211 @@ export class IndyVdrAnonCredsRegistry implements AnonCredsRegistry {
...
+          },
+          revocationRegistryDefinitionState: {
+            revocationRegistryDefinition,
+            state: 'failed',
+            reason: `error resolving schema with id ${credentialDefinition.schemaId}: ${resolutionMetadata.error} ${resolutionMetadata.message}`,
+          },
+        }
+      }
+
+      const schemaSeqNo = schemaMetadata.indyLedgerSeqNo
```
TimoGlastra, line 639: We can extract the schema seqNo from the cred def id

TimoGlastra: So we don't have to fetch the schema

packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
```diff
@@ -552,48 +582,211 @@ export class IndyVdrAnonCredsRegistry implements AnonCredsRegistry {
...
-        revocationRegistryId: legacyRevocationRegistryId,
-        toTs: timestamp,
-      })
+      let writeRequest: CustomRequest
+      let didIndyRevocationRegistryDefinitionId: string
 
-      agentContext.config.logger.trace(
-        `Submitting get revocation registry delta request for revocation registry '${revocationRegistryId}' to ledger`
+      // TODO: this will bypass caching if done on a higher level.
+      const { credentialDefinition, resolutionMetadata } = await this.getCredentialDefinition(
```
TimoGlastra, line 605: Why do we need to fetch the cred def? We can extract all needed parts from the cred def id. And indy will reject if something is wrong

packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
```diff
@@ -643,25 +819,173 @@ export class IndyVdrAnonCredsRegistry implements AnonCredsRegistry {
...
+      const didIndyRevocationRegistryEntryId = getDidIndyRevocationRegistryEntryId(
+        namespace,
+        namespaceIdentifier,
+        schemaSeqNo,
+        credentialDefinitionTag,
+        revocationRegistryTag
+      )
+
+      if (revocationRegistryNamespace && revocationRegistryNamespace !== namespace) {
+        // throw error
```
TimoGlastra, line 883: These are still todo. If we do it in separate pr we should remove it for now

TimoGlastra (approved):
Sorry a few more comments. But rest looks good!!

PR #1699, opened 2024-01-18
feat: did rotate

genaris:
Adds support for [Aries RFC 0794](https://github.com/hyperledger/aries-rfcs/tree/main/features/0794-did-rotate). 

API changes in Connections module are quite simple:

- **rotate**:  given a connection id, rotate our did to a new one. We can specify an external did (that we control) or create a peer did (did:peer:4 by default). Routing setting can be also set.
Former and new dids' keys will be marked as tags for the connection, in order to accept messages that may be received out of order (e.g. a message using previous keys after receiving the acknowledge)
- **hangup**: It will terminate a connection. Can also delete it (and remove any mediation key). If not deleted (default behaviour), it will still be able to receive messages but will not allow to send messages through it

There are also some fixes related to the changes introduced in #1550, related to cases where external dids are specified.

For the moment, when a rotation is being done, previous dids related to the connection are not removed, meaning that it is still possible to receive messages using former keys. There is a new API method called **removePreviousDids** that allows to explicitly remove any previous did associated to a given connection.

We should also consider that the keys related to dids are not actually being deleted from the wallet (this is something pending from general connection deletion flows), which is something that certainly would be nice to add in a further PR.


swcurran:
FYI @dbluhm @Jsyro

dbluhm:
Just to comment on what I think we'll end up doing in ACA-Py on the question of when to delete keys, ACA-Py has a cache for mapping from inbound keys to a did then to a connection. The cache has a TTL of one hour by default. So I think we'll commit the changes immediately and then let the cache handle any potentially straggling out-of-order messages. We'll see how that works out in implementation.

packages/core/src/modules/connections/ConnectionsApi.ts
```diff
@@ -278,6 +287,57 @@ export class ConnectionsApi {
     return message
   }
 
+  public async rotate(options: { connectionId: string; did?: string; routing?: Routing }) {
```
TimoGlastra, line 290:
```suggestion
  public async rotate(options: { connectionId: string; toDid?: string; routing?: Routing }) {
```

packages/core/src/modules/connections/ConnectionsApi.ts
```diff
@@ -278,6 +287,57 @@ export class ConnectionsApi {
...
+    const connection = await this.connectionService.getById(this.agentContext, connectionId)
+
+    if (did && routing) {
+      throw new AriesFrameworkError(`'routing' is disallowed when defining 'did'`)
+    }
+
+    const message = await this.didRotateService.createRotate(this.agentContext, {
+      connection,
+      did,
+      routing: routing || (await this.routingService.getRouting(this.agentContext, {})),
```
TimoGlastra, line 301: Do we also want to create routing when we have a toDid specified?

packages/core/src/modules/connections/ConnectionsApi.ts
```diff
@@ -394,6 +454,36 @@ export class ConnectionsApi {
...
+   * Note: this will delete both ours and theirs previous dids
+   *
+   * @param connectionId
+   */
+  public async removePreviousDids(options: { connectionId: string }) {
+    const connection = await this.connectionService.getById(this.agentContext, options.connectionId)
+
+    for (const previousDid of connection.previousDids) {
+      const did = await this.didResolverService.resolve(this.agentContext, previousDid)
+      if (!did.didDocument) break
```
TimoGlastra, line 470: Should the did doc also be removed? Once we support dids othee than did:peer we will need to update this probably as you may not want to remove your public did (we could pass in a method array or something?)

removePreviousDids({ methods: ['key', 'peer', 'jwk'] })

packages/core/src/modules/connections/DidExchangeProtocol.ts
```diff
@@ -194,31 +187,33 @@ export class DidExchangeProtocol {
     // Get DID Document either from message (if it is a supported did:peer) or resolve it externally
     const didDocument = await this.resolveDidDocument(agentContext, message)
 
-    if (isValidPeerDid(didDocument.id)) {
```
TimoGlastra, line 197: Does this mean we now support connections with other dids than did:peer?

packages/core/src/modules/connections/__tests__/did-rotate.e2e.test.ts
```diff
@@ -0,0 +1,385 @@
...
+      const oldDid = aliceBobConnection!.did
+      expect(bobAliceConnection!.theirDid).toEqual(oldDid)
+
+      // Send message to initial did
+      await bobAgent.basicMessages.sendMessage(bobAliceConnection!.id, 'Hello initial did')
+
+      await waitForBasicMessage(aliceAgent, { content: 'Hello initial did' })
+
+      // Do did rotate
+      const { did: newDid } = await aliceAgent.connections.rotate({ connectionId: aliceBobConnection!.id })
```
TimoGlastra, line 91: maybe we can also rename the return value. Maybe newDid, or toDid

packages/core/src/modules/connections/handlers/DidRotateAckHandler.ts
```diff
@@ -0,0 +1,17 @@
+import type { MessageHandler, MessageHandlerInboundMessage } from '../../../agent/MessageHandler'
+import type { DidRotateService } from '../services'
+
+import { RotateAckMessage } from '../messages'
```
TimoGlastra, line 4:
```suggestion
import { DidRotateAckMessage } from '../messages'
```

packages/core/src/modules/connections/handlers/DidRotateHandler.ts
```diff
@@ -0,0 +1,26 @@
+import type { MessageHandler, MessageHandlerInboundMessage } from '../../../agent/MessageHandler'
+import type { DidRotateService } from '../services'
+import type { ConnectionService } from '../services/ConnectionService'
+
+import { AriesFrameworkError } from '../../../error'
+import { RotateMessage } from '../messages'
```
TimoGlastra, line 6:
```suggestion
import { DidRotateMessage } from '../messages'
```

packages/core/src/modules/connections/services/DidRotateService.ts
```diff
@@ -0,0 +1,265 @@
...
+  ) {
+    const { connection, did, routing } = options
+
+    const config = agentContext.dependencyManager.resolve(ConnectionsModuleConfig)
+
+    // Do not allow to receive concurrent did rotation flows
+    const didRotateMetadata = connection.metadata.get(ConnectionMetadataKeys.DidRotate)
+
+    if (didRotateMetadata) {
+      this.logger.warn(`There is already an existing opened did rotation flow for connection id ${connection.id}`)
```
TimoGlastra, line 57: Should this return? As you said do not allow concurrent flows

tests/transport/SubjectInboundTransport.ts
```diff
@@ -48,7 +48,16 @@ export class SubjectInboundTransport implements InboundTransport {
           })
         }
 
-        await messageReceiver.receiveMessage(message, { session })
+        // This emits a custom error in order to easily catch in E2E tests when a message
+        // reception throws an error. TODO: Remove as soon as agent throws errors automatically
```
TimoGlastra, line 52: Yeah it would be good if we can extend the current event to also emit when the processing went wrong. There's an issue open for it

TimoGlastra (approved):
No blockers, just some nits

packages/core/src/modules/connections/services/DidRotateService.ts
genaris: Yes, I think it is better to throw error in that case. I didn't want to be so strict, in order to allow restarting the flow in case the ack is never received (or the rotate message couldn't be delivered), but I'll add a method at `DidRotateService` level to clear rotation data so it will be somehow possible from a client to reset the flow without getting into details about the metadata.

packages/core/src/modules/connections/DidExchangeProtocol.ts
genaris: I didn't do tests other than using the 'InMemory DID registry', but it should work, especially now we have the did caching (although for messaging maybe we should use a long-term cache instead of the default of 5 minutes).

TimoGlastra: Okay makes sense.

Yes the 5 minutes was really to prevent multiple loads within a single exchange, without needing to worry about DIDs changing over time

TimoGlastra: We could look at fancy caching strategies for this, like `stale-while-revalidate`: https://web.dev/articles/stale-while-revalidate

packages/core/src/modules/connections/ConnectionsApi.ts
genaris: For the moment the method is only intended to remove the relationship between the connection and its previous dids, meaning that records will not be deleted. This is actually the same thing that happens when we delete a connection. 

IMHO we should revisit this process in order to clean up the underlying keys as well, but as you say, now we are supporting more DID methods and flows we should carefully analyze when this data is safe to be removed.

packages/core/src/modules/connections/ConnectionsApi.ts
genaris: If `toDid` is specified, routing is disallowed (the same thing as we did in #1550).

TimoGlastra: Yes if routing is not provided we will go to `|| (await this.routingService.getRouting(this.agentContext, {})),`, so it willl still be generated right? I think we shouldn't call `getRouting` in this case

genaris: I'm under a heavy jetlag right now but I think that case will not happen, as a few lines before that we have:

```ts
if (toDid && routing) {
      throw new AriesFrameworkError(`'routing' is disallowed when defining 'did'`)
    }
```

TimoGlastra: Yes but if `routing` is not defined, this will not throw an error. Then later on it will do `routing || (await this.routingService.getRouting(this.agentContext, {}))`. Routing will never be defined when `toDid` is defined, so that will mean it will call `this.routingService`.

I think `|| (await this.routingService.getRouting(this.agentContext, {}))` should ONLY be called if `toDid` is not defined.

```
if (toDid && routing) {
  throw new AriesFrameworkError(`'routing' is disallowed when defining 'did'`)
}

let routingToUse = routing
if (!toDid && !routing) {
  routingToUse = await this.routingService.getRouting(this.agentContext, {})
}
```

TimoGlastra: Maybe I'm missing something here, but I think your influenced by the jetlag 😅

genaris: Ooh I see. It is creating an unnecessary routing 🤦. Thanks for the explanation, I think I should have some rest 😄.

PR #1705, opened 2024-01-24
feat: support short legacy connectionless invitations

TimoGlastra:
When a legacy connectionless invitation is hosted and shared as a short url. E.g. in the case of https://github.com/bcgov/vc-authn-oidc

### DIFF
--- a/packages/core/src/utils/parseInvitation.ts
+++ b/packages/core/src/utils/parseInvitation.ts
@@ -59,6 +59,9 @@ export const parseInvitationJson = (invitationJson: Record<string, unknown>): Ou
     const outOfBandInvitation = convertToNewInvitation(invitation)
     outOfBandInvitation.invitationType = InvitationType.Connection
     return outOfBandInvitation
+  } else if (invitationJson['~service']) {
+    // This is probably a legacy connectionless invitation
+    return transformLegacyConnectionlessInvitationToOutOfBandInvitation(invitationJson)
   } else {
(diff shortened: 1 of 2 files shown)


packages/core/src/utils/parseInvitation.ts
```diff
@@ -105,6 +108,33 @@ export const oobInvitationFromShortUrl = async (response: Response): Promise<Out
...
+  // This destructuring removes the ~service property from the message, and
+  // we can can use messageWithoutService to create the out of band invitation
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const { '~service': service, ...messageWithoutService } = messageJson
+
+  // transform into out of band invitation
+  const invitation = new OutOfBandInvitation({
+    // The label is currently required by the OutOfBandInvitation class, but not according to the specification.
+    // FIXME: In 0.5.0 we will make this optional: https://github.com/hyperledger/aries-framework-javascript/issues/1524
+    label: '',
```
genaris, line 128: If we can merge the optional label PR first, we can simply avoid using this empty label. See https://github.com/openwallet-foundation/credo-ts/pull/1680#issuecomment-1914514095

genaris (approved):

packages/core/src/utils/parseInvitation.ts
TimoGlastra: I'll update once #1680 is merged

PR #1708, opened 2024-01-28
feat: openid4vc and sd-jwt-vc support

TimoGlastra:
Supersedes #1667, as there were issues with DCO sign-off

It works e2e with openid4vc issuance, and verification with SD-JWT-VC, BUTT there's still some todos, but If possible I'd like to do it in follow up PRs as this one is already huge.

Some of the todos:
- add events so you can hook into the state changes of offers/requests
- Update some FIXMEs once PRs have been merged into PEX/OID4VC/SD-JWT libraries (it's currently all working, but sometimes a bit hacky)
- add support for request_uri as the QR for presentation requests can get quite large at the moment
- Add improvements to the API so it's known as a holder which field will be selectively disclosed (you now just get the record)
- Issue that awesome-qr MUST be installed on react native
- support jwk thumbprint in Sphereon siop/oid4vp libs (currently only dids are supported for proving, opened https://github.com/Sphereon-Opensource/SIOP-OID4VP/issues/67, but this requires some substantial work)
- Allow presenting multiple SD-JWT vcs in a single request. My first option was to go for creating multiple VPs for a single presentation definitoin and submission, but this is not supported by PEX/OID4VP libraries from Sphereon (See https://github.com/Sphereon-Opensource/SIOP-OID4VP/issues/62). It does support multiple Presentation definitions, so maybe that's the way to go (although I'd like using one PD more)
- Replace the in-memory session with something persistent

Here's a list of PRs that if released will allow for some cleanup on our side (but we can do that after this PR is merged):
- https://github.com/berendsliedrecht/sd-jwt-ts/pull/35
- https://github.com/berendsliedrecht/sd-jwt-ts/pull/33
- https://github.com/berendsliedrecht/sd-jwt-ts/pull/34
- https://github.com/Sphereon-Opensource/PEX/pull/140
- https://github.com/Sphereon-Opensource/SIOP-OID4VP/pull/71
- https://github.com/Sphereon-Opensource/SIOP-OID4VP/pull/70


packages/core/src/modules/sd-jwt-vc/__tests__/sdjwtvc.fixtures.ts
```diff
@@ -0,0 +1,17 @@
+export const simpleJwtVc =
```
berendsliedrecht, line 1: Are these just created by the framework? If not, would be good to link to the source for where they were found.

packages/core/src/modules/sd-jwt-vc/repository/SdJwtVcRecord.ts
```diff
@@ -0,0 +1,66 @@
...
+    }
+  }
+
+  public getTags() {
+    const sdJwtVc = SdJwtVc.fromCompact(this.compactSdJwtVc)
+
+    return {
+      ...this._tags,
+      vct: sdJwtVc.getClaimInPayload<string>('vct'),
+      sdAlg: (sdJwtVc.payload._sd_alg as string | undefined) ?? 'sha-256',
```
berendsliedrecht, line 58: `getClaimInPayload` can also be used here

packages/core/src/modules/vc/data-integrity/SignatureSuiteRegistry.ts
```diff
@@ -27,12 +27,15 @@ export class SignatureSuiteRegistry {
...
+  /**
+   * @deprecated recommended to always search by key type instead as that will have broader support
+   */
   public getByVerificationMethodType(verificationMethodType: string) {
     return this.suiteMapping.find((x) => x.verificationMethodTypes.includes(verificationMethodType))
   }
 
   public getByKeyType(keyType: KeyType) {
-    return this.suiteMapping.find((x) => x.keyTypes.includes(keyType))
+    return this.suiteMapping.filter((x) => x.keyTypes.includes(keyType))
```
berendsliedrecht, line 38: Do we want the name of the API to indicate that it will return multiple?

packages/core/src/modules/vc/data-integrity/models/W3cJsonLdVerifiableCredential.ts
```diff
@@ -55,4 +60,8 @@ export class W3cJsonLdVerifiableCredential extends W3cCredential {
   public get encoded() {
     return this.toJson()
   }
+
+  public get jsonCredential(): W3cJsonCredential {
```
berendsliedrecht, line 64: Can't the `this.encoded` method just be used for this? Or is the casting something that should stay unique for this method

packages/core/src/utils/path.ts
```diff
@@ -7,3 +7,29 @@
...
   return path.substring(0, Math.max(path.lastIndexOf('/'), path.lastIndexOf('\\')))
 }
+
+/**
+ * Combine multiple uri parts into a single uri taking into account slashes.
+ *
+ * @param parts the parts to combine
+ * @returns the combined url
+ */
+export function joinUriParts(base: string, parts: string[]) {
```
berendsliedrecht, line 17: Might be good to call `trim()` all strings before checking whether they end with a `/`, or start.

packages/openid4vc/README.md
```diff
@@ -0,0 +1,177 @@
...
+#### Requirements
+
+Before a credential can be requested, you need the issuer URI. This URI starts with `openid-initiate-issuance://` and is provided by the issuer. The issuer URI is commonly acquired by scanning a QR code.
+
+#### Module registration
+
+In order to get this module to work, we need to inject it into the agent. This makes the module's functionality accessible through the agent's `modules` api.
+
+```ts
+import { OpenId4VcHolderModule } from '@aries-framework/openid4vc-holder'
```
berendsliedrecht, line 52:
```suggestion
import { OpenId4VcHolderModule } from '@aries-framework/openid4vc'
```

right?

packages/openid4vc/package.json
```diff
@@ -10,26 +10,31 @@
...
-    "directory": "packages/openid4vc-client"
+    "directory": "packages/openid4vc"
   },
   "scripts": {
     "build": "yarn run clean && yarn run compile",
     "clean": "rimraf ./build",
     "compile": "tsc -p tsconfig.build.json",
     "prepublishOnly": "yarn run build",
-    "test": "jest"
+    "test": "jest --forceExit --detectOpenHandles"
```
berendsliedrecht, line 24: Was this just required for debugging or do we always need it for this?

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
```diff
@@ -0,0 +1,719 @@
...
+import { OpenId4VciCredentialFormatProfile } from '../shared'
+import {
+  getTypesFromCredentialSupported,
+  handleAuthorizationDetails,
+  getOfferedCredentials,
+} from '../shared/issuerMetadataUtils'
+import { getSupportedJwaSignatureAlgorithms } from '../shared/utils'
+
+import {
+  type OpenId4VciAuthCodeFlowOptions,
```
berendsliedrecht, line 66: Shouldn't we use `import type { ...`? Or will this work always now? Vaguely remember that this caused some issues with me before, but it might've been resolved now.

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
```diff
@@ -0,0 +1,719 @@
...
+    const client = await OpenID4VCIClient.fromURI({
+      uri: credentialOffer,
+      resolveOfferUri: true,
+      retrieveServerMetadata: true,
+    })
+
+    if (!client.credentialOffer?.credential_offer) {
+      throw new AriesFrameworkError(`Could not resolve credential offer from '${credentialOffer}'`)
+    }
+    const credentialOfferPayload: OpenId4VciCredentialOfferPayload = client.credentialOffer?.credential_offer
```
berendsliedrecht, line 103:
```suggestion
    const credentialOfferPayload: OpenId4VciCredentialOfferPayload = client.credentialOffer.credential_offer
```

Assertion above should have confirmed it exists

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
```diff
@@ -0,0 +1,719 @@
...
+    }
+  }
+
+  public async resolveAuthorizationRequest(
+    agentContext: AgentContext,
+    resolvedCredentialOffer: OpenId4VciResolvedCredentialOffer,
+    authCodeFlowOptions: OpenId4VciAuthCodeFlowOptions
+  ): Promise<OpenId4VciResolvedAuthorizationRequest> {
+    const { credentialOfferPayload, metadata, offeredCredentials } = resolvedCredentialOffer
+    const codeVerifier = `${await agentContext.wallet.generateNonce()}${await agentContext.wallet.generateNonce()}`
```
berendsliedrecht, line 173: Any performance increase here when we use `Promise.allSettled`?

packages/openid4vc/src/openid4vc-issuer/repository/OpenId4VcIssuerRecord.ts
```diff
@@ -0,0 +1,60 @@
...
+   * The fingerprint (multibase encoded) of the public key used to sign access tokens for
+   * this issuer.
+   */
+  accessTokenPublicKeyFingerprint: string
+
+  credentialsSupported: OpenId4VciCredentialSupportedWithId[]
+  display?: OpenId4VciIssuerMetadataDisplay[]
+}
+
+export class OpenId4VcIssuerRecord extends BaseRecord<DefaultOpenId4VcIssuerRecordTags> {
```
berendsliedrecht, line 29: I think this could benefit from a comment. We never really had an `issuer` record for DIDComm and I am a bit confused whether an issuer should only have one record or multiple for different servers (if that even makes sense)?

packages/openid4vc/src/openid4vc-issuer/router/accessTokenEndpoint.ts
```diff
@@ -0,0 +1,161 @@
+import type { OpenId4VcIssuanceRequest } from './requestContext'
+import type { AgentContext } from '@aries-framework/core'
+import type { JWTSignerCallback } from '@sphereon/oid4vci-common'
+import type { NextFunction, Response, Router } from 'express'
```
berendsliedrecht, line 4: For now it is okay to leave it like this, but I feel that including express in here is not a good solution in the long run. There are many other options out there and especially when using other runtimes like Deno.

packages/openid4vc/src/openid4vc-verifier/OpenId4VcSiopVerifierService.ts
```diff
@@ -0,0 +1,364 @@
...
+      throw new AriesFrameworkError(`Unable to find request state for correlationId '${correlationId}'`)
+    }
+
+    const requestClientId = await requestSessionState.request.getMergedProperty<string>('client_id')
+    const requestNonce = await requestSessionState.request.getMergedProperty<string>('nonce')
+    const requestState = await requestSessionState.request.getMergedProperty<string>('state')
+    const presentationDefinitionsWithLocation = await requestSessionState.request.getPresentationDefinitions()
+
+    if (!requestNonce || !requestClientId || !requestState) {
+      throw new AriesFrameworkError(
```
berendsliedrecht, line 132: It would be good to log the requestNonce, requestClientId and requestState here so the user knows which one is undefined.

packages/openid4vc/src/openid4vc-verifier/OpenId4VcSiopVerifierService.ts
```diff
@@ -0,0 +1,364 @@
...
+    }
+
+    const requestSessionState = await sessionManager.getRequestStateByCorrelationId(correlationId)
+    if (!requestSessionState) {
+      throw new AriesFrameworkError(`Unable to find request state for correlationId '${correlationId}'`)
+    }
+
+    const requestClientId = await requestSessionState.request.getMergedProperty<string>('client_id')
+    const requestNonce = await requestSessionState.request.getMergedProperty<string>('nonce')
+    const requestState = await requestSessionState.request.getMergedProperty<string>('state')
```
berendsliedrecht, line 128:
```suggestion
    const requestClientId = await requestSessionState.request.getMergedProperty<string | undefined>('client_id')
    const requestNonce = await requestSessionState.request.getMergedProperty<string | undefined>('nonce')
    const requestState = await requestSessionState.request.getMergedProperty<string | undefined>('state')
```

Or does the get method return `<T | undefined>`?

packages/openid4vc/src/openid4vc-verifier/OpenId4VcSiopVerifierService.ts
```diff
@@ -0,0 +1,364 @@
...
+      builder.withSignature(suppliedSignature)
+
+      _clientId = suppliedSignature.did
+    }
+
+    if (!_clientId) {
+      throw new AriesFrameworkError("Either 'requestSigner' or 'clientId' must be provided.")
+    }
+
+    // FIXME: we now manually remove did:peer, we should probably allow the user to configure this
```
berendsliedrecht, line 244: `did:peer` could still be valid right? Haven't seen it in any other context other than didcomm, but if you have exchanged it before hand it'll work. 

It can be left like this, as the user should likely define it themselves and by default just allow for all supported resolver methods.

packages/openid4vc/src/openid4vc-verifier/OpenId4VcVerifierModule.ts
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  public async initialize(rootAgentContext: AgentContext): Promise<void> {
+    this.configureRouter(rootAgentContext)
+  }
+
+  /**
+   * Registers the endpoints on the router passed to this module.
+   */
+  private configureRouter(rootAgentContext: AgentContext) {
```
berendsliedrecht, line 57: Have not checked it out in great detail, but will the express servers work fine if you are an issuer AND verifier?

packages/openid4vc/src/openid4vc-verifier/repository/OpenId4VcVerifierRecord.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export interface OpenId4VcVerifierRecordProps {
+  id?: string
+  createdAt?: Date
+  tags?: TagsBase
+
+  verifierId: string
+}
+
+export class OpenId4VcVerifierRecord extends BaseRecord<DefaultOpenId4VcVerifierRecordTags> {
```
berendsliedrecht, line 19: same comment as for the issuer record.

packages/openid4vc/src/shared/utils.ts
```diff
@@ -0,0 +1,103 @@
...
+    // get the key from the verification method and use the first supported signature algorithm
+    key = getKeyFromVerificationMethod(verificationMethod)
+    const _alg = getJwkClassFromKeyType(key.keyType)?.supportedSignatureAlgorithms[0]
+    if (!_alg) throw new AriesFrameworkError(`No supported signature algorithms for key type: ${key.keyType}`)
+
+    alg = _alg
+    kid = verificationMethod.id
+    did = verificationMethod.controller
+  } else {
+    throw new AriesFrameworkError("Unsupported jwt issuer method. Only 'did' is supported.")
```
berendsliedrecht, line 60:
```suggestion
    throw new AriesFrameworkError(`Unsupported jwt issuer method ('${jwtIssuer.method}'). Only 'did' is supported.`)
```

berendsliedrecht (approved):
Really great work @TimoGlastra and @2mau !!!

There are indeed quite some FIXME's, TODO's and NOTE's in there but I am perfectly fine with fixing them in other PRs as this PR already made my browser lag a whole bunch ;).

I think it would be good if another person went over it as with such a big PR, you tend to miss some details in some places.

packages/cheqd/src/anoncreds/utils/identifiers.ts
```diff
@@ -9,18 +9,23 @@ const IDENTIFIER = `((?:${ID_CHAR}*:)*(${ID_CHAR}+))`
 const PATH = `(/[^#?]*)?`
 const QUERY = `([?][^#]*)?`
 const VERSION_ID = `(.*?)`
+const FRAGMENT = `([#].*)?`
```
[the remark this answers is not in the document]
TimoGlastra: Patches that I found while working on this PR. I tested issuance of OpenID4VC with cheqd, and it was failing because of this.

And instead of making the right choice to create a separate PR i added it all to this one 🙃

packages/core/src/modules/dif-presentation-exchange/DifPresentationExchangeService.ts
```diff
@@ -327,14 +210,38 @@ export class DifPresentationExchangeService {
...
 
-    for (const vpf of verifiablePresentationResultsWithFormat) {
-      const { verifiablePresentationResult } = vpf
-      presentationSubmission.descriptor_map.push(...verifiablePresentationResult.presentationSubmission.descriptor_map)
-    }
+    verifiablePresentationResultsWithFormat.forEach(({ verifiablePresentationResult }, index) => {
+      // FIXME: path_nested should not be used for sd-jwt.
+      // Can be removed once https://github.com/Sphereon-Opensource/PEX/pull/140 is released
+      const descriptorMap = verifiablePresentationResult.presentationSubmission.descriptor_map.map((d) => {
+        const descriptor = { ...d }
```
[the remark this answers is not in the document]
TimoGlastra: I mainly didn't want to overwrite the descriptor value as I call delete, so I treat it as an immutable object. I think that should be fine?

packages/openid4vc/src/openid4vc-verifier/OpenId4VcSiopVerifierService.ts
TimoGlastra: Yes method returns it as optional type

packages/core/src/modules/dif-presentation-exchange/DifPresentationExchangeService.ts
```diff
@@ -545,4 +494,78 @@ export class DifPresentationExchangeService {
 
     return verificationMethod
   }
+
+  /**
+   * Queries the wallet for credentials that match the given presentation definition. This only does an initial query based on the
+   * schema of the input descriptors. It does not do any further filtering based on the constraints in the input descriptors.
+   */
+  private async queryCredentialForPresentationDefinition(
```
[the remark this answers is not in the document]
TimoGlastra: Yeah right now we use the Indy/Askar WQL format, which is very similar to how Mongo query language works: https://www.mongodb.com/docs/manual/tutorial/query-documents/

I think for now it's ok to assume a backend will implement this (as we've done that for all other records as well).

I do want to revamp this a lot more, as It's really not scalable and every time we need to query something we don't have as tag we need to re-store all records...

packages/core/src/modules/sd-jwt-vc/SdJwtVcService.ts
```diff
@@ -0,0 +1,425 @@
...
+    } as const
+
+    const payload = {
+      iat: verifierMetadata.issuedAt,
+      nonce: verifierMetadata.nonce,
+      aud: verifierMetadata.audience,
+
+      // FIXME: _sd_hash is missing. See
+      // https://github.com/berendsliedrecht/sd-jwt-ts/issues/8
+      _sd_hash: TypedArrayEncoder.toBase64URL(Hasher.hash(compactDerivedSdJwtVc, sdAlg)),
```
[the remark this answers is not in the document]
TimoGlastra: Once there's a stable release for sd-jwt yes, will do in follow up PR

packages/core/src/modules/sd-jwt-vc/__tests__/sdjwtvc.fixtures.ts
TimoGlastra: Yes created by the framework. Would be good to have external fixtures as well

packages/core/src/modules/sd-jwt-vc/repository/SdJwtVcRecord.ts
TimoGlastra: It will throw when not found, and then it becomes this multi-line monster

```
let sdAlg: string
try {
  sdAlg = sdJwtVc.getClaimInPayload('_sd_alg')
} catch {
  sdAlg = 'sha-256'
}
```

packages/core/src/modules/vc/data-integrity/models/W3cJsonLdVerifiableCredential.ts
TimoGlastra: I want to have the same methods availalbe on both JWT and JSON-LD as otherwise you have to do instance checks. This way typescript is smart enough to see `jsonCredentail` is availalbe on both.

While encoded is string for JWT and object for JSON-LD.

I'm also thinking of adding `encoded` to SdJwtVc in AFJ as that will allow you to call .encoded no matter if you have an SD-JWT, JWT or JSON-LD, making it a lot easier to work with

packages/openid4vc/src/shared/utils.ts
TimoGlastra: The PAIN of typescript making it 'never' and thus throwing an error here :(

Basically with TS this code can never be reached, but in JS land it can ...

packages/openid4vc/src/openid4vc-verifier/OpenId4VcVerifierModule.ts
TimoGlastra: Yes, you just have to make sure it's exposed on a different path. See https://github.com/animo/openid4vc-playground/blob/b2b039bd316914aaedfda752b1a99e9326c5ba55/agent/src/server.ts#L28-L29

packages/openid4vc/package.json
TimoGlastra: No not needed

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
TimoGlastra: I think the issue we had recently was to do `export type * from` which is TS 5+. But I usually avoid this as well jsut to prevent TS version issues. Will update

packages/openid4vc/src/openid4vc-issuer/repository/OpenId4VcIssuerRecord.ts
TimoGlastra: Will add. The generic thing is that for OID4VC you need to expos metadata files. Each issuer needs to host this metadata. This is not the case for DIDComm where we can just have one `/didcomm` endpoint. So we create a record per openid issuer/verifier that you want, and each tenant can create multiple issuers/verifiers

packages/openid4vc/src/openid4vc-issuer/router/accessTokenEndpoint.ts
TimoGlastra: Yes I agree, but as we're already using express for DIDComm I didn't see a good reason to do it different for oid4vc.

packages/core/src/modules/sd-jwt-vc/repository/SdJwtVcRecord.ts
berendsliedrecht: Ah right. Likely a specific method for the hashing algo would be best that defaults to Sha256.

packages/core/src/modules/vc/data-integrity/models/W3cJsonLdVerifiableCredential.ts
berendsliedrecht: Ah I see, that makes sense!

PR #1710, opened 2024-01-29
feat(dids)!: did caching

TimoGlastra:
Adds support for caching of did resolver results. I saw that in a lot flows we're resoling the same did multiple times, and thus sometimes hitting the ledger 5 times for the same did. This adds a default cache of 5 minutes for did documents.

You can disable this by passing options to resolve, and I've disabled it for dids where it doesn't make sense: `did:peer`, `did:key` and `did:jwk`.

BREAKING CHANGE: it is now required to set the `allowsCaching` on a `DidResolver` implementation. When set to `true` the `DidResolverService` will allow the results from the `DidResolver` implementation to be cached. When `false`, a result from the `DidResolver` will NEVER be cached. It is advised to not allow caching for dynamically generated or stored did documents (such as `did:peer`, `did:key` and `did:jwk`), but do allow caching for resolvers that are expensive to call (think a ledger did method such as `did:cheqd` or `did:indy`).

### DIFF
--- a/packages/core/src/modules/dids/domain/DidResolver.ts
+++ b/packages/core/src/modules/dids/domain/DidResolver.ts
@@ -3,6 +3,8 @@ import type { ParsedDid, DidResolutionResult, DidResolutionOptions } from '../ty
 
 export interface DidResolver {
   readonly supportedMethods: string[]
+  readonly allowsCaching: boolean
+
   resolve(
     agentContext: AgentContext,
     did: string,
--- a/packages/anoncreds-rs/tests/LocalDidResolver.ts
+++ b/packages/anoncreds-rs/tests/LocalDidResolver.ts
@@ -4,6 +4,7 @@ import { DidsApi } from '@aries-framework/core'
 
 export class LocalDidResolver implements DidResolver {
   public readonly supportedMethods = ['sov', 'indy']
+  public readonly allowsCaching = false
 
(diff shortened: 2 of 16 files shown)


packages/core/src/modules/dids/domain/DidResolver.ts
```diff
@@ -3,6 +3,8 @@ import type { ParsedDid, DidResolutionResult, DidResolutionOptions } from '../ty
 
 export interface DidResolver {
   readonly supportedMethods: string[]
+  readonly allowsCaching: boolean
```
genaris, line 6: Since we are forced to add this flag in case we have a custom DID resolver, would this be a breaking change? It's not a problem but if that's the case should be added to the release notes.

genaris (approved):
Great addition! I like the flexibility of choosing when we want to use it or not, and defining custom expiration time (useful for some particular cases).

packages/core/src/modules/dids/domain/DidResolver.ts
TimoGlastra: I can make it optional and set a default?

But might be good to make it explicit and a breaking change. Should be easy to configure.

What do you prefer?

genaris: I think it is fine to make it explicit. We are also updating some other interfaces in this release as well (like the ones for AnonCreds registries), so it should not be a problem I guess.

PR #1711, opened 2024-01-29
fix: support all minor versions handshake

TimoGlastra:
Fixes a regression that popped up after introducing support for didexchange 1.1 and invitations using didexchange 1.0 now not being accepted anymore. I found it in our Paradym Wallet when updating, while it received an invitation from Paradym server still on 0.4.2

I've updated the `HandshakeProtocol` enum to include the minor version as `.x` now, and we don't care anymore about the minor version. When creating an out of band invitation we map the `.x` version to the current minor version supported by the agent, and when receiving we match the minor version to any minor version we support.

This results in:
- ✅ receive `https://didcomm.org/didexchange/1.1` while current supported version is `https://didcomm.org/didexchange/1.1`
- ✅ receive `https://didcomm.org/didexchange/1.0` while current supported version is `https://didcomm.org/didexchange/1.1`
- ✅ receive `https://didcomm.org/didexchange/1.1` while current supported version is `https://didcomm.org/didexchange/1.0`
- ❌  receive `https://didcomm.org/didexchange/2.0` while current supported version is `https://didcomm.org/didexchange/1.0`


genaris (approved):
Good timing for this fix, as I experienced the same issue when interacting with ACA-Py and was about to see how to solve it in an elegant way.

I think it would be nice to have something like the `ParsedMessageType` for the handshake protocols that can return the part we want but maybe it's not worthy, considering that we likely won't touch this code anymore as soon as we move to DIDComm V2.

I left some comments after my review in terms of variable/method naming but feel free to ignore them if they are clear for you. I'm OK as long as the code does its job properly 😄.

packages/core/src/agent/MessageHandlerRegistry.ts
```diff
@@ -1,9 +1,11 @@
 import type { AgentMessage } from './AgentMessage'
 import type { MessageHandler } from './MessageHandler'
+import type { ParsedDidCommProtocolUri } from '../utils/messageType'
 
 import { injectable } from 'tsyringe'
+import { version } from 'uuid'
```
TimoGlastra, line 6:
```suggestion
```

PR #1714, opened 2024-01-30
chore: make deprecated packages private

TimoGlastra:
Making these packages temporarily private before the 0.5.0 to see if we will release them under 0.5.0

Otherwise It'd be great if we can keep the `@credo-ts` scope on NPM as clean as possible:
- `@credo-ts/sd-jwt-vc` -> integrated into core in #1708 
- `@credo-ts/anoncreds-rs` -> I'd like to merge this with `@credo-ts/anoncreds` if we remove indy-sdk
- `@credo-ts/indy-sdk` -> we're trying to deprecate this one, so no need to publish it anymore

If we end up still releasing them for 0.5.0 I'll just create a PR again to set them to not private

### DIFF
--- a/packages/anoncreds-rs/package.json
+++ b/packages/anoncreds-rs/package.json
@@ -3,6 +3,7 @@
   "main": "build/index",
   "types": "build/index",
   "version": "0.4.2",
+  "private": true,
   "files": [
     "build"
   ],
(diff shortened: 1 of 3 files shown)


TimoGlastra:
Curious what you think of this one:
`@credo-ts/anoncreds-rs` -> I'd like to merge this with `@credo-ts/anoncreds` if we remove indy-sdk

To me it feels like it should be one package

karimStekelenburg (approved):

PR #1720, opened 2024-01-31
feat: add Multikey as supported vm type

dbluhm:
This PR adds support for the `Multikey` verification method type defined in [vc-data-integrity](https://www.w3.org/TR/vc-data-integrity/#multikey). Similar to `JsonWebKey2020`, it can be used to represent multiple key types.

This is perhaps a naive implementation. This is my first time digging around inside of credo. Please let me know if there's anything I need to improve!


packages/core/src/modules/dids/domain/key-type/ed25519.ts
```diff
@@ -40,6 +44,10 @@ export const keyDidEd25519: KeyDidMapping = {
       return getKeyFromJsonWebKey2020(verificationMethod)
     }
 
+    if (isMultikey(verificationMethod)) {
```
TimoGlastra, line 47: We also have this in keyDidMapping (a higher level), so i don't know when it'll reach this, but i see we have the same duplication for JsonWebKey2020 so it's okay for now.

We can refactor it a bit later on

packages/core/src/modules/dids/domain/verificationMethod/Multikey.ts
```diff
@@ -0,0 +1,50 @@
...
+ * Get a key from a Multikey verification method.
+ */
+export function getKeyFromMultikey(verificationMethod: VerificationMethod & { type: 'Multikey' }) {
+  if (!verificationMethod.publicKeyMultibase) {
+    throw new AriesFrameworkError(
+      `Missing publicKeyMultibase on verification method with type ${VERIFICATION_METHOD_TYPE_MULTIKEY}`
+    )
+  }
+
+  return Key.fromFingerprint(verificationMethod.publicKeyMultibase)
```
TimoGlastra, line 49: fingerprint was supposed to give a fingerprint of the key, and it also is a multibase encoding. Might be good to add a toMultibase and fromMultibase options and also support e.g. base64 encoding as described in the di spec. But fine for now

TimoGlastra (approved):

TimoGlastra:
Thanks 🙏

PR #1734, opened 2024-02-01
fix(anoncreds): pass along options for registry and status list

berendsliedrecht:


TimoGlastra (approved):
Maybe a test would be good?

PR #1739, opened 2024-02-05
feat: add goal codes to v2 protocols

TimoGlastra:
Adds the `goal` and `goal_code` properties to all the messages and interfaces for v2 protocols.

`goal_code` was already enabled for proof messages, but not goal. No validation, no logic to it, just something an user of Credo can use to add context to an exchange. 

Goal codes are adopted by v2 of the protocols.


karimStekelenburg (approved):
LGTM

PR #1744, opened 2024-02-06
feat: anoncreds w3c migration

auer-martin:


packages/core/src/modules/credentials/formats/dataIntegrity/DataIntegrityCredentialFormat.ts
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * This defines the module payload for calling CredentialsApi.offerCredential
+ * or CredentialsApi.negotiateProposal
+ */
+export interface DataIntegrityOfferCredentialFormat {
+  credential: W3cCredential | JsonObject
+  bindingRequired: boolean
+  anonCredsLinkSecretBindingMethodOptions?: AnonCredsLinkSecretBindingMethodOptions
+  didCommSignedAttachmentBindingMethodOptions?: DidCommSignedAttachmentBindingMethodOptions
```
TimoGlastra, line 41:
```suggestion
  anonCredsLinkSecretBinding?: AnonCredsLinkSecretBindingMethodOptions
  didCommSignedAttachmentBinding?: DidCommSignedAttachmentBindingMethodOptions
```

packages/core/src/modules/dif-presentation-exchange/DifPresentationExchangeService.ts
```diff
@@ -384,10 +399,38 @@ export class DifPresentationExchangeService {
...
+      )
+
+      return inputDescriptor?.format === 'di_vp' &&
+        verifiableCredentials.credential.credential instanceof W3cJsonLdVerifiableCredential
+        ? verifiableCredentials.credential.credential.cryptoSuites
+        : []
+    })
+
+    const commonCryptoSuites = cryptosuites.reduce((a, b) => a.filter((c) => b.includes(c)))
+    if (commonCryptoSuites.length === 0 || !commonCryptoSuites.includes('anoncreds-2023')) return undefined
```
TimoGlastra, line 420:
```suggestion
    if (commonCryptoSuites.length === 0 || !commonCryptoSuites.includes('anoncreds-2023')) return false
```

packages/core/src/modules/dif-presentation-exchange/DifPresentationExchangeService.ts
```diff
@@ -48,6 +55,8 @@ import {
   getSphereonOriginalVerifiableCredential,
 } from './utils'
 
+// export type ProofStructure = Record<string, Record<string, Array<W3cCredentialRecord>>>
```
TimoGlastra, line 58:
```suggestion
```

packages/core/src/modules/proofs/formats/dif-presentation-exchange/DifPresentationExchangeProofFormatService.ts
```diff
@@ -218,6 +221,24 @@ export class PresentationExchangeProofFormatService implements ProofFormatServic
...
+  ) {
+    if (presentation.claimFormat !== ClaimFormat.LdpVp) return false
+
+    const descriptorMap = presentationSubmission.descriptor_map
+
+    const verifyUsingDataIntegrity = descriptorMap.every((descriptor) => descriptor.format === ClaimFormat.DiVp)
+    if (!verifyUsingDataIntegrity) return false
+
+    if (Array.isArray(presentation.proof)) {
+      return presentation.proof.some((proof) => proof.cryptosuite?.includes('anoncreds-2023'))
```
TimoGlastra, line 236: Maybe we can make `anoncreds-2023` a const that we use everyweyre as it's now harcoded in quite some places

packages/core/src/modules/vc/W3cCredentialService.ts
```diff
@@ -178,6 +178,7 @@ export class W3cCredentialService {
     const w3cCredentialRecord = new W3cCredentialRecord({
       tags: { expandedTypes },
       credential: options.credential,
+      anonCredsCredentialRecordOptions: options.anonCredsCredentialRecordOptions,
```
TimoGlastra, line 181:
```suggestion
```

packages/core/src/modules/vc/W3cCredentialServiceOptions.ts
```diff
@@ -190,4 +191,5 @@ export interface W3cJsonLdVerifyPresentationOptions extends W3cVerifyPresentatio
 
 export interface StoreCredentialOptions {
   credential: W3cVerifiableCredential
+  anonCredsCredentialRecordOptions?: AnonCredsCredentialRecordOptions
```
TimoGlastra, line 194:
```suggestion
```

TimoGlastra (dismissed):
Lot of work @2mau!!! I've left quite some comments. Most of them are small housekeeping fixes, some require some more substantial refacotrs. I'm fine with possibly moving some improvements over to after this is merged and released, but especially the API surface can sometimes use some clenaup I think. The logic itself makes sense though and I think you made the right abstractions, nice job 👍

packages/anoncreds/src/anoncreds-rs/AnonCreds2023DataIntegrityService.ts
```diff
@@ -0,0 +1,580 @@
...
+    presentationSubmission: PresentationSubmission,
+    credentials: JsonObject[]
+  ) => {
+    const credentialsProve: CredentialProve[] = []
+    const schemaIds = new Set<string>()
+    const credentialDefinitionIds = new Set<string>()
+    const credentialsWithMetadata: CredentialWithMetadata[] = []
+
+    const hash = Hasher.hash(TypedArrayEncoder.fromString(presentationDefinition.id), 'sha-256')
+    const nonce = new BigNumber(hash).toString().slice(0, 32)
```
TimoGlastra, line 390: BTW -- I think this should be documented as it can lead to issues with verification if the nonce is different during verification.

I think we should use the `challenge` the presentation request: https://github.com/hyperledger/aries-rfcs/blob/main/features/0510-dif-pres-exch-attach/README.md#request-presentation-attachment-format 

We already pass this to the `w3cCredentialService.verifyPresentation` in the `DifPresentationExchangeProofFormatService` and I think we can pass it here as well.

MAYBE we need to define that if the nonce is not a 32 bytes long number, you need to hash the `challenge` from DIF PEX, and take the first 32 bytes of the bignumber of the hash (I don't know exactly what bignumber does).

Opened an issue in AnonCreds spec as well: https://github.com/hyperledger/anoncreds-spec/issues/198

TimoGlastra:
@2mau https://github.com/openwallet-foundation/credo-ts/pull/1754 has just been merged. The tests were flaky and almost always failing. There are some changes however to e2e tests vs unit tests.

Could you make sure that all the tests you edited / added are following the new pattern:
- e2e test: uses indy / cheqd / postgres / http / ws
- unit test: other tests, can run without internet and external services

If a test is a e2e test it should end with .e2e.test.ts

An easy way to check is running the unit tests (`yarn test:unit`) without any of the docker-compose services running and any tests that fail because the service is not running should use `.e2e.test.ts`. Other way around is a bit harder, and is a bit more manual. Not sure how much of tests you wrote are using in memory anoncreds  registry (preffered) vs indy/cheqd registry (for a few e2e tests)

packages/core/src/modules/credentials/formats/dataIntegrity/DataIntegrityCredentialFormat.ts
```diff
@@ -0,0 +1,66 @@
...
+export interface DataIntegrityOfferCredentialFormat {
+  credential: W3cCredential | JsonObject
+  bindingRequired: boolean
+  anonCredsLinkSecretBinding?: AnonCredsLinkSecretCreateOfferOptions
+  didCommSignedAttachmentBinding?: DidCommSignedAttachmentCreateOfferOptions
+}
+
+/**
+ * This defines the module payload for calling CredentialsApi.acceptRequest. No options are available for this
+ * method, so it's an empty object
```
TimoGlastra, line 42:
```suggestion
 * This defines the module payload for calling CredentialsApi.acceptRequest.
```

packages/cheqd/tests/helpers/cheqdAnonCredsSetup.ts
```diff
@@ -0,0 +1,356 @@
...
+    holderVerifierConnectionId: holderVerifierConnection?.id,
+    verifierHolderConnectionId: verifierHolderConnection?.id,
+  } as unknown as SetupAnonCredsTestsReturn<VerifierName, CreateConnections>
+}
+
+export async function prepareForAnonCredsIssuance(
+  agent: Agent,
+  { attributeNames, issuerId }: { attributeNames: string[]; issuerId: string }
+) {
+  //const key = await agent.wallet.createKey({ keyType: KeyType.Ed25519 })
```
TimoGlastra, line 284:
```suggestion
```

packages/cheqd/tests/helpers/cheqdAnonCredsSetup.ts
```diff
@@ -0,0 +1,356 @@
...
+    options: {
+      network: 'testnet',
+      methodSpecificIdAlgo: 'uuid',
+    },
+  })
+  expect(did.didState).toMatchObject({ state: 'finished' })
+  return did.didState.did as string
+}
+
+export async function setupAnonCredsTests<
```
TimoGlastra, line 156: This is a lot of duplication for swapping out cheqd vs indy-vdr. Would it be easier to combine them and add some conditional steps?  Basically everything in this file is duplicated except for the cheqd specific stuff (which is very minimal)

packages/cheqd/tests/cheqd-data-integrity.e2e.test.ts
```diff
@@ -0,0 +1,273 @@
...
+    await anonCredsFlowTest({
+      issuerId,
+      credentialDefinitionId,
+      issuerHolderConnectionId,
+      holderIssuerConnectionId,
+      issuerReplay,
+      holderReplay,
+      issuer: issuerAgent,
+      holder: holderAgent,
+    })
```
TimoGlastra, line 67: as mentioned in previous review. If we only have on test there's no need to extract it into a method. Makes it harder to udnerstand the flow IMO

packages/cheqd/tests/cheqd-data-integrity.e2e.test.ts
```diff
@@ -0,0 +1,273 @@
...
+  let issuerAgent: AnonCredsTestsAgent
+  let holderAgent: AnonCredsTestsAgent
+  let credentialDefinitionId: string
+  let issuerHolderConnectionId: string
+  let holderIssuerConnectionId: string
+
+  let issuerReplay: EventReplaySubject
+  let holderReplay: EventReplaySubject
+
+  const inMemoryRegistry = new InMemoryAnonCredsRegistry()
```
TimoGlastra, line 33: Not used right?

packages/anoncreds/tests/fixtures/presentation-definition.ts
```diff
@@ -0,0 +1,56 @@
+import type { PresentationDefinitionV1 } from '@sphereon/pex-models'
```
TimoGlastra, line 1: Same here, would be good to use as much types from the `dif-presentation-exchange` module in core, to prevent dependency on the sphereon libs all across the codebase (same like we try to hide AnonCreds RS implementation behind AnonCreds generic interface)

packages/anoncreds/tests/data-integrity.e2e.test.ts
```diff
@@ -0,0 +1,303 @@
...
+import { InMemoryAnonCredsRegistry } from '../../anoncreds/tests/InMemoryAnonCredsRegistry'
+import { createDidKidVerificationMethod } from '../../core/tests'
+import { waitForCredentialRecordSubject, waitForProofExchangeRecord } from '../../core/tests/helpers'
+
+import { setupAnonCredsTests } from './anoncredsSetup'
+import { presentationDefinition } from './fixtures/presentation-definition'
+
+const issuerId = 'did:indy:local:LjgpST2rjsoxYegQDRm7EL'
+
+describe('anoncreds w3c data integrity e2e tests', () => {
```
TimoGlastra: This can be non .e2e.test.ts as it doesn't depend on the indy-ledger right?

packages/anoncreds/tests/InMemoryAnonCredsRegistry.ts
```diff
@@ -19,22 +19,32 @@ import type {
...
-import { Hasher } from '@credo-ts/core'
+import { Hasher, utils } from '@credo-ts/core'
 import BigNumber from 'bn.js'
 
 import {
   getDidIndyCredentialDefinitionId,
   getDidIndyRevocationRegistryDefinitionId,
   getDidIndySchemaId,
 } from '../../indy-vdr/src/anoncreds/utils/identifiers'
+import {} from '../src'
```
TimoGlastra, line 30:
```suggestion
```

packages/anoncreds/tests/InMemoryAnonCredsRegistry.ts
```diff
@@ -152,56 +178,77 @@ export class InMemoryAnonCredsRegistry implements AnonCredsRegistry {
...
-      namespaceIdentifier,
-      indyLedgerSeqNo,
-      options.credentialDefinition.tag
-    )
-
-    this.credentialDefinitions[didIndyCredentialDefinitionId] = options.credentialDefinition
+    const schemaId = options.credentialDefinition.schemaId
+
+    let credentialDefinitionId: string
+    if (isIndyDid(schemaId) || isUnqualifiedSchemaId(schemaId)) {
```
TimoGlastra, line 186: There should be no case where the schemaId is unqualified

packages/anoncreds/src/utils/w3cAnonCredsUtils.ts
```diff
@@ -0,0 +1,201 @@
...
+    revocationRegistryId,
+    credentialRevocationId,
+    linkSecretId,
+    methodName,
+  } = options
+
+  const issuerId = w3cCredentialRecord.credential.issuerId
+
+  const anonCredsCredentialRecordTags: AnonCredsCredentialTags = {
+    anonCredsCredentialId: w3cCredentialRecord.id,
```
TimoGlastra, line 166: Same here. If it's the same as the id of the record, there's no need to add the tag

packages/anoncreds/src/utils/metadata.ts
```diff
@@ -16,14 +38,8 @@ export const AnonCredsCredentialMetadataKey = '_anoncreds/credential'
...
- * MUST be used with {@link AnonCredsCredentialMetadataKey}
+ * MUST be used with {@link W3cAnoncredsCredentialMetadata}
  */
-export interface AnonCredsCredentialMetadata {
-  schemaId?: string
-  credentialDefinitionId?: string
-  revocationRegistryId?: string
-  credentialRevocationId?: string
-}
+export const W3cAnonCredsCredentialMetadataKey = '_w3c/AnonCredsMetadata'
```
TimoGlastra, line 45: for consistency 
```suggestion
export const W3cAnonCredsCredentialMetadataKey = '_w3c/anonCredsMetadata'
```

packages/anoncreds/src/utils/anonCredsObjects.ts
```diff
@@ -0,0 +1,95 @@
...
+      `RevocationRegistryDefinition not found for id ${revocationRegistryDefinitionId}: ${result.resolutionMetadata.message}`
+    )
+  }
+
+  const indyNamespace = result.revocationRegistryDefinitionMetadata.didIndyNamespace
+
+  return {
+    revocationRegistryDefinition: result.revocationRegistryDefinition,
+    revocationRegistryDefinitionId,
+    indyNamespace: indyNamespace && typeof indyNamespace === 'string' ? indyNamespace : undefined,
```
TimoGlastra, line 69: In what case is indyNamespace defined, but not string?
I.e. what is the difference with:
```suggestion
    indyNamespace
```

TimoGlastra: Or is it to get typing for TS right?

packages/anoncreds/src/updates/0.4-0.5/anonCredsCredentialRecord.ts
```diff
@@ -0,0 +1,163 @@
...
+    qualifiedSchemaIssuerId = getQualifiedDidIndyDid(legacyTags.schemaIssuerId, indyNamespace)
+  } else {
+    qualifiedCredentialDefinitionId = legacyTags.credentialDefinitionId
+    qualifiedIssuerId = legacyTags.issuerId
+    qualifiedRevocationRegistryId = legacyTags.revocationRegistryId
+    qualifiedSchemaId = legacyTags.schemaId
+    qualifiedSchemaIssuerId = legacyTags.schemaIssuerId
+  }
+
+  const anonCredsRsHolderService = agentContext.dependencyManager.resolve(AnonCredsRsHolderService)
```
TimoGlastra, line 93: should not depend on anocreds-rs directly

packages/anoncreds/src/updates/0.4-0.5/anonCredsCredentialRecord.ts
```diff
@@ -0,0 +1,163 @@
...
+
+  if (
+    !isUnqualifiedCredentialDefinitionId(legacyTags.credentialDefinitionId) &&
+    !isUnqualifiedIndyDid(legacyTags.issuerId)
+  ) {
+    if (isIndyDid(legacyTags.issuerId)) {
+      indyNamespace = getIndyNamespaceFromIndyDid(legacyTags.issuerId)
+    }
+  } else {
+    indyNamespace = await getIndyNamespace(agentContext, legacyTags.credentialDefinitionId, legacyTags.issuerId)
```
TimoGlastra, line 74: This can return no value. In that case we don't want to continue with migration to a w3c record right? Or am i missing a step here where it would prevent from continuing if it can't extract the indyNamespace (the `getIndyNamespace` seems to catch and not re-throw)

packages/anoncreds/src/models/W3cAnonCredsMetadata.ts
TimoGlastra, line 1: empty file

packages/anoncreds/src/formats/LegacyIndyProofFormatService.ts
```diff
@@ -488,32 +488,16 @@ export class LegacyIndyProofFormatService implements ProofFormatService<LegacyIn
...
-        throw new CredoError(
-          `Credential definition not found for id ${credentialDefinitionId}: ${credentialDefinitionResult.resolutionMetadata.message}`
-        )
+      const credentialDefinitionResult = await fetchCredentialDefinition(agentContext, credentialDefinitionId)
+      if (isUnqualifiedCredentialDefinitionId(credentialDefinitionResult.credentialDefinitionId)) {
+        credentialDefinitions[credentialDefinitionId] = credentialDefinitionResult.credentialDefinition
+      } else {
+        credentialDefinitions[getUnQualifiedDidIndyDid(credentialDefinitionId)] =
+          getUnqualifiedDidIndyCredentialDefinition(credentialDefinitionResult.credentialDefinition)
       }
```
TimoGlastra, line 500: This should not support qualifieid id. So we need to get back the 

```
      if (!isUnqualifiedCredentialDefinitionId(credentialDefinitionId)) {
        throw new CredoError(`${credentialDefinitionId} is not a valid legacy indy credential definition id`)
      }
```

code and then we can remove the if / else here. This is `LegacyIndyProofFormatService` which only supports unqualified.

packages/anoncreds/src/formats/LegacyIndyProofFormatService.ts
```diff
@@ -456,23 +464,15 @@ export class LegacyIndyProofFormatService implements ProofFormatService<LegacyIn
...
-      if (!isUnqualifiedSchemaId(schemaId)) {
-        throw new CredoError(`${schemaId} is not a valid legacy indy schema id`)
-      }
-
-      const schemaRegistry = registryService.getRegistryForIdentifier(agentContext, schemaId)
-      const schemaResult = await schemaRegistry.getSchema(agentContext, schemaId)
-
-      if (!schemaResult.schema) {
-        throw new CredoError(`Schema not found for id ${schemaId}: ${schemaResult.resolutionMetadata.message}`)
+      const schemaResult = await fetchSchema(agentContext, schemaId)
```
TimoGlastra, line 470: same here

packages/anoncreds/src/formats/LegacyIndyProofFormatService.ts
```diff
@@ -456,23 +464,11 @@ export class LegacyIndyProofFormatService implements ProofFormatService<LegacyIn
...
    */
   private async getSchemas(agentContext: AgentContext, schemaIds: Set<string>) {
-    const registryService = agentContext.dependencyManager.resolve(AnonCredsRegistryService)
-
     const schemas: { [key: string]: AnonCredsSchema } = {}
 
     for (const schemaId of schemaIds) {
-      if (!isUnqualifiedSchemaId(schemaId)) {
-        throw new CredoError(`${schemaId} is not a valid legacy indy schema id`)
-      }
```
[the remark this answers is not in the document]
TimoGlastra: This service should not support qualified ids. It's a legacy service (LegacyIndyProofFormatService), and was from before qualifieid identifiers existed.

packages/anoncreds/src/formats/DataIntegrityCredentialFormatService.ts
```diff
@@ -0,0 +1,1195 @@
...
+      missingBindingMethod ||
+      invalidDataModelVersions ||
+      invalidLinkSecretBindingMethod ||
+      invalidDidCommSignedAttachmentBindingMethod
+    ) {
+      throw new ProblemReportError('Invalid credential offer', {
+        problemCode: CredentialProblemReportReason.IssuanceAbandoned,
+      })
+    }
+  }
```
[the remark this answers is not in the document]
TimoGlastra: This is not resolved?

packages/anoncreds/src/anoncreds-rs/__tests__/AnonCredsRsServices.test.ts
```diff
@@ -58,6 +71,22 @@ const agentContext = getAgentContext({
...
+      {
+        suiteClass: Ed25519Signature2018,
+        proofType: 'Ed25519Signature2018',
+        verificationMethodTypes: [
+          VERIFICATION_METHOD_TYPE_ED25519_VERIFICATION_KEY_2018,
+          VERIFICATION_METHOD_TYPE_ED25519_VERIFICATION_KEY_2020,
+        ],
+        keyTypes: [KeyType.Ed25519],
+      },
+    ],
```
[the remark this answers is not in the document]
TimoGlastra: Would be good to find out why -- Then we can maybe make some improvements. As we now have to add this 12 lines long code to a lot of files. Or we make it static and use that everywhere. Just a lot of duplication which we will have to change in a lot of places that don't even use Ed25519Signature2018

packages/anoncreds/src/anoncreds-rs/AnonCredsRsHolderService.ts
```diff
@@ -336,26 +505,69 @@ export class AnonCredsRsHolderService implements AnonCredsHolderService {
...
+          options.schemaId && isUnqualifiedSchemaId(options.schemaId) ? options.schemaId : undefined,
+        anonCredsUnqualifiedIssuerId:
+          options.issuerId && isUnqualifiedIndyDid(options.issuerId) ? options.issuerId : undefined,
+        anonCredsUnqualifiedSchemaIssuerId:
+          options.schemaIssuerId && isUnqualifiedIndyDid(options.schemaIssuerId) ? options.schemaIssuerId : undefined,
+        anonCredsUnqualifiedCredentialDefinitionId:
+          options.credentialDefinitionId && isUnqualifiedCredentialDefinitionId(options.credentialDefinitionId)
+            ? options.credentialDefinitionId
+            : undefined,
+      })
```
TimoGlastra, line 540: How does it deal with false being passed? I think that can result in issues?

Maybe we can make this a bit more of a builder flow? Even if it works with Askar, we're basically filtering for tags having value `false`.

TimoGlastra (commented):
Lots of good updates! I left some additional comments. Also some comments I left on previous review seemed unresolved but marked as resolved

packages/anoncreds/src/anoncreds-rs/AnonCreds2023DataIntegrityService.ts
TimoGlastra: I think this is great 👍

packages/anoncreds/src/anoncreds-rs/AnonCredsRsHolderService.ts
```diff
@@ -254,73 +333,200 @@ export class AnonCredsRsHolderService implements AnonCredsHolderService {
...
+          ? {
+              issuerId: getNonQualifiedId(credential.issuerId),
+              credentialDefinitionId: getNonQualifiedId(credentialDefinitionId),
+              schemaId: getNonQualifiedId(credentialDefinition.schemaId),
+              schemaIssuerId: getNonQualifiedId(schema.issuerId),
+              revocationRegistryId: revocationRegistryId ? getNonQualifiedId(revocationRegistryId) : undefined,
+            }
+          : undefined,
+      },
+    })
```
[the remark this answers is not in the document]
TimoGlastra: Your current approach looks good to me! 👍

packages/anoncreds/src/anoncreds-rs/AnonCredsRsHolderService.ts
auer-martin: what do you mean with false being passed?

packages/anoncreds/src/utils/anonCredsObjects.ts
auer-martin: it's unkown. I changed it and cast it to string | undefined

packages/anoncreds/src/formats/LegacyIndyProofFormatService.ts
auer-martin: how should we store credentials received via the legacyIndyCredentialFormatService?

packages/anoncreds/src/anoncreds-rs/AnonCredsRsHolderService.ts
TimoGlastra: Nevermind. Didn't understand the code correctly

packages/anoncreds/src/formats/DataIntegrityCredentialFormatService.ts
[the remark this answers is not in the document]
karimStekelenburg: @auer-martin do you know what to do here? If not, let me know!

karimStekelenburg (approved):

PR #1745, opened 2024-02-06
feat: optional backup on storage migration

genaris:
Make storage backup optional during wallet storage migration process. This is mainly to allow storage migration for those wallets that do not support exporting (like Askar wallets using PosgreSQL). 

@TimoGlastra I have to review it a bit more and add some tests, but wanted to open it now so you can check if it fits the changes you are doing on your side.


packages/anoncreds/src/updates/__tests__/0.3.test.ts
```diff
@@ -82,7 +82,7 @@ describe('UpdateAssistant | AnonCreds | v0.3.1 - v0.4', () => {
       },
     ])
 
-    await updateAssistant.update('0.4')
+    await updateAssistant.update({ updateToVersion: '0.4' })
```
TimoGlastra, line 85: note to self: update in update docs

packages/core/src/agent/AgentConfig.ts
```diff
@@ -71,6 +71,10 @@ export class AgentConfig {
     return this.initConfig.autoUpdateStorageOnStartup ?? false
   }
 
+  public get backupBeforeStorageUpdate() {
```
TimoGlastra, line 74: One thing to consider though that not creating a backup may be seen as advanced and you should use the update assistant in this case? But if it's useful to add this to the default agent startup, then I'm ok with it 👍

packages/core/src/storage/migration/UpdateAssistant.ts
```diff
@@ -107,8 +107,12 @@ export class UpdateAssistant<Agent extends BaseAgent<any> = BaseAgent> {
     return neededUpdates
   }
 
-  public async update(updateToVersion?: UpdateToVersion) {
+  public async update(options?: { updateToVersion?: UpdateToVersion; backupBeforeUpdate?: boolean }) {
     const updateIdentifier = Date.now().toString()
+    const updateToVersion = options?.updateToVersion
+
+    // By default do a backup first (should be explicitly disabled in case the wallet backend does not support export)
+    const doBackup = options?.backupBeforeUpdate ?? true
```
TimoGlastra, line 115:
```suggestion
    const createBackup = options?.backupBeforeUpdate ?? true
```

TimoGlastra (approved):

PR #1746, opened 2024-02-06
docs: OpenID Demo folder Readme

eduelias:
I was going to open an issue about the readme but I, instead, decided to do the decent thing: To open a PR with the fix.

The README is still a bit vague because I could not go further than the presentation resolving, as it seems that the HTTP/Express server is still not fully working on the demo. 

I'll open a issue (Or a PR, if I find how to fix it) soon.

I hope my wording is correct!


TimoGlastra:
Thanks for the PR @eduelias

> as it seems that the HTTP/Express server is still not fully working on the demo.

Hmm that's odd. I'll have to look into it

TimoGlastra (approved):

eduelias:
@TimoGlastra First, thanks for your approval! Now I see that the CI for Node 20.x is failing in this PR but I saw that you fixed it in yours, right? Can you, when the fix is merged, update this MR too, please? Thanks! :D 

I'm still trying to identify the problem with the 404 on the `holder.acceptPresentationRequest` method call.

TimoGlastra:
Found the issue @eduelias, pushed the fix to this branch. The verifier router was not passed to the verifier module, meaning the endpoints were not exposed 🤦 

Thanks for looking into this. Will merge now!

PR #1752, opened 2024-02-08
fix: did:peer:2 creation and parsing

### LINKED ISSUE #1682: Multiple services are encoded within `did:peer:2` incorrectly
According to the recent clarification in [Peer DID Method Specification](https://identity.foundation/peer-did-method-spec/#generating-a-didpeer2), multiple services must be encoded individually:
>When encoding multiple services, they MUST be encoded individually and concatenated together. The services MUST NOT be encoded as a JSON list.

Please also see the example in the end of the [section](https://identity.foundation/peer-did-method-spec/#generating-a-didpeer2). The clarification was made in https://github.com/decentralized-identity/peer-did-method-spec/commit/49a7c727a6205ba1e2a70ffc98c1a9f06930bff9.

However, in the latest release of Aries Framework JavaScript (v0.4.2) multiple services are encoded within `did:peer:2` all together as a JSON array:
https://github.com/openwallet-foundation/agent-framework-javascript/blob/v0.4.2/packages/core/src/modules/dids/methods/peer/peerDidNumAlgo2.ts#L141

This should be fixed in accordance with the latest specification.

genaris:
Changes in did:peer:2 creation and parsing as specified in https://identity.foundation/peer-did-method-spec/#method-2-multiple-inception-key-without-doc. Basically:

- When creating did:peer:2, encode multiple services separately (i.e. in concatenated 'S' tokens instead of a single token containing the list). Verification methods are named using '#key-N' pattern and ordered
- When resolving a did:peer:2, name verification methods with the pattern "did#key-N" rather than "did#fingerprint"

I need to add some more tests and use cases check if it might represent an issue with AFJ 0.4.2 (apparently not).

Closes #1682


packages/core/src/modules/dids/methods/peer/__tests__/__fixtures__/didPeer2Ez6L.json
```diff
@@ -2,21 +2,21 @@
   "id": "did:peer:2.Ez6LSbysY2xFMRpGMhb7tFTLMpeuPRaqaWM1yECx2AtzE3KCc.Vz6MkqRYqQiSgvZQdnBytw86Qbs2ZWUkGv22od935YF4s8M7V.Vz6MkgoLTnTypo3tDRwCkZXSccTPHRLhF4ZnjhueYAFpEX6vg.SeyJ0IjoiZG0iLCJzIjoiaHR0cHM6Ly9leGFtcGxlLmNvbS9lbmRwb2ludCIsInIiOlsiZGlkOmV4YW1wbGU6c29tZW1lZGlhdG9yI3NvbWVrZXkiXSwiYSI6WyJkaWRjb21tL3YyIiwiZGlkY29tbS9haXAyO2Vudj1yZmM1ODciXX0",
   "authentication": [
     {
-      "id": "did:peer:2.Ez6LSbysY2xFMRpGMhb7tFTLMpeuPRaqaWM1yECx2AtzE3KCc.Vz6MkqRYqQiSgvZQdnBytw86Qbs2ZWUkGv22od935YF4s8M7V.Vz6MkgoLTnTypo3tDRwCkZXSccTPHRLhF4ZnjhueYAFpEX6vg.SeyJ0IjoiZG0iLCJzIjoiaHR0cHM6Ly9leGFtcGxlLmNvbS9lbmRwb2ludCIsInIiOlsiZGlkOmV4YW1wbGU6c29tZW1lZGlhdG9yI3NvbWVrZXkiXSwiYSI6WyJkaWRjb21tL3YyIiwiZGlkY29tbS9haXAyO2Vudj1yZmM1ODciXX0#6MkqRYqQiSgvZQdnBytw86Qbs2ZWUkGv22od935YF4s8M7V",
+      "id": "#key-2",
```
TimoGlastra, line 5: It would be good to keep full ids here, at least in the 'resolved' variant as using relative identifiers is not valid and might lead to unexpected behavior

TimoGlastra (approved):

packages/core/src/modules/dids/methods/peer/__tests__/__fixtures__/didPeer2Ez6L.json
genaris: I think for the resolution it is OK to keep them, as you say. For creation, [the spec](https://identity.foundation/peer-did-method-spec/#creating-a-did) now states: _"All identifiers within this document MUST be relative. For example, the id of a verificationMethod might be #key-1 instead of something like did:example:abc123#key-1."_

In did:peer:2, key encoding ordering is important, as the resolver will need to reconstruct relative references (e.g. encoded within services) using "key-N" pattern, which was the issue found by @dbluhm when testing against ACA-Py.

packages/core/src/modules/dids/methods/peer/peerDidNumAlgo2.ts
```diff
@@ -115,20 +115,23 @@ export function didDocumentToNumAlgo2Did(didDocument: DidDocument) {
...
+      keys.push({ id: entry.id, encoded })
     })
-
-    // Add all encoded keys
-    did += encoded.join('')
   }
 
+  // Add all encoded keys ordered by their id (#key-1, #key-2, etc.)
+  did += keys
+    .sort((a, b) => a.id.localeCompare(b.id))
```
genaris, line 131: I think this will fail in case of >= 10 keys within a did:peer, so we should use a regex/remove '#key-' prefix and do the sorting numerically

TimoGlastra: Aren't the keys already sorted by default? I think we always process them start to end right?

packages/core/src/modules/dids/methods/peer/__tests__/__fixtures__/didPeer2Ez6L.json
TimoGlastra: I think this should say "All identifiers within _the encoded_ document MUST be relative". Otherwise it goes against the did-core specification, which mentions:

> The value of the [id](https://www.w3.org/TR/did-core/#dfn-id) property for a [verification method](https://www.w3.org/TR/did-core/#dfn-verification-method) MUST be a [string](https://infra.spec.whatwg.org/#string) that conforms to the rules in Section [3.2 DID URL Syntax](https://www.w3.org/TR/did-core/#did-url-syntax).

There's not really a reason to not follow the did-core spec here right?

TimoGlastra: The did:peer spec also mentions:

> This document should look almost exactly like the final resolved DID Document you desire but with a few key differences:

So I think it is talking about the encoded variant. The resolved DIDDocument should still follow the did-core spec rules I make up of this.

TimoGlastra:
@genaris any big changes still needed for this PR?

packages/core/src/modules/dids/methods/peer/peerDidNumAlgo2.ts
genaris: Given that our input is a modeled `DidDocument`, In `didDocumentToNumAlgo2Did()` we are sorting by key purpose (first assertionMethods, then keyAgreement, authentication, etc.), so it might happen that a key referenced by a service will not be encoded in the order we expect (e.g. mixing up verification key and key agreement key, or using key from another service). This is mainly the reason for this concern.

genaris:
> @genaris any big changes still needed for this PR?

I think there are no more changes needed from my side, unless we find a better way of handling these changes in the spec. I don't like so much the idea of limiting the key ids to the pattern 'key-N' (any call an existing app makes to `agent.dids.create()` for did:peer:2 not taking this into consideration will fail), but given the current state I think it safer to do so instead of doing some 'smart transformation' within did peer creation process (e.g. turning any key from input DidDocument into 'key-N' and make sure every reference to it is updated accordingly).

TimoGlastra:
> e.g. turning any key from input DidDocument into 'key-N' and make sure every reference to it is updated accordingly

Agreed. I've had this with several other did methods though, and it may be interesting to look at a better did builder pattern so you can input what type of did doc you want and the did registrar turns it into something compatible with the method.

E.g. i want two verificationMethods, one with keyType ed25519, the other one with this specific key. And most fields will just be optional and populated later on.

packages/core/src/modules/dids/methods/peer/peerDidNumAlgo2.ts
```diff
@@ -115,20 +115,35 @@ export function didDocumentToNumAlgo2Did(didDocument: DidDocument) {
...
 
-      return encoded
+      keys.push({ id: entry.id, encoded })
     })
+  }
 
-    // Add all encoded keys
-    did += encoded.join('')
+  const prefix = 'key-'
+  if (!keys.every((key) => key.id.split('#')[1]?.startsWith(prefix))) {
```
TimoGlastra, line 130: Might be better/cleaner to use parseDid here?

TimoGlastra (approved):

PR #1753, opened 2024-02-09
feat: added support for DRPC protocol

wadeking98:
### Send DRPC request and wait for response:
``` typescript
const listenResponse = await aliceAgentmodules.drpc.sendRequest(connectionRecord.id, {
  jsonrpc:'2.0', 
  method:'hello', 
  id:1
})
const response = await listenResponse()
```
After the listener sends the response, the sender will receive a curried function that will resolve to the response

### Listen for DRPC request and send response:
``` typescript
const { request, sendResponse } = await faberAgent.modules.drpc.recvRequest()
await sendResponse({
  jsonrpc:'2.0', 
  result:'Hello', 
  id:1
})
```

### DRPC notification
to send a DRPC notification simply set the `id` to be `null`  
the response will then be an empty object

### DRPC batch requests
DRPC supports sending multiple request objects over a single connection
``` typescript
const batchListen = await aliceAgent.modules.drpc.sendRequest(aliceConnection.id, [
  { jsonrpc: '2.0', method: 'hello', id: 1 },
  { jsonrpc: '2.0', method: 'add', params: [2, 3, 7], id: 2 },
  { jsonrpc: '2.0', method: 'parseFoo', params: { foo: 'bar' }, id: 3 },
  { jsonrpc: '2.0', method: 'error', id: 4 },
])
const batchResponse = await batchListen()
```
the response will be an array of DRPC responses


packages/core/src/modules/drpc/__tests__/drpc-messages.e2e.test.ts
```diff
@@ -0,0 +1,199 @@
...
+  afterEach(async () => {
+    await faberAgent.shutdown()
+    await faberAgent.wallet.delete()
+    await aliceAgent.shutdown()
+    await aliceAgent.wallet.delete()
+  })
+
+  test('Alice and Faber exchange messages', async () => {
+    testLogger.test('Alice sends message to Faber')
+    faberAgent.drpcMessages.createDRPCMethodHandler('hello', async (message) => {
```
amanji, line 58: I may have interpreted the RFC differently, or maybe this is how it's intended to work in `credo-ts`, but in ACA-Py the agent doesn't directly handle the RPC-JSON object (i.e. it doesn't handle the RPC method execution, it simply forwards that through to a controller. The controller is responsible for handling the RPC message accordingly and formulating a response. So, in this case, would an application running `credo-ts` have to register a method handler directly with the agent in order to use DRPC?

I'm quoting this line from the [Key Concepts](https://github.com/hyperledger/aries-rfcs/blob/bfa0bb23bda457a92a2a37a33b396f43ae84eb03/features/0804-didcomm-rpc/README.md#key-concepts) of RFC 0804:

>... The interpretation of the request, how to carry out the request, the content of the response, and the interpretation of the response, are all up to the business logic (controllers) of the participating agents...

amanji:
I may have interpreted the RFC differently, or maybe this is how it's intended to work in `credo-ts`, but in ACA-Py the agent doesn't directly handle the JSON-RPC object (i.e. it doesn't handle the RPC method execution) it simply forwards that through to a controller. The controller is responsible for handling the JSON-RPC accordingly and formulating a response that the agent will convey back. So, in this case, would an application (ex. a wallet) running `credo-ts` have to register a method handler directly with the agent in order to use DRPC?

I'm quoting this line from the [Key Concepts](https://github.com/hyperledger/aries-rfcs/blob/bfa0bb23bda457a92a2a37a33b396f43ae84eb03/features/0804-didcomm-rpc/README.md#key-concepts) of RFC 0804:

> ... The interpretation of the request, how to carry out the request, the content of the response, and the interpretation of the response, are all up to the business logic (controllers) of the participating agents...

Perhaps ACA-Py's implementation is more simplistic though, the agent simply handles the DRPC Request and Response messages and ensures it is valid in accordance with the JSON-RPC specification and the Aries DRPC Protocol.

wadeking98:
Do you mean this line here: `The interpretation of the request, how to carry out the request, the content of the response, and the interpretation of the response, are all up to the business logic (controllers)`

I think it's a bit up to interpretation. From one perspective the agent is kind of doing the RPC execution because the handler function is being executed by the agent... but from another perspective it's the controller that's handling RPC execution since it's the controller's function that's being passed into the agent and used as the RPC handler.  

It seems to me it could be argued either way based on semantics, but ultimately its the business logic that defines the handler and so its the business logic that decides "how the request is interpreted", "how to carry out the request", etc. even though the agent is the one executing the handler.

wadeking98:
Also the methodHandler approach provides some encapsulation which will make the business logic on the controller simpler. I'm open to forwarding all DRPC messages to the controller, but I think this approach makes things a bit tidier IMHO

amanji:
Also I didn't mean to post my first message three times 😅

wadeking98:
Maybe we could get @swcurran 's opinion on this?

packages/core/src/modules/drpc/DRPCMessagesApi.ts
```diff
@@ -0,0 +1,173 @@
...
+      return {
+        jsonrpc: '2.0',
+        id: message.id,
+        error: { code: DRPCErrorCode.INTERNAL_ERROR, message: 'Internal error', data: error.message },
+      }
+    }
+  }
+
+  public async sendDRPCRequest(connectionId: string, request: DRPCRequest): Promise<DRPCResponse> {
+    const messageId = uuid()
```
genaris, line 102: Is there any need of defining `messageId` here? It can be calculated internally in  `createRequestMessage` and use `message.id` in the filter by threadId.

packages/core/src/modules/drpc/DRPCMessagesApi.ts
```diff
@@ -0,0 +1,173 @@
...
+    } catch (error) {
+      return {
+        jsonrpc: '2.0',
+        id: message.id,
+        error: { code: DRPCErrorCode.INTERNAL_ERROR, message: 'Internal error', data: error.message },
+      }
+    }
+  }
+
+  public async sendDRPCRequest(connectionId: string, request: DRPCRequest): Promise<DRPCResponse> {
```
genaris, line 101: Since you are awaiting for a response, IMHO there should be a timeout setting. Otherwise, the execution will block in case the other party does not respond. Maybe it can follow the pattern we do in DiscoverFeaturesApi's `queryFeatures()` ( `awaitDisclosures` setting).

packages/core/src/modules/drpc/DRPCMessagesApi.ts
```diff
@@ -0,0 +1,173 @@
...
+    try {
+      const { message: drpcMessage, record: drpcMessageRecord } = await this.drpcMessageService.createRequestMessage(
+        this.agentContext,
+        request,
+        connection,
+        messageId
+      )
+      await this.sendDRPCMessage(connection, drpcMessage, drpcMessageRecord)
+    } catch {
+      throw new Error('Invalid DRPC Request')
```
genaris, line 113: Maybe we can leave sendDRPCMessage throw its error? Because it can be of different natures (e.g. network error).

packages/core/src/modules/drpc/DRPCMessagesApi.ts
```diff
@@ -0,0 +1,173 @@
...
+    const outboundMessageContext = new OutboundMessageContext(message, {
+      agentContext: this.agentContext,
+      connection,
+      associatedRecord: messageRecord,
+    })
+
+    await this.messageSender.sendMessage(outboundMessageContext)
+  }
+
+  public createDRPCMethodHandler(
```
genaris, line 159:
```suggestion
  public addMethodHandler(
```

genaris: Probably would be nice to add some validation if a handler for a given method is already registered, as only one per method is allowed, right?

packages/core/src/modules/drpc/repository/DRPCMessageRecord.ts
```diff
@@ -0,0 +1,51 @@
...
+export interface DRPCMessageStorageProps {
+  id?: string
+  connectionId: string
+  role: DRPCMessageRole
+  tags?: CustomDRPCMessageTags
+  content: DRPCRequestMessage | DRPCResponseMessage
+}
+
+export class DRPCMessageRecord extends BaseRecord<DefaultBasicMessageTags, CustomDRPCMessageTags> {
+  public content!: DRPCRequestMessage | DRPCResponseMessage
```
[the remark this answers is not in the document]
genaris: In addition, it will be useful to save the `state` (as defined in the RFC). This way you can easily tell if a given request was already answered.

packages/core/src/modules/drpc/DRPCMessagesApi.ts
```diff
@@ -0,0 +1,173 @@
...
+        }
+        const responses = await Promise.all(futures)
+        try {
+          await this.sendDRPCResponse(
+            drpcMessageRecord.connectionId,
+            message.id,
+            responses.length === 1 ? responses[0] : responses
+          )
+        } catch {
+          await this.sendDRPCResponse(drpcMessageRecord.connectionId, message.id, {
```
genaris, line 59: Will it suffice to send a single item in this error response even if there are multiple requests in the same message?

packages/core/src/modules/drpc/DRPCMessagesApi.ts
```diff
@@ -0,0 +1,173 @@
...
+    agentContext: AgentContext
+  ) {
+    this.drpcMessageService = drpcMessageService
+    this.messageSender = messageSender
+    this.connectionService = connectionService
+    this.agentContext = agentContext
+    this.registerMessageHandlers(messageHandlerRegistry)
+
+    // listen for incoming drpc requests and forward them to appropriate handlers
+    this.drpcMessageService.createMessageListener(async ({ message, drpcMessageRecord }) => {
```
genaris, line 44: For this particular case I don't see why it is needed to add a listener. It can be handled directly within `DrpcMessageService` in a `processRequest` method (provided drpcMethodHandlers are managed or accesible by it).

There you can also do other general checks, like the existence of a record with the given threadId (as stated in another comment) and verify that every request has an id in the case of a batch request.

packages/core/src/modules/drpc/services/DRPCMessageService.ts
```diff
@@ -0,0 +1,122 @@
...
+  private eventEmitter: EventEmitter
+
+  public constructor(drpcMessageRepository: DRPCMessageRepository, eventEmitter: EventEmitter) {
+    this.drpcMessageRepository = drpcMessageRepository
+    this.eventEmitter = eventEmitter
+  }
+
+  public async createRequestMessage(
+    agentContext: AgentContext,
+    message: DRPCRequest,
```
genaris, line 27: In case of a batch request are we validating or automatically adding an id for each entry?

packages/core/src/modules/drpc/messages/DRPCMessage.ts
```diff
@@ -0,0 +1,93 @@
...
+  result?: any
+  error?: DRPCResponseError
+  id: string | number | null
+}
+
+export class DRPCRequestMessage extends AgentMessage {
+  public readonly allowDidSovPrefix = true
+
+  /**
+   * Create new BasicMessage instance.
```
genaris, line 47: Copy-paste issue 😛

packages/core/src/modules/drpc/messages/DRPCMessage.ts
```diff
@@ -0,0 +1,93 @@
...
+  @Expose({ name: 'request' })
+  @IsValidDRPCRequest()
+  public request!: DRPCRequest
+}
+
+export class DRPCResponseMessage extends AgentMessage {
+  public readonly allowDidSovPrefix = true
+
+  /**
+   * Create new BasicMessage instance.
```
genaris, line 73: 😄

genaris (changes requested):
Thanks @wadeking98 ! I'm pretty excited about this new protocol and having it supported in Credo.

I've left a good amount of comments in the code but in general I think it will be important to put all this stuff in a separate module instead of adding it to core. It will need a bit of overhead but should not be hard if you take `action-menu` or `question-answer` as a base.

Other than that, I think the API is quite simple and I find interesting that you manage everything within the framework. It can be also a good idea to leave the possibility of handling methods from the controller side (e.g. a config setting that does not automatically respond according to the registered method handlers) but of course that can be left for a further PR in case is proven to be useful in real-world usage.

BTW I don't know what others think but I think 'DRPC' in class names should be changed to 'Dprc' to be more aligned with our coding style (e.g. classes with `Did*` or `OpenId4Vc*`

packages/core/src/modules/drpc/DRPCMessagesApi.ts
```diff
@@ -0,0 +1,173 @@
...
+import { uuid } from '../../utils/uuid'
+import { ConnectionService } from '../connections'
+
+import { DRPCMessageRole } from './DRPCMessageRole'
+import { DRPCMessageHandler } from './handlers'
+import { DRPCErrorCode, DRPCRequestMessage, DRPCResponseMessage } from './messages'
+import { DRPCMessageService } from './services'
+
+@injectable()
+export class DRPCMessagesApi {
```
TimoGlastra, line 20:
```suggestion
export class DrpcMessagesApi {
```

To follow conventions in Credo

packages/core/src/modules/drpc/DRPCMessagesApi.ts
TimoGlastra: I'm thinking whether a handler should be registered at all in Credo? 

I think the normal event flow may work well for this:
- receive rpc message, get notified using agent.events.on
- execute code as part of handler, this is not handled by credo
- call api to send response.

packages/core/src/modules/drpc/DRPCMessagesApi.ts
TimoGlastra: Is rpc meant to be request -> response, or also to support async flows? If request/ response i think timeout should suffice (which we can make configurable on the module and request method). Otherwise async (response should be catched using events) may work better.

packages/core/src/agent/BaseAgent.ts
```diff
@@ -118,6 +121,7 @@ export abstract class BaseAgent<AgentModules extends ModulesMap = EmptyModuleMap
       this.mediationRecipient,
       this.messagePickup,
       this.basicMessages,
+      this.drpcMessages,
```
TimoGlastra, line 124: I'm not sure this protocol should be part of core. Do you think it can be moved to a separate module?

packages/core/src/modules/drpc/DRPCMessagesApi.ts
TimoGlastra: Especially with multi tenancy the add handler may not be the best fit (also think about multi-instance deployments). So i'd prefer to have just the message sending/receiving handled by credo and the actual procedure be executed based on events

packages/core/src/modules/drpc/DRPCMessagesApi.ts
amanji: The way it is implemented in ACA-Py and described in the RFC (at least how I've interpreted) it is async.

packages/core/src/modules/drpc/DRPCMessagesApi.ts
amanji: I didn't even consider the aspect of multi tenancy so that's also a very good point :).

packages/core/src/modules/drpc/DRPCMessagesApi.ts
genaris: I think that even the "Messages" word can be skipped (e.g. DrpcApi, DrpcService, etc.)

packages/core/src/modules/drpc/DRPCMessageRole.ts
```diff
@@ -0,0 +1,4 @@
+export enum DRPCMessageRole {
+  Sender = 'sender',
```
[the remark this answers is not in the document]
wadeking98: This is a little different than client/server. The role applies to the message record itself, when the client sends a drpc request then the record will be in the `sender` role on the client's end and the `reciever` role on the server's end. Vice versa for the response sent by the server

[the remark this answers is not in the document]
TimoGlastra: We also have a DidCommMessageRecord available that you can use to store/retrieve exchanged didcomm messages. We usually do that over adding the message to the exchange record itself, and that also has a role property (sender / receiver)

packages/core/src/modules/drpc/services/DRPCMessageService.ts
```diff
@@ -0,0 +1,122 @@
...
+        message,
+        drpcMessageRecord,
+        removeListener: () => this.eventEmitter.off(DRPCMessageEventTypes.DRPCMessageStateChanged, listener),
+      })
+    }
+    this.eventEmitter.on(DRPCMessageEventTypes.DRPCMessageStateChanged, listener)
+  }
+
+  /**
+   * @todo use connection from message context
```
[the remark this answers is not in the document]
wadeking98: ah, that was an old comment that got copied over accidentally from basic massages

packages/core/src/modules/drpc/services/DRPCMessageService.ts
wadeking98: in the agent we don't really care about the id in each DRPC request/response object. those are more for business logic to consume. So we leave it up to the business logic to validate/add ids for the DRPC messages

wadeking98:
@genaris @TimoGlastra The validate CI Stage is failing after updating the module README for some reason but other than that it's ready for re-review

wadeking98:
> Did you run `yarn format` locally?

That did the trick, should be all ready now

wadeking98:
@TimoGlastra @genaris would you mind doing a quick re-review when you have a second?

packages/core/src/modules/connections/ConnectionsApi.ts
```diff
@@ -43,7 +43,6 @@ export interface SendPingOptions {
   responseRequested?: boolean
   withReturnRouting?: boolean
 }
```
berendsliedrecht, line 45: Can this newline be added back?

packages/drpc/CHANGELOG.md
```diff
@@ -0,0 +1,98 @@
+# Change Log
```
berendsliedrecht, line 1: This can be removed IIRC.

packages/drpc/src/DrpcModule.ts
```diff
@@ -0,0 +1,31 @@
...
+import { DrpcMessageRepository } from './repository'
+import { DrpcService } from './services'
+
+export class DrpcModule implements Module {
+  public readonly api = DrpcApi
+
+  /**
+   * Registers the dependencies of the drpc message module on the dependency manager.
+   */
+  public register(dependencyManager: DependencyManager, featureRegistry: FeatureRegistry) {
```
berendsliedrecht, line 16: WOuld be good to add the `experimental` log in the beginning. And we can remove it after some public use (like at 0.6.0 or something).

packages/drpc/src/DrpcRole.ts
```diff
@@ -0,0 +1,4 @@
+export enum DrpcRole {
```
berendsliedrecht, line 1: Can these be relocated under `drpc/src/models`? Slightly more consistent with the rest.

packages/drpc/src/DrpcState.ts
```diff
@@ -0,0 +1,5 @@
+export enum DrpcState {
```
berendsliedrecht, line 1: Relocation also applies here.

packages/drpc/src/messages/DrpcRequest.ts
```diff
@@ -0,0 +1,73 @@
+import type { ValidationArguments, ValidationOptions } from 'class-validator'
+
+import { IsValidMessageType, parseMessageType, AgentMessage } from '@credo-ts/core'
+import { Expose } from 'class-transformer'
+import { ValidateBy, ValidationError, buildMessage } from 'class-validator'
+
+export function IsValidDrpcRequest(validationOptions?: ValidationOptions): PropertyDecorator {
+  return function (target: any, propertyKey: string | symbol) {
```
berendsliedrecht, line 8: Is this a validator requirement that `target` is `any` or can we use `unknown` as well?

berendsliedrecht: Would also be good to move the validation functions to another file and just keep this the request class.

packages/drpc/src/messages/DrpcResponse.ts
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  return false
+}
+
+function isValidDrpcResponseError(error: any): boolean {
+  return typeof error === 'object' && error !== null && 'code' in error && 'message' in error
+}
+
+export enum DrpcErrorCode {
```
berendsliedrecht, line 69: This would also be better in a separate file under `models`.

packages/drpc/src/messages/DrpcResponse.ts
```diff
@@ -0,0 +1,112 @@
...
+
+export interface DrpcResponseObject {
+  jsonrpc: string
+  result?: any
+  error?: DrpcResponseError
+  id: string | number | null
+}
+
+export class DrpcResponseMessage extends AgentMessage {
+  public readonly allowDidSovPrefix = true
```
berendsliedrecht, line 94: Is this used anywhere?

berendsliedrecht (approved):
Minor notes, mainly for consistency.

packages/drpc/src/DrpcModule.ts
genaris: I think it is important to add this mark, especially because the DRPC protocol itself is not yet merged into Aries RFCs repo, so it's likely to have some updates at some point.

packages/drpc/src/DrpcState.ts
```diff
@@ -0,0 +1,5 @@
...
+  RequestSent = 'request-sent',
+  RequestRecieved = 'request-received',
```
genaris, line 3:
```suggestion
  RequestReceived = 'request-received',
```

packages/drpc/src/services/DrpcService.ts
```diff
@@ -0,0 +1,178 @@
...
+      const { drpcMessageRecord } = event.payload
+      await callback({
+        drpcMessageRecord,
+        removeListener: () => this.eventEmitter.off(DrpcResponseEventTypes.DrpcResponseStateChanged, listener),
+      })
+    }
+    this.eventEmitter.on(DrpcResponseEventTypes.DrpcResponseStateChanged, listener)
+  }
+
+  public async recieveResponse(messageContext: InboundMessageContext<DrpcResponseMessage>) {
```
genaris, line 80:
```suggestion
  public async receiveResponse(messageContext: InboundMessageContext<DrpcResponseMessage>) {
```

packages/drpc/src/repository/DrpcMessageRecord.ts
```diff
@@ -0,0 +1,75 @@
...
+}
+
+export type DrpcMessageTags = RecordTags<DrpcMessageRecord>
+
+export interface DrpcMessageStorageProps {
+  id?: string
+  connectionId: string
+  role: DrpcRole
+  tags?: CustomDrpcMessageTags
+  message: DrpcRequest | DrpcResponse
```
genaris, line 21: IMHO it would be clearer if we'd store separate `request?: DrpcRequest` and `response?: DrpcResponse`

The term `message` is mostly used (in credo context) for DIDComm messages, so using that here for another kind of object can be confusing.

packages/drpc/src/services/DrpcService.ts
```diff
@@ -0,0 +1,178 @@
...
+export class DrpcService {
+  private drpcMessageRepository: DrpcMessageRepository
+  private eventEmitter: EventEmitter
+
+  public constructor(drpcMessageRepository: DrpcMessageRepository, eventEmitter: EventEmitter) {
+    this.drpcMessageRepository = drpcMessageRepository
+    this.eventEmitter = eventEmitter
+  }
+
+  public async createRequestMessage(agentContext: AgentContext, message: DrpcRequest, connectionId: string) {
```
genaris, line 25:
```suggestion
  public async createRequestMessage(agentContext: AgentContext, request: DrpcRequest, connectionId: string) {
```

packages/drpc/src/services/DrpcService.ts
```diff
@@ -0,0 +1,178 @@
...
+      role: DrpcRole.Client,
+    })
+
+    await this.drpcMessageRepository.save(agentContext, drpcMessageRecord)
+    this.emitStateChangedEvent(agentContext, drpcMessageRecord)
+
+    return { message: drpcMessage, record: drpcMessageRecord }
+  }
+
+  public async createResponseMessage(agentContext: AgentContext, message: DrpcResponse, drpcRecord: DrpcMessageRecord) {
```
genaris, line 42:
```suggestion
  public async createResponseMessage(agentContext: AgentContext, response: DrpcResponse, drpcRecord: DrpcMessageRecord) {
```

packages/drpc/tests/drpc-messages.e2e.test.ts
```diff
@@ -0,0 +1,282 @@
...
+  const responseListener = await sender.modules.drpc.sendRequest(connectionRecord.id, message)
+  const { request, sendResponse } = await receiver.modules.drpc.recvRequest()
+  const result = await handleMessageOrError(messageHandlers, request as DrpcRequestObject)
+  await sendResponse(result as DrpcResponseObject)
+
+  const helloRecord = await responseListener()
+  return helloRecord as DrpcResponseObject
+}
+
+const sendAndRecieveBatch = async (
```
genaris, line 62:
```suggestion
const sendAndReceiveBatch = async (
```

packages/drpc/src/DrpcApi.ts
```diff
@@ -0,0 +1,152 @@
...
+
+      this.drpcMessageService.createResponseListener(listener)
+    })
+  }
+
+  /**
+   * Listen for a request and returns the request object and a function to send the response
+   * @returns the request object and a function to send the response
+   */
+  public async recvRequest(): Promise<{
```
genaris, line 85: This naming reminds me socket's recv() call, so I guess the intention is to be used like that. In this case, to also mimic that function, I think it will be nice to make it possible to define an optional timeout so it's not forced to block

packages/drpc/src/DrpcApi.ts
```diff
@@ -0,0 +1,152 @@
...
+  public async sendRequest(connectionId: string, request: DrpcRequest): Promise<() => Promise<DrpcResponse>> {
+    const connection = await this.connectionService.getById(this.agentContext, connectionId)
+    const { message: drpcMessage, record: drpcMessageRecord } = await this.drpcMessageService.createRequestMessage(
+      this.agentContext,
+      request,
+      connection.id
+    )
+    const messageId = drpcMessage.id
+    await this.sendMessage(connection, drpcMessage, drpcMessageRecord)
+    return this.recvResponse.bind(this, messageId)
```
genaris, line 53: This is a great pattern! Seems a very handy for the user.

packages/drpc/src/DrpcApi.ts
```diff
@@ -0,0 +1,152 @@
...
+    })
+  }
+
+  /**
+   * Sends a drpc response to a connection
+   * @param connectionId the connection id to use
+   * @param threadId the thread id to respond to
+   * @param response the drpc response object to send
+   */
+  private async sendResponse(connectionId: string, threadId: string, response: DrpcResponse): Promise<void> {
```
genaris, line 117: It might be a good idea to use an options object here, to avoid unnecessary breaking changes to the API in the future. Something like:

 ```suggestion
  private async sendResponse(options: { connectionId: string, threadId: string, response: DrpcResponse }): Promise<void> {
```

**Review by genaris — CHANGES_REQUESTED** [review]
I have a few more comments but overall looks great!

File: packages/drpc/src/messages/DrpcRequest.ts (hunk 21 above)
Comment (wadeking98) [reply packages/drpc/src/messages/DrpcRequest.ts:8]: it has to be any or Object

File: packages/drpc/src/messages/DrpcResponse.ts (hunk 23 above)
Comment (wadeking98) [reply packages/drpc/src/messages/DrpcResponse.ts:94]: It was part of the subclass AgentMessage but I don't think it's actually needed

File: packages/drpc/src/messages/DrpcRequest.ts (hunk 21 above)
Comment (wadeking98) [reply packages/drpc/src/messages/DrpcRequest.ts:8]: I think it expects the target to be type `Object` but having `Object` as a type creates a linting error

File: packages/drpc/src/DrpcApi.ts [hunk 33]
```diff
@@ -0,0 +1,184 @@
...
+      }
+    })
+  }
+
+  /**
+   * Listen for a request and returns the request object and a function to send the response
+   * @param timeout the time in seconds to wait for a request
+   * @returns the request object and a function to send the response
+   */
+  public async recvRequest(timeout?: number): Promise<
```
genaris, line 96:
```suggestion
  public async recvRequest(timeoutMs?: number): Promise<
```

And I promise this is my last comment 😄.

Just for consistency with other timeouts within the framework: it will be good if we can express it in milliseconds

wadeking98: no worries lol

genaris (approved):

genaris (approved):

PR #1755, opened 2024-02-12
feat(mesage-pickup): option for awaiting completion

genaris:
Add a blocking mode for `MessagePickupApi.pickupMessages()`, where the agent will wait until all messages have been effectively retrieved and therefore the message pickup process is completed.

This tries to solve two main situations:

1) Retrieve queued messages BEFORE setting live mode in agents configured with `MediatorPickupStrategy.PickUpV2LiveMode`. This allows supporting the case where the mediator does not support `live-delivery-change` message (which is the case of AFJ <= 0.4.2 and AFAIK ACA-Py's Pick-up v2 plug-in). In such case, mediator will send a problem report that will be logged in the recipient but should not prevent the pick-up flow from being started
2) Facilitate the creation of a process dedicated to message pickup only. For instance, this can be used in mobile agents that are initiated from a background notification handler.

I did not take too much effort on the tests yet as I see there are some improvements in #1754 so maybe I can incorporate them once it's reviewed and merged.

### DIFF
--- a/packages/core/src/modules/message-pickup/protocol/v1/V1MessagePickupProtocol.ts
+++ b/packages/core/src/modules/message-pickup/protocol/v1/V1MessagePickupProtocol.ts
@@ -1,8 +1,10 @@
 import type { AgentContext } from '../../../../agent'
 import type { AgentMessage } from '../../../../agent/AgentMessage'
+import type { AgentMessageReceivedEvent } from '../../../../agent/Events'
 import type { FeatureRegistry } from '../../../../agent/FeatureRegistry'
 import type { InboundMessageContext } from '../../../../agent/models/InboundMessageContext'
 import type { DependencyManager } from '../../../../plugins'
+import type { MessagePickupCompletedEvent } from '../../MessagePickupEvents'
 import type { MessagePickupRepository } from '../../storage/MessagePickupRepository'
 import type {
   DeliverMessagesProtocolOptions,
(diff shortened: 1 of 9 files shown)


packages/core/src/modules/message-pickup/protocol/v1/V1MessagePickupProtocol.ts
```diff
@@ -127,4 +133,40 @@ export class V1MessagePickupProtocol extends BaseMessagePickupProtocol {
...
+          contextCorrelationId: messageContext.agentContext.contextCorrelationId,
+        },
+      })
+    })
+
+    // A Batch message without messages at all means that we are done with the
+    // message pickup process (Note: this is not optimal since we'll always doing an extra
+    // Batch Pickup. However, it is safer and should be faster than waiting an entire loop
+    // interval to retrieve more messages)
+    if (messages.length === 0) {
```
TimoGlastra, line 159: If messages.length is smaller than the batch size we can also assume the message pickup is complete right?

TimoGlastra (approved):
Thansk for looking into this! How would it exactly work when you want to use a 0.4.2 mediator? pickup v2 and it will automatically use implicit afterwards?

packages/core/src/modules/message-pickup/protocol/v1/V1MessagePickupProtocol.ts
genaris: Yes! The issue is that I cannot be 100% sure of the batch size requested in the originating batch message (I can assume it from agent config but user can actually override it)

TimoGlastra:
Okay makes sense! Maybe we should deprecate implicit mode in 0.6 and remove in 0.7 (or deprecate in 0.5.x and remove in 0.6).

Maybe we should also deprecate pickup v1 then, as there's not really a lot of adoption, and also not really any reason not to use v2

PR #1774, opened 2024-02-20
fix(anoncreds): check whether the tails location is a URL

berendsliedrecht:
When you register a tailsfile, but you want to have an external endorser sign this transaction, this bug will happen. When creating the TXN, it will first replace the `tailsFileLocation` from a file path to a URL and upload it. After the endorser signed the TXN, it will be submitted again, running through the same code, and the `tailsFileLocation` will be replaced from a `URL` to another `URL` and uploaded again.

Best solution I could come up with to check wether the tails file is already uploaded.

TBH, it would be far better to have a different API for the `internal` vs `external`. Right now there is quite some code (like storing the revocation registry definition private parts) and now this that should **only** happen in either.

Also, reusing the `tailsFileLocation` for the URL and the file path does not really have my preference. Maybe in `0.6` we could introduce a `tailsFileFsLocation` and `tailsFileRemoteLocation`? Maybe this is a bit too overengineerd, but it was the reason why there was this bug in credo.

### DIFF
--- a/packages/anoncreds/src/AnonCredsApi.ts
+++ b/packages/anoncreds/src/AnonCredsApi.ts
@@ -21,7 +21,7 @@ import type {
 import type { Extensible } from './services/registry/base'
 import type { SimpleQuery } from '@credo-ts/core'
 
-import { AgentContext, inject, injectable } from '@credo-ts/core'
+import { AgentContext, inject, injectable, isUri } from '@credo-ts/core'
 
 import { AnonCredsModuleConfig } from './AnonCredsModuleConfig'
 import { AnonCredsStoreRecordError } from './error'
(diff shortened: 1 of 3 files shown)


packages/anoncreds/src/AnonCredsApi.ts
```diff
@@ -398,13 +398,20 @@ export class AnonCredsApi {
...
-        revocationRegistryDefinition,
-      })
-      revocationRegistryDefinition.value.tailsLocation = tailsFileUrl
+      // A check is done whether the tails location is a URL.
+      // This state may occur when the `registerRevocationRegistryDefinition` is called
+      // once to create the TXN and once to actually submit it.
+      // When the TXN is created, it will be a file path and this code will be called.
+      // When the submission is done, it already has a URL as the location and does not
+      // need to be reuploaded
+      if (!isUri(localTailsLocation)) {
```
genaris, line 409: I'm confused here. This `localTailsLocation` comes from the result of `this.anonCredsIssuerService.createRevocationRegistryDefinition`, which will create the object by calling anoncreds-rs `RevocationRegistryDefinition.create()` method. How is it possible for it to return an URL?

genaris (commented):
I agree this `tailsLocation` thing is quite confusing. Probably we'll need to rethink about it not only in credo but also in anoncreds-rs.

I'm not sure I properly understood the flow you are describing, but I feel that the URI check might be more appropriate to be added directly into the `tailsFileService` implementation. But surely I'm missing something, please elaborate 😄.

packages/anoncreds/src/AnonCredsApi.ts
berendsliedrecht: Hmmm yeah you are correct :'). I interpreted the code a bit wrong and assumed it would just use the passed arguments, but it creates it again (which is an issue as well).

I will close this PR and open an issue for this as currently has a bug where signing the transaction by an external endorser (i.e. calling `registerRevocationRegistryDefinition` twice) does not work. The credential will contain invalid revocation data.

PR #1780, opened 2024-02-29
fix(core): query credential and proof records by correct DIDComm role

berendsliedrecht:
- Added an additional requirement to the query to find an agent message. This was added so you can technically offer a credential, and request a proof, to yourself. No changes in the current behavior.


packages/anoncreds/src/protocols/proofs/v1/V1ProofProtocol.ts
```diff
@@ -220,7 +222,7 @@ export class V1ProofProtocol extends BaseProofProtocol implements ProofProtocol<
       await didCommMessageRepository.saveOrUpdateAgentMessage(agentContext, {
         agentMessage: proposalMessage,
         associatedRecordId: proofRecord.id,
-        role: DidCommMessageRole.Sender,
+        role: DidCommMessageRole.Receiver,
```
berendsliedrecht, line 225: I think this one was set incorrectly, a double-check at the review would be good!

TimoGlastra (approved):

PR #1783, opened 2024-03-01
fix: query the record by credential and proof role

berendsliedrecht:
- Missed in the last PR regarding issuance to self. We also have to make sure when finding the credential, or proof, exchange record, that the role is correct otherwise we will find multiple instances with the same connection id and thread id.


packages/core/src/modules/credentials/protocol/v2/__tests__/V2CredentialProtocolCred.test.ts
```diff
@@ -372,6 +374,8 @@ describe('credentialProtocol', () => {
       // then
       expect(credentialRepository.findSingleByQuery).toHaveBeenNthCalledWith(1, agentContext, {
         threadId: 'somethreadid',
+        connectionId: 'issuer',
+        role: undefined,
```
TimoGlastra, line 378: Shouldn't the role be issuer here?

berendsliedrecht: Yeah I messed something up. I opened a new PR.

berendsliedrecht:
continues in #1784

PR #1784, opened 2024-03-01
fix: query the record by credential and proof role

berendsliedrecht:
- Missed in the last PR regarding issuance to self. We also have to make sure when finding the credential, or proof, exchange record, that the role is correct otherwise we will find multiple instances with the same connection id and thread id.
- Fixed some of the deprecated jest functions (just the ones I came across).

### DIFF
--- a/packages/core/src/modules/credentials/protocol/BaseCredentialProtocol.ts
+++ b/packages/core/src/modules/credentials/protocol/BaseCredentialProtocol.ts
@@ -142,12 +142,10 @@ export abstract class BaseCredentialProtocol<CFs extends CredentialFormatService
 
     agentContext.config.logger.debug(`Processing problem report with message id ${credentialProblemReportMessage.id}`)
 
-    const credentialRecord = await this.getByThreadIdConnectionIdAndRole(
-      agentContext,
-      credentialProblemReportMessage.threadId,
-      undefined,
-      connection.id
-    )
+    const credentialRecord = await this.getByProperties(agentContext, {
+      threadId: credentialProblemReportMessage.threadId,
+      connectionId: connection.id,
+    })
 
     // Update record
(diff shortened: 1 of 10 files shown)


packages/core/src/modules/credentials/protocol/BaseCredentialProtocol.ts
```diff
@@ -275,41 +277,49 @@ export abstract class BaseCredentialProtocol<CFs extends CredentialFormatService
...
    * @param connectionId The connection id
+   * @param role The role of the record, i.e. holder or issuer
    * @param threadId The thread id
+   *
    * @throws {RecordNotFoundError} If no record is found
    * @throws {RecordDuplicateError} If multiple records are found
    * @returns The credential record
    */
-  public getByThreadAndConnectionId(
+  public getByThreadIdConnectionIdAndRole(
```
genaris, line 287: I think it is a bit confusing this method to be called `getByThreadIdConnectionIdAndRole` if it's expecting the role _before_ the connectionId. What about swapping the parameters, renaming it or use an options object like we usually do?

berendsliedrecht: Yeah I didn't do an object to not make too many changes. But I think that that would be the correct call (using a bunch of optional parameter causes a bug in this PR earlier and an object would've caught that). 

I do think when it's an object I will rename it to `getByQueryFilter` or something equivalent. This would allow to have more options in the future as well. Also happy to pick another name :).

berendsliedrecht: @genaris I updated it to `getByProperties` and `findByProperties`. This'll make it expandable in the future as well. Is that good?

genaris: Oh, I was just complaining about the order of the parameters, but like this is better of course!

genaris (approved):

PR #1786, opened 2024-03-01
fix: stopped recvRequest from receiving outbound messages

wadeking98:
Previously the drpc method `recvRequest` would receive messages that had just been send by the agent. This is a quick fix to ensure the method on receives incoming requests from other agents


genaris (approved):

packages/drpc/src/DrpcApi.ts
```diff
@@ -109,7 +110,7 @@ export class DrpcApi {
         removeListener: () => void
```
TimoGlastra, line 110: Not related to this PR, but I'm not a big fan on reicvRequest over receiveRequest, that feels more consistent with other methods in Credo (we almost never abbreviate)

PR #1788, opened 2024-03-06
fix: websocket outbound transport

auer-martin:
The previous approach was only possible in Node.
The websocket implementation of RN follows this specification. https://developer.mozilla.org/en-US/docs/Web/API/WebSocket. 
The Promise was therefore never resolved as the callback does not exist.

### DIFF
--- a/packages/core/src/transport/WsOutboundTransport.ts
+++ b/packages/core/src/transport/WsOutboundTransport.ts
@@ -68,18 +68,14 @@ export class WsOutboundTransport implements OutboundTransport {
     const isNewSocket = !this.hasOpenSocket(socketId)
     const socket = await this.resolveSocket({ socketId, endpoint, connectionId })
 
-    return new Promise<void>((resolve, reject) =>
-      socket.send(Buffer.from(JSON.stringify(payload)), (err) => {
-        // If the socket was created for this message and we don't have return routing enabled
-        // We can close the socket as it shouldn't return messages anymore
-        if (isNewSocket && !outboundPackage.responseRequested) {
-          socket.close()
-        }
-
-        if (err) return reject(err)
-        resolve()
-      })
-    )
+    // If the socket was created for this message and we don't have return routing enabled
+    // We can close the socket as it shouldn't return messages anymore
+    // make sure to use the socket in a manner that is compliant with the https://developer.mozilla.org/en-US/docs/Web/API/WebSocket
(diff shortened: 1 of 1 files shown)


packages/core/src/transport/WsOutboundTransport.ts
```diff
@@ -68,18 +68,14 @@ export class WsOutboundTransport implements OutboundTransport {
...
-
-        if (err) return reject(err)
-        resolve()
-      })
-    )
+    // If the socket was created for this message and we don't have return routing enabled
+    // We can close the socket as it shouldn't return messages anymore
+    // make sure to use the socket in a manner that is compliant with the https://developer.mozilla.org/en-US/docs/Web/API/WebSocket
+    // (React Native) and https://github.com/websockets/ws (NodeJs)
+    socket.send(Buffer.from(JSON.stringify(payload)))
```
berendsliedrecht, line 75: We technically could send just a string instead of a buffer with both these interfaces, but it should not matter that much.

berendsliedrecht (approved):

genaris:
Oh that API difference is unfortunate! This error handling was added recently to improve the test suites. Not sure if removing it will affect a lot (because there were other improvements done by @TimoGlastra at the same time), but if it does we can create a `TestWsOutboundTransporter` or so to manage it.

That being said, if WebSocket API differs in each platform, probably we should look at making a dedicated implementation for each (like we do for inbound transports). The good thing about such idea is that the infamous `AgentDependencies` object may become smaller, since we won't need a WebSocketClass: if later on we do the same with HttpOutboundTransport, we can remove fetch and if we feel brave enough to move FileSystem dependency to AnonCreds (where it is actually used, mainly for tails files), the only remaining dependency would be the EventEmitter... Just thinking! 🤔

packages/core/src/transport/WsOutboundTransport.ts
TimoGlastra: Is the method sync in RN? And if we don't provide a callback, is it sync in NodeJS as well?

berendsliedrecht: Haven't checked it, but there is likely an overload to not provide a callback as you don't have to handle the res/err if you don't want. But yeah it would be good to do a test run with this in both environments.

auer-martin: I think this should be fine. It's also the behavior that was there up until 2-3 weeks ago.

PR #1795, opened 2024-03-17
fix(openid4vc): several fixes and improvements

TimoGlastra:
Enable `issuanceMetadata` to be provided during offer creation, to add some context to an issuance session. Can be used to already provide the credential data, store an user id so the data can be fetched at time of issuance, or other metadata.

This was already supported, but just not exposed in the API.

Non breaking and can be released as 0.5.1


packages/core/src/utils/transformers.ts
```diff
@@ -30,6 +30,7 @@ export function MetadataTransformer() {
  */
 export function DateTransformer() {
   return Transform(({ value, type }) => {
+    if (value === undefined) return undefined
```
berendsliedrecht, line 33: Should this not throw an error if the value is required? Not too familiar with class-transformer. But it seems like this could cause bugs, right?

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerService.ts
```diff
@@ -158,6 +164,15 @@ export class OpenId4VcIssuerService {
 
     const issuer = await this.getIssuerByIssuerId(agentContext, options.issuanceSession.issuerId)
 
+    const cNonce = getCNonceFromCredentialRequest(credentialRequest)
+    if (issuanceSession.cNonce !== cNonce) {
+      throw new CredoError('The cNonce in the credential request does not match the cNonce in the issuance session.')
+    }
+
+    if (!issuanceSession.cNonceExpiresAt || Date.now() > issuanceSession.cNonceExpiresAt.getTime()) {
```
berendsliedrecht, line 172: Might be good to separate the `!issuanceSession.cNonceExpiresAt` statement and give a more verbose error message which explains that the `cNonce` was not included, but is required.

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerServiceOptions.ts
```diff
@@ -37,6 +38,14 @@ export interface OpenId4VciCreateCredentialOfferOptions {
...
 
   preAuthorizedCodeFlowConfig: OpenId4VciPreAuthorizedCodeFlowConfig
+
+  /**
+   * Metadata about the issuance, that will be stored in the issuance session record and
+   * passed to the credential request to credential mapper. This can be used to e.g. store an
+   * user identifier so user data can be fetched in the credential mapper, or the actual credential
+   * data.
+   */
+  issuanceMetadata?: Record<string, unknown>
```
berendsliedrecht, line 48: What is the main difference between this and metadata on a record? Would it be possible to type this or give some info about this?

packages/openid4vc/src/openid4vc-issuer/__tests__/openid4vc-issuer.test.ts
```diff
@@ -381,16 +385,23 @@ describe('OpenId4VcIssuer', () => {
...
+          myIssuance: 'metadata',
+        })
+
+        return {
+          format: 'jwt_vc',
+          credential: new W3cCredential({
+            type: openBadgeCredential.types,
+            issuer: new W3cIssuer({ id: issuerDid }),
+            credentialSubject: new W3cCredentialSubject({ id: holderDid }),
+            issuanceDate: w3cDate(Date.now()),
```
berendsliedrecht, line 400: Not related to your PR, but looking at all the `W3cXXX` classes, it might be worth considering to make the `W3cDate` a class as well.

berendsliedrecht (approved):
Couple of minor points, LGTM!

packages/core/src/utils/transformers.ts
TimoGlastra: AFAIK transformation is separate from validation. I think it first transforms and then if isOptional is not provided and you validate it will throw an error

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerServiceOptions.ts
TimoGlastra: Yeah was thinking about it as well, but as it's part if the core flow i was thinking it deserves it's own property.

The type is completely up to the user, so it's not possible to type it further.

Metadata is usually quite hidden, and i see it more for additional/internal extra data. 

This metadata should be easily accessible by the user of the framework to help them in knowing what credential to issue

packages/openid4vc/src/openid4vc-issuer/__tests__/openid4vc-issuer.test.ts
TimoGlastra: Yeah maybe, or we can make it a date instance and do the transformation for the user

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerService.ts
TimoGlastra: Hmm yeah should not really happen this case. They're always set together, mostly to make TS happy here, but i can split them up

packages/openid4vc/src/openid4vc-issuer/__tests__/openid4vc-issuer.test.ts
TimoGlastra: I'm going to address this in a separate PR, as it will require quite some changes across the codebase 😊

PR #1798, opened 2024-03-19
feat: allow ignoring revocation status

auer-martin:
…r request


packages/core/src/modules/dif-presentation-exchange/DifPresentationExchangeService.ts
```diff
@@ -67,10 +67,27 @@ export class DifPresentationExchangeService {
 
   public async getCredentialsForRequest(
     agentContext: AgentContext,
-    presentationDefinition: DifPresentationExchangeDefinition
+    presentationDefinition: DifPresentationExchangeDefinition,
+    options?: {
+      filterByNonRevocationRequirements: boolean
```
TimoGlastra, line 72: I know we used this parameter before, but it's hard to understand the property name for me.

// defaults to false
"ignoreRevocationStatus": true

packages/core/src/modules/dif-presentation-exchange/DifPresentationExchangeService.ts
```diff
@@ -67,10 +67,27 @@ export class DifPresentationExchangeService {
...
+
+    if (options?.filterByNonRevocationRequirements === false) {
+      pd = {
+        ...pd,
+        input_descriptors: (pd.input_descriptors = pd.input_descriptors.map((inputDescriptor) => {
+          return {
+            ...inputDescriptor,
+            constraints: { ...inputDescriptor.constraints, statuses: undefined },
+          }
+        })),
```
TimoGlastra, line 85:
```suggestion
        input_descriptors: pd.input_descriptors.map((inputDescriptor) => {
          return {
            ...inputDescriptor,
            constraints: { ...inputDescriptor.constraints, statuses: undefined },
          }
        }),
```

packages/core/src/modules/proofs/formats/dif-presentation-exchange/DifPresentationExchangeProofFormatService.ts
```diff
@@ -354,27 +356,38 @@ export class DifPresentationExchangeProofFormatService
...
     const ps = this.presentationExchangeService(agentContext)
     const { presentation_definition: presentationDefinition } =
       requestAttachment.getDataAsJson<DifPresentationExchangeRequest>()
 
-    const credentialsForRequest = await ps.getCredentialsForRequest(agentContext, presentationDefinition)
+    // Set default values
+    const { filterByNonRevocationRequirements = true } = proofFormats?.presentationExchange ?? {}
+    const credentialsForRequest = await ps.getCredentialsForRequest(agentContext, presentationDefinition, {
+      filterByNonRevocationRequirements,
+    })
```
TimoGlastra, line 390: Will it work if we make a presentation that does not adhere to non-revocation requirements?

packages/core/src/modules/dif-presentation-exchange/DifPresentationExchangeService.ts
berendsliedrecht:
```suggestion
        input_descriptors: pd.input_descriptors.map((inputDescriptor) => ({
            ...inputDescriptor,
            constraints: { ...inputDescriptor.constraints, statuses: undefined },
          }),
```

PR #1800, opened 2024-03-25
feat: credentials api decline offer report

TimoGlastra:
Adds same functionality available in proofs api, allowing to send a problem report when declining a credential offer. Also adds support for sending a problem report if no connection is available on the exchange (when doing e.g. message attached to oob invitation)

### DIFF
--- a/packages/core/src/modules/credentials/CredentialsApi.ts
+++ b/packages/core/src/modules/credentials/CredentialsApi.ts
@@ -16,6 +16,7 @@ import type {
   SendCredentialProblemReportOptions,
   DeleteCredentialOptions,
   SendRevocationNotificationOptions,
+  DeclineCredentialOfferOptions,
 } from './CredentialsApiOptions'
 import type { CredentialProtocol } from './protocol/CredentialProtocol'
 import type { CredentialFormatsFromProtocols } from './protocol/CredentialProtocolOptions'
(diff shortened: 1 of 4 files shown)


packages/core/src/modules/credentials/CredentialsApi.ts
```diff
@@ -324,12 +325,19 @@ export class CredentialsApi<CPs extends CredentialProtocol[]> implements Credent
...
+    credentialRecordId: string,
+    options?: DeclineCredentialOfferOptions
+  ): Promise<CredentialExchangeRecord> {
     const credentialRecord = await this.getById(credentialRecordId)
     credentialRecord.assertState(CredentialState.OfferReceived)
 
     // with version we can get the Service
     const protocol = this.getProtocol(credentialRecord.protocolVersion)
+    if (options?.sendProblemReport) {
+      await this.sendProblemReport({ credentialRecordId, description: 'Offer declined' })
```
genaris, line 338: I think it would be nice to make it possible to specify the description (and/or the code), as this can be triggered for different reasons (user declined, VDR not supported, timeout, etc) and credential protocols are currently making this very 'open'. This comment of course also applies to the proofs module counterpart.

genaris (approved):
I have a suggestion but of course can wait for a further PR or even make it in next major release if it might represent a breaking change (in case of proofs API)

packages/core/src/modules/credentials/CredentialsApi.ts
TimoGlastra: I was thinking of it as well yes but was lazy 😅

I'll make the changes (for now with extra problemReportDescription param that is only used when you also send a problem report) later we can make it an object

PR #1803, opened 2024-03-27
fix: anoncreds w3c migration metadata

auer-martin:


packages/anoncreds/src/updates/0.4-0.5/anonCredsCredentialRecord.ts
```diff
@@ -96,6 +96,10 @@ async function migrateLegacyToW3cCredential(agentContext: AgentContext, legacyRe
     credential: w3cJsonLdCredential,
   })
 
+  for (const [key, meta] of Object.entries(legacyRecord.metadata.data)) {
+    w3cCredentialRecord.metadata.set(key, meta)
+  }
```
TimoGlastra, line 101: Is there any metadata that can clash? I.e. metadata fields on the W3C record? Or do we set those after so it should be good?

Or are there any metadata fields that were on anoncreds record, but not needed anymore on the w3c record (the ones we created with `_anoncreds` and `_internal` prefixes?

auer-martin: the new w3c-anoncreds metadata is set afterward, so that should not be a problem.

I did a quick check for metadata but could not find anything.

TimoGlastra (approved):

PR #1805, opened 2024-03-27
fix: remove strict w3c subjectId uri validation

auer-martin:


TimoGlastra (approved):
We should probably add some extra checks in other places, but for now this is great a good fix 👍

PR #1808, opened 2024-03-28
fix(cheqd): do not crash agent if cheqd down

TimoGlastra:
Adds some error handling to make sure the agent initialization does not fail if a cheqd network is down. As this prevents wallets and servers from starting.

Also updates test to not rely on cheqd testnet, but rather a local network so we're not dependant on external services

### DIFF
--- a/docker-compose.arm.yml
+++ b/docker-compose.arm.yml
@@ -30,3 +30,19 @@ services:
     platform: linux/amd64
     ports:
       - '26657:26657'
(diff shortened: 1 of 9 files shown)


berendsliedrecht (approved):
That's all that is required?

TimoGlastra:
oh damn forgot to push the important part 🤦

berendsliedrecht (approved):

PR #1817, opened 2024-04-03
fix: update cheqd to 2.4.2

TimoGlastra:
This updates the cheqd SDK to a new release that is compatible with node version 2 (used by cheqd testnet and soon mainnet).

~~Still need to figure out the React Native changes so draft for now~~

For react native I will update the docs, but we can already merge this PR.

Also pinned the Node.JS version as it's suddenly broken on newer minor versions due to ffi-napi ( https://github.com/nodejs/node/issues/52240)


berendsliedrecht (approved):
Can't wait to not use ffi-napi anymore...

PR #1819, opened 2024-04-04
feat: query connections by theirLabel and alias

genaris:
This was mainly created to fix https://github.com/openwallet-foundation/credo-ts-ext/issues/267, as it might be useful to query connections by these parameters.

The bad thing about it is that previously created connections won't work with these query parameters, unless we do a migration script (which of course should be reserved to a major version).


GHkrishna:
Any specific reason why we have not merged this?

It would be better to sort new connections than to not be able sort any.

TimoGlastra:
The reason we didn't merge it, as adding it doesn't solve it for already stored records. We can probably merge it now, and do the migration later on?

GHkrishna:
Okay, that seems a bit far but makes sense to add it in 0.6 :)

genaris:
Labeled to 0.6. Need to add a migration script first

packages/didcomm/src/DidCommModule.ts
```diff
@@ -118,5 +119,10 @@ export class DidCommModule implements Module {
       toVersion: '0.5',
       doUpdate: updateV0_4ToV0_5,
     },
+    {
```
TimoGlastra: One thing i was thinking of is that I'd like migrations to not be required for one major upgrade of Credo. So that 0.6 can run on 0.5 and 0.6 storage. Esp. since we don't have any important migrations in this release (yet). 

I'll see if i can implement something that allows the past two versions before locking the agent instead of just the last. We just need to make sure the agent logic is not relying on the last version storage but also can deal with the new storage. 


In e.g. openid4vc module i have been doing something like this:
https://github.com/openwallet-foundation/credo-ts/blob/321b2e8cd7dbf6a4a1321757ae4b7b2167ee2f63/packages/openid4vc/src/openid4vc-issuer/repository/OpenId4VcIssuerRecord.ts#L106-L121

Which supports both formats, and I will then write a migration script as part of the 0.6 release. But credo can still work with the 0.5 structure.

TimoGlastra:
We should make sure these tags are also present in the drizzle storage

PR #1827, opened 2024-04-10
fix: oid4vp can be used separate from idtoken

TimoGlastra:
The current implementation always requires ID Token to be used with VP Token. That is not required, and was a misunderstanding on my side.

This PR addresses it but it needs https://github.com/Sphereon-Opensource/SIOP-OID4VP/pull/76 in SIOP-OID4VP to be merged and released first


TimoGlastra:
I've temporarily applied a fork of SIOP-OID4VP as dependency. We should wait with merging until the changes are released so keeping in draft until then

https://github.com/animo/SIOP-OID4VP#release

berendsliedrecht (approved):

PR #1829, opened 2024-04-11
feat: openid4vc issued state per credential

auer-martin:


packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerService.ts
```diff
@@ -520,6 +521,19 @@ export class OpenId4VcIssuerService {
...
+      if (credentialHasAlreadyBeenIssued) {
+        throw new CredoError(`The requested credential with id '${requestedCredentialId}' has already been issued.`)
+      }
+
+      const updatedIssuanceSession = await this.openId4VcIssuanceSessionRepository.getById(
+        agentContext,
+        issuanceSession.id
+      )
+      updatedIssuanceSession.issuedCredentials.push(requestedCredentialId)
+      await this.openId4VcIssuanceSessionRepository.update(agentContext, updatedIssuanceSession)
```
TimoGlastra, line 535: A few notes on this:
- I think we should do the update after the mapper has been called. What if the mapper errors, then we should not update the state already
- We currently support draft 11 which in the request doesn't include the offered id specifically, but we match based on the types. this means we can have multiple matches and picking the first one is potentially not the one that the mapper will issue. One way to possible solve this is to add a new required return param `credentialSupportedId` to indicate the credential that you issued. Then we can after include that in the record (if it matches one from the offer list). We could make it optional and take the first one if there's one entry, and throw an error otherwise, or we just make it required.
- I think we should maybe filter `offeredCredentialsMatchingRequest` based on the already issued ids?
- How does this deal with an `id` being offered multiple times in an offer? I think we should disallow that for now (add a check to createOffer method that all entries in the array must be unique)

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuanceSessionState.ts
```diff
@@ -4,6 +4,7 @@ export enum OpenId4VcIssuanceSessionState {
   AccessTokenRequested = 'AccessTokenRequested',
   AccessTokenCreated = 'AccessTokenCreated',
   CredentialRequestReceived = 'CredentialRequestReceived',
-  CredentialIssued = 'CredentialIssued',
+  CredentialsPartiallyIssued = 'CredentialsPartiallyIssued',
```
TimoGlastra, line 7: We need to somehow handle already stored records. Maybe in the session record we can do an `@Transform` and if the value is `CredentialIssued` we update it to `Completed` for now? Or maybe better to `Error`?

TimoGlastra (commented):
Nice 👍

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuanceSessionState.ts
auer-martin: not really sure how you to update the record itself

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerService.ts
```diff
@@ -177,12 +184,42 @@ export class OpenId4VcIssuerService {
     }
 
     const vcIssuer = this.getVcIssuer(agentContext, issuer)
+    const issuerMetadata = this.getIssuerMetadata(agentContext, issuer)
+    const holderBinding = await this.getHolderBindingFromRequest(credentialRequest as OpenId4VciCredentialRequest)
```
TimoGlastra, line 188: At this stage this binding hasn't been verified yet, I was thinking we do the check before we call `mapper` , but stilli in the callback we pass to the sphereon lib. That way we know the request etc.. is already verified. But maybe there's something preventing that?

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerServiceOptions.ts
```diff
@@ -101,7 +101,10 @@ export type OpenId4VciCredentialRequestToCredentialMapper = (options: {
   credentialsSupported: OpenId4VciCredentialSupported[]
 }) => Promise<OpenId4VciSignCredential> | OpenId4VciSignCredential
 
-export type OpenId4VciSignCredential = OpenId4VciSignSdJwtCredential | OpenId4VciSignW3cCredential
+export type OpenId4VciSignCredential = { credentialSupportedId: string } & (
```
TimoGlastra, line 104: Now if you import the `OpenId4VciSignSdJwtCredential` it doesn't include the `credentialSupportedId`. I think either we need a base that `OpenId4VciSignSdJwtCredential` and `OpenId4VciSignW3cCredential` extend from, or we should do a bit of duplication

packages/openid4vc/src/openid4vc-issuer/__tests__/openid4vc-issuer.test.ts
```diff
@@ -566,7 +569,7 @@ describe('OpenId4VcIssuer', () => {
           throw new Error('Not implemented')
         },
       })
-    ).rejects.toThrow('No offered credentials match the credential request.')
+    ).rejects.toThrow('Not implemented')
```
TimoGlastra, line 572: This is not what we're testing here, why is the error changed?

auer-martin: because the order changed when doing the mapping before the callback. now the mapping fails

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerService.ts
TimoGlastra: In the `getCredentialDataSupplier` we have access to the `issuanceSessionRecord` right? 

So everything you now moved out of the `getCredentialDataSupplier` ca be in the `getCredentialDataSupplier`?

packages/openid4vc/src/openid4vc-issuer/__tests__/openid4vc-issuer.test.ts
TimoGlastra: but then we're breaking the flow, it should error when there's no credentials matching the credential request, not call the mapper.

TimoGlastra (approved):

PR #1832, opened 2024-04-12
feat(tenants): return value from withTenatnAgent

TimoGlastra:
Has been boterhing me a lot, as it makes the function a lot more complex to use and thus I used `getTenantAgent`, but this way it's easer to use `withTenantAgent,` which always makes sure the session is closed


berendsliedrecht (approved):

sudddy:
Great! About to raise a issue on this and saw this fixed. Thanks @TimoGlastra

PR #1833, opened 2024-04-13
fix(anoncreds): unqualified revocation registry processing

genaris:
I ran into issues issues while receiving legacy Indy credentials with revocation support, e.g. receiving the following error when using [BC Digital Trust Showcase](https://digital.gov.bc.ca/digital-trust/showcase/dashboard):

> CredoError: QEquAHkM35w4XVT3Ku5yat:4:QEquAHkM35w4XVT3Ku5yat:3:CL:567943:student_card:CL_ACCUM:dd6cecd9-a71b-4af4-a8f6-26a3fee2e073 is not a valid did:indy did\n    at parseIndyDid at getUnQualifiedDidIndyDid at getW3cRecordAnonCredsTags

I found that record metadata was populated with `revocationRegistryId` from the W3C credential object, which in this case would be an unqualified identifier. As `getUnQualifiedDidIndyDid` expected a qualified identifier, this fails. In addition, the utility method `getQualifiedDidIndyDid` was not taking into account credentialDefinitionTag while generating the identifier, resulting also in an error.

In this PR this is fixed by explicitly specifying the `revocationRegistryId` when calling `AnonCredsRsHolderService.storeW3cCredential` rather than taking it from the W3C credential. I'm not sure if this is the most elegant solution though, so any suggestion on that will be more than appreciated.

Another a bit controversial thing I changed here is to delay the `w3cCredentialService.storeCredential()` call to the last minute, in order to prevent any possible error to be thrown that could leave a `W3cCredentialRecord` without tags and metadata.

### DIFF
--- a/packages/anoncreds/src/anoncreds-rs/AnonCredsRsHolderService.ts
+++ b/packages/anoncreds/src/anoncreds-rs/AnonCredsRsHolderService.ts
@@ -365,10 +365,18 @@ export class AnonCredsRsHolderService implements AnonCredsHolderService {
       schema: AnonCredsSchema
       credentialDefinition: AnonCredsCredentialDefinition
       revocationRegistryDefinition?: AnonCredsRevocationRegistryDefinition
+      revocationRegistryId?: string
       credentialRequestMetadata: AnonCredsCredentialRequestMetadata
     }
   ) {
-    const { credential, credentialRequestMetadata, schema, credentialDefinition, credentialDefinitionId } = options
+    const {
+      credential,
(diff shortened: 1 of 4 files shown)


TimoGlastra (approved):
Great fixes 👍

TimoGlastra:
@auer-martin this might be related to https://github.com/openwallet-foundation/bifold-wallet/pull/1115#issuecomment-2062595023?

auer-martin:
Yes, that probably is the issue 👍🏻

PR #1836, opened 2024-04-20
fix: persist leading zero

sairanjit:
- This fixes the removal of the leading zeroes

### DIFF
--- a/packages/anoncreds/src/utils/credential.ts
+++ b/packages/anoncreds/src/utils/credential.ts
@@ -15,7 +15,6 @@ const isString = (value: unknown): value is string => typeof value === 'string'
 const isNumber = (value: unknown): value is number => typeof value === 'number'
 const isBoolean = (value: unknown): value is boolean => typeof value === 'boolean'
 const isNumeric = (value: string) => /^-?\d+$/.test(value)
-const isDecimal = (value: string): boolean => /\d+\.\d+/.test(value)
 const hasLeadingZero = (value: string) => value.length > 1 && value.startsWith('0')
 
 const isInt32 = (number: number) => {
@@ -48,11 +47,6 @@ export function encodeCredentialValue(value: unknown) {
 
   // If value is int32 return as number string
   if (isNumber(value) && isInt32(value)) {
-    return value.toString()
-  }
-
-  // If value is a number string with leading zero and not a decimal return as number string
-  if (isString(value) && !isNaN(Number(value)) && !isDecimal(value) && hasLeadingZero(value)) {
     return value.toString()
   }
 
--- a/packages/anoncreds/src/utils/credential.ts
+++ b/packages/anoncreds/src/utils/credential.ts
@@ -15,6 +15,8 @@ const isString = (value: unknown): value is string => typeof value === 'string'
 const isNumber = (value: unknown): value is number => typeof value === 'number'
 const isBoolean = (value: unknown): value is boolean => typeof value === 'boolean'
 const isNumeric = (value: string) => /^-?\d+$/.test(value)
+const isDecimal = (number: number) => !Number.isInteger(number)
+const hasLeadingZero = (value: string) => value.length > 1 && value.startsWith('0')
 
 const isInt32 = (number: number) => {
   const minI32 = -2147483648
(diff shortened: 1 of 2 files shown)


packages/anoncreds/src/utils/credential.ts
```diff
@@ -49,6 +51,11 @@ export function encodeCredentialValue(value: unknown) {
     return value.toString()
   }
 
+  // If value is a number string with leading zero and not a decimal return as number string
+  if (isString(value) && !isNaN(Number(value)) && !isDecimal(value) && hasLeadingZero(value)) {
+    return value.toString()
+  }
+
```
TimoGlastra, line 58: If the value is not a number we still need to encode it as a number.

So i would expand the next if statement to also add && !hadLeadingZero.

Otherwise it will reach the end as be encoded as any other string (by hashing)

TimoGlastra (changes requested):

packages/anoncreds/src/utils/credential.ts
TimoGlastra: You either need to encode it as a number and remove the leading zero, or you need to encode it as a normal string

sairanjit: I think adding `!hasLeadingZero(value)` in the next if statement resolves it.

packages/anoncreds/src/utils/credential.ts
```diff
@@ -49,8 +51,20 @@ export function encodeCredentialValue(value: unknown) {
...
```
TimoGlastra, line 58: This can be removed now right?

sairanjit: Okay If I remove that then the number `0678` is hashed instead it should return a number right ?

TimoGlastra: A number with a leading zero is not a valid number i think for AnonCreds

@berendsliedrecht we had an issue with leading 0 before in AnonCreds. What's the behavior there so we can match it?

sairanjit: @TimoGlastra This PR [fix leading zero bug](https://github.com/hyperledger/anoncreds-rs/pull/255) in AnonCreds has some fixes.

TimoGlastra: Yeah so it trims the leading 0's. We'd have to fix it in anoncreds as well. 

I think it's best to treat strings with leading 0's as a string. If you want it as a number, just remove the leading 0's manually.

sairanjit: Yes, I updated the PR with the required changes now

TimoGlastra: I think we also have to adjust in AnonCreds RS, as that's where the values are transformed for the credential. 

We should keep behavior the same beteeen the two

berendsliedrecht: according to https://github.com/hyperledger/aries-rfcs/blob/50d148b812c45af3fc847c1e7033b084683dceb7/features/0592-indy-attachments/README.md#notes-on-encoding-claims we should keep the leading zeroes in the raw values and strip them in the encoded value. 

> I think it's best to treat strings with leading 0's as a string.

That would mean that integer strings with leading zeroes can not be used for predicates as the encoded value will be used for that. 

> The use of AnonCreds predicates, such as proving "older than 21" based on a date of birth claim without sharing the date of birth, is based on an expression involving the encoded value. Thus, only raw integers or string integers can be used in AnonCreds predicates.

I think we should follow the structure for the encoding claims (I think which already happens correctly in anoncreds-rs) and we should maybe open an issue if a claim requires leading zeroes if that can be enabled somehow.

berendsliedrecht: nothing, this PR contradicts the definition of the RFC. There is absolutely a valid use case for leading zero numbers in claims, but that should change in the RFC and then we can update accordingly. I would not want to derive from the specs.

sairanjit: @berendsliedrecht I agree with you if we need to update the logic we should change in RFC first. @TimoGlastra Can we close this PR as of now ?

PR #1842, opened 2024-04-25
fix: return w3c for anoncreds

TimoGlastra:


genaris (commented):
This seems a super-elegant fix! But I'm not sure if it will work, since the `credentialRecordId` of the created `W3CCredentialRecord` seems to be different from the previous one. Am I right?

TimoGlastra:
Hmm but in that case we also aren't able to determine the id in the migration script for 0.6? As there's not link anymore between the W3C record and the old record?

TimoGlastra:
But you're right, good catch @genaris. The ID is different

TimoGlastra:
Not sure how to fix this... I think we shouldn't hold up the 0.5.2 release for it but good to properly fix. I'm going to think about it in more detail

genaris:
> Hmm but in that case we also aren't able to determine the id in the migration script for 0.6? As there's not link anymore between the W3C record and the old record?

Yes, unfortunately at the moment I don't see any straightforward upgrade path for those who are already using 0.5.1. In our case we got the relationship by using the issuance date, since we were storing some info within the record metadata, but I don't know if there is some other way to link both records once the anoncreds one is deleted.

PR #1844, opened 2024-04-25
fix(anoncreds): credential exchange record migration

genaris:
As discussed in today's meeting, in order to properly bind existing Credential Exchange Records to the credentials stored as a result of them.


TimoGlastra (approved):
LGTM 👍

PR #1847, opened 2024-04-29
fix: store recipient keys by default

TimoGlastra:
When integrating the new `invitationDid` with a `did:cheqd` DID, I noticed that DIDComm wouldn't work because the recipientKeyFingerprints weren't added.

To make it easier, it is now calculated in the did record if the did document is provided, and I've also updated the `did:indy` registrar to store the did document. It's very expensive to resolve these documents, and this will allow us in a future update to shortcut did docs created by us to first look for the local record.

But for now, it makes it easy to use any kind of did with DIDComm.


genaris (approved):
Nice optimization!

PR #1850, opened 2024-04-30
fix: cheqd create from did document

TimoGlastra:
Some typing improvements and added a test for creating a cheqd did based on a did document.

Also added a failing test for adding a P256 key. this is failing because the signature is invalid according to the cheqd node. Have done quite some debugging and even by trying the cheqd signer (using an Askar generated key), but haven't been able to get it to work in any way. Will add a follow up Pr with tests once I know more.

Also sadly we can't use DIDcomm with cheqd dids, as they strip out the recipientKeys property from the didcomm service :(


berendsliedrecht:
> Also added a failing test for adding a P256 key. this is failing because the signature is invalid according to the cheqd node. Have done quite some debugging and even by trying the cheqd signer (using an Askar generated key), but haven't been able to get it to work in any way. Will add a follow up Pr with tests once I know more.

It could be the signature format? IIRC Askar supports verifying multiple formats, but (of course) only outputs it in a single format. Could be `ASN.1 der`. Just something I also ran into when working on secure env, but it could very well be something else.

packages/cheqd/src/anoncreds/utils/identifiers.ts
```diff
@@ -44,7 +45,7 @@ export function parseCheqdDid(didUrl: string): ParsedCheqdDid | null {
     const parts: ParsedCheqdDid = {
       did: `did:cheqd:${sections[1]}:${sections[2]}`,
       method: 'cheqd',
-      network: sections[1],
+      network: sections[1] as `${CheqdNetwork}`,
```
berendsliedrecht, line 48: Do we need to do string interpolation for the cast?

berendsliedrecht (approved):

packages/cheqd/src/anoncreds/utils/identifiers.ts
TimoGlastra: It's not an enum value, but rather the string value. I don't think it really matters, but i also use it as otherwise it will complain if you don't do `CheqdNetowrk.Testnet` (which we don't want in the public api, just `testnet` is easier)

PR #1856, opened 2024-05-03
feat: allow serving dids from did record

TimoGlastra:
It's a bit weird that for DIDs we created we have to resolve them from the ledger quite often. We already introduced using a cache, which mainly resolved the issue where we would resolve a did multiple times in a session. 

This will help with keeping things fast and local when working with did documents locally (e.g. signing credentials, etc..). We also noticed that some blockchains need some time to process did documents, and that it prevents you from already using the key in a did document. So using the local did record can speed things up quite a lot.

We will have to make improvements still to e.g. the cheqd anoncreds resolver as that bypasses this layer (which is good as it needs metadata from the resolver). But there may be some additional optimizations we can do to keep signing completely local (which has significant impact on performance).

In v0.6.0 we can make the property `allowsLocalDidRecord` on a did resolver required. For now it defaults to false (to not break existing logic) and have enabled it for all remote dids (so did:key, and did:jwk, did:peer are always served from the resolver as it's computed a lot of the times)

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -2,10 +2,10 @@ name: Continuous Integration
 
 on:
   pull_request:
-    branches: [main, "**-pre"]
+    branches: [main, '**-pre']
     types: [opened, synchronize, reopened, labeled]
   push:
-    branches: [main, "**-pre"]
+    branches: [main, '**-pre']
   workflow_dispatch:
 
 env:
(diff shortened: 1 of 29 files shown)


genaris (approved):
I was doing pretty much the same thing for did:web (in my case, extending WebDidResolver class and overriding the resolve method), where I think this is most relevant, but I like this generic approach!

PR #1859, opened 2024-05-08
fix: process problem report message

### LINKED ISSUE #1858: Issue while declining the credential offer
```
ERROR: Error handling message with type https://didcomm.org/issue-credential/2.0/problem-report {
  "message": {
    "@type": "https://didcomm.org/issue-credential/2.0/problem-report",
    "@id": "f5fa72ac-f98e-4857-8442-5442e06cab55",
    "description": {
      "en": "Declined by holder",
      "code": "issuance-abandoned"
    },
    "~thread": {
      "thid": "510ecce1-818b-4751-8c03-244ded51b842",
      "pthid": "fbc2a473-d306-478e-a558-9f25bf1ec310"
    }
  },
  "error": {
    "name": "RecordNotFoundError",
    "message": "CredentialRecord: No record found for given query '{\"connectionId\":\"a283bfab-fd15-46ba-8dae-f0bd249350f0\",\"threadId\":\"510ecce1-818b-4751-8c03-244ded51b842\"}'",
  },
  "senderKey": "z6MkfxkeseKXHYJ37vzXXJPM7wKYRVMivZNkWqNXCC2ifvBx",
  "recipientKey": "z6Mkvr76YcHTfc1uRRDa8XMifKxsTA2AWHXBKe4uo7kNXhJ1",
  "connectionId": "a283bfab-fd15-46ba-8dae-f0bd249350f0"
}
```

This is happening because there is no connection attached for credential record in the credential repository

 - We can fix this by removing the connectionId while querying here
 https://github.com/openwallet-foundation/credo-ts/blob/main/packages/core/src/modules/credentials/protocol/BaseCredentialProtocol.ts#L147

sairanjit:
### Summary

Fixes #1858

### DIFF
--- a/packages/core/src/modules/credentials/protocol/BaseCredentialProtocol.ts
+++ b/packages/core/src/modules/credentials/protocol/BaseCredentialProtocol.ts
@@ -138,13 +138,10 @@ export abstract class BaseCredentialProtocol<CFs extends CredentialFormatService
   ): Promise<CredentialExchangeRecord> {
     const { message: credentialProblemReportMessage, agentContext } = messageContext
 
-    const connection = messageContext.assertReadyConnection()
-
     agentContext.config.logger.debug(`Processing problem report with message id ${credentialProblemReportMessage.id}`)
 
     const credentialRecord = await this.getByProperties(agentContext, {
       threadId: credentialProblemReportMessage.threadId,
-      connectionId: connection.id,
     })
 
     // Update record
--- a/packages/core/src/modules/credentials/protocol/BaseCredentialProtocol.ts
+++ b/packages/core/src/modules/credentials/protocol/BaseCredentialProtocol.ts
@@ -150,7 +150,9 @@ export abstract class BaseCredentialProtocol<CFs extends CredentialFormatService
     })
 
     // Assert
-    await connectionService.assertConnectionOrOutOfBandExchange(messageContext)
+    await connectionService.assertConnectionOrOutOfBandExchange(messageContext, {
+      expectedConnectionId: credentialRecord.connectionId,
+    })
 
     //  This makes sure that the sender of the incoming message is authorized to do so.
     if (!credentialRecord?.connectionId) {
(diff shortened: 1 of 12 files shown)


packages/core/src/modules/credentials/protocol/BaseCredentialProtocol.ts
```diff
@@ -138,13 +138,10 @@ export abstract class BaseCredentialProtocol<CFs extends CredentialFormatService
...
     const { message: credentialProblemReportMessage, agentContext } = messageContext
 
-    const connection = messageContext.assertReadyConnection()
-
     agentContext.config.logger.debug(`Processing problem report with message id ${credentialProblemReportMessage.id}`)
 
     const credentialRecord = await this.getByProperties(agentContext, {
       threadId: credentialProblemReportMessage.threadId,
-      connectionId: connection.id,
     })
```
TimoGlastra, line 145: This is dangerous as it allows anyone to send a response to the credential exchange by knowing the thread id. For this reason we don't support problem reports currently for connectionsless exchanges.

If we want to support this, we need to handle it the same as is done in the processRequest message. Which consists of:
- query the record without connection id
- then if the record has a connection id:
  - check if it matches the connection id from the exchange
- if the record does not have a connection id:
  - check that the problem report is authorized based on the out of band exchange (the parentThreadId also need to match)
  - if the incoming message has a connection associated, update the connectionId in the exchange record

This is already implemented like this for processRequest, and I think you can copy most of the logic

TimoGlastra (changes requested):

TimoGlastra:
Hmm it seems there's a small issue with the logic (not from your code. I'll push some changes to this PR tomorrow to fix that as well)

TimoGlastra:
Pushed some changes @sairanjit

sairanjit:
@TimoGlastra I removed this check 
```typescript
const connection = messageContext.assertReadyConnection()
```
As it was failing for connection less flows

TimoGlastra (approved):

PR #1865, opened 2024-05-14
Feat: json ld fixes and aca-py interop fixes

KolbyRKunz:
This PR fixes some minor typos and issues with JSON-LD between aca-py and credo. It additionally updates did:peer:2 to ignore unknown service objects, and allow multiple services to be on a did:peer:2 did. These fixes overall allows interop with aca-py's implementation of JSON-LD over didCOMM.


packages/core/src/modules/connections/DidExchangeProtocol.ts
```diff
@@ -523,64 +523,73 @@ export class DidExchangeProtocol {
...
+        } catch (e) {
+          this.logger.warn(`Document does not contain didRotate. Error: ${e.message}`)
+        }
       }
     }
 
     // Now resolve the document related to the did (which can be either a public did or an inline did)
     try {
-      return await agentContext.dependencyManager.resolve(DidsApi).resolveDidDocument(message.did)
+      if (message.did.startsWith('did:'))
```
genaris, line 583: In which cases the `did` field in a DID Exchange Request or Response would not start by `did:`?

packages/core/src/modules/connections/DidExchangeProtocol.ts
```diff
@@ -523,64 +523,73 @@ export class DidExchangeProtocol {
...
+                signerKeys
+              )} invitationKeys:${JSON.stringify(invitationKeysBase58)}`,
+              {
+                problemCode: DidExchangeProblemReportReason.ResponseNotAccepted,
+              }
+            )
           }
-        )
+        } catch (e) {
+          this.logger.warn(`Document does not contain didRotate. Error: ${e.message}`)
```
genaris, line 576: I think this warning should be shown when there is not any did rotate attachment (i.e. in an `else` block in this case), but I don't see why we should be catching the validations we do here (like the did, JWS header, etc.)

packages/core/src/modules/dids/domain/verificationMethod/Bls12381G1Key2020.ts
```diff
@@ -32,9 +32,17 @@ export function isBls12381G1Key2020(verificationMethod: VerificationMethod): ver
...
+    const key = Key.fromFingerprint(verificationMethod.publicKeyMultibase)
+    if (key.keyType === KeyType.Bls12381g1) return key
+    else
+      throw new CredoError(
+        `Unexpected key type from resolving multibase encoding, key type was ${key.keyType} but expected ${KeyType.Bls12381g1}}`
+      )
   }
 
-  return Key.fromPublicKeyBase58(verificationMethod.publicKeyBase58, KeyType.Bls12381g1)
+  throw new CredoError('verification method is missing publicKeyBase58')
```
genaris, line 47: As now we also accept multibase we can update the error?

```suggestion
  throw new CredoError('verification method is missing publicKeyBase58 or publicKeyMultibase')
```

packages/core/src/modules/vc/data-integrity/W3cJsonLdCredentialService.ts
```diff
@@ -339,12 +339,23 @@ export class W3cJsonLdCredentialService {
     agentContext: AgentContext,
     verificationMethod: string
   ): Promise<Key> {
-    const documentLoader = this.w3cCredentialsModuleConfig.documentLoader(agentContext)
-    const verificationMethodObject = await documentLoader(verificationMethod)
-    const verificationMethodClass = JsonTransformer.fromJSON(verificationMethodObject.document, VerificationMethod)
-
-    const key = getKeyFromVerificationMethod(verificationMethodClass)
-    return key
+    if (!verificationMethod.startsWith('did:peer')) {
```
genaris, line 342: What about using `isValidPeerDid`?

genaris (commented):
Thanks @KolbyRKunz ! Good fixes here. I left some minor comments and also doubts on DID Exchange message handling but overall looks good.

packages/core/src/modules/dids/domain/verificationMethod/EcdsaSecp256k1VerificationKey2019.ts
```diff
@@ -43,9 +43,17 @@ export function isEcdsaSecp256k1VerificationKey2019(
  * Get a key from a EcdsaSecp256k1VerificationKey2019 verification method.
  */
 export function getKeyFromEcdsaSecp256k1VerificationKey2019(verificationMethod: EcdsaSecp256k1VerificationKey2019) {
-  if (!verificationMethod.publicKeyBase58) {
-    throw new CredoError('verification method is missing publicKeyBase58')
+  if (verificationMethod.publicKeyBase58) {
+    return Key.fromPublicKeyBase58(verificationMethod.publicKeyBase58, KeyType.K256)
+  }
+  if (verificationMethod.publicKeyMultibase) {
```
TimoGlastra, line 49: Don't we have this check on level higher in getKeyFromVerificationMethod? We can parse this generically and thus don't have to support it for each key type separately (you can look at JsonWebKey2020 as an example, as it behaves the same)

packages/core/src/modules/dids/methods/peer/peerDidNumAlgo2.ts
```diff
@@ -69,10 +69,13 @@ export function didToNumAlgo2DidDocument(did: string) {
...
         service = expandServiceAbbreviations(service)
-
         service.id = `${did}#${service.type.toLowerCase()}-${serviceIndex++}`
-
-        didDocument.addService(JsonTransformer.fromJSON(service, DidDocumentService))
+        try {
+          didDocument.addService(JsonTransformer.fromJSON(service, DidDocumentService))
+        } catch (e) {
+          //Ignore a service if we do not recognize it
+          serviceIndex--
```
TimoGlastra, line 77: We should nog skip the service index i think, as then the index references will be messed up. I think we can skip adding it to the document, but also skip the index

TimoGlastra: Also furious when does this happen? If a json transform fails? I don't think we should just swallow such an error. Can you explain why this is needed?

KolbyRKunz: Looking at this I realize this was a bit of a hold over from implementing this functionality in another library that had much stricter serialization rules than typescript does, specifically with unknown additional keys appearing causing errors. With the regex adjustment this is redundant for Credo.

packages/core/src/modules/dids/domain/verificationMethod/EcdsaSecp256k1VerificationKey2019.ts
KolbyRKunz: From what I am seeing the `getKeyFromVerificationMethod` function will check if it is multikey in type and then attempt to process the multikey encoding. The issue I was running into in testing was that this function was not the function used everywhere for resolving keys from verification methods. Specifically while processing JSON-LD elsewhere there is code attempting to resolve a specific key type from a verification method that is of type multikey and so we would fail to resolve there. Aca-py by default will now use the multikey encoding and I believe there is a push to make that the standard encoding. I think having this on per-key basis and in the `getKeyFromVerificationMethod` function is more fault tolerant for resolving keys and is the only way to ensure the key you are resolving is of a specific type.

packages/core/src/modules/connections/DidExchangeProtocol.ts
KolbyRKunz: This came up as an issue with testing against aca-py and being sent an unqualified did in response to using `did:peer:1`. Aca-py supports processing `did:peer:1` but will not send one back which is strange behavior on their part. I would suggest taking this change for now and opening an issue with Aca-py to try and point out this behavior. I might not be right but I was under the impression that unqualified dids were going to the way side here soon.

genaris: Well, I'd rather suggest to fix the issue in ACA-Py so both are working as expected :stuck_out_tongue: . Jokes apart, we are not supporting unqualified DIDs in DID Exchange protocol so I think it is better to keep it like that.

packages/core/src/modules/vc/data-integrity/W3cJsonLdCredentialService.ts
KolbyRKunz: I have updated this to just check it if starts with `did:` as I am uncertain if it will always be a `did:peer`. I have left the split in the else branch as it is only needed there.

packages/core/src/modules/dids/methods/peer/peerDidNumAlgo2.ts
KolbyRKunz: Removed this logic as it is not needed in Credo because serialization is more fault tolerant than I thought at development time.

packages/anoncreds/src/utils/w3cAnonCredsUtils.ts
```diff
@@ -166,6 +167,14 @@ export function getStoreCredentialOptions(
...
 }
 
+// The issuer of the schema does not always match the issuer of the credential definition thus the unqualified schema id needs to be derived from both values
+function getUnqualifiedSchemaId(schemaIssuerId: string, schemaId: string) {
+  const schemaDid = schemaIssuerId.split(':')[3]
+  const split = getUnQualifiedDidIndyDid(schemaId).split(':')
+  split[0] = schemaDid
+  return split.join(':')
+}
+
```
KolbyRKunz, line 177: I made a comment in the discord about this but it looks like Credo is not reconstructing the schema id correctly in some cases. I have found that his will work but I cannot track if it is a issuer higher up in the code or not. I would be happy to fix it higher up if anyone see how to.

packages/core/src/modules/dids/domain/verificationMethod/EcdsaSecp256k1VerificationKey2019.ts
TimoGlastra: Hmm okay -- ideally we update the calls that use the specific getKeyFromVerificationMethod and use the generic one instead. that way we encourage moving towards this approach rather than adding the multikey to format/key-type specific transformers

packages/anoncreds/src/utils/w3cAnonCredsUtils.ts
```diff
@@ -20,6 +20,7 @@ import {
   isUnqualifiedRevocationRegistryId,
   isIndyDid,
   getUnQualifiedDidIndyDid,
+  isUnqualifiedIndyDid,
```
TimoGlastra, line 23: This is unrelated to this PR right? And probably fixed by some of the other PRs opened by @genaris?

packages/core/src/modules/dids/methods/peer/didPeer.ts
```diff
@@ -3,7 +3,7 @@ import { CredoError } from '../../../../error'
 import { getAlternativeDidsForNumAlgo4Did } from './peerDidNumAlgo4'
 
 const PEER_DID_REGEX = new RegExp(
-  '^did:peer:(([01](z)([1-9a-km-zA-HJ-NP-Z]{5,200}))|(2((.[AEVID](z)([1-9a-km-zA-HJ-NP-Z]{5,200}))+(.(S)[0-9a-zA-Z=]*)?))|([4](z[1-9a-km-zA-HJ-NP-Z]{46})(:z[1-9a-km-zA-HJ-NP-Z]{6,}){0,1}))$'
+  '^did:peer:(([01](z)([1-9a-km-zA-HJ-NP-Z]{5,200}))|(2((.[AEVID](z)([1-9a-km-zA-HJ-NP-Z]{5,200}))+(.(S)[0-9a-zA-Z=]*)*))|([4](z[1-9a-km-zA-HJ-NP-Z]{46})(:z[1-9a-km-zA-HJ-NP-Z]{6,}){0,1}))$'
```
TimoGlastra, line 6: What does this change in regex mean?

packages/core/src/modules/dids/methods/peer/peerDidNumAlgo2.ts
TimoGlastra: The code is still here?

packages/core/src/modules/vc/data-integrity/W3cJsonLdCredentialService.ts
```diff
@@ -339,12 +339,23 @@ export class W3cJsonLdCredentialService {
...
+    } else {
+      const [did, keyid] = verificationMethod.split('#')
+      const didsApi = agentContext.dependencyManager.resolve(DidsApi)
+      const doc = await didsApi.resolve(did)
+      if (doc.didDocument) {
+        const verificationMethodClass = doc.didDocument.dereferenceKey(keyid)
+        return getKeyFromVerificationMethod(verificationMethodClass)
+      }
+      throw new CredoError(`Could not resolve verification method with id ${verificationMethod}`)
+    }
```
TimoGlastra, line 358: Why is this needed? The document loader should be able to load did documents, so I'm not sure if we want to bypass it here

packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
```diff
@@ -268,7 +268,7 @@ export class IndyVdrAnonCredsRegistry implements AnonCredsRegistry {
       agentContext.config.logger.trace(
         `Submitting get credential definition request for credential definition '${credentialDefinitionId}' to ledger '${pool.indyNamespace}'`
       )
-      const response = await pool.submitRequest(request)
+      const response: GetCredentialDefinitionResponse = await pool.submitRequest(request)
```
TimoGlastra, line 271: Is this needed? The response type should automatically be inferred from the request

packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
```diff
@@ -978,9 +978,11 @@ export class IndyVdrAnonCredsRegistry implements AnonCredsRegistry {
       return null
     }
 
+    const schemaDid = response.result.data?.txn.metadata['from'] as string
```
TimoGlastra, line 981: this also works?

```suggestion
    const schemaDid = response.result.data?.txn.metadata.from as string
```

TimoGlastra: Also -- I think this is solved by https://github.com/openwallet-foundation/credo-ts/pull/1884

TimoGlastra (changes requested):
Hey @KolbyRKunz, left another review. I think it would be great to get it merged. if you can:
- remove the changes not related to JSON-LD (anoncreds / indy-did) etc.. I think these have been solved by other PRs
- update the getKeyFromVerificationMethod to support multi-key and update all places that use the specific getKeyFromXXXX methods to use getKeyFromVerificationMethod instead.
- revert the change to the peerDidNumalgo2.ts file
- revert the change to the `getPublicKeyFromVerificationMethod` or explain specifically why this change is needed (as the did resolving is handled by the document loader)

packages/core/src/modules/connections/DidExchangeProtocol.ts
```diff
@@ -523,58 +523,56 @@ export class DidExchangeProtocol {
...
+              signerKeys
+            )} invitationKeys:${JSON.stringify(invitationKeysBase58)}`,
+            {
+              problemCode: DidExchangeProblemReportReason.ResponseNotAccepted,
+            }
+          )
+        }
+      } else {
+        this.logger.warn(`Document does not contain didRotate`)
       }
```
TimoGlastra, line 576: I don't think this change is correct. `extractResolvableDidDocument` is called when the did document is not `did:peer:1`. If we resolve the method, we REQUIRE the rotate attachment to be present. Could you explain a bit more why this change is needed?

packages/core/src/modules/dids/methods/peer/peerDidNumAlgo2.ts
KolbyRKunz: With my comment below about did:peer:2 adding multiple services this was added because some of the agents and mediator I tested against included didCommV2 services that were intended to be ignored by the agent if they do not use didCommV2.  Have this try catch here successfully ignores it and keeps the indexes of the services we do recognize consistent.

packages/core/src/modules/dids/methods/peer/didPeer.ts
KolbyRKunz: This regex is specifically changed for `did:peer:2`. With the updates to the protocol of `did:peer:2` the services on the encoded version are not always a list and can be multiple service appended instead.

packages/core/src/modules/vc/data-integrity/W3cJsonLdCredentialService.ts
KolbyRKunz: The document loader would resolve the did but not return the full did Document with the verification information applied. This might have changed since my testing because I added this while testing with `0.5.1`

packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
KolbyRKunz: It was not automatically inferring the type for me when I was developing, not sure if it was an issue with my environment or not.

packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
KolbyRKunz: I was planning on remove this because of the merged PR

packages/core/src/modules/connections/DidExchangeProtocol.ts
KolbyRKunz: The test case I ran into was agents or mediators using did:peer:2 without the intention of doing a did rotate or `alsoKnownAs`. It also is a part of the didExchange RFC that the value is optional for any reason so for interoperability sake I think this should remain.

TimoGlastra: If the did is not rotated from the invitation we can indeed ignore it. But if the did is rotated, not doing this check means the connection can be hijacked. I could send a response even though you created the invitation, because we don't require a signature from the invitation did. If the did is the same, then we don't have that issue and i'm ok with not requiring did rotate. But the current code doesn't take this into consideration

packages/core/src/modules/dids/domain/verificationMethod/EcdsaSecp256k1VerificationKey2019.ts
KolbyRKunz: I added a change that checks if the type is multikey or if the multikey property is defined then to use that on the multikey resolution. After merging main I can no longer find the place that used the ed25519 key resolution directly so hopefully that should now be resolved.

packages/anoncreds/src/utils/w3cAnonCredsUtils.ts
KolbyRKunz: This got added further down in the file just for the case that the did is an unqualified indy did already. The code retrieves the unqualified indy did using the indy did but if it starts as an unqualified indy did the check fails. even though the `getUnqualifiedDidIndyDid` accounts for this case internally.

KolbyRKunz:
@TimoGlastra Sorry for the long time between reviews. I have made most of the recommended changes and commented on the ones that I did not. Hopefully that can get this PR on. Not sure if/when I will be available to update this PR again just so you are aware. Hopefully this helps :)

packages/core/src/modules/connections/DidExchangeProtocol.ts
KolbyRKunz: I might be confused but the warn only happens if the attachment is not present, the else block is for the if statement on line 526 where it checks if the `didRotateAttachment` is defined. The signature is always checked if the rotate attachment is present. The else block with the warn was a suggestion made by @genaris when this PR was initially made.

TimoGlastra: The check that is now in place always allows didRotate to not be present, meaning we could just remove the check. So we need to be more specific in checking when didRotate needs to be present: when you rotate your did from the invitation. Even though the Aries RFC mentions it's optional, it's super important for security to do this check. did_rotate was added specifically for this reason, and it's only optional to not introduce a breaking change in the RFC. Before did rotate was introduced, it was basically not possible to rotate a did from the invitation if you were not using diddoc~attach and sign it.

If the did is rotated from the invitation, either a signed diddoc~attach must be present OR a signed did_rotate~attach must be present

packages/core/src/modules/dids/methods/peer/peerDidNumAlgo2.ts
TimoGlastra: But by doing this the references between implementations will mess up. We would skip didcomm v2 and others would not skip it and thus the same id could point to different services. I think we should not do this

packages/core/src/modules/vc/data-integrity/W3cJsonLdCredentialService.ts
TimoGlastra: I'm sorry, could you elaborate a bit further? What do you mean with "not return the full did Document with the verification information applied"? What does it return?

packages/core/src/modules/dids/methods/peer/peerDidNumAlgo2.ts
TimoGlastra: AFter looking at this in more depth -- we don't handle the DIDComm v2 serviceEndpoint being an object. So instead of doing this, we need to fix that parsing of serviceEndpoint instead

TimoGlastra: Do you have example did:peer:2 document with DIDComm v2 service for test? I have a PR locally that correctly fixes https://github.com/openwallet-foundation/credo-ts/issues/1789, and thus should also fix this issue

packages/core/src/modules/dids/methods/peer/didPeer.ts
TimoGlastra: Do you have an example did:peer so I can add a test covering this?

TimoGlastra:
Okay with #1903 and #1902, the following features from this PR remain:
- support array in did:peer:2? -> Need an example did so we can add a test
- did dodcument loader with JSON-LD not using the JSON-LD loader -> need more input on what the issue here is exactly. As there could be different behaviour in not using the JSON-LD loader
- did rotate handling. We should check that if the did is rotated from the invitation, the did_rotate is present and valid. Only if the did is not rotated, is it OK to skip the did_rotate attachment.

All other changes should be removed (please look at changed files tab and undo all changes not needed)

KolbyRKunz:
An example did:peer:2 is this one `did:peer:2.Ez6LSe3YyteKQVXSgGfZyCzbLq9K34zjAfap4EuaWJjbo5Gwk.Vz6MkvLvjZVc9jWJVs8M3qNZeRsnXXZGFjzcRJDH9vvdJMgHB.SeyJ0IjoiZG0iLCJzIjp7InVyaSI6Imh0dHBzOi8vdXMtZWFzdC5wcm92ZW4ubWVkaWF0b3IuaW5kaWNpb3RlY2guaW8vbWVzc2FnZSIsImEiOlsiZGlkY29tbS92MiIsImRpZGNvbW0vYWlwMjtlbnY9cmZjMTkiXX19.SeyJ0IjoiZG0iLCJzIjp7InVyaSI6IndzczovL3dzLnVzLWVhc3QucHJvdmVuLm1lZGlhdG9yLmluZGljaW90ZWNoLmlvL3dzIiwiYSI6WyJkaWRjb21tL3YyIiwiZGlkY29tbS9haXAyO2Vudj1yZmMxOSJdfX0.SeyJzIjogImh0dHBzOi8vdXMtZWFzdC5wcm92ZW4ubWVkaWF0b3IuaW5kaWNpb3RlY2guaW8vbWVzc2FnZSIsICJhIjogWyJkaWRjb21tL2FpcDEiLCJkaWRjb21tL2FpcDI7ZW52PXJmYzE5Il0sICJyZWNpcGllbnRLZXlzIjogWyIja2V5LTIiXSwgInQiOiAiZGlkLWNvbW11bmljYXRpb24ifQ.SeyJzIjogIndzczovL3dzLnVzLWVhc3QucHJvdmVuLm1lZGlhdG9yLmluZGljaW90ZWNoLmlvL3dzIiwgImEiOiBbImRpZGNvbW0vYWlwMSIsImRpZGNvbW0vYWlwMjtlbnY9cmZjMTkiXSwgInJlY2lwaWVudEtleXMiOiBbIiNrZXktMiJdLCAidCI6ICJkaWQtY29tbXVuaWNhdGlvbiJ9`

It has the multiple service endpoints that require the regex change and it also has didCommV2 endpints

KolbyRKunz:
The document loader issue I was running into was that it would resolve a didDoc but the verification arrays would come back as empty services, they would only contain ids. Something like `{ id: "did:peer:2:someRandomDid#key-1`}, the actual keys would not be included making it impossible to sign the payload correctly.

KolbyRKunz:
I changed the logic on the didExchange issue so that if we are getting a response message we now require that either a `didRotate` or `didDoc` attachment is added to the message so that we have a signature to validate against. I have also removed the changes that are made redundant by the other PR's that you mentioned @TimoGlastra. I also want to point out the addition of the buffer import in the indy-vdr package, without that import it is not possible to resolve an abbreviated verkey on a did:sov in React Native. Thanks for working through these fixes with me.

TimoGlastra:
Could you confirm whether this looks OK as resolved variant of the DID you provided?

```ts
{
      "@context": [
        "https://w3id.org/did/v1"
      ],
      "id": "did:peer:2.Ez6LSe3YyteKQVXSgGfZyCzbLq9K34zjAfap4EuaWJjbo5Gwk.Vz6MkvLvjZVc9jWJVs8M3qNZeRsnXXZGFjzcRJDH9vvdJMgHB.SeyJ0IjoiZG0iLCJzIjp7InVyaSI6Imh0dHBzOi8vdXMtZWFzdC5wcm92ZW4ubWVkaWF0b3IuaW5kaWNpb3RlY2guaW8vbWVzc2FnZSIsImEiOlsiZGlkY29tbS92MiIsImRpZGNvbW0vYWlwMjtlbnY9cmZjMTkiXX19.SeyJ0IjoiZG0iLCJzIjp7InVyaSI6IndzczovL3dzLnVzLWVhc3QucHJvdmVuLm1lZGlhdG9yLmluZGljaW90ZWNoLmlvL3dzIiwiYSI6WyJkaWRjb21tL3YyIiwiZGlkY29tbS9haXAyO2Vudj1yZmMxOSJdfX0.SeyJzIjogImh0dHBzOi8vdXMtZWFzdC5wcm92ZW4ubWVkaWF0b3IuaW5kaWNpb3RlY2guaW8vbWVzc2FnZSIsICJhIjogWyJkaWRjb21tL2FpcDEiLCJkaWRjb21tL2FpcDI7ZW52PXJmYzE5Il0sICJyZWNpcGllbnRLZXlzIjogWyIja2V5LTIiXSwgInQiOiAiZGlkLWNvbW11bmljYXRpb24ifQ.SeyJzIjogIndzczovL3dzLnVzLWVhc3QucHJvdmVuLm1lZGlhdG9yLmluZGljaW90ZWNoLmlvL3dzIiwgImEiOiBbImRpZGNvbW0vYWlwMSIsImRpZGNvbW0vYWlwMjtlbnY9cmZjMTkiXSwgInJlY2lwaWVudEtleXMiOiBbIiNrZXktMiJdLCAidCI6ICJkaWQtY29tbXVuaWNhdGlvbiJ9",
      "service": [
        {
          "type": "DIDCommMessaging",
          "serviceEndpoint": {
            "uri": "https://us-east.proven.mediator.indiciotech.io/message",
            "accept": [
              "didcomm/v2",
              "didcomm/aip2;env=rfc19"
            ]
          },
          "id": "did:peer:2.Ez6LSe3YyteKQVXSgGfZyCzbLq9K34zjAfap4EuaWJjbo5Gwk.Vz6MkvLvjZVc9jWJVs8M3qNZeRsnXXZGFjzcRJDH9vvdJMgHB.SeyJ0IjoiZG0iLCJzIjp7InVyaSI6Imh0dHBzOi8vdXMtZWFzdC5wcm92ZW4ubWVkaWF0b3IuaW5kaWNpb3RlY2guaW8vbWVzc2FnZSIsImEiOlsiZGlkY29tbS92MiIsImRpZGNvbW0vYWlwMjtlbnY9cmZjMTkiXX19.SeyJ0IjoiZG0iLCJzIjp7InVyaSI6IndzczovL3dzLnVzLWVhc3QucHJvdmVuLm1lZGlhdG9yLmluZGljaW90ZWNoLmlvL3dzIiwiYSI6WyJkaWRjb21tL3YyIiwiZGlkY29tbS9haXAyO2Vudj1yZmMxOSJdfX0.SeyJzIjogImh0dHBzOi8vdXMtZWFzdC5wcm92ZW4ubWVkaWF0b3IuaW5kaWNpb3RlY2guaW8vbWVzc2FnZSIsICJhIjogWyJkaWRjb21tL2FpcDEiLCJkaWRjb21tL2FpcDI7ZW52PXJmYzE5Il0sICJyZWNpcGllbnRLZXlzIjogWyIja2V5LTIiXSwgInQiOiAiZGlkLWNvbW11bmljYXRpb24ifQ.SeyJzIjogIndzczovL3dzLnVzLWVhc3QucHJvdmVuLm1lZGlhdG9yLmluZGljaW90ZWNoLmlvL3dzIiwgImEiOiBbImRpZGNvbW0vYWlwMSIsImRpZGNvbW0vYWlwMjtlbnY9cmZjMTkiXSwgInJlY2lwaWVudEtleXMiOiBbIiNrZXktMiJdLCAidCI6ICJkaWQtY29tbXVuaWNhdGlvbiJ9#didcommmessaging-0"
        },
        {
          "type": "DIDCommMessaging",
          "serviceEndpoint": {
            "uri": "wss://ws.us-east.proven.mediator.indiciotech.io/ws",
            "accept": [
              "didcomm/v2",
              "didcomm/aip2;env=rfc19"
            ]
          },
          "id": "did:peer:2.Ez6LSe3YyteKQVXSgGfZyCzbLq9K34zjAfap4EuaWJjbo5Gwk.Vz6MkvLvjZVc9jWJVs8M3qNZeRsnXXZGFjzcRJDH9vvdJMgHB.SeyJ0IjoiZG0iLCJzIjp7InVyaSI6Imh0dHBzOi8vdXMtZWFzdC5wcm92ZW4ubWVkaWF0b3IuaW5kaWNpb3RlY2guaW8vbWVzc2FnZSIsImEiOlsiZGlkY29tbS92MiIsImRpZGNvbW0vYWlwMjtlbnY9cmZjMTkiXX19.SeyJ0IjoiZG0iLCJzIjp7InVyaSI6IndzczovL3dzLnVzLWVhc3QucHJvdmVuLm1lZGlhdG9yLmluZGljaW90ZWNoLmlvL3dzIiwiYSI6WyJkaWRjb21tL3YyIiwiZGlkY29tbS9haXAyO2Vudj1yZmMxOSJdfX0.SeyJzIjogImh0dHBzOi8vdXMtZWFzdC5wcm92ZW4ubWVkaWF0b3IuaW5kaWNpb3RlY2guaW8vbWVzc2FnZSIsICJhIjogWyJkaWRjb21tL2FpcDEiLCJkaWRjb21tL2FpcDI7ZW52PXJmYzE5Il0sICJyZWNpcGllbnRLZXlzIjogWyIja2V5LTIiXSwgInQiOiAiZGlkLWNvbW11bmljYXRpb24ifQ.SeyJzIjogIndzczovL3dzLnVzLWVhc3QucHJvdmVuLm1lZGlhdG9yLmluZGljaW90ZWNoLmlvL3dzIiwgImEiOiBbImRpZGNvbW0vYWlwMSIsImRpZGNvbW0vYWlwMjtlbnY9cmZjMTkiXSwgInJlY2lwaWVudEtleXMiOiBbIiNrZXktMiJdLCAidCI6ICJkaWQtY29tbXVuaWNhdGlvbiJ9#didcommmessaging-1"
        },
        {
          "serviceEndpoint": "https://us-east.proven.mediator.indiciotech.io/message",
          "accept": [
            "didcomm/aip1",
            "didcomm/aip2;env=rfc19"
          ],
          "recipientKeys": [
            "#key-2"
          ],
          "type": "did-communication",
          "id": "did:peer:2.Ez6LSe3YyteKQVXSgGfZyCzbLq9K34zjAfap4EuaWJjbo5Gwk.Vz6MkvLvjZVc9jWJVs8M3qNZeRsnXXZGFjzcRJDH9vvdJMgHB.SeyJ0IjoiZG0iLCJzIjp7InVyaSI6Imh0dHBzOi8vdXMtZWFzdC5wcm92ZW4ubWVkaWF0b3IuaW5kaWNpb3RlY2guaW8vbWVzc2FnZSIsImEiOlsiZGlkY29tbS92MiIsImRpZGNvbW0vYWlwMjtlbnY9cmZjMTkiXX19.SeyJ0IjoiZG0iLCJzIjp7InVyaSI6IndzczovL3dzLnVzLWVhc3QucHJvdmVuLm1lZGlhdG9yLmluZGljaW90ZWNoLmlvL3dzIiwiYSI6WyJkaWRjb21tL3YyIiwiZGlkY29tbS9haXAyO2Vudj1yZmMxOSJdfX0.SeyJzIjogImh0dHBzOi8vdXMtZWFzdC5wcm92ZW4ubWVkaWF0b3IuaW5kaWNpb3RlY2guaW8vbWVzc2FnZSIsICJhIjogWyJkaWRjb21tL2FpcDEiLCJkaWRjb21tL2FpcDI7ZW52PXJmYzE5Il0sICJyZWNpcGllbnRLZXlzIjogWyIja2V5LTIiXSwgInQiOiAiZGlkLWNvbW11bmljYXRpb24ifQ.SeyJzIjogIndzczovL3dzLnVzLWVhc3QucHJvdmVuLm1lZGlhdG9yLmluZGljaW90ZWNoLmlvL3dzIiwgImEiOiBbImRpZGNvbW0vYWlwMSIsImRpZGNvbW0vYWlwMjtlbnY9cmZjMTkiXSwgInJlY2lwaWVudEtleXMiOiBbIiNrZXktMiJdLCAidCI6ICJkaWQtY29tbXVuaWNhdGlvbiJ9#did-communication-2"
        },
        {
          "serviceEndpoint": "wss://ws.us-east.proven.mediator.indiciotech.io/ws",
          "accept": [
            "didcomm/aip1",
            "didcomm/aip2;env=rfc19"
          ],
          "recipientKeys": [
            "#key-2"
          ],
          "type": "did-communication",
          "id": "did:peer:2.Ez6LSe3YyteKQVXSgGfZyCzbLq9K34zjAfap4EuaWJjbo5Gwk.Vz6MkvLvjZVc9jWJVs8M3qNZeRsnXXZGFjzcRJDH9vvdJMgHB.SeyJ0IjoiZG0iLCJzIjp7InVyaSI6Imh0dHBzOi8vdXMtZWFzdC5wcm92ZW4ubWVkaWF0b3IuaW5kaWNpb3RlY2guaW8vbWVzc2FnZSIsImEiOlsiZGlkY29tbS92MiIsImRpZGNvbW0vYWlwMjtlbnY9cmZjMTkiXX19.SeyJ0IjoiZG0iLCJzIjp7InVyaSI6IndzczovL3dzLnVzLWVhc3QucHJvdmVuLm1lZGlhdG9yLmluZGljaW90ZWNoLmlvL3dzIiwiYSI6WyJkaWRjb21tL3YyIiwiZGlkY29tbS9haXAyO2Vudj1yZmMxOSJdfX0.SeyJzIjogImh0dHBzOi8vdXMtZWFzdC5wcm92ZW4ubWVkaWF0b3IuaW5kaWNpb3RlY2guaW8vbWVzc2FnZSIsICJhIjogWyJkaWRjb21tL2FpcDEiLCJkaWRjb21tL2FpcDI7ZW52PXJmYzE5Il0sICJyZWNpcGllbnRLZXlzIjogWyIja2V5LTIiXSwgInQiOiAiZGlkLWNvbW11bmljYXRpb24ifQ.SeyJzIjogIndzczovL3dzLnVzLWVhc3QucHJvdmVuLm1lZGlhdG9yLmluZGljaW90ZWNoLmlvL3dzIiwgImEiOiBbImRpZGNvbW0vYWlwMSIsImRpZGNvbW0vYWlwMjtlbnY9cmZjMTkiXSwgInJlY2lwaWVudEtleXMiOiBbIiNrZXktMiJdLCAidCI6ICJkaWQtY29tbXVuaWNhdGlvbiJ9#did-communication-3"
        }
      ],
      "authentication": [
        {
          "id": "did:peer:2.Ez6LSe3YyteKQVXSgGfZyCzbLq9K34zjAfap4EuaWJjbo5Gwk.Vz6MkvLvjZVc9jWJVs8M3qNZeRsnXXZGFjzcRJDH9vvdJMgHB.SeyJ0IjoiZG0iLCJzIjp7InVyaSI6Imh0dHBzOi8vdXMtZWFzdC5wcm92ZW4ubWVkaWF0b3IuaW5kaWNpb3RlY2guaW8vbWVzc2FnZSIsImEiOlsiZGlkY29tbS92MiIsImRpZGNvbW0vYWlwMjtlbnY9cmZjMTkiXX19.SeyJ0IjoiZG0iLCJzIjp7InVyaSI6IndzczovL3dzLnVzLWVhc3QucHJvdmVuLm1lZGlhdG9yLmluZGljaW90ZWNoLmlvL3dzIiwiYSI6WyJkaWRjb21tL3YyIiwiZGlkY29tbS9haXAyO2Vudj1yZmMxOSJdfX0.SeyJzIjogImh0dHBzOi8vdXMtZWFzdC5wcm92ZW4ubWVkaWF0b3IuaW5kaWNpb3RlY2guaW8vbWVzc2FnZSIsICJhIjogWyJkaWRjb21tL2FpcDEiLCJkaWRjb21tL2FpcDI7ZW52PXJmYzE5Il0sICJyZWNpcGllbnRLZXlzIjogWyIja2V5LTIiXSwgInQiOiAiZGlkLWNvbW11bmljYXRpb24ifQ.SeyJzIjogIndzczovL3dzLnVzLWVhc3QucHJvdmVuLm1lZGlhdG9yLmluZGljaW90ZWNoLmlvL3dzIiwgImEiOiBbImRpZGNvbW0vYWlwMSIsImRpZGNvbW0vYWlwMjtlbnY9cmZjMTkiXSwgInJlY2lwaWVudEtleXMiOiBbIiNrZXktMiJdLCAidCI6ICJkaWQtY29tbXVuaWNhdGlvbiJ9#key-2",
          "type": "Ed25519VerificationKey2018",
          "controller": "did:peer:2.Ez6LSe3YyteKQVXSgGfZyCzbLq9K34zjAfap4EuaWJjbo5Gwk.Vz6MkvLvjZVc9jWJVs8M3qNZeRsnXXZGFjzcRJDH9vvdJMgHB.SeyJ0IjoiZG0iLCJzIjp7InVyaSI6Imh0dHBzOi8vdXMtZWFzdC5wcm92ZW4ubWVkaWF0b3IuaW5kaWNpb3RlY2guaW8vbWVzc2FnZSIsImEiOlsiZGlkY29tbS92MiIsImRpZGNvbW0vYWlwMjtlbnY9cmZjMTkiXX19.SeyJ0IjoiZG0iLCJzIjp7InVyaSI6IndzczovL3dzLnVzLWVhc3QucHJvdmVuLm1lZGlhdG9yLmluZGljaW90ZWNoLmlvL3dzIiwiYSI6WyJkaWRjb21tL3YyIiwiZGlkY29tbS9haXAyO2Vudj1yZmMxOSJdfX0.SeyJzIjogImh0dHBzOi8vdXMtZWFzdC5wcm92ZW4ubWVkaWF0b3IuaW5kaWNpb3RlY2guaW8vbWVzc2FnZSIsICJhIjogWyJkaWRjb21tL2FpcDEiLCJkaWRjb21tL2FpcDI7ZW52PXJmYzE5Il0sICJyZWNpcGllbnRLZXlzIjogWyIja2V5LTIiXSwgInQiOiAiZGlkLWNvbW11bmljYXRpb24ifQ.SeyJzIjogIndzczovL3dzLnVzLWVhc3QucHJvdmVuLm1lZGlhdG9yLmluZGljaW90ZWNoLmlvL3dzIiwgImEiOiBbImRpZGNvbW0vYWlwMSIsImRpZGNvbW0vYWlwMjtlbnY9cmZjMTkiXSwgInJlY2lwaWVudEtleXMiOiBbIiNrZXktMiJdLCAidCI6ICJkaWQtY29tbXVuaWNhdGlvbiJ9",
          "publicKeyBase58": "GtfgyFMiPxp2kdWM9oboanEXhyzQL7N4cCNE6efHSTVo"
        }
      ],
      "keyAgreement": [
        {
          "id": "did:peer:2.Ez6LSe3YyteKQVXSgGfZyCzbLq9K34zjAfap4EuaWJjbo5Gwk.Vz6MkvLvjZVc9jWJVs8M3qNZeRsnXXZGFjzcRJDH9vvdJMgHB.SeyJ0IjoiZG0iLCJzIjp7InVyaSI6Imh0dHBzOi8vdXMtZWFzdC5wcm92ZW4ubWVkaWF0b3IuaW5kaWNpb3RlY2guaW8vbWVzc2FnZSIsImEiOlsiZGlkY29tbS92MiIsImRpZGNvbW0vYWlwMjtlbnY9cmZjMTkiXX19.SeyJ0IjoiZG0iLCJzIjp7InVyaSI6IndzczovL3dzLnVzLWVhc3QucHJvdmVuLm1lZGlhdG9yLmluZGljaW90ZWNoLmlvL3dzIiwiYSI6WyJkaWRjb21tL3YyIiwiZGlkY29tbS9haXAyO2Vudj1yZmMxOSJdfX0.SeyJzIjogImh0dHBzOi8vdXMtZWFzdC5wcm92ZW4ubWVkaWF0b3IuaW5kaWNpb3RlY2guaW8vbWVzc2FnZSIsICJhIjogWyJkaWRjb21tL2FpcDEiLCJkaWRjb21tL2FpcDI7ZW52PXJmYzE5Il0sICJyZWNpcGllbnRLZXlzIjogWyIja2V5LTIiXSwgInQiOiAiZGlkLWNvbW11bmljYXRpb24ifQ.SeyJzIjogIndzczovL3dzLnVzLWVhc3QucHJvdmVuLm1lZGlhdG9yLmluZGljaW90ZWNoLmlvL3dzIiwgImEiOiBbImRpZGNvbW0vYWlwMSIsImRpZGNvbW0vYWlwMjtlbnY9cmZjMTkiXSwgInJlY2lwaWVudEtleXMiOiBbIiNrZXktMiJdLCAidCI6ICJkaWQtY29tbXVuaWNhdGlvbiJ9#key-1",
          "type": "X25519KeyAgreementKey2019",
          "controller": "did:peer:2.Ez6LSe3YyteKQVXSgGfZyCzbLq9K34zjAfap4EuaWJjbo5Gwk.Vz6MkvLvjZVc9jWJVs8M3qNZeRsnXXZGFjzcRJDH9vvdJMgHB.SeyJ0IjoiZG0iLCJzIjp7InVyaSI6Imh0dHBzOi8vdXMtZWFzdC5wcm92ZW4ubWVkaWF0b3IuaW5kaWNpb3RlY2guaW8vbWVzc2FnZSIsImEiOlsiZGlkY29tbS92MiIsImRpZGNvbW0vYWlwMjtlbnY9cmZjMTkiXX19.SeyJ0IjoiZG0iLCJzIjp7InVyaSI6IndzczovL3dzLnVzLWVhc3QucHJvdmVuLm1lZGlhdG9yLmluZGljaW90ZWNoLmlvL3dzIiwiYSI6WyJkaWRjb21tL3YyIiwiZGlkY29tbS9haXAyO2Vudj1yZmMxOSJdfX0.SeyJzIjogImh0dHBzOi8vdXMtZWFzdC5wcm92ZW4ubWVkaWF0b3IuaW5kaWNpb3RlY2guaW8vbWVzc2FnZSIsICJhIjogWyJkaWRjb21tL2FpcDEiLCJkaWRjb21tL2FpcDI7ZW52PXJmYzE5Il0sICJyZWNpcGllbnRLZXlzIjogWyIja2V5LTIiXSwgInQiOiAiZGlkLWNvbW11bmljYXRpb24ifQ.SeyJzIjogIndzczovL3dzLnVzLWVhc3QucHJvdmVuLm1lZGlhdG9yLmluZGljaW90ZWNoLmlvL3dzIiwgImEiOiBbImRpZGNvbW0vYWlwMSIsImRpZGNvbW0vYWlwMjtlbnY9cmZjMTkiXSwgInJlY2lwaWVudEtleXMiOiBbIiNrZXktMiJdLCAidCI6ICJkaWQtY29tbXVuaWNhdGlvbiJ9",
          "publicKeyBase58": "3NNpNLWYQ4iwBHCCgM5PWZ6ZDrC3xyduMvrppGxGMuAz"
        }
      ]
    }
```

KolbyRKunz:
@TimoGlastra That didDoc looks correct to me

packages/core/src/modules/connections/DidExchangeProtocol.ts
```diff
@@ -504,10 +504,13 @@ export class DidExchangeProtocol {
...
-    return isDid(message.did, 'peer') && getNumAlgoFromPeerDid(message.did) === PeerDidNumAlgo.GenesisDoc
-      ? this.extractAttachedDidDocument(agentContext, message, invitationKeysBase58)
-      : this.extractResolvableDidDocument(agentContext, message, invitationKeysBase58)
+    // Not all agents use didRotate yet, some may still send a didDoc attach with various did types
+    // we should check if the didDoc attach is there and if not require that the didRotate be present
+    if (message.didDoc) {
+      return this.extractAttachedDidDocument(agentContext, message, invitationKeysBase58)
+    } else {
+      return this.extractResolvableDidDocument(agentContext, message, invitationKeysBase58)
+    }
```
TimoGlastra, line 513: Sorry to be so nitpicky and pedandit, but I want to be careful with this, as it can have a lot of impact on security. 

If we allow any did document to be passed, it means that e.g. for a did:web document, I could provide it as attachment heren, and how this is now coded we will store the document as is. That means I could impersonate a did:web document, as we now don't need to resolve it. This means we bypass the ledger security layer most dids provide.

So I think we should implement the logic as follows:
- check if the did is resolvable using a did resolver (all dids except for did:peer:1)
  - if it is resolvable, and the did is the same as in the invitation we will resolve the DID, without requireing `did_rotate~attach` or `did_doc~attach`.
  - if it is resolvable, and the did is different from the did in the invitation, verify the `did_rotate~attach` has been signed with the invitation did (as we do now) and resolve the did
  - if it is not resolvable (so did:peer:1), we will extract the did from the `did_doc~attach` and verify the signature

I would be curious: in ACA-Py do you use `did_doc~attach` also with did:peer:4?

TimoGlastra:
> The document loader issue I was running into was that it would resolve a didDoc but the verification arrays would come back as empty services, they would only contain ids. Something like { id: "did:peer:2:someRandomDid#key-1"}, the actual keys would not be included making it impossible to sign the payload correctly.

Can you provide me with a sample input that would cause this? I think it is because of missing `@context` values. Would like to fix it with the JSON-LD loader and add some tests for it.

PR #1881, opened 2024-05-27
chore: updating dependabot file to support GHA, TS, JS, dev container and docker packages

rajpalc7:


rajpalc7:
Hi @WadeBarnes , this is ready for review now.

WadeBarnes (approved):
LGTM, but I'll leave the final review and approval to the regular maintainers.

genaris (changes requested):
Thanks @rajpalc7 ! Could you fix the prettier errors? It can be done by running `yarn format`.

TimoGlastra (changes requested):
Can we make sure to group all the dependencies into one PR? I'm really overwhelmed by the 100s of dependabot PR notifications I have to deal with on a daily basis and it distracts from more meaningful PRs

Also -- can we configure it to only make __patch__ updates for production dependencies? I'm not really intersted in updates for dev-deps and would rather update these manually once in a while.

Also -- is there a way to fully automate the merging of patches if the CI succeeds?

Issue #1007, opened 2022-09-02
Support pagination and offset when querying records


TimoGlastra:
Don't think offset is supported by indy-sdk but should be by aries-askar. We need to provide Apis in the storage service (and in higher level APIs) to support pagination and offset when querying records. This will become more important, especially with server deployments

genaris:
+1000 to this. Not only server but also in mobile deployments as well, as easily there could be thousands of records of single types, even filtered. For instance, if a wallet holds chat messages it's not convenient to load all of them and keep in React Native state (especially in low-end devices).

PR #1883, opened 2024-05-28
feat: add pagination params to storage service

mrlunin:
This is mainly an implementation for #1007.

Aries Askar supports pagination using offset and limit, so I added additional parameters `$offset` and `$limit` (with `$` to not overlap with some existing record tags) to the Storage service and added implementation to AskarStorageService.

I added this because we have server deployment and to avoid querying all records at once.


TimoGlastra (changes requested):
I like the simplicity of the solution, but i don't think it should be added as tags. But rather as separate options

packages/core/src/storage/StorageService.ts
```diff
@@ -12,7 +12,13 @@ interface AdvancedQuery<T extends BaseRecord> {
   $not?: Query<T>
 }
 
-export type Query<T extends BaseRecord<any, any, any>> = AdvancedQuery<T> | SimpleQuery<T>
+interface PaginationQuery {
```
genaris, line 15: I agree with @TimoGlastra in that maybe using an extra query options parameters would be cleaner. What about something like the following?

```ts
export type QueryOptions = {
  limit?: number
  offset?: number
}

findByQuery(agentContext: AgentContext, recordClass: BaseRecordConstructor<T>, query: Query<T>, queryOptions?: QueryOptions): Promise<T[]>
```

mrlunin:
@TimoGlastra @genaris I've added an additional parameter `QueryOptions` to the `findByQuery`. Can you check the code?

TimoGlastra (approved):
I like it! Thanks @mrlunin 🙏

TimoGlastra:
There's several tests failing @mrlunin, if you address these we can merge this :)

Issue #1874, opened 2024-05-20
Wrong schema Id is stored for credentials


sairanjit:
While storing the credential in wallet if the`did` of `schemaId` and `credDefId` are different then the wallet uses the did of `credDefId` for `schemaId`. 

- The issue is here https://github.com/openwallet-foundation/credo-ts/blob/main/packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts#L274
- Same `namespaceIdentifier` is passed to function `this.fetchIndySchemaWithSeqNo` but it should be of the `schemaId`

From where can we use the same `namespaceIdentifier` for `schemaId` ?

TimoGlastra:
Ah good catch. We should probably add tests where the schema creator is different from the cred def creator.

Would you be able to create a fix, and also add some tests?

sairanjit:
Yeah Sure @TimoGlastra

pallavighule:
@sairanjit @TimoGlastra  I am looking into this . I will raise the PR soon

PR #1884, opened 2024-05-29
fix: wrong schema id is stored for credentials

pallavighule:
### What  ###

- Fixed the issue that wrong schema id is stored for credentials
- Written a test case to register and resolve a credential definition (internal,credDefDid != schemaDid)

### DIFF
--- a/packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
+++ b/packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
@@ -287,9 +287,8 @@ export class IndyVdrAnonCredsRegistry implements AnonCredsRegistry {
       }
 
       // Format the schema id based on the type of the credential definition id
-      const unqualifiedDid = schema.schema.schemaId.split(':')[0]
       const schemaId = credentialDefinitionId.startsWith('did:indy')
-        ? getDidIndySchemaId(pool.indyNamespace, unqualifiedDid, schema.schema.name, schema.schema.version)
+        ? getDidIndySchemaId(pool.indyNamespace, schema.schema.issuerId, schema.schema.name, schema.schema.version)
         : schema.schema.schemaId
 
       return {
@@ -989,7 +988,7 @@ export class IndyVdrAnonCredsRegistry implements AnonCredsRegistry {
         attr_name: schema.data.attr_names,
         name: schema.data.name,
         version: schema.data.version,
-        issuerId: did,
+        issuerId: schemaDid,
         seqNo,
       },
       indyNamespace: pool.indyNamespace,
--- a/packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
+++ b/packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
@@ -979,8 +979,8 @@ export class IndyVdrAnonCredsRegistry implements AnonCredsRegistry {
     }
 
     const schema = response.result.data?.txn.data as SchemaType
-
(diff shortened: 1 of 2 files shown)


pallavighule:
@TimoGlastra @genaris  I have fixed the  issue #1874  .Please review the PR.

packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
```diff
@@ -979,8 +980,8 @@ export class IndyVdrAnonCredsRegistry implements AnonCredsRegistry {
     }
 
     const schema = response.result.data?.txn.data as SchemaType
-
-    const schemaId = getUnqualifiedSchemaId(did, schema.data.name, schema.data.version)
+    const schemaDid = response.result.data?.txn.metadata.from as string
+    const schemaId = getUnqualifiedSchemaId(schemaDid, schema.data.name, schema.data.version)
 
     return {
       schema: {
```
genaris, line 987: I understand from this code that the actual issuer Id for this schema is the one in the "from" field. So shouldn't we also put this value into the `issuerId` of the returned object?

packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
```diff
@@ -287,8 +287,9 @@ export class IndyVdrAnonCredsRegistry implements AnonCredsRegistry {
       }
 
       // Format the schema id based on the type of the credential definition id
+      const unqualifiedDid = schema.schema.schemaId.split(':')[0]
       const schemaId = credentialDefinitionId.startsWith('did:indy')
-        ? getDidIndySchemaId(pool.indyNamespace, namespaceIdentifier, schema.schema.name, schema.schema.version)
+        ? getDidIndySchemaId(pool.indyNamespace, unqualifiedDid, schema.schema.name, schema.schema.version)
```
genaris, line 292: If my comment below regarding returning the correct issuerId in `fetchIndySchemaWithSeqNo` is correct, I think we could simply `schema.schema.issuerId`, so there will not be need of creating the unqualifiedDid variable.

genaris (commented):
Thanks! I have a couple questions regarding the returned value of `fetchIndySchemaWithSeqNo` that could make the code even more simple (if I'm right).

packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
pallavighule: Yes, thank you for the suggestion. I will make the necessary changes.

packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
pallavighule: Yes right , I will make the necessary changes.

packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
pallavighule: @genaris  I have made the necessary changes

packages/indy-vdr/src/anoncreds/IndyVdrAnonCredsRegistry.ts
pallavighule: I have made the necessary changes

genaris (approved):
LGTM!

It would be good to get a double check by @KolbyRKunz (who also experienced this issue recently)

KolbyRKunz:
LGTM!

PR #1887, opened 2024-06-02
fix: unqualified indy revRegDefId in migration

### LINKED ISSUE #1882: Error deleting AnonCreds credentials after upgrading from 0.4.2 to 0.5.3
Credo raises an error when attempting to delete a credential in 0.5.3 that was received in 0.4.2

Error originates in @credo-ts/anoncreds V1CredentialProtocol.ts
`[CredoError: Unsupported credential record type anoncreds for v1 issue credential protocol (need w3c)]`

This occurs when calling `agent.credentials.deleteById(credential.id)` from Bifold. I might be wrong but I believe we were using V1 protocol with AnonCreds in Credo/AFJ 0.4.2 so I'm confused why it appears to no longer be supported.

Of `credentialRecord.credentials` the `credentialRecord` part seems to be deleted successfully but not the `.credentials` part. I think, higher up in Credo, it's failing during 'deleting associated credentials' loop [here](https://github.com/openwallet-foundation/credo-ts/blob/dcb363dc7e7abbbcce8b3a344a487a186eb5c88e/packages/core/src/modules/credentials/protocol/BaseCredentialProtocol.ts#L257)

Thank you for any help and let me know if I can provide more info or assist in any way

genaris:
Sadly, the fix done in 0.5.3 for Indy credentials migration was not enough for some Indy revocable credentials, since it is failing to apply qualified/unqualified revocation registry identifier tags into Credential Exchange record and, therefore, not effectively updating its bindings to the W3C credential.

This happens, for instance, when migrating from 0.4.2 to 0.5.3 the credentials received in [BCGov Showcase](https://digital.gov.bc.ca/digital-trust/showcase/) and attempting to delete the credentials associated to a credential exchange. 

Fixes #1882

cc @wadeking98 @bryce-mcmath

### DIFF
--- a/packages/anoncreds/src/models/exchange.ts
+++ b/packages/anoncreds/src/models/exchange.ts
@@ -52,6 +52,8 @@ export interface AnonCredsCredential {
   values: Record<string, AnonCredsCredentialValue>
   signature: unknown
   signature_correctness_proof: unknown
+  rev_reg?: unknown
+  witness?: unknown
 }
 
 export interface AnonCredsProof {
--- a/packages/anoncreds/src/updates/0.4-0.5/__tests__/w3cCredentialRecordMigration.test.ts
+++ b/packages/anoncreds/src/updates/0.4-0.5/__tests__/w3cCredentialRecordMigration.test.ts
@@ -294,7 +294,7 @@ async function testMigration(
   const anonCredsRecord = new AnonCredsCredentialRecord({
     credential: {
       schema_id: schemaId,
-      cred_def_id: credentialDefinitionState.credentialDefinitionId,
+      cred_def_id: credentialDefinitionId,
       values: {
         name: {
           raw: 'John',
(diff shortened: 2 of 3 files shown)


bryce-mcmath:
I added a patch with these changes today and so far deleting credentials appears to be working flawlessly now! Thank you for this, much appreciated

TimoGlastra (approved):
Seems to be some nasty bugs related to the unqualified/qualified identifiers and migration of anoncreds credentials.

Sorry for the inconvenience from our side in releasing this code too soon before it was properly tested, lesson learned! 

Thanks for looking into this @genaris! 🙌

TimoGlastra:
Hmm seems the DCO check is not working (also not in Askar repo)

genaris:
> **DCO ** Expected — Waiting for status to be reported

It seems this problem should be fixed shortly: https://github.com/dcoapp/app/issues/211

PR #1888, opened 2024-06-02
fix: WebSocket priority in Message Pick Up V2

### LINKED ISSUE #1877: Mediator in Live delivery mode not getting Live Mode sessions
When using the `MediatorPickupStrategy` as `PickUpV2LiveMode` the mobile agent is not able to receive the messages immediately

- @genaris I think we are just monitoring the websocket events but not connecting the to mediator with websocket here - https://github.com/openwallet-foundation/credo-ts/blob/main/packages/core/src/modules/routing/MediationRecipientApi.ts#L270

genaris:
Pick Up V2 Live Mode requires the use of a long-lived transport session, and currently we are not checking this. So here we are making it mandatory for Live mode and preferred for Pick Up V1/V2 'manual' modes.

Fixes #1877


packages/core/src/modules/message-pickup/MessagePickupApi.ts
```diff
@@ -255,18 +257,19 @@ export class MessagePickupApi<MPPs extends MessagePickupProtocol[] = [V1MessageP
...
   public async setLiveDeliveryMode(options: SetLiveDeliveryModeOptions): Promise<SetLiveDeliveryModeReturnType> {
     const connectionRecord = await this.connectionService.getById(this.agentContext, options.connectionId)
-
     const protocol = this.getProtocol(options.protocolVersion)
     const { message } = await protocol.setLiveDeliveryMode(this.agentContext, {
       connectionRecord,
       liveDelivery: options.liveDelivery,
     })
 
+    // Live mode requires a long-lived transport session, so we'll require WebSockets to send this message
```
TimoGlastra, line 266: Not for now, but would be good if we can configure this on the transport class, so we can add the requirement here and filter dynamicslly

TimoGlastra:
Is this ready for review / to be merged @genaris ?

genaris:
> Is this ready for review / to be merged @genaris ?

Yes, I think it's fine. I wanted to do something fancy with the tests but didn't have time these days.

TimoGlastra (approved):

PR #1891, opened 2024-06-06
fix: AnonCreds proof requests with unqualified dids

genaris:
Following issues found during integration with BCGov Wallet, where AnonCreds with unqualified identifiers are being used, here we try to adjust the identifiers used in a Present Proof flow depending on the proof request contents.

Basically, we analyze whether there is any restriction that contains an unqualified did and, if true, assume that we should use unqualified identifiers. Otherwise, we use the default ones (qualified).

I'm testing with BCGov wallet's [main](https://github.com/bcgov/bc-wallet-mobile) branch in both verifier and holder side and seems to work fine (at least with a basic presentation request using Person credential from [BC Wallet Showcase](https://digital.gov.bc.ca/digital-trust/showcase/).

### DIFF
--- a/.github/workflows/continuous-deployment.yml
+++ b/.github/workflows/continuous-deployment.yml
@@ -84,13 +84,13 @@ jobs:
           echo version="${NEW_VERSION}" >> "$GITHUB_OUTPUT"
 
       - name: Create Tag
-        uses: mathieudutour/github-tag-action@v6.1
+        uses: mathieudutour/github-tag-action@v6.2
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           custom_tag: ${{ steps.new-version.outputs.version }}
 
       - name: Create Release
-        uses: softprops/action-gh-release@v1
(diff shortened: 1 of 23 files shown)


genaris:
cc @bryce-mcmath @wadeking98 hopefully this can solve the remaining issues with that awful thing popularly known as 'unqualified hell'.

TimoGlastra (approved):
Another great fix, thanks @genaris! 

Do we need a simple test to cover this? (I'm fine with merging it as is to get it fixed ASAP)

genaris:
> Another great fix, thanks @genaris!
> 
> Do we need a simple test to cover this? (I'm fine with merging it as is to get it fixed ASAP)

I've added some based on `v2-indy-proofs.e2e.test.ts` (basically a copy-paste, maybe we can improve it in a future when reorganizing core modules). They fail when running with previous code and OK with these fixes, so it seems we are OK!

I'd like to wait for @bryce-mcmath feedback before merging to make sure we are handling all relevant cases.

PR #1893, opened 2024-06-07
fix(anoncreds): combine creds into one proof

TimoGlastra:
Combines attributes and predicates from the same credential into one proof. This way the verifier can assert that e.g. a predicate of age > 15 and attribute of name are from the same cred.

Based on discussion in https://github.com/hyperledger/anoncreds-rs/issues/292

Would be great if there's a bit more control of it in the proof request syntax, as you now have to parse the presentation/proof structure to know whether multiple attributes/predicates come from the same credential (which I suspect no-one does, so request a predicate and attribute at the same time is actually not very secure it seems as you don't know whether they are from the same cred).


packages/anoncreds/src/anoncreds-rs/AnonCredsRsHolderService.ts
```diff
@@ -199,21 +200,52 @@ export class AnonCredsRsHolderService implements AnonCredsHolderService {
...
-        credentials.push(await credentialEntryFromAttribute(attribute))
-        credentialsProve.push({ entryIndex, isPredicate: false, referent, reveal: attribute.revealed })
-        entryIndex = entryIndex + 1
+
+        // If the credentialId with the same timestamp is already present, we will use the existing entry, so that the proof is created
+        // showing the attributes come from the same cred, rather than different ones.
+        const existingCredentialIndex = credentials.findIndex(
+          (credential) =>
+            credential.credentialId === attribute.credentialId &&
+            attribute.timestamp === credential.credentialEntry.timestamp
```
berendsliedrecht, line 214: Won't the `credentialId` suffice? Why is the timestamp included?

berendsliedrecht (approved):

packages/anoncreds/src/anoncreds-rs/AnonCredsRsHolderService.ts
TimoGlastra: I wasn't 100 sure, but as we create the revocation state based on the timestamp, it would be possible to have different requirements.

E.g. age > 15 nonRevoked between 100 and 1000 (timestamp)
and name = Berend. nonRevoked between 10.000 and 100.000 (timestamp)..

These would create different revocation states, and therefore I decided to also include timestamp. It's very very rare to have different timestamps, but this at least makes sure the presentation is valid

berendsliedrecht: Ah okay, makes sense! good catch!

PR #1894, opened 2024-06-08
feat: add message handler middleware and fallback

### LINKED ISSUE #97: add agent middleware
As a developer I sometimes want to apply custom logic for all incoming messages. We already have handlers that allow to implement logic for a custom message type. I think it would be beneficial to introduce the concept of middleware into aries-framework-javascript.

aries-framework-dotnet also implements the concept of middleware which we can take as [an example](https://github.com/hyperledger/aries-framework-dotnet/blob/fb5d3081f2740c9b02e7bfbc82a9c61f6de4a936/src/Hyperledger.Aries/Agents/Abstractions/IAgentMiddleware.cs)

### Checklist

- [ ] create `AgentMiddleware` interface
- [ ] call middleware before/after passing to handler (see question below)
    - I think `Dispatcher` would be the best place to call middleware
- [ ] add function to register middleware to `Agent` (like`registerHandlers`)
    - one function to register them all, or function that can be called multiple times?

### Questions

- When should middleware be called?
    - Before passing to handler / after passing to handler / both?
    - .NET runs the middleware after passing a message to the handler. 

### Use cases

Probably a lot more, but this are two I can think of at the moment.

- [#96: Check connection acknowledgement for all messages, not just trust ping and ack](https://github.com/hyperledger/aries-framework-javascript/issues/96)
- [Message Tracing](https://github.com/hyperledger/aries-rfcs/blob/master/features/0034-message-tracing/README.md)

TimoGlastra:
This PR adds two things:
- middleware for the message handler
- fallback message handler for when no message handler is registered for the incoming message type

You could quite easily solve 2 (fallback message handler) with 1 (middleware), but it's meant to be a convenience method so you can easily handle incoming messages without a handler (i suspect for a lot of use case it should suffice). But the middleware gives that extra power to e.g. modify the incoming message. 

It would be nice to also start adopting the middleware in more places, e.g. the message sender (to modify messages before they are sent).

Our use case for adding this at the moment is that we want to handle messages that have no message class supported. I haven't added any top-level APIs yet (only on the dependency manager) as it's good to fiddle around for a bit maybe first. One thing I think that might be worth considering is adding a top level method where you can register a handler for incoming messages. So you can 'extend' the framework without having to write custom classes. E.g. a didcommCustomMessaging module, where you can send and receive messages. 

Then this work would be more the internal API to hook into the receiving / sending process but for most cases we can just expose it as an easy to use API.

The API can be used as follows at the moment:

```ts
agent.dependecyManager.registerMessageHandlerMiddleware(async (inboundMessageContext: InboundMessageContext, next) => {
// this middleware will just silently ignore any incoming messages without a handler
  if (!inboundMessageContext.messageHandler) {
   console.log('no message handler registered').

   // if returned, and next is not called, other middleware WON'T be called
   return
 }

  await next()
})
```

And for the fallback handler

```ts
agent.dependecyManager.setFallbackMessageHandler(async (messageContext: InboundMessageContext) => {

  await writeMessageToDatabase(inboundMessageContext.message)

  // reply with the same message that we received, because why not? :)
  return getOutboundMessageContext(inboundMessageContext.agentContext, {
    message: inboundMessageContext.message,
   connectionRecord: inboundMessageContext.connection,
 })
})
```

No breaking changes

Fixes #97 

cc @Tommylans

### DIFF
--- a/packages/core/src/agent/Dispatcher.ts
+++ b/packages/core/src/agent/Dispatcher.ts
@@ -38,9 +38,14 @@ class Dispatcher {
   }
 
   private defaultHandlerMiddleware: MessageHandlerMiddleware = async (inboundMessageContext, next) => {
-    const messageHandler =
-      inboundMessageContext.messageHandler ??
-      inboundMessageContext.agentContext.dependencyManager.fallbackMessageHandler
+    let messageHandler = inboundMessageContext.messageHandler
+
+    if (!messageHandler && inboundMessageContext.agentContext.dependencyManager.fallbackMessageHandler) {
+      messageHandler = {
+        supportedMessages: [],
+        handle: inboundMessageContext.agentContext.dependencyManager.fallbackMessageHandler,
+      }
+    }
 
     if (!messageHandler) {
       throw new ProblemReportError(
--- a/packages/core/src/agent/Dispatcher.ts
+++ b/packages/core/src/agent/Dispatcher.ts
@@ -1,16 +1,19 @@
 import type { AgentMessage } from './AgentMessage'
 import type { AgentMessageProcessedEvent } from './Events'
+import type { MessageHandlerMiddleware } from './MessageHandlerMiddleware'
 import type { InboundMessageContext } from './models/InboundMessageContext'
 
 import { InjectionSymbols } from '../constants'
-import { CredoError } from '../error/CredoError'
+import { CredoError } from '../error'
 import { Logger } from '../logger'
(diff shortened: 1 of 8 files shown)


genaris (approved):
Great job @TimoGlastra ! I think I can use it to ignore duplicated incoming messages: I can define it as a middleware that I'll use initially on my app but can also contribute back as an extension middleware that anybody can use if they want.

Looking forward for outbound message middleware support as well. We can use them to set some optional attributes of DIDComm messages (specially V2) such as creation time.

PR #1896, opened 2024-06-11
fix: Moved the middleware to the MessageHandlerRegistry

Tommylans:
The fallback message handler and the middleware were lost in the tenant agent. I moved the middleware and the fallback handler to the `MessageHandlerRegistry` so it will be available in the child container.


TimoGlastra (commented):
Makes sense actually to put them there! :)

PR #1901, opened 2024-06-16
chore: changesets and pnpm

TimoGlastra:
In an attempt to fix the wrong alpha publishing (under 0.5.1-alpha, which should be 0.5.4-alpha) and to simplify the publishing setup I took a look at [changesets](https://github.com/changesets/changesets). Changesets is simliar to conventional commits, but requires you to be more specific in which changes should be included in the changelog (so not every commit is nessearily a change that needs to be described).

I also updated to pnpm as it gives us the tools to work with monorepos and remove the need for lerna. I also updated TS as there were some fixes in the latest TypeScript about type references (which we raised because of stricter PNPM dependency installs).

In a next PR I'd like to replace ESlint and Prettier with Biome. We've been slowly migrating all our repositories to Biome and it's A LOT faster, simpler to setup, and only requires 1 dep instead of multiple + plugins.

The changeset setup we have now, will also allow us to do pre-release branch again, which is great for a next major release, so we can keep publishing on 0.5. If we remove the need for squash merging, and enable merge commits, we will also have an easier time in keeping the branches in sync (and because conventional commits are not required, anymore less need to keep the commit history clean).

A PR will be opened every time a change is ready to be puslihed, and whenver we're ready to release it, we can just merge the PR. We will still have alpha releases on every commit -- however they are formatted differently, e.g. `0.6.0-alpha-20240616132541`. This is so we don't have to look at NPM for the last released version, but we just include the timestamp (latest timestamp is latest alpha).


TimoGlastra:
As you can see, there's also a new bot that will inform users whether a changeset has been added. If not needed, you can ignore it. It also adds a link that easy allows us to create a PR with a changeset, so a contributor doesn't have to do it

packages/bbs-signatures/README.md
```diff
@@ -52,7 +53,7 @@ When using Credo inside the React Native environment, temporarily, a dependency
     }
 ```
 
-#### npm
+#### npm / pnpm
 
 ```diff
 +   "overrides": {
```
berendsliedrecht, line 59: pnpm needs `pnpm.overrides.XXX`

package.json
```diff
@@ -13,29 +13,30 @@
...
   "devDependencies": {
+    "@changesets/cli": "^2.27.5",
     "@hyperledger/aries-askar-nodejs": "^0.2.1",
+    "@jest/types": "^29.6.3",
     "@types/bn.js": "^5.1.5",
     "@types/cors": "^2.8.10",
     "@types/eslint": "^8.21.2",
     "@types/express": "^4.17.13",
-    "@types/jest": "^29.5.11",
+    "@types/jest": "^29.5.12",
```
berendsliedrecht, line 39: types/jest and jest/types are both required?

package.json
```diff
@@ -54,14 +55,14 @@
     "eslint-plugin-workspaces": "^0.8.0",
     "express": "^4.17.1",
     "jest": "^29.7.0",
-    "lerna": "^6.5.1",
+    "lerna": "^8.1.3",
```
berendsliedrecht, line 58: Can be removed

package.json
TimoGlastra: yes, sadly. Was getting issues with one of them not being defined (I think this has to do with the strict dependency resolution from pnpm, as `@jest/types` was not needed before)

berendsliedrecht (approved):

TimoGlastra:
@genaris are you also OK with leveraging changesets and pnpm? As this is a bit of a bigger restructure would like your input on this as well :)

PR #1903, opened 2024-06-16
fix: pex query fix

TimoGlastra:
A small fix from #1865 to make it easier to get the remaining issues of that PR addressed


packages/core/src/modules/sd-jwt-vc/repository/SdJwtVcRecord.ts
```diff
@@ -5,6 +5,7 @@ import type { Constructable } from '../../../utils/mixins'
 import { decodeSdJwtSync } from '@sd-jwt/decode'
 
 import { BaseRecord } from '../../../storage/BaseRecord'
+import { Query, SimpleQuery } from '../../../storage/StorageService'
```
genaris, line 8: It seems to be unused

genaris (approved):

PR #1913, opened 2024-06-21
fix: minor should be patch

TimoGlastra:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -35,7 +35,6 @@ jobs:
         run: pnpm install --frozen-lockfile
 
       - name: Create Release Pull Request or Publish to npm
-        if: github.ref == 'refs/heads/main'
         id: changesets
         uses: changesets/action@v1
         with:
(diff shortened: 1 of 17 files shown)


.github/workflows/release.yml
```diff
@@ -80,6 +79,13 @@ jobs:
           pnpm changeset version --snapshot alpha
           pnpm build
           pnpm changeset publish --tag alpha
+
+          CURRENT_PACKAGE_VERSION=$(node -p "require('./packages/core/package.json').version")
```
berendsliedrecht, line 83: Ah thats a bit meh, but if it works we can clean it up later if we want to.

berendsliedrecht (approved):

.github/workflows/release.yml
TimoGlastra: Yeah it's a bit meh, they don't create a tag/commit when you do a snapshot (it's seems to be unusual to create tags for that)

PR #1920, opened 2024-06-25
build: fix build issue in 0.5.5

TimoGlastra:
Fixes the build issue introduced in 0.5.5. We'll have to do a 0.5.6 release sadly.

It does add a few nice improvements though:
- now also includes src in package.json
- source maps point to source files, so whenever you now click on a type / method in e.g. vscode from credo you will be shown the source TS code rather than the build js code. Making debugging a lot easier I'd say :)

### DIFF
--- a/.changeset/silly-cooks-compete.md
+++ b/.changeset/silly-cooks-compete.md
@@ -1,0 +1,18 @@
+---
+'@credo-ts/indy-sdk-to-askar-migration': patch
+'@credo-ts/question-answer': patch
+'@credo-ts/bbs-signatures': patch
+'@credo-ts/react-native': patch
+'@credo-ts/action-menu': patch
+'@credo-ts/anoncreds': patch
+'@credo-ts/openid4vc': patch
+'@credo-ts/indy-vdr': patch
+'@credo-ts/tenants': patch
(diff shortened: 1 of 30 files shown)


berendsliedrecht (approved):
How will this influence the bundle size? Including the `src` directory can have quite an impact, right?

TimoGlastra:
Yeah it has some impact on the bundle size as you include the src, but I think it's well worth it because of the goto source now actually works and that makes debugging a lot easier.

berendsliedrecht:
> Yeah it has some impact on the bundle size as you include the src, but I think it's well worth it because of the goto source now actually works and that makes debugging a lot easier.

If it is included in the final production bundle, I am not sure I think it is worth it. Unpacked size of `@credo-ts/core` is already 4.7MB. Now, with the `src` it will be more than double so likely something around 10MB (for just core!) and if we also include atleast `libaskar.so` we will add another 5-7 MB... This is getting really big IMO.

TimoGlastra:
Okay -- now released 0.5.6 as there was a build issue with 0.5.5. I'll look into it again. I like it when packages publish source code, but if it's not desired it's easy to not do it

PR #1922, opened 2024-06-25
chore: sign-off release commit and do not publish src

TimoGlastra:
- do not publish src
- sign-off release commits
- create single tag and github release
- do not skip ci


.changeset/commit.js
```diff
@@ -0,0 +1,24 @@
+const { execSync } = require('node:child_process')
```
berendsliedrecht, line 1: AH thats annoying... Does changesets still offer enough value if we have to do things like this?

berendsliedrecht (approved):

.changeset/commit.js
TimoGlastra: Not sure -- it seems the amount of scripts is now similar to our previous approach. It is a bit less hacky which I like. But I'll keep looking further to an alternative. Maybe NX, or back to the lerna setup.

PR #1938, opened 2024-07-02
feat(askar): create keys in the secure element

berendsliedrecht:
- Waiting for the GH release of askar.

### DIFF
--- a/packages/askar/src/wallet/AskarBaseWallet.ts
+++ b/packages/askar/src/wallet/AskarBaseWallet.ts
@@ -24,6 +24,7 @@ import {
   WalletError,
   Key,
   TypedArrayEncoder,
+  KeyBackend,
 } from '@credo-ts/core'
 import { CryptoBox, Store, Key as AskarKey, keyAlgFromString } from '@hyperledger/aries-askar-shared'
 import BigNumber from 'bn.js'
@@ -35,6 +36,7 @@ import {
   isKeyTypeSupportedByAskarForPurpose,
   keyTypesSupportedByAskar,
 } from '../utils'
+import { convertToAskarKeyBackend } from '../utils/askarKeyBackend'
 
 import { didcommV1Pack, didcommV1Unpack } from './didcommV1'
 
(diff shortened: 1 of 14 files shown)


packages/core/src/crypto/KeyBackend.ts
```diff
@@ -0,0 +1,4 @@
+export enum KeyBackend {
+  Software = 'sofware',
+  SecureElement = 'secure_element',
+}
```
TimoGlastra, line 4:
```suggestion
export enum KeyBackend {
  Software = 'sofware',
  /**
  * Generate the key in a secure element. Need to be supported by the Wallet implementation. If using Aries Askar, Secure Element is available on iOS and Android in React Native
  */
  SecureElement = 'secure_element',
}
```

packages/askar/src/wallet/AskarBaseWallet.ts
```diff
@@ -125,7 +126,12 @@ export abstract class AskarBaseWallet implements Wallet {
    * Create a key with an optional seed and keyType.
    * The keypair is also automatically stored in the wallet afterwards
    */
-  public async createKey({ seed, privateKey, keyType }: WalletCreateKeyOptions): Promise<Key> {
+  public async createKey({
+    seed,
+    privateKey,
+    keyType,
+    keyBackend = KeyBackend.Software,
```
TimoGlastra, line 133: We don't pass it?

TimoGlastra (changes requested):

packages/askar/src/wallet/AskarBaseWallet.ts
berendsliedrecht: Need to wait for the library update :). Wanted to have to PR ready as I was waiting.

packages/core/src/crypto/KeyBackend.ts
berendsliedrecht: I redefined the enum in credo so it would not be askar specific. I will add a comment to the KeyBackend enum, but I will slightly reword it.

genaris: But the names `sofware` and `secure_element' are hardcoded in Askar, right? So in some way it depends on Askar (otherwise it would not be snake_case :smile: ).

Note the typo in 'sofware' (missing t).

berendsliedrecht: `software` and `secure_element` are also used by askar as it would not make sense to choose something else. If we want to support other providers which define different identifiers for their backend we can create a mapping function.

For now, this is fine IMO. (good catch as well on the `sofware` ;) ).

TimoGlastra: > as it would not make sense to choose something else

It would, as Credo API != Askar API

TimoGlastra (approved):
noice noice

PR #1944, opened 2024-07-08
fix(dids): allow to pass in a key instance when creating a did document

berendsliedrecht:

### DIFF
--- a/packages/core/src/modules/dids/methods/peer/PeerDidRegistrar.ts
+++ b/packages/core/src/modules/dids/methods/peer/PeerDidRegistrar.ts
@@ -1,5 +1,5 @@
 import type { AgentContext } from '../../../../agent'
-import type { KeyType } from '../../../../crypto'
+import type { Key, KeyType } from '../../../../crypto'
 import type { Buffer } from '../../../../utils'
 import type { DidRegistrar } from '../../domain/DidRegistrar'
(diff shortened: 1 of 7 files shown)


packages/core/src/modules/dids/methods/peer/PeerDidRegistrar.ts
```diff
@@ -37,23 +37,27 @@ export class PeerDidRegistrar implements DidRegistrar {
...
+
+        if (!key && keyType) {
+          key = await agentContext.wallet.createKey({
+            keyType,
+            seed,
+            privateKey,
+          })
+        }
+
+        if (!key) {
```
TimoGlastra, line 50: need to handle where both keyType and key are passed, but they don't match

TimoGlastra: also need to handle the case where privateKey/seed is passed and key as well. That should cause an error. As it results in ambigious behaviour (I have no idea without looking at the code which will take precedence)

TimoGlastra: (same for all files)

TimoGlastra (changes requested):

TimoGlastra:
can you also add a changest?

berendsliedrecht:
> can you also add a changest?

how?

TimoGlastra:
image

TimoGlastra (approved):

PR #1948, opened 2024-07-16
fix: check handshake protocols array is not empty

genaris:
This is related to https://github.com/bcgov/vc-authn-oidc/issues/583, where apparently ACA-Py is returning a present but empty handshake_protocols array in connection-less use case. I cannot think of a valid use case for an empty array so I think it is correct to consider it as undefined if it comes like that.

### DIFF
--- a/packages/core/src/modules/oob/OutOfBandApi.ts
+++ b/packages/core/src/modules/oob/OutOfBandApi.ts
@@ -557,7 +557,7 @@ export class OutOfBandApi {
           `Connection already exists and reuse is enabled. Reusing an existing connection with ID ${existingConnection.id}.`
         )
 
-        if (!messages) {
+        if (!messages || messages?.length === 0) {
           this.logger.debug('Out of band message does not contain any request messages.')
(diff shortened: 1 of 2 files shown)


packages/core/src/modules/oob/OutOfBandApi.ts
```diff
@@ -548,7 +548,7 @@ export class OutOfBandApi {
 
     await this.outOfBandService.updateState(this.agentContext, outOfBandRecord, OutOfBandState.PrepareResponse)
 
-    if (handshakeProtocols) {
+    if (handshakeProtocols && handshakeProtocols.length > 0) {
```
TimoGlastra, line 551: Should we do the same with attached requests for consistency then?

TimoGlastra (approved):

packages/core/src/modules/oob/OutOfBandApi.ts
genaris: Good point!

PR #1954, opened 2024-07-17
feat: AnonCreds revocation registry and statuslist

Tommylans:
The implementation is based on the logic described in the cheqd documentation about the registry and the status list https://docs.cheqd.io/product/advanced/anoncreds/revocation-status-list#tying-creddef-revregdef-and-statuslist-objects-together


packages/cheqd/src/anoncreds/services/CheqdAnonCredsRegistry.ts
```diff
@@ -305,24 +366,35 @@ export class CheqdAnonCredsRegistry implements AnonCredsRegistry {
...
+        throw new CredoError(`Revocation registry definition not found for id: ${revocationRegistryId}`)
+
+      const revocationRegistryDefinitionName = revocationRegistryDefinition.revocationRegistryDefinitionMetadata.name
+      if (!revocationRegistryDefinitionName)
+        throw new CredoError(`Revocation registry definition name not found for id: ${revocationRegistryId}`)
+
       const response = await cheqdDidResolver.resolveResource(
         agentContext,
-        `${revocationRegistryId}&resourceType=anonCredsStatusList&resourceVersionTime=${timestamp}`
+        `${parsedDid.did}?resourceType=anonCredsStatusList&resourceVersionTime=${timestamp}&resourceName=${revocationRegistryDefinitionName}`
```
TimoGlastra, line 388: Does the revocation registry definition name need to be unique? couldn't we use the revocationRegistryId as the name of the status list?

packages/cheqd/src/anoncreds/services/CheqdAnonCredsRegistry.ts
```diff
@@ -349,7 +421,79 @@ export class CheqdAnonCredsRegistry implements AnonCredsRegistry {
...
+
+      const response = await cheqdDidRegistrar.createResource(
+        agentContext,
+        revocationStatusList.issuerId,
+        revocationStatusListResource
+      )
+      if (response.resourceState.state !== 'finished')
+        throw new CredoError(response.resourceState.reason ?? 'Unknown error')
+
+      // It's not possible to get the timestamp from the response, so we set it to the current time
```
TimoGlastra, line 464: Why not? The timestamp is important for some pieces, so I think we need to get the timestamp is will be used when you resolve the status list (then we can get the timestamp using response.resourceMetadata.created.getTime()?)

TimoGlastra (approved):

packages/cheqd/src/anoncreds/services/CheqdAnonCredsRegistry.ts
Tommylans: Based on the documentation of cheqd about anoncreds revocation. The statuslist and the revocation registry definition name needs to be the same as the credential definition name. Which is currently the case in the implementation.

packages/cheqd/src/anoncreds/services/CheqdAnonCredsRegistry.ts
Tommylans: The cheqd sdk currently doesn't return the creation date when you create a new resource, but it's returned when you are querying a resource. So the only way is to query it just after creation? Also when you look at the raw response of the cheqd sdk of the creation it doesn't return a creation.

But for fetching the statuslist it only needs a timestamp which is close to the resource it doesn't really have to be exact.

So what do you think we should do in this case?

PR #1966, opened 2024-07-25
feat: dpop support

auer-martin:
blocked by for now https://github.com/Sphereon-Opensource/OID4VC/pull/131

### DIFF
--- a/packages/openid4vc/src/openid4vc-holder/OpenId4VcHolderApi.ts
+++ b/packages/openid4vc/src/openid4vc-holder/OpenId4VcHolderApi.ts
@@ -149,9 +149,9 @@ export class OpenId4VcHolderApi {
     const {
       access_token: accessToken,
       c_nonce: cNonce,
-      dPoPJwk,
+      dpop,
     } = await this.openId4VciHolderService.requestAccessToken(this.agentContext, options)
-    return { accessToken, cNonce, dPoPJwk }
+    return { accessToken, cNonce, dpop }
   }
 
   /**
@@ -161,14 +161,14 @@ export class OpenId4VcHolderApi {
    * @param options.tokenResponse Obtained through @see requestAccessToken
    */
   public async requestCredentials(options: OpenId4VciRequestCredentialOptions) {
-    const { resolvedCredentialOffer, cNonce, accessToken, dPoPJwk, ...credentialRequestOptions } = options
+    const { resolvedCredentialOffer, cNonce, accessToken, dpop, ...credentialRequestOptions } = options
 
     return this.openId4VciHolderService.acceptCredentialOffer(this.agentContext, {
       resolvedCredentialOffer,
       acceptCredentialOfferOptions: credentialRequestOptions,
       accessToken,
       cNonce,
-      dPoPJwk,
+      dpop,
     })
   }
 
--- a/packages/openid4vc/src/openid4vc-holder/OpenId4VcHolderApi.ts
+++ b/packages/openid4vc/src/openid4vc-holder/OpenId4VcHolderApi.ts
@@ -146,11 +146,12 @@ export class OpenId4VcHolderApi {
    * @param options.code The authorization code obtained via the authorization request URI
    */
   public async requestToken(options: OpenId4VciRequestTokenOptions): Promise<OpenId4VciRequestTokenResponse> {
-    const { access_token: accessToken, c_nonce: cNonce } = await this.openId4VciHolderService.requestAccessToken(
-      this.agentContext,
-      options
-    )
-    return { accessToken, cNonce }
+    const {
+      access_token: accessToken,
+      c_nonce: cNonce,
+      dPoPJwk,
+    } = await this.openId4VciHolderService.requestAccessToken(this.agentContext, options)
+    return { accessToken, cNonce, dPoPJwk }
   }
 
   /**
@@ -160,13 +161,14 @@ export class OpenId4VcHolderApi {
    * @param options.tokenResponse Obtained through @see requestAccessToken
    */
   public async requestCredentials(options: OpenId4VciRequestCredentialOptions) {
-    const { resolvedCredentialOffer, cNonce, accessToken, ...credentialRequestOptions } = options
+    const { resolvedCredentialOffer, cNonce, accessToken, dPoPJwk, ...credentialRequestOptions } = options
 
     return this.openId4VciHolderService.acceptCredentialOffer(this.agentContext, {
       resolvedCredentialOffer,
       acceptCredentialOfferOptions: credentialRequestOptions,
       accessToken,
       cNonce,
+      dPoPJwk,
     })
   }
 
--- a/packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
+++ b/packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
@@ -424,6 +424,7 @@ export class OpenId4VciHolderService {
       const credential = await this.handleCredentialResponse(agentContext, credentialResponse, {
         verifyCredentialStatus: verifyCredentialStatus ?? false,
         credentialIssuerMetadata: metadata.credentialIssuerMetadata,
+        format: offeredCredentialConfiguration.format as OpenId4VciCredentialFormatProfile,
       })
 
       this.logger.debug('Full credential', credential)
@@ -615,6 +616,7 @@ export class OpenId4VciHolderService {
     options: {
       verifyCredentialStatus: boolean
       credentialIssuerMetadata: OpenId4VciIssuerMetadata
+      format: OpenId4VciCredentialFormatProfile
     }
   ): Promise<OpenId4VciCredentialResponse> {
     const { verifyCredentialStatus, credentialIssuerMetadata } = options
@@ -635,7 +637,7 @@ export class OpenId4VciHolderService {
           }
         : undefined
 
-    const format = credentialResponse.successBody.format
+    const format = options.format
     if (format === OpenId4VciCredentialFormatProfile.SdJwtVc) {
       if (typeof credentialResponse.successBody.credential !== 'string')
         throw new CredoError(
--- a/packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
+++ b/packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
@@ -69,6 +69,7 @@ import {
   PARMode,
   post,
   EndpointMetadataResult,
+  DPoPResponseParams,
 } from '@sphereon/oid4vci-common'
 
 import { OpenId4VciCredentialFormatProfile } from '../shared'
@@ -313,7 +314,7 @@ export class OpenId4VciHolderService {
     const { metadata, credentialOfferRequestWithBaseUrl } = resolvedCredentialOffer
 
     // acquire the access token
-    let accessTokenResponse: OpenIDResponse<AccessTokenResponse>
+    let accessTokenResponse: OpenIDResponse<AccessTokenResponse, DPoPResponseParams>
 
     const accessTokenClient = new AccessTokenClient()
 
@@ -354,7 +355,10 @@ export class OpenId4VciHolderService {
 
     this.logger.debug('Requested OpenId4VCI Access Token.')
 
-    return { ...accessTokenResponse.successBody, ...(dpopJwk && { dpop: { dpopJwk: dpopJwk } }) }
+    return {
+      ...accessTokenResponse.successBody,
+      ...(dpopJwk && { dpop: { dpopJwk: dpopJwk, dpopNonce: accessTokenResponse.params?.dpop?.dpopNonce } }),
+    }
   }
 
   public async acceptCredentialOffer(
@@ -365,7 +369,7 @@ export class OpenId4VciHolderService {
       resolvedAuthorizationRequestWithCode?: OpenId4VciResolvedAuthorizationRequestWithCode
       accessToken?: string
       cNonce?: string
-      dpop?: { dpopJwk: Jwk }
+      dpop?: { dpopJwk: Jwk; dpopNonce?: string }
     }
   ) {
     const { resolvedCredentialOffer, acceptCredentialOfferOptions } = options
@@ -410,7 +414,7 @@ export class OpenId4VciHolderService {
       ? {
           access_token: options.accessToken,
           c_nonce: options.cNonce,
-          ...(options.dpop && { dpop: { dpopJwk: options.dpop.dpopJwk } }),
+          ...(options.dpop && { dpop: { dpopJwk: options.dpop.dpopJwk, dpopNonce: options.dpop?.dpopNonce } }),
         }
       : await this.requestAccessToken(agentContext, tokenRequestOptions)
 
@@ -484,7 +488,7 @@ export class OpenId4VciHolderService {
 
         createDpopOpts = {
           jwtIssuer: { alg: alg as unknown as SigningAlgo, jwk: jwk.toJson() },
-          jwtPayloadProps: { accessToken: options.accessToken },
+          jwtPayloadProps: { accessToken: tokenResponse.access_token, nonce: tokenResponse.dpop?.dpopNonce },
           createJwtCallback: getCreateJwtCallback(agentContext),
         }
       }
@@ -690,7 +694,7 @@ export class OpenId4VciHolderService {
 
   private async handleCredentialResponse(
     agentContext: AgentContext,
-    credentialResponse: OpenIDResponse<CredentialResponse>,
+    credentialResponse: OpenIDResponse<CredentialResponse, DPoPResponseParams>,
     options: {
       verifyCredentialStatus: boolean
       credentialIssuerMetadata: OpenId4VciIssuerMetadata
--- a/packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
+++ b/packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
@@ -285,9 +285,7 @@ export class OpenId4VciHolderService {
 
     if (!dPoPSigningAlgValuesSupported) return undefined
 
-    const alg = dPoPSigningAlgValuesSupported
-      .flatMap((alg) => alg)
-      .find((alg) => getJwkClassFromJwaSignatureAlgorithm(alg))
+    const alg = dPoPSigningAlgValuesSupported.find((alg) => getJwkClassFromJwaSignatureAlgorithm(alg))
 
     const JwkClass = alg ? getJwkClassFromJwaSignatureAlgorithm(alg) : undefined
 
@@ -323,7 +321,6 @@ export class OpenId4VciHolderService {
     const createDPoPOptions = await this.getCreateDPoPOptions(agentContext, metadata)
     const dPoPJwk = createDPoPOptions ? getJwkFromJson(createDPoPOptions.jwtIssuer.jwk) : undefined
 
-    resolvedCredentialOffer.metadata.credentialIssuerMetadata
     if (resolvedAuthorizationRequest) {
       const { codeVerifier, redirectUri } = resolvedAuthorizationRequest
       accessTokenResponse = await accessTokenClient.acquireAccessToken({
--- a/packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
+++ b/packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
@@ -17,19 +17,13 @@ import type {
   OpenId4VciCredentialSupported,
   OpenId4VciIssuerMetadata,
 } from '../shared'
-import type { AgentContext, JwaSignatureAlgorithm, Key, JwkJson } from '@credo-ts/core'
-import type {
-  AccessTokenResponse,
-  CredentialResponse,
-  Jwt,
-  OpenIDResponse,
-  AuthorizationDetails,
-  AuthorizationDetailsJwtVcJson,
-  CredentialIssuerMetadataV1_0_11,
-  CredentialIssuerMetadataV1_0_13,
-} from '@sphereon/oid4vci-common'
 
 import {
+  AgentContext,
+  JwaSignatureAlgorithm,
+  Key,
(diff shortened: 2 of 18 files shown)


packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
```diff
@@ -263,6 +273,44 @@ export class OpenId4VciHolderService {
...
+    }
+  ) {
+    const dPoPSigningAlgValuesSupported =
+      metadata.authorizationServerMetadata?.dpop_signing_alg_values_supported ??
+      metadata.credentialIssuerMetadata.dpop_signing_alg_values_supported
+
+    if (!dPoPSigningAlgValuesSupported) return undefined
+
+    const alg = dPoPSigningAlgValuesSupported
+      .flatMap((alg) => alg)
```
TimoGlastra, line 289: Why is this needed? It's already string array right?

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
```diff
@@ -271,6 +319,11 @@ export class OpenId4VciHolderService {
     let accessTokenResponse: OpenIDResponse<AccessTokenResponse>
 
     const accessTokenClient = new AccessTokenClient()
+
+    const createDPoPOptions = await this.getCreateDPoPOptions(agentContext, metadata)
```
TimoGlastra, line 323: In credo we usually remove the capitalization if it's being used aa a word.

So createDpopOptions. Then you always know when a word starts and stops.

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
```diff
@@ -271,6 +319,11 @@ export class OpenId4VciHolderService {
...
+    const dPoPJwk = createDPoPOptions ? getJwkFromJson(createDPoPOptions.jwtIssuer.jwk) : undefined
+
+    resolvedCredentialOffer.metadata.credentialIssuerMetadata
```
TimoGlastra, line 326:
```suggestion
```

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
```diff
@@ -297,7 +352,7 @@ export class OpenId4VciHolderService {
 
     this.logger.debug('Requested OpenId4VCI Access Token.')
 
-    return accessTokenResponse.successBody
+    return { ...accessTokenResponse.successBody, dPoPJwk }
```
TimoGlastra, line 355: This is infortunate, but i don't think we should mix it. Ideally we return a new object with keys: {response: {body,headers,code}, dpop: {jwk}}


Then It's more extendable. 

Maybe deprecate and add a new one?  Or we keep it like this for now and upgrade when moving to 0.6


But still would like to have dpop.jwk so we can extend dpop values as well.

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
```diff
@@ -308,6 +363,7 @@ export class OpenId4VciHolderService {
       resolvedAuthorizationRequestWithCode?: OpenId4VciResolvedAuthorizationRequestWithCode
       accessToken?: string
       cNonce?: string
+      dPoPJwk?: Jwk
```
TimoGlastra, line 366: dpopJwk

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
```diff
@@ -347,7 +405,7 @@ export class OpenId4VciHolderService {
     } as OpenId4VciTokenRequestOptions
 
     const tokenResponse = options.accessToken
-      ? { access_token: options.accessToken, c_nonce: options.cNonce }
+      ? { access_token: options.accessToken, c_nonce: options.cNonce, dPoPJwk: options.dPoPJwk }
       : await this.requestAccessToken(agentContext, tokenRequestOptions)
```
TimoGlastra, line 409: For the service we can already change it (only api we need to worry about breaking changes)

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
```diff
@@ -412,10 +470,24 @@ export class OpenId4VciHolderService {
         .withToken(tokenResponse.access_token)
 
       const credentialRequestClient = credentialRequestBuilder.build()
+
+      let createDPoPOptions: CreateDPoPClientOptions | undefined
+      if (options.dPoPJwk) {
```
TimoGlastra, line 475: Isn't the dpop jwk also included in the access token cnf? In that case, can't we dynamically detect it and then use dpop?

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderServiceOptions.ts
```diff
@@ -112,6 +112,7 @@ export interface OpenId4VciCredentialRequestOptions extends Omit<OpenId4VciAccep
   resolvedCredentialOffer: OpenId4VciResolvedCredentialOffer
   accessToken: string
   cNonce?: string
+  dPoPJwk?: Jwk
```
TimoGlastra, line 115: dpopJwk ( you get it)

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerApi.ts
```diff
@@ -76,6 +76,7 @@ export class OpenId4VcIssuerApi {
       issuer.credentialConfigurationsSupported = undefined
     }
     issuer.display = options.display
+    issuer.dPoPSigningAlgValuesSupported = options.dPoPSigningAlgValuesSupported
```
TimoGlastra, line 79: I feel like we'll have more issuer metadata properties over time that we want to add. Should we have a generic way for it?

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerService.ts
```diff
@@ -303,6 +303,7 @@ export class OpenId4VcIssuerService {
     const openId4VcIssuerBase = {
       issuerId: options.issuerId ?? utils.uuid(),
       display: options.display,
+      dPoPSigningAlgValuesSupported: options.dPoPSigningAlgValuesSupported,
```
TimoGlastra, line 306:
```suggestion
      dpopSigningAlgValuesSupported: options.dPoPSigningAlgValuesSupported,
```

packages/openid4vc/src/openid4vc-issuer/router/accessTokenEndpoint.ts
```diff
@@ -120,6 +122,30 @@ export function handleTokenRequest(config: OpenId4VciAccessTokenEndpointConfig)
     const issuerMetadata = openId4VcIssuerService.getIssuerMetadata(agentContext, issuer)
     const accessTokenSigningKey = Key.fromFingerprint(issuer.accessTokenPublicKeyFingerprint)
 
+    let dPoPJwk: JWK | undefined
+    if (request.headers.dpop) {
+      try {
+        const fullUrl = request.protocol + '://' + request.get('host') + request.originalUrl
```
TimoGlastra, line 128: Need to use the base url from module configuration. This will break

packages/openid4vc/src/openid4vc-issuer/router/accessTokenEndpoint.ts
```diff
@@ -120,6 +122,30 @@ export function handleTokenRequest(config: OpenId4VciAccessTokenEndpointConfig)
     const issuerMetadata = openId4VcIssuerService.getIssuerMetadata(agentContext, issuer)
     const accessTokenSigningKey = Key.fromFingerprint(issuer.accessTokenPublicKeyFingerprint)
 
+    let dPoPJwk: JWK | undefined
+    if (request.headers.dpop) {
```
TimoGlastra, line 126: Can we require dpop?

packages/openid4vc/src/openid4vc-issuer/router/verifyResourceRequest.ts
```diff
@@ -27,6 +36,15 @@ export async function verifyAccessToken(
     },
   })
 
+  const fullUrl = request.protocol + '://' + request.get('host') + request.originalUrl
```
TimoGlastra, line 39: Same here . Need to use baseUrl from configuration

packages/openid4vc/src/openid4vc-issuer/router/verifyResourceRequest.ts
```diff
@@ -27,6 +36,15 @@ export async function verifyAccessToken(
...
+  await verifyResourceDPoP(
```
TimoGlastra, line 40: I know it looks a bit weird, but consistency is king imo

```suggestion
  await verifyResourceDpop(
```

packages/openid4vc/src/shared/utils.ts
```diff
@@ -124,13 +134,20 @@ export async function openIdTokenIssuerToJwtIssuer(
...
+    }
+
     if (!issuer.startsWith('https://')) {
       throw new CredoError('The X509 certificate issuer must be a HTTPS URI.')
     }
 
     if (leafCertificate.sanUriNames?.includes(issuer)) {
       return {
         ...openId4VcTokenIssuer,
+        alg: alg as unknown as SigningAlgo,
```
TimoGlastra, line 150: I think the lib should just support string. As you can use any algorithm and this is a bit annoying to cast everytime

packages/openid4vc/tests/openid4vc.e2e.test.ts
```diff
@@ -218,6 +223,7 @@ describe('OpenId4Vc', () => {
       },
     })
     const openIdIssuerTenant2 = await issuerTenant2.modules.openId4VcIssuer.createIssuer({
+      dPoPSigningAlgValuesSupported: ['EdDSA'],
```
TimoGlastra, line 226: Thinking in general whether this is something you want to configure per issuer? It will make uodates in multi-tenancy quite cumbersome. But i also see the case for having different values per issuer

TimoGlastra (commented):
Nice very good stuff martin!!'

Only thinkg i'm not sure about is the retuen value of requestAccesToken. Bit unfortunate we already released the previous pr as now it will be a breaking change...

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
auer-martin: no

packages/openid4vc/src/openid4vc-issuer/router/accessTokenEndpoint.ts
auer-martin: Well, we could implement it. But in theory, the client can opt into it for more security.
Also, supporting dpop without having clients use it is perfectly viable.
If we implement it we should probably create a config field allowing users to opt into that strict behaviour

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
auer-martin: I would leave it as is for now. Maybe change with 0.6 then.

> But still would like to have dpop.jwk so we can extend dpop values as well. 
On the client side you only receive the dpop not really things that can be extended IMO

TimoGlastra: Until we maybe want to return the dpop that was sent, or the htu, or anything.. I think we've had a few times now where not adding an extra layer of nesting resulted in using having to make breaking changes. Rather take that into account now

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
TimoGlastra: Interesting. For sd-jwt they include the jwk directly but here a fingerpring...

packages/openid4vc/src/openid4vc-issuer/router/accessTokenEndpoint.ts
TimoGlastra: For the funke pid issuer it is a requirement. There's going to be cases where the issuer wants to require it (fine to keep it out for now, but something we can add later. Maybe on the issuer record, or per credential offer)

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
```diff
@@ -266,14 +275,60 @@ export class OpenId4VciHolderService {
...
 
     const accessTokenClient = new AccessTokenClient()
+
+    const createDPoPOpts = await this.getCreateDPoPOptions(agentContext, metadata)
+
+    let dpopJwk: Jwk | undefined
+    if (createDPoPOpts) {
+      if (!createDPoPOpts.jwtIssuer.jwk.kty) {
+        throw new CredoError('Missing required key type (kty) in the jwk.')
+      }
```
TimoGlastra, line 329: why is this needed? In theory the jwk should have this property, so I'm not sure why we check for this specific property

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
```diff
@@ -266,14 +275,60 @@ export class OpenId4VciHolderService {
...
+      dpopJwk = getJwkFromJson(createDPoPOpts.jwtIssuer.jwk as JwkJson)
```
TimoGlastra, line 330: Why first transform from JWK class to json and then back to jwk, shouldn't the createDpopOptions just return the instance?

packages/openid4vc/src/openid4vc-holder/OpenId4VcHolderApi.ts
```diff
@@ -146,11 +146,12 @@ export class OpenId4VcHolderApi {
...
-    const { access_token: accessToken, c_nonce: cNonce } = await this.openId4VciHolderService.requestAccessToken(
-      this.agentContext,
-      options
-    )
-    return { accessToken, cNonce }
+    const {
+      access_token: accessToken,
+      c_nonce: cNonce,
+      dpop,
+    } = await this.openId4VciHolderService.requestAccessToken(this.agentContext, options)
```
TimoGlastra, line 153: We can change the service interface just fine, so let's change the service return value to:
```
{
  accessToken,
  cNonce,
  dpop,
  response // should be cloned response (using .clone)
}
```

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
```diff
@@ -311,6 +371,7 @@ export class OpenId4VciHolderService {
       resolvedAuthorizationRequestWithCode?: OpenId4VciResolvedAuthorizationRequestWithCode
       accessToken?: string
       cNonce?: string
+      dpop?: { dpopJwk: Jwk; dpopNonce?: string }
```
TimoGlastra, line 374: just thinking: is it needed to prefix dpop if we already are in `dpop` object?

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
```diff
@@ -351,7 +414,11 @@ export class OpenId4VciHolderService {
     } as OpenId4VciTokenRequestOptions
 
     const tokenResponse = options.accessToken
-      ? { access_token: options.accessToken, c_nonce: options.cNonce }
+      ? {
+          access_token: options.accessToken,
+          c_nonce: options.cNonce,
+          ...(options.dpop && { dpop: { dpopJwk: options.dpop.dpopJwk, dpopNonce: options.dpop?.dpopNonce } }),
```
TimoGlastra, line 420: this won't work?

```suggestion
          ...(options.dpop && { dpop: options.dpop  }),
```

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
```diff
@@ -425,10 +492,24 @@ export class OpenId4VciHolderService {
...
+      if (tokenResponse.dpop) {
+        const jwk = tokenResponse.dpop.dpopJwk
+        const alg = jwk.supportedSignatureAlgorithms[0]
+
+        createDpopOpts = {
+          jwtIssuer: { alg: alg as unknown as SigningAlgo, jwk: jwk.toJson() },
+          jwtPayloadProps: { accessToken: tokenResponse.access_token, nonce: tokenResponse.dpop?.dpopNonce },
+          createJwtCallback: getCreateJwtCallback(agentContext),
+        }
+      }
```
TimoGlastra, line 506: Why not reuse `this.getCreateDPoPOptions` here? Also this could give an error when multiple alg values are supported for a jwk type, but the server only supports a specific one. So reusing the dpop alg values supported logic from existing method makes sense?

Also, accessing index [0] is unsafe again

TimoGlastra: Also should we check if dpop is required? Can we see that using the access token? Or how do we know it is required?

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerApi.ts
```diff
@@ -63,21 +63,22 @@ export class OpenId4VcIssuerApi {
...
+      issuer.credentialsSupported = credentialsSupportedV13ToV11(credentialConfigurationsSupported)
     } else {
-      issuer.credentialsSupported = options.credentialsSupported
+      issuer.credentialsSupported = credentialsSupported
       issuer.credentialConfigurationsSupported = undefined
     }
-    issuer.display = options.display
 
-    return this.openId4VcIssuerService.updateIssuer(this.agentContext, issuer)
+    return this.openId4VcIssuerService.updateIssuer(this.agentContext, Object.assign(issuer, issuerOptions))
```
TimoGlastra, line 81: I'd rather manually assign the props that can be updated instead of doing object.assign. Especially since if you're not using TypeScript it will just assign any property to the record that you pass here. You could provide e.g. `type: 'Random'` and it will overwrite the type of the record.

packages/openid4vc/src/shared/utils.ts
```diff
@@ -66,23 +67,29 @@ export function getVerifyJwtCallback(agentContext: AgentContext): VerifyJwtCallb
...
         payload: JwtPayload.fromJson(jwt.payload),
         key,
       })
 
       return jws
     } else if (jwtIssuer.method === 'jwk') {
-      const key = getJwkFromJson(jwtIssuer.jwk).key
+      if (!jwtIssuer.jwk.kty) {
+        throw new CredoError('Missing required key type (kty) in the jwk.')
+      }
```
TimoGlastra, line 88: again, why this check?

TimoGlastra (approved):

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
auer-martin: Hmmm. I don't think so because we create the options which are passed to sphereons library and there we don't have the jwk class

packages/openid4vc/src/openid4vc-holder/OpenId4VcHolderApi.ts
auer-martin: ? what should the response contain? the whole response?

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
auer-martin: if the tokenresponse contains dpop it is required

packages/openid4vc/src/shared/utils.ts
auer-martin: can be undefined in sphereon

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
TimoGlastra: But token response is passed by the user

packages/openid4vc/src/shared/utils.ts
TimoGlastra: But in this case we have a jwk class instance right? So it won't happen?

packages/openid4vc/src/openid4vc-holder/OpenId4VcHolderApi.ts
TimoGlastra: Yes it would be a start at solving https://github.com/openwallet-foundation/credo-ts/issues/1965, which was requested. Returning the response means better extensibility as you can extract the headers, etc..

auer-martin: In the PR where I implemented the tokenRequestfunctionality, I had that behaviour, and you made me change it.

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
auer-martin: Hmm, IMO, the user should not modify the object unless he knows what he is doing.
Also in theory it is valid to remove it since an authorization server can protect multiple credential issuers,
for some of them use of dpop may be optional even if the authorization server requires dpop.

packages/openid4vc/src/openid4vc-holder/OpenId4VcHolderApi.ts
TimoGlastra: Okay then i misunderstood the previous implementation. What i don't want is requiring the whole access token response as input to the request credentials method.

I want to have separation between the required fields for credo to work, and additional params you may need for extension. And as I said in the previous PR, i'm not a fan of mixing the successBody with extra params we extracted from e.g. the header.

TimoGlastra:
Made a small mistake in my PR on jwk thumbprint, so we need to wait for https://github.com/Sphereon-Opensource/OID4VC/pull/140

PR #1974, opened 2024-07-31
fix: remove unused openid token issuer requirement

auer-martin:


packages/openid4vc/src/openid4vc-holder/OpenId4vcSiopHolderService.ts
```diff
@@ -115,19 +117,26 @@ export class OpenId4VcSiopHolderService {
...
+          'Unable to create authorization response. openIdTokenIssuer MUST be supplied when no presentation is active and the ResponseType includes id_token.'
+        )
+      }
+
+      this.assertValidTokenIssuer(authorizationRequest, openIdTokenIssuer)
     }
 
-    this.assertValidTokenIssuer(authorizationRequest, openIdTokenIssuer)
-    const openidProvider = await this.getOpenIdProvider(agentContext)
+    const jwtIssuer =
```
TimoGlastra, line 130: Does this now also support jwk as id token issuer?

auer-martin: no

TimoGlastra (approved):

PR #1983, opened 2024-08-01
fix(openid4vc): v11 metadata type and transformation

TimoGlastra:
There were some issues in the transformation logic between v11 and v13 metadata. I fixedthe issue and also updated the code to use v13 by default as the v13 format is richer and thus we make sure we won't lose any metadata when transforming from v11 -> 13

### DIFF
--- a/packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
+++ b/packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
@@ -12,7 +12,6 @@ import type {
   OpenId4VciTokenRequestOptions,
 } from './OpenId4VciHolderServiceOptions'
 import type {
-  OpenId4VciCredentialConfigurationsSupported,
   OpenId4VciCredentialConfigurationSupported,
   OpenId4VciCredentialSupported,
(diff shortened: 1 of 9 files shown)


packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
```diff
@@ -357,24 +351,18 @@ export class OpenId4VciHolderService {
...
-          }
-
-          return true
-        }) ?? offeredCredentials
+    const credentialConfigurationToRequest =
+      credentialsToRequest?.map((id, i) => {
+        if (!offeredCredentialConfigurations[id]) {
+          const offeredCredentialIds = Object.keys(offeredCredentialConfigurations).join(', ')
+          throw new CredoError(
+            `Credential to request '${credentialsToRequest[i]}' is not present in offered credentials. Offered credentials are ${offeredCredentialIds}`
```
berendsliedrecht, line 359:
```suggestion
            `Credential to request '${id}' is not present in offered credentials. Offered credentials are ${offeredCredentialIds}`
```

berendsliedrecht (approved):

PR #2006, opened 2024-08-21
fix: iso compliant auth requests

auer-martin:


packages/openid4vc/src/openid4vc-verifier/OpenId4VcSiopVerifierService.ts
```diff
@@ -97,9 +100,31 @@ export class OpenId4VcSiopVerifierService {
...
+        clientIdScheme = 'x509_san_uri'
+      } else {
+        throw new CredoError(
+          `With jwtIssuer 'method' 'x5c' the jwtIssuer's 'issuer' field must either match the match a sanDnsName (FQDN) or sanUriName in the leaf x509 chain's leaf certificate.`
+        )
+      }
+
+      if (clientIdScheme === 'x509_san_dns') {
+        clientId = getDomainFromUrl(jwtIssuer.issuer)
+        authorizationResponseUrl = jwtIssuer.issuer
```
TimoGlastra, line 123: Do we check if the jwt.issuer matches the auth response url? I think overwriting it can lead to weird behavior, as then the endpoint isn't reachable anymore

packages/openid4vc/src/openid4vc-verifier/OpenId4VcSiopVerifierService.ts
```diff
@@ -97,9 +100,31 @@ export class OpenId4VcSiopVerifierService {
...
+      } else {
+        clientId = jwtIssuer.issuer
+        authorizationResponseUrl = clientId
```
TimoGlastra, line 126: Same here, is this safe? We don't check if clientId === authorizationResponseUrl?

packages/openid4vc/tests/openid4vc.e2e.test.ts
```diff
@@ -771,15 +777,17 @@ describe('OpenId4Vc', () => {
         requestSigner: {
           method: 'x5c',
           x5c: [rawCertificate],
-          issuer: 'https://example.com/hakuna/matadata',
```
TimoGlastra, line 774: 😂

packages/openid4vc/src/openid4vc-verifier/OpenId4VcSiopVerifierService.ts
```diff
@@ -97,9 +100,32 @@ export class OpenId4VcSiopVerifierService {
...
 
+    let authorizationResponseUrl = joinUriParts(this.config.baseUrl, [
+      options.verifier.verifierId,
+      this.config.authorizationEndpoint.endpointPath,
+    ])
+
     if (jwtIssuer.method === 'x5c') {
-      clientId = jwtIssuer.issuer
-      clientIdScheme = jwtIssuer.clientIdScheme
+      if (!jwtIssuer.issuer.startsWith(authorizationResponseUrl)) {
```
TimoGlastra, line 109: Is starts with enough? Don't they have to exactly match?

PR #2007, opened 2024-08-21
feat(cheqd): Add CheqdApi

DaevMithran:
This PR introduces the exposure of the `createResource` function from the `CheqdDidRegistrar` class using `CheqdApi` class, enabling the creation of custom DID-Linked Resources (DLRs). With this enhancement, users can now create and manage DLRs such as trust registries, images, files, and other custom resources as needed.

### DIFF
--- a/packages/cheqd/src/CheqdApi.ts
+++ b/packages/cheqd/src/CheqdApi.ts
@@ -1,0 +1,18 @@
+import { AgentContext } from '@credo-ts/core'
+import { injectable } from 'tsyringe'
+
+import { CheqdCreateResourceOptions, CheqdDidRegistrar } from './dids'
+
+@injectable()
+export class CheqdApi {
+  private agentContext: AgentContext
+
+  public constructor(agentContext: AgentContext) {
+    this.agentContext = agentContext
+  }
+
+  public async createResource(did: string, options: CheqdCreateResourceOptions) {
+    const cheqdDidRegistrar = this.agentContext.dependencyManager.resolve(CheqdDidRegistrar)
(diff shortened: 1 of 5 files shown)


packages/cheqd/src/CheqdApi.ts
```diff
@@ -0,0 +1,18 @@
...
+
+@injectable()
+export class CheqdApi {
+  private agentContext: AgentContext
+
+  public constructor(agentContext: AgentContext) {
+    this.agentContext = agentContext
+  }
+
+  public async createResource(did: string, options: CheqdCreateResourceOptions) {
```
TimoGlastra, line 14: In the underlying method, all fields except data are optional. Is this correct? Do we need to provide some simplified API top level? OR is it correct that all fields are optiona? (so just providing data should succeed)

DaevMithran: No @TimoGlastra, ideally the agent should provide the below attributes

- name
- resourceType
- data
-  id

They are defaulted to an empty string and will be rejected from the ledger side. Updated `CheqdCreateResourceOptions` with a stricter schema

TimoGlastra (approved):
Thinking of it now, if we have createResource, shouldn't we also add a getResource method?

TimoGlastra:
I'll merge now, but would  be great if you could also add the getResource method :)

TimoGlastra:
@DaevMithran can you accept my PR to your branch to add a changelog entry? https://github.com/cheqd/credo-ts/pull/6

TimoGlastra:
We can merge after taht

PR #2009, opened 2024-08-23
Question about oid4vc demo

Patrik-Stas:
Hi,
this is not PR I am necessary looking to merge, but rather ask a question about the code and changes illustrated here.

My issue is illustrated by scenario:
- I have a user which somehow already authenticated themselves and I want to issue credential with their `name` 
- The user calls some endpoint to get the credential offer, which end up calling into `function createCredentialOffer` in the demo - i have generated random `userId`, as illustration of the situation where I know this information
- Credential offer is created here, but later on, when `getCredentialRequestToCredentialMapper` is called, how am I supposed to know who is actually calling that endpoint? I might need that information to actually look up the user personal information to fill out the credential payload.
- Well, I got it working if you look at the code, the simple way I could find, but I am in doubt this is the recommended way of using credo.

### DIFF
--- a/demo-openid/src/Issuer.ts
+++ b/demo-openid/src/Issuer.ts
@@ -48,12 +48,47 @@ export const credentialsSupported = [
   universityDegreeCredentialSdJwt,
 ] satisfies OpenId4VciCredentialSupportedWithId[]
 
+// string: string dictionary
+const issuanceToUserId = {} as Record<string, string>
+
+function getStringAfterLastSlash(credentialOfferUriEncoded: string): string {
+  // Decode the URL-encoded credential_offer_uri
+  const credentialOfferUri = decodeURIComponent(credentialOfferUriEncoded)
+
+  // Extract the part after the last '/' in the decoded credential_offer_uri
+  const parts = credentialOfferUri.split('/')
+  return parts[parts.length - 1]
+}
+
+function getNameSurname(userId: string) {
+  return {
+    name: `Alice${userId}`,
+    surname: `Doe${userId}`,
+  }
+}
+
+function credentialOfferOfferUri(credentialOffer: string) {
+  const url = new URL(credentialOffer)
+  const credentialOfferUri = url.searchParams.get('credential_offer_uri')
+  if (!credentialOfferUri) {
+    throw new Error('credential_offer_uri parameter not found in the input string.')
+  }
+  return credentialOfferUri
+}
+
 function getCredentialRequestToCredentialMapper({
   issuerDidKey,
 }: {
   issuerDidKey: DidKey
 }): OpenId4VciCredentialRequestToCredentialMapper {
-  return async ({ holderBinding, credentialConfigurationIds }) => {
+  return async ({ issuanceSession, holderBinding, credentialConfigurationIds }) => {
+    const issuanceId = getStringAfterLastSlash(issuanceSession.credentialOfferUri)
+    const userId = issuanceToUserId[issuanceId]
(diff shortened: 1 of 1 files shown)


TimoGlastra (commented):
There's a few approaches. 

The first is kinda like what you're doing now and that is storing the sessionId of the offer outside of credo (think in a normal database). When you issue the credential you fetch this. This way no correlatable data is stored in Credo about the issuance at all.

Another approach is that you can provide issuance metadata (issuanceMetadata) when creating an offer which is a generic object. You can read this value at time of issuance and get the needed data. Either an userId to fetch the data, or the data itself.

Hope this makes sense. We kept it a bit more flexible on how you associate a session, partly so you don't have to store any PII in Credo, but also to already accommodate for the authorization code flow (where the credential data is only known later in the flow once the user is authorized)

Patrik-Stas:
I see, thanks @TimoGlastra 

I just realized I could have done it simpler in the example, as `createCredentialOffer` returns not only the offer itself, but also `issuanceSession: OpenId4VcIssuanceSessionRecord`, which contains `id`.

I been playing with this demo and it's really helpful to get oid4vc stuff up and running with credo, but I would like to suggest some improvements, based on my own journey. Let me know which of these make sense to you.

### Demo

- Adjusting this PR to demonstrate offer-request correlation via `issuanceSession.id` and issuing claim values pre-defined upon creation of credential offer (hence bit more reflecting circumstances of typical pre-authorization flow)
- For `jwt` credentials, include some claims in credentials (setting up claims is only demonstrated for `sd-jwt`)
- Demonstrate displaying credential values on holder side
- Demonstrate display values shared to verifier in presentation 

### Logging

Additionally I found myself doing lot of console.logging to see the exchanges between the parties to understand:
- which endpoints were called
- with what payload
- what response was returned
It would be cool to have option to enable logging of all incoming and outgoing HTTP requests for all these 3 agents. I guess that would be matter of tweaking implementations for Inbounds/Outbounds abstractions? Would that be good way to go about it?

On that note, it would be IMO also nice if credo also printed list of endpoints it's exposing. I have to pass credo my express app to register bunch of endpoints, but it takes digging through code to find out what they are, what the URL parameters represents etc.

### ? Decoupling express server ?
Now this is broader point, but I am wondering if there's any plans in terms of how InboundTransport are approached. Currently the http implementation is hardcoded for `express` and I am not sure how doable it is to implement my own inbound transport. Looking at the implementation, some kind of inversion of control seem to be potentially applicable here. But this is was quite vague, I am just kinda prompting your thoughts here :-)

TimoGlastra:
Thank for the suggestions. I think your suggestions for the demo make sense (and happy to accept a PR for it if you have the time )


Also on the logging, fully agree, it would be good if we can add some extra logging to give more visibility. Could you maybe open a separate issue for the HTTP logging (so request logging + exposed endpoint logging)?

On the decoupling of the express server i agree this would be nice. However it's not my / our highest priority at the moment. There's additional complexity involved with this. I do think we should extract some logic from the http server as it's indeed very dependant on internal framework logic. Maybe we can create a base HTTP implementation and provide a default express one. if you want a custom you can extend / import the base one and add your specific http server logic. 

Again thanks for the feedback, it all makes sense, and I'm happy to review PRs for it. These are just some areas that we havne't been able to spend much time on recently (very busy with extending OID4VC / EU standard support).

Patrik-Stas:
Cool, thanks for feedback! I'll try to address some of the points with PRs.

PR #2019, opened 2024-08-30
feat: optional x5c issuer field

auer-martin:
…equests


TimoGlastra (commented):
i think besides the mdoc change, this should be able to go into main right?

TimoGlastra (approved):

PR #2023, opened 2024-09-02
fix: duplicate offer being created on issuers side while negotiating offer as a holder

### LINKED ISSUE #2022: Duplicate credential record as Issuer when OOB offer and negotiateOffer as Holder
# Issue
- Two credential record can be seen of you create an OOB offer as an Issuer and negotiate it as a holder after accepting the Invitation

## Reproduce
- Issue credential OOB as a Issuer
- Accept OOB invitation as holder
- Negotiate offer as a holder
- Two records can be found when Issuer processes the proposal

GHkrishna:
## What
- Fixes #2022 

## How
- Initially during processing proposal in cases where an OOB credential is being offered.
- If holder, accepts OOB invitation and negotiates the offer, the connection details are appended on the holders side.
- But for issuer we are unable to update the credential record with the connection details.
- Hence, the issuer considers this credential proposal as a new flow and creates a new credential record with connection details. This causes the flow to break further in the processing when two records against a single threadId are found

### DIFF
--- a/packages/core/src/modules/credentials/protocol/v2/V2CredentialProtocol.ts
+++ b/packages/core/src/modules/credentials/protocol/v2/V2CredentialProtocol.ts
@@ -186,15 +186,6 @@ export class V2CredentialProtocol<CFs extends CredentialFormatService[] = Creden
 
     // credential record already exists
     if (credentialRecord) {
-      //  This makes sure to associate the connectionId
-      if (!credentialRecord?.connectionId) {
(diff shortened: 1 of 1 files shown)


TimoGlastra:
Previously we haven't supported proposals for connectionless exchanges. We probably need some additional checks to ensure the exchange can't be hijacked by another connection. You can look at the other methods. And we need to set the connection id after fetching if it's available

GHkrishna:
@TimoGlastra is this the right approach to associate the connectionId once we get the credentialRecord

TimoGlastra (approved):
Thank you!

PR #2024, opened 2024-09-02
feat: support w3c revocation

KulkarniShashank:
### What ###
This PR introduces support for W3C credential revocation. The following changes have been made:

Refactor functionality for Verify Credential and Verify Presentation.
Integrated with the Bit String Status List for checking and updating credential status.

### How ###
**Revocation Flow**: Implemented the revocation flow by checking a credential status and updating the credential issuance process to handle revocable credentials.
**Integration**: Integrated the new revocation features with the existing system, ensuring compatibility and proper handling of revoked credentials.


packages/core/src/modules/vc/data-integrity/W3cJsonLdCredentialService.ts
```diff
@@ -265,6 +304,52 @@ export class W3cJsonLdCredentialService {
         challenge: options.challenge,
         domain: options.domain,
         documentLoader: this.w3cCredentialsModuleConfig.documentLoader(agentContext),
+        checkStatus: async ({ credential }: { credential: W3cJsonCredential }) => {
```
TimoGlastra, line 307: Can you extract this into a method and reuse that for both checkStatus methods?

packages/core/src/modules/vc/data-integrity/W3cJsonLdCredentialService.ts
```diff
@@ -265,6 +304,52 @@ export class W3cJsonLdCredentialService {
...
+            const bitStringCredential =
+              (await bitStringStatusListCredential.json()) as unknown as BitStringStatusListCredential
+            const encodedBitString = bitStringCredential.credential.credentialSubject.encodedList
+            const compressedBuffer = Uint8Array.from(atob(encodedBitString), (c) => c.charCodeAt(0))
+
+            // Decompress using pako
+            const decodedBitString = pako.ungzip(compressedBuffer, { to: 'string' })
+            const statusListIndex = Number(credential.credentialStatus.statusListIndex)
+
+            if (statusListIndex < 0 || statusListIndex >= decodedBitString.length) {
```
TimoGlastra, line 341: We should validate the bitstring status list credential as well

packages/core/src/modules/vc/models/credential/W3cJsonCredential.ts
```diff
@@ -10,4 +10,59 @@ export interface W3cJsonCredential {
   expirationDate?: string
   credentialSubject: SingleOrArray<JsonObject>
   [key: string]: unknown
+  credentialStatus?: SingleOrArray<CredentialStatus>
+}
+
+type CredentialStatusType = 'BitstringStatusListEntry'
```
TimoGlastra, line 16: Can we move all the status list logic related to bit string status list to a separate file?

packages/core/src/modules/vc/data-integrity/W3cJsonLdCredentialService.ts
```diff
@@ -265,6 +304,52 @@ export class W3cJsonLdCredentialService {
...
+                'Verifying credential status as an array for JSON-LD credentials is currently not supported'
+              )
+            }
+
+            // Ensure credentialStatus contains the necessary properties
+            if (!credential.credentialStatus || credential.credentialStatus.statusListIndex === undefined) {
+              throw new CredoError('Invalid credential status format')
+            }
+
+            const credentialStatusURL = credential.credentialStatus.statusListCredential
```
TimoGlastra, line 321: We should first check the type fiets to make sure this is a bit string status type. There could be others and those are not supported at the moment.

TimoGlastra (changes requested):
Thanks for this! Left some comments

KulkarniShashank:
@TimoGlastra As per your suggestions, I have made the changes. Please review them and let me know your thoughts.

packages/core/src/modules/vc/data-integrity/W3cJsonLdCredentialService.ts
KulkarniShashank: @TimoGlastra I have developed a common function for this purpose. However, in the **checkStatus** method, there are specific conditions related to credential status that are dependent on the payload received from the **verifyCredential** and **verifyPresentation** functions. Therefore, the **checkStatus** method invokes the common function, as it contains the identical logic.

Tweeddalex:
Out of interest @KulkarniShashank will this work with cheqd DIDs and on-ledger Bitstring Status Lists?

KulkarniShashank:
> Out of interest @KulkarniShashank will this work with cheqd DIDs and on-ledger Bitstring Status Lists?

Its can work for all the DIDs that support JSON-LD credential format

TimoGlastra:
However @Tweeddalex this PR does not support cheqd yet. The current implementation only suppors https and uses `fetch` to retrieve the status list. For cheqd support we'd have to fetch the status list differently.

packages/core/src/modules/vc/data-integrity/W3cJsonLdCredentialService.ts
```diff
@@ -265,6 +265,14 @@ export class W3cJsonLdCredentialService {
         challenge: options.challenge,
         domain: options.domain,
         documentLoader: this.w3cCredentialsModuleConfig.documentLoader(agentContext),
+        checkStatus: async ({ credential }: { credential: W3cJsonCredential }) => {
+          if ('credentialStatus' in credential) {
+            await verifyBitStringCredentialStatus(credential, agentContext)
+          }
```
TimoGlastra, line 271: verifyPresentation also has options.verifyCredentialStatus, can you add that check here as well?

packages/core/src/modules/vc/models/credential/W3cCredentialStatus.ts
```diff
@@ -1,23 +1,61 @@
...
+  @IsString()
+  public type!: CredentialStatusType
+
+  @IsEnum(CredentialStatusPurpose)
+  @IsString()
+  public statusPurpose!: CredentialStatusPurpose
+
   @IsString()
-  public type!: string
+  public statusListIndex!: string
```
TimoGlastra, line 42: Looking at the VC Data model specification, these properties are not required fro the `credentialStatus` property. Only `id` and `type` are. 

We should have a generic base class and allow multiple extension points. You can look at e.g. how we solved this with the DidDocument service property, where there's a base service class and multiple possible extensions that are matched based on the `type` property

packages/core/src/modules/vc/models/credential/W3cCredentialStatus.ts
```diff
@@ -1,23 +1,61 @@
...
+  try {
+    await validateOrReject(entry)
+    return true
+  } catch (errors) {
+    agentContext.config.logger.debug(`Credential status validation failed: ${errors}`, {
+      stack: errors,
+    })
+    throw new CredoError(`Invalid credential status type: ${errors}`)
+  }
 }
```
TimoGlastra, line 61: This validation is automatically done when transforming into a specific class instance type, so I don't think it's needed if you follow the did document service approach

packages/core/src/modules/vc/models/credential/W3cJsonCredentialStatus.ts
```diff
@@ -0,0 +1,55 @@
...
+
+// Define an interface for the `credential` object that uses `CredentialSubject`
+export interface Credential {
+  credentialSubject: CredentialSubject
+}
+
+// Use the `Credential` interface within `BitStringStatusListCredential`
+export interface BitStringStatusListCredential {
+  credential: Credential
+}
```
TimoGlastra, line 55: All these interface names should be bitstring status list specific. It is important to make a distinction between bitstring status list, and possible other `credentialStatus` methods that will be added in the future.

packages/core/src/modules/vc/data-integrity/VerifyBitStringCredentialStatus.ts
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  try {
+    return (await response.json()) as BitStringStatusListCredential
+  } catch (error) {
+    throw new CredoError('Failed to parse the bit string status list credential')
+  }
+}
+
+export const verifyBitStringCredentialStatus = async (credential: W3cJsonCredential, agentContext: AgentContext) => {
```
TimoGlastra, line 28: This doesn't validate the signature of the status list credential yet

packages/core/src/modules/credentials/CredentialsApi.ts
```diff
@@ -515,6 +525,87 @@ export class CredentialsApi<CPs extends CredentialProtocol[]> implements Credent
...
 
+  /**
+   * Revoke a credential by issuer
+   * associated with the credential record.
+   *
+   * @param credentialRecordId The id of the credential record for which to revoke the credential
+   * @returns Revoke credential notification message
+   *
+   */
+  public async revokeJsonLdCredential(options: RevokeCredentialOption): Promise<{ message: string }> {
```
TimoGlastra, line 536: there is a lot of dupcliation in this file with the already present methods. Some suggestions:
- move all the credential status logic into a specific folder in the w3c module
- add the revocation related methods to the w3c credentails api, not the DIDcomm credentials api
- the only thing we should add it here is a `sendRevocationNotification` method (which we can extend to support non-anoncreds revocadion notification messages as well)
- making a POST to the credential status list url to update the status list seems implementation specific and may not work for all implementations

genaris:
Superseded by #2072

PR #2025, opened 2024-09-04
feat:Cached tentant record for context

sagarkhole4:
The getAgentContextForContextCorrelationId function caches the tenant record to avoid the expensive operation of retrieving it from the database each time.

### DIFF
--- a/packages/tenants/src/context/TenantAgentContextProvider.ts
+++ b/packages/tenants/src/context/TenantAgentContextProvider.ts
@@ -22,6 +22,7 @@ import {
   isValidJweStructure,
   JsonEncoder,
   isJsonObject,
+  CacheModuleConfig,
 } from '@credo-ts/core'
 
 import { TenantAgent } from '../TenantAgent'
@@ -55,13 +56,26 @@ export class TenantAgentContextProvider implements AgentContextProvider {
   }
 
   public async getAgentContextForContextCorrelationId(contextCorrelationId: string) {
+    this.logger.debug('Inside getAgentContextForContextCorrelationId')
     // It could be that the root agent context is requested, in that case we return the root agent context
     if (contextCorrelationId === this.rootAgentContext.contextCorrelationId) {
       return this.rootAgentContext
     }
 
-    // TODO: maybe we can look at not having to retrieve the tenant record if there's already a context available.
-    const tenantRecord = await this.tenantRecordService.getTenantById(this.rootAgentContext, contextCorrelationId)
+    const cache = this.rootAgentContext.dependencyManager.resolve(CacheModuleConfig).cache
+
+    this.logger.debug('Getting tenantRecord from cache')
+    let tenantRecord: TenantRecord | null = await cache.get(
+      this.rootAgentContext,
+      `contextCorrelationId-${contextCorrelationId}`
+    )
+    if (!tenantRecord) {
+      // TODO: maybe we can look at not having to retrieve the tenant record if there's already a context available.
+      this.logger.debug('TenantRecord not found in cache')
+      tenantRecord = await this.tenantRecordService.getTenantById(this.rootAgentContext, contextCorrelationId)
+      await cache.set(this.rootAgentContext, `contextCorrelationId-${contextCorrelationId}`, tenantRecord)
+      this.logger.debug(`Cached tenantRecord '${contextCorrelationId}'`)
+    }
     const shouldUpdate = !isStorageUpToDate(tenantRecord.storageVersion)
 
     // If the tenant storage is not up to date, and autoUpdate is disabled we throw an error


packages/tenants/src/context/TenantAgentContextProvider.ts
```diff
@@ -55,13 +56,26 @@ export class TenantAgentContextProvider implements AgentContextProvider {
...
+    this.logger.debug('Getting tenantRecord from cache')
+    let tenantRecord: TenantRecord | null = await cache.get(
+      this.rootAgentContext,
+      `contextCorrelationId-${contextCorrelationId}`
+    )
+    if (!tenantRecord) {
+      // TODO: maybe we can look at not having to retrieve the tenant record if there's already a context available.
+      this.logger.debug('TenantRecord not found in cache')
+      tenantRecord = await this.tenantRecordService.getTenantById(this.rootAgentContext, contextCorrelationId)
+      await cache.set(this.rootAgentContext, `contextCorrelationId-${contextCorrelationId}`, tenantRecord)
```
TimoGlastra, line 76: I think we should only put the tenant record in the cache after we have run the checks for agent storage updates etc.. (so after getContextForSession, and before return the agent context)

packages/tenants/src/context/TenantAgentContextProvider.ts
```diff
@@ -55,13 +56,26 @@ export class TenantAgentContextProvider implements AgentContextProvider {
...
-    const tenantRecord = await this.tenantRecordService.getTenantById(this.rootAgentContext, contextCorrelationId)
+    const cache = this.rootAgentContext.dependencyManager.resolve(CacheModuleConfig).cache
+
+    this.logger.debug('Getting tenantRecord from cache')
+    let tenantRecord: TenantRecord | null = await cache.get(
+      this.rootAgentContext,
+      `contextCorrelationId-${contextCorrelationId}`
+    )
+    if (!tenantRecord) {
+      // TODO: maybe we can look at not having to retrieve the tenant record if there's already a context available.
```
TimoGlastra, line 73: this can be removed

packages/tenants/src/context/TenantAgentContextProvider.ts
```diff
@@ -55,13 +56,26 @@ export class TenantAgentContextProvider implements AgentContextProvider {
...
+      this.logger.debug('TenantRecord not found in cache')
```
TimoGlastra, line 74: It would be helpful to add the ids etc.. to the log messages. So `TenantRecord with id ${contextCorrelationId} not found in cache` (same for the other log messages)

packages/tenants/src/context/__tests__/TenantAgentContextProvider.test.ts
```diff
@@ -62,6 +63,11 @@ describe('TenantAgentContextProvider', () => {
 
       const tenantAgentContext = jest.fn() as unknown as AgentContext
 
+      container.registerInstance(CacheModuleConfig, {
+        cache: new CacheModule({
+          cache: new InMemoryLruCache({ limit: 100 }),
+        }).config.cache as unknown as CacheModule,
+      })
```
TimoGlastra, line 70: I think you can simplify this

Does this work?

```suggestion
      container.registerInstance(CacheModuleConfig, new CacheModuleConfig({
        cache: new InMemoryLruCache({ limit: 100 })
       })
```

sagarkhole4: Yes, This test case is working I have tested it on my local machine.

packages/tenants/src/context/TenantAgentContextProvider.ts
sagarkhole4: Updated all logs

packages/tenants/src/context/TenantAgentContextProvider.ts
sagarkhole4: Removed.

packages/tenants/src/context/TenantAgentContextProvider.ts
sagarkhole4: I appreciate your suggestion. In line with that, I modified the code.

TimoGlastra (approved):
Thanks a lot!

sagarkhole4:
> pnpm install
Runed  pnpm install and committed pnpm-lock file.

TimoGlastra:
Tests are now failing as removing a tenant does not remove it from the cache. I think we should remove the tenant record from the cache if a tenant is deleted. I think for every update to the tenant record, we should update the entry in the cache.

And finally, we store the wallet key for the tenant in the wallet record, and so putting it in the cache is not safe I think.  Any thoughts on that?

TimoGlastra:
I think we should maybe only use caching if using Askar profile as multitenant scheme, as in that case the wallet key is not used. But otherwise I'm not sure it's safe to store the wallet key in the cache

sagarkhole4:
> Tests are now failing as removing a tenant does not remove it from the cache. I think we should remove the tenant record from the cache if a tenant is deleted. I think for every update to the tenant record, we should update the entry in the cache.
> 
> And finally, we store the wallet key for the tenant in the wallet record, and so putting it in the cache is not safe I think. Any thoughts on that?

I agree that we should ensure the tenant record is removed from the cache when a tenant is deleted. To address this, we can delete the cached record just before the deletion or update operation is performed. This way, the cache stays in sync with the actual database state.

Regarding the storage of sensitive data like wallet keys, you're right—storing this kind of information directly in the cache could pose a security risk. To mitigate this, we could consider encrypting the data before caching it, ensuring that any sensitive information remains secure.

As fetching data from the database can be resource-intensive, especially in a production environment, caching is a valuable mechanism to enhance performance. By keeping the cache updated efficiently, we can reduce database load and improve response times.

PR #2026, opened 2024-09-09
fix: add config to process messages concurrently

sairanjit:
- Add support to process the messages concurrently

cc: @Zzocker : Taking it further from #1276 

 @TimoGlastra From the earlier [conversations](https://github.com/openwallet-foundation/credo-ts/pull/1276#issuecomment-1423965933) how can we handle the queue from inbound transport? 
- Can we use the event emitter ?


packages/core/src/agent/AgentConfig.ts
```diff
@@ -75,6 +75,10 @@ export class AgentConfig {
     return this.initConfig.backupBeforeStorageUpdate ?? true
   }
 
+  public get processMessagesConcurrently() {
```
TimoGlastra, line 78:
```suggestion
  public get processDidCommMessagesConcurrently() {
```

packages/core/src/types.ts
```diff
@@ -83,6 +83,7 @@ export interface InitConfig {
   connectionImageUrl?: string
   autoUpdateStorageOnStartup?: boolean
   backupBeforeStorageUpdate?: boolean
+  processMessagesConcurrently?: boolean
```
TimoGlastra, line 86:
```suggestion
  processDidCommMessagesConcurrently?: boolean
```

packages/node/src/transport/WsInboundTransport.ts
```diff
@@ -58,13 +68,28 @@ export class WsInboundTransport implements InboundTransport {
...
+            session: session,
+          },
+        })
+
+        // Wait for message to be processed
+        await firstValueFrom(
+          agent.events.observable<AgentMessageProcessedEvent>(AgentEventTypes.AgentMessageProcessed).pipe(
+            filter((e) => e.type === AgentEventTypes.AgentMessageProcessed),
+            filter((e) => e.payload.message.id === encryptedMessage.id),
+            filter((e) => e.payload.message.type === encryptedMessage.type),
```
TimoGlastra, line 89: an encrypted message does not have an `id` and `type` right?

packages/node/src/transport/HttpInboundTransport.ts
```diff
@@ -52,10 +60,24 @@ export class HttpInboundTransport implements InboundTransport {
...
 
+        // Wait for message to be processed
+        await firstValueFrom(
+          agent.events.observable<AgentMessageProcessedEvent>(AgentEventTypes.AgentMessageProcessed).pipe(
+            filter((e) => e.type === AgentEventTypes.AgentMessageProcessed),
+            filter((e) => e.payload.message.id === encryptedMessage.id),
+            filter((e) => e.payload.message.type === encryptedMessage.type),
+            first()
+          )
+        )
```
TimoGlastra, line 79: If the message is processed very quickly, this won't work, we need to start the event listener before. Also if processing somehow fails this will never emit. Do we have the right guardrails in place that this event will always be emitted? Otherwise we may want to add a timeout?

TimoGlastra (commented):
Good updates! Thanks

packages/node/src/transport/WsInboundTransport.ts
sairanjit: - But here there are no common properties I can map to. Can you suggest me a way guard here?

packages/node/src/transport/HttpInboundTransport.ts
sairanjit: I have added a timeout of `10` seconds.

packages/node/src/transport/HttpInboundTransport.ts
```diff
@@ -51,11 +59,32 @@ export class HttpInboundTransport implements InboundTransport {
...
+
+        observable
+          .pipe(
+            first((e) => e.type === AgentEventTypes.AgentMessageProcessed),
+            timeout({
+              first: 10000, // timeout after 10 seconds
+              meta: 'HttpInboundTransport.start',
+            })
+          )
+          .subscribe(subject)
```
TimoGlastra, line 75: This won't work, as there could be multiple messages processed. I think we need to assign each message an unique id, or we can lookg at the encrypted message and include that in the event as well. But we need some identifier

sairanjit: - Updated the `AgentMessageProcessedEvent` to have optional parameter of the `encryptedMessage`

packages/node/src/transport/HttpInboundTransport.ts
```diff
@@ -51,11 +59,33 @@ export class HttpInboundTransport implements InboundTransport {
...
-          session,
+        const encryptedMessage = JSON.parse(message) as EncryptedMessage
+
+        const observable = agent.events.observable<AgentMessageProcessedEvent>(AgentEventTypes.AgentMessageProcessed)
+        const subject = new ReplaySubject(1)
+
+        observable
+          .pipe(
+            filter((e) => e.type === AgentEventTypes.AgentMessageProcessed),
+            filter((e) => deepEquality(e.payload.encryptedMessage, encryptedMessage)),
```
TimoGlastra, line 70: I think we can just check for equality right (===)? AS the object will be passed around? Using deep object equality can be expensive

packages/core/src/agent/Dispatcher.ts
```diff
@@ -64,7 +65,7 @@ class Dispatcher {
     await next()
   }
 
-  public async dispatch(messageContext: InboundMessageContext): Promise<void> {
+  public async dispatch(messageContext: InboundMessageContext, encryptedMessage?: EncryptedMessage): Promise<void> {
```
TimoGlastra, line 68: What if we add `encryptedMessage` to the `InboundMessageContext` ( as it's related to the incoming message)

packages/node/src/transport/WsInboundTransport.ts
```diff
@@ -58,13 +68,36 @@ export class WsInboundTransport implements InboundTransport {
...
+
+        const observable = agent.events.observable<AgentMessageProcessedEvent>(AgentEventTypes.AgentMessageProcessed)
+        const subject = new ReplaySubject(1)
+
+        observable
+          .pipe(
+            filter((e) => e.type === AgentEventTypes.AgentMessageProcessed),
+            filter((e) => deepEquality(e.payload.encryptedMessage, encryptedMessage)),
+            timeout({
+              first: 10000, // timeout after 10 seconds
```
TimoGlastra, line 85: I think it's less relevant here to await the processing? As we don't do anything afterwards

packages/node/src/transport/HttpInboundTransport.ts
```diff
@@ -51,11 +59,33 @@ export class HttpInboundTransport implements InboundTransport {
...
+            timeout({
+              first: 10000, // timeout after 10 seconds
```
TimoGlastra, line 72: Let's add this as a param to the constructor  of the HttpInboundTransport class?

TimoGlastra:
@sairanjit can you resolve the conflicts? I think we're good to merge!

sairanjit:
@TimoGlastra Resolved conflicts now

PR #2027, opened 2024-09-11
feat: Auth code flow (not working)

auer-martin:
This PR adds the base for the openid4vci auth code flow
Currently, the token response is not successful.
It also adds a Provider to the openid demo to test the authorization code flow


TimoGlastra:
Closing because of #2056

PR #2028, opened 2024-09-11
fix: handling of diddoc content as string

dbluhm:
This PR fixes a TODO item left in the Indy DID Resolution code. nym.diddocContent is in fact a string and not a json object when retrieved from the ledger.


TimoGlastra (approved):
Thanks!

genaris (approved):

PR #2029, opened 2024-09-11
feat: add Ed25519Signature2020 support

dbluhm:
This is a somewhat naive implementation of adding Ed25519Signature2020 support to the W3cCredentialsModule. Feedback is most welcome and any direction you'd be willing to offer on what other things are required to get this to a merge-able state would be appreciated.

### DIFF
--- a/packages/core/src/modules/vc/data-integrity/signature-suites/ed25519/Ed25519Signature2020.ts
+++ b/packages/core/src/modules/vc/data-integrity/signature-suites/ed25519/Ed25519Signature2020.ts
@@ -1,0 +1,236 @@
+import type { DocumentLoader, JsonLdDoc, Proof, VerificationMethod } from '../../jsonldUtil'
+import type { JwsLinkedDataSignatureOptions } from '../JwsLinkedDataSignature'
+
+import { MultiBaseEncoder, TypedArrayEncoder } from '../../../../../utils'
+import { CREDENTIALS_CONTEXT_V1_URL, SECURITY_CONTEXT_URL } from '../../../constants'
+import { _includesContext } from '../../jsonldUtil'
+import jsonld from '../../libraries/jsonld'
+import { JwsLinkedDataSignature } from '../JwsLinkedDataSignature'
+
+import { ED25519_SUITE_CONTEXT_URL_2020 } from './constants'
+import { ed25519Signature2020Context } from './context2020'
+
+type Ed25519Signature2020Options = Pick<
+  JwsLinkedDataSignatureOptions,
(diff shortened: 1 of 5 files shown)


packages/core/src/modules/vc/data-integrity/signature-suites/ed25519/Ed25519Signature2020.ts
```diff
@@ -0,0 +1,239 @@
...
+      documentLoader: options.documentLoader,
+    })
+
+    // convert Ed25519VerificationKey2020 to Ed25519VerificationKey2018
+    if (_isEd2020Key(verificationMethod) && _includesEd2020Context(verificationMethod)) {
+      // -- convert multibase to base58 --
+      let publicKeyBuffer = MultiBaseEncoder.decode(verificationMethod.publicKeyMultibase).data
+      if ((verificationMethod.publicKeyMultibase as string).startsWith('z6Mk')) {
+        publicKeyBuffer = publicKeyBuffer.slice(2)
+      }
```
TimoGlastra, line 101: We have better tools to convert from multibase to base58:

(from fingerprint is not the best name and we should change that, but it does mutlibase decoding and handles the prefix for multiple key types with differeing prefix lengths)

```
Key.fromFingerprint(verificationMethod.publicKeyMultibase).publicKeyBase58
```

TimoGlastra (changes requested):
Nice! Thanks for this. One suggestion on the key transformation

packages/core/src/modules/vc/data-integrity/signature-suites/ed25519/Ed25519Signature2020.ts
dbluhm: Thanks! What are your thoughts on whether we have to be defensive here and check for whether the multicodec is present? My memory of the specs lead me to go the defensive approach but going back to the spec (https://www.w3.org/community/reports/credentials/CG-FINAL-di-eddsa-2020-20220724/#ed25519verificationkey2020), it looks like multicodec is explicitly required now. I might just be remembering wrong or thinking of a different spec that was more vague.

For contrast, in ACA-Py, we are currently taking the defensive approach.

TimoGlastra: Hmm I know they changed it for old verification methods, which I thought was not possible (in my assumption Ed25519VerificationKey2018 would ALWAYS use publicKeyBase58). So I'm not wondering whether we should have a way more generic approach to extract the key material from a verification method?

TimoGlastra (approved):

TimoGlastra:
Thanks 🙌

PR #2037, opened 2024-09-13
fix: return error response to avoid hanging

ja-bravo:
This PR updates the access token endpoint so it returns when an error happens, closing the connection. 

This fixes a crash when the end user takes a good while to fetch the credential after creating an offer, triggering a `ERR_HTTP_HEADERS_SENT: Cannot set headers after they are sent to the client` error.

We talked about this before and it was unclear if this change would break anything else but we've been running this successfully for months so I think it's okay.


TimoGlastra (commented):
I think the issue with this approach is that the agent context session won't be closed correctly (which is why we call `next()`.

I think we need to find out what is modifying the response after the error has been sent. Or allow returning early, but still having the afterware called

TimoGlastra:
I have fixed this now in the authorization code flow PR (#2056)

ja-bravo:
Thank you for that @TimoGlastra, I haven't had a chance to take a look at this since

PR #2045, opened 2024-09-27
feat: add direct ecdh-es jwe encryption/decryption

TimoGlastra:
it's very specific to what we need for OID4VC now. At some point we should make one generic encryption/decryption api in the wallet api, and then build a JWE layer on top of it.

the method is optional for now but implemented in askar wallet so it's not really a breaking change


swcurran:
“create-spotty-hounds.md”?  I gather that is a generated name?  😆

PR #2046, opened 2024-09-27
Jarm

auer-martin:


packages/openid4vc/src/openid4vc-verifier/OpenId4VcSiopVerifierService.ts
```diff
@@ -422,15 +485,29 @@ export class OpenId4VcSiopVerifierService {
...
 
+    const mode =
+      !responseMode || responseMode === 'direct_post'
+        ? SphereonResponseMode.DIRECT_POST
+        : SphereonResponseMode.DIRECT_POST_JWT
+
+    let jarmEncryptionJwk: (JwkJson & { kid: string; use: 'enc' }) | undefined
+
+    if (mode === SphereonResponseMode.DIRECT_POST_JWT) {
+      const key = await agentContext.wallet.createKey({ keyType: KeyType.P256 })
```
TimoGlastra, line 496: I think key type should match the key type being used to sign the request (it could even be the same key?!)

packages/openid4vc/src/openid4vc-verifier/router/authorizationEndpoint.ts
```diff
@@ -15,34 +20,98 @@ export interface OpenId4VcSiopAuthorizationEndpointConfig {
...
+      if (request.body.response) {
+        const res = await RP.processJarmAuthorizationResponse(request.body.response, {
+          getAuthRequestPayload: async (input) => {
+            verificationSession = await getVerificationSession(agentContext, {
+              verifierId: verifier.verifierId,
+              state: input.state,
+              nonce: input.nonce,
+            })
+
+            const res = await AuthorizationRequest.fromUriOrJwt(verificationSession.authorizationRequestJwt)
```
TimoGlastra, line 86:
```suggestion
            const req = await AuthorizationRequest.fromUriOrJwt(verificationSession.authorizationRequestJwt)
```

packages/openid4vc/src/openid4vc-verifier/router/authorizationEndpoint.ts
```diff
@@ -15,34 +20,98 @@ export interface OpenId4VcSiopAuthorizationEndpointConfig {
...
+          },
+          decryptCompact: decryptJarmResponse(agentContext),
+        })
+
+        authorizationResponsePayload = res.authResponseParams as AuthorizationResponsePayload
+      } else {
+        authorizationResponsePayload = request.body
+      }
+
+      verificationSession = await getVerificationSession(agentContext, {
```
TimoGlastra, line 101: We already fetch ot above. Can't we reuse that?

packages/openid4vc/tests/openid4vc.e2e.test.ts
```diff
@@ -704,7 +708,7 @@ describe('OpenId4Vc', () => {
     })
   })
 
-  it('e2e flow with verifier endpoints verifying a sd-jwt-vc with selective disclosure', async () => {
+  it('e2e flow (jarm) with verifier endpoints verifying a sd-jwt-vc with selective disclosure', async () => {
```
TimoGlastra, line 711: Can we also keep a non-jarm test?

TimoGlastra (approved):
Nice!! Some small remarks

TimoGlastra:
Oh also: should we add a simple check to make sure the jarm enc alg is supported etc? Also we don't add the required metadata params i think?

PR #2049, opened 2024-10-03
chore: add raw filepath for logo in readme

cre8:
Add the raw file path for the logo

The current path will point to the file in the repository, but not the the path that will return the image to display it. Github itself has some auto correct for this, but when you are referencing the readme from another source, it will fail to load it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,8 @@
   <br />
   <img
     alt="Credo Logo"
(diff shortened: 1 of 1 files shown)


TimoGlastra (approved):

TimoGlastra:
Thanks!

cre8:
@TimoGlastra I saw there was a linebreak when I added it... The link has to be in the same line and not in a new one... sry for not previewing it first!

PR #2052, opened 2024-10-07
feat: Trusted certificates hook for presentations

Tommylans:
Gives the ability to dynamically set the trusted Certificates for a specific proof request.


packages/core/src/modules/vc/W3cCredentialServiceOptions.ts
```diff
@@ -181,6 +181,8 @@ interface W3cVerifyPresentationOptionsBase {
 
 export interface W3cJwtVerifyPresentationOptions extends W3cVerifyPresentationOptionsBase {
```
TimoGlastra, line 182: I had to think for a bit on what I think of this API. And I think it may be a bit confusing. Especially since `correlationId` is something internal to sphereon libraries, and proofRecordId it's not obvious this is for DIDcomm.

So..  what if we make it a `verificationContext`, which is optional, and has a unified interface that can be reused.

```
interface VerificationContext {
  /**
   * The `id` of the `ProofRecord` that this verification is bound to.
   */
  didcommProofRecordId?: string

  /**
   * The `id` of the `OpenId4VcVerificationSessionRecord` that this verification is bound to.
   */
   openId4VcVerificationSessionId?: string
}
```

I think we could also extract the `trustedCertificates`, but it would be good to also add these, as then you can pass them when calling the top-level APIs (w3c verification) (i think Sd-JWT is already possible right?)

And we will only call the get trusted certificates method if you didn't provide certificates (so if you call sdJWt.verify with certificates we won't call it). But otherwise we will

packages/core/src/modules/vc/jwt-vc/W3cJwtCredentialService.ts
```diff
@@ -308,13 +309,21 @@ export class W3cJwtCredentialService {
       const proverPublicKey = getKeyFromVerificationMethod(proverVerificationMethod)
       const proverPublicJwk = getJwkFromKey(proverPublicKey)
 
+      const getTrustedCertificatesForProof = agentContext.dependencyManager.isRegistered(X509ModuleConfig)
+        ? agentContext.dependencyManager.resolve(X509ModuleConfig).getTrustedCertificatesForProof
```
TimoGlastra, line 313: If we add the `verificationContextt`,  I think `getTrustedCertificatesForVerification`. makes sense here?

packages/core/src/modules/vc/jwt-vc/W3cJwtCredentialService.ts
```diff
@@ -308,13 +309,21 @@ export class W3cJwtCredentialService {
...
+        : undefined
+
       let signatureResult: VerifyJwsResult | undefined = undefined
       try {
         // Verify the JWS signature
         signatureResult = await this.jwsService.verifyJws(agentContext, {
           jws: presentation.jwt.serializedJwt,
           // We have pre-fetched the key based on the singer/holder of the presentation
           jwkResolver: () => proverPublicJwk,
+          trustedCertificates: await getTrustedCertificatesForProof?.(agentContext, {
```
TimoGlastra, line 323:
```suggestion
          trustedCertificates: options.trustedCertificates ??  await getTrustedCertificatesForProof?.(agentContext, {
```

packages/core/src/modules/x509/X509ModuleConfig.ts
```diff
@@ -1,22 +1,43 @@
...
+}
+
 export interface X509ModuleConfigOptions {
   /**
    *
    * Array of trusted base64-encoded certificate strings in the DER-format.
    */
   trustedCertificates?: [string, ...string[]]
+
+  getTrustedCertificatesForProof?(
```
TimoGlastra, line 15:
```suggestion
  /**
   * Optional callback method that will be called to dynamically get trusted certificates for a verification.
   * It will always provide the `agentContext` allowing to dynamically set the trusted certificates for a tenant.
   * If available the associated record id is also provided allowing to filter down trusted certificates to a single
   * exchange.
   */
  getTrustedCertificatesForVerification?(
```

TimoGlastra:
I like the general appraoch. Just some commetns to make it a bit more specific  / unified and so we can extend it with other records as well in the future (e.g. issuance will also need this probably)

packages/core/src/modules/vc/jwt-vc/W3cJwtCredentialService.ts
```diff
@@ -308,13 +309,21 @@ export class W3cJwtCredentialService {
...
+            proofRecordId: options.proofRecordId,
+            correlationId: options.correlationId,
+          }),
```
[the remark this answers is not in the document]
Tommylans: Ah I think I get your idea based on the other comment

packages/core/src/modules/vc/W3cCredentialServiceOptions.ts
Tommylans: We can also add the already given trustedCertificates here so it's the choice by the user to combine it with the hook ones or completely override them

packages/core/src/modules/vc/jwt-vc/W3cJwtCredentialService.ts
```diff
@@ -308,13 +309,20 @@ export class W3cJwtCredentialService {
       const proverPublicKey = getKeyFromVerificationMethod(proverVerificationMethod)
       const proverPublicJwk = getJwkFromKey(proverPublicKey)
 
+      const getTrustedCertificatesForProof = agentContext.dependencyManager.isRegistered(X509ModuleConfig)
```
TimoGlastra, line 312: forVerification

TimoGlastra:
one final remark, otherwise LGTM!

packages/core/src/modules/vc/jwt-vc/W3cJwtCredentialService.ts
Tommylans: Done

TimoGlastra (approved):

PR #2054, opened 2024-10-08
feat: mdoc-support

auer-martin:


packages/openid4vc/src/openid4vc-holder/OpenId4vcSiopHolderService.ts
```diff
@@ -109,6 +110,14 @@ export class OpenId4VcSiopHolderService {
           challenge: nonce,
           domain: clientId,
           presentationSubmissionLocation: DifPresentationExchangeSubmissionLocation.EXTERNAL,
+          openid4vp: {
+            clientId,
+            verifierGeneratedNonce: nonce,
+            mdocGeneratedNonce: await agentContext.wallet.generateNonce(),
+            responseUri:
+              authorizationRequest.authorizationRequestPayload.response_uri ??
+              authorizationRequest.authorizationRequestPayload.response_uri,
```
TimoGlastra, line 119: this is redundant

packages/openid4vc/src/openid4vc-holder/OpenId4vcSiopHolderService.ts
```diff
@@ -109,6 +110,14 @@ export class OpenId4VcSiopHolderService {
           challenge: nonce,
           domain: clientId,
           presentationSubmissionLocation: DifPresentationExchangeSubmissionLocation.EXTERNAL,
+          openid4vp: {
```
TimoGlastra, line 113: how are these values mapped to the presentation? Ideally we don't add oid4vp specific properties to the pex service, but i can understand if it's not possible otherwise

packages/openid4vc/src/openid4vc-holder/OpenId4vcSiopHolderService.ts
```diff
@@ -109,6 +110,14 @@ export class OpenId4VcSiopHolderService {
...
+            clientId,
+            verifierGeneratedNonce: nonce,
```
TimoGlastra, line 115: I think this should be extracted from the `challenge` property in the pex service itself. As it's now redundant, and can lead to inconsistencies.

packages/openid4vc/src/openid4vc-holder/OpenId4vcSiopHolderService.ts
```diff
@@ -109,6 +110,14 @@ export class OpenId4VcSiopHolderService {
...
+            mdocGeneratedNonce: await agentContext.wallet.generateNonce(),
```
TimoGlastra, line 116: we need to make sure to reuse this nonce when encrypting the resposne (we generate a new nonce there at the moment, but these should be the same)

packages/openid4vc/src/shared/transform.ts
```diff
@@ -41,6 +45,8 @@ export function getSphereonVerifiablePresentation(
     return JsonTransformer.toJSON(verifiablePresentation) as SphereonW3cVerifiablePresentation
   } else if (verifiablePresentation instanceof W3cJwtVerifiablePresentation) {
     return verifiablePresentation.serializedJwt
+  } else if (verifiablePresentation instanceof MdocVerifiablePresentation) {
+    throw new CredoError('Mdoc verifiable credential is not yet supported.')
```
TimoGlastra, line 49:
```suggestion
    throw new CredoError('Mdoc verifiable presentation is not yet supported.')
```

packages/core/src/modules/dif-presentation-exchange/DifPresentationExchangeService.ts
```diff
@@ -185,6 +183,42 @@ export class DifPresentationExchangeService {
...
+          )
+        }
+        const mdocRecord = presentationToCreate.verifiableCredentials[0].credential
+        if (!openid4vp) {
+          throw new DifPresentationExchangeError('Missing openid4vp options for creating MDOC presentation.')
+        }
+
+        const { deviceResponseBase64Url, presentationSubmission } = await MdocDeviceResponse.openId4Vp(agentContext, {
+          mdocs: [Mdoc.fromBase64Url(mdocRecord.base64Url)],
+          presentationDefinition: presentationDefinition as DifPresentationExchangeDefinitionV2,
```
TimoGlastra, line 199: what if v1 is used?

packages/core/src/modules/dif-presentation-exchange/DifPresentationExchangeService.ts
```diff
@@ -185,6 +183,42 @@ export class DifPresentationExchangeService {
...
+        verifiablePresentationResultsWithFormat.push({
+          verifiablePresentationResult: {
+            presentationSubmission: presentationSubmission,
+            verifiablePresentation: deviceResponseBase64Url,
+            presentationSubmissionLocation: PresentationSubmissionLocation.EXTERNAL,
+          },
+          claimFormat: presentationToCreate.claimFormat,
+        })
+
+        continue
```
TimoGlastra, line 214: I think instead of `continue` we could use an else clause maybe? I had a hard time understanding this code, until i spotted this `continue` statement

packages/core/src/modules/dif-presentation-exchange/utils/credentialSelection.ts
```diff
@@ -24,13 +28,25 @@ export async function getCredentialsForRequest(
...
+  const { mdocPresentationDefinition, nonMdocPresentationDefinition } =
+    MdocDeviceResponse.partitionPresentationDefinition(presentationDefinition)
+
+  const selectResultsRaw = pex.selectFrom(nonMdocPresentationDefinition, encodedCredentials)
+
+  const selectResults: CredentialRecordSelectResults = {
     ...selectResultsRaw,
+    areRequiredCredentialsPresent:
+      mdocPresentationDefinition.input_descriptors.length >= 1
+        ? 'warn' // we don't know yet wheater the required credentials are present
```
TimoGlastra, line 46: when will we know this? (i don't see warn changed to error / info?)

packages/core/src/modules/mdoc/Mdoc.ts
```diff
@@ -0,0 +1,121 @@
...
+    return this.issuerSignedDocument.issuerSigned.issuerAuth.decodedPayload.validityInfo
+  }
+
+  public get deviceSignedNamespaces(): MdocNameSpaces {
+    if (this.issuerSignedDocument instanceof DeviceSignedDocument === false) {
+      throw new MdocError(`Cannot get 'device-namespaces from a IssuerSignedDocument. Must be a DeviceSignedDocument.`)
+    }
+
+    return this.issuerSignedDocument.allDeviceSignedNamespaces
+  }
```
TimoGlastra, line 59: isn't device signed the presentation?

packages/core/src/modules/mdoc/Mdoc.ts
```diff
@@ -0,0 +1,121 @@
...
+    const document = new Document(docType, mdocContext)
+      .useDigestAlgorithm('SHA-256')
+      .addValidityInfo(validityInfo)
+      .addDeviceKeyInfo({ deviceKey: holderPublicJwk.toJson() })
+
+    for (const [namespace, namespaceRecord] of Object.entries(namespaces)) {
+      document.addIssuerNameSpace(namespace, namespaceRecord)
+    }
+
+    const cert = X509Certificate.fromEncodedCertificate(issuerCertificate)
```
TimoGlastra, line 79: i see we're mostly using string certificates here. Do we want to use that as the primary method to pass around certificate in Credo (i haven't made my mind up yet, just thinking 🤔 ).

packages/core/src/modules/mdoc/Mdoc.ts
```diff
@@ -0,0 +1,121 @@
...
+    const issuerPrivateJwk = await getJwkFromKey(options.issuerKey ?? cert.publicKey)
```
TimoGlastra, line 80: Do we check if the issuerKey and cert key match? Can't we always just extract the key from the cert?

packages/core/src/modules/mdoc/Mdoc.ts
```diff
@@ -0,0 +1,121 @@
...
+    const issuerSignedDocument = await document.sign(
+      {
+        issuerPrivateKey: issuerPrivateJwk.toJson(),
```
TimoGlastra, line 83: this is the public key jwk, not the private key jwk right? I think the name is confusing. It's important we don't have to pass the private key, so we can support HSM

packages/core/src/modules/mdoc/Mdoc.ts
```diff
@@ -0,0 +1,121 @@
...
+        alg: issuerPrivateJwk.supportedSignatureAlgorithms[0] as 'ES256' | 'ES384' | 'ES512' | 'EdDSA',
```
TimoGlastra, line 84: I don't like the casting here, it can lead to bugs in the future. We should check if index 0 exists, and then also whether it's a supported alg by mdoc (or does it throw if the provided alg is not supported?)

packages/core/src/modules/mdoc/Mdoc.ts
```diff
@@ -0,0 +1,121 @@
...
+        issuerCertificate,
+        kid: cert.publicKey.fingerprint,
+      },
+      mdocContext
+    )
+
+    return new Mdoc(issuerSignedDocument)
+  }
+
+  public async verify(agentContext: AgentContext, options?: MdocVerifyOptions): Promise<boolean> {
```
TimoGlastra, line 94: just as an FYI trusted certificaes will need to be aligned with https://github.com/openwallet-foundation/credo-ts/pull/2052 once merged

packages/core/src/modules/mdoc/Mdoc.ts
```diff
@@ -0,0 +1,121 @@
...
+        {
+          trustedCertificates: trustedCerts.map((cert) => X509Certificate.fromEncodedCertificate(cert).rawCertificate),
+          issuerAuth: this.issuerSignedDocument.issuerSigned.issuerAuth,
+          disableCertificateChainValidation: false,
+          now: options?.now,
+        },
+        getMdocContext(agentContext)
+      )
+
+      await verifier.verifyData({ mdoc: this.issuerSignedDocument }, mdocContext)
```
TimoGlastra, line 115: does it return a verification object or something? Would be good to return more than true/false, but also something we can add later

packages/core/src/modules/mdoc/MdocApi.ts
```diff
@@ -0,0 +1,75 @@
...
+    this.mdocService = mdocService
+  }
+
+  /**
+   * Create a new Mdoc, with a spcific doctype, namespace, and validity info.
+   *
+   * @param options {MdocCreateOptions}
+   * @returns {Promise<Mdoc>}
+   */
+  public async create(options: MdocCreateOptions) {
```
TimoGlastra, line 30: does create also sign? I think in SD JWT API we use `sign` so might be nice to stay consistent?

packages/core/src/modules/mdoc/MdocContext.ts
```diff
@@ -0,0 +1,136 @@
...
+          const { data } = mac0.getRawSigningData()
+          return await agentContext.wallet.sign({
+            data: Buffer.from(data),
+            key: getJwkFromJson(jwk as JwkJson).key,
+          })
+        },
+        verify: async (input) => {
+          const { mac0, jwk, options } = input
+          const { data, signature, alg } = mac0.getRawVerificationData(options)
+          return await verifyWithJwk({ jwk, signature, data, alg })
```
TimoGlastra, line 58: does this still use the credo wallet to verify the signature? I can't see the context or something passed around. It is important. 

Aslo it's prob ok for now, but we should try to stay away from doing crypto operations outside of the wallet. Also with e.g. the p256 shared secret calcuation etc.. The more we can do in the wallet, the better.

So if you can provide a list of all cryptographic operations that are needed, we can look to move them to the wallet over time

packages/core/src/modules/mdoc/MdocContext.ts
```diff
@@ -0,0 +1,136 @@
...
+
+    x509: {
+      getIssuerNameField: (input) => {
+        const { certificate, field } = input
+        const x509Certificate = X509Certificate.fromRawCertificate(certificate)
+        return x509Certificate.getIssuerNameField(field)
+      },
+      getPublicKey: async (input) => {
+        const certificate = new x509.X509Certificate(input.certificate)
+        const key = await importX509({
```
TimoGlastra, line 86: what do these do? and why are they part of the protokol library?

packages/core/src/modules/mdoc/MdocDeviceResponse.ts
```diff
@@ -0,0 +1,193 @@
...
+    // TODO: Multiple Mdocs can have different device keys.
+    const mso = mdoc.documents[0].issuerSigned.issuerAuth.decodedPayload
+    const deviceKeyInfo = mso.deviceKeyInfo
+    if (!deviceKeyInfo?.deviceKey) {
+      throw new CredoError('Device key info is missing')
+    }
+
+    const publicDeviceJwk = COSEKey.import(deviceKeyInfo.deviceKey).toJWK()
+
+    deviceKeyInfo.deviceKey
```
TimoGlastra, line 139:
```suggestion
```

packages/core/src/modules/mdoc/MdocDeviceResponse.ts
```diff
@@ -0,0 +1,193 @@
...
+    for (const [nameSpace, nameSpaceValue] of Object.entries(options.deviceNameSpaces ?? {})) {
+      deviceResponseBuilder.addDeviceNameSpace(nameSpace, nameSpaceValue)
+    }
+
+    const deviceResponseMdoc = await deviceResponseBuilder.sign(getMdocContext(agentContext))
+
+    return {
+      deviceResponseBase64Url: TypedArrayEncoder.toBase64URL(deviceResponseMdoc.encode()),
+      presentationSubmission: MdocDeviceResponse.createPresentationSubmission({
+        id: 'MdocPresentationSubmission ' + agentContext.wallet.generateNonce(),
```
TimoGlastra, line 155: I think we can also just use an uuid?

```suggestion
        id: 'MdocPresentationSubmission ' + await agentContext.wallet.generateNonce(),
```

packages/core/src/modules/mdoc/MdocDeviceResponse.ts
```diff
@@ -0,0 +1,193 @@
...
+
+  public static async verify(agentContext: AgentContext, options: MdocDeviceResponseVerifyOptions) {
+    const verifier = new Verifier()
+    const mdocContext = getMdocContext(agentContext)
+
+    const trustedCerts =
+      options?.trustedCertificates ?? agentContext.dependencyManager.resolve(X509ModuleConfig).trustedCertificates
+
+    if (!trustedCerts) {
+      throw new MdocError('No trusted certificates found. Cannot verify mdoc.')
```
TimoGlastra, line 169: same with integraiont of pr https://github.com/openwallet-foundation/credo-ts/pull/2052

packages/core/src/modules/mdoc/MdocContext.ts
auer-martin: We just need the key derivation.

packages/core/src/modules/mdoc/MdocContext.ts
auer-martin: Please take a look at my comment. There is an issue with how we represent the key as jwk. It probably should be compressed. Nothing from the protokoll library should remain here.

packages/core/src/modules/mdoc/Mdoc.ts
auer-martin: I don't mind. We essentially also set trustedCerts as strings. That's why I have chosen this approach

```
export interface X509ModuleConfigOptions {
  /**
   *
   * Array of trusted base64-encoded certificate strings in the DER-format.
   */
  trustedCertificates?: [string, ...string[]]

```

packages/core/src/modules/mdoc/Mdoc.ts
auer-martin: yes. It can also include device-signed namespaces. Essentially self-attested claims.

packages/openid4vc/src/openid4vc-holder/OpenId4vcSiopHolderService.ts
auer-martin: I did this intentionally. I don't think we should use a challenge property. openid4vp is very specific. To me it would be completely unclear why the verification of a device response will fail after changing a e.g. domain/challenge property which have no relation at all to Mdoc?

packages/openid4vc/src/openid4vc-holder/OpenId4vcSiopHolderService.ts
auer-martin: Yeah, I also don't like it, but I don't see an easy better way to propagete the values. There are used to calculate the session transcript

packages/core/src/modules/mdoc/Mdoc.ts
auer-martin: I also think this should work. Not sure why it doesn't. It may be related to the issue I mentioned below with the jwk representation of keys.

packages/core/src/modules/mdoc/MdocContext.ts
TimoGlastra: Which comment? I can't find it

auer-martin: //expected
        //{
        //kty: 'EC',
        //x: 'OFBq4YMKg4w5fTifsytwBuJf_7E7VhRPXiNm52S3q1E',
        //y: 'EyIAXV8gyt5FcRsYHhz4ryz97rjL0uogxHO6jMZr3bg',
        //crv: 'P-256'
        //}

        //actual
        //{
        //kty: 'EC',
        //crv: 'P-256',
        //x: 'OFBq4YMKg4w5fTifsytwBuJf_7E7VhRPXiNm52S3q1ETIgBdXyDK3kVxGxgeHPiv',
        //y: 'LP3uuMvS6iDEc7qMxmvduNeBp_oWscK1x-3_1KKYDayIctdDcpXHi8HcbehAfVIK'
        //}

        //const comp = X509Certificate.fromRawCertificate(input.certificate)
        //const x = getJwkFromKey(comp.publicKey).toJson()
        //// eslint-disable-next-line @typescript-eslint/no-unused-vars
        //const { use, ...jwk } = x
        //return jwk

packages/openid4vc/src/openid4vc-holder/OpenId4vcSiopHolderService.ts
auer-martin: What do you think?

TimoGlastra: Hmm I think it's quite common to use one nonce and apply that in all places that need a nonce. That way you keep consistency. we use the presentation exchange nonce in the w3c vp signature creation, so i don't see why we can't use it in the mdoc presentation creation.

I rather reuse the existing property, as there's really no reason to use a different nonce for the mdoc presentation.

TimoGlastra: because the nonce for a w3c jwt vp / sd-jwt-vc is also based on the challenge. We just call it challenge in our api, but all VPs use this as the nonce/challenge.

packages/openid4vc/src/openid4vc-holder/OpenId4vcSiopHolderService.ts
TimoGlastra: okay and this is openid4vp specific, or is it mdoc specific?

packages/core/src/modules/x509/X509ServiceOptions.ts
```diff
@@ -22,6 +22,7 @@ export interface X509CreateSelfSignedCertificateOptions {
   notBefore?: Date
   notAfter?: Date
   name?: string
```
TimoGlastra, line 24: I think name can just be a string with `CN=X, N=X` I think that might be nicer to keep it like that. Otherwise we'll have to add a custom prop for everything

TimoGlastra: or we need to nest / rename it

packages/core/src/modules/mdoc/Mdoc.ts
```diff
@@ -0,0 +1,121 @@
...
+ * which are the actual credentials being issued to holders.
+ */
+export class Mdoc {
+  public base64Url: string
+  private constructor(private issuerSignedDocument: IssuerSignedDocument) {
+    const issuerSigned = issuerSignedDocument.prepare().get('issuerSigned')
+    this.base64Url = TypedArrayEncoder.toBase64URL(cborEncode(issuerSigned))
+  }
+
+  public static _interalFromIssuerSignedDocument(issuerSignedDocument: IssuerSignedDocument) {
```
TimoGlastra, line 25: i think if it's public we shouldn't call it _internal, but just fromIssuerSignedDocument. Or we don't  allow creating it from IssuerSignedDocument at all (but something else).

packages/core/src/modules/mdoc/Mdoc.ts
TimoGlastra: but shouldn't this be on the mdoc presentation then?

packages/core/src/modules/mdoc/Mdoc.ts
TimoGlastra: string is fine 👍

packages/core/src/modules/mdoc/MdocOptions.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+export type MdocDeviceResponseVerifyOptions = {
+  mdoc: Mdoc
+  trustedCertificates?: [string, ...string[]]
+  sessionTranscriptOptions: MdocOpenId4VpSessionTranscriptOptions
+  /**
+   * The base64Url-encoded device response string.
+   */
+  deviceResponse: string
```
TimoGlastra, line 34: Why do we need to add both the mdoc and the deviceResponse base64? The encoded device response also contains the mdoc right? 

If it's for internal APIs it's ok, but for public APIs I'd like to avoid options that could conflict (what if they don't match?)

packages/core/src/modules/mdoc/MdocOptions.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  /**
+   * The private part of the ephemeral key used in the session where the DeviceResponse was obtained. This is only required if the DeviceResponse is using the MAC method for device authentication.
```
TimoGlastra, line 37: This is not the private part. It's a public key that can be used during verificadtion to know which private key to use.

```suggestion
   * The public ephemeral key used in the session where the DeviceResponse was obtained. This is only required if the DeviceResponse is using the MAC method for device authentication.
```

Also, isn't the public key added in the device respnose, or do we need to make sure it's MAC'd to this specific key?

packages/core/src/modules/mdoc/MdocOptions.ts
```diff
@@ -0,0 +1,56 @@
...
+  docType: 'org.iso.18013.5.1.mDL' | (string & {})
+  validityInfo?: Partial<ValidityInfo>
+  namespaces: { [namespace: string]: Record<string, unknown> }
+
+  /**
+   *
+   * The trusted base64-encoded issuer certificate string in the DER-format.
+   */
+  issuerCertificate: string
+  holderPublicKey: Key
```
TimoGlastra, line 54: Either both key, or both PublicKey, but issuerKey and holderPublicKey doesn't make sense I think

```suggestion
  holderKey: Key
```

packages/core/src/modules/mdoc/MdocService.ts
```diff
@@ -0,0 +1,78 @@
...
+  public createMdoc(agentContext: AgentContext, options: MdocCreateOptions) {
+    return Mdoc.create(agentContext, options)
+  }
+
+  public async verifyMdoc(agentContext: AgentContext, mdoc: Mdoc, options: MdocVerifyOptions) {
+    return await mdoc.verify(agentContext, options)
+  }
+
+  public async createDeviceResponse(agentContext: AgentContext, options: MdocDeviceResponseOpenId4VpOptions) {
+    return MdocDeviceResponse.openId4Vp(agentContext, options)
```
TimoGlastra, line 41: Is device response an openid4vp specific object, if we're using BLE don't we use device response? If so I don't think the underlying API should be called `openId4Vp`?

TimoGlastra: Or should this method be called `createOpenId4VpDeviceResponse`?

packages/core/src/modules/mdoc/__tests__/mdoc.deviceResponse.openid4vp.test.ts
TimoGlastra, line 1: nice but usually we name tests mdoc-deviceResponse-openid4vc.test.ts instead of using dots (which are more for extension types)

packages/core/src/modules/mdoc/__tests__/mdoc.deviceResponse.openid4vp.test.ts
```diff
@@ -0,0 +1,305 @@
...
+
+    const issuerPrivateAskar = AskarKey.fromJwk({ jwk: Jwk.fromJson(ISSUER_PRIVATE_KEY_JWK) })
+    const issuerPrivateKey = await agent.context.wallet.createKey({
+      keyType: KeyType.P256,
+      privateKey: Buffer.from(issuerPrivateAskar.secretBytes),
+    })
+
+    // this is the ISSUER side
+    {
+      mdoc = await Mdoc.create(agent.context, {
```
TimoGlastra: do we also want to add an oid4vci flow with mdoc? OR lateR?

packages/core/src/modules/mdoc/__tests__/mdoc.deviceResponse.openid4vp.test.ts
```diff
@@ -0,0 +1,305 @@
...
+      [
+        'mdocGeneratedNonce',
+        {
+          clientId,
+          responseUri,
+          verifierGeneratedNonce,
+          mdocGeneratedNonce: 'wrong',
+        },
+      ] as const,
+    ].forEach(([name, values]) => {
```
TimoGlastra, line 267: I think this might look a bit clear?

```suggestion
    Object.entries({
      clientId: {
          clientId: 'wrong',
          responseUri,
          verifierGeneratedNonce,
          mdocGeneratedNonce,
        } 
        }).forEach(([name, values]) => {
```

TimoGlastra (commented):
Nice 👍

packages/openid4vc/src/openid4vc-holder/OpenId4vcSiopHolderService.ts
auer-martin: Hmm I guess depends on the viewpoint the data we need is expected to come for openid4vp, but we need it to create mdoc presentations.

TimoGlastra: Okay so if we implement BLE exchange we wouldn't need these values?

packages/core/src/modules/mdoc/__tests__/mdoc.deviceResponse.openid4vp.test.ts
auer-martin: I would do that later. Let's first get the basis merged?

packages/core/src/modules/mdoc/Mdoc.ts
auer-martin: doesnt return anything atm.

packages/openid4vc/src/openid4vc-holder/OpenId4vcSiopHolderService.ts
auer-martin: No we need it to compute the session transcript bytes. 

```
SessionTranscriptBytes = #6.24(bstr .cbor SessionTranscript)
SessionTranscript = [
    DeviceEngagementBytes,
    EReaderKeyBytes,
    Handover
]
DeviceEngagementBytes = #6.24(bstr .cbor DeviceEngagement)
Handover = QRHandover / NFCHandover
QRHandover= null
NFCHandover = [
    bstr        ; Binary value of the Handover Select Message
    bstr / null ; Binary value of the Handover Request Message,
; shall be null if NFC Static Handover was used DeviceEngagement is defined in 8.2.1.1.
]
```

packages/core/src/modules/mdoc/Mdoc.ts
auer-martin: A presentation is a set of mdoc's. If we want to go that route we need to split Mdoc class up into IssuerSigned / Device signed document class.

packages/core/src/modules/mdoc/MdocService.ts
auer-martin: renamed to createOpenId4VpDeviceResponse

packages/core/src/modules/mdoc/MdocOptions.ts
auer-martin: with berends fix we now just need 

```
  issuerCertificate: string
  holderKey: Key
```

packages/core/src/modules/mdoc/Mdoc.ts
auer-martin: works now with the fixes regarding key compression

packages/core/src/modules/mdoc/MdocOptions.ts
auer-martin: I removed this for now. As currently we never use mac authentication

packages/core/src/modules/mdoc/Mdoc.ts
```diff
@@ -0,0 +1,176 @@
...
+  public async verify(
+    agentContext: AgentContext,
+    options?: MdocVerifyOptions
+  ): Promise<{ isValid: true } | { isValid: false; error: string }> {
+    let trustedCerts: [string, ...string[]] | undefined
+
+    if (options?.trustedCertificates) {
+      trustedCerts = options.trustedCertificates
+    } else if (options?.verificationContext) {
+      agentContext.dependencyManager.resolve(X509ModuleConfig).getTrustedCertificatesForVerification
```
TimoGlastra, line 145:
```suggestion
```

packages/core/src/modules/mdoc/Mdoc.ts
```diff
@@ -0,0 +1,176 @@
...
+    try {
+      const verifier = new Verifier()
+      await verifier.verifyIssuerSignature(
+        {
+          trustedCertificates: trustedCerts.map((cert) => X509Certificate.fromEncodedCertificate(cert).rawCertificate),
+          issuerAuth: this.issuerSignedDocument.issuerSigned.issuerAuth,
+          disableCertificateChainValidation: false,
+          now: options?.now,
+        },
+        getMdocContext(agentContext)
```
TimoGlastra, line 167: Not important but We also create the context above, can we reuse it?

packages/core/src/modules/mdoc/MdocOptions.ts
```diff
@@ -0,0 +1,58 @@
+import type { Mdoc } from './Mdoc'
+import type { Key } from '../../crypto/Key'
+import type { DifPresentationExchangeDefinition } from '../dif-presentation-exchange'
+import type { ValidityInfo, MdocNameSpaces } from '@protokoll/mdoc-client'
+
+export type { MdocNameSpaces } from '@protokoll/mdoc-client'
+
+export interface MdocVerificationContext {
```
TimoGlastra, line 8: Can't we reuse global other verification context for this?

TimoGlastra (approved):

packages/core/src/modules/mdoc/MdocOptions.ts
auer-martin: For the w3cCredentials the context looks like this. Or what global context do you mean?

export interface VerificationContext {
  /**
   * The `id` of the `ProofRecord` that this verification is bound to.
   */
  didcommProofRecordId?: string

  /**
   * The `id` of the `OpenId4VcVerificationSessionRecord` that this verification is bound to.
   */
  openId4VcVerificationSessionId?: string
}

PR #2069, opened 2024-10-25
feat: openid4vci mdoc-issuanc

auer-martin:


demo-openid/src/Issuer.ts
```diff
@@ -110,6 +126,25 @@ function getCredentialRequestToCredentialMapper({
...
+      const holderKey = await agentContext.dependencyManager
+        .resolve(MdocService)
+        .getKeyFromMdocCredentialHolderBinding(agentContext, holderBinding)
+
+      return {
+        credentialSupportedId: universityDegreeCredentialMdoc.id,
+        format: ClaimFormat.MsoMdoc,
+        docType: universityDegreeCredentialMdoc.doctype,
+        issuerCertificate: trustedCertificates[0],
+        holderKey,
```
TimoGlastra, line 139: Can't we return the holder binding like we do with sd-jwt?

packages/core/src/modules/mdoc/MdocOptions.ts
```diff
@@ -56,3 +57,15 @@ export type MdocSignOptions = {
   issuerCertificate: string
   holderKey: Key
 }
+
+export type MdocCredentialHolderDidBinding = {
+  method: 'did'
+  didUrl: string
+}
```
TimoGlastra, line 64: Can you bind a mdoc to a did? I think only a key right?

packages/core/src/modules/mdoc/MdocService.ts
```diff
@@ -37,6 +40,29 @@ export class MdocService {
     return await mdoc.verify(agentContext, options)
   }
 
+  public async getKeyFromMdocCredentialHolderBinding(
+    agentContext: AgentContext,
+    holderBinding: MdocCredentialHolderBinding
+  ) {
+    let holderKey: Key
+    if (holderBinding.method !== 'jwk') {
```
TimoGlastra, line 48:
```suggestion
    if (holderBinding.method === 'did') {
```


But again, i don't think we should allow the did holder binding without the did actually being put in the mdoc (like is the case with sd-jwt, the binding actually represents what gets put in the cred, did or jwk). So do we need mdoc holder binding at all? I think jwk is closest to what gets put in the actual mdoc

TimoGlastra (approved):
Nice!! Some small suggestions

demo-openid/src/Issuer.ts
auer-martin: We can, but then we need to decide internally how to e.g. extract a key from a did-holder biding with mdoc, which I think should be up to the user?

TimoGlastra: Hmm how to extract the key is generally straigtforward. You resolve the did, find the key with the key id, transform it to a key instance (like you do now). I think we should just not allow did holder binding when using mdoc?

TimoGlastra: hmm but I understand now what you mean, it will break the API. Okay let's keep it like this 👍 

Could we also add the `key` to the binding / credential request mapper? Then you at least don't have to call the mdoc service.

TimoGlastra:
LGTM but CI is failing

TimoGlastra:
Still failing :(

PR #2071, opened 2024-10-29
fix: mdoc fixes

TimoGlastra:
Two fixes for mdoc:
- update mdoc client to fix react native issues
- proper handling of cose_key (gave issues with BDR pid issuer)


TimoGlastra:
No, but it's weird that you get that error as the library it's complaining about is not a module.

Probably something related to ts-node 🤔

PR #2072, opened 2024-10-29
feat: support w3c revocation

GHkrishna:
## What
This PR introduces support for W3C credential revocation. The following changes have been made:

Refactor functionality for Verify Credential and Verify Presentation.
Integrated with the Bit String Status List for checking and updating credential status.

## How
Revocation Flow: Implemented the revocation flow by checking a credential status and updating the credential issuance process to handle revocable credentials.
Integration: Integrated the new revocation features with the existing system, ensuring compatibility and proper handling of revoked credentials.


GHkrishna:
Hello @TimoGlastra, wanting to create a separate PR for w3c revocation, due to some conflicts in the original PR #2024 .
The work until now, I have updated with the origin authors name.

packages/core/src/modules/vc/models/credential/w3c-credential-status/bitstring-status-list/BitStringStatusList.ts
```diff
@@ -0,0 +1,145 @@
...
+import { IsEnum, IsNumber, IsString } from 'class-validator'
+
+import { JsonLdCredentialDetail, W3cCredentialSubject } from '../../../../../..'
+import { W3cCredential } from '../../W3cCredential'
+import { W3cCredentialStatus } from '../W3cCredentialStatus'
+
+// The purpose can be anything apart from this as well
+export enum BitstringStatusListCredentialStatusPurpose {
+  'revocation' = 'revocation',
+  'suspension' = 'suspension',
```
TimoGlastra:
```suggestion
  Revocation = 'revocation',
  Suspension = 'suspension',
```

packages/core/src/modules/vc/models/credential/w3c-credential-status/bitstring-status-list/BitStringStatusList.ts
```diff
@@ -0,0 +1,145 @@
...
+
+// // Define an interface for the `credential` object that uses `CredentialSubject`
+// export interface Credential {
+//   credentialSubject: CredentialSubject
+// }
+
+// // Use the `Credential` interface within `BitStringStatusListCredential`
+// export interface BitStringStatusListCredential {
+//   credential: Credential
+// }
```
TimoGlastra: Can you remove unused code?

packages/core/src/modules/vc/models/credential/w3c-credential-status/bitstring-status-list/BitStringStatusList.ts
```diff
@@ -0,0 +1,145 @@
...
+    issuanceDate: string
+    credentialSubject: BitStringStatusListCredentialSubject
+  }) {
+    super({
+      ...options,
+      type: ['VerifiableCredential', BitStringStatusListEntry.type],
+    })
+  }
+
+  @IsString()
```
TimoGlastra: This shouldn't be a string? It's an object right?

packages/core/src/modules/vc/models/credential/w3c-credential-status/bitstring-status-list/VerifyBitStringCredentialStatus.ts
```diff
@@ -0,0 +1,73 @@
...
+  url: string
+): Promise<BitStringStatusListCredential> => {
+  const response = await agentContext.config.agentDependencies.fetch(url, { method: 'GET' })
+
+  if (!response.ok) {
+    throw new CredoError(`Failed to fetch bit string status list. HTTP Status: ${response.status}`)
+  }
+
+  try {
+    return (await response.json()) as BitStringStatusListCredential
```
TimoGlastra, line 31: The bitstrintstatuslistcredential is a class. We should transform and validate the json to a class instance

packages/core/src/modules/vc/jwt-vc/W3cJwtCredentialService.ts
```diff
@@ -538,4 +539,11 @@ export class W3cJwtCredentialService {
 
     return verificationMethod
   }
+
+  // temporarily disable no unused var
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  public async revokeCredential(_agentContext: AgentContext, _options: RevokeCredentialOptions) {
+    // revoke jwt cred
+    throw new CredoError(`Revocation support not implemented for jwtVc`)
```
TimoGlastra: Why add these methods if both jwt and jsonld are not implemented?

packages/core/src/modules/vc/W3cCredentialService.ts
```diff
@@ -186,6 +187,35 @@ export class W3cCredentialService {
...
+   *
+   * @param credentialRecordId The id of the credential record for which to revoke the credential
+   * @returns Revoke credential notification message
+   *
+   */
+  public async revokeCredential<Format extends ClaimFormat.JwtVp | ClaimFormat.LdpVp>(
+    agentContext: AgentContext,
+    options: RevokeCredentialOptions
+  ) {
+    const credentialRecod = await this.getCredentialRecordById(agentContext, options.credentialRecordId)
```
sairanjit, line 202: - I think we should not use the credentialRecord to get tags as credentialRecord may or may not be present in wallet. To get the claimFormat we can use options

packages/core/src/modules/vc/models/credential/w3c-credential-status/bitstring-status-list/BitStringStatusList.ts
sairanjit, line 11:
```suggestion
  Revocation = 'revocation',
  Suspension = 'suspension',
```

sairanjit (changes requested):

packages/core/src/modules/vc/W3cCredentialService.ts
GHkrishna: Ahh yes you are right. Also as the specs suggest the issuer may not always be the one revoking the credential. Will change it

GHkrishna: Updated this

GHkrishna:
As per the discussion in the community call(with @TimoGlastra and @genaris ), the index allocation logic and the logic to revoke credential should be part of the revokers preference and credo must only be responsible for signing the BitSting status list credential and issuing a revocable credential containing a valid credential status property.

GHkrishna:
Hello @genaris @TimoGlastra 
Need your feedback on the PR, since, it has a risk of getting stale.

I understand you guys are quite occupied with all the interesting work going on in credo. Since this PR is quite raw, your feedback is quite essential to accommodate changes based on them.

Thank you once again.

packages/core/src/modules/vc/models/credential/w3c-credential-status/bitstring-status-list/BitStringStatusList.ts
```diff
@@ -0,0 +1,131 @@
...
+export interface BitStringStatusListStatusMessage {
+  // a string representing the hexadecimal value of the status prefixed with 0x
+  status: string
+  // a string used by software developers to assist with debugging which SHOULD NOT be displayed to end users
+  message?: string
+  // We can have some key value pairs as well
+  [key: string]: unknown
+}
+
+export interface IBitStringStatusListEntryOptions {
```
genaris:
```suggestion
export interface BitStringStatusListEntryOptions {
```

Mostly for consistency with other interface naming along the repo

packages/didcomm/src/modules/credentials/formats/jsonld/JsonLdCredentialFormat.ts
```diff
@@ -1,5 +1,6 @@
 import type { JsonObject, SingleOrArray, W3cIssuerOptions } from '@credo-ts/core'
 import type { CredentialFormat } from '../CredentialFormat'
+import { CredentialStatusBasedOnType } from '@credo-ts/core/build/modules/vc/models/credential/w3c-credential-status/W3cCredentialStatus'
```
genaris: If it's not the case, I think it would be good to expose `CredentialStatusBasedOnType` so it is easily used by other modules.

packages/core/package.json
```diff
@@ -58,6 +58,8 @@
     "lru_map": "^0.4.1",
     "make-error": "^1.3.6",
     "object-inspect": "^1.10.3",
+    "pako": "^2.1.0",
```
genaris, line 62: Does pako work fine in react native? How does it compare with other modules such as [fflate](https://www.npmjs.com/package/fflate)? I'm asking this because we want core to be as smaller as possible and it looks like pako is a bit heavy.

packages/core/src/modules/vc/models/credential/w3c-credential-status/W3cCredentialStatus.ts
```diff
@@ -0,0 +1,35 @@
...
+      this.type = options.type
+    }
+  }
+
+  @IsUri()
+  @IsString()
+  public id!: string
+
+  @IsString()
+  // @IsEnum(W3cCredentialStatusSupportedTypes, { message: 'Invalid credential status type' })
```
genaris: We can maybe remove this commented line?

packages/core/src/modules/vc/models/credential/w3c-credential-status/bitstring-status-list/BitStringStatusList.ts
GHkrishna: Agreed n updated

packages/core/package.json
GHkrishna: Okay, that's a good suggestion, I'm all in on keeping core as small as possible, 
Pako did work fine with an android device.
I was not aware about fflate, will have a look at it.

GHkrishna:
> I think it would be good to add some tests as well.

Yes sure, I just wanted to make sure, we are on the right track before starting with the test cases. I think this is a good time to start now.

GHkrishna:
Ariel, currently I'm adding testcases for revocable credential in addition to that of the existing ones(non-revocable w3c) instead of replacing/editing them.

packages/core/package.json
TimoGlastra: Pako is also used by SD-JWT JS library, works in react native, and since we already depend on it is maybe not a bad choice. I have no idea on the size .

TimoGlastra:
@GHkrishna could you update the PR with the latest main, and fix the CI tests?

PR #2080, opened 2024-11-01
feat: openid4vp-mdoc

auer-martin:


packages/core/package.json
```diff
@@ -35,16 +35,17 @@
...
     "@peculiar/x509": "^1.11.0",
-    "@protokoll/mdoc-client": "0.2.35",
+    "@protokoll/mdoc-client": "0.2.36",
     "@sd-jwt/core": "^0.7.0",
     "@sd-jwt/decode": "^0.7.0",
     "@sd-jwt/jwt-status-list": "^0.7.0",
     "@sd-jwt/sd-jwt-vc": "^0.7.0",
     "@sd-jwt/types": "^0.7.0",
     "@sd-jwt/utils": "^0.7.0",
     "@sphereon/pex": "5.0.0-unstable.18",
```
TimoGlastra, line 45: can you update this to unstable `5.0.0-unstable.25`

packages/core/package.json
```diff
@@ -35,16 +35,17 @@
...
+    "@sphereon/kmp-mdl-mdoc": "0.2.0-SNAPSHOT.22",
```
TimoGlastra, line 46: We don't need this right?

packages/core/src/modules/dif-presentation-exchange/utils/credentialSelection.ts
```diff
@@ -41,6 +42,11 @@ export async function getCredentialsForRequest(
 
   const selectResults: CredentialRecordSelectResults = {
     ...selectResultsRaw,
+    areRequiredCredentialsPresent:
+      nonMdocPresentationDefinition.input_descriptors.length === 0 &&
+      mdocPresentationDefinition.input_descriptors.length > 0
+        ? Status.INFO
+        : selectResultsRaw.areRequiredCredentialsPresent,
```
TimoGlastra, line 49: i don't understand this? mdoc also handle mdoc selecting now right? And what if the mdoc is not available, it should'nt be info the nright?

packages/core/src/modules/dif-presentation-exchange/utils/transform.ts
```diff
@@ -32,8 +34,10 @@ export function getSphereonOriginalVerifiablePresentation(
     verifiablePresentation instanceof W3cJsonLdVerifiablePresentation
   ) {
     return verifiablePresentation.encoded as SphereonOriginalVerifiablePresentation
-  } else if (verifiablePresentation instanceof MdocVerifiablePresentation) {
-    throw new CredoError('Mdoc verifiable presentation is not yet supported by Sphereon.')
+  } else if (verifiablePresentation instanceof MdocDeviceResponse) {
+    return com.sphereon.mdoc.data.device.DeviceResponseCbor.Static.cborDecode(
+      new Int8Array(TypedArrayEncoder.fromBase64(verifiablePresentation.base64Url))
+    )
```
TimoGlastra, line 40: original can just be a string for PEX / ssi-types now. That way we do'nt have to depend on / import the sphereon mdl lib

packages/core/src/modules/mdoc/MdocDeviceResponse.ts
```diff
@@ -26,7 +27,41 @@ import { getMdocContext } from './MdocContext'
...
+  public static isBase64DeviceResponse(base64Url: string) {
+    try {
+      const parsed = parseDeviceResponse(TypedArrayEncoder.fromBase64(base64Url))
+      if (parsed.status === MDocStatus.OK) return true
+      return false
+    } catch (error) {
+      // no-op
+    }
+
+    return false
```
TimoGlastra, line 41:
```suggestion
    try {
      const parsed = parseDeviceResponse(TypedArrayEncoder.fromBase64(base64Url))
      return parsed.status === MDocStatus.OK
    } catch (error) {
      return false
    }
```

TimoGlastra: Also seems quite expensive. That we try to parse. But maybe the best way. Are we doing in places:
```
if (isBase64DeviceReponse(data)) {
   parseBase64DeviceResponse()
}
```

Because in that case a `tryParseBase64DeviceResponse` that returns null on error might be better?

packages/core/src/modules/mdoc/MdocDeviceResponse.ts
```diff
@@ -26,7 +27,41 @@ import { getMdocContext } from './MdocContext'
...
+  }
+
+  public static fromBase64Url(base64Url: string) {
+    const parsed = parseDeviceResponse(TypedArrayEncoder.fromBase64(base64Url))
+    if (parsed.status !== MDocStatus.OK) {
```
TimoGlastra, line 46: no extra error provided in parsed?

packages/core/src/modules/mdoc/MdocDeviceResponse.ts
```diff
@@ -26,7 +27,41 @@ import { getMdocContext } from './MdocContext'
...
+      throw new MdocError(`Parsing Mdoc Device Response failed.`)
+    }
+
+    const documents = parsed.documents.map((doc) => {
+      const prepared = doc.prepare()
+      const docType = prepared.get('docType') as string
+      const issuerSigned = cborEncode(prepared.get('issuerSigned'))
+      const deviceSigned = cborEncode(prepared.get('deviceSigned'))
+
+      return Mdoc.fromIssuerSignedDocument(
```
TimoGlastra, line 56: fromIssuerSigned and then pasing deviceSigned is a bit weird maybe?

packages/core/src/modules/mdoc/MdocDeviceResponse.ts
```diff
@@ -113,7 +148,18 @@ export class MdocDeviceResponse {
 
     const inputDescriptor = this.assertMdocInputDescriptor(options.inputDescriptor)
     const _mdoc = parseIssuerSigned(TypedArrayEncoder.fromBase64(mdoc.base64Url), mdoc.docType)
-    return mdocLimitDisclosureToId({ mdoc: _mdoc, inputDescriptor })
+
+    const disclosure = mdocLimitDisclosureToId(_mdoc, inputDescriptor)
```
TimoGlastra, line 152: Id is maybe a bit offputting as `inputDescriptor` ( i thought which ID?). So mabye just `mdocLimitDisclosureToInputDescriptor`?

packages/openid4vc/package.json
```diff
@@ -27,12 +27,13 @@
...
   "dependencies": {
     "@credo-ts/core": "workspace:*",
-    "@sphereon/did-auth-siop": "0.16.1-next.168",
-    "@sphereon/oid4vc-common": "0.16.1-next.168",
-    "@sphereon/oid4vci-client": "0.16.1-next.168",
-    "@sphereon/oid4vci-common": "0.16.1-next.168",
-    "@sphereon/oid4vci-issuer": "0.16.1-next.168",
-    "@sphereon/ssi-types": "^0.30.1",
+    "@sphereon/did-auth-siop": "0.16.1-fix.173",
+    "@sphereon/kmp-mdl-mdoc": "0.2.0-SNAPSHOT.22",
```
TimoGlastra, line 31: again, would be good to not depend on this.

packages/openid4vc/package.json
```diff
@@ -27,12 +27,13 @@
...
+    "@sphereon/oid4vc-common": "0.16.1-fix.173",
+    "@sphereon/oid4vci-client": "0.16.1-fix.173",
+    "@sphereon/oid4vci-common": "0.16.1-fix.173",
+    "@sphereon/oid4vci-issuer": "0.16.1-fix.173",
+    "@sphereon/ssi-types": "0.30.2-next.129",
```
TimoGlastra, line 36: also update this one

packages/openid4vc/tests/openid4vc.e2e.test.ts
```diff
@@ -1748,4 +1743,735 @@ describe('OpenId4Vc', () => {
...
+            _sd: [expect.any(String), expect.any(String)],
+            _sd_alg: 'sha-256',
+            cnf: {
+              kid: 'did:key:z6MkpGR4gs4Rc3Zph4vj8wRnjnAxgAPSxcR8MAVKutWspQzc#z6MkpGR4gs4Rc3Zph4vj8wRnjnAxgAPSxcR8MAVKutWspQzc',
+            },
+            iat: expect.any(Number),
+            iss: 'did:key:z6MkrzQPBr4pyqC776KKtrz13SchM5ePPbssuPuQZb5t4uKQ',
+            vct: 'OpenBadgeCredential2',
+            degree: 'bachelor2',
+          },
```
TimoGlastra, line 2462: we should spit this up at some point 😅 but ok for now

packages/openid4vc/src/shared/transform.ts
```diff
@@ -78,9 +76,12 @@ export function getVerifiablePresentationFromSphereonWrapped(
     } satisfies SdJwtVc
   } else if (wrappedVerifiablePresentation.format === 'mso_mdoc') {
     if (typeof wrappedVerifiablePresentation.original !== 'string') {
```
TimoGlastra, line 78: i think the original will ALWAYS be string if we also change the other code to always provide a string from the start. I think it's ok  as we don't need to import the library here though. Just want to be careful in depending on two mdl librs. Ideally we jus tprovide string and get back string from PEX. For the rest it should be a black box (so we're also not prone to changes to their library)

packages/openid4vc/src/openid4vc-verifier/OpenId4VcSiopVerifierService.ts
```diff
@@ -608,9 +624,28 @@ export class OpenId4VcSiopVerifierService {
 
         let isValid: boolean
 
-        // TODO: it might be better here to look at the presentation submission to know
-        // If presentation includes a ~, we assume it's an SD-JWT-VC
-        if (typeof encodedPresentation === 'string' && encodedPresentation.includes('~')) {
```
TimoGlastra, line 613: let's keep sd-jwt first, as it's cheaper to check for inclusion of `~` than to parse the mdoc.

Also this is what i meant in other comment with check if `isBase64DeviceResponse` and then `MdocDeviceResponse.fromBase64Url`. Might as well save the result from the first call and then using that to not do the whole cbor decoding twice.

packages/openid4vc/src/openid4vc-verifier/OpenId4VcSiopVerifierService.ts
```diff
@@ -608,9 +624,28 @@ export class OpenId4VcSiopVerifierService {
...
+        if (typeof encodedPresentation === 'string' && MdocDeviceResponse.isBase64DeviceResponse(encodedPresentation)) {
+          if (!options.responseUri || !options.mdocGeneratedNonce) {
+            isValid = false
```
TimoGlastra, line 629: let's add a reason to the return value. as this is very hard to debug that you didn't provide these values: https://github.com/Sphereon-Opensource/OID4VC/blob/3b516bef69484cf7af370ea0213e486f4f701d28/packages/siop-oid4vp/lib/authorization-response/types.ts#L94

packages/openid4vc/src/openid4vc-verifier/OpenId4VcSiopVerifierService.ts
```diff
@@ -553,9 +563,9 @@ export class OpenId4VcSiopVerifierService {
         // to fix that in Sphereon lib
         client_id: clientId,
         passBy: PassBy.VALUE,
```
TimoGlastra, line 565: This one is also included in the final metadata. I think we can jsut remove it?

packages/openid4vc/src/openid4vc-verifier/OpenId4VcSiopVerifierService.ts
```diff
@@ -229,6 +232,7 @@ export class OpenId4VcSiopVerifierService {
     const requestClientId = await authorizationRequest.getMergedProperty<string>('client_id')
     const requestNonce = await authorizationRequest.getMergedProperty<string>('nonce')
     const requestState = await authorizationRequest.getMergedProperty<string>('state')
+    const responseUri = await authorizationRequest.getMergedProperty<string>('response_uri')
```
TimoGlastra, line 235: in other code you tried both `response_uri` and `redirect_uri` is it ok to jsut take the response uri here?

TimoGlastra (approved):
Good stuff, nice!!

packages/openid4vc/src/openid4vc-verifier/OpenId4VcSiopVerifierService.ts
auer-martin: mdoc requires using response_uri. So I think we should be fine

packages/core/src/modules/mdoc/MdocDeviceResponse.ts
auer-martin: No, not really atm

packages/core/src/modules/dif-presentation-exchange/utils/credentialSelection.ts
auer-martin: we set it to error below when we also check if all the mdoc credentials are available.

packages/openid4vc/src/openid4vc-verifier/OpenId4VcSiopVerifierService.ts
auer-martin: that does not work. Many failing tests.

TimoGlastra: Hmm okay, this should be fixed in sphereon's lib then.

packages/openid4vc/src/openid4vc-verifier/OpenId4VcSiopVerifierService.ts
```diff
@@ -624,7 +642,26 @@ export class OpenId4VcSiopVerifierService {
...
+                clientId: options.audience,
+                mdocGeneratedNonce: options.mdocGeneratedNonce,
+                responseUri: options.responseUri,
+                verifierGeneratedNonce: options.nonce,
+              },
+              verificationContext: {
+                openId4VcVerificationSessionId: options.correlationId,
+              },
+            })
+            isValid = true
```
TimoGlastra, line 662: By the way, does thi throw if it fails, or should we set the `isValid` based on the verify result?

PR #2088, opened 2024-11-09
feat! auth code flow

TimoGlastra:
See #2056


packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
```diff
@@ -555,63 +495,108 @@ export class OpenId4VciHolderService {
...
-    // if (!supportedVerificationMethods.includes(verificationMethod.type)) {
-    //   const supportedVerificationMethodsString = supportedVerificationMethods.join(', ')
-    //   throw new CredoError(
-    //     `Verification method uses verification method type '${verificationMethod.type}', but only '${supportedVerificationMethodsString}' verification methods are supported for key type '${JwkClass.keyType}'`
-    //   )
-    // }
+    const alg = jwk.supportedSignatureAlgorithms.find((alg) => signatureAlgorithms.includes(alg))
+    if (!alg) {
+      // Should not happen, to make ts happy
+      throw new CredoError(`Unable to determien alg for key type ${jwk.keyType}`)
```
berendsliedrecht, line 583:
```suggestion
      throw new CredoError(`Unable to determine alg for key type ${jwk.keyType}`)
```

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
```diff
@@ -3,82 +3,65 @@ import type {
   OpenId4VciAuthCodeFlowOptions,
```
berendsliedrecht, line 3: Not really super concrete feedback but this file is getting very large and has some complex to read error handling. It might be good in the future to just create single use functions for the validation, i.e. `getDPoPSigningAlgorithm`, which handles the error as well. Reading the main functionality is a lot easier then.

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderServiceOptions.ts
```diff
@@ -43,44 +37,47 @@ export interface OpenId4VciNotificationMetadata {
...
-
 export type OpenId4VciRequestTokenResponse = {
   accessToken: string
   cNonce?: string
-  dpop?: { jwk: Jwk; nonce?: string }
+  dpop?: OpenId4VciDpopRequestOptions
 }
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type UnionToArrayUnion<T> = T extends any ? T[] : never
```
berendsliedrecht, line 47: which type does not extend `any`? `never`?

packages/openid4vc/src/openid4vc-holder/__tests__/fixtures.ts
```diff
@@ -39,6 +42,8 @@ export const matrrLaunchpadDraft11JwtVcJson = {
       {
         id: 'b4c4cdf5-ccc9-4945-8c19-9334558653b2',
         format: 'ldp_vc',
+        // FIXME: should we make this optional? Seems a lot of types break on this
```
berendsliedrecht, line 45:
```suggestion
        // TODO: should we make this optional? Seems a lot of types break on this
```

Do we differentiate between FIXME and TODO? We should keep it conistent if they mean the same.

Might also be nice to have something like `// TODO(oid): bla bla`

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerModule.ts
```diff
@@ -116,21 +127,38 @@ export class OpenId4VcIssuerModule implements Module {
...
       const { agentContext } = getRequestContext(req)
       await agentContext.endSession()
+
       next()
     })
 
     // This one will be called for all errors that are thrown
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    contextRouter.use(async (_error: unknown, req: OpenId4VcIssuanceRequest, _res: unknown, next: any) => {
+    contextRouter.use(async (_error: unknown, req: OpenId4VcIssuanceRequest, res: Response, next: any) => {
```
berendsliedrecht, line 148: next is `NextFunction` if I am not mistaken

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerService.ts
```diff
@@ -3,97 +3,90 @@ import type {
   OpenId4VciCreateCredentialOfferOptions,
   OpenId4VciCreateIssuerOptions,
   OpenId4VciPreAuthorizedCodeFlowConfig,
-  OpenId4VcIssuerMetadata,
-  OpenId4VciSignSdJwtCredential,
-  OpenId4VciSignW3cCredential,
-  OpenId4VciSignMdocCredential,
+  OpenId4VciSignW3cCredentials,
```
berendsliedrecht, line 6: Same comment as with the holder service file. Would really benefit from a split up in files and some helper functions to make it a lot easier to read.

packages/openid4vc/src/openid4vc-issuer/repository/OpenId4VcIssuanceSessionRecord.ts
```diff
@@ -1,39 +1,104 @@
...
+   * is used to bind the subsequent Authorization Request with the Credential Issuer to a context set up during previous steps.
+   */
+  issuerState?: string
+}
+
+export interface OpenId4VcIssuanceSessionPresentation {
+  /**
+   * Whether presentation during issuance is required.
+   */
+  required: true
```
berendsliedrecht, line 33: should this not be a `boolean`?

packages/openid4vc/src/openid4vc-issuer/router/accessTokenEndpoint.ts
```diff
@@ -1,224 +1,197 @@
...
+    const fullRequestUrl = joinUriParts(issuerMetadata.credentialIssuer.credential_issuer, [
+      config.accessTokenEndpointPath,
+    ])
+    const requestLike = {
+      headers: new Headers(request.headers as Record<string, string>),
+      method: request.method as HttpMethod,
+      url: fullRequestUrl,
+    } as const
+
+    // What error does this throw?
```
berendsliedrecht, line 51: does this need a TODO?

berendsliedrecht (approved):
There are quite some TODO's still, but I assume everything that should be fixed before merging will be fixed :). Good addition @TimoGlastra !!

demo-openid/package.json
```diff
@@ -10,15 +10,19 @@
   "license": "Apache-2.0",
   "scripts": {
     "issuer": "ts-node src/IssuerInquirer.ts",
+    "provider": "tsx src/Provider.ts",
```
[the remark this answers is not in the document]
TimoGlastra: ah yeah, provider needs esm support which does not work with ts-node. I'll update them all to tsx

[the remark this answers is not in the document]
TimoGlastra: Oh wait the issue is:
- tsx supports esm but no decorators (so works for provider)
- ts-node does nont support esm, but does support decorators (so works for issuer/holder/verifier)
-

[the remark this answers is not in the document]
TimoGlastra: did my best but no luck. We will either have to update to the new decorator support (still to experimental i think), or move away from decorators.

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
```diff
@@ -382,27 +316,22 @@ export class OpenId4VciHolderService {
...
-    const { credentialsToRequest, credentialBindingResolver, verifyCredentialStatus } = acceptCredentialOfferOptions
+    const { metadata, offeredCredentialConfigurations } = resolvedCredentialOffer
+    const { credentialConfigurationIds, credentialBindingResolver, verifyCredentialStatus } =
+      acceptCredentialOfferOptions
+    const client = this.getClient(agentContext)
 
-    if (credentialsToRequest?.length === 0) {
-      this.logger.warn(`Accepting 0 credential offers. Returning`)
-      return []
+    if (credentialConfigurationIds?.length === 0) {
```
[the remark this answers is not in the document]
TimoGlastra: Yes because then we take all offered credentials. So we only want to catch the case here where the user provided credential configurations, but it's an empty array

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
TimoGlastra: Yes agreed -- it's a mostly from the pattern we had that all logic is in the service, but it's getting a bit out of hand. Do you want me to tackle as part of this PR?

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderServiceOptions.ts
TimoGlastra: It maps from `TypeA | TypeB` to `Array<TypeA> | Array<TypeB>`. I don't really know why it works, but it's different that doing `Array<TypeA | TypeB>`

packages/openid4vc/src/openid4vc-issuer/repository/OpenId4VcIssuanceSessionRecord.ts
TimoGlastra: The object will never be defined it it's not true. So i think `true` is more clear here. Alternative is empty object to indicate it's required but that felt not explicit.

Or another top-level prop for `presentationRequired` and then `presentation` for the details, but then you can still have that `presentation` is undefined (in TS) even if `required` is true. So no if you check for `required == true`, the other props will also be defined

packages/openid4vc/src/openid4vc-holder/__tests__/fixtures.ts
TimoGlastra: They are different. I fixed quite a few and renamed some to todo

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
berendsliedrecht: Can be a separate PR :)

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderServiceOptions.ts
berendsliedrecht: Ahh okay. Maybe add a comment with this in there? Was quite confusing haha

PR #2108, opened 2024-11-25
fix(askar): secure environment get public key is now async

berendsliedrecht:


packages/askar/package.json
```diff
@@ -46,7 +46,7 @@
   },
   "peerDependencies": {
     "@hyperledger/aries-askar-shared": "^0.2.3",
-    "@animo-id/expo-secure-environment": "^0.0.1-alpha.0"
+    "@animo-id/expo-secure-environment": "*"
```
TimoGlastra, line 49: this can cause issues, it's not fine to just use any version?

packages/askar/src/wallet/AskarBaseWallet.ts
```diff
@@ -206,7 +206,7 @@ export abstract class AskarBaseWallet implements Wallet {
 
         // Generate a hardware-backed P-256 keypair
         secureEnvironment.generateKeypair(kid)
```
TimoGlastra, line 208: Should also be awaited

TimoGlastra, line 208:
```suggestion
        await secureEnvironment.generateKeypair(kid)
```

PR #2117, opened 2024-11-27
fix: assert valid jarm response

auer-martin:


TimoGlastra:
Maybe we can add a test covering this?

auer-martin:
I added a basic test. It works, but I am not sure if we should include it like this.

packages/openid4vc/tests/openid4vc.e2e.test.ts
```diff
@@ -1857,6 +1858,112 @@ describe('OpenId4Vc', () => {
...
+    requestPayload.response_mode = ResponseMode.DIRECT_POST
+
+    await expect(
+      holder.agent.modules.openId4VcHolder.acceptSiopAuthorizationRequest({
+        authorizationRequest: resolvedAuthorizationRequest.authorizationRequest,
+        presentationExchange: {
+          credentials: selectedCredentials,
+        },
+      })
+    ).rejects.toThrow()
```
TimoGlastra, line 1964: I think this test makes sense and is a good addition 👍 . Is there anything returned in the error indicating we're missing response encryption? Would be nice if this `toThrow` contains something that asserts it's the JARM that is failing here. (it also helps the client interacting with Credo as a verifier to see what's missing)

auer-martin: Nah, there was not. I added a cause property now.

packages/openid4vc/src/shared/router/context.ts
```diff
@@ -83,7 +83,7 @@ export function sendErrorResponse(
   error: unknown,
   additionalPayload?: Record<string, unknown>
 ) {
-  const body = { error: message, ...additionalPayload }
+  const body = { error: message, ...(error instanceof Error && { cause: error.message }), ...additionalPayload }
```
TimoGlastra, line 86: This is not compliant with the spec. there is error and error_description. But we should be cautious with just returning the error message for unknown errors.

For oid4vci i reworked it to now throw a `Oauth2ServerErrorResponseError ` and in the place where `sendErrorResponse` is called i first check for   instanceof Oauth2ServerErrorResponseError and then call `sendOauth2ErrorResponse`. Then we make sure only error messages that should be leaked to the client are leaked. (Because we only construct a `Oauth2ServerErrorResponseError` if needed. Can you udpate?

TimoGlastra (approved):

PR #2124, opened 2024-12-05
fix: Anoncreds Offer Attachment Type

gmulhearn:
issue-credential-v2 anoncreds didcomm attachments are sending with `anoncreds/credential@v1.0` format type rather than `anoncreds/credential-offer@v1.0`

spec:
https://github.com/hyperledger/aries-rfcs/blob/main/features/0771-anoncreds-attachments/README.md#credential-offer-format

### DIFF
--- a/packages/anoncreds/src/formats/AnonCredsCredentialFormatService.ts
+++ b/packages/anoncreds/src/formats/AnonCredsCredentialFormatService.ts
@@ -567,7 +567,7 @@ export class AnonCredsCredentialFormatService implements CredentialFormatService
     // if the proposal has an attachment Id use that, otherwise the generated id of the formats object
     const format = new CredentialFormatSpec({
       attachmentId: attachmentId,
-      format: ANONCREDS_CREDENTIAL,
+      format: ANONCREDS_CREDENTIAL_OFFER,
     })
 
     const offer = await anonCredsIssuerService.createCredentialOffer(agentContext, {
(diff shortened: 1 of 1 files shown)


packages/anoncreds/src/formats/AnonCredsCredentialFormatService.ts
```diff
@@ -567,7 +567,7 @@ export class AnonCredsCredentialFormatService implements CredentialFormatService
     // if the proposal has an attachment Id use that, otherwise the generated id of the formats object
     const format = new CredentialFormatSpec({
       attachmentId: attachmentId,
-      format: ANONCREDS_CREDENTIAL,
+      format: ANONCREDS_CREDENTIAL_OFFER,
```
TimoGlastra, line 570: I wanted to make sure this won't break previous versions, but we basically allow any of the identifiers for all messags, which is a bit weird but in this case helps a lot (but probably also why we didn't dettect this for so long)

TimoGlastra (approved):

packages/anoncreds/src/formats/AnonCredsCredentialFormatService.ts
gmulhearn: yep makes sense. i'm not sure if you guys had been interop testing with acapy, but they also have some minor issues with the attachment naming which i imagine would cause interop issues (see https://github.com/openwallet-foundation/acapy/pull/3374). Hopefully after that we can have some good interop. this is specifically coming up as i've been testing out cheqd anoncreds (being introduced to acapy soon).

PR #2126, opened 2024-12-05
fix: Revocation Index 0 anoncreds-rs support

### LINKED ISSUE #2125: bug: anoncreds-rs issuance of revocation index 0 silently issues non-revokable cred
I was finding that when i issued a revocable cred with index 0, the issued credential would be issued without any attachment to a revocation registry (non-revocable). e.g.:
```ts
  const credEx = await credentials.offerCredential({
    protocolVersion: "v2",
    connectionId: connId,
    credentialFormats: {
      anoncreds: {
        attributes: [
          { name: "name", value: "Jane Doe" },
          { name: "age", value: "23" },
        ],
        revocationRegistryIndex: 0,
        revocationRegistryDefinitionId: revRegDefId,
      },
    },
  });
```

I believe this is because the anoncreds-rs issuer service only assembles `CredentialRevocationConfig` if `revocationRegistryIndex` is truthy - so it skips `CredentialRevocationConfig` construction if `revocationRegistryIndex` is `0`. Problem resolved when i moved to index `1`

https://github.com/openwallet-foundation/credo-ts/blob/9d934afc87eb7dad87fd65681ada4df0333bd439/packages/anoncreds/src/anoncreds-rs/AnonCredsRsIssuerService.ts#L323

gmulhearn:
Closes #2125

### DIFF
--- a/packages/anoncreds/src/anoncreds-rs/AnonCredsRsIssuerService.ts
+++ b/packages/anoncreds/src/anoncreds-rs/AnonCredsRsIssuerService.ts
@@ -78,10 +78,10 @@ export class AnonCredsRsIssuerService implements AnonCredsIssuerService {
 
     let createReturnObj:
       | {
-        credentialDefinition: CredentialDefinition
-        credentialDefinitionPrivate: CredentialDefinitionPrivate
-        keyCorrectnessProof: KeyCorrectnessProof
-      }
+          credentialDefinition: CredentialDefinition
+          credentialDefinitionPrivate: CredentialDefinitionPrivate
+          keyCorrectnessProof: KeyCorrectnessProof
+        }
       | undefined
     try {
       createReturnObj = CredentialDefinition.create({
@@ -114,9 +114,9 @@ export class AnonCredsRsIssuerService implements AnonCredsIssuerService {
 
     let createReturnObj:
       | {
-        revocationRegistryDefinition: RevocationRegistryDefinition
-        revocationRegistryDefinitionPrivate: RevocationRegistryDefinitionPrivate
-      }
+          revocationRegistryDefinition: RevocationRegistryDefinition
+          revocationRegistryDefinitionPrivate: RevocationRegistryDefinitionPrivate
+        }
       | undefined
     try {
       createReturnObj = RevocationRegistryDefinition.create({
--- a/packages/anoncreds/src/anoncreds-rs/AnonCredsRsIssuerService.ts
+++ b/packages/anoncreds/src/anoncreds-rs/AnonCredsRsIssuerService.ts
@@ -78,10 +78,10 @@ export class AnonCredsRsIssuerService implements AnonCredsIssuerService {
 
     let createReturnObj:
       | {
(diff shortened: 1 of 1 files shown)


packages/anoncreds/src/anoncreds-rs/AnonCredsRsIssuerService.ts
```diff
@@ -320,7 +320,7 @@ export class AnonCredsRsIssuerService implements AnonCredsIssuerService {
       }
 
       let revocationConfiguration: CredentialRevocationConfig | undefined
-      if (revocationRegistryDefinitionId && revocationStatusList && revocationRegistryIndex) {
+      if (revocationRegistryDefinitionId && revocationStatusList && revocationRegistryIndex !== undefined) {
```
TimoGlastra, line 323: I just remembered it starts at one by convetion, so it will still throw later on i think. But this at least fixes your issue. 

We might want to add an error saying index 0 is not allowed?

https://github.com/hyperledger/anoncreds-rs/issues/337

TimoGlastra (approved):

packages/anoncreds/src/anoncreds-rs/AnonCredsRsIssuerService.ts
gmulhearn: oh damn! i didn't know this. i guess atleast with this change it will make it's way down into anoncreds-rs, where it will then throw an error - rather than current behaviour where it silently becomes non-revokable.

FYI i just tried running this fix with an index 0 input. i get the error:
```
AnoncredsError: Invalid state: Revocation index is outside of valid range
```

gmulhearn:
I'm getting the same int test failure on my other PR: https://github.com/openwallet-foundation/credo-ts/pull/2124 . 

seems unrelated?

packages/anoncreds/src/anoncreds-rs/AnonCredsRsIssuerService.ts
TimoGlastra: okay then let's merge this, because it improves a lot

TimoGlastra:
> I'm getting the same int test failure on my other PR: #2124 .
> 
> seems unrelated?

YEs we need to fix the CI, if it's only the unrelated issue i can merge it!

packages/anoncreds/src/anoncreds-rs/AnonCredsRsIssuerService.ts
genaris: I'm a bit late into the conversation but yes, this came originally from the fact that an index = 0 is not allowed. 

I think it would be great if you can add an specific check for `revocationRegistryIndex === 0` with an appropriate error message, since this is confusing for those who are not familiar with this convention (which BTW is only shyly mentioned in the [spec](https://hyperledger.github.io/anoncreds-spec/): _"Each credential issued using the Revocation Registry is given its own index (1 to the capacity of the Revocation Registry into the array, the index of the point for that credential."_)

PR #2127, opened 2024-12-05
refactor!: didcomm module

genaris:
Initial work on extracting DIDComm out of core. This PR moves all DIDComm-related modules to a dedicated didcomm package, where they are located alongside a base `DidCommModule` that exposes and API and config to set up basic DIDComm features (e.g. register inbound/outbound transports, endpoints and DIDComm flags).

To not overcomplicate this big refactoring, no further API and usage changes where applied, so most changes to use DIDComm in an Agent are:

1) We need to explicitly create in Agent constructor all modules we use (apart from base DidCommModule): ProofsModule, ConnectionsModule, BasicMessagesModule, etc. The exposed `getDefaultDidcommModules()` can help to facilitate typical Agent creation before doing a further refactoring
2) Move DIDComm-related AgentConfig parameters to the DidCommModuleConfig
3) Update calls to Agent's registerInboundTransport/registerOutboundTransport, endpoints, features and other DIDComm-related properties to `agent.modules.didcomm.xxx`
4) Add `.modules` to all calls to DIDComm related modules (e.g. `agent.proofs.method()` to `agent.modules.proofs.method()`)

And those are pretty much the changes required to use it coming from previous code base.

Some notes I took:

- Since DIDComm is not initialized at BaseAgent.initialize() anymore, its dependant modules cannot access objects like Feature Registry at module `register()`. So protocol and other features are actually registered at each module's initialize()` method. Not a big problem I think but it is a new measure to take into account
- Due to the way it is implemented, `mediationRecipient` needs to be explicitly initialized (i.e. call its initialize() API method). Otherwise, provisioning and pickup will not work
- Tenant records only consider base Agent config. Therefore, it is not possible to specify a particular connection image URL for each tenant in a multi-tenant configuration (or at least that's what I think)
- Some types were duplicated (e.g. EncryptedMessage, PlaintextMessage, DidCommService) to avoid a dependency cycle between core and didcomm
- For the moment, the very minimum setup needs to instantiate at least DIDComm, Connections, OOB and Message Pickup. Dependency on the latter might be easy to remove but we will probably need to embed most functionality from connections and oob in the base module.
- `node` package depends on didcomm, due to the definition of HTTP and WebSocket Inbound transports. We can create a separate package for that, or leave this work for a full refactor of this package (and react-native), since most of them are used on DIDComm and AnonCreds related modules
- I didn't want to make this PR heavier than it is, so tests have been mostly adapted to just work. But we'll need to reorganize them and also simplify them a bit, since there are a good amount of overlap and this makes the test suite to be slower and harder to maintain


TimoGlastra:
@genaris I see tests etc are passing. Do you think we could already merge a (partial) version of this soon, so we can get most of the large file changes merged, and also start work on some other tickets (such as #2145 and #2157).

genaris:
> @genaris I see tests etc are passing. Do you think we could already merge a (partial) version of this soon, so we can get most of the large file changes merged, and also start work on some other tickets (such as #2145 and #2157).

I've updated the description and marked this PR as ready for review. Both tests and demos are working, so I think it is in a status where it can be safe to merge. There is a good number of things to do yet before the official 0.6.0 release but I think it is a good start.

.changeset/purple-donkeys-camp.md
```diff
@@ -0,0 +1,67 @@
...
+'@credo-ts/didcomm': minor
+'@credo-ts/tenants': minor
+'@credo-ts/askar': minor
+'@credo-ts/cheqd': minor
+'@credo-ts/core': minor
+'@credo-ts/drpc': minor
+'@credo-ts/node': minor
+---
+
+DIDComm has been extracted out of the Core. This means that now all DIDComm related modules (e.g. proofs, credentials) must be explicitly added when creating an `Agent` instance. Therefore, their API will be accesable under `agent.modules.[moduleAPI]` instead of `agent.[moduleAPI]`. Some `Agent` DIDComm-related properties and methods where also moved to the API of a new DIDComm module (e.g. `agent.registerInboundTransport` turned into `agent.modules.didcomm.registerInboundTransport`).
```
TimoGlastra, line 16: This is not for this PR, but I think before 0.6 we should look at making:
- `agent.didcomm` possible (nesting is getting quite large)
- nest didcomm modules under `agent.didcomm`, also because `agent.credentials` / `agent.modules.credentials` is misleading, since it's didcomm specific. I think registerin
- Adopting the conenctions / oob modules into the didcomm module by default, and you can configure it in the didcomm module config again. I think this simplifies setup.
- Registering custom didcomm protocols on the didcomm module rather than the agent? This may solve some of the initialization and feature registry issues. We can create a `DidcommModule` interface that has special feature registry etc.

packages/core/src/modules/dids/__tests__/DidsApi.test.ts
```diff
@@ -13,7 +13,7 @@ import {
 
 const agentOptions = getInMemoryAgentOptions('DidsApi')
 
-const agent = new Agent(agentOptions)
+const agent: Agent = new Agent(agentOptions)
```
TimoGlastra, line 16:
```suggestion
const agent = new Agent(agentOptions)
```

packages/core/src/modules/dids/methods/peer/createPeerDidDocumentFromServices.ts
```diff
@@ -1,4 +1,4 @@
-import type { ResolvedDidCommService } from '../../../didcomm'
+import type { ResolvedDidCommService } from '../../../../types'
```
TimoGlastra, line 1: I think it makes sense to keep this type in Core, and import it from the didcomm module (does not have to be done in this PR).

Because we support DidComm service for Did document, and thus I think it's ok to make this part of core

tests/transport/SubjectInboundTransport.ts
```diff
@@ -19,18 +18,19 @@ export class SubjectInboundTransport implements InboundTransport {
     this.ourSubject = ourSubject
   }
 
-  public async start(agent: Agent) {
+  public async start(agent: AgentContext) {
```
TimoGlastra, line 21: Is it not possible to provide the agent here anymore?

packages/tenants/src/services/__tests__/TenantService.test.ts
```diff
@@ -28,18 +28,20 @@ describe('TenantRecordService', () => {
     jest.clearAllMocks()
   })
 
+  // FIXME: connectionImageUrl is now part of DIDComm module. Tenants records do not currently
```
TimoGlastra, line 31: For the new crypto api i've also been looking to solve this problem, and how i tried to solve it is that each module can hook into the the storing process of the tenant config. It's a bit tricky though, as it makes all modules somewhat intertwined with the tenants module ...

packages/core/src/storage/migration/updates/0.3-0.3.1/index.ts
```diff
@@ -1,7 +1,10 @@
 import type { BaseAgent } from '../../../../agent/BaseAgent'
+import type { UpdateConfig } from '../../updates'
 
 import { migrateDidRecordToV0_3_1 } from './did'
 
-export async function updateV0_3ToV0_3_1<Agent extends BaseAgent>(agent: Agent): Promise<void> {
+// FIXME: optional update config parameter
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export async function updateV0_3ToV0_3_1<Agent extends BaseAgent>(agent: Agent, _config: UpdateConfig): Promise<void> {
```
TimoGlastra, line 8: what is this fixme? I'm not sure i understand

packages/core/src/constants.ts
```diff
@@ -1,5 +1,5 @@
 export const InjectionSymbols = {
-  MessagePickupRepository: Symbol('MessagePickupRepository'),
+  MessagePickupRepository: Symbol('MessagePickupRepository'), // TODO: Move to DIDComm
```
TimoGlastra, line 2: I think we should get rid of the injectionsymbols, and move it to the corresponding packages/files.

In most cases it's not needed to use the specific injection token publicy (as you can get it from the agent config / agent context directly)

TimoGlastra (approved):

TimoGlastra:
Left some comments, but all are fine to fix later. Very nice work!! 🙌 

Maybe we can open an issue and add your + mine comments so we can keep track of todos?

.changeset/purple-donkeys-camp.md
genaris: Yeah good ideas there! Initially I made a "DidCommModule" but I decided to leave it for a further PR to not overcomplicate it. But for sure it is something that makes sense, especially to have a tidy Agent API.

Same thing about the transport config, which is something not so evident for developers.

tests/transport/SubjectInboundTransport.ts
genaris: The problem is that we start transports at DIDComm module's initialize(), which receives the AgentContext as an argument. Before this PR we started transport in `Agent` so we were able to call `transport.start(this)`.

Do you see a potential issue on this change, besides the fact that it breaks the API?

packages/core/src/storage/migration/updates/0.3-0.3.1/index.ts
genaris: Me neither... that's the problem of make big refactors in the spare time :stuck_out_tongue_closed_eyes:

packages/core/src/constants.ts
genaris: I agree.. InjectionSymbols are somewhat a pattern that we used at the beginning of Credo development but now they can be replaced (at least most of them).

In this particular case probably we can work with events, that will be catched by the Message Pickup module in order to decouple both. Surely I'll think about it in a further PR in the incoming days.

packages/tenants/src/services/__tests__/TenantService.test.ts
genaris: Well, maybe tenants could become a core feature? :thinking:

Issue #2135, opened 2024-12-18
Unable to reinitialize agent once shut downed


GHkrishna:
## What:
Reinitializing agent with `await agent.initialize()` after `await agent.shutdown()` gives an indyVdr error due to this implementation. Should we just return instead of throwing error:
```typescript
public connect() {
    if (this._pool) {
      throw new IndyVdrError('Cannot connect to pool, already connected.')
    }

    this._pool = new PoolCreate({
      parameters: {
        transactions: this.config.genesisTransactions,
      },
    })
  }
```

This becomes,
```typescript
public connect() {
    if (this._pool) {
      return
    }

    this._pool = new PoolCreate({
      parameters: {
        transactions: this.config.genesisTransactions,
      },
    })
  }
```
## To reproduce
1. await agent.shutdown()
2. await agent.initialize()

## Version 
"@credo-ts/indy-vdr": "0.5.13"

TimoGlastra:
I think the better approach would be to shutdown the pool at agent shutdown and remove the value initialization

GHkrishna:
The issue for shutting down of pool still exists in IndyVDR. For now I think we can remove the initialized value and set it to null. And once it is resolved in VDR we can retro fit it. What do you think?

TimoGlastra:
Yes i think that makes sense. But since we can't close it we should reuse the existing pool on initialize again

GHkrishna:
Closing as completed by https://github.com/openwallet-foundation/credo-ts/pull/2137

PR #2137, opened 2024-12-27
chore: fixes restart without indy pool disconnect

GHkrishna:
## What
Fixes: Unable to reinitialize agent once shut down #2135

### DIFF
--- a/packages/indy-vdr/src/pool/IndyVdrPool.ts
+++ b/packages/indy-vdr/src/pool/IndyVdrPool.ts
@@ -41,7 +41,6 @@ export interface IndyVdrPoolConfig {
 
 export class IndyVdrPool {
   private _pool?: indyVdrPool
-  private _archievedPool?: indyVdrPool
   private poolConfig: IndyVdrPoolConfig
   public authorAgreement?: AuthorAgreement | null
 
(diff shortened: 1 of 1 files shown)


sairanjit:
@GHkrishna When is the pool `close` method called?

GHkrishna:
> @GHkrishna When is the pool `close` method called?

Tha pool close from the IndyVDR is not initiated for now as it seems to be fixed first. We are just resting the pool variable, as if the pool was never initialised. But since we know we did, we'll reuse the same pool.

sairanjit:
Yes @GHkrishna We have to close the pool. We can do that by subscribing to the `stop$` subject.

packages/indy-vdr/src/pool/IndyVdrPool.ts
```diff
@@ -61,6 +62,12 @@ export class IndyVdrPool {
       throw new IndyVdrError('Cannot connect to pool, already connected.')
```
TimoGlastra, line 62: what if we just return if the pool is already defined? Then we don't have to store the archived pool, and we don't set it to undefined on close (until we fix indy-vdr)

GHkrishna: True, that's probably the easiest way around, thanks :sweat_smile:

TimoGlastra (approved):

GHkrishna:
@TimoGlastra need your help with the failing E2E Tests. Ig our change shouldn't affect it.

PR #2139, opened 2025-01-06
fix(cheqd): cheqd revocationRegistryDefinition resource name

DaevMithran:
This PR fixes resource name collisions for revocation registries with different tags.

Creating two revocation registries with same name's will lead to updating the resource. Adding credential definition tag in the resource name will fix this issue


packages/cheqd/src/anoncreds/services/CheqdAnonCredsRegistry.ts
```diff
@@ -308,8 +312,8 @@ export class CheqdAnonCredsRegistry implements AnonCredsRegistry {
 
       const revocationRegistryDefinitionResource = {
         id: utils.uuid(),
-        name: credentialDefinitionName as string,
-        resourceType: 'anonCredsRevocRegDef',
+        name: `${credentialDefinitionName}-${revocationRegistryDefinition.tag}`,
```
TimoGlastra, line 315: we had issues with long names in the past, for which I think we came up with the solution to hash. Also, does this break already existing revocation registry definitions? Can we resolve both previously created ones, and the new one? And should the spec be updated?

packages/cheqd/package.json
```diff
@@ -26,7 +26,7 @@
     "test": "jest"
   },
   "dependencies": {
-    "@cheqd/sdk": "^2.5.1",
+    "@cheqd/sdk": "^2.6.0",
```
TimoGlastra, line 29: We should properly test this in RN. We have to pin everything to a very specific version or the SDK breaks. Even with the current version it throws an unhandled promise rejection on startup (due to wasm errors) but it somehow still works for resolving.

We haven't had the time to properly look into it, but before we can merge this we should first test it in react native (actually we should pin the version to 2.5.1 for now, instead of using ^2.5.1 until it's fixed)

TimoGlastra (commented):
Thanks. Some notes

packages/cheqd/src/anoncreds/services/CheqdAnonCredsRegistry.ts
DaevMithran: Right, as the credential definition name is hashed already in credo will this still be a problem?

No resolution is backward compatible. we are resolving using the resource id -> get the resource name from metadata to resolve the status lists.

packages/cheqd/package.json
DaevMithran: Reverting it! It would be very helpful if you can create an issue on our [sdk](https://github.com/cheqd/sdk) with context-rich logs or reproduction steps. Thanks

packages/cheqd/src/anoncreds/services/CheqdAnonCredsRegistry.ts
TimoGlastra: Okay thanks for the answers! I'm not sure, what is the max length of the tag? I think the credential definition name is max 32 because it's hashed. With the dash, the tag can max be 31 characters then. I think it should suffice, but if we really want it to be error prone we should maybe re-hash?

DaevMithran: Both the approaches sound good to me. Let me know if I should change it to the hashed one

TimoGlastra: Yes let's do the hash. Or the cheqd spec should mention it can max be 31 characters long, but I think that's more prone to errors down the line. So hashing seems most stable to me

packages/cheqd/package.json
DaevMithran: @TimoGlastra after our mainnet upgrade we got some feedback that Cheqd transactions from credo are failing. This must be due to insufficient gas after introducing our fee market module, In our latest sdk we bumped the default gas fees, I think we should bump it. We have not made any major dependency changes from 2.5.1 to 2.6.0

packages/cheqd/src/anoncreds/services/CheqdAnonCredsRegistry.ts
DaevMithran: This is complete

TimoGlastra:
Sorry about the delay for this. It seems CI is now failing due to support missing for BigInt:

> Error: ../node_modules/.pnpm/@cheqd+ts-proto@2.4.0/node_modules/@cheqd/ts-proto/google/protobuf/timestamp.ts(176,29): error TS2737: BigInt literals are not available when targeting lower than ES2020.

I've check and BigInt is available in Node.JS and React Native now so I think we can update our target to ES2020 so resolve this.

TimoGlastra:
I've created this PR to update the target: https://github.com/openwallet-foundation/credo-ts/pull/2156. Once that is merged we can merge this PR. 

I will also try to get a PR pushed with support for the latest version of the Cheqd SDK, as well as the related cosmo dependencies. I'll then create an example app with the issues we encounter with cheqd in react native

TimoGlastra:
Found a first issue with updating the Cheqd SDK: https://github.com/cheqd/sdk/issues/400

DaevMithran:
> Found a first issue with updating the Cheqd SDK: [cheqd/sdk#400](https://github.com/cheqd/sdk/issues/400)

@TimoGlastra for cjs in cheqd. we have fixed them at `version: 2.x` or `tag: cjs` in sdk and ts-proto and we are maintaining it actively. The latest versions only support esm (some packages may not be compatible in react native like cosmjs).

TimoGlastra:
The latest version (5.x) also published a CJS version still, however it doesn't work correctly as ts-proto does not publish CJS on the latest version. So i think it makes sense to remove the CJS build from the newer (5.x) releases.

TimoGlastra:
https://github.com/openwallet-foundation/credo-ts/pull/2156 has been merged, can you update this branch with main?

TimoGlastra:
sorry one more thing 😬 , Can you accept this PR (i can't make updates to your branch if you create the PR from an org account)

https://github.com/cheqd/credo-ts/pull/8

TimoGlastra:
@DaevMithran it seems a test related to cheqd failed

DaevMithran:
> @DaevMithran it seems a test related to cheqd failed

@TimoGlastra Can we rereun it. 

The error was `A jest worker process (pid=7356) crashed for an unknown reason: exitCode=0` only in **E2E(20)**

TimoGlastra:
Thanks so much for this @DaevMithran 🙌

PR #2156, opened 2025-01-27
build: update target for tsc compiler to ES2020

TimoGlastra:
Generally this should not have an impact for the supported environments (Node.JS / React Native). However this will have to be tested in React Native


genaris (approved):
LGTM. I think I can test it on React Native 0.75 this week just to make sure everything is OK.

PR #2161, opened 2025-01-28
refactor!: migrate from .afj to .credo

### LINKED ISSUE #1715: Migrate from `.afj` data folder to `.credo`
For now we kept `.afj` to not cause issues and unexpected breakages, but would be nice to use `.credo`, at least for new wallets

sairanjit:
### Summary

- Fixes #1715 

Questions
- Do we need to delete the `.afj` folder or old wallet after migrating to `.credo` folder ?


sairanjit:
cc: @TimoGlastra @genaris

sairanjit:
I have tested on node and Android 

Screenshot 2025-01-29 at 5 24 47 PM

TimoGlastra:
Thanks for this PR @sairanjit, I think it would be good to wait for a bit with releasing this, as it makes testing with 0.6 unstable versions easier (due to that a migration script can only be run once per version). We should probably look into multi-updates at some point

packages/node/src/NodeFileSystem.ts
```diff
@@ -19,16 +19,61 @@ export class NodeFileSystem implements FileSystem {
...
+    this.dataPath = options?.baseDataPath ? `${options?.baseDataPath}/.credo` : `${homedir()}/.credo/data`
+    this.cachePath = options?.baseCachePath ? `${options?.baseCachePath}/.credo` : `${homedir()}/.credo/cache`
+    this.tempPath = `${options?.baseTempPath ?? tmpdir()}/.credo`
+  }
+
+  /**
+   * Migrate data from .afj to .credo if .afj exists.
+   * Copy the contents from old directory (.afj) to new directory (.credo).
+   */
+  public async migrateWalletToCredoFolder(walletId: string) {
```
TimoGlastra, line 38: I think the file system should stay very generic. So instead of adding this method `migrateWalletToCredoFolder` in the file system (which then needs to be implemented by every file system), I'd move it to the migration method. This can then call the needed methods `exists` and `copyDirectory` (i think this method can also copy files now? Should we create one generic `copy` method that can do both directories (using `recursive` property) and files?)

packages/node/src/NodeFileSystem.ts
```diff
@@ -19,16 +19,61 @@ export class NodeFileSystem implements FileSystem {
...
+          to: this.tempPath,
+        },
+      ]
+
+      for await (const path of pathsToMigrate) {
+        // Migrate if the old paths exist
+        if (await this.exists(path.from)) {
+          await this.copyDirectory(path.from, path.to)
+        }
+      }
```
TimoGlastra, line 67: Should we delete the old files afterwards? Or how do we clean up?

packages/react-native/src/ReactNativeFileSystem.ts
```diff
@@ -13,26 +13,88 @@ export class ReactNativeFileSystem implements FileSystem {
...
-      : `${RNFS.TemporaryDirectoryPath}/.afj${Platform.OS === 'android' ? '/temp' : ''}`
+      ? `${options?.baseTempPath}/.credo`
+      : `${RNFS.TemporaryDirectoryPath}/.credo${Platform.OS === 'android' ? '/temp' : ''}`
+  }
+
+  /**
+   * Migrate data from .afj to .credo if .afj exists.
+   * Copy the contents from old directory (.afj) to new directory (.credo).
+   */
+  public async migrateWalletToCredoFolder() {
```
TimoGlastra, line 42: same here, this can be moved to migration logic so we don't have to duplicate

packages/core/src/storage/migration/updates/0.5-0.6/migrateToCredoFolder.ts
```diff
@@ -0,0 +1,36 @@
...
+import type { FileSystem } from '../../../FileSystem'
+
+import { InjectionSymbols } from '../../../../constants'
+import { CredoError } from '../../../../error'
+
+/**
+ * Migrates the sqlite folder location from .afj to .credo in the storage directory in node and react native.
+ *
+ */
+export async function migrateToCredoFolder<Agent extends BaseAgent>(agent: Agent) {
```
TimoGlastra, line 11: due to the sensivity in doing migrations it's always good to test these migrations extensively. For all other migrations and version changes we have tests (per migration method, and for a version migration in general), can you add these?

TimoGlastra (changes requested):
One thing I'm not sure about is how this works:
- We initialize the agent and open the wallet (this will use the new paths, so a new wallet will be created)
- The wallet will be initialized with storage version 0.6
- No updates need to happen because we're on the new path
- we now have a new empty wallet

Or am i missing something here?

sairanjit:
Based on the discussion in the Credo Contributors meeting on 30th January 2025 we can keep `.afj` folder due to the complexity involved in the #1715  and close this PR

PR #2166, opened 2025-01-30
feat!: Use did:peer:4 by default for DID Exchange

genaris:
Up to now, we were using did:peer:1 by default when creating DID Exchange requests. This PR changes this to use the newer did:peer:4, unless the `peerNumAlgoForDidExchangeRequests` in DIDComm connection module config is explicitly set to another one.

Note that this PR also applies some fixes found during this update:

- In DID Exchange Response creation, ensure that invitation key is used to sign DID rotation attachment, since in some cases, such as using public DIDs or implicit invitations, we were using the keys present in the services from rotated DID (or the ones from `routing` parameter). I applied this to `did_doc~attach` as well, although in the [RFC](https://github.com/hyperledger/aries-rfcs/blob/main/features/0023-did-exchange/README.md) it does not explicitly mention that it should use the invitation keys. @TimoGlastra please let me know what's your interpretation for this.
- When receiving an implicit invitation, we were setting `theirLabel` for the new connection with the value coming from `label`, which is supposed to be the opposite (our label instead). So now we put `alias` as `theirLabel` , as it usually happens to connections created with explicit invitations


packages/core/tests/oob.test.ts
```diff
@@ -767,7 +767,7 @@ describe('out of band', () => {
 
       // Use the invitation did from the first connection to create the second connection
       const outOfBandRecord2 = await faberAgent.modules.oob.createInvitation({
-        invitationDid: outOfBandRecord1.outOfBandInvitation.invitationDids[0],
+        invitationDid: faberAliceConnection.did,
```
TimoGlastra, line 770: This is different right? InvitationDid is usually not the final did

packages/didcomm/src/modules/connections/DidExchangeProtocol.ts
```diff
@@ -284,32 +286,33 @@ export class DidExchangeProtocol {
...
+    // DID Rotate attachment should be signed with invitation keys
+    const invitationRecipientKeys = outOfBandRecord.outOfBandInvitation
+      .getInlineServices()
+      .map((s) => s.recipientKeys)
+      .reduce((acc, curr) => acc.concat(curr), [])
+
+    // In case the DID Exchange started with an implicit invitation, take all recipient keys from the corresponding DID
+    for (const did of outOfBandRecord.outOfBandInvitation.getDidServices()) {
+      invitationRecipientKeys.push(
+        ...(await getDidDocumentForCreatedDid(agentContext, parseDid(did).did)).recipientKeys.map(
```
TimoGlastra, line 298: Do we always need to do this, or only for implicit? Because now we run this always?

packages/didcomm/src/modules/connections/services/helpers.ts
```diff
@@ -132,12 +132,14 @@ export function routingToServices(routing: Routing): ResolvedDidCommService[] {
...
-
-  if (!didRecord?.didDocument) {
-    throw new CredoError(`Could not get DidDocument for created did ${did}`)
+  if (!didRecord) {
+    throw new CredoError(`Could not find created did ${did}`)
   }
-  return didRecord.didDocument
+
+  const didsApi = agentContext.dependencyManager.resolve(DidsApi)
+  return await didsApi.resolveDidDocument(did)
```
TimoGlastra, line 142: Hmm 🤔 

Are there cases where we don't store our own did?

TimoGlastra (approved):
Can you add a changeset for each of the changes in this pr?

packages/core/tests/oob.test.ts
genaris: In this case it should be the same, since we are not setting it as an multiUse invitation. I did this trick because `OutOfBandInvitation.invitationdDids` is dynamically generating peer:2 DIDs and this does not work directly when querying peer:4 DIDs.

But I'll take a second look to it to see if everything is in place.

packages/didcomm/src/modules/connections/DidExchangeProtocol.ts
genaris: I think the comment is misleading. It is useful for resolvable DIDs in general.

packages/didcomm/src/modules/connections/services/helpers.ts
genaris: Not really, but I wanted to take all the advantages of `DidsApi.resolveDidDocument()` (such as dynamically generating the did:peer:4 DID Document) making sure that it would only resolve DIDs created by us (I can propagate a filtering parameter but it would modify a good number of methods).

TimoGlastra: This is okay! Just curious :)

TimoGlastra: It also has the benefit of caching, and even DIDs created by us, but not managed by us (e.g. think a did web for which we have one key, but it's stored outside of credo as it's regularly updated)

PR #2167, opened 2025-01-30
fix: exclude test types from core

genaris:
In Core we are including test types in the source code, and since now we have dependency on DIDComm module it was messing up the core package build process (it creates 2 directories, one with core and another with didcomm).

Thanks @sairanjit for noticing and showing the solution for this!

### DIFF
--- a/packages/core/tsconfig.build.json
+++ b/packages/core/tsconfig.build.json
@@ -5,5 +5,5 @@
     "outDir": "./build"
   },
 
-  "include": ["src/**/*", "types"]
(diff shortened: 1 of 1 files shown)


sairanjit (approved):

TimoGlastra (approved):
> and since now we have dependency on DIDComm module

Does core depend on didcomm?

genaris:
> > and since now we have dependency on DIDComm module
> 
> Does core depend on didcomm?

By mistake, due to this jest types file that I'm excluding here.

PR #2168, opened 2025-01-31
feat: remove MessagePickupRepository injection symbol

genaris:
This PR removes MessagePickupRepository injection symbol, which makes DIDComm base module to not depend on Message Pickup. Instead, `MessageSender` emits an internal event when a message is meant to be queued for pickup (e.g. when the recipient does have a transport queue endpoint). Such event is catched by `MessagePickupModule` and effectively added to the queue there.

The main advantage for now (apart from removing an InjectionSymbol) is that we can define a very basic DIDComm Agent that only includes the very basic DIDComm modules (didcomm, oob and connections).

I want to take this as an initial step to refactor the relationship between modules within DIDComm. For instance, I think it would be nice to put in MessagePickupModule the logic for pickup clients that currently resides on `MediationRecipientApi`, and maybe we can even reduce the amount of modules by having a single mediation module that will be used for the Mediation Coordination protocol, leaving the base mediation features (records, Forward message handling) to the base DIDComm module, along with connections/oob.

### DIFF
--- a/packages/didcomm/src/MessageSender.ts
+++ b/packages/didcomm/src/MessageSender.ts
@@ -1,6 +1,6 @@
 import type { AgentMessage } from './AgentMessage'
 import type { EnvelopeKeys } from './EnvelopeService'
-import type { AgentMessageSentEvent } from './Events'
+import type { AgentMessageSentEvent, QueuePackedMessageForPickupEvent } from './Events'
 import type { TransportSession } from './TransportService'
 import type { ConnectionRecord } from './modules/connections/repository'
 import type { OutOfBandRecord } from './modules/oob/repository'
@@ -33,7 +33,6 @@ import { DID_COMM_TRANSPORT_QUEUE } from './constants'
 import { ReturnRouteTypes } from './decorators/transport/TransportDecorator'
 import { MessageSendingError } from './errors'
 import { OutboundMessageContext, OutboundMessageSendStatus } from './models'
-import { MessagePickupRepository } from './modules/message-pickup/storage'
 import { DidCommDocumentService } from './services/DidCommDocumentService'
 
 export interface TransportPriorityOptions {
@@ -45,7 +44,6 @@ export interface TransportPriorityOptions {
 export class MessageSender {
   private envelopeService: EnvelopeService
   private transportService: TransportService
-  private messagePickupRepository: MessagePickupRepository
   private logger: Logger
   private didResolverService: DidResolverService
   private didCommDocumentService: DidCommDocumentService
@@ -55,7 +53,6 @@ export class MessageSender {
   public constructor(
     envelopeService: EnvelopeService,
     transportService: TransportService,
-    @inject(InjectionSymbols.MessagePickupRepository) messagePickupRepository: MessagePickupRepository,
     @inject(InjectionSymbols.Logger) logger: Logger,
     didResolverService: DidResolverService,
     didCommDocumentService: DidCommDocumentService,
@@ -63,7 +60,6 @@ export class MessageSender {
   ) {
     this.envelopeService = envelopeService
     this.transportService = transportService
-    this.messagePickupRepository = messagePickupRepository
     this.logger = logger
     this.didResolverService = didResolverService
     this.didCommDocumentService = didCommDocumentService
@@ -185,7 +181,7 @@ export class MessageSender {
     // If the other party shared a queue service endpoint in their did doc we queue the message
     if (queueService) {
       this.logger.debug(`Queue packed message for connection ${connection.id} (${connection.theirLabel})`)
-      await this.messagePickupRepository.addMessage({
+      this.emitQueuePackedMessageForPickupEvent(agentContext, {
         connectionId: connection.id,
         recipientDids: [verkeyToDidKey(recipientKey)],
         payload: encryptedMessage,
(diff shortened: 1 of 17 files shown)


packages/didcomm/src/MessageSender.ts
```diff
@@ -341,7 +337,10 @@ export class MessageSender {
       }
 
       const encryptedMessage = await this.envelopeService.packMessage(agentContext, message, keys)
-      await this.messagePickupRepository.addMessage({
+
+      // This is an internal event that will be catched by Message Pickup module and queued, in case the
+      // Agent supports it
+      this.emitQueuePackedMessageForPickupEvent(agentContext, {
```
TimoGlastra, line 343: I generally like this new approach, but one downside is that if no message pickup module is registered the queued messages will be lost. I think if we receive a message that can't be delivered we should probably handle that in some way(maybe we can check if there's any subscribers?)

TimoGlastra: Also, one more consideration is that we've noticed that in serverless architectures asynchronous events don't really work well. You don't know when an incoming request is done processing. 

That's why I wanted to look into reducing the number of events emitted, but I wasn't sure yet what a good approach would be. I think callbacks make sense, but is also what we had previously (we could of course have no handler by default, and the message pickup module can register it, but then we're basically back at the injection in a different way).

Thoughts?

packages/didcomm/src/modules/message-pickup/MessagePickupModuleConfig.ts
```diff
@@ -36,8 +38,12 @@ export interface MessagePickupModuleConfigOptions<MessagePickupProtocols extends
 export class MessagePickupModuleConfig<MessagePickupProtocols extends MessagePickupProtocol[]> {
   private options: MessagePickupModuleConfigOptions<MessagePickupProtocols>
 
+  #messagePickupRepository: MessagePickupRepository
+
   public constructor(options: MessagePickupModuleConfigOptions<MessagePickupProtocols>) {
     this.options = options
+    // Message Pickup queue: use provided one or in-memory one if no injection symbol is yet defined
+    this.#messagePickupRepository = options.messagePickupRepository ?? new InMemoryMessagePickupRepository()
```
TimoGlastra, line 46: Should we change this to something like messageQueueRepository? or queuedMessageRepository?

packages/didcomm/src/modules/message-pickup/storage/InMemoryMessagePickupRepository.ts
```diff
@@ -17,10 +17,10 @@ interface InMemoryQueuedMessage extends QueuedMessage {
 
 @injectable()
 export class InMemoryMessagePickupRepository implements MessagePickupRepository {
-  private logger: Logger
+  private logger?: Logger
   private messages: InMemoryQueuedMessage[]
 
-  public constructor(@inject(InjectionSymbols.Logger) logger: Logger) {
+  public constructor(@inject(InjectionSymbols.Logger) logger?: Logger) {
```
TimoGlastra, line 23: Why optional?

TimoGlastra: If we don't have access to this, we should just pass the agent context to all the methods in this class I think

TimoGlastra (approved):
Nice! only one bigger concern re the events and serverless architectures. 

Otherwise, would be really interested to also think more about the multi-instance session management. We're now discussing it in mediator repo (not sure what came out of the WG call yesterday?), but it's probably relevant to solve in Credo core (at least provide the needed interfaces).

TimoGlastra:
Can you add changeset(s) for the changes?

packages/didcomm/src/modules/message-pickup/storage/InMemoryMessagePickupRepository.ts
genaris: Yeah, I didn't want to change all `MessagePickupRepository` interface methods only for this basic implementation. But I think resistance is futile: a good refactoring of all of this will be needed before 0.6.0 release!

packages/didcomm/src/modules/message-pickup/MessagePickupModuleConfig.ts
genaris: This subtle question made me think that the approach we are following is not correct: this message queue is part of DIDComm core, regardless of the message pickup mechanism we use (for instance, the infamous "implicit" mode present in some mediators do not use Message Pickup protocol at all, but they do access this element to send queued messages to their clients at the moment they connect to a WebSocket).

In such case, this [QueueTransportMessageRepository](https://github.com/decentralized-identity/didcomm-messaging/blob/main/extensions/return_route/main.md#queue-transport) must be defined in the base DIDComm module configuration, and therefore much of the things done in this PR would not make sense.

packages/didcomm/src/MessageSender.ts
genaris: Say you don't like this approach without saying you don't like this approach! :laughing: 

Seriously speaking, I think we can solve this problem in different ways, and surely we can avoid using async events for that. If we leave this "Queue Transport repository" as a core feature of DIDComm, we can queue messages synchronously. Also, if we consider mediation (the concept itself, not the coordination protocols) as a core feature, we can also setup there different queue transport policies, like the one you mentioned some time ago about only queuing messages for agents we are a mediator of.

genaris:
For the moment I'll close this PR in favour of another one where we can refactor DIDComm modules in such a way that core features and concepts are part of the base module, and we leave protocol-specific parts to the corresponding modules.

packages/didcomm/src/MessageSender.ts
TimoGlastra: > Say you don't like this approach without saying you don't like this approach! 😆

I really like getting rid of the injection symbols :) 

But i've also had some bad experience trying to make credo work in serverless environments and it sadly just didn't work :(

packages/didcomm/src/modules/message-pickup/MessagePickupModuleConfig.ts
TimoGlastra: I agree. Routing/forwarding/queuing is part of core. Mediator management and pickup not! :)

Issue #2145, opened 2025-01-16
Switch from prettier + eslint to Biome


TimoGlastra:
Example of setup can be found here: https://github.com/openwallet-foundation-labs/oid4vc-ts/blob/main/biome.json

https://github.com/openwallet-foundation-labs/oid4vc-ts/blob/main/package.json#L14-L15

Things to consider:
- migration of all ESLint rules to Biome. Some may not be needed, but e.g. no restricted globals would help: https://github.com/openwallet-foundation/credo-ts/blob/main/.eslintrc.js#L78

Wait for all big PRs that are open to be merged?
- [ ] https://github.com/openwallet-foundation/credo-ts/pull/2127

genaris:
Everything that helps to improve CI performance will be well received, and certainly eslint is getting very, very slow since some time. I don't have experience with Biome but I think we could give a try.

berendsliedrecht:
Basically all projects I start now use Biome and i strongly prefer it over prettier+eslint. It does not require 10 plugins for everything to work, it is insanely fast and way easier to configure.

I think we can introduce it with just the config some rules off. And slowly in following PRs turn each rule on with the fixes. Might be a nicer review process because of that.

TimoGlastra:
@sairanjit the big didcomm PR got merged, so feel free to take a stab at the biome setup!

sairanjit:
Sure @TimoGlastra I will pick this soon.

TimoGlastra:
This is done 🙌 (thanks again @sairanjit )

PR #2175, opened 2025-02-04
chore: add biomejs

sairanjit:
### Summary

- Use biomejs for Linting and formatting #2145 

Tasks
- [ ] Fix formatting
- [ ] Fix Linting
- [x] Migrate initial required rules to biomejs
- [ ] Update `eslint-disable-next-line` to `biome-ignore`
- [x] Update the scripts in package.json
- [x] Update commands in CI/CD


sairanjit:
cc: @TimoGlastra @berendsliedrecht This is a starting point for migrating to `biomejs`. Do you have any thoughts on additional points I should consider beyond what I’ve already mentioned?

TimoGlastra:
I think the list looks good! Could you create a list of which eslint rules you haven't added biome rules for yet?

I think for me I'd like to have support for:
- import/no-cycle
- @typescript-eslint/consistent-type-imports
- @typescript-eslint/no-non-null-assertion
- import/no-extraneous-dependencies
- no-restricted-imports

sairanjit:
@TimoGlastra These are the rules I have configured as of now.

- **Rules configured**
	✅ @typescript-eslint/no-unsafe-declaration-merging
	✅ @typescript-eslint/no-use-before-define
	✅ no-console
	✅ @typescript-eslint/consistent-type-imports
	✅ @typescript-eslint/no-non-null-assertion
	✅ import/no-extraneous-dependencies
	✅ no-restricted-imports
	✅ no-restricted-globals

- **Rules Not available**
	❌ @typescript-eslint/explicit-function-return-type
	❌ @typescript-eslint/explicit-module-boundary-types
	❌ @typescript-eslint/explicit-member-accessibility
	❌ @typescript-eslint/ban-ts-comment
	❌ @typescript-eslint/no-floating-promises
	❌ import/no-cycle
	❌ import/newline-after-import
	❌ import/order
	❌ no-restricted-syntax
	
Once you confirm it I will run `pnpm style:fix` and update this PR

TimoGlastra:
> import/order

This has a generic rule, which is fine. Is that enabled? 

https://biomejs.dev/analyzer/import-sorting/

biome.json
```diff
@@ -0,0 +1,130 @@
...
+  "overrides": [
+    {
+      "include": ["packages/core/**"],
+      "linter": {
+        "rules": {
+          "style": {
+            "noRestrictedGlobals": {
+              "level": "error",
+              "options": {
+                "deniedGlobals": ["Buffer"]
```
TimoGlastra, line 79: I think we can extend this to all packages. But i see it was not like that in eslint before. So probably good to do it in a separate PR to avoid too many changes

biome.json
```diff
@@ -0,0 +1,130 @@
...
+        "demo/**",
+        "demo-openid/**",
+        "scripts/**",
+        "**/tests/**",
+        "tests/**"
+      ],
+      "linter": {
+        "rules": {
+          "correctness": {
+            "noUndeclaredDependencies": "off"
```
TimoGlastra, line 121: Can we enable this for all files with dev dependencies set to false, and then here with dev dependencies true?

https://biomejs.dev/linter/rules/no-undeclared-dependencies/#options

sairanjit: This options are available from biome version `2.0.0`

biome.json
sairanjit: Sure

biome.json
sairanjit: Also there is a bug now for `noUndeclaredDependencies` https://github.com/biomejs/biome/issues/4992 which will be fixed in `2.0.0`

sairanjit:
I broke the rebase and merge. I will create a new PR

biome.json
TimoGlastra: Ah would be good to use the latest version immediately? Otherwise we have to update again after this pr

sairanjit: I think we should wait till 2.0 release https://github.com/biomejs/biome/milestone/6 and comeback again to this

TimoGlastra: Ah didn't see it was not released yet. Makes sense 👍

PR #2186, opened 2025-02-11
feat: add did:webvh as a DID resolver

brianorwhatever:
I'll admit I don't know how to use `pnpm` so I worry the `package.json` changes aren't required it seems to have sorted and upgraded some unrelated packages. Any help is appreciated!


packages/core/src/modules/dids/methods/webvh/WebVHDidResolver.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  public readonly allowsCaching = true
+  public readonly allowsLocalDidRecord = true
+
+  // FIXME: Would be nice if we don't have to provide a did resolver instance
+  // private _resolverInstance = new Resolver()
+  // private resolver = didWeb.getResolver()
+
+  public async resolve(agentContext: AgentContext, did: string): Promise<DidResolutionResult> {
+    const result = await resolveDID(did)
```
TimoGlastra, line 21: This should probably have a try catch, as it seems the `resolveDID` can throw?

TimoGlastra (commented):
Hey @brianorwhatever, thanks so much for this PR!

I have some doubts with adding the didwebvh-ts library as is:
- It seems to depend on some dependencies that may not work well in React Native. Have you already tested the library to work in React Native? Libraries that may cause issues are nanoid ([fixed recently](https://github.com/ai/nanoid/issues/468), but not released yet), elysia (web framework?), multiformats (we had to remove this dependency and re-implement some stuff, not sure what parts of multiformats it uses). It's important that the core package works in all environments.
- crypto is handled by the didwebvh-ts libary. Generally we want all crypto to be handled by the wallet / crypto implementation of Credo and thus it'd be great if the crypto is pluggable . This is important as it's often required for organizations to have all crypto handled in a single place. Generally all our lower-level libs we create nowadays (sd-jwt vc, mdoc, oid4vc-ts) all don't come with any crypto support, and it needs to be provided dyncamically.

I understand this might not be so easy to fix, but unless we can meet these requirements I think it's better to create this as an extension module ,and we can clearly mention in the docs that this module depends on outside crypto /may not work in React Native. 

Curious to hear your thoughts

genaris (commented):
Thanks Brian! I think did:webvh resolving capabilities is a great addition to Credo. The problem I see is that, currently, `didwebvh-ts` seems to be too backend-oriented, including the `elysia` dependency (used to expose the resolver as a web server).

Do you think it would be possible to split this library into a pure JS-client package and a separate app that manages the REST API to resolve DIDs in a backend? I think this would be useful not only for us but also for other client-side apps.

We could also reimplement the resolver logic within Credo Core. I'd like to avoid that because it can be harder to maintain afterwards as did:webvh spec evolves, but can be of course more efficient in terms of dependency tree.

packages/core/src/modules/dids/methods/webvh/WebVHDidResolver.ts
brianorwhatever: thanks, will update!

brianorwhatever:
@TimoGlastra @genaris ok, great feedback thank you! I have not yet tested in React Native as I am having trouble getting bifold to use my local credo package.. so definitely could have some errors there.

Goal is definitely to have `did:webvh` resolvable in all environments. I was wondering if it would be better to have this as part of core or an extension as I see DID methods that follow both patterns. There will eventually be some resource resolution that comes along with this so maybe it would be better to follow the extension approach similar to how cheqd is implemented?

Here is the list of issues I will tackle based on the above comments:
1. remove `nanoid` dependency (totally unecessary)
2. move web server resolver into separate package
3. remove multiformats dependency (I've also had my fair share of struggles with this package.. should be able to reimplement what we need)
4. add pluggable crypto (will take a look at sd-jwt vc, mdoc, oid4vc-ts)

Please let me know if you think of anything else I should address

PR #2191, opened 2025-02-13
chore(mdoc): update dependency

berendsliedrecht:
- Update mdoc dep to 0.4.0
- create device response returns bytes and not base64 encoded


TimoGlastra (approved):

TimoGlastra:
Test related to oid4vc now fails. That normally doesn't happen

PR #2197, opened 2025-02-18
feat(x509): Update certificate creation API

berendsliedrecht:
Update the X.509 certificate creation api to not only create self-signed certs anymore.


packages/core/src/modules/x509/X509Certificate.ts
```diff
@@ -1,41 +1,55 @@
...
   id_ce_authorityKeyIdentifier,
+  id_ce_extKeyUsage,
   id_ce_keyUsage,
   id_ce_subjectAltName,
   id_ce_subjectKeyIdentifier,
   SubjectPublicKeyInfo,
 } from '@peculiar/asn1-x509'
 import * as x509 from '@peculiar/x509'
 
+x509.KeyUsageFlags
```
TimoGlastra, line 15:
```suggestion

```

I think this doesn't do anything right?

packages/core/src/modules/x509/X509Certificate.ts
```diff
@@ -47,6 +61,16 @@ export enum KeyUsage {
...
 }
 
+export enum X509ExtendedKeyUsage {
+  ServerAuth = '1.3.6.1.5.5.7.3.1',
+  ClientAuth = '1.3.6.1.5.5.7.3.2',
+  CodeSigning = '1.3.6.1.5.5.7.3.3',
+  EmailProtection = '1.3.6.1.5.5.7.3.4',
+  TimeStamping = '1.3.6.1.5.5.7.3.8',
+  OcspSigning = '1.3.6.1.5.5.7.3.9',
+  MdlDs = '1.0.18013.5.1.2',
```
TimoGlastra, line 71: Casing is intentional here?

packages/core/src/modules/x509/X509Certificate.ts
```diff
@@ -228,6 +311,11 @@ export class X509Certificate {
     return certificate.subject
   }
 
+  public get issuer() {
+    const certificate = new x509.X509Certificate(this.rawCertificate)
```
TimoGlastra, line 315: Should we store an instance of the x509 lib cert in this class? I'm not sure on the overhead of parsing

TimoGlastra (approved):

packages/core/src/modules/x509/X509Certificate.ts
berendsliedrecht: yes

PR #2204, opened 2025-02-22
feat: allow kid (not a did) to be combined with x5c/jwk

TimoGlastra:
This is a pattern commonly used and breaks interop with Credo.

### DIFF
--- a/packages/core/src/crypto/JwsService.ts
+++ b/packages/core/src/crypto/JwsService.ts
@@ -211,17 +211,15 @@ export class JwsService {
   }
 
   private buildProtected(options: JwsProtectedHeaderOptions) {
-    const x5cJwkCount = [options.jwk, options.x5c].filter(Boolean).length
-
     // FIXME: checking for kid starting with '#' is not good.
     // but now we don't really limit that kid (did key reference)
     // cannot be combined with x5c/jwk
     if (options.kid?.startsWith('did:')) {
-      if (x5cJwkCount > 0) {
+      if (options.jwk || options.x5c) {
         throw new CredoError("When 'kid' is a did, 'jwk' and 'x5c' cannot be provided.")
       }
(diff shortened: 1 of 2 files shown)


packages/core/src/crypto/JwsService.ts
```diff
@@ -241,16 +250,25 @@ export class JwsService {
...
 
-    if ([protectedHeader.jwk, protectedHeader.kid, protectedHeader.x5c].filter(Boolean).length > 1) {
-      throw new CredoError('Only one of jwk, kid and x5c headers can and must be provided.')
+    const x5cJwkCount = [protectedHeader.jwk, protectedHeader.x5c].filter(Boolean)
+
+    // FIXME: checking for kid starting with '#' is not good.
+    // but now we don't really limit that kid (did key reference)
+    // cannot be combined with x5c/jwk
+    if (typeof protectedHeader.kid === 'string' && protectedHeader.kid?.startsWith('did:')) {
+      if (x5cJwkCount.length > 0) {
```
auer-martin, line 259: I think just doing this would be cleaner

```ts
if (protectedHeader.jwk || protectedHeader.x5c) is cleaner here
```
and this below
```ts
(protectedHeader.jwk && protectedHeader.x5c)
```

packages/core/src/crypto/JwsService.ts
```diff
@@ -241,16 +250,25 @@ export class JwsService {
...
+        throw new CredoError("When 'kid' is a did, 'jwk' and 'x5c' cannot be provided.")
+      }
+    } else if (x5cJwkCount.length > 1) {
+      throw new CredoError("Header must contain one of 'kid' with a did value, 'x5c', or 'jwk'.")
```
auer-martin, line 263: Header must contain one of 'x5c', 'jwk' or 'kid' with a did value,.

auer-martin (approved):

PR #2205, opened 2025-02-22
fix: support verification of certificate chain without root

TimoGlastra:
… the root certificate is not included in the certificate chain but present in the trusted certificates


auer-martin (approved):
LGTM

PR #2208, opened 2025-02-28
chore(cheqd): Bump @cheqd/sdk

DaevMithran:


DaevMithran:
> Thanks for the update @DaevMithran. It looks like we have an issue with E2E tests. Maybe because the changes in default rpcUrl?

@genaris  Fixed it thanks

TimoGlastra (approved):

TimoGlastra:
It would be great if we can test this in React Native before merging.



We had to pin the SDK to very specific versions: https://github.com/animo/paradym-wallet/blob/d41a957bbfbf6167993ec319903d0f19ee6c44ab/pnpm-workspace.yaml#L39-L40

PR #2210, opened 2025-03-02
chore: biomejs more updates

sairanjit:
### Summary

- [x] Updated all `foreach` with biome errors to `forof`
- [x] Enable recommended rules - By default it is `true`
- [x] Fixed `noConfusingVoidType` rule
- [x] Why "noUnsafeDeclarationMerging": "off" - Added biome-ignore for the required APIs
- [x] Added `noRestrictedGlobals` rule for `Buffer` globally

### DIFF
--- a/packages/didcomm/src/handlers/MessageHandler.ts
+++ b/packages/didcomm/src/handlers/MessageHandler.ts
@@ -4,7 +4,7 @@ import type { InboundMessageContext, OutboundMessageContext } from '../models'
 export interface MessageHandler {
   readonly supportedMessages: readonly ConstructableAgentMessage[]
 
-  handle(messageContext: InboundMessageContext): Promise<OutboundMessageContext | void>
+  handle(messageContext: InboundMessageContext): Promise<OutboundMessageContext | undefined>
 }
 
 /**
--- a/.changeset/polite-bees-serve.md
+++ b/.changeset/polite-bees-serve.md
@@ -1,0 +1,9 @@
+---
+"@credo-ts/action-menu": minor
+"@credo-ts/anoncreds": minor
+"@credo-ts/askar": minor
+"@credo-ts/core": patch
+"@credo-ts/didcomm": minor
+---
+
+- messagehandler should return undefined if it doesn't want to response with a message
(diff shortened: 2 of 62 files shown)


packages/didcomm/src/handlers/MessageHandler.ts
```diff
@@ -4,7 +4,7 @@ import type { InboundMessageContext, OutboundMessageContext } from '../models'
 export interface MessageHandler {
   readonly supportedMessages: readonly ConstructableAgentMessage[]
 
-  handle(messageContext: InboundMessageContext): Promise<OutboundMessageContext | void>
+  handle(messageContext: InboundMessageContext): Promise<OutboundMessageContext | undefined>
```
TimoGlastra, line 7: This is a breaking change. I think a good one, but maybe you can add a changeset entry that messagehandler should return undefined if it doesn't want to response with a message

TimoGlastra (approved):

packages/didcomm/src/handlers/MessageHandler.ts
sairanjit: I am currently changing all the handlers to support this. Which of the following approach is the the best/correct to go with ?
1. Make the function return type as `Promise<undefined>` 
```typescript
public async processPerform(): Promise<undefined> {

    ...
}  
```
2. Purposefully `return undefined` from the function
```typescript
public async processPerform() {

    ...
    return undefined
}
```

TimoGlastra: I think option 2 makes sense!

TimoGlastra:
Hey, tested the biome stuff locally. I think to make it easier to find current issue we should:
- set noUndeclaredDependencies to off for now until v2 is released
- set useImportType to off. This is recommended by biome for projects using decorators.

Then we can fix the remaining warnings (Which are only a few)

sairanjit:
> * useImportType

Sure I will do that

packages/didcomm/src/handlers/MessageHandler.ts
sairanjit: Added tis changes now

PR #2213, opened 2025-03-05
chore: remove warnings

TimoGlastra:
Removes all warnings to not clutter the output with existing code.

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -75,7 +75,7 @@
       "performance": {
         "noDelete": {
           "fix": "none",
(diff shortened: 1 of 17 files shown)


biome.json
```diff
@@ -75,7 +75,7 @@
       "performance": {
         "noDelete": {
           "fix": "none",
-          "level": "warn"
```
sairanjit, line 78: Ah missed this from my side. Thanks for fixing.

genaris (approved):

PR #2226, opened 2025-03-12
fix: presenting proof for non Mdoc VCs

### LINKED ISSUE #2222: Unable to present OpenID credential
Using the latest stable release 0.5.13 I am able to receive proof request for w3C/sd-jwt credentials with no issues
However when attempting to send the presentation to the verifier credo raises the following exception

"Cannot read property 'map' of undefined" 

Triggered from 
```
agent.modules.openId4VcHolder.acceptSiopAuthorizationRequest({
        authorizationRequest,
        presentationExchange: {
          credentials,
        },
```

Tested with several verifiers including "ssi.dutchblockchaincoalition.org" with no hope
Also this only happens in version 0.5.13, not in 0.5.11 which works fine

I also tried to update to version "0.5.14-alpha-20250224091806" which is used by paradym wallet but couldn't find this version in the npm registry

MosCD3:
This fixes #2222


TimoGlastra:
Is this issue also present in the latest 0.6 alpha? Since this PR won't affect 0.5.x releases

MosCD3:
We ended up patching 0.5.13 to get Bifold to work with SD-JWT and JWTs

MosCD3:
@TimoGlastra Any thing needs to be done for this fix to go in?

TimoGlastra:
I'm not really sure what is causing this error.

The pex library should return an array. And the version being used in for pex 0.6 is different than 0.5

So:
- does this error only occur on 0.5 or also 0.6? If only on 0.5 it should not go into the 0.6 branch
- can you provide a reproducible example where no array is returned? I would like to add a test.

So before merging i'd like to understand the root cause, as the pex library should return an array in all cases

MosCD3:
@TimoGlastra Couple of things here
1st, this is not about PEX, there is an issue with logic itself
Observe when its MDOC, the object is structured this way
```
 verifiablePresentationResultsWithFormat.push({
          verifiablePresentationResult: {
            presentationSubmission: presentationSubmission,
            verifiablePresentations: [deviceResponseBase64Url],
            presentationSubmissionLocation,
          },
          claimFormat: presentationToCreate.claimFormat,
        })
```
  When its not, the object is structured this way
```
   verifiablePresentationResultsWithFormat.push({
          verifiablePresentationResult,
          claimFormat: presentationToCreate.claimFormat,
        })
```
So, if PEX library returns and array regardless, this function has to be refactored to always construct the object ` verifiablePresentationResultsWithFormat` consistent regardless

About testing, I can't see any docs or sample implementation for credo-6 to try with. Please guide me if you have more info

TimoGlastra:
This doesn't seem to be an issue in the main branch anymore. If an error occurs, it's because a mismatch in the PEX library being used

PR #2229, opened 2025-03-13
fix: didcomm public did x25519 for connection

TimoGlastra:
Fixes an issue where a public/custom did document is used for connection invitations, where the didcomm service uses a X25519 key instead of an Ed25519 key. We now reuse the existing DidcommDocumentService to resolve the X25519 key to an Ed25519 key


berendsliedrecht (approved):

genaris (approved):
A tricky one, certainly! 

I think we didn't find this issue before because:
- `didexchange-numalgo.test.ts` does test different situations, but in every case the invitation is created with a peer did
- did:peer:4 ACA-Py interop works, since it is the usually the one who generates the invitation and the responsible for creating DID Exchange Response message

So I think we can add some cases in the test suite where `createInvitation` is called with an invitationDid or so that can reflect this issue more.

I'm wondering if there is a way we can fix this in order to let 0.6 agents work with current 0.5.13. Probably we should backport the fix to 0.5 branch along with the fix for revocation notification.

TimoGlastra:
> I'm wondering if there is a way we can fix this in order to let 0.6 agents work with current 0.5.13. Probably we should backport the fix to 0.5 branch along with the fix for revocation notification.

What do you mean with this? As i understand it, this fix will let 0.6 agents work with current 0.5.13 agent right?

TimoGlastra:
Yeah to not break anything we're going to keep using did:peer:1 for a while still, until we can assume everything is updated to 0.6 😬

PR #2233, opened 2025-03-22
fix: support non-self-signed x509 certificate is trusted

TimoGlastra:
Two noteworthy things:
- The slice of the parsedChain took all the certificates from root up until the trusted one (if trusted certificates were provided), not from the trusted certificate until the leaf certificate. Probably because the order of certificates is reversed between X509 and JWS, which is confusing. The certificate signatures were still validated, however it meant that for the leaf certificate the time based verification was not performed correctly if trusted certificates were provided
- Fixing the above highlighted another issue: verifying a chain where we trust a non-self-signed intermediate certificate would fail. I adapted the code to correctly try to extract the certificate from the parent in the chain (we start verification from the cert we trust, and try extract the key that signed the cert from the more root cert). However this approach won't fully work if the root certificate is not present. To not break receiving mDOCs from unknown issuers (see https://github.com/openid/OpenID4VCI/issues/62 for context), we now skip the signature verification of the first certificate in the chain we build if it's not self-signed, if it's trusted and we don't have the parent signer key. i guess since the cert is already trusted it's ok-ish. But we might need to change this behaviour in the future, and instead throw an error?


packages/core/src/modules/x509/X509Certificate.ts
```diff
@@ -57,22 +57,18 @@ export enum X509ExtendedKeyUsage {
 export type X509CertificateOptions = {
   publicKey: Key
   privateKey?: Uint8Array
-  extensions?: Array<x509.Extension>
-  rawCertificate: Uint8Array
+  x509Certificate: x509.X509Certificate
```
berendsliedrecht, line 60: I don't think it should be a required property right? Makes it very strict, we can build the x509 cert from the options, right?

berendsliedrecht: Ah I see its just a change from the raw certificate.

Maybe we should allow to pass in the raw or the x509.X509Certificate? Otherwise the user would need to install the @peculiar lib

berendsliedrecht (approved):

packages/core/src/modules/x509/X509Certificate.ts
TimoGlastra: So i'll make the constructor private, and you can use the static members (fromEncodedCertificate, fromRawCertificate)

PR #2234, opened 2025-03-24
fix: Update package.json in demo-openid in order to run provider script

pmccabensds:
I was messing around with the open-id demo and noticed a small error in the config to run the provider


genaris (approved):
Thanks! Can you fix the DCO issue? https://github.com/openwallet-foundation/credo-ts/pull/2234/checks?check_run_id=39292259097

TimoGlastra:
Seems to be fixed in main

PR #2237, opened 2025-03-24
fix: additional unqualified did:sov support with anoncreds

jamshale:
This is for a problem we were seeing with the bc wallet when trying to use the anoncreds proof format with unqualified did sov's. I believe this should be supported by credo due to all the utility functions for the conversion.

### DIFF
--- a/packages/anoncreds/src/anoncreds-rs/AnonCredsRsHolderService.ts
+++ b/packages/anoncreds/src/anoncreds-rs/AnonCredsRsHolderService.ts
@@ -143,9 +143,10 @@ export class AnonCredsRsHolderService implements AnonCredsHolderService {
           }
         }
 
-        const { linkSecretId, revocationRegistryId, credentialRevocationId } = getAnoncredsCredentialInfoFromRecord(
+        const proofUsesUnqualifiedIdentifiers = proofRequestUsesUnqualifiedIdentifiers(proofRequest)
+        const { linkSecretId, schemaId, credentialDefinitionId, revocationRegistryId, credentialRevocationId } = getAnoncredsCredentialInfoFromRecord(
           credentialRecord,
-          proofRequestUsesUnqualifiedIdentifiers(proofRequest)
+          proofUsesUnqualifiedIdentifiers
         )
 
         // TODO: Check if credential has a revocation registry id (check response from anoncreds-rs API, as it is
@@ -186,7 +187,12 @@ export class AnonCredsRsHolderService implements AnonCredsHolderService {
                   credential: credentialRecord.credential as W3cJsonLdVerifiableCredential,
                 })
               : (credentialRecord.credential as AnonCredsCredential)
-
+          
+          if (proofUsesUnqualifiedIdentifiers) {
+            credential.schema_id = schemaId
+            credential.cred_def_id = credentialDefinitionId
+            credential.rev_reg_id = credentialRevocationId != null ? credentialDefinitionId : undefined 
+          }
           return {
             linkSecretId,
             credentialId: attribute.credentialId,
--- a/packages/anoncreds/src/anoncreds-rs/AnonCredsRsHolderService.ts
+++ b/packages/anoncreds/src/anoncreds-rs/AnonCredsRsHolderService.ts
@@ -144,10 +144,8 @@ export class AnonCredsRsHolderService implements AnonCredsHolderService {
         }
 
         const proofUsesUnqualifiedIdentifiers = proofRequestUsesUnqualifiedIdentifiers(proofRequest)
-        const { linkSecretId, schemaId, credentialDefinitionId, revocationRegistryId, credentialRevocationId } = getAnoncredsCredentialInfoFromRecord(
-          credentialRecord,
-          proofUsesUnqualifiedIdentifiers
-        )
+        const { linkSecretId, schemaId, credentialDefinitionId, revocationRegistryId, credentialRevocationId } =
+          getAnoncredsCredentialInfoFromRecord(credentialRecord, proofUsesUnqualifiedIdentifiers)
 
         // TODO: Check if credential has a revocation registry id (check response from anoncreds-rs API, as it is
         // sending back a mandatory string in Credential.revocationRegistryId)
@@ -187,11 +185,11 @@ export class AnonCredsRsHolderService implements AnonCredsHolderService {
                   credential: credentialRecord.credential as W3cJsonLdVerifiableCredential,
                 })
               : (credentialRecord.credential as AnonCredsCredential)
-          
+
           if (proofUsesUnqualifiedIdentifiers) {
             credential.schema_id = schemaId
             credential.cred_def_id = credentialDefinitionId
-            credential.rev_reg_id = credentialRevocationId != null ? credentialDefinitionId : undefined 
+            credential.rev_reg_id = credentialRevocationId != null ? credentialDefinitionId : undefined
           }
           return {
             linkSecretId,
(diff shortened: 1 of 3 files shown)


swcurran:
@jamshale — sorry for the obvious question, but while this handles unqualified `did:sov` identifiers, are qualified `did:sov` identifiers also handled?  Just want to make sure this doesn’t just fix one use case and break another.

Thanks for getting this one in there!

swcurran:
@bryce-mcmath — what is the path to get this into Bifold Wallet and BC Wallet?  Is there a version of Credo-TS currently in BC Wallet to which a patch can be applied?  Or do we have to wait for a release of the latest Credo, and then wait until Bifold Wallet updates to that latest release?  Thanks.

jamshale:
When I tried to create did:sov qualified identifiers they did match a registry and issuance failed. 

I think that may be a different problem to solve. Hopefully just changing the registry pattern to match did:sov and unqualified dids. I haven't tried to figure that out yet. 

I think this would break a different case. I don't think qualified did:sov's are currently supported in anoncreds format for credo or acapy.

bryce-mcmath:
@swcurran for Bifold and it's extensions, going from Credo 0.5.13 to 0.6.x is going to be a sizeable jump. Should be simple to get this into a patch for the 0.5.13 that Bifold and BC Wallet uses though

TimoGlastra:
We have a branch to make patch releases for credo 0.15, i think this is a good candidate!

bryce-mcmath:
@TimoGlastra do you mean the `v0.5.x` branch? I can't find anything close to 0.15

TimoGlastra (commented):
I'm not 100% what this PR aims to fix.

Is it credentials that were previously issued in unqualified format and now are requested as qualified format? Because the AnonCresdProofFormatService should not be used with unqualified identifiers. So the request and response should use qualified.

Also Credo stores the qualified identifers for quite a while already, so it would be for credentials that were issued a while back, correct?

We also have migration scripts in place, but since those require internet there could be chance not all credentials are migrated.

Just trying to understand the exact flow that triggers this issue

jamshale:
Acapy is sending unqualified did:sov identifiers through the anoncreds format. This allows the proof request to find the credentials. 

We did try with fully qualified did:sov identifiers as well, but credo wouldn't accept it. Acapy aims to have fully qualified identifiers but doesn't yet.

This seems like this easiest way to allow proof requests with unqualified did:sov identifiers. It looked like this was intended already as credential issuance worked, and there were already util functions for the conversion.

bryce-mcmath:
@TimoGlastra there is steps to reproduce at the top of this ticket in the first comment (just use Credo rather than Bifold / BC Wallet, they have the same issue) https://github.com/openwallet-foundation/acapy/issues/3576#issue-2911707592

genaris:
IIRC, our understanding when AnonCreds format was that everybody would use it with fully qualified identifiers. The logic for legacy, unqualified dids was there mainly for the indy credential/proof format services, which were there to support existing deployments.

But certainly I don't find in [AnonCreds spec](https://hyperledger.github.io/anoncreds-spec/#issuer-identifiers) any restriction to do use unqualified identifiers, especially for this did:sov:

> AnonCreds identifiers MUST be a Uniform Resource Identifier (URI) conformant with [RFC3986](https://www.rfc-editor.org/rfc/rfc3986), although one notable exception is permitted. The exception is that for backwards compatibility, the Indy DIDs used in the early (pre did:indy) open source Hyperledger Indy implementation are permitted.

I think it is unfortunate that the community has started using this anoncreds attachment format with unqualified identifiers, but that's how it is :cry:.

TimoGlastra:
> Acapy is sending unqualified did:sov identifiers through the anoncreds format. This allows the proof request to find the credentials.

In this case the LegacyIndyProofFormatService should be used. We support this proof format for V2 protocols as well. I think it's important to keep the distinction between the indy proof format (unqualified) and anoncreds proof format (qualified)

> We did try with fully qualified did:sov identifiers as well, but credo wouldn't accept it. Acapy aims to have fully qualified identifiers but doesn't yet.

did:sov identifiers are not valid sadly. It needs to be a qualified did:indy indentifier in this case.

> This seems like this easiest way to allow proof requests with unqualified did:sov identifiers. It looked like this was intended already as credential issuance worked, and there were already util functions for the conversion.

I think the easiest approach would be to keep using the legacy indy proof format for now. I'm not 100% why credential issuance worked (did you use legacy indy credential foramt, or also the anoncreds credential format), in case of the latter it should not work with unqualified identifiers.

Credo supports the unqualified format still, so I don't see why that couldn't be used? We have a lot of utils for conversion, but this is all based on the assumption that indy servides use unqualified and anoncreds services use qualified. The logic is to account for e.g. a credential received in unqualfied format and share it qualified format. 

Hope this makes sense

jamshale:
Hmm. Ok. We definitely used the `anoncreds` format for issuance and it worked fine. I was under the impression that there was an exception for old type dids in the spec https://hyperledger.github.io/anoncreds-spec/#anoncreds-identifiers and that the format should handle unqualified. Is there somewhere where this is explained. I haven't found it.

In acapy we wanted to be able to upgrade to use anoncreds format with existing did:sov did's because it would prevent needing to specify indy or anoncreds format for different types of dids.

I guess if this change isn't acceptable we can force acapy controllers to use the indy format for any unqualifed did:sov's.

TimoGlastra:
If this is what is needed to get interop I'm fine with merging it. The changes are really small. 👍 

However, we have been discussing removing support for v1 versions of the protocols and also  support for unqualified identifiers in the near future. It would be great if you could keep us posted on the status of qualified identifier support in ACA-Py (i was of the impression this was already supported).

TimoGlastra:
And I must say i didn't know about the following statement in the spec. Technically what you're trying to do seems valid, I guess I just hoped it wouldn't be true 😅 :

> AnonCreds identifiers MUST be a Uniform Resource Identifier (URI) conformant with [RFC3986](https://www.rfc-editor.org/rfc/rfc3986), although one notable exception is permitted. The exception is that for backwards compatibility, the Indy DIDs used in the early (pre did:indy) open source Hyperledger Indy implementation are permitted.

TimoGlastra (approved):

jamshale:
It is supported now but the did:sov method is widely used and has always been unqualified. 

Thanks for your help. I want to try and get acapy more on the same page as credo. We will push to discontinue any unqualifed did usage. I do very much agree any applications should avoid using these and rotating existing dids to qualified if possible.

TimoGlastra:
> I do very much agree any applications should avoid using these and rotating existing dids to qualified if possible

There's not really a need to rotate. The same hosted did can be resolved as an unqualified did, or a qualified did:indy did. So it's mostly about updating the application logic to handle did:indy DIDs correctly.

TimoGlastra:
`@credo-ts/*` version `0.5.14-alpha-20250325224809` should be available in few minutes, which contain the fixes. If you have other small issues in 0.5.13 we can backport these as well

swcurran:
Awesom @TimoGlastra — thanks!  We’ll look more at unqualified identifiers to continue the move away from them.

@bryce-mcmath — let us know if you can test the fix in BC Wallet and move towards a published version of the app with a fix.

bryce-mcmath:
> @bryce-mcmath — let us know if you can test the fix in BC Wallet and move towards a published version of the app with a fix.

I plan to get the patch in today or latest tomorrow, will ping you once it's in Testflight

jamshale:
I think I'm just confused what the difference between did:sov and did:indy really is. I don't know all the history. I was going off this was the difference https://hyperledger.github.io/indy-did-method/#nym-transaction-version. But it seems like what are unqualified did:sov's in acapy are did:indy in credo?

@bryce-mcmath Thanks. There's still the issue with revocation notification. I'm trying to create a fix just in acapy but have hit a roadblock with getting the correct namespace.

PR #2238, opened 2025-03-24
feat(webvh): Added did:webvh resolver

brianorwhatever:
@genaris @TimoGlastra as discussed I have pulled the crypto out of didwebvh-ts while also removing all of the dependencies. Please take a look at this new PR


packages/webvh/src/WebvhModuleConfig.ts
```diff
@@ -0,0 +1,22 @@
...
+export class WebvhModuleConfig {
+  private options: WebvhModuleConfigOptions
+
+  public constructor(options?: WebvhModuleConfigOptions) {
+    this.options = options ?? {}
+  }
+
+  /** See {@link WebvhModuleConfigOptions.baseUrl} */
+  public get baseUrl() {
+    return this.options.baseUrl ?? 'https://webvh.io'
```
TimoGlastra, line 20: What does this default domain do? Is the baseUrl for when you create DIDs? I think having no default value and allowing it to be passed when creating one makes sense.

packages/webvh/src/dids/WebvhDidRegistrar.ts
```diff
@@ -0,0 +1,239 @@
...
+            didDocumentMetadata: {},
+            didRegistrationMetadata: {},
+            didState: {
+              state: 'failed',
+              reason: 'Invalid private key provided',
+            },
+          }
+        }
+        const privateKey = Buffer.from(MultiBaseEncoder.decode(privateKeyMultibase).data)
+        if (privateKeyMultibase && !isValidPrivateKey(privateKey, KeyType.Ed25519)) {
```
TimoGlastra, line 46: Is Ed25519 required? Just curious, we'd like to use it with P256 as well

packages/webvh/src/dids/WebvhDidRegistrar.ts
```diff
@@ -0,0 +1,239 @@
...
+          return {
+            didDocumentMetadata: {},
+            didRegistrationMetadata: {},
+            didState: {
+              state: 'failed',
+              reason: 'Invalid private key provided',
+            },
+          }
+        }
+        const privateKey = Buffer.from(MultiBaseEncoder.decode(privateKeyMultibase).data)
```
TimoGlastra, line 45: Is it possible to pass an existing public key, but not the private key? We usually create a key first, then use that to create a diddoc

(This is fine to add later, just trying to understand how it works)

packages/webvh/src/dids/WebvhDidRegistrar.ts
```diff
@@ -0,0 +1,239 @@
...
+          didRegistrationMetadata: {},
+          didState: {
+            state: 'failed',
+            reason: 'Provide a didDocument or at least one verificationMethod with seed in secret',
+          },
+        }
+      }
+
+      // Register the DID using didwebvh-ts
+      const response = await createDID()
```
TimoGlastra, line 79: Shouldn't we somehow pass the did document here?

packages/webvh/src/dids/WebvhDidRegistrar.ts
```diff
@@ -0,0 +1,239 @@
...
+          didRegistrationMetadata: {},
+          didState: {
+            state: 'failed',
+            reason: 'Provide a valid didDocument',
+          },
+        }
+      }
+
+      // Update the DID using didwebvh-ts
+      const response = await updateDID({doc: options.didDocument.toJSON()})
```
TimoGlastra, line 128: I think we should first fetch the record to see if it exists

packages/webvh/src/dids/WebvhDidRegistrar.ts
```diff
@@ -0,0 +1,239 @@
...
+        didState: {
+          state: 'failed',
+          reason: `unknownError: ${error.message}`,
+        },
+      }
+    }
+  }
+
+  public async deactivate(agentContext: AgentContext, options: WebvhDidDeactivateOptions): Promise<DidDeactivateResult> {
+    // const didRepository = agentContext.dependencyManager.resolve(DidRepository)
```
TimoGlastra, line 160: If this is not supported yet we can return an error for now and remove this code

packages/webvh/src/dids/WebvhDidRegistrar.ts
```diff
@@ -0,0 +1,239 @@
...
+export interface WebvhDidDeactivateOptions extends DidCreateOptions {
+  method: 'webvh'
+  did: string
+  options?: {
+    versionId?: string
+  }
+}
+
+interface IVerificationMethod {
+  type: string
```
TimoGlastra, line 235: Type must be MultiKey right?

packages/webvh/src/dids/didWebvhUtil.ts
```diff
@@ -0,0 +1,94 @@
...
+    })
+    this.agentContext = agentContext
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  public async sign(input: SigningInput): Promise<SigningOutput> {
+    throw new Error('Not implemented')
+  }
+
+  public async verify(signature: Uint8Array, message: Uint8Array, publicKey: Uint8Array): Promise<boolean> {
```
TimoGlastra, line 25: It's probably good if the interface from did web vh passes the keytype as well, so it's ready to support additional key types as well

packages/webvh/src/dids/didWebvhUtil.ts
```diff
@@ -0,0 +1,94 @@
...
+
+      const key = new Key(publicKey, KeyType.Ed25519)
+
+      return await this.agentContext.wallet.verify({
+        key,
+        data: Buffer.from(message),
+        signature: Buffer.from(signature),
+      })
+    } catch (error) {
+      // Log error in a non-production environment
```
TimoGlastra, line 39: You can use agentContext.config.logger.error for this, and can be disabled by the user of the framework

packages/webvh/src/dids/didWebvhUtil.ts
```diff
@@ -0,0 +1,94 @@
+import { type AgentContext, Buffer, DidDocument, JsonTransformer, KeyType, Key } from '@credo-ts/core'
+import { createDID, AbstractCrypto, type SigningOutput, type SigningInput } from 'didwebvh-ts'
+
+export class DIDWebvhCrypto extends AbstractCrypto {
+  private agentContext?: AgentContext
+
+  public constructor(agentContext?: AgentContext) {
```
TimoGlastra, line 7: I think we can make agent context required

TimoGlastra (commented):
The new approach is awesome, thanks @brianorwhatever. I left some comments. It seems some parts are todo, do you want to address these as part of this PR, or in another one?

packages/webvh/src/dids/WebvhDidRegistrar.ts
brianorwhatever: Yes unfortunately ED25519 is required for did:webvh for update keys. We did it that way to standardize it on one interoperable crypto suite you can of course use any keys in a DID document you desire.

brianorwhatever:
@TimoGlastra I forgot to mention I was focussing solely on resolver in this PR. I added some of the base stuff for the registrar, which will be coming soon. I'm happy to either remove that stuff or fix it up later. Higher priority is resolving DID documents and resources.

TimoGlastra:
Ah that makes sense @brianorwhatever. Let's focus this pr on the resolving in that case and we can add the registration parts next 👍

brianorwhatever:
OK, great I will remove the stuff about the registrar. Unfortunately, I don't think that this is 0.3 compatible I can compare the differences and see what would be required to make it work.

brianorwhatever:
Hey @TimoGlastra, quick question about verifying Data Integrity proofs in a custom setup.

We're building an `AnonCredsRegistry` for `did:webvh` (`WebVhAnonCredsRegistry`). The objects we fetch (schemas, cred defs, etc.) are structured as simple JSON-LD "AttestedResources" which basically wrap the `content` (the actual AnonCreds object) and include a standard DI `proof` (e.g., `eddsa-jcs-2022`).

Inside our registry logic (in a helper `_resolveAndValidateAttestedResource`), we're already handling:

*   Resolving the `did:webvh:.../resource/hash` URI.
*   Parsing the JSON-LD.
*   Checking the `type` includes `AttestedResource`.
*   Verifying a multihash of the canonicalized `content` matches the `hash` in the URI.
*   Checking the `issuerId` in the `content` matches the DID from the URI.

The last piece is actually verifying the `proof` signature itself. We looked into `DataIntegrityService` and `W3cCredentialService`, but they didn't seem designed for verifying proofs on arbitrary signed objects like this (it's not strictly a VC/VP).

We also considered using the signature suites directly (like `Ed25519Signature2020`), getting them from the `SignatureSuiteRegistry`, and calling their `verifyProof` method. This looks possible but feels like we might be duplicating internal logic, especially around preparing the data correctly for the specific suite and ensuring the `documentLoader` (which uses our `WebvhDidResolver`) is passed through correctly to resolve the `verificationMethod`.

**Is there a recommended/simpler way in Credo to verify a DI proof on a custom JSON-LD object like this, making sure key resolution via the `verificationMethod` works as expected?** Or is using the signature suite directly the intended path here?

Currently have it stubbed out, but want to make sure we implement the crypto verification correctly using Credo's infrastructure.

Thanks!

brianorwhatever:
@TimoGlastra @genaris this is ready for a review. I have updated the didwebvh-ts package to the new version which supports version 1.0 of the spec and addressed any comments in this thread. Thanks!

brianorwhatever:
The integration tests probably don't need to be checked in they are testing real resource resolution and those files will likely disappear one day

packages/webvh/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "@credo-ts/webvh",
+  "version": "0.4.0",
```
TimoGlastra: Should be synced with the other packages

```suggestion
  "version": "0.5.13",
```

packages/webvh/src/anoncreds/services/WebVhAnonCredsRegistry.ts
```diff
@@ -0,0 +1,568 @@
...
+// Simple multibase decoder for base58btc (z prefix)
+function decodeMultibase(data: string): { data: Uint8Array; baseName: string } {
+  if (data[0] === 'z') {
+    // base58btc encoding
+    const base58Data = data.substring(1)
+    const decoded = decodeFromBase58(base58Data)
+    return { data: decoded, baseName: 'base58btc' }
+  }
+  throw new Error(`Unsupported multibase prefix '${data[0]}'`)
+}
```
TimoGlastra: We have a multibase decoder (https://github.com/openwallet-foundation/credo-ts/blob/879ed2c9178684d14e6850cbcdb02d6832e0d730/packages/core/src/utils/MultiBaseEncoder.ts#L48), as well as a Base58 decoder (https://github.com/openwallet-foundation/credo-ts/blob/879ed2c9178684d14e6850cbcdb02d6832e0d730/packages/core/src/utils/TypedArrayEncoder.ts#L48)

packages/webvh/src/anoncreds/utils/base58.ts
TimoGlastra, line 1: Can be replaced with TypedArrayEncoder

packages/webvh/src/anoncreds/utils/multihash.ts
TimoGlastra, line 1: We have a multihash implementation already: https://github.com/openwallet-foundation/credo-ts/blob/879ed2c9178684d14e6850cbcdb02d6832e0d730/packages/core/src/utils/MultiHashEncoder.ts.

If there's any features missing in our multi hash implementation, it would be better to extend the existing functionaility, this way we prevent duplication of logic

packages/webvh/src/anoncreds/utils/transform.ts
```diff
@@ -0,0 +1,161 @@
...
+  IsObject,
+  Validate,
+  ValidatorConstraint,
+  ValidatorConstraintInterface,
+  ValidationArguments,
+  IsNumber,
+} from 'class-validator'
+
+@ValidatorConstraint({ name: 'isStringOrStringArray', async: false })
+export class IsStringOrStringArray implements ValidatorConstraintInterface {
```
TimoGlastra: We already have this validation implemented, it would prefer we export it from core and reuse it: https://github.com/openwallet-foundation/credo-ts/blob/879ed2c9178684d14e6850cbcdb02d6832e0d730/packages/core/src/utils/transformers.ts#L66

packages/webvh/src/anoncreds/utils/transform.ts
```diff
@@ -0,0 +1,161 @@
...
+}
+
+export class WebVhSchemaContent {
+  @IsArray()
+  @IsString({ each: true })
+  @ArrayMinSize(1) // Example validation: ensure attrNames is not empty
+  @Type(() => String)
+  public attrNames!: string[]
+
+  @IsString() // Example validation
```
TimoGlastra: "Example validation" can be removed everywhere i think?

packages/webvh/src/anoncreds/utils/transform.ts
```diff
@@ -0,0 +1,161 @@
...
+  public type!: string[] // Keep as array internally after normalization
+
+  @IsString()
+  public id!: string
+
+  @Expose()
+  @Transform(({ value }) => {
+    // Determine which class to use based on the content
+    if (value && 'attrNames' in value) {
+      return Object.assign(new WebVhSchemaContent(), value)
```
TimoGlastra: Does this stil trigger the validation i you're doing it like this? Usually i use the JsonTransformer, but if this works, it's ok 👍

packages/webvh/src/dids/WebvhDidResolver.ts
```diff
@@ -0,0 +1,128 @@
...
+
+import { DidDocument } from '@credo-ts/core'
+import { resolveDID } from 'didwebvh-ts'
+
+import { DIDWebvhCrypto } from './didWebvhUtil'
+
+export class WebvhDidResolver implements DidResolver {
+  public readonly supportedMethods = ['webvh']
+  public readonly allowsCaching = false
+  public readonly allowsLocalDidRecord = false
```
TimoGlastra: Any reason these are both false? For the `did:web` one they are both true, I'd assume the same values here

packages/webvh/src/dids/WebvhDidResolver.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  public async resolve(agentContext: AgentContext, did: string): Promise<DidResolutionResult> {
+    const didDocumentMetadata = {}
+
+    try {
+      return await this.resolveDidDoc(agentContext, did)
+    } catch (error) {
+      agentContext.config.logger.error(`Error resolving DID: ${error}`)
```
TimoGlastra:
```suggestion
      agentContext.config.logger.error(`Error resolving DID ${did}: ${error}`)
```

packages/webvh/src/dids/WebvhDidResolver.ts
```diff
@@ -0,0 +1,128 @@
...
+    try {
+      agentContext.config.logger.debug(`Attempting to resolve resource: ${resourceUrl}`)
+      
+      // Use the getBaseUrl method to parse the did:webvh resource URL
+      const httpsUrl = this.getBaseUrl(resourceUrl)
+      
+      agentContext.config.logger.debug(`Fetching resource from: ${httpsUrl}`)
+      
+      // Fetch the resource using native fetch
+      const response = await fetch(httpsUrl)
```
TimoGlastra: Should use fetch from the agent config. `agentContext.config.agentDependencies.fetch`

packages/webvh/src/dids/didWebvhUtil.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  public constructor(agentContext: AgentContext) {
+    super({
+      verificationMethod: {
+        id: 'did:webvh:123',
+        controller: 'did:webvh:123',
+        type: 'Ed25519VerificationKey2020',
+        publicKeyMultibase: '123',
+        secretKeyMultibase: '123',
+      },
```
TimoGlastra: Why do we set it to these sample values? Should this be optional in the `AbstractCrypto`?

packages/webvh/src/dids/didWebvhUtil.ts
```diff
@@ -0,0 +1,119 @@
...
+    throw new Error('Not implemented')
+  }
+
+  private async isKmsAvailable(): Promise<boolean> {
+    try {
+      return !!(this.agentContext.dependencyManager.resolve(Kms.KeyManagementApi))
+    } catch {
+      return false
+    }
+  }
```
TimoGlastra: Kms will always be available, so i think this check is redundant

packages/webvh/src/dids/didWebvhUtil.ts
```diff
@@ -0,0 +1,119 @@
...
+      })
+
+      return verificationResult.verified
+    } catch (error) {
+      this.agentContext.config.logger.error('KMS verification failed:', error)
+      return false
+    }
+  }
+
+  private async verifyWithLegacyMethod(signature: Uint8Array, message: Uint8Array, publicKey: Uint8Array): Promise<boolean> {
```
TimoGlastra: This should be removed, if you want this to work with 0.5.x as well, you can make another PR to the `0.5.x` branch based on the old KMS

packages/webvh/src/dids/didWebvhUtil.ts
```diff
@@ -0,0 +1,119 @@
...
+      this.agentContext.config.logger.error('Legacy verification failed:', error)
+      return false
+    }
+  }
+
+  public async verify(signature: Uint8Array, message: Uint8Array, publicKey: Uint8Array): Promise<boolean> {
+    try {
+      if (!this.agentContext) {
+        throw new Error('Agent context is required')
+      }
```
TimoGlastra: Agent context is required in the contructor, is there a specific reason for adding this check?

packages/webvh/src/dids/didWebvhUtil.ts
```diff
@@ -0,0 +1,119 @@
...
+        return await this.verifyWithKms(signature, message, publicKey)
+      } else {
+        this.agentContext.config.logger.debug('KMS not available, using legacy verification method')
+        return await this.verifyWithLegacyMethod(signature, message, publicKey)
+      }
+    } catch (error) {
+      // Log error in a non-production environment
+      if (process.env.NODE_ENV !== 'production') {
+        this.agentContext.config.logger.error('Error verifying signature:', error)
+      }
```
TimoGlastra:
```suggestion
        this.agentContext.config.logger.error('Error verifying signature:', error)
```

packages/webvh/src/dids/didWebvhUtil.ts
```diff
@@ -0,0 +1,119 @@
...
+      return false
+    }
+  }
+}
+
+export interface IDidDocOptions {
```
TimoGlastra: Where is this interface used? We generally don't use `I` prefix for interfaces, and also write out the full name, so `DidDocumentOptions`, but might be good to make it more specific if it's related to WebVh.

packages/webvh/tests/setup.ts
```diff
@@ -0,0 +1,32 @@
+import { DidDocument, DidDocumentService, VerificationMethod } from '@credo-ts/core'
```
TimoGlastra: Setup is usually run before tests by JEst, but this file does not do any setup, i think it's more a test util. Maybe we can rename it, and then remove it from the jest.config?

packages/webvh/tests/webvh-did-resolver.test.ts
```diff
@@ -0,0 +1,75 @@
...
+  beforeAll(async () => {
+    const agentOptions = getAgentOptions('WebVH DID Resolver Test', {}, {}, getWebvhModules())
+    agent = new Agent({...agentOptions, modules: { ...getWebvhModules(), inMemory: new InMemoryWalletModule() }})
+
+    const initPromise = agent.initialize()
+    const timeoutPromise = new Promise((_, reject) =>
+      setTimeout(() => reject(new Error('Agent initialization timed out')), 5000)
+    )
+
+    await Promise.race([initPromise, timeoutPromise])
```
TimoGlastra: Any reason for this approach? Since you mock the resolve call, i don't see any reason why initialize would time out?

packages/webvh/tests/webvh-did-resolver.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    try {
+      const didResolutionResult = await agent.dids.resolve(validDid)
+      res = didResolutionResult
+    } catch (e) {
+      err = e
+    }
+
+    expect(err).toBeUndefined()
+    expect(res?.didDocument).toBeDefined()
+    expect(res?.didDocument?.id).toBe('did:webvh:QmdmPkUdYzbr9txmx8gM2rsHPgr5L6m3gHjJGAf4vUFoGE:domain.example')
```
TimoGlastra:
```suggestion
      const didResolutionResult = await agent.dids.resolve(validDid)
    expect(didResolutionResult.didDocument).toBeDefined()
 expect(didResolutionResult.didDocument?.id).toBe('did:webvh:QmdmPkUdYzbr9txmx8gM2rsHPgr5L6m3gHjJGAf4vUFoGE:domain.example')
```

I think this tests exactly the same?

packages/webvh/tsconfig.build.json
```diff
@@ -0,0 +1,13 @@
+{
+  "extends": "../../tsconfig.build.json",
+  "compilerOptions": {
+    "outDir": "./build",
+    "baseUrl": ".",
+    "skipDefaultLibCheck": true,
+    "paths": {
+      "@credo-ts/core": ["../core/build"],
+      "@credo-ts/core/*": ["../core/build/*"]
+    }
```
TimoGlastra: this should not be needed

packages/webvh/src/anoncreds/services/__tests__/mock-resources.ts
```diff
@@ -0,0 +1,100 @@
...
+    type: 'DataIntegrityProof',
+    cryptosuite: 'eddsa-jcs-2022',
+    proofPurpose: 'assertionMethod',
+    proofValue: 'z2P3cR46Qt8r3Zc47EHxou6JkU7sTWd8PFZpQwZR7k8MU2Sntt325dQ1u3bn2ZoQUvoCFdNHWYbmsf5FzpqaF1n41',
+    verificationMethod:
+      'did:webvh:QmRxso8yoATm66gKhp3AKbPSH6ys4XcNVgKT786M99JRpN:id.anoncreds.vc:demo:863862bf-cd3b-44e3-89d4-0a2d7f5cc8d4#key-01',
+  },
+}
+
+export const MockRevRegDefResource = {
```
TimoGlastra: Just for consistency. We only use PascalCase for types/classes usually (same for other mocks in this file)

```suggestion
export const mockRevRegDefResource = {
```

packages/webvh/src/anoncreds/services/__tests__/WebVhAnonCredsRegistry.test.ts
```diff
@@ -0,0 +1,681 @@
...
+        return mockDidsApi
+      }
+      if (tokenString.includes('KeyManagementApi')) {
+        return mockKeyManagementApi
+      }
+      if (token === WebvhDidResolver || tokenString.includes('WebvhDidResolver')) {
+        return { resolveResource: mockResolveResource }
+      }
+      return originalResolve(token)
+    })
```
TimoGlastra: I think something like this would also work:

```suggestion
    agentContext = getAgentContext({ 
      agentConfig,
      registerInstances: [
        [DidsApi, mockDidsApi],
        [KeyManagementApi, mockKeyManagementApi],
        [WebvhDidResolver, { resolveResource: mockResolveResource }]
      ]
      })
```

Not sure about this line:
```
    // Register the mocked resolver instance
    agentContext.dependencyManager.registerInstance(WebvhDidResolver, new WebvhDidResolver())
```

As afterwards you overwrite it here:
```
      if (token === WebvhDidResolver || tokenString.includes('WebvhDidResolver')) {
        return { resolveResource: mockResolveResource }
      }
```

You can update the entry in `registerInstances` for `WebvhDidResolver` with the value you like.

packages/webvh/src/anoncreds/services/WebVhAnonCredsRegistry.ts
```diff
@@ -0,0 +1,568 @@
...
+      if (!isVerified) {
+        agentContext.config.logger.error('Signature verification failed')
+        return false
+      }
+
+      agentContext.config.logger.debug('DataIntegrityProof verification successful')
+      return true
+
+    } catch (error) {
+      agentContext.config.logger.error('Error during proof validation', {
```
TimoGlastra: For all error messages here it would be useful to be more specific. If you see this in the context of the full credo agent, it can be hard to understand where a message originates from

```suggestion
      agentContext.config.logger.error('Error during did:webvh resource proof validation', {
```

packages/webvh/src/dids/WebvhDidResolver.ts
brianorwhatever: That was temporary and I forgot to update. true seems fine to me will set them to that.

packages/webvh/src/dids/didWebvhUtil.ts
brianorwhatever: looks like it wasn't being used at all anymore. removing

packages/webvh/src/dids/didWebvhUtil.ts
brianorwhatever: Can you point me to an example of the old KMS? I couldn't find any. This was specifically not working in BC Wallet which is on 0.5.13.

Is `main` branch for version 0.6.0?

packages/webvh/src/anoncreds/services/__tests__/WebVhAnonCredsRegistry.test.ts
brianorwhatever: worked great. thanks!

brianorwhatever:
Thank you @TimoGlastra I believe I addressed all of your comments other than the legacy KMS stuff. Need a bit more help getting that plugged in properly

packages/webvh/src/dids/didWebvhUtil.ts
TimoGlastra: Yes the main branch is for 0.6.0. 

If we want this to be available in both i think the best method is to:
- remove all legacy KMs stuff from this PR
- merge into main
- create a PR to 0.5.x branch, with changes for the legacy KMS

But when targeting the main branch, we should only use the new KMS API

packages/webvh/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "clean": "rimraf -rf ./build",
+    "compile": "tsc -p tsconfig.build.json",
+    "prepublishOnly": "pnpm run build",
+    "test": "jest",
+    "test:debug": "node --inspect-brk node_modules/.bin/jest --runInBand --config jest.config.js"
+  },
+  "dependencies": {
+    "@credo-ts/anoncreds": "workspace:*",
+    "@credo-ts/core": "workspace:*",
+    "@multiformats/base-x": "^4.0.1",
```
TimoGlastra: I think this can be removed now?

packages/webvh/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "@stablelib/ed25519": "^2.0.2",
```
TimoGlastra: This one also

For the legacy KMS, we also supported Ed25519 ,so it should not be needed to use this (and we prefer doing everything through the pluggable crypto interface)

packages/webvh/src/anoncreds/services/WebVhAnonCredsRegistry.ts
```diff
@@ -0,0 +1,552 @@
...
+        if (error.message.includes('parse resource data')) errorType = 'invalidJson'
+        else if (error.message.includes('resolve') || error.message.includes('missing data')) errorType = 'notFound'
+        else if (error.message.includes('attested')) errorType = 'invalid'
+        else if (error.message.includes('proof')) errorType = 'invalid'
+        else if (error.message.includes('hash mismatch')) errorType = 'invalid'
+        else if (error.message.includes('not a valid schema')) errorType = 'invalid'
+        else if (error.message.includes('Invalid schema ID')) errorType = 'invalidDid'
+        else if (error.message.includes('Issuer ID mismatch')) errorType = 'invalid'
+        else if (error.message.includes('missing a valid issuerId')) errorType = 'invalid'
+      }
```
TimoGlastra: Where are these error messsages constructed? We don't use error messages anywhere to handle errors as it's very brittle. If you want to catch these cases, best i think is to create and throw specific errors that you can check with instanceof

packages/webvh/src/dids/didWebvhUtil.ts
brianorwhatever: Perfect, that makes sense. I'm in the process of removing the legacy stuff and I think I have an idea of how to get 0.5 working. Just created a PR for didwebvh-ts that removes the requirement of passing a VM into the constructor so once that lands I'll have this PR in great shape

brianorwhatever:
@TimoGlastra I think this is ready to be looked at again now. Working on a 0.5.x version now

brianorwhatever:
legacy version created in #2339

swcurran:
@genaris @TimoGlastra @cvarjao @berendsliedrecht -- can someone please review this and the #2339 PR?  We would really like to get this into Bifold and beyond.

packages/webvh/src/dids/WebvhDidResolver.ts
```diff
@@ -0,0 +1,128 @@
...
```
genaris:
```suggestion
      const response = await agentContext.config.agentDependencies.fetch(httpsUrl)
```

packages/webvh/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "@credo-ts/webvh",
+  "version": "0.6.0",
```
genaris:
```suggestion
  "version": "0.5.13",
```

Probably it is not a problem, but just to make it consistent with other packages.

genaris (approved):
Great job here! LGTM

Looking forward to work also in the DID and resource creation/update!

genaris:
@brianorwhatever are you able to address the two minor changes I've spotted in my review? I wanted to do it myself but I cannot commit to your branch because it is in an organization fork.

packages/webvh/package.json
brianorwhatever: This PR is for 0.6.0 while #2339 is for 0.5.x

packages/webvh/src/dids/WebvhDidResolver.ts
brianorwhatever: resolved this in both PRs

brianorwhatever:
@genaris I have updated both this and #2339 with the fetch dependency from agent context

packages/webvh/package.json
genaris: I understand, but all other modules in main branch do have this version. Please update it to 0.5.13 so all of them are in sync.

brianorwhatever: ah ok I understand now. Done!

genaris:
> @genaris I have updated both this and #2339 with the fetch dependency from agent context

Thanks! It looks like formatting is failing: it should be enough to just run `pnpm style:fix` and commit the changes. Hopefully this is the last piece before merging!

genaris:
@brianorwhatever I've created a PR in your branch to fix some remaining issues regarding the types: https://github.com/aviarytech/credo-ts/pull/2

brianorwhatever:
thanks @genaris!

Issue #2206, opened 2025-02-24
Problem receiving revocation notification


barnabef:
Here is the setup.  We are using Hyperledger Acapy version 1.2.0 with askar-anoncred as our controller.  We are using credo-ts v 0.5.13 as wallet holder.

The flow is quite simple.  The Acapy agent sends a credential offer.  The user accepts the credential in Credo.  Then Acapy revokes the credential it sends a notification to the Credo holder, but there's this error when trying to get it from the storage:
```
INFO: Processing revocation notification v2 {
2025-02-24T16:04:29.552264300Z   "message": {
2025-02-24T16:04:29.552282690Z     "@type": "https://didcomm.org/revocation_notification/2.0/revoke",
2025-02-24T16:04:29.552291303Z     "@id": "c3256534-d68f-42e2-ad22-cb0c7149b8ca",
2025-02-24T16:04:29.552295793Z     "revocation_format": "indy-anoncreds",
2025-02-24T16:04:29.552299852Z     "credential_id": "Uzy1MJK1dQ6eoX265pUTLx:4:Uzy1MJK1dQ6eoX265pUTLx:3:CL:19:default:CL_ACCUM:0::3",
2025-02-24T16:04:29.552304001Z     "comment": "Revocation"
2025-02-24T16:04:29.552334319Z   }
2025-02-24T16:04:29.552339540Z }
2025-02-24T16:04:29.558320243Z TRACE: Getting record by query for revocation notification: {
2025-02-24T16:04:29.558432306Z   "$or": [
2025-02-24T16:04:29.558456042Z     {
2025-02-24T16:04:29.558464475Z       "anonCredsRevocationRegistryId": "Uzy1MJK1dQ6eoX265pUTLx:4:Uzy1MJK1dQ6eoX265pUTLx:3:CL:19:default:CL_ACCUM:0",
2025-02-24T16:04:29.558469585Z       "anonCredsCredentialRevocationId": "3",
2025-02-24T16:04:29.558473657Z       "connectionId": "dbc36568-c1b5-4d67-9049-68b73b861d1a"
2025-02-24T16:04:29.558477991Z     },
2025-02-24T16:04:29.558481875Z     {
2025-02-24T16:04:29.558485384Z       "anonCredsUnqualifiedRevocationRegistryId": "Uzy1MJK1dQ6eoX265pUTLx:4:Uzy1MJK1dQ6eoX265pUTLx:3:CL:19:default:CL_ACCUM:0",
[...]

PR #2240, opened 2025-03-27
fix: unqualified anoncreds revocation notification support

genaris:
This is a fix focused mainly for 0.5.x deployments to address the issue #2206. 

The proposed solution here is to search for W3C Credential Records matching the revocation notification only in case we don't find any matching Credential Exchange record. The opposite (search first for W3C Credential records) would work, but force us to get all Credential Exchange records and filter them (i.e. no optimized query using tags), since `RevocationNotificationReceivedEvent` requires to send the associated CredentialExchangeRecord.

I think for 0.6 we could aim to update this API to focus on W3CCredentialRecord directly, meaning that we will need to do some updates to the data model (e.g. W3CCredentialRecord will need to hold revocation info and connectionId).

@jJoLuPuma @jamshale @barnabef @zoblazo I didn't test it deeply yet but, if you have the chance of patching your 0.5.13 with this and test it, please let me know to make sure this does the trick.


genaris:
This is how the changes look like right now for a 0.5.13 credo: https://github.com/openwallet-foundation/credo-ts/compare/v0.5.x...genaris:fix/v0.5x-unqualified-anoncreds-revocation-support?expand=1

packages/didcomm/src/modules/credentials/protocol/revocation-notification/services/RevocationNotificationService.ts
```diff
@@ -64,25 +72,61 @@ export class RevocationNotificationService {
...
+        'Credential Exchange Record not found. Getting W3C credential record by query for revocation notification:',
+        w3cCredentialQuery
+      )
+
+      const w3cCredentialRepository = agentContext.dependencyManager.resolve(W3cCredentialRepository)
+      const w3cCredentialRecord = await w3cCredentialRepository.getSingleByQuery(agentContext, w3cCredentialQuery)
+
+      // Find credential exchange record associated with this credential
+      credentialExchangeRecord =
+        (await this.credentialRepository.getAll(agentContext)).find((record) =>
```
TimoGlastra, line 101: Is there no way to find the credential exchange record associated with a credential record?

This seems inefficient to run for every query ... :/

genaris: Yes, that's why I tried to avoid running it in cases not affected by this problem (e.g. using qualified identifiers).

Probably we have some other tricks to do, but I don't think in practice it will be so critical, since this code runs on holder side and I don't expect users to hold thousands of credentials using unqualified identifiers.

Something not done in this PR (at the moment) are changes on `AnonCredsCredentialFormatService`: we can also add the `anonCredsUnqualified*` tags to CredentialExchangeRecords. This would at least prevent this code from running in 0.5.14 agents.

jamshale:
@bryce-mcmath Were you able to test this? I was wondering if this could still get merged into the patch as it is the last thing allowing this scenario to work and do our migrations incrementally. 

I don't have a good understanding of credo storage, but, I agree that the performance doesn't seem to be much of an issue if it only does this when unqualified id's are present and is only on the holder side.

bryce-mcmath:
> @bryce-mcmath Were you able to test this? I was wondering if this could still get merged into the patch as it is the last thing allowing this scenario to work and do our migrations incrementally.
> 
> I don't have a good understanding of credo storage, but, I agree that the performance doesn't seem to be much of an issue if it only does this when unqualified id's are present and is only on the holder side.

It's already in our patches in the Testflight BC Wallet if you want to double check anything in particular. Seems to be working well so far

jamshale:
OK, great. Thank you.

genaris:
@TimoGlastra since this fix can be useful for some users and appearently works well for their use case, do you think we can re-target this PR to 0.5.x branch so it can be included in the next patch release for them?

I can create a different one for `main`, refactoring the revocation notification support to use W3CCredentialRecord directly. Then we'll add a migration script to upgrade everything in 0.6.

packages/didcomm/src/modules/credentials/protocol/revocation-notification/services/RevocationNotificationService.ts
TimoGlastra: I think we should limit the calling of this to:
- filter by state (credential received and completed)
- only if the incoming revocation id is unqualified

genaris: Took some time but now it's filtered by state and executed only in case of unqualified revocation registry ids

PR #2246, opened 2025-04-03
chore: update ec-compression library

berendsliedrecht:


TimoGlastra (approved):
Should we get an 0.1.0 release for ec-compression? I think would be good to have a stable release

berendsliedrecht:
> Should we get an 0.1.0 release for ec-compression? I think would be good to have a stable release

Yeah I think that is fine

PR #2247, opened 2025-04-03
fix: assign proper rev_reg_id

bryce-mcmath:
Tiny fix for a bug we found while testing this patch


bryce-mcmath:
@genaris @TimoGlastra I don't know if you're tagged already but in case you aren't ^

TimoGlastra (approved):

TimoGlastra:
I think this PR should also go to main branch? Can you do that @bryce-mcmath?

PR #2250, opened 2025-04-07
fix(cheqd): Populate contexts for JsonLD issuance

DaevMithran:
we are facing an issue in JSONLD issuance with credo, and issue is with the default document loader
```
DidDocument {
	context: ['https://www.w3.org/ns/did/v1', 'https://w3id.org/did/v1', 'https://w3id.org/security/suites/jws-2020/v1', 'https://w3id.org/security/suites/ed25519-2018/v1'],
	id: 'did:cheqd:testnet:151734ba-d9c8-4bf6-a83c-bc4098b10ac6',
	controller: [],
	verificationMethod: [ VerificationMethod {
		id: 'did:cheqd:testnet:151734ba-d9c8-4bf6-a83c-bc4098b10ac6#z6MktzU5kM5XdEKmGrZwXfQEa3jZAyexhmP7iNDMotin4GcR',
		type: 'Ed25519VerificationKey2018',
		controller: 'did:cheqd:testnet:151734ba-d9c8-4bf6-a83c-bc4098b10ac6',
		publicKeyBase58: 'FYD3A6q6HgqJAMjEr6SPixBZMQP7Ht8m2MJRyckm93q3'
	}],
	authentication: ['did:cheqd:testnet:151734ba-d9c8-4bf6-a83c-bc4098b10ac6#z6MktzU5kM5XdEKmGrZwXfQEa3jZAyexhmP7iNDMotin4GcR'],
	assertionMethod: ['did:cheqd:testnet:151734ba-d9c8-4bf6-a83c-bc4098b10ac6#z6MktzU5kM5XdEKmGrZwXfQEa3jZAyexhmP7iNDMotin4GcR']
}
```

This is the input for jsond.frame in [default document loader](https://github.com/openwallet-foundation/credo-ts/blob/e7b59000502bb893a26e12200ea0f701237b972e/packages/core/src/modules/vc/data-integrity/libraries/documentLoader.ts#L43)

The result is missing the publicKeyBase58, same is the case for other signature suites.
```
{
	'@context': ['https://w3id.org/did/v1', 'https://w3id.org/security/suites/jws-2020/v1'],
	id: 'did:cheqd:testnet:efa002a1-ea59-4a87-9fe9-1be2919886d7#z6MkqzAn7GL2srNkmFN8iPXfPKCHfHvY5eTeRUFTeN1xYLsa',
	type: 'Ed25519VerificationKey2018',
	controller: 'did:cheqd:testnet:efa002a1-ea59-4a87-9fe9-1be2919886d7'
}
```

### DIFF
--- a/packages/cheqd/src/dids/CheqdDidRegistrar.ts
+++ b/packages/cheqd/src/dids/CheqdDidRegistrar.ts
@@ -44,6 +44,11 @@ import {
 
 export class CheqdDidRegistrar implements DidRegistrar {
   public readonly supportedMethods = ['cheqd']
+  private contextMapping = {
+    Ed25519VerificationKey2018: ED25519_SUITE_CONTEXT_URL_2018,
+    Ed25519VerificationKey2020: ED25519_SUITE_CONTEXT_URL_2020,
+    JsonWebKey2020: SECURITY_JWS_CONTEXT_URL,
+  }
 
   public async create(agentContext: AgentContext, options: CheqdDidCreateOptions): Promise<DidCreateResult> {
     const didRepository = agentContext.dependencyManager.resolve(DidRepository)
@@ -106,12 +111,6 @@ export class CheqdDidRegistrar implements DidRegistrar {
         }
       }
 
-      const contextMapping = {
-        Ed25519VerificationKey2018: ED25519_SUITE_CONTEXT_URL_2018,
-        Ed25519VerificationKey2020: ED25519_SUITE_CONTEXT_URL_2020,
-        JsonWebKey2020: SECURITY_JWS_CONTEXT_URL,
-      }
-
       // Normalize context to an array
       const contextSet = new Set<string>(
         typeof didDocument.context === 'string'
@@ -122,7 +121,7 @@ export class CheqdDidRegistrar implements DidRegistrar {
       )
 
       for (const verificationMethod of didDocument.verificationMethod || []) {
-        const contextUrl = contextMapping[verificationMethod.type as keyof typeof contextMapping]
+        const contextUrl = this.contextMapping[verificationMethod.type as keyof typeof this.contextMapping]
         if (contextUrl) {
           contextSet.add(contextUrl)
         }
@@ -131,7 +130,7 @@ export class CheqdDidRegistrar implements DidRegistrar {
       // Add the context to the did document
       // NOTE: cheqd sdk uses https://www.w3.org/ns/did/v1 while Credo did doc uses https://w3id.org/did/v1
       // We should align these at some point. For now we just return a consistent value.
-      didDocument.context = Array.from(contextSet.add('https://www.w3.org/ns/did/v1'))
+      didDocument.context = Array.from(contextSet.add('https://w3id.org/did/v1'))
 
       const didDocumentJson = didDocument.toJSON() as DIDDocument
 
@@ -198,6 +197,14 @@ export class CheqdDidRegistrar implements DidRegistrar {
             },
           }
         }
+        // Normalize existing context to an array
+        const contextSet = new Set<string>(
+          typeof didDocument.context === 'string'
+            ? [didDocument.context]
+            : Array.isArray(didDocument.context)
+            ? didDocument.context
+            : []
+        )
 
         if (verificationMethod) {
           const privateKey = verificationMethod.privateKey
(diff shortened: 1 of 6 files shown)


TimoGlastra:
So it seems that the document that is passed to the frame method only contains the following context:

```json
"@context": [
        "https://w3id.org/did/v1",
        "https://w3id.org/security/suites/jws-2020/v1"
],
```

And thus it's missing the required `https://w3id.org/security/suites/ed25519-2018/v1` context (which is present in the input you posted above).

On uniresolver (https://dev.uniresolver.io/) it does return the correct ed25519 context.

It seems that we directly use the document received from the SDK, as I can't find where we set the @context on the document when we resolve a cheqd did

tweeddalex:
Thanks @TimoGlastra we're now working on a fix in our SDK to align the `context` handling with how our DID Resolver works!

DaevMithran:
@TimoGlastra the resolveDid in cheqd populates the context as expected. But we noticed the contexts were missing in the didRecord saved within the wallet. For issuance shouldn't the flow resolve the DID instead of relying on the didRecord locally for the DIDDocument

TimoGlastra:
@DaevMithran for now we assume all updates are made through credo and thus the local record should match. It's very inefficient to resolve the did record every time you want to issue. We could do a periodic fetch, but I'd rather rely on the local record. I think we should fix the inconsistency between the cheqd ledger record and the local did record

packages/cheqd/src/dids/CheqdDidRegistrar.ts
```diff
@@ -25,8 +25,14 @@ import {
   getKeyFromVerificationMethod,
   JsonTransformer,
   VerificationMethod,
+  SECURITY_JWS_CONTEXT_URL,
 } from '@credo-ts/core'
 
+import {
+  ED25519_SUITE_CONTEXT_URL_2018,
+  ED25519_SUITE_CONTEXT_URL_2020,
+} from '../../../core/src/modules/vc/data-integrity/signature-suites/ed25519/constants'
```
TimoGlastra, line 34: this won't work on publish, as it's importing from another package

sownak: Thanks Timo.. this is updated and I have merged latest 0.5.x branch as well

DaevMithran: @TimoGlastra Should we create a separate PR targeting main branch too?

TimoGlastra: Yes that'd be great @DaevMithran

packages/cheqd/src/dids/CheqdDidRegistrar.ts
```diff
@@ -178,6 +196,14 @@ export class CheqdDidRegistrar implements DidRegistrar {
             },
           }
         }
+        // Normalize existing context to an array
```
DaevMithran, line 199: @sownak Don't think this is this needed?

packages/cheqd/src/dids/CheqdDidRegistrar.ts
```diff
@@ -196,6 +222,12 @@ export class CheqdDidRegistrar implements DidRegistrar {
             keyType: KeyType.Ed25519,
             privateKey: privateKey,
           })
+          const contextUrl = this.contextMapping[verificationMethod.type as keyof typeof this.contextMapping]
```
DaevMithran, line 225: This should be moved below maybe after the update operation. You can see there are some additions to the verificationMethod in the below step.

packages/cheqd/src/dids/CheqdDidRegistrar.ts
DaevMithran, line 199: I think it's better to initialize an empty set as there can be some verificationMethods removed. We can compute the entire context after the update

packages/cheqd/src/dids/CheqdDidRegistrar.ts
sownak: Agreed, this is done.

packages/cheqd/src/dids/CheqdDidRegistrar.ts
sownak: `didDocument = options.didDocument` in L186 so the current value of didDocument is from the inputs, so I would keep that same

TimoGlastra:
For 0.5.x I'd like to overwrite the context to match what cheqd is using (in cheqd module), but not change the default, since we don't want to push such a breaking change in the 0.5.x release channel (but rather do that in 0.6). Would it be possible to revert the changing of the default context url for the 0.5.x branch (I accepted the change for the 0.6 PR). 

I see you already added a changset entry here 👍

sownak:
@TimoGlastra The default context changes have been reverted.

PR #2252, opened 2025-04-09
chore: add missing test script in package-core

GHkrishna:
## What
- Add missing `test` script in core package


TimoGlastra (commented):
Usually we run tests from root. So not all packages have this script. Did you run into issues?

GHkrishna:
No issues as such.
Actually I wanted to run tests in core only. However, I was not able to even though it had required test files.
All the other packages has the test script apart from `core`, so this is mostly for consistency reasons.

TimoGlastra (approved):

PR #2278, opened 2025-05-09
docs(changeset): Intialize feemarket module in cheqd sdk

DaevMithran:


.changeset/many-sheep-fix.md
```diff
@@ -0,0 +1,5 @@
+---
+'@credo-ts/cheqd': minor
```
TimoGlastra, line 2: Should be patch
```suggestion
'@credo-ts/cheqd': patch
```

DaevMithran: Fixed it

PR #2288, opened 2025-05-16
feat: redis cache

berendsliedrecht:


packages/core/src/modules/cache/CachedStorageService.ts
```diff
@@ -0,0 +1,77 @@
...
+  private cache(agentContext: AgentContext) {
+    return agentContext.resolve(CacheModuleConfig).cache
+  }
+
+  private getCacheKey(options: { type: string; id: string }) {
+    return `${options.type}:${options.id}`
+  }
+
+  public async save(agentContext: AgentContext, record: T): Promise<void> {
+    if (record.useCache) {
```
TimoGlastra, line 19: I would call it `allowCache`, because it's only using cache if the use cached storage serivice is enabled

packages/core/src/modules/cache/CachedStorageService.ts
```diff
@@ -0,0 +1,77 @@
...
+      await this.cache(agentContext).remove(agentContext, this.getCacheKey({ ...recordClass, id }))
+    }
+    return await this.storageService.deleteById(agentContext, recordClass, id)
+  }
+
+  public async getById(agentContext: AgentContext, recordClass: BaseRecordConstructor<T>, id: string): Promise<T> {
+    if (recordClass.useCache) {
+      const cachedValue = await this.cache(agentContext).get<T>(agentContext, `${recordClass.type}:${id}`)
+
+      // TODO: class transform
```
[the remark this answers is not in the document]
TimoGlastra: can you add some minimal tests, so we know at least the base flow works?

packages/redis-cache-nodejs/src/RedisCache.ts
```diff
@@ -0,0 +1,99 @@
...
+    }
+  }
+
+  public async remove(agentContext: AgentContext, key: string): Promise<void> {
+    const client = await this.client()
+    const namespacedKey = this.getNamespacedKey(agentContext, key)
+    await client.del(namespacedKey)
+  }
+
+  public async destroy(agentContext: AgentContext) {
```
TimoGlastra, line 73: Is this part of the cache interface? or for testing?

TimoGlastra: Becuase it won't work for multi-tenancy cases i think?

berendsliedrecht: `destroy` might be a bad name, but disconnects and removes all the tenants key/values.

packages/core/src/modules/cache/CachedStorageService.ts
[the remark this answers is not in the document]
berendsliedrecht: Yeah, I will.

packages/redis-cache-nodejs/src/RedisCache.ts
```diff
@@ -0,0 +1,99 @@
...
+    const client = await this.client()
+    let cursor = '0'
+
+    do {
+      const [nextCursor, keys] = await client.scan(
+        cursor,
+        'MATCH',
+        `${agentContext.contextCorrelationId}:*`,
+        'COUNT',
+        '100'
```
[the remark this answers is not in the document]
berendsliedrecht: limit size

packages/redis-cache-nodejs/src/RedisCache.ts
TimoGlastra: but if it also disconnects, that impacts more than just the tenant? There could be other processes currently doing a query that will then suddenly be aborted?

packages/redis-cache-nodejs/src/RedisCache.ts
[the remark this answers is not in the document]
TimoGlastra: So what if there's more than 100 entries?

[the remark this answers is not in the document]
TimoGlastra: If we want to add this to the base class (and it's not for test utils) this is not a complete approach

packages/redis-cache-nodejs/src/RedisCache.ts
berendsliedrecht: The client checks with a ping if it is still alive. It's less efficient, though.  

I can make it delete the records, but not disconnect,

TimoGlastra: Yes but that's before the query starts. I'm not sure what the behviour is if you call disconnect in the middile of a query while waiting for a response?

I want to avoid this:

- Start destroy
- start remove keys command
- Start request 1
- Request 1 pings
- Request 1 send requests, waiting for response
- remove keys comlete, destory session
- Error because session closed while request 1 was processing

TimoGlastra: But not disconnecting will solve it

TimoGlastra (approved):
addressed feedback

PR #2289, opened 2025-05-16
feat: native SQLite and Postgres database support based on Drizzle ORM

TimoGlastra:
Still WIP.

Will have no breaking API changes and thus will be a drop-in replacement of Askar (not considering the migration from Askar to the new DB).

Can add support for other database types in the future as well, given we define the table structure for all records


TimoGlastra:
@genaris @berendsliedrecht I think I have most of the foundation/scaffolding in place, and now most of the remaining work will be in adding the missing records for sqlite/postgres.

Maybe you can take a look at the current usage documentation and let me know what you think of the process: https://github.com/TimoGlastra/aries-framework-javascript/blob/feat/drizzle-database/packages/drizzle-storage/README.md

To get a feel what writing a drizzle integration for a custom module looks like, you can look at e.g. the action-menu one: https://github.com/TimoGlastra/aries-framework-javascript/tree/feat/drizzle-database/packages/drizzle-storage/src/action-menu

Basically you define:
- for each record
	- record adapter (bridge between Credo `Record` class and the Drizzle world
	- sqlite schema
	- postgres schema
- meta information such as the bundle `name` (.e.g `core` or `anoncreds` or `my-custom-module`), along with migration information.

During development you can generate migrations using the CLI, and then optinoally modify them. Then in your project, you need to apply migrations based on the modules you use. This allows us to move a lot of the manual TypeScript based migrations to SQL. We should still only push breaking migrations during breaking releases ,and should probably start adopting the multi-deployments breaking chanegs. (so introduce a new api in 0.6.x, and support the old and new database model in 0.6.x), then in 0.7 we only support the new database model. This will allow seemless migrations in cloud envrionments. For React Native you can see in the docs how the migrations will work based on drizzle.


This PR is dependant on #2317 since it does not work with Node 18

packages/drizzle-storage/README.md
```diff
@@ -0,0 +1,207 @@
...
+- The database records (models) to register.
+
+```ts
+import { didcommDrizzleRecords } from "@credo-ts/drizzle-storage/didcomm";
+
+const agent = new Agent({
+  modules: {
+    drizzleStorage: new DrizzleStorageModule({
+      database,
+      records,
```
berendsliedrecht: Started with reviewing the README, but is the `records` the import that is not used, `didcommDrizzleRecords` or is it a list of records that extend BaseRecord? Maybe a short short description of where the records come from.

.gitignore
```diff
@@ -11,4 +11,6 @@ coverage
 logs.txt
 logs/
 .pnpm-store
-ngrok.auth.yml
\ No newline at end of file
+ngrok.auth.yml
+cli-build
+./packages/drizzle-storage/local.db
```
berendsliedrecht: Can't we glob ignore just `*.db`?

package.json
```diff
@@ -16,9 +16,9 @@
     "style:fix": "biome check --write --unsafe",
     "clean": "pnpm -r --parallel run clean",
     "build": "pnpm -r --parallel run build",
-    "test:unit": "jest --testPathIgnorePatterns 'e2e.test.ts$'",
-    "test:e2e": "jest --testMatch '**/?(*.)e2e.test.ts'",
-    "test": "jest",
+    "test:unit": "NODE_OPTIONS=--experimental-vm-modules jest --testPathIgnorePatterns 'e2e.test.ts$'",
```
berendsliedrecht: What does this add?

package.json
```diff
@@ -50,7 +50,23 @@
     "tsyringe": "^4.8.0",
     "typescript": "~5.5.2",
     "undici": "^6.20.1",
-    "ws": "^8.13.0"
+    "ws": "^8.13.0",
+    "@credo-ts/core": "workspace:*",
```
berendsliedrecht: Why are they all added to the root package.json?

packages/drizzle-storage/README.md
```diff
@@ -0,0 +1,207 @@
...
+- `dialect` - the dialect to use for migrations. Either `postgres` or `sqlite` and MUST match with the database being used during runtime in your Credo agent.
+- `database-url` - the url to connect to the database. You can also provide this using the `DRIZZLE_DATABASE_URL` environment variable (recommended).
+
+If the `@credo-ts/drizzle-storage` package is already installed in your project, you can directly use it to run the CLI. You can also use `npx` to execute the CLI, but make sure that you install the correct version matching the version that is used for your runtime agent. You also need to make sure that all extension modules are installed so the migration files can be resolved (which can be done using `--project` in `npx`).
+
+```sh
+# or npm or yarn
+pnpm @credo-ts/drizzle-storage@^0.x migrate --bundle core --bundle didcomm --bundle anoncreds --database-url postgresql://postgres:postgres@localhost:5432/postgres`
```
berendsliedrecht, line 133: Are we are 100% certain that the ordering of the CLI arguments is retained?

packages/drizzle-storage/cli/generate-migrations.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    for (const dialect of dialects) {
+      const dialectBundle = bundle.migrations[dialect]
+      const schemaFile = await resolveSchemaFile(dialectBundle.schemaModule)
+      const drizzleMigrationsFolder = getMigrationsDirectory(schemaFile, dialectBundle.migrationsPath)
+
+      const migrateResult = spawnSync(
+        'npx',
+        [
+          'ts-node',
```
berendsliedrecht, line 29: Can't we use `tsx`? I think it is a bit faster

berendsliedrecht:
I left some initial comments on some basic setup items, but mainly looking at the `action-menu` bundle I really like the approach. It seems quite easily reproducible for custom packages and not too much overhead. Maybe at some point the schemas for the db can me more tightly coupled to the records, maybe something like inferring a type based on the schema and comparing that type to the record type (because I don't think we can generate a record based on the schema, or at least that seems also like a mess in code to me).

But just so I understand it, it will not, currently, include the migration from askar to sqlite, but it will migrate the records from askar with credo 0.5.x to askar with credo 0.6.x? And it will be for new agents to not have to require askar anymore, right?

berendsliedrecht:
I left some initial comments on some basic setup items, but mainly looking at the `action-menu` bundle I really like the approach. It seems quite easily reproducible for custom packages and not too much overhead. Maybe at some point the schemas for the db can me more tightly coupled to the records, maybe something like inferring a type based on the schema and comparing that type to the record type (because I don't think we can generate a record based on the schema, or at least that seems also like a mess in code to me).

But just so I understand it, it will not, currently, include the migration from askar to sqlite, but it will migrate the records from askar with credo 0.5.x to askar with credo 0.6.x? And it will be for new agents to not have to require askar anymore, right?

packages/drizzle-storage/README.md
TimoGlastra: This was an outdated example, you provide it as `bundles` now

package.json
TimoGlastra: It is needed for the in-memory postgres database:

```
TypeError: A dynamic import callback was invoked without --experimental-vm-modules
    at importModuleDynamicallyCallback (node:internal/modules/esm/utils:263:11)
    at Sa (/Users/timo/Developer/Work/Projects/AFJ/aries-framework-javascript/node_modules/.pnpm/@electric-sql+pglite@0.3.1/node_modules/@electric-sql/pglite/dist/index.cjs:10:3986)
    at Et.Ha (/Users/timo/Developer/Work/Projects/AFJ/aries-framework-javascript/node_modules/.pnpm/@electric-sql+pglite@0.3.1/node_modules/@electric-sql/pglite/dist/index.cjs:25:73096)
    at processTicksAndRejections (node:internal/process/task_queues:95:5) {
  code: 'ERR_VM_DYNAMIC_IMPORT_CALLBACK_MISSING_FLAG'
}
```

it's just needed for the tests. Otherwise we need to use a real postgres database for tests

package.json
TimoGlastra: So we can get rid of using `paths` in `tsconfig.json`, as we now use `exports` map in package.json (which can have separate exports in  `publishConfig`). So the `exports` points to `src` and the `publishConfig.exports` points to `build`. 

But for this to work the package needs to be resolvable in normal node_modules way, and thus we need to add it as a dependency to root (for the tests to be able to resolve it).

I like the approach more since it relies on native node_module resolutions now rather than mapping it based on custom `paths` in tsconfig

packages/drizzle-storage/README.md
TimoGlastra: Yes

packages/drizzle-storage/cli/generate-migrations.ts
TimoGlastra: no, not until we get rid of using the legacy experimental decorators. See https://github.com/openwallet-foundation/credo-ts/issues/2320 and https://github.com/openwallet-foundation/credo-ts/issues/2157


We really need to invest some time into getting rid of it. Maybe now that we have the KMS and storage revamped, i think it should be one of the higher priority tickets, so we can get rid of some legacyc packages and move to zod. But it's a very heavy task

TimoGlastra:
> it will be for new agents to not have to require askar anymore, right?

yes

> it will not, currently, include the migration from askar to sqlite,

correct, i will probably create a separate repo for that which allows for a one-time migration.

package.json
TimoGlastra: updated it to use a real postgres database

PR #2290, opened 2025-05-16
fix(cheqd): Populate DID Document Contexts For JsonLD Issuance

DaevMithran:


packages/cheqd/tests/cheqd-did-resolver.e2e.test.ts
```diff
@@ -75,7 +75,11 @@ describe('Cheqd DID resolver', () => {
     })
     expect(JsonTransformer.toJSON(resolveResult)).toMatchObject({
       didDocument: {
-        '@context': ['https://www.w3.org/ns/did/v1', 'https://w3id.org/security/suites/jws-2020/v1'],
+        '@context': [          
+          'https://w3id.org/did/v1',
```
TimoGlastra, line 79: why do we now have two did contexts? I don't think this will work with a JSON ld processor?

TimoGlastra: Any reason cheqd uses the other one? The spec seems to mostly refer to the one used by credo.

Did you try it in a json-ld resolver with both contexts, does it cause issues?

We probably need a way to avoid this?

tweeddalex: Hey @TimoGlastra it seems that there might be an inconsistency between the `context` that Credo populates, and the `context` that is in the DID Core and DID Resolution specs.

Across the latter, it seems that "https://www.w3.org/ns/did/v1" is the default. Maybe this is a change that is needed on the Credo side?

tweeddalex: Tagging @berendsliedrecht here as well

tweeddalex: Hey @TimoGlastra @berendsliedrecht I confirmed with Markus Sabadello which is the correct `context` and he said that the "[https://www.w3.org/ns/did/v1](https://www.w3.org/ns/did/v1)" URL that the cheqd SDK populates is correct. 

This means that the auto-population within Credo should be updated to reflect this URL.

In the meantime, is there a workaround @DaevMithran @sownak - can we override the Credo `context` population or look to update the default `context` without breaking anything?

TimoGlastra (approved):
One minor comment, but otherwise LGTM. I would like @genaris to also review given the breaking change to the did document context. I don't think it should cause any issues with different did:peer:1 document hashes being generated?

For completeness can you add two changesets (you can add a changeset by running `pnpm changeset`)?
- one for the core package, a minor release explaining that the default context for did document has changed from xx to xx.
- one for the cheqd package, a patch released explaining the fix of this PR

packages/cheqd/src/dids/CheqdDidRegistrar.ts
```diff
@@ -182,6 +177,25 @@ export class CheqdDidRegistrar implements DidRegistrar {
...
+      // Normalize context to an array
+      const contextSet = new Set<string>(
+        typeof didDocument.context === 'string'
+          ? [didDocument.context]
+          : Array.isArray(didDocument.context)
+          ? didDocument.context
+          : []
+      )
+
+      for (const verificationMethod of didDocument.verificationMethod || []) {
```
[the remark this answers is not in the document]
tweeddalex: @sownak if you could make this change, that would be ideal

tweeddalex:
Hey @TimoGlastra, given the reviews and approvals - I think we should now be able to get this one merged 🚀

PR #2292, opened 2025-05-21
feat(didcomm): Queue Transport repository

genaris:
A follow-up from https://github.com/openwallet-foundation/credo-ts/pull/2168 where we found that it would be a good idea to move "Message Pickup Repository" to DIDComm main module, calling it "Queue Transport Repository" (to be consistent with naming used in [DIDComm spec](https://github.com/decentralized-identity/didcomm-messaging/blob/main/extensions/return_route/main.md#queue-transport). 

Good thing about this is that now we could get rid of `InjectionSymbols.MessagePickupRepository`.

Here we also add AgentContext to all methods of the newly defined QueueTransportRepository interface.


packages/didcomm/src/__tests__/MessageSender.test.ts
```diff
@@ -28,18 +27,18 @@ import { MessageSender } from '../MessageSender'
...
+import { DidCommModuleConfig } from '../DidCommModuleConfig'
 import { DummyTransportSession } from './stubs'
 
 jest.mock('../TransportService')
 jest.mock('../EnvelopeService')
 jest.mock('../../../core/src/modules/dids/DidsApi')
 jest.mock('../services/DidCommDocumentService')
 
-const logger = testLogger
+const _logger = testLogger
```
TimoGlastra, line 41: Still used?

packages/didcomm/src/transport/queue/QueueTransportRepository.ts
```diff
@@ -0,0 +1,18 @@
+import { AgentContext } from '@credo-ts/core'
+import type {
+  AddMessageOptions,
+  GetAvailableMessageCountOptions,
+  RemoveMessagesOptions,
+  TakeFromQueueOptions,
+} from './QueueTransportRepositoryOptions'
+import type { QueuedMessage } from './QueuedMessage'
+
+export interface QueueTransportRepository {
```
TimoGlastra, line 10: Should we take into account an agent that does not want to queue? Can we disable it instead of using in memory (which could be a memory leak)



also should we add some extra metadata to the queued messgages? So like a state (awaiting confirmation, or the date/time since we are awaiting confirmation) or do we see this more as a repository specific thing to ha e?

TimoGlastra (approved):

PR #2295, opened 2025-05-22
feat: create transport session repository

sairanjit:


TimoGlastra:
Discussed on the WG. Overall approach makes sense, but we need to think about how the current TransportSession (bound to Http/Ws) relates to a TransportSession we need for instance sending.

So the implementer of the transport session repository should also have control over the actual transport sessions.

Or it could be that we still have Ws/Http session, and the custom repository implementation can add other types as well for other instances of Credo.

TimoGlastra:
@genaris to take a look at the PR and give some ideas. Goal is to make it very open to different implementations / architectures. single vs multi instance. Or handling sending within agents or through an external server (e.g. socketdock).

PR #2303, opened 2025-05-27
feat: custom verification callback for the authorization request on the holder side for verification

berendsliedrecht:

### DIFF
--- a/packages/openid4vc/src/openid4vc-holder/OpenId4vpHolderService.ts
+++ b/packages/openid4vc/src/openid4vc-holder/OpenId4vpHolderService.ts
@@ -23,7 +23,6 @@ import {
   DifPresentationExchangeService,
   DifPresentationExchangeSubmissionLocation,
   Hasher,
-  JwsService,
   Kms,
   TypedArrayEncoder,
   injectable,
@@ -146,18 +145,15 @@ export class OpenId4VpHolderService {
     if (options?.verifyAuthorizationRequestCallback) {
       try {
         await options.verifyAuthorizationRequestCallback({
+          agentContext,
           authorizationRequest:
             verifiedAuthorizationRequest.authorizationRequestPayload as Openid4vpAuthorizationRequest,
-          jwsService: agentContext.resolve(JwsService),
           client,
         })
       } catch (e) {
-        throw new CredoError(
-          `verificationAuthorizationCallback returned false. User-provided validation failed. cause: ${e}`,
-          {
-            cause: e,
-          }
-        )
+        throw new CredoError(`error during call to User-provided verificationAuthorizationCallback. Cause: ${e}`, {
+          cause: e,
+        })
       }
     }
 
--- a/packages/openid4vc/src/openid4vc-holder/OpenId4vpHolderService.ts
+++ b/packages/openid4vc/src/openid4vc-holder/OpenId4vpHolderService.ts
@@ -28,7 +28,6 @@ import {
   injectable,
 } from '@credo-ts/core'
 import {
-  Openid4vpAuthorizationRequest,
   Openid4vpAuthorizationResponse,
   Openid4vpClient,
   VpToken,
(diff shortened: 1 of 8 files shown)


berendsliedrecht:
> can you add a changeset (just know i will keep commenting this on every PR, so might be good to just add it to every PR by default 😅)

Haha yes, I will!

packages/openid4vc/tests/openid4vc.e2e.test.ts
```diff
@@ -353,6 +367,233 @@ describe('OpenId4Vc', () => {
...
+        verifyAuthorizationRequestCallback: () => {},
+      })
+    ).resolves.toBeDefined()
+  })
+
+  it('e2e flow with tenants, holder verification callback for authorization request succeeds', async () => {
+    const holderTenant = await holder.agent.modules.tenants.getTenantAgent({ tenantId: holder1.tenantId })
+
+    await expect(
+      holderTenant.modules.openId4VcHolder.resolveOpenId4VpAuthorizationRequest(
```
berendsliedrecht, line 463: Main interesting code, i.e. validation the registration certificate, happens here.

packages/openid4vc/src/openid4vc-holder/OpenId4vpHolderServiceOptions.ts
```diff
@@ -1,20 +1,32 @@
...
-import type { OpenId4VpAuthorizationRequestPayload } from '../shared'
+import { Openid4vpAuthorizationRequestDcApi, ResolvedOpenid4vpAuthorizationRequest } from '@openid4vc/openid4vp'
+import type { OpenId4VpAuthorizationRequestPayload, Openid4vpAuthorizationRequest } from '../shared'
 
 // TODO: export from oid4vp
 export type ParsedTransactionDataEntry = NonNullable<ResolvedOpenid4vpAuthorizationRequest['transactionData']>[number]
 
+export type VerifyAuthorizationRequestOptions = {
+  authorizationRequest: Openid4vpAuthorizationRequest | Openid4vpAuthorizationRequestDcApi
+  jwsService: JwsService
```
TimoGlastra, line 19: I think it makes more sense to pass the `agentContext` as the first param to the callback. The user can then inject anything they need. Attestations could be other things than JWTs as well, so it's too focused on a specific case.

If the context is passed the user can just do `agentContext.resolve(JwsService)`

packages/openid4vc/src/openid4vc-holder/OpenId4vpHolderServiceOptions.ts
```diff
@@ -1,20 +1,32 @@
...
 } from '@credo-ts/core'
-import { ResolvedOpenid4vpAuthorizationRequest } from '@openid4vc/openid4vp'
-import type { OpenId4VpAuthorizationRequestPayload } from '../shared'
+import { Openid4vpAuthorizationRequestDcApi, ResolvedOpenid4vpAuthorizationRequest } from '@openid4vc/openid4vp'
+import type { OpenId4VpAuthorizationRequestPayload, Openid4vpAuthorizationRequest } from '../shared'
 
 // TODO: export from oid4vp
 export type ParsedTransactionDataEntry = NonNullable<ResolvedOpenid4vpAuthorizationRequest['transactionData']>[number]
 
+export type VerifyAuthorizationRequestOptions = {
```
TimoGlastra, line 17: I feel we're still missing relevant information from the `OpenId4VpResolvedAuthorizationRequest` interface here.

Also we should try to stay as consistent as possible with `OpenId4VpResolvedAuthorizationRequest`.

So 
```ts
authorizationRequestPayload: OpenId4VpAuthorizationRequestPayload
```

and ideally:

```
  /**
   * Metadata about the signed authorization request.
   *
   * Only present if the authorization request was signed
   */
  signedAuthorizationRequest?: {
    signer: VerifiedJarRequest['signer']
    payload: VerifiedJarRequest['jwt']['payload']
    header: VerifiedJarRequest['jwt']['header']
  }
```

also add the origin (in case of DC-API):
```
  /**
   * Origin of the request, to be used with Digital Credentials API
   */
  origin?: string
```

Maybe we can just pass the same interface? Or do we call this before the DCQL and Presentation Exchange Queries are run? (in that case we can omit these from the interface). But i would rather not add `client` here if it's not also available in the resolved value.


https://github.com/openwallet-foundation/credo-ts/blob/main/packages/openid4vc/src/openid4vc-holder/OpenId4vpHolderServiceOptions.ts#L22

packages/openid4vc/src/openid4vc-holder/OpenId4vpHolderService.ts
```diff
@@ -141,6 +143,24 @@ export class OpenId4VpHolderService {
...
+      try {
+        await options.verifyAuthorizationRequestCallback({
+          authorizationRequest:
+            verifiedAuthorizationRequest.authorizationRequestPayload as Openid4vpAuthorizationRequest,
+          jwsService: agentContext.resolve(JwsService),
+          client,
+        })
+      } catch (e) {
+        throw new CredoError(
+          `verificationAuthorizationCallback returned false. User-provided validation failed. cause: ${e}`,
```
TimoGlastra, line 156: The method does not return false anymore

```suggestion
          `error during call to user-provided verifyAuthorizationRequestCallback. cause: ${e}`,
```

packages/openid4vc/src/openid4vc-holder/OpenId4vpHolderServiceOptions.ts
berendsliedrecht: Yes, makes sense

TimoGlastra (approved):
One remaining comment

PR #2309, opened 2025-06-02
feat: add verfier attestations as a verifier

berendsliedrecht:

### DIFF
--- a/packages/openid4vc/src/openid4vc-verifier/OpenId4VpVerifierService.ts
+++ b/packages/openid4vc/src/openid4vc-verifier/OpenId4VpVerifierService.ts
@@ -128,6 +128,11 @@ export class OpenId4VpVerifierService {
     if (version === 'v1.draft21' && options.dcql) {
       throw new CredoError(`OpenID4VP version '${version}' cannot be used with dcql. Use version 'v1.draft24' instead.`)
     }
+    if (version === 'v1.draft21' && options.verifierAttestations) {
+      throw new CredoError(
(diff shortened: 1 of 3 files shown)


packages/openid4vc/src/openid4vc-verifier/OpenId4VpVerifierService.ts
```diff
@@ -219,6 +224,7 @@ export class OpenId4VpVerifierService {
       response_mode: responseMode,
       response_type: 'vp_token',
       client_metadata,
+      verifier_attestations: options.verifierAttestations,
```
TimoGlastra, line 242: Should we check that if `verifier_attestations` are provided we make sure that at least the `credential_ids` are present in the provided pex/dcql query?

packages/openid4vc/src/openid4vc-verifier/OpenId4VpVerifierServiceOptions.ts
```diff
@@ -35,6 +42,8 @@ export interface OpenId4VpCreateAuthorizationRequestOptions {
    */
   transactionData?: TransactionDataEntry[]
 
+  verifierAttestations?: VerifierAttestations
```
TimoGlastra, line 47: A short documentation string will help.

TimoGlastra (approved):

packages/openid4vc/src/openid4vc-verifier/OpenId4VpVerifierService.ts
berendsliedrecht: Yeah that would be a good validation to do.

PR #2310, opened 2025-06-02
chore: remove verify auth callback

berendsliedrecht:
- **chore: revert verifyAuthorizationRequestCallback**
- **docs(changeset):  Revert the verifyAuthorizationCallback. Should be called outside of the framework**

### DIFF
--- a/.changeset/gold-wombats-think.md
+++ b/.changeset/gold-wombats-think.md
@@ -1,0 +1,5 @@
+---
+"@credo-ts/openid4vc": patch
+---
+
+Revert the verifyAuthorizationCallback. Should be called outside of the framework


.changeset/gold-wombats-think.md
```diff
@@ -0,0 +1,5 @@
+---
+"@credo-ts/openid4vc": patch
+---
+
+Revert the verifyAuthorizationCallback. Should be called outside of the framework
```
TimoGlastra, line 5: i think we can just remove the previous changeset that was added (and this one), so we don't clutter the changelog

packages/openid4vc/tests/openid4vc.e2e.test.ts
```diff
@@ -437,20 +437,6 @@ pUGCFdfNLQIgHGSa5u5ZqUtCrnMiaEageO71rjzBlov0YUH4+6ELioY=
 
     await verifierTenant1.endSession()
     await verifierTenant2.endSession()
```
TimoGlastra, line 439: this test was added for this purpose right, so we cna probably remove it? Or did you adapt an existing test?

TimoGlastra (approved):

packages/openid4vc/tests/openid4vc.e2e.test.ts
berendsliedrecht: Adapted existing.

Issue #1960, opened 2024-07-19
Error Revoking Multiple AnonCreds Credentials


tarunvaddeSoul:
When attempting to revoke multiple anoncreds credentials, I encounter errors related to invalid client requests. Below are the specific error messages:

```
{"op":"REJECT","identifier":"JM9L6HL2QCexjbn9WB46h9","reqId":1721286371992373000,"reason":"client request invalid: InvalidClientRequest('JM9L6HL2QCexjbn9WB46h9', 1721286371992373000, 'Issued indices from txn: {0, 3, 4} are not present in the current revoked list from state: [1]')","code":null}


{"op":"REJECT","identifier":"JM9L6HL2QCexjbn9WB46h9","reqId":1721286460910490000,"reason":"client request invalid: InvalidClientRequest('JM9L6HL2QCexjbn9WB46h9', 1721286460910490000, 'Issued indices from txn: {0, 2, 3} are not present in the current revoked list from state: [1]')","code":null}


{"op":"REJECT","identifier":"JM9L6HL2QCexjbn9WB46h9","reqId":1721286535864486000,"reason":"client request invalid: InvalidClientRequest('JM9L6HL2QCexjbn9WB46h9', 1721286535864486000, 'Issued indices from txn: {0, 2, 4} are not present in the current revoked list from state: [1]')","code":null}
```

## Steps to reproduce: 

1. Clone the repository [git clone https://github.com/tarunvaddeSoul/credo-anoncreds-revocation-issue
](URL)
2. cd credo-anoncreds-revocation-issue
3. npm install
4. Additions in .env
```
TAILS_SERVER_BASE_URL=<your_tails_server_base_url>
TAILS_DIRECTORY_PATH=<your_tails_directory_path>
```
5. Start the project with `npm start`

## Environment

- Node.js version: v18.17.1
- npm version: 9.6.7
- Tails server: Running from `credo-ts/samples/tails`

Arif-Shahriar028:
@tarunvaddeSoul could you find any solution of this? I am facing the same issue when trying to revoke multiple credentials.

tarunvaddeSoul:
@Arif-Shahriar028 I didn't look into it after raising this issue since I didn't get any response

Issue #1807, opened 2024-03-27
Revocation Clarification (does not work after the first revocation in a list)


vk1992op:
TLDR: Revocation works fine for the first credential but if I try to revoke a second one in the same status list the ledger returns: 
```
"state\":\"failed\",\"reason\":\"unknownError: Request failed: client request invalid: InvalidClientRequest('7wV7McGyL5MnfpDF321NS3', 1711548598037245847, 'Issued indices from txn: {0, 3, 4} are not present in the current revoked list from state: [1]')\"
```

There is a possibility that I do not fully understand how exactly the flow for revocation should work, so I will write down my flow.

## Setup

I have done the flow as described in the test of the anoncreds package

1. register cred def
2. register revRegDef (max Number: 5)
3. register status list 
4. issue a credential as revocable on that credDef with that revRegDef, with index 0
5. issue another one, with index 1
6. ...etc. (5 credentials total)
7. revoke credential at index 0 -> OK
8. check status of credential at index 0 -> OK
9. revoke credential at index 2 (or any other) -> Error

**However**
If I revoke all other credentials in the list (all except the one at index 0, which is already revoked) -> OK, everything works as expected and all credentials are successfully revoked.

For the exact calling of methods and parameters I am using the tests in the anoncreds package as a reference.

Could it be that the status list has to be used to revoke all credentials inside it, but if that is the case then why does it successfully revoke one but only the first time?

TimoGlastra:
Thanks for the detailed description @vk1992op!

Could you provide a repo with a minimal reproduction in Node.JS? That way i can easily dig in and see if there's a bug or whether your flow needs tweaking.

vk1992op:
https://github.com/vk1992op/revocation-demo

This is a cut-down version of the project I am working on, you can follow the readme, and you should have it up and running in a few steps.

A few important additional notes:

#### you will need your own S3 compatible bucket for the revocation tails (s3, backblaze b2, cloudflare, digital ocean)
#### the base url for the postman calls will be: http://0.0.0.0:8002 (if port is not changed in env)


#### the files of interest are: 
  -  /libs/askar/src/credo/revocation/TailFileService
  -  /libs/askar/src/askar/agent.service  -> offerCredential function (indices for the status list are automatically incremented and revocation definitions are automatically registered once the previous ones are full); revokeCredentialById function
  -  /libs/askar/src/askar-rest/rest.controller -> here you can find the various requests and what function of the service they use

[...]

PR #2319, opened 2025-06-11
fix: correct indyVdrCreateLatestRevocationDelta to allow individual credential revocation/unrevocation

### LINKED ISSUE #2318: Bug: Incorrect delta calculation in indyVdrCreateLatestRevocationDelta prevents individual credential revocation/unrevocation
## Summary

It is currently **not possible to revoke more than one credential individually** in a revocation registry using Credo, due to a bug in the delta calculation logic in `indyVdrCreateLatestRevocationDelta` (`packages/indy-vdr/src/anoncreds/utils/transform.ts`).  
This results in errors such as:
```
Issued indices from txn: {0, 3, 4} are not present in the current revoked list from state: [1]
```
- Reproducible by following the steps in [#1960](https://github.com/openwallet-foundation/credo-ts/issues/1960) and [#1807](https://github.com/openwallet-foundation/credo-ts/issues/1807).

---

## Steps to Reproduce

1. Clone https://github.com/tarunvaddeSoul/credo-anoncreds-revocation-issue
2. Follow the README and setup instructions.
3. Issue several revocable credentials with the same revocation registry.
4. Attempt to revoke credentials individually, one after another.

**Expected:** Each credential can be revoked or unrevoked independently.  
**Actual:** Only the first credential can be revoked. Further attempts to revoke others individually fail, or only mass revocation/unrevocation works.

---

### Previous Code
```typescript
// Incorrect: Used previousDelta.issued instead of comparing to previous state
if (revocationStatus === RevocationState.Active && !previousDelta.issued.includes(idx)) {
    issued.push(idx);
}
```
### What it did?

- For each index:
If the current status is Active (not revoked) and this index was not in the previous `issued` list, add it to the new `issued` list.

### Implication

- This logic would add any index that is currently active and was not previously issued.
- If `previousDelta.issued` is empty, this would add all active indexes to `issued`, even if they were not changed since the last delta.
```ts
// Example:

// You have a revocation registry for 5 credentials.

// The current revocation status list is:
const revocationStatusList = [0, 1, 0, 1, 0];
// 0 = active (not revoked), 1 = revoked

// The previous delta is:
const previousDelta = {
  issued: [],
  revoked: [1, 3],
};
// No new credentials were issued.
// Indexes 1 and 3 were revoked previously.

// Now, since previousDelta.issued is an empty array,
// the check `!previousDelta.issued.includes(idx)` will return true for every index.

// So for each index where revocationStatus === RevocationState.Active (i.e., 0, 2, 4),
// the logic incorrectly does:
issued.push(idx);

// Final result:
issued = [0, 2, 4];
// All currently active credentials are added,
// even though they were never changed or unrevoked.

// When you submit issued = [0, 2, 4] to the ledger as part of a revocation delta, the ledger interprets this as a request to "unrevoke" (activate) credentials at indexes 0, 2, and 4.

// Example error you see: Issued indices from txn: {0, 2, 4} are not present in the current revoked list from state: [...]

```
- It does not check if the index was previously revoked, so it can't distinguish between "newly unrevoked" and "always active".

### Fixed Code
```typescript
// Correct: Only unrevoked indexes (were revoked, now active) are added
if (revocationStatus === RevocationState.Active && previousDelta.revoked.includes(idx)) {
    issued.push(idx); // UNREVOKE
}
```
### What this fixes?
- For each index in the current revocationStatusList: If the status is **Active** (0) and the index was previously **revoked**, it means this credential was just **unrevoked** (so add to **issued**).

---

## Why Was the Bug Happening?

- The old code assumed the last delta's `issued`/`revoked` arrays represented the full current state, but they only represent the *change* from the last update.
- This caused the ledger to reject requests that tried to issue (unrevoke) indices that were not previously revoked, or revoke indices that were already revoked.
- See also: [#1960](https://github.com/openwallet-foundation/credo-ts/issues/1960), [#1807](https://github.com/openwallet-foundation/credo-ts/issues/1807)

---

## Solution

- The delta calculation must compare the **current and previous status lists** (bit arrays), not just the previous delta's change sets.
- The revised code above enables correct per-credential revocation and unrevocation.

---

## References

- [#1960](https://github.com/openwallet-foundation/credo-ts/issues/1960) (detailed steps and repo to reproduce)
- [#1807](https://github.com/openwallet-foundation/credo-ts/issues/1807) (similar issue by another user)

---

tarunvaddeSoul:
## Summary

Fixes a bug in `indyVdrCreateLatestRevocationDelta` where the delta calculation incorrectly included all active indexes in `issued` if `previousDelta.issued` was empty, preventing correct individual credential revocation/unrevocation.

Closes #2318, #1960, #1807 

## Details

- The previous logic added all active indexes to `issued` if they were not in `previousDelta.issued`, even if they were never revoked.
- The new logic only adds indexes to `issued` if they were previously revoked and are now active (i.e., actually unrevoked).
- Updated/added tests to cover individual revoke and unrevoke scenarios.

## Testing

- All existing and new tests pass.
- See `transformers.test.ts` for new/updated test cases.

## Additional Context

See linked issues for detailed reproduction steps and discussion.

### DIFF
--- a/packages/indy-vdr/src/anoncreds/utils/__tests__/transform.test.ts
+++ b/packages/indy-vdr/src/anoncreds/utils/__tests__/transform.test.ts
@@ -102,7 +102,7 @@ describe('transform', () => {
           delta,
           true
         )
-      ).toThrowError()
+      ).toThrow()
     })
   })
 
@@ -170,7 +170,7 @@ describe('transform', () => {
 
       const { revoked, issued } = indyVdrCreateLatestRevocationDelta(accum, revocationStatusList, delta)
 
-      expect(issued).toStrictEqual([1, 2, 3, 4])
+      expect(issued).toStrictEqual([])
       expect(revoked).toStrictEqual([6, 7, 8, 9])
     })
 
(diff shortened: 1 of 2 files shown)


tarunvadde:
@TimoGlastra Hi, Timo. Could we please get this checked on urgency?

TimoGlastra:
Apologies for the late review @tarunvaddeSoul, i missed this PR. Thank you!

tarunvadde:
Hi, @TimoGlastra 

Thanks for merging — is there a planned release schedule for @credo-ts/indy-vdr including this change?

TimoGlastra:
You can download the latest alpha (automatically released for every commit). If you want it in 0.5.x, you'd need ti make the same pr to 0.5.x branch and we can backport
It and release it there as well

tarunvadde:
My apologies for the late response, Timo. I will do that, thank you.

PR #2324, opened 2025-06-25
feat: native SQLite and Postgres database support based on Drizzle ORM

TimoGlastra:
Supersedes #2289

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -55,10 +55,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # NOTE: starting from node 22.18 there's some issues with type-stripping
-        # Fixing it breaks node 20 support, so for now pinning to node 22.17
-        # https://github.com/nodejs/node/issues/59364
-        node-version: [20, 22.17]
+        node-version: [20, 22]
         # Each shard runs a set of the tests
         # Make sure to UPDATE THE TEST command with the total length of
         # the shards if you change this!!
@@ -91,8 +88,15 @@ jobs:
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
 
+      # Issue with node 22 starting 22.18
+      # https://github.com/nodejs/node/issues/59364
+      - name: Run tests
+        run: NODE_OPTIONS="$NODE_OPTIONS --no-experimental-strip-types" pnpm test:unit --coverage --forceExit --shard=${{ matrix.shard }}/3
+        if: matrix.node-version == 22
+
       - name: Run tests
         run: pnpm test:unit --coverage --forceExit --shard=${{ matrix.shard }}/3
+        if: matrix.node-version != 22
 
       # Upload coverage for shard
       - run: mv coverage/coverage-final.json coverage/${{ matrix.shard }}.json
@@ -110,9 +114,6 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # NOTE: starting from node 22.18 there's some issues with type-stripping
-        # Fixing it breaks node 20 support, so for now pinning to node 22.17
-        # https://github.com/nodejs/node/issues/59364
         node-version: [20, 22]
 
     steps:
@@ -147,8 +148,15 @@ jobs:
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
 
+      # Issue with node 22 starting 22.18
+      # https://github.com/nodejs/node/issues/59364
+      - name: Run tests
+        run: NODE_OPTIONS="$NODE_OPTIONS --no-experimental-strip-types" pnpm test:e2e --coverage --forceExit
+        if: matrix.node-version == 22
+
       - name: Run tests
         run: pnpm test:e2e --coverage --forceExit
+        if: matrix.node-version != 22
 
       # Upload coverage for e2e
       - run: mv coverage/coverage-final.json coverage/e2e.json
@@ -166,9 +174,6 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # NOTE: starting from node 22.18 there's some issues with type-stripping
-        # Fixing it breaks node 20 support, so for now pinning to node 22.17
-        # https://github.com/nodejs/node/issues/59364
         node-version: [20, 22]
 
     steps:
@@ -203,8 +208,15 @@ jobs:
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
 
+      # Issue with node 22 starting 22.18
+      # https://github.com/nodejs/node/issues/59364
+      - name: Run tests
+        run: NODE_OPTIONS="$NODE_OPTIONS --no-experimental-strip-types" pnpm test:drizzle --coverage --forceExit
+        if: matrix.node-version == 22
+
       - name: Run tests
         run: pnpm test:drizzle --coverage --forceExit
+        if: matrix.node-version != 22
 
       # Upload coverage for e2e
       - run: mv coverage/coverage-final.json coverage/drizzle.json


packages/drizzle-storage/src/adapter/BaseDrizzleRecordAdapter.ts
```diff
@@ -0,0 +1,319 @@
...
+        throw new RecordNotFoundError(`record with id ${id} not found.`, {
+          recordType: this.recordType,
+        })
+      }
+
+      return
+    }
+
+    if (isDrizzleSqliteDatabase(this.database)) {
+      const deleted = await this.database
```
vinaysingh8866, line 294: Hi @TimoGlastra, found small issue 
The SQLite database delete method is using the wrong table reference 
```
if (isDrizzleSqliteDatabase(this.database)) {
  const deleted = await this.database
    .delete(this.table.postgres)  // Should be this.table.sqlite
    .where(
```

TimoGlastra: Thanks @vinaysingh8866, i have resolved the issue!

packages/anoncreds/tests/legacyAnonCredsSetup.ts
```diff
@@ -60,6 +60,12 @@ import {
   parseIndySchemaId,
 } from '../src'
 
+import { DrizzleStorageModule } from '../../drizzle-storage/src'
```
berendsliedrecht, line 63: should this not import from @credo-ts/drizze?

packages/anoncreds/tests/legacyAnonCredsSetup.ts
```diff
@@ -313,7 +322,16 @@ export async function setupAnonCredsTests<
   attributeNames?: string[]
   preCreatedDefinition?: PreCreatedAnonCredsDefinition
   createConnections?: CreateConnections
+  useDrizzleStorage?: boolean | 'postgres' | 'sqlite'
```
berendsliedrecht:
```suggestion
  useDrizzleStorage?: false | 'postgres' | 'sqlite'
```

slightly more correct, right?

packages/core/tests/helpers.ts
```diff
@@ -105,10 +109,14 @@ export function getAgentOptions<AgentModules extends AgentModulesInput | EmptyMo
...
-  { requireDidcomm = false, inMemory = true }: { requireDidcomm?: boolean; inMemory?: boolean } = {}
+  {
+    requireDidcomm = false,
+    inMemory = true,
+    drizzle,
+  }: { requireDidcomm?: boolean; inMemory?: boolean; drizzle?: AnyDrizzleDatabase } = {}
 ): {
   config: InitConfig
-  modules: AgentModules & DefaultAgentModulesInput
+  modules: AgentModules & DefaultAgentModulesInput & { drizzle?: DrizzleStorageModule }
```
berendsliedrecht, line 119: Should this type not be on the DefaultAgentModulesInput?

packages/drizzle-storage/src/DrizzleRecord.ts
```diff
@@ -0,0 +1,69 @@
...
+    }
+  }
+}
+
+export interface DrizzleRecord {
+  postgres: Record<string, unknown>
+  sqlite: Record<string, unknown>
+  adapter?: new (
+    // biome-ignore lint/suspicious/noExplicitAny: <explanation>
+    database: any
```
berendsliedrecht, line 67: Won't `AnyDrizzleDatabase` work here?

packages/drizzle-storage/src/action-menu/action-menu/sqlite.ts
```diff
@@ -0,0 +1,34 @@
+import { foreignKey, sqliteTable, text, unique } from 'drizzle-orm/sqlite-core'
```
berendsliedrecht: bit confused about the file path here, why the second action-menu folder?

packages/drizzle-storage/src/adapter/BaseDrizzleRecordAdapter.ts
```diff
@@ -0,0 +1,319 @@
+/**
```
berendsliedrecht: Maybe add this comment to the abstract class

berendsliedrecht (approved):
Looks good!

One thing that I was wondering about, is that we are very tightly coupled now to Drizzle (naming and without any major abstraction). We tend to rewrite things once in a while :'). Do you think that if that were to happen, that we would be fine? Or would it help to have a thin layer between Drizzle, or would you prefer just a complete new package as it must be quite thight to the ORM?

packages/anoncreds/tests/legacyAnonCredsSetup.ts
TimoGlastra: You can provide true and it will use `sqlite` by default, but we can also make it explicit? But in that case we could also do `useDrizzleStorage?: 'postgres' | 'sqlite'`

packages/drizzle-storage/src/DrizzleRecord.ts
TimoGlastra: No you get the `could be instantiated with a different ...` something error. I've tried a few things, and couldn't get it to work since drizzle's database supports a lot of dialtects and drivers.

packages/drizzle-storage/src/action-menu/action-menu/sqlite.ts
TimoGlastra: Action menu package, action menu record.

So for didcomm pckage we have `didcomm/connection`. Didn't want to break consistency here for the duplicate.

TimoGlastra:
> One thing that I was wondering about, is that we are very tightly coupled now to Drizzle (naming and without any major abstraction). We tend to rewrite things once in a while :'). Do you think that if that were to happen, that we would be fine? Or would it help to have a thin layer between Drizzle, or would you prefer just a complete new package as it must be quite thight to the ORM?

I think since the migrations are very tied to Drizzle it's good to not abstract it away IMO. If we were to move to another provider, the nice things is that at least the databases structure can stay the same, and only the management tool changes, so I thought it would be fine to couple it.

packages/anoncreds/tests/legacyAnonCredsSetup.ts
berendsliedrecht: Yeah I think that that is better.

packages/drizzle-storage/src/DrizzleRecord.ts
berendsliedrecht: Ah.... yeah I do not like that error.

packages/drizzle-storage/src/action-menu/action-menu/sqlite.ts
berendsliedrecht: Ahh makes sense. Might be overkill, but have you considered `connection-record` as the file name?

packages/core/tests/helpers.ts
TimoGlastra: Noo, since drizzle is not a default module (it's an extension module)

packages/drizzle-storage/README.md
```diff
@@ -0,0 +1,217 @@
...
+### Bundles
+
+The bundles provide the agent with the needed context about which record types are used. For each of the modules you use with Credo that provides Record classes, the respective drizzle record MUST be registered, otherwise you cannot use the record, and an error will be thrown when such a record is retrieved, stored, updated or deleted. The `coreDrizzleBundle` is always registered, as these are required to use Credo.
+
+The following bundles are provided out of the box by the Drizzle storage module:
+
+- `coreDrizzleBundle` from `@credo-ts/drizzle-storage/core` (`core`)
+- `didcommDrizzleBundle` from `@credo-ts/drizzle-storage/didcomm` (`didcomm`)
+- `actionMenuDrizzleBundle` from `@credo-ts/drizzle-storage/action-menu` (`action-menu`)
+- `anonCredsDrizzleBundle` from `@credo-ts/drizzle-storage/anoncreds` (`anoncreds`)
```
genaris, line 98: I think you have to add also the ones for drpc, openid4vc, etc. (added after the writing of this README)

PR #2325, opened 2025-06-27
feat(cheqd): Add Token StatusList Service

DaevMithran:


packages/cheqd/src/tokenStatusList/CheqdTokenStatusListService.ts
```diff
@@ -0,0 +1,159 @@
+import { AgentContext, CredoError, getKeyFromVerificationMethod, Buffer, utils } from '@credo-ts/core'
+import { StatusListPayload, StatusListToken } from './types'
+import { createEmptyBitmap, decodeBitmap, encodeBitmap, isBitSet, setBit } from './utils/bitmap'
+import { CheqdCreateResourceOptions, CheqdDidResolver } from '../dids'
+import { TokenStatusList } from './types/tokenStatusList'
+import { CheqdApi } from '../CheqdApi'
+import { parseCheqdDid } from '../anoncreds/utils/identifiers'
+import base64url from 'base64url'
```
TimoGlastra, line 8: We have base64url support in the `TypedArrayEncoder` in the core package

packages/cheqd/src/tokenStatusList/CheqdTokenStatusListService.ts
```diff
@@ -0,0 +1,159 @@
...
+    statusListId: string
+  ): Promise<{
+    metadata: any
+    bitmap: Buffer
+    jwt: string
+  }> {
+    const api = agentContext.dependencyManager.resolve(CheqdApi)
+    const resource = await api.resolveResource(statusListId)
+    const jwt = resource.resource?.data.toString()
+    const payload = JSON.parse(Buffer.from(jwt!.split('.')[1], 'base64').toString()) as StatusListPayload
```
TimoGlastra, line 22: WE have tools like the `Jwt.fromSerializedJwt` to parse JWTs. 

Also should we not verify it first?

packages/cheqd/src/tokenStatusList/CheqdTokenStatusListService.ts
```diff
@@ -0,0 +1,159 @@
...
+    await api.createResource(did, resource)
+
+    return {
+      jwt,
+      metadata: {
+        statusListId: resource.id,
+        issuedAt: payload.iat,
+        size: size,
+      },
+    }
```
TimoGlastra, line 69: We have quite some utils also to create JWTs. See the Jwt service, also we should probably make a generic status list service that creates the status list (like we have an SD-JWT VC service) which works with a `signer.method` (see sd-jwt-vc service) which can be a cheqd did or a x509 cert for example. 

Making a cheqd focused status list service will just result in duplication. So I don't think we can merge the PR like this, and it would first need to be refactored to a generic status list service.

Then i see two options for the cheqd integration:
- you call two methods
	- we have a `agent.modules.tokenStatusList.createStatusList`
	- we have a `agent.modules.cheqd.uploadStatusList`
- we create one wrapper method
  - `agent.modules.cheqd.createStatustList` - which call the `agent.modules.tokenStatusList.createStatusList` and then uploads it as a cheqd resource, but still using the generic implementation

packages/cheqd/src/tokenStatusList/CheqdTokenStatusListService.ts
```diff
@@ -0,0 +1,159 @@
+import { AgentContext, CredoError, getKeyFromVerificationMethod, Buffer, utils } from '@credo-ts/core'
+import { StatusListPayload, StatusListToken } from './types'
+import { createEmptyBitmap, decodeBitmap, encodeBitmap, isBitSet, setBit } from './utils/bitmap'
```
TimoGlastra, line 3: SD-JWT VC library already has utils to create status lists. 

```
      const statusListValues = new Array(body.size).fill(0)
      const statusList = new StatusList(statusListValues, 1)
```

Or updating an existing one:
```ts
import { getListFromStatusListJWT } from '@sd-jwt/jwt-status-list'

const statusList = getListFromStatusListJWT(currentStatusListJwt)
const parsedStatusListJwt = Jwt.fromSerializedJwt(currentStatusListJwt)

// Set the revoked indices to 1
for (const revokedIndex of body.revokedIndices) {
  statusList.setStatus(revokedIndex, 1)
}
```

packages/cheqd/src/tokenStatusList/CheqdTokenStatusListService.ts
DaevMithran: @TimoGlastra Is this the right way to create JWTs?
```
        const jwsService = agentContext.dependencyManager.resolve(JwsService)
        const jwt = await jwsService.createJwsCompact(agentContext, {
            payload: jwtPayload,
            keyId: issuer.publicJwk.keyId,
            protectedHeaderOptions: {
                alg: issuer.alg,
                typ: "statuslist+jwt"
            }
        })
```

DaevMithran: @TimoGlastra to publish status lists in different registries. How about following the pattern similar to registrar

`agent.modules.tokenStatusList.registerRegistry([new CheqdStatusListRegistry(), new HttpStatusListRegistry]); etc.
`
and the functions within TokenStatusListService call registry.publish(), registry.retreive() etc

DaevMithran:
Closing and implementing new approach suggested in https://github.com/openwallet-foundation/credo-ts/pull/2326

PR #2326, opened 2025-07-02
feat: Add Token Status List support for SDJWT VC

DaevMithran:
This PR adds support for **Token Status Lists** in SD-JWT VC credentials, based on the specification (\[[spec link](https://www.ietf.org/archive/id/draft-ietf-oauth-status-list-02.html)]).

#### Features:

* **Token Status List integration**: Enables status management for SD-JWT VCs using the Token Status List mechanism defined in the spec.
* **Pluggable registry system**:
  The approach to configuring **Token Status List registries** follows the same pattern used for registrars and resolvers. This ensures flexibility and consistency within the framework.

#### Default Behavior:

* By default, the module uses the `HttpTokenStatusListRegistry` to publish and manage status lists via HTTP endpoints.

#### Custom Registry Support:

* Consumers can override the default behavior by implementing the `TokenStatusListRegistryInterface` and configuring custom registries at module startup.
* Example:

  ```ts
  sdJwtVc: new SdJwtVcModule({
    registries: [new CheqdTokenStatusListRegistry()],
  })
  ```

### DIFF
--- a/packages/cheqd/src/wrappers/CheqdTokenStatusListRegistry.ts
+++ b/packages/cheqd/src/wrappers/CheqdTokenStatusListRegistry.ts
@@ -2,15 +2,15 @@ import {
   type AgentContext,
   CredoError,
   Jwt,
-  PublishTokenStatusListOptions,
+  type PublishTokenStatusListOptions,
   parseDid,
-  SdJwtVcIssuerDid,
-  TokenStatusListRegistry,
+  type SdJwtVcIssuerDid,
+  type TokenStatusListRegistry,
   TypedArrayEncoder,
   utils,
 } from '@credo-ts/core'
 import { cheqdSdkAnonCredsRegistryIdentifierRegex, parseCheqdDid } from '../anoncreds/utils/identifiers'
-import { CheqdCreateResourceOptions, CheqdDidRegistrar, CheqdDidResolver } from '../dids'
+import { type CheqdCreateResourceOptions, CheqdDidRegistrar, CheqdDidResolver } from '../dids'
 
 export class CheqdTokenStatusListRegistry implements TokenStatusListRegistry {
   methodName = 'cheqd'
--- a/packages/core/src/modules/sd-jwt-vc/SdJwtVcModuleConfig.ts
+++ b/packages/core/src/modules/sd-jwt-vc/SdJwtVcModuleConfig.ts
@@ -1,4 +1,4 @@
-import { TokenStatusListRegistry } from './credential-status'
+import type { TokenStatusListRegistry } from './credential-status'
 import { HttpTokenStatusListRegistry } from './credential-status/token-status-list/http/HttpTokenStatusListRegistry'
 
 /**
--- a/packages/core/src/modules/sd-jwt-vc/credential-status/token-status-list/http/HttpTokenStatusListRegistry.ts
+++ b/packages/core/src/modules/sd-jwt-vc/credential-status/token-status-list/http/HttpTokenStatusListRegistry.ts
@@ -1,8 +1,8 @@
 import { AgentContext } from '../../../../../agent'
 import { CredoError } from '../../../../../error'
 import { fetchWithTimeout } from '../../../../../utils/fetch'
-import { SdJwtVcIssuerX5c } from '../../../SdJwtVcOptions'
-import { PublishTokenStatusListOptions, TokenStatusListRegistry } from '../TokenStatusListRegistry'
+import type { SdJwtVcIssuerX5c } from '../../../SdJwtVcOptions'
+import type { PublishTokenStatusListOptions, TokenStatusListRegistry } from '../TokenStatusListRegistry'
 
 export class HttpTokenStatusListRegistry implements TokenStatusListRegistry {
   public readonly methodName = 'http'
--- a/packages/cheqd/tests/cheqd-token-status-list.e2e.test.ts
+++ b/packages/cheqd/tests/cheqd-token-status-list.e2e.test.ts
@@ -1,4 +1,4 @@
-import { Agent, SdJwtVcIssuer, TokenStatusListService } from '@credo-ts/core'
+import { Agent, type SdJwtVcIssuer, TokenStatusListService } from '@credo-ts/core'
 import { getAgentOptions } from '../../core/tests/helpers'
 import type { CheqdDidCreateOptions } from '../src'
 import { cheqdPayerSeeds, getCheqdModules } from './setupCheqdModule'
--- a/packages/core/src/modules/sd-jwt-vc/SdJwtVcModule.ts
+++ b/packages/core/src/modules/sd-jwt-vc/SdJwtVcModule.ts
@@ -2,7 +2,7 @@ import type { DependencyManager, Module } from '../../plugins'
 import { TokenStatusListService } from './credential-status'
 import { SdJwtVcRepository } from './repository'
 import { SdJwtVcApi } from './SdJwtVcApi'
-import { SdJwtVcModuleConfig, SdJwtVcModuleConfigOptions } from './SdJwtVcModuleConfig'
+import { SdJwtVcModuleConfig, type SdJwtVcModuleConfigOptions } from './SdJwtVcModuleConfig'
 import { SdJwtVcService } from './SdJwtVcService'
 
 /**
--- a/packages/core/src/modules/sd-jwt-vc/credential-status/token-status-list/TokenStatusListRegistry.ts
+++ b/packages/core/src/modules/sd-jwt-vc/credential-status/token-status-list/TokenStatusListRegistry.ts
@@ -1,6 +1,6 @@
-import { BitsPerStatus } from '@sd-jwt/jwt-status-list'
+import type { BitsPerStatus } from '@sd-jwt/jwt-status-list'
 import type { AgentContext } from '../../../../agent'
-import { SdJwtVcIssuer } from '../../SdJwtVcOptions'
+import type { SdJwtVcIssuer } from '../../SdJwtVcOptions'
 
 export interface PublishTokenStatusListOptions {
   uri?: string
--- a/packages/core/src/modules/sd-jwt-vc/credential-status/token-status-list/TokenStatusListService.ts
+++ b/packages/core/src/modules/sd-jwt-vc/credential-status/token-status-list/TokenStatusListService.ts
@@ -6,10 +6,10 @@ import { CredoError } from '../../../../error'
 import { dateToSeconds } from '../../../../utils'
 import { DidsApi, getPublicJwkFromVerificationMethod, parseDid } from '../../../dids'
 import { SdJwtVcModuleConfig } from '../../SdJwtVcModuleConfig'
-import { SdJwtVcIssuer } from '../../SdJwtVcOptions'
+import type { SdJwtVcIssuer } from '../../SdJwtVcOptions'
 import { extractKeyFromIssuer } from '../../utils'
 import { TokenStatusListError } from './TokenStatusListError'
-import {
+import type {
   PublishTokenStatusListOptions,
   TokenStatusListJwtPayload,
   TokenStatusListRegistry,


tweeddalex:
Hey @TimoGlastra I think this is ready for your review if you get capacity! Would be awesome to have this added, and can be consumed directly by Paradym

packages/core/src/modules/sd-jwt-vc/credential-status/token-status-list/http/HttpTokenStatusListRegistry.ts
```diff
@@ -0,0 +1,43 @@
...
+    _jwt: string,
+    _options: PublishTokenStatusListOptions
+  ): Promise<string> {
+    throw new Error('To be Implemented')
+  }
+
+  /**
+   * Retrieve a token status list JWT from the registry
+   */
+  async retrieve(agentContext: AgentContext, uri: string): Promise<string> {
```
TimoGlastra: Maybe use `resolve` or `get` to be more consistent with existing registry interfaces we have for DIDs/Anoncreds?

packages/core/src/modules/sd-jwt-vc/credential-status/token-status-list/TokenStatusList.ts
```diff
@@ -0,0 +1,237 @@
...
+              const publicJwk = getPublicJwkFromVerificationMethod(verificationMethod)
+
+              return {
+                alg,
+                method: issuer.method,
+                didUrl,
+                jwk: publicJwk,
+              }
+            }
+          : undefined,
```
TimoGlastra, line 225: If the jwt is now signed with something else than `did` we don't check if it's actually signed by the SdJwtVcIssuer.

I think we should use `jwsSigner` instead and provide it beforehand.

packages/core/src/modules/sd-jwt-vc/credential-status/token-status-list/TokenStatusList.ts
```diff
@@ -0,0 +1,237 @@
...
+      } else if (isURL(uri)) {
+        method = 'http'
+      } else {
+        throw new TokenStatusListError('Status List Uri is not supported')
+      }
+    } else if (issuer && issuer.method === 'did') {
+      method = parseDid(issuer.didUrl).method
+    } else {
+      throw new TokenStatusListError('Status List Uri is not provided')
+    }
```
TimoGlastra: Adding this logic here, makes the registry not really extendable by nature. 

I think it would make more sense to provide an uri to the registry and match based on regex (like we do with anoncreds). Both the `iss`/`uri` values can be passed to it.

packages/core/src/modules/sd-jwt-vc/credential-status/token-status-list/TokenStatusList.ts
```diff
@@ -0,0 +1,237 @@
...
+      },
+    } satisfies TokenStatusListJwtPayload)
+
+    const jwt = await jwsService.createJwsCompact(agentContext, {
+      payload: jwtPayload,
+      keyId: issuerKey.publicJwk.keyId,
+      protectedHeaderOptions: {
+        alg: issuerKey.alg,
+        typ: header.typ,
+        kid: issuerKey.kid,
```
TimoGlastra, line 141: In case the status list is signed by a did, this should be the did key reference, not the KMS key reference. Is that the case when a did is used?

packages/core/src/modules/sd-jwt-vc/credential-status/token-status-list/TokenStatusList.ts
```diff
@@ -0,0 +1,237 @@
...
+    for (const revokedIndex of options.indices) {
+      statusList.setStatus(revokedIndex, 1)
+    }
+
+    const issuerKey = await sdJwtService.extractKeyFromIssuer(agentContext, issuer, true)
+
+    // construct jwt payload
+    const jwtPayload = new JwtPayload({
+      iss,
+      iat: Math.floor(Date.now() / 1000),
```
TimoGlastra, line 126: Maybe we can move `dateToSeconds` from openid4vc package to core utils and use that?

packages/core/src/modules/sd-jwt-vc/credential-status/token-status-list/TokenStatusList.ts
```diff
@@ -0,0 +1,237 @@
...
+
+    const { indices, publish, ...publishOptions } = options
+
+    const currentStatusListJwt = await this.getStatusList(agentContext, options.uri)
+    const parsedStatusListJwt = Jwt.fromSerializedJwt(currentStatusListJwt)
+
+    // extract and validate iss
+    const iss = parsedStatusListJwt.payload.iss as string
+    if (issuer.method === 'did' && parseDid(issuer.didUrl).did !== iss) {
+      throw new TokenStatusListError('Invalid issuer')
```
TimoGlastra: Good to be more explicit in error messages. E.g. `Unable to update status list '${uri}'. Expected current 'iss' value to match the did of the issuer ('${did}')`

packages/core/src/modules/sd-jwt-vc/credential-status/token-status-list/TokenStatusList.ts
TimoGlastra, line 1: Since this exports TokenStatusListService, i think it should also be named TokenStatusListService.ts

packages/core/src/modules/sd-jwt-vc/credential-status/token-status-list/TokenStatusList.ts
```diff
@@ -0,0 +1,237 @@
...
+  uri: string
+  indices: number[]
+  publish: boolean
+}
+
+/**
+ * @internal
+ */
+@injectable()
+export class TokenStatusListService {
```
TimoGlastra, line 39: Thinking whether we should make this an external module, and not depend it on SD-jWT VC, but for now it's fine i think.

packages/core/src/modules/sd-jwt-vc/credential-status/token-status-list/TokenStatusList.ts
```diff
@@ -0,0 +1,237 @@
...
+    agentContext: AgentContext,
+    issuer: SdJwtVcIssuer,
+    options: CreateTokenStatusListOptions
+  ): Promise<CreateTokenStatusListResult> {
+    const sdJwtService = agentContext.dependencyManager.resolve(SdJwtVcService)
+    const jwsService = agentContext.dependencyManager.resolve(JwsService)
+
+    const statusList = new StatusList(new Array(options.size).fill(0), 1)
+
+    const issuerKey = await sdJwtService.extractKeyFromIssuer(agentContext, issuer, true)
```
TimoGlastra: We have a circular dependnecy here:
- TokenStatusListService depends on SdJwtVcService
- SdJwtVcService depends on TokenStatusListService

I think we should extract `extractKeyFromIssuer` into a separate reusable util that can be used by both services (i think it makes sense SdJwtVcService depends on TokenStatusListService).

packages/core/src/modules/sd-jwt-vc/SdJwtVcModuleConfig.ts
```diff
@@ -0,0 +1,42 @@
+import { TokenStatusListRegistry } from './credential-status'
+import { HttpTokenStatusListRegistry } from './credential-status/token-status-list/http/HttpTokenStatusListRegistry'
+
+/**
+ * SdJwtVcModuleConfigOptions defines the interface for the options of the SdJwtVcModuleConfig class.
+ * This can contain optional parameters that have default values in the config class itself.
+ */
+export interface SdJwtVcModuleConfigOptions {
+  /**
+   * List of token status list registries that should be used by the dids module. The registrar must
```
TimoGlastra: still mentions dids

packages/core/src/modules/sd-jwt-vc/SdJwtVcModuleConfig.ts
```diff
@@ -0,0 +1,42 @@
...
+   * be an instance of the {@link TokenStatusListRegistry} interface.
+   *
+   * If no registries are provided, the default registries will be used. `PeerTokenStatusListRegistry` and `KeyTokenStatusListRegistry`
+   * will ALWAYS be registered, as they are needed for connections, mediation and out of band modules to function.
```
TimoGlastra: outdated

packages/cheqd/src/wrappers/CheqdTokenStatusListRegistry.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    const { didDocument } = await didResolver.resolve(agentContext, did, parsedDid)
+    if (!didDocument) throw new CredoError(`Unable to resolve DID Document for ${did}`)
+
+    const method = didDocument.verificationMethod?.find((vm) => vm.id === issuer.didUrl)
+    if (!method) throw new CredoError(`No verification method found for issuer ${issuer.didUrl}`)
+
+    // Upload to Cheqd
+    const resourcePayload: CheqdCreateResourceOptions = {
+      collectionId: did.split(':')[3],
```
TimoGlastra: can't we extract this from `parsedDid`?

packages/core/src/modules/sd-jwt-vc/credential-status/token-status-list/TokenStatusList.ts
DaevMithran: There could be multiple handlers for each uri according to regex is it? 

I was thinking for all http uri's they would be consistent and followed a pattern similar to did registrar, anyone can register a custom HTTPTokenStatusListRegistry and the different DID methods it can support.

packages/core/src/modules/sd-jwt-vc/credential-status/token-status-list/TokenStatusList.ts
DaevMithran: @TimoGlastra Could you share some references for `jwsSigner`

packages/core/src/modules/sd-jwt-vc/credential-status/token-status-list/TokenStatusList.ts
DaevMithran: @TimoGlastra Made the changes

packages/core/src/modules/sd-jwt-vc/credential-status/token-status-list/TokenStatusList.ts
Zzocker: In the test, did is being used for signing. Below is an example of the statusList created.
 ## Header
```json
{
  "alg": "EdDSA",
  "typ": "statuslist+jwt",
  "kid": "#key-1"
}
```

## Body
```
{
  "status_list": {
    "bits": 1,
    "lst": "eNpjYAAAAAIAAQ"
  },
  "iss": "did:cheqd:testnet:804f2bd3-ccb7-40bf-aec1-be71e4af525e",
  "iat": 1769542847
}
```

## DID Document
```json
{
    "@context": [
        "https://www.w3.org/ns/did/v1",
        "https://w3id.org/security/suites/jws-2020/v1"
    ],
    "id": "did:cheqd:testnet:804f2bd3-ccb7-40bf-aec1-be71e4af525e",
    "controller": [
        "did:cheqd:testnet:804f2bd3-ccb7-40bf-aec1-be71e4af525e"
    ],
    "verificationMethod": [
        {
            "id": "did:cheqd:testnet:804f2bd3-ccb7-40bf-aec1-be71e4af525e#key-1",
            "type": "JsonWebKey2020",
            "controller": "did:cheqd:testnet:804f2bd3-ccb7-40bf-aec1-be71e4af525e",
            "publicKeyJwk": {
                "crv": "Ed25519",
                "kty": "OKP",
                "x": "dle5ymF3KSXDnlKBvd0fB_CRjjZlRXV4yzhukuhguSU"
            }
        }
    ],
    "authentication": [
        "did:cheqd:testnet:804f2bd3-ccb7-40bf-aec1-be71e4af525e#key-1"
    ],
    "assertionMethod": [
        "did:cheqd:testnet:804f2bd3-ccb7-40bf-aec1-be71e4af525e#key-1"
    ]
}
```

PR #2327, opened 2025-07-03
Incorrect definition of field in OpenId4VcIssuerModule

tnotheis:
I assume that using a `=` at this place was just a mistake, and you actually wanted to use `:`.


tnotheis:
I'm not sure whether this should be declared as a chore or a fix. Feel free to change it however you like.

tnotheis:
Nvm, I just noticed that it was intended. 😄

PR #2329, opened 2025-07-08
made a start at mdoc update

TimoGlastra:


berendsliedrecht:
I am a bit biased, so maybe we need a review from @hacdias or @genaris

TimoGlastra:
> I think due to the `validityInfo.validUntil` being changed from optional to mandatory it is a breaking change :(.

Did it have a default before? I think it's required according to spec right? In that case it's more a fix? If there was a default we can just set that default from credo now until 0.7?

genaris:
> > I think due to the `validityInfo.validUntil` being changed from optional to mandatory it is a breaking change :(.
> 
> Did it have a default before? I think it's required according to spec right? In that case it's more a fix? If there was a default we can just set that default from credo now until 0.7?

I agree that it would be good if we can leave it as optional but setting a "safe" (if any) value, and marking it to be mandatory from 0.7

berendsliedrecht:
The output has changed quite a bit I have noticed from the tests as we did directly pass the mdoc libraries types to the end-user, meaning now that the library changed, we also have breaking changes. This should be added to 0.7.0 sadly.

TimoGlastra:
Okay, so then the question is, when can we push 0.7? 😬 if we just do it for the mdoc lib, it will be a really small update (which is probably a good thing!)

genaris:
> Okay, so then the question is, when can we push 0.7? 😬 if we just do it for the mdoc lib, it will be a really small update (which is probably a good thing!)

Yes, why not? If we can also add other dependency updates, notably the newer askar and anoncreds with koffi, will be a really great update. It would be good to say goodbye to the good-old patched ref-napi/ffi-napi that served for more than 3 years!

TimoGlastra:
Yeah i think a dependency upgrade release would be good 👍 

I think we still need to release anoncreds wrapper 0.4

berendsliedrecht:
I will do one more walk-through of the mdoc classes credo exposes, it should never expose the `owf/mdoc` types

TimoGlastra:
The device response verification now contains this:
 // NOTE: we split so that we can use individual trusted entities for each mdoc
const splittedDeviceResponses = this.splitIntoSingleDocumentResponses()

We are also doing that on the OID4VP layer (where we need it), but I don't think this is needed in the DeviceResponse class, did you add this @berendsliedrecht? it does not seems to be there in main branch

TimoGlastra:
Just triggered release of mdoc 0.6.0.

Done: https://github.com/openwallet-foundation-labs/mdoc-ts/releases/tag/v0.6.0

berendsliedrecht:
> The device response verification now contains this: // NOTE: we split so that we can use individual trusted entities for each mdoc const splittedDeviceResponses = this.splitIntoSingleDocumentResponses()
> 
> We are also doing that on the OID4VP layer (where we need it), but I don't think this is needed in the DeviceResponse class, did you add this @berendsliedrecht? it does not seems to be there in main branch

Didn't add that, no. I can remove it however :).

PR #2335, opened 2025-07-17
feat: support openid4vp v1

TimoGlastra:
Updates openid4vp to support OpenID4VP 1.0

Also adds support for redirect_uri (unsigned client id prefix)


packages/core/src/modules/dcql/DcqlService.ts
```diff
@@ -119,131 +190,248 @@ export class DcqlService {
...
 
-        return [credential_query_id, { ...result, all }]
+        return [credential_query_id, updatedResult]
       })
     )
 
     return {
       ...queryResult,
       credential_matches: matchesWithRecord,
-    }
```
hacdias, line 208: Why is `as DcqlQueryResult` necessary? If the types of the values are correct, it should not be necessary, right?

packages/core/src/modules/dcql/DcqlService.ts
```diff
@@ -321,83 +487,158 @@ export class DcqlService {
...
+      dcqlPresentation,
+      encodedDcqlPresentation,
     }
   }
 
   public getEncodedPresentations(dcqlPresentation: DcqlPresentation): DcqlEncodedPresentations {
-    return Object.fromEntries(Object.entries(dcqlPresentation).map(([key, value]) => [key, value.encoded]))
+    return Object.fromEntries(
+      Object.entries(dcqlPresentation).map(([key, value]) => [key, mapSingleOrArray(value, (v) => v.encoded)])
+    ) as DcqlEncodedPresentations
```
hacdias: Is the ` as DcqlEncodedPresentations` needed?

packages/openid4vc/src/openid4vc-holder/OpenId4vpHolderService.ts
```diff
@@ -372,9 +465,9 @@ export class OpenId4VpHolderService {
           presentationDefinition:
             authorizationRequestPayload.presentation_definition as unknown as DifPresentationExchangeDefinition,
           challenge: nonce,
-          domain: clientId,
+          domain: audience,
```
hacdias, line 454: Just to double check: does this function expect `domain` to be prefixed with `origin:` in newer drafts?

packages/openid4vc/src/openid4vc-holder/OpenId4vpHolderService.ts
```diff
@@ -401,11 +494,26 @@ export class OpenId4VpHolderService {
       const { encodedDcqlPresentation } = await this.dcqlService.createPresentation(agentContext, {
         credentialQueryToCredential: credentialsWithTransactionData,
         challenge: nonce,
-        domain: clientId,
-        openid4vp: openid4vpOptions,
+        domain: audience,
```
hacdias, line 483: Same question here.

packages/openid4vc/src/openid4vc-holder/OpenId4vpHolderServiceOptions.ts
```diff
@@ -64,6 +68,25 @@ export interface OpenId4VpResolvedAuthorizationRequest {
...
+     * The effective client id, taking into account default values and different draft versions.
+     */
+    effectiveClientId: ResolvedOpenid4vpAuthorizationRequest['client']['effective']
+
+    /**
+     * The client id metadata.
+     *
+     * In case of 'openid_federation' client id prefix, this will be the metadata from the federation.
+     * clientMetadata?: ClientMetadata
+     */
```
hacdias, line 87: Shouldn't this line be uncommented? Or is this not yet implemented?

packages/openid4vc/src/openid4vc-verifier/OpenId4VpVerifierService.ts
```diff
@@ -308,20 +338,28 @@ export class OpenId4VpVerifierService {
...
+            this.decodePresentation(agentContext, {
+              presentation,
+              format:
+                queryCredential.format === 'mso_mdoc'
+                  ? ClaimFormat.MsoMdoc
+                  : queryCredential.format === 'dc+sd-jwt' || queryCredential.format === 'vc+sd-jwt'
+                    ? ClaimFormat.SdJwtVc
+                    : queryCredential.format === 'jwt_vc_json'
+                      ? ClaimFormat.JwtVp
+                      : ClaimFormat.LdpVp,
```
hacdias: Not the biggest fan of such long ternary operators. I've seen this in other places too, but I wonder if a static object mapping the format string to the `ClaimFormat` wouldn't make this much more readable.

packages/openid4vc/src/openid4vc-verifier/OpenId4VpVerifierService.ts
```diff
@@ -308,20 +338,28 @@ export class OpenId4VpVerifierService {
...
+            })
+          ),
         ]
       })
-    )
+    ) as DcqlPresentation
```
hacdias: I would prefer to have this in the return type of the function than using an `as`. I've had cases in the past where asserting to a type forcefully was actually hiding errors about missing mandatory fields:

```ts
type Test = {
    a: string
    b: number
}

// No error, but faulty definition
const x = {
    a: 'Test'
} as Test

// Correct error
const y: Test = {
    a: 'Test'
}
```

This comment applies to all other places where there's a type assertion.

packages/openid4vc/src/openid4vc-verifier/OpenId4VpVerifierService.ts
```diff
@@ -445,51 +497,65 @@ export class OpenId4VpVerifierService {
...
+              presentations.map((presentation) =>
+                this.verifyPresentation(agentContext, {
+                  format:
+                    queryCredential.format === 'mso_mdoc'
+                      ? ClaimFormat.MsoMdoc
+                      : queryCredential.format === 'dc+sd-jwt' || queryCredential.format === 'vc+sd-jwt'
+                        ? ClaimFormat.SdJwtVc
+                        : queryCredential.format === 'jwt_vc_json'
+                          ? ClaimFormat.JwtVp
+                          : ClaimFormat.LdpVp,
```
hacdias: Same comment as before.

hacdias (commented):
Just some questions (because I'm not yet very knowledgeable about this codebase), some nits and some assertion concerns.

packages/core/src/modules/dcql/DcqlService.ts
TimoGlastra: It's to do with mapping NonEmptyArray (map loses the NonEmpty). I added a new `mapNonEmptyArray` function that retains this

packages/openid4vc/src/openid4vc-holder/OpenId4vpHolderService.ts
TimoGlastra: Yes! The client id prefix needs to be part of the audience

packages/openid4vc/src/openid4vc-holder/OpenId4vpHolderService.ts
TimoGlastra: Yes!

packages/openid4vc/src/openid4vc-holder/OpenId4vpHolderServiceOptions.ts
TimoGlastra: It's only implemented in #2093, so that's why it's not added yet. (i will integrate it into that PR)

packages/openid4vc/src/openid4vc-verifier/OpenId4VpVerifierService.ts
TimoGlastra: Again due to non empty array types and mapping. Updated it

TimoGlastra:
@hacdias i addressed your feedback

hacdias (approved):

PR #2337, opened 2025-07-26
fix: use compressed legacy key id

TimoGlastra:
Legacy key ids uses compressed form for EC keys


packages/core/src/modules/kms/jwk/kty/okp/Ed25519PublicJwk.ts
```diff
@@ -26,6 +26,13 @@ export class Ed25519PublicJwk implements PublicJwkType<Jwk> {
     }
   }
 
+  /**
+   * Not supported for Ed25519
+   */
+  public get compressedPublicKey() {
+    return null
```
berendsliedrecht, line 33: Ed25519 public keys can be compressed, any reason why it is null?

TimoGlastra: Hmm okay. Did we use the compressed key for Ed25519 keys as well for the Askar key name? I don't think so right?

TimoGlastra: And do we have a way to return the compressed key here? Since the EC compression lib only works for EC keys right?

berendsliedrecht: ed25519 is an ec key, however it is based on a twisted edwards curve, not a weierstrass curve like secp. As long as the public key is a point on the elliptic curve (which is the case for ed25519, p256, etc.), it is compressable. 

However, the compression of ed25519 is different from p256. So the library, currently, does not support compression of ed25519 keys.

With Askar, we did use compressed ed25519 keys (I think uncompressed form is way more uncommon compared to secp).

askar:

https://github.com/openwallet-foundation/askar/blob/0b919bb58205416df7cd2cc1a24ce52373cbe105/askar-crypto/src/alg/ed25519.rs#L38

and that is:

https://docs.rs/ed25519-dalek/latest/ed25519_dalek/constant.PUBLIC_KEY_LENGTH.html

berendsliedrecht: We should add the compression/decompression to the library, but since it is almost always compressed, we could also check the length to be 32 and if so, return the key otherwise, null.

TimoGlastra: Hmm okay but that complicates things, because currently `publicKey` assumes uncpompressed. But maybe we should make `publicKey` the default encoding, and we add both `compressedPublicKey` and `uncompressedPublicKey`?

berendsliedrecht: publicKey to default encoding makes sense, we just have to add it  as a tsdoc to the class. 

For secp it is a bit difficult for me to say which is the default encoding, so IMO whatever is easier for credo, for ed/x25519 we can just do compressed and for RSA we do uncompressed.

PR #2339, opened 2025-07-28
feat: did:webvh resolver for legacy 0.5.16 credo

brianorwhatever:
This removes the use of Kms and uses Askar wallet instead


packages/webvh/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "compile": "tsc -p tsconfig.build.json",
+    "prepublishOnly": "pnpm run build",
+    "test": "jest",
+    "test:debug": "node --inspect-brk node_modules/.bin/jest --runInBand --config jest.config.js"
+  },
+  "dependencies": {
+    "@credo-ts/anoncreds": "workspace:*",
+    "@credo-ts/askar": "workspace:*",
+    "@credo-ts/core": "workspace:*",
+    "@credo-ts/node": "workspace:*",
```
genaris: This dependency might imply issues when running on React Native. Are you sure you need it?

packages/webvh/src/dids/WebvhDidResolver.ts
```diff
@@ -0,0 +1,128 @@
...
+    try {
+      agentContext.config.logger.debug(`Attempting to resolve resource: ${resourceUrl}`)
+      
+      // Use the getBaseUrl method to parse the did:webvh resource URL
+      const httpsUrl = this.getBaseUrl(resourceUrl)
+      
+      agentContext.config.logger.debug(`Fetching resource from: ${httpsUrl}`)
+      
+      // Fetch the resource using native fetch
+      const response = await fetch(httpsUrl)
```
genaris:
```suggestion
      const response = await agentContext.config.agentDependencies.fetch(httpsUrl)
```

genaris (changes requested):

packages/webvh/package.json
genaris: @brianorwhatever can you check this one? I'm afraid that it could cause issues in React Native apps like the ones based on Bifold

brianorwhatever: I am using `new AskarWallet(` which requires `new agentDependencies.FileSystem()` as a parameter which is the only time this is being used. I think I'm either not supposed to use AskarWallet as I am or there is a way to get around this?

genaris: What about using this.agentContext.wallet instead?

brianorwhatever: @genaris Thanks! that worked great

packages/webvh/package.json
```diff
@@ -0,0 +1,38 @@
...
+    "build": "pnpm run clean && pnpm run compile",
+    "clean": "rimraf -rf ./build",
+    "compile": "tsc -p tsconfig.build.json",
+    "prepublishOnly": "pnpm run build",
+    "test": "jest",
+    "test:debug": "node --inspect-brk node_modules/.bin/jest --runInBand --config jest.config.js"
+  },
+  "dependencies": {
+    "@credo-ts/anoncreds": "workspace:*",
+    "@credo-ts/askar": "workspace:*",
```
genaris: I think we can also remove askar dependency now we are using Agent's wallet, right?

brianorwhatever: yep good call! thanks

genaris (approved):
Thanks @brianorwhatever ! Can you fix the lint error? With that I think we are good to go.

brianorwhatever:
@genaris I think we are good to go now. I had to make some changes to the cheqd package in [13d04db](https://github.com/openwallet-foundation/credo-ts/pull/2339/commits/13d04dbb423dbb72645b00ff277bc452fdf0be29).. not sure if I should have done that but couldn't get the green checks without

PR #2343, opened 2025-08-01
fix: incorrect DCQL query in OpenID4VC demo

Magnus-Kuhn:
In the OpenID4VC demo, presentation with DCQL failed with the holder unable to handle the request.

Two corrections to the DCQL query in the demo are made:
* The `id` is changed to not use the forbidden character `+`
* The credential type query is changed to match the credential type issued in the demo


genaris (commented):
Thanks @Magnus-Kuhn ! Can you fix the DCO issue?

PR #2348, opened 2025-08-06
feat: emit rotation event on hangup

rmlearney-digicatapult:
At present, `processHangup` succeeds without emitting any event that can be captured by an external component

Added standard `emitDidRotatedEvent`


rmlearney-digicatapult:
Non-semantic single commit

PR #2349, opened 2025-08-06
feat: emit rotation event on hangup

rmlearney-digicatapult:
At present, `processHangup` succeeds without emitting any event that can be captured by an external component

Added standard `emitDidRotatedEvent` so the rotation to an undefined did can be captured and processed as a disconnection event in a client


rmlearney-digicatapult:
Having some issues with 'files changed = 29' after merging latest  `v0.5.x` :(

genaris:
> Having some issues with 'files changed = 29' after merging latest `v0.5.x` :(

I think it is OK again now. Ready to update some tests :smile:

PR #2350, opened 2025-08-07
feat: support deferred credential issuance

hacdias:
This PR adds support fopr deferred credential issuance within the OpenID4VCI specification. Some notes:

- Built on top of #2335. To look at the correct diff of this PR check: https://github.com/TimoGlastra/aries-framework-javascript/compare/feat/oid4vp-1.0...hacdias:credo-ts:feat/deferred-credential-issuance?expand=1
- Requires https://github.com/openwallet-foundation-labs/oid4vc-ts/pull/86 to first be merged, released, and then bumped here.

There are still a few small questions left to answer:

1. ~~Are deferred and partial issuance compatible with each other? I assumed that no, since the specification was not clear enough for me. However, if they **are** compatible, I need to update the code to allow multiple transaction IDs per issuance session.~~
2. I'm not sure if we need to change something in the token creation logic. Do we need to make them live longer?
3. ~~I did not change anything in the validity of an issuance session. On the deferred credential issuance endpoint, I just do not check for the expiration time, since it's no longer  relevant. Is there any cleanup of the issuance session repository? Do I need to make sure it does not get deletedr in some way?~~


.changeset/ripe-lights-appear.md
```diff
@@ -0,0 +1,5 @@
+---
+"@credo-ts/openid4vc": minor
+---
+
+Add support for deferred credential issuance according to Draft 16 of the OpenID for Verifiable Credential Issuance specification.
```
TimoGlastra: Do we also add support for Draft 16? And if so can we add a changeset for that

hacdias: Looking at the document history, I think we are compatible with Draft 16. The only thing we don't support is fetching (and serving) signed metadata, which is not mandatory: https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-signed-metadata.

hacdias: I can quickly take a look and add support for that on `oid4vc`

hacdias: I updated the changeset to mention that we support Draft 16.

hacdias:
@TimoGlastra I'm opening this for review. Please note that there are still some open questions that I was hoping getting some feedback on: https://github.com/openwallet-foundation/credo-ts/pull/2350#issue-3299772172

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
```diff
@@ -488,13 +491,101 @@ export class OpenId4VciHolderService {
...
+        ? {
+            ...options.dpop,
+            nonce: dpopNonce,
+          }
+        : undefined,
+      cNonce,
+    }
+  }
+
+  public async acceptDeferredCredentialOffer(
```
TimoGlastra: You alraedy accepted the credential offer in this case right, so should it be just `retrieveDeferredCredentials`?

packages/openid4vc/src/openid4vc-issuer/repository/OpenId4VcIssuanceSessionRecord.ts
```diff
@@ -134,6 +147,16 @@ export interface OpenId4VcIssuanceSessionRecordProps {
    */
   presentation?: OpenId4VcIssuanceSessionPresentation
 
+  /**
+   * When the issuance is deferred, this is the transaction id that is used to
+   * link the deferred credential requests to the issuance session.
+   */
+  // TODO: I'm assuming it is not possible to have multiple deferred credential
+  // requests within the same issuance session. If that is not the case, this needs
+  // to be reworked.
```
TimoGlastra: i think it's possible theoretically? If you offer both drivers license and PID and both need a few days to process? If we want to allow it is another question, I'm not sure 🤔

packages/openid4vc/src/openid4vc-issuer/repository/OpenId4VcIssuanceSessionRecord.ts
```diff
@@ -134,6 +147,16 @@ export interface OpenId4VcIssuanceSessionRecordProps {
...
+  transactionId?: string
+  transactionData?: OpenId4VcIssuanceSessionRecordTransactionData
```
TimoGlastra: Maybe we can make the transactionId part of the transactionData and make it a required param? That way if `transactionData.transactionId` is defined we also know transactionData is defined

TimoGlastra: we can also make it an array just in case and only allow a single one for now, but then at least the data structure will stay the same?

hacdias: Yes, I can do that :) I had done it this way before I noticed that we filter (with the query) on the "tags" 😅 Initially, I assumed it was on the fields themselves.

packages/openid4vc/src/openid4vc-issuer/repository/OpenId4VcIssuanceSessionRecord.ts
hacdias: Yeah, exactly, theoretically seems like it should be possible, but whether or not it will actually happen, is another question.

I can take a second look and see if I can enable this to work with multiple transactions.

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuanceSessionState.ts
```diff
@@ -12,6 +12,7 @@ export enum OpenId4VcIssuanceSessionState {
 
   CredentialRequestReceived = 'CredentialRequestReceived',
   CredentialsPartiallyIssued = 'CredentialsPartiallyIssued',
+  Deferred = 'Deferred',
```
TimoGlastra: If we have one deferred and one issued, will the state be CredentialsPartiallyIssued or Deferred?

hacdias: I will rework this to work with multiple deferrals, which also means that I will be ditching this state (https://github.com/openwallet-foundation/credo-ts/pull/2350#discussion_r2266741065).

At the moment, you could only have a single deferred credential with the request and nothing else.

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerService.ts
```diff
@@ -324,42 +327,184 @@ export class OpenId4VcIssuerService {
...
+        expectedLength: options.issuanceSession.transactionData.numberOfCredentials,
+      })
+
+      deferredCredentialResponse = vcIssuer.createDeferredCredentialResponse({
+        credentials: credentials.credentials,
+      })
+
+      issuanceSession.issuedCredentials.push(credentialConfigurationId)
+
+      await this.updateState(agentContext, issuanceSession, OpenId4VcIssuanceSessionState.Completed)
```
TimoGlastra: Completed does not take into account multipoe configurations being offered in a single offer, do we only allow a single credential configuration id if using deferred (i don't think we do since you only trigger the deferred flow when the credential request to credential mapper is called, so this should match the logic of the normal flow and set to partially issued or completed?)

hacdias: See above: https://github.com/openwallet-foundation/credo-ts/pull/2350#discussion_r2266759406

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerServiceOptions.ts
```diff
@@ -273,9 +288,43 @@ export interface OpenId4VciCredentialRequestToCredentialMapperOptions {
...
+  deferredCredentialRequest: OpenId4VciDeferredCredentialRequest
+}
+
+export type OpenId4VciDeferredCredentialRequestToCredentialMapper = (
+  options: OpenId4VciDeferredCredentialRequestToCredentialMapperOptions
+) =>
+  | Promise<OpenId4VciSignCredentials>
+  | OpenId4VciSignCredentials
+  | Promise<OpenId4VciDeferredCredentials>
+  | OpenId4VciDeferredCredentials
```
TimoGlastra: Maybe we can add a `type` (or maybe a less overused term) that we can use to detect sign/defer. That way we don't have to check for `'interval' in signOrDefer` and it's also more explicit from the user in the mapper that the credential issuance is deferred.

packages/openid4vc/src/openid4vc-issuer/repository/OpenId4VcIssuanceSessionRecord.ts
```diff
@@ -97,6 +98,18 @@ export type DefaultOpenId4VcIssuanceSessionRecordTags = {
...
+  // The expected number of credentials that will be issued in this transaction
+  numberOfCredentials: number
+
+  // The format of the credential request that was used to create the issuance session
+  requestFormat?: CredentialRequestFormatSpecific
+
+  // The credential configuration and its id
+  credentialConfigurationId?: string
+  credentialConfiguration?: CredentialConfigurationSupportedWithFormats
+}
```
TimoGlastra: Do we need to store the configuration data on this record? I think all other flows will fail if you update a configuration during the process of issuance, but this handles that by storing the configuration. 

I'm wondering if we should, and if so, whether we don't want to make it generic? Just some thoughts

packages/openid4vc/src/openid4vc-issuer/repository/OpenId4VcIssuanceSessionRecord.ts
```diff
@@ -97,6 +98,18 @@ export type DefaultOpenId4VcIssuanceSessionRecordTags = {
   presentationAuthSession?: string
 }
 
+export interface OpenId4VcIssuanceSessionRecordTransactionData {
+  // The expected number of credentials that will be issued in this transaction
+  numberOfCredentials: number
+
+  // The format of the credential request that was used to create the issuance session
+  requestFormat?: CredentialRequestFormatSpecific
```
TimoGlastra: same here, do we need to store this?

packages/openid4vc/src/openid4vc-issuer/repository/OpenId4VcIssuanceSessionRecord.ts
```diff
@@ -97,6 +98,18 @@ export type DefaultOpenId4VcIssuanceSessionRecordTags = {
...
+
+  // The credential configuration and its id
+  credentialConfigurationId?: string
```
TimoGlastra: I think we should store the configuration id for the transaction (and so make this required), so we know which configuration id the deferred issuance is assocaited with.

packages/openid4vc/src/openid4vc-issuer/repository/OpenId4VcIssuanceSessionRecord.ts
hacdias: I don't think it is strictly necessary, no. I can remove it.

packages/openid4vc/src/openid4vc-issuer/repository/OpenId4VcIssuanceSessionRecord.ts
hacdias: This is the answer for the past 3 questions: I'm not sure.

They're all used when calling `getCredentialConfigurationsForRequest`, whose result is needed to call `getSignedCredentials`. But if we remove the `requestFormat` **and** the `credentialConfiguration`, we the first call will fail.

My initial idea was to preserve the data from the original credential request.

This is also an answer to your other 3 questions.

hacdias: Because we need the credential configuration to subsequently call `getSignedCredentials`

packages/openid4vc/src/openid4vc-issuer/repository/OpenId4VcIssuanceSessionRecord.ts
hacdias: See https://github.com/openwallet-foundation/credo-ts/pull/2350#discussion_r2266804133

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuanceSessionState.ts
TimoGlastra: We should probably keep state somewhere (like we have with the issuedCredentials array), so we know if the request was already processed or not)

TimoGlastra:
> Are deferred and partial issuance compatible with each other? I assumed that no, since the specification was not clear enough for me. However, if they are compatible, I need to update the code to allow multiple transaction IDs per issuance session.


Partial issuance is something we added in Credo. It's if you offer multiple configurations in a offer, and only a subset of those have been issued. Every issuance flow that offers multiple credentials will reach this state, so yes I'd say we have to make them compatible with each other.

> I'm not sure if we need to change something in the token creation logic. Do we need to make them live longer?

This is a good question. We could make them live longer, but i think we should also consider to add support for refresh tokens. (this will enable credential refresh, something we want to support as well). Using a refresh token you can get a new access token.

Some considerations to this:
- We should add the option to issue a refresh token that is also bound to the `pre-authorized_code` or `issuer_state` and has the same `sub`.
- For deferred issuance we should probably require refresh tokens
- We should allow exchanging a refresh token (with same DPoP and Client Attestation) for an access token

> I did not change anything in the validity of an issuance session. On the deferred credential issuance endpoint, I just do not check for the expiration time, since it's no longer relevant. Is there any cleanup of the issuance session repository? Do I need to make sure it does not get deletedr in some way?

We use it in e.g. Paradym to remove expired sessions. So i think we should update this as well. But we probably want to have longer expiration for the deferred issuance than the issuance session as a whole 🤔

packages/openid4vc/src/openid4vc-issuer/repository/OpenId4VcIssuanceSessionRecord.ts
hacdias: See https://github.com/openwallet-foundation/credo-ts/pull/2350#discussion_r2266804133

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuanceSessionState.ts
hacdias: Definitely. I just meant removing it as a state. See [`ca620f4` (#2350)](https://github.com/openwallet-foundation/credo-ts/pull/2350/commits/ca620f4125dd7c05e2afb351279d570142bc5209)

packages/openid4vc/src/openid4vc-issuer/repository/OpenId4VcIssuanceSessionRecord.ts
TimoGlastra: I think we should store the `credentialConfigurationId` and retrieve the `credentialConfiguration` from the stored issuer metadata (by calling issuerService.getIssuerMetadata). That way we don't need `requestFormat`, and we can minimze the amount of data we have to store for each transaction

TimoGlastra:
Also mainly a question on the overall flow. If you want to use deferred issuance in credo you should:
- implement the mapper to return deferred credential (an interval)
- store the keys/attestation associated with the request outside of credo for processing, along with a transaction id
- Implement the deferred mapper to return the credential at a later stage?

hacdias:
> implement the mapper to return deferred credential (an interval)

Yes, and transaction id. I considered generating the transaction id in Credo itself, but that'd make it more complicated, since we'd need to "give" that information back to the implementer.

> store the keys/attestation associated with the request outside of credo for processing, along with a transaction id

Yes.

> Implement the deferred mapper to return the credential at a later stage?

Yes

TimoGlastra:
> I considered generating the transaction id in Credo itself, but that'd make it more complicated, since we'd need to "give" that information back to the implementer.

Makes sense, users just need to be careful to genreate unique transaction ids since it can cause issues if they clash

packages/openid4vc/src/openid4vc-issuer/repository/OpenId4VcIssuanceSessionRecord.ts
hacdias: I've just improved this and only retain the config id.

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerServiceOptions.ts
hacdias: I added `type`: `credentials` or `deferral`

hacdias:
> We use it in e.g. Paradym to remove expired sessions. So i think we should update this as well. But we probably want to have longer expiration for the deferred issuance than the issuance session as a whole 🤔

At the moment, `issuanceSession` does not really have an "expiry date". We just use `createdAt` with `statefulCredentialOfferExpirationInSeconds`.

**Update**: I introduced a `expiresAt` variable, which is initially set to `createdAt + statefulCredentialOfferExpirationInSeconds` (preserving current behavior). Every time it gets deferred, we change `expiresAt` to be the maximum between:

1. Current `expiresAt`
2. Current time + interval * 2
3. Current time + statefulCredentialOfferExpirationInSeconds

This accounts for multiple deferred credentials with potentially longer intervals, hence we keep checking what's currently set as expiry time. `3` is just a measure to not allow it to be too short, and allow time for the wallet to get a new token and fetch the creedentials.

hacdias:
> but i think we should also consider to add support for refresh tokens. (this will enable credential refresh, something we want to support as well). Using a refresh token you can get a new access token.

I will work on adding support for refresh tokens on https://github.com/openwallet-foundation-labs/oid4vc-ts

TimoGlastra:
> > but i think we should also consider to add support for refresh tokens. (this will enable credential refresh, something we want to support as well). Using a refresh token you can get a new access token.
> 
> 
> 
> I will work on adding support for refresh tokens on https://github.com/openwallet-foundation-labs/oid4vc-ts

We can also add it in another PR, since we won't have to change the logic for deferred issuance for it

hacdias:
@TimoGlastra yes, let's merge this one first. 

I will work on the refresh tokens in. a separate PR.

packages/openid4vc/src/openid4vc-issuer/repository/OpenId4VcIssuanceSessionRecord.ts
```diff
@@ -147,6 +161,12 @@ export class OpenId4VcIssuanceSessionRecord extends BaseRecord<DefaultOpenId4VcI
   public static readonly type = 'OpenId4VcIssuanceSessionRecord'
   public readonly type = OpenId4VcIssuanceSessionRecord.type
 
+  /**
+   * Expiry time for the issuance session. This can change dynamically during
+   * the session lifetime, based on the possible deferrals.
+   */
+  public expiresAt!: Date
```
```suggestion
  /**
   * Expiry time for the issuance session. This can change dynamically during
   * the session lifetime, based on the possible deferrals.
   * 
   * @since 0.6
   */
  public expiresAt?: Date
```

Should be optional, since previous records won't have this value set

packages/openid4vc/src/openid4vc-issuer/router/credentialEndpoint.ts
```diff
@@ -246,7 +243,12 @@ export function configureCredentialEndpoint(router: Router, config: OpenId4VcIss
         )
       }
 
+      const createdAt = new Date()
+      const expiresAt = addSecondsToDate(createdAt, config.statefulCredentialOfferExpirationInSeconds)
```
TimoGlastra, line 251: Do we handle all places where we handle expiration? Since we also have authorization code expiration etc.

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerServiceOptions.ts
```diff
@@ -273,33 +296,68 @@ export interface OpenId4VciCredentialRequestToCredentialMapperOptions {
...
   options: OpenId4VciCredentialRequestToCredentialMapperOptions
-) => Promise<OpenId4VciSignCredentials> | OpenId4VciSignCredentials
+) => CanBePromise<OpenId4VciSignCredentials> | CanBePromise<OpenId4VciDeferredCredentials>
+
+export interface OpenId4VciDeferredCredentialRequestToCredentialMapperOptions {
+  agentContext: AgentContext
+
+  /**
+   * Authorization associated with the credential request
+   * TODO: needed?
```
TimoGlastra: Not sure, do we add it in the normal `CredentialRequestToCredentialMapperOptions`?

hacdias: We do

packages/openid4vc/src/openid4vc-issuer/router/credentialEndpoint.ts
hacdias: Yes,  check the last commit

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerService.ts
```diff
@@ -324,42 +332,210 @@ export class OpenId4VcIssuerService {
...
+          dcql: response.dcql,
+        }
+      } else {
+        throw new CredoError(
+          `Verified authorization response for verification session with id '${session.id}' does not have presentationExchange or dcql defined.`
+        )
+      }
+    }
+
+    const signOptionsOrDeferral = await mapper({
```
TimoGlastra, line 369: Do we handle somewhere that a credential cannot be requested multiple times while it's deferred? I think we some logic in place that we check that a requeted credential is not already in `issuedCredentials` on the record, but since we don't add it until after the credential is issued we should check that there's also not a transaction for this configuration id in place.

hacdias: Fixed https://github.com/openwallet-foundation/credo-ts/pull/2350/commits/bfbf4110676fbb7c7f7531fd379fa0c9236f36b1

hacdias:
@TimoGlastra just rebased with latest updates from `main` :)

Sshovon:
Hey @hacdias , I ran into a couple of runtime errors while testing credential retrieval from an offer. Initially I got, 
```
TypeError: expiresAt.getTime is not a function.
``` 
 After converting `expiresAt` to a Date with `new Date(expiresAt).getTime()`, the next error appeared:
 ```
 TypeError: options.preAuthorizedCodeExpiresAt.getTime is not a function
 ```
It looks like in some code paths these fields are `strings` rather than `Date` instances. 
I think these issues are connected with this PR. Please ignore if they are not. Thanks.

hacdias:
@Sshovon thanks, I think this should fix it: https://github.com/openwallet-foundation/credo-ts/pull/2357

PR #2353, opened 2025-08-11
feat: (main) decline request

rmlearney-digicatapult:
Linked to issue https://github.com/openwallet-foundation/credo-ts/issues/2338

Have created a method in ConnectionsAPI.ts for gracefully declining an incoming connection request in the request-received state by sending a problem report to the requester and marking the connection as `abandoned` on our side. The OOB invite state is also transitioned to `Done` on our side if not reusable to prevent replay.

Submitting for your review and insights


packages/didcomm/src/modules/connections/ConnectionsApi.ts
```diff
@@ -220,6 +223,75 @@ export class ConnectionsApi {
...
+    if (connectionRecord.protocol === HandshakeProtocol.DidExchange) {
+      await this.didExchangeProtocol.createResponse(
+        this.agentContext,
+        connectionRecord,
+        outOfBandRecord,
+        routing
+      )
+    } else {
+      throw new CredoError(`Connection protocol is ${connectionRecord.protocol} not didexchange`)
+    }
```
TimoGlastra, line 268: I'm not sure we should trigger the whole response creation logic just to decline the request? Since we don't send the response, we don't send the signed attachments from the response, and thus it won't be possible for the other party to verify the problem report came from the inviter.

The only way to do that is by wrapping the problem report using the same key that was included in the invitation.

packages/didcomm/src/modules/connections/__tests__/connection-manual.test.ts
```diff
@@ -147,9 +170,82 @@ describe('Manual Connection Flow', () => {
...
+    const [bobMessage] = bobSpy.mock.calls[0]
+
+    expect(bobSpy2).toHaveBeenCalled()
+    const [spy2] = bobSpy2.mock.calls
+    // expect(bobConnectionSpy).toHaveBeenCalled()
+    // const [spyMessage] = bobConnectionSpy.mock.calls[0]
+    
+    console.debug(aliceMessage)
+    console.log(bobMessage)
+    console.log(spy2)
```
TimoGlastra, line 234: This needs to be clean up. I am curious, how is the decline request currently handled by Bob? Since it should not allow the problem report to be assocaited with the invitation/connection, since we can't verify whether the sender of the problem report is the inviter based on the current code in declineRequest

TimoGlastra (changes requested):
Thanks for this PR, i have some concerns on the security of this implementation

packages/didcomm/src/modules/connections/__tests__/connection-manual.test.ts
rmlearney-digicatapult: The tests were WIP, but I agree the problem report must contain something known to both Alice and Bob which would either be the `invitationId` from the OOB record, or the `threadId` from the connectionRecord. I set the threadId of the problem report in the `declineRequest` method to equal that of the connectionRecord.

packages/didcomm/src/modules/connections/ConnectionsApi.ts
rmlearney-digicatapult: Are you able to give me any pointers on how to do that more efficiently? I assumed the creation of the outbound message would automatically wrap/sign the message with the correct keys.

PR #2354, opened 2025-08-11
fix(webvh): eddsa-jcs-2022 transformation algorithm

PatStLouis:


packages/webvh/src/anoncreds/services/WebVhAnonCredsRegistry.ts
```diff
@@ -479,32 +486,18 @@ export class WebVhAnonCredsRegistry implements AnonCredsRegistry {
...
-      }
-
-      // Canonicalize the document using JCS
-      const canonicalizedDocument = canonicalize(documentToVerify)
-      const documentBytes = TypedArrayEncoder.fromString(canonicalizedDocument)
+      // For eddsa-jcs-2022 hashing algorithm:
+      // https://www.w3.org/TR/vc-di-eddsa/#hashing-eddsa-jcs-2022
+      const proofConfig = (({ proofValue, ...o }) => o)(proof);
+      const proofConfigHash = HashData(canonicalize(proofConfig));
+      const transformedDocumentHash = HashData(canonicalize(unsecuredDocument));
```
TimoGlastra: We have a hasher already in utils, which doesn't add another dependency

packages/webvh/src/anoncreds/services/WebVhAnonCredsRegistry.ts
```diff
@@ -103,7 +104,8 @@ export class WebVhAnonCredsRegistry implements AnonCredsRegistry {
       agentContext.config.logger.trace(`Resource ${resourceId} attestation found.`)
 
       // 2. Check for valid proof
-      const isProofValid = await this.verifyProof(agentContext, resourceObject.proof, resourceObject.content)
+      const unsecuredDocument = (({ proof, ...o }) => o)(resourceObject);
```
TimoGlastra: Isn't this the same as

```suggestion
      const { proof, ...unsecuredDocument } =  resourceObject
```

packages/webvh/src/anoncreds/services/WebVhAnonCredsRegistry.ts
```diff
@@ -479,32 +486,18 @@ export class WebVhAnonCredsRegistry implements AnonCredsRegistry {
...
-          created: proof.created || new Date().toISOString(),
-        },
-      }
-
-      // Canonicalize the document using JCS
-      const canonicalizedDocument = canonicalize(documentToVerify)
-      const documentBytes = TypedArrayEncoder.fromString(canonicalizedDocument)
+      // For eddsa-jcs-2022 hashing algorithm:
+      // https://www.w3.org/TR/vc-di-eddsa/#hashing-eddsa-jcs-2022
+      const proofConfig = (({ proofValue, ...o }) => o)(proof);
```
TimoGlastra: same here:
```suggestion
      const { proofValue, ...proofConfig } = proof;
```

packages/webvh/src/anoncreds/services/WebVhAnonCredsRegistry.ts
PatStLouis: thanks for the comment, this is my first time with the credo project so I'm getting familiar with the codebase, I have found the existing hasher function along with other things, I'll make sure to make this PR as consistent with the rest of the project as possible please bear with me!

packages/webvh/src/anoncreds/services/WebVhAnonCredsRegistry.ts
```diff
@@ -479,32 +486,18 @@ export class WebVhAnonCredsRegistry implements AnonCredsRegistry {
...
+      const proofConfigHash = HashData(canonicalize(proofConfig));
+      const transformedDocumentHash = HashData(canonicalize(unsecuredDocument));
+      const hashData = proofConfigHash+transformedDocumentHash;
```
genaris: I'm not completely sure about the procedure you need to achieve with the hashes but I think this code may help you to reach it without any external dependency:

```suggestion
      const proofConfigHash = Hasher.hash(canonicalize(proofConfig), 'sha-256')
      const transformedDocumentHash = Hasher.hash(canonicalize(unsecuredDocument), 'sha-256')
      const hashData = Buffer.concat([proofConfigHash, transformedDocumentHash])
```

You just need to import Hasher from @credo-ts/core.

genaris (changes requested):

PatStLouis:
@genaris @TimoGlastra I'm having an issue with the askar js wrapper. When I try to instantiate a key from a public key bytes with the `ed25519` algorithm, i get `AskarError: Unknown key algorithm`. I can't reproduce it locally, this error is only raised when using credo. You can see the line here:
https://github.com/OpSecId/credo-ts/blob/e39b68dc83dbe7ff1eef23caa71568798315ee13/packages/webvh/src/cryptosuites/eddsa-jcs-2022.ts#L94

Looking at the enum from the library, it should be supported? Any pointers?

Here's the error trace:
```
      AskarError: Unknown key algorithm
          at NodeJSAskar.getAskarError (/home/ubuntu/repos/credo-ts/node_modules/.pnpm/@openwallet-foundation+askar-nodejs@0.3.2/node_modules/@openwallet-foundation/askar-nodejs/src/NodeJSAskar.ts:216:12)
          at NodeJSAskar.handleError (/home/ubuntu/repos/credo-ts/node_modules/.pnpm/@openwallet-foundation+askar-nodejs@0.3.2/node_modules/@openwallet-foundation/askar-nodejs/src/NodeJSAskar.ts:222:16)
          at NodeJSAskar.keyFromPublicBytes (/home/ubuntu/repos/credo-ts/node_modules/.pnpm/@openwallet-foundation+askar-nodejs@0.3.2/node_modules/@openwallet-foundation/askar-nodejs/src/NodeJSAskar.ts:587:10)
          at Function.fromPublicBytes (/home/ubuntu/repos/credo-ts/node_modules/.pnpm/@openwallet-foundation+askar-shared@0.3.2/node_modules/@openwallet-foundation/askar-shared/src/crypto/Key.ts:40:26)
          at EddsaJcs2022Cryptosuite.proofVerification (/home/ubuntu/repos/credo-ts/packages/webvh/src/cryptosuites/eddsa-jcs-2022.ts:95:21)
          at async EddsaJcs2022Cryptosuite.verifyProof (/home/ubuntu/repos/credo-ts/packages/webvh/src/cryptosuites/eddsa-jcs-2022.ts:119:22)
          at async WebVhAnonCredsRegistry.verifyProof (/home/ubuntu/repos/credo-ts/packages/webvh/src/anoncreds/services/WebVhAnonCredsRegistry.ts:414:34)
          at async Object.<anonymous> (/home/ubuntu/repos/credo-ts/packages/webvh/src/anoncreds/services/__tests__/WebVhAnonCredsRegistry.test.ts:415:22) {
        code: 8
      }
```

packages/webvh/src/anoncreds/services/WebVhAnonCredsRegistry.ts
PatStLouis: thanks, I ended up importing the Sha256 hasher directly from the @credo-ts/core

packages/webvh/src/anoncreds/services/WebVhAnonCredsRegistry.ts
PatStLouis: I've imported the Sha256 hasher from the credo core

packages/webvh/src/anoncreds/services/WebVhAnonCredsRegistry.ts
PatStLouis: applied suggestion, thanks

packages/webvh/src/anoncreds/services/WebVhAnonCredsRegistry.ts
PatStLouis: applied suggestion, thanks

PatStLouis:
I ended up with a direct implementation of the "crypto" module for creating the ed25519 public key from the publicKeyMultibase value. I believe there is a bug in the askar library when creating a key from public key bytes which makes it not recognize the key algorithm passed.

packages/webvh/src/cryptosuites/eddsa-jcs-2022.ts
```diff
@@ -0,0 +1,141 @@
+import { Buffer } from 'buffer'
+import { createPublicKey, verify } from 'crypto'
```
TimoGlastra: You should not use crypto directly, we have a `kms` API that should handle all crypto operations (there's a lot of places in the codebase where you can see it being used)

packages/webvh/src/cryptosuites/eddsa-jcs-2022.ts
```diff
@@ -0,0 +1,141 @@
...
+    const canonicalProofConfig = canonicalize(proofConfig)
+    return canonicalProofConfig
+  }
+
+  public hashing(transformedDocument: string, canonicalProofConfig: string) {
+    // https://www.w3.org/TR/vc-di-eddsa/#hashing-eddsa-jcs-2022
+    const hasher = new Sha256()
+    const encoder = new TextEncoder()
+    const transformedDocumentHash = hasher.hash(encoder.encode(transformedDocument))
+    const proofConfigHash = hasher.hash(encoder.encode(canonicalProofConfig))
```
TimoGlastra: You should import `Hasher` from core and then pass the hasher algoirthm. And for encoding, you should use the `TypedArrayEncoder`. (TextEncoder is not available in all platforms)

packages/webvh/src/cryptosuites/eddsa-jcs-2022.ts
```diff
@@ -0,0 +1,141 @@
...
+    const publicKey = createPublicKey({
+      key: Buffer.concat([
+        Buffer.from([0x30, 0x2a, 0x30, 0x05, 0x06, 0x03, 0x2b, 0x65, 0x70, 0x03, 0x21, 0x00]),
+        Buffer.from(publicKeyBytes),
+      ]),
+      format: 'der',
+      type: 'spki',
+    })
+    return publicKey
+  }
```
TimoGlastra: You can use `getPublicJwkFromVerificationMethod` to get the publicJwk from the verification method, which then can do all these transformations. It's better to reuse the existing utils for this, as there's a lot of transformation logic already in place

packages/webvh/src/cryptosuites/eddsa-jcs-2022.ts
PatStLouis: I've implemented the KMS verify function, however it's not longer able to verify the proofs, could you let me know what is wrong with my implementation?

packages/webvh/src/cryptosuites/eddsa-jcs-2022.ts
PatStLouis: I've imported the Hasher function and TypedArrayEncoder

packages/webvh/src/cryptosuites/eddsa-jcs-2022.ts
PatStLouis: I've implemented getPublicJwkFromVerificationMethod

packages/webvh/src/cryptosuites/eddsa-jcs-2022.ts
PatStLouis: Verification now works as expected

packages/webvh/src/cryptosuites/eddsa-jcs-2022.ts
```diff
@@ -0,0 +1,106 @@
...
+    this.didApi = agentContext.dependencyManager.resolve(DidsApi)
+    this.keyApi = agentContext.dependencyManager.resolve(Kms.KeyManagementApi)
+  }
+
+  public async _logError(error: string) {
+    this.agentContext.config.logger.error(error)
+  }
+
+  public async _publicJwkFromId(verificationMethodId: string): Promise<PublicJwk> {
+    const didDocument = new DidDocument(await this.didApi.resolveDidDocument(verificationMethodId))
```
TimoGlastra: this method already returns an instance of DidDocument
```suggestion
    const didDocument = await this.didApi.resolveDidDocument(verificationMethodId)
```

packages/webvh/src/cryptosuites/eddsa-jcs-2022.ts
```diff
@@ -0,0 +1,106 @@
...
+  public transformation(unsecuredDocument: object, options: ProofOptions) {
+    // https://www.w3.org/TR/vc-di-eddsa/#transformation-eddsa-jcs-2022
+    if (options.type !== 'DataIntegrityProof') {
+      this._logError('PROOF_VERIFICATION_ERROR')
+      throw new Error('PROOF_VERIFICATION_ERROR')
+    }
+    if (options.cryptosuite !== 'eddsa-jcs-2022') {
+      this._logError('PROOF_VERIFICATION_ERROR')
+      throw new Error('PROOF_VERIFICATION_ERROR')
+    }
```
TimoGlastra: Should throw `CredoError` at least, and useful to make the error messsages more descriptive

packages/webvh/src/anoncreds/services/WebVhAnonCredsRegistry.ts
```diff
@@ -95,65 +94,55 @@ export class WebVhAnonCredsRegistry implements AnonCredsRegistry {
...
-        throw new CredoError(`Could not extract DID from resource ID: ${resourceId}`)
-      }
-      const expectedIssuerDid = resourceDidMatch[1]
-
-      if (contentIssuerId !== expectedIssuerDid) {
-        throw new CredoError(
-          `Issuer ID mismatch for ${resourceId}. Expected: ${expectedIssuerDid}, Actual: ${contentIssuerId}`
-        )
-      }
-      agentContext.config.logger.trace(`Resource ${resourceId} issuer ID matches DID.`)
```
TimoGlastra, line 155: Why is this check removed?

packages/webvh/src/cryptosuites/eddsa-jcs-2022.ts
PatStLouis: it seems to be returning an object

packages/webvh/src/cryptosuites/eddsa-jcs-2022.ts
PatStLouis: I can do, I was trying to be as verbatim as the specification

brianorwhatever: done

packages/webvh/src/cryptosuites/eddsa-jcs-2022.ts
brianorwhatever: everything is an object in JS land 😅 
I think you might have been having troubles in the test that was mocking the DID Doc returning a regular object. I've fixed this so the mock returns a new DidDocument

PatStLouis: it was having a hard time finding the `dereferenceVerificationMethod` on the response of the `resolveDidDocument` request. I might of misunderstood the error

TimoGlastra (approved):
Thanks for the quick iterations

packages/webvh/src/anoncreds/services/WebVhAnonCredsRegistry.ts
PatStLouis: I suppose we can keep it, I removed it because not all attested resource will have a `content.issuerId` property

TimoGlastra:
Can you run the formatter once more?

packages/webvh/src/anoncreds/services/WebVhAnonCredsRegistry.ts
TimoGlastra: We should probably keep it for the anoncreds resources then? 

Because this is the anoncreds registry all resources will have an issuer id right?

TimoGlastra:
Ah forogt about the comment on anoncreds registry. I think it's important to keep the issuerId check, but @brianorwhatever probably knows better

packages/webvh/src/anoncreds/services/WebVhAnonCredsRegistry.ts
PatStLouis: yeah we can probably enable it back, I think only the schema/cred def have the issuerId, not the revocation objects. we can keep it here or move it to the `getSchema` `getCredDef` calls respectively.

brianorwhatever:
I will move the issuer check to `getSchema` and `getCredDef`. Do I need to figure out what is happening with the failing tests? It seems they are intermittent and unrelated to this code.

genaris:
> I will move the issuer check to `getSchema` and `getCredDef`. Do I need to figure out what is happening with the failing tests? It seems they are intermittent and unrelated to this code.

Yeah, don't worry about them, as long as the ones for your module (and related ones) are working fine. Usually we need to re-run some suites before merging.

brianorwhatever:
Alright then I think this is ready for another review

packages/webvh/src/anoncreds/services/__tests__/WebVhTransform.test.ts
```diff
@@ -9,15 +9,15 @@ describe('WebVhTransform', () => {
     const resource = JsonTransformer.fromJSON(mockSchemaResource, WebVhResource)
 
     expect(resource).toBeInstanceOf(WebVhResource)
-    expect(resource['@context']).toEqual(['https://w3id.org/security/data-integrity/v2'])
+    // expect(resource['@context']).toEqual(['https://w3id.org/security/data-integrity/v2'])
```
genaris: I've noticed that the new mocked resources include `https://opsecid.github.io/attested-resource/v1` in their context. Is this correct? 

If that's the case, shall we uncomment and update this line or remove it?

packages/webvh/src/anoncreds/services/__tests__/WebVhTransform.test.ts
```diff
@@ -27,14 +27,14 @@ describe('WebVhTransform', () => {
     const resource = JsonTransformer.fromJSON(mockCredDefResource, WebVhResource)
 
     expect(resource).toBeInstanceOf(WebVhResource)
-    expect(resource['@context']).toEqual(['https://w3id.org/security/data-integrity/v2'])
+    // expect(resource['@context']).toEqual(['https://w3id.org/security/data-integrity/v2'])
```
genaris: Same as above

genaris (approved):
Just a doubt regarding mocked resources but LGTM!

packages/webvh/src/anoncreds/services/__tests__/WebVhTransform.test.ts
brianorwhatever: done

packages/webvh/src/anoncreds/services/__tests__/WebVhTransform.test.ts
brianorwhatever: done

brianorwhatever:
@genaris added the temp opsecid context URL. also updated the version to a new, react-native enabled build of `didwebvh-ts`

genaris:
> @genaris added the temp opsecid context URL. also updated the version to a new, react-native enabled build of `didwebvh-ts`

OK let's go! I think we'll need to do the backport to 0.5.x as well to facilitate integration testing.

brianorwhatever:
thanks @genaris! I am just looking at that now. I see opportunity for the differences to be encapsulated into the WebvhDidCrypto file and have the eddsa-jcs-2022 file. For now though I'll make changes to both files and we can look at doing that refactor in the next iteration.

PR #2357, opened 2025-08-13
fix(openid4vc): correctly transform date on IssuanceSessionRecord

hacdias:
Should fix https://github.com/openwallet-foundation/credo-ts/pull/2350#issuecomment-3180592711


hacdias:
Could we merge it?

PR #2358, opened 2025-08-13
feat(openid4vc): add refresh token support

hacdias:
Adds support for refresh tokens in the OpenID4VC server/client.

### DIFF
--- a/packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
+++ b/packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
@@ -16,6 +16,7 @@ import {
   parseDid,
 } from '@credo-ts/core'
 import {
+  CallbackContext,
   Jwk,
   Oauth2Client,
   RequestDpopOptions,
@@ -54,6 +55,7 @@ import type {
   OpenId4VciRetrieveAuthorizationCodeUsingPresentationOptions,
   OpenId4VciSendNotificationOptions,
   OpenId4VciSupportedCredentialFormats,
+  OpenId4VciTokenRefreshOptions,
   OpenId4VciTokenRequestOptions,
 } from './OpenId4VciHolderServiceOptions'
 
@@ -354,6 +356,43 @@ export class OpenId4VciHolderService {
           dpop,
           txCode: options.txCode,
         })
+
+    return {
+      ...result,
+      dpop: dpop
+        ? {
+            ...result.dpop,
+            alg: dpop.signer.alg as Kms.KnownJwaSignatureAlgorithm,
+            jwk: Kms.PublicJwk.fromUnknown(dpop.signer.publicJwk),
+          }
+        : undefined,
+    }
+  }
+
+  public async refreshAccessToken(agentContext: AgentContext, options: OpenId4VciTokenRefreshOptions) {
+    const oauth2Client = this.getOauth2Client(agentContext, {
+      clientAttestation: options.walletAttestationJwt,
+      clientId: options.clientId,
+    })
+
+    const dpop = options.dpop
+      ? await this.getDpopOptions(agentContext, {
+          ...options.dpop,
+          dpopSigningAlgValuesSupported: [options.dpop.alg],
+        })
+      : undefined
+
+    const authorizationServerMetadata = getAuthorizationServerMetadataFromList(
+      options.issuerMetadata.authorizationServers,
+      options.authorizationServer ?? options.issuerMetadata.authorizationServers[0].issuer
+    )
+
+    const result = await oauth2Client.retrieveRefreshTokenAccessToken({
+      authorizationServerMetadata,
+      refreshToken: options.refreshToken,
+      dpop,
+      resource: options.issuerMetadata.credentialIssuer.credential_issuer,
+    })
 
     return {
       ...result,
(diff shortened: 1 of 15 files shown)


packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
```diff
@@ -367,6 +368,55 @@ export class OpenId4VciHolderService {
...
+  public async refreshAccessToken(agentContext: AgentContext, options: OpenId4VciTokenRefreshOptions) {
+    const oauth2Client = this.getOauth2Client(agentContext)
+
+    const authorizationServerMetadata = getAuthorizationServerMetadataFromList(
+      options.issuerMetadata.authorizationServers,
+      options.issuerMetadata.authorizationServers.find(
+        (server) =>
+          // TODO: add type in @openid4vc/oauth2
+          Array.isArray(server.grant_types_supported) && server.grant_types_supported.includes('refresh_token')
+      )?.issuer ?? options.issuerMetadata.authorizationServers[0].issuer
```
hacdias: ℹ️ See https://github.com/openwallet-foundation-labs/oid4vc-ts/pull/89.

packages/openid4vc/src/openid4vc-issuer/router/accessTokenEndpoint.ts
```diff
@@ -215,13 +269,15 @@ export function handleTokenRequest(config: OpenId4VcIssuerModuleConfig) {
         clientId: issuanceSession.clientId,
 
         additionalAccessTokenPayload: {
-          'pre-authorized_code':
-            grant.grantType === preAuthorizedCodeGrantIdentifier ? grant.preAuthorizedCode : undefined,
+          'pre-authorized_code': issuanceSession.preAuthorizedCode,
           issuer_state: issuanceSession.authorization?.issuerState,
```
hacdias: ❓ Can this change cause issues? Do we have any flow where you might have both set? My issue here was to know what to set in the case of refresh tokens.

packages/openid4vc/src/shared/callbacks.ts
```diff
@@ -395,9 +395,8 @@ export function getOid4vcCallbacks(
     hash: (data, alg) => Hasher.hash(data, alg.toLowerCase()),
     generateRandom: (length) => kms.randomBytes({ length }),
     signJwt: getOid4vcJwtSignCallback(agentContext),
-    clientAuthentication: () => {
-      throw new CredoError('Did not expect client authentication to be called.')
-    },
+    // TODO?
+    clientAuthentication: clientAuthenticationAnonymous(),
```
hacdias: Is this fine?

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderServiceOptions.ts
```diff
@@ -178,6 +179,20 @@ export type OpenId4VciTokenRequestOptions =
   | OpenId4VciPreAuthorizedTokenRequestOptions
   | OpenId4VcAuthorizationCodeTokenRequestOptions
 
+export type OpenId4VciTokenRefreshOptions = {
+  refreshToken: string
+
+  /**
+   * The issuer metadata.
+   */
+  issuerMetadata: IssuerMetadataResult
```
hacdias, line 188: Do we maybe want to include the metadata in the refresh token itself? Then, we'd need to make it a JWT. An issue could be that the metadata changes.

My reasoning would be that the caller would first fetch the (potentially updated) metadata, and then refresh the token.

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
TimoGlastra: Multiple authorization servers could support it. We need to check based on the current access token which server to use.

Not sure if we can know it from the token, in which case we will have to store it alongside the refresh token. As this can lead to weird errors

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
```diff
@@ -367,6 +368,55 @@ export class OpenId4VciHolderService {
...
+          dpopSigningAlgValuesSupported: [options.dpop.alg],
+        })
+      : // We should be careful about this case. It could just be the user didn't correctly
+        // provide the DPoP from the auth response. In which case different DPoP will be used
+        // However it might be that they only use DPoP for the token request (esp in pre-auth case)
+        isDpopSupported.supported
+        ? await this.getDpopOptions(agentContext, {
+            dpopSigningAlgValuesSupported: isDpopSupported.dpopSigningAlgValuesSupported,
+          })
+        : undefined
```
TimoGlastra: When we have an access token already, we already have the dpop key, so i don't
Think we have to generate a new one (servers should verify that you use the same dpop key to request the new access token as the initial one, so you're sure it's still the same client)

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderServiceOptions.ts
TimoGlastra: Would this work with other implementations still if we add it to the metadata? 

It's a good question what should happen with refresh, but i think we should just always use the latest metadata (at least for now) so updating metadata means that the refresh token will also use the updated metadata

packages/openid4vc/src/openid4vc-issuer/repository/OpenId4VcIssuanceSessionRecord.ts
```diff
@@ -40,6 +40,11 @@ export interface OpenId4VcIssuanceSessionAuthorization {
    */
   codeExpiresAt?: Date
 
+  /**
+   * Refresh token that can be used to refresh the access token.hat can be used to refresh the access token.
```
```suggestion
   * Refresh token that can be used to refresh the access token.
```

TimoGlastra: I don't think we should store the refresh token, but rather make it a signed jwt with the needed metadata. That way it's more flexible and we can also reuse it for refresh. 

We also don't store the access token.

packages/openid4vc/src/openid4vc-issuer/router/accessTokenEndpoint.ts
```diff
@@ -176,6 +196,40 @@ export function handleTokenRequest(config: OpenId4VcIssuerModuleConfig) {
               }
             : undefined,
         })
+      } else if (grant.grantType === refreshTokenGrantIdentifier) {
```
TimoGlastra, line 203: We might want to add an option whether to issue a refresh token, and how long the refresh token is valid.

Also do we issue a new refresh token every time you use a refresh token? In that case we also need to configure it, but maybe just issuing one refresh token that is valid for x time is sufficient?

packages/openid4vc/src/openid4vc-issuer/router/accessTokenEndpoint.ts
TimoGlastra: An offer can contain both

It is a change in behaviour, since we only add the used method to the token.

We can base it on the current access token? That is used for refresh

packages/openid4vc/src/shared/callbacks.ts
TimoGlastra: Generally we override it in specific places so we know in what places we need ti handle auth. 

Where is it called? I assume we want to pass the wallet attestation there as well (similar to how access token works)

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
hacdias: > We need to check based on the current access token which server to use.

Do you mean whether the pre-authorization server or the other one should be used?

If we need to have the possibility of both methods, we'll need to actually have multiple refresh tokens as a JWT, with enough information to connect it back to a session **and** the correct authentication method.

packages/openid4vc/src/openid4vc-issuer/router/accessTokenEndpoint.ts
hacdias: > Also do we issue a new refresh token every time you use a refresh token? In that case we also need to configure it, but maybe just issuing one refresh token that is valid for x time is sufficient?

No, we only do 1 at the time, since it gets stored in the session itself. And yesterday we had discussed that SPRIN-D only uses one, so I assumed we wanted to do something similar.

I think issuing new refresh tokens with a validity (configurable) is easier. Then we should make it a signed JWT, and it should contain the necessary metadata in order to be able to recreate a new access token.

packages/openid4vc/src/openid4vc-issuer/router/accessTokenEndpoint.ts
hacdias: I will update this once i make them a JWT

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
TimoGlastra: This is the holder side, so we can't assume the issuer is also credo. There could be 100 authorization servers in the list, and tbe issuer can indicate in the offer which one should be used, so on the holder side we need to store which authorization servers the access token is associated with

hacdias:
Hmm, I see a lot of failing tests. I will take a look.

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerService.ts
```diff
@@ -1067,6 +1067,95 @@ export class OpenId4VcIssuerService {
...
+    const key = issuer.resolvedAccessTokenPublicJwk
+    const refreshToken = await jwsService.createJwsCompact(agentContext, {
+      keyId: key.keyId,
+      payload: JwtPayload.fromJson({
+        iss: issuerMetadata.credentialIssuer.credential_issuer,
+        aud: issuerMetadata.credentialIssuer.credential_issuer,
+        exp: dateToSeconds(expiresAt),
+        issuer_state: options.issuerState,
+        'pre-authorized_code': options.preAuthorizedCode,
+      }),
```
TimoGlastra: We should probably bind the refresh token to the DPoP key as well, so that we can check when the refresh token is used that it's still being used with the DPoP key.

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerModuleConfig.ts
```diff
@@ -226,12 +242,31 @@ export class OpenId4VcIssuerModuleConfig {
...
+    return this.options.refreshTokenExpiresInSeconds ?? DEFAULT_REFRESH_TOKEN_EXPIRES_IN
+  }
+
+  /**
+   * Whether to generate refresh tokens for issuance sessions. This can be useful if
+   * you're planning on deferring the issuance of credentials.
+   *
+   * @default false
+   */
+  public get generateRefreshTokens(): boolean {
```
TimoGlastra: I think it's useful to be able to set this on a per-issuance session basis as well, is that possible? At least in Paradym I see us making deferred issuance a thing you configure on a credential template, and thus it's not a global setting

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerService.ts
```diff
@@ -1067,6 +1067,95 @@ export class OpenId4VcIssuerService {
...
+        typ: 'credo+refresh_token',
+        kid: key.keyId,
+        alg: key.signatureAlgorithm,
+      },
+    })
+
+    return refreshToken
+  }
+
+  public async verifyRefreshToken(agentContext: AgentContext, issuer: OpenId4VcIssuerRecord, token: string) {
```
TimoGlastra: If we do bind the refresh token to the DPoP, we should verify it here as well.

Reference for german architecture: https://gitlab.opencode.de/bmi/eudi-wallet/eidas-2.0-architekturkonzept/-/blob/main/architecture-proposal/flows/PID-IssuerSigned-cloud.md#issuance-of-the-refresh-token

We can also defer this to a later PR, but if we don't bind the refresh token to the DPoP as well, the refresh token loses a lot of security that the Access token does have.

hacdias: I will add that in this PR

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerModuleConfig.ts
hacdias: It is possible

TimoGlastra (approved):

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerModuleConfig.ts
hacdias: I made this an option when creating the credential offer.

The only question that remains is regarding dynamic issuance: should refresh tokens just be disabled for those? Or should we keep a global config (like now), but specific to dynamic issuance? I'd keep it simple for now and keep it disabled, and only add if needed.

hacdias:
@TimoGlastra I added the DPoP to the the refresh token, as well as the validation.

Re refresh token configuration: https://github.com/openwallet-foundation/credo-ts/pull/2358#discussion_r2276586468

TimoGlastra (approved):
Good stuff!

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerModuleConfig.ts
TimoGlastra: I agree

PR #2361, opened 2025-08-13
refactor!: prefix DIDComm module and main classes

genaris:
A very tough one!

Following up #2160, the goal here is to rename DIDComm-related modules and man classes/interfaces to help identifying them and reducing conflicts and ambiguities with other modules and/or protocols. 

So here's a summary of the work done so far:

- All `didcomm` package modules, APIs, models and services that are used for dependency injection now include `DidComm` in its naming (most of them are prefixed, while others just include it somewhere in the name for a more appropriate naming, e.g. OutboundDidCommTransport`)
- DIDComm-related events have also changed their text string to make it possible to distinguish them from events triggered by other protocols
- DIDComm credentials module API has been updated to use the term `credentialExchangeRecord` instead of `credentialRecord`, since it is usually confused with W3cCredentialRecords (and potentially other kind of credential records we might have). I took this opportunity to also update `declineOffer` options structure to match DIDComm proofs module API
- DIDComm-related records were renamed, but their type is still the original one (e.g. `CredentialRecord`, `BasicMessageRecord`). Probably it's worth to take this major release to do the migration, but I'm afraid that it will be a bit risky, so I'm hesitating to do so or leaving it for some other major upgrade (if we think it's actually needed)
- I didn't rename any other DIDComm-related packages, such as action-menu, dprc and question-answer. Should I?

Surely there are several things I forgot to refactor, and others that do not have a very consistent naming. Open to get some help from other eyes (mines are completely wasted now, but I'll review this PR as soon as they recover :smile: )

PR #2362, opened 2025-08-14
feat!: drop label and connectionImageUrl from agent config

genaris:
Another task from https://github.com/openwallet-foundation/credo-ts/issues/2160. 

For `connectionImageUrl` there is no major issue but for label there are some concerns that could result on a problem:
- Some connection invitation related API methods now require label to be specified (if users do not want to set anything, they can just set them to "")
- In processes that are handling DIDComm connection protocols automatically, we won't be able to let users to specify their particular label. We'll always have the possibility to disable automatic connection/invitation acceptance, but it will not be the case for the auto-provisioning mechanism (i.e. when `mediationInvitationUrl` is defined). In all those cases, a default empty label will be used
- Label in Tenants is kept left for the moment, since probably it can be useful to identify tenants in a friendlier manner than their ID. Also I'm not sure how to deal with existing tenant records that include this field

### DIFF
--- a/demo-openid/src/BaseAgent.ts
+++ b/demo-openid/src/BaseAgent.ts
@@ -35,7 +35,6 @@ export class BaseAgent<AgentModules extends ModulesMap> {
     this.app = express()
 
     const config = {
-      label: name,
       allowInsecureHttpUrls: true,
       logger: new ConsoleLogger(LogLevel.off),
     } satisfies InitConfig
--- a/demo/src/Alice.ts
+++ b/demo/src/Alice.ts
@@ -26,7 +26,9 @@ export class Alice extends BaseAgent {
   }
 
   private async receiveConnectionRequest(invitationUrl: string) {
-    const { connectionRecord } = await this.agent.modules.oob.receiveInvitationFromUrl(invitationUrl)
+    const { connectionRecord } = await this.agent.modules.oob.receiveInvitationFromUrl(invitationUrl, {
+      label: 'alice',
+    })
     if (!connectionRecord) {
       throw new Error(redText(Output.NoConnectionRecordFromOutOfBand))
     }
--- a/demo/src/BaseAgent.ts
+++ b/demo/src/BaseAgent.ts
@@ -1,4 +1,3 @@
-import type { InitConfig } from '@credo-ts/core'
 import type { DidCommModuleConfigOptions } from '@credo-ts/didcomm'
 import type { IndyVdrPoolConfig } from '@credo-ts/indy-vdr'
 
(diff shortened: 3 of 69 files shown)


TimoGlastra:
> In processes that are handling DIDComm connection protocols automatically, we won't be able to let users to specify their particular label. We'll always have the possibility to disable automatic connection/invitation acceptance, but it will not be the case for the auto-provisioning mechanism (i.e. when mediationInvitationUrl is defined). In all those cases, a default empty label will be used

Do you have some examples where this applied(there's a lot of changed files 😬 )

genaris:
> > In processes that are handling DIDComm connection protocols automatically, we won't be able to let users to specify their particular label. We'll always have the possibility to disable automatic connection/invitation acceptance, but it will not be the case for the auto-provisioning mechanism (i.e. when mediationInvitationUrl is defined). In all those cases, a default empty label will be used
> 
> Do you have some examples where this applied(there's a lot of changed files 😬 )

Oh this is nothing compared to #2361 ! :rofl: 

Actually, the main issue is at `packages/didcomm/src/modules/routing/MediationRecipientModule.ts` when mediationInvitationUrl is defined, in `onInitializeContext()`.

All the other cases are covered, provided the user is always forced to specify a non-undefined `label` for both invitation reception and creation.

.changeset/seven-dryers-carry.md
```diff
@@ -0,0 +1,15 @@
+---
+"@credo-ts/indy-sdk-to-askar-migration": patch
+"@credo-ts/anoncreds": patch
+"@credo-ts/openid4vc": patch
+"@credo-ts/didcomm": patch
+"@credo-ts/tenants": patch
+"@credo-ts/askar": patch
+"@credo-ts/core": patch
+---
```
TimoGlastra: should be minor

.changeset/seven-dryers-carry.md
```diff
@@ -0,0 +1,15 @@
...
+"@credo-ts/indy-sdk-to-askar-migration": patch
+"@credo-ts/anoncreds": patch
+"@credo-ts/openid4vc": patch
+"@credo-ts/didcomm": patch
+"@credo-ts/tenants": patch
+"@credo-ts/askar": patch
+"@credo-ts/core": patch
```
TimoGlastra:
```suggestion
"@credo-ts/indy-sdk-to-askar-migration": minor
"@credo-ts/anoncreds": minor
"@credo-ts/openid4vc": minor
"@credo-ts/didcomm": minor
"@credo-ts/tenants": minor
"@credo-ts/askar": minor
"@credo-ts/core": minor
```

PR #2363, opened 2025-08-14
fix: webvh di proof verification v0.5.x

brianorwhatever:
Follow up to #2354 for 0.5.x


genaris (approved):

swcurran:
Thanks @genaris -- would love to see 0.5.17 released with this -- https://github.com/openwallet-foundation/credo-ts/pull/2351.  Nudge nudge -- in the nicest way possible :-).

TimoGlastra:
Did you already test the latest alpha release in Bifold? Usually we try to test it in at least one  implementation before making a release (and our wallet is already on 0.6.x alpha track).

Basically there's no difference between the latest alpha on NPM and the next 0.5.17 release if we trigger it now.

https://www.npmjs.com/package/@credo-ts/webvh/v/0.5.17-alpha-20250814185552

swcurran:
@TimoGlastra -- thanks for the guidance.  I wasn't sure how final releases get triggered.

@brianorwhatever -- can you please do the Bifold testing with the alpha and let us know when done?

PR #2368, opened 2025-08-15
feat: store kms key id on credential record and batch records

TimoGlastra:
Wnat to open this to get some feedback.

It contains two things:
- It allows multiple credetnails to be stored in a single record. This aligns with how OpenID4VCI batch issuance works (you get multiple instances of the same credentail, to use a new one for each presentation).
- It stores the kms key id on the record

The reason to store the kms key id, is that (as we've noticed with our own issuer) we maybe can't rely on the issuer including `kid` we send in the JWK in the credential itself. This causes issues, since we need the kid to sign the presentation.

The reason to store multiple instances in one record, is that all credentials in one batch should be seen as canonical, so storing them as separate records just clutters the credential storage (we have been using this approach using metadata for a while already).

If we think both are a good approach, i can work this out further, including:
- using a new instance every time we present a credential if new instances are still availalbe (configurable)
- extracting the kms key id after we received the credential (@hacdias we need to think how this will work with deferred issuance, since we send the proof during the request, but only create the record later, so we need to store the kms key id somewhere between request and actual isuance in case of deferred).

Curious to also hear your thought @genaris.

### DIFF
--- a/packages/drizzle-storage/src/core/w3c-v2-credential-record/postgres.ts
+++ b/packages/drizzle-storage/src/core/w3c-v2-credential-record/postgres.ts
@@ -13,7 +13,6 @@ export const w3cV2Credential = pgTable(
   {
     ...getPostgresBaseRecordTable(),
 
-
     credentialInstances: jsonb('credential_instances').$type<W3cV2CredentialRecordInstances>().notNull(),
 
     // Default Tags


hacdias:
Storing the KMS Key ID, and multiple instances in one record sound both sensible ideas to me.

> * using a new instance every time we present a credential if new instances are still availalbe (configurable)

This also seems the most logical option to me.

> * extracting the kms key id after we received the credential (@hacdias we need to think how this will work with deferred issuance, since we send the proof during the request, but only create the record later, so we need to store the kms key id somewhere between request and actual issuance in case of deferred).

We could either create some repository in the holder for pending issuance sessions, or make it an additional argument/option the user needs to pass along. I think both are sensible options, but the second is probably easier to manage. Or rather: we don't really need to manage it, but the user. We could require that, in the case of deferred issuance, the user needs to provide the proofs to the `handleCredentialResponse` function.

TimoGlastra:
> Or rather: we don't really need to manage it, but the user. We could require that, in the case of deferred issuance, the user needs to provide the proofs to the handleCredentialResponse function.

I think for starters this is fine. Then we can add it in our wallet, experiment with it, and if we have a model that works we can always add it to Credo to make the deferred issuance integration easier 👍

packages/askar/src/storage/AskarStorageService.ts
```diff
@@ -65,6 +69,46 @@ export class AskarStorageService<T extends BaseRecord> implements StorageService
     }
   }
 
+  /**
+   * Update the record by id with a lock based on the value returned in `updateCallback`.
+   *
+   * NOTE that this has no effect when SQLite is used, as SQLite does not support row level
+   * locking
+   */
+  public async updateByIdWithLock(
```
hacdias, line 78: I like this.

packages/drizzle-storage/src/core/w3c-v2-credential-record/postgres.ts
```diff
@@ -12,7 +13,9 @@ export const w3cV2Credential = pgTable(
   {
     ...getPostgresBaseRecordTable(),
 
-    credential: jsonb().$type<W3cVerifiableCredential['encoded']>().notNull(),
+    // credential: jsonb().$type<W3cVerifiableCredential['encoded']>().notNull(),
```
hacdias: Should we remove this comment?


```suggestion
```

packages/core/src/storage/StorageService.ts
```diff
@@ -76,6 +76,34 @@ export interface StorageService<T extends BaseRecord<any, any, any>> {
...
+  /**
+   * Retrieve the record with by id, and provide it in the callback for update.
+   * The returned record will be stored.
+   *
+   * The purpose of this method is to allow storage services that support locking
+   * to lock the record, preventing concurrent processes from overwriting updates
+   * to the record.
+   *
+   * Note that locking a record can result in deadlocks, and slow down processes.
+   * It's recommended to minimize the side effects performed in the `updateCallback`
```
hacdias, line 88: Maybe a future idea would be to set some timeout in the callback. Not sure which mechanisms exist that would make that easy to do, but something like: if updateCallback doesn't resolve in 30 seconds, the lock is released and an error is thrown.

packages/core/src/storage/StorageService.ts
```diff
@@ -76,6 +76,34 @@ export interface StorageService<T extends BaseRecord<any, any, any>> {
...
+   *
+   * TODO: should we allow partial updates for backend that support it? E.g. with drizzle
+   * we can update just a value which makes locking less needed in some cases.
```
hacdias, line 91: What do you mean by partial values? You mean specific fields?

packages/openid4vc/tests/openid4vc-wallet-key-attestation.e2e.test.ts
```diff
@@ -30,13 +30,22 @@ const universityDegreeCredentialConfigurationSupportedMdoc = {
         user_authentication: ['iso_18045_high'],
       },
     },
+    attestation: {
+      proof_signing_alg_values_supported: ['ES256', 'EdDSA'],
+      key_attestations_required: {
+        // TODO: enum for iso enum values
+        key_storage: ['iso_18045_high'],
+        user_authentication: ['iso_18045_high'],
```
hacdias, line 38: Are we bringing this later?

packages/openid4vc/tests/openid4vp-v1.e2e.test.ts
```diff
@@ -454,18 +478,27 @@ pUGCFdfNLQIgHGSa5u5ZqUtCrnMiaEageO71rjzBlov0YUH4+6ELioY=
...
-                },
+                ],
               },
             },
           ],
         },
       },
     })
 
+    // biome-ignore lint/suspicious/noExplicitAny: no explanation
```
hacdias, line 493: ✨ _suspicious_ ✨

packages/core/src/storage/StorageService.ts
TimoGlastra: Yeah like a normal `UPDATE X SET Y = Z`. However in this case I'd still want a lock, to prevent credentials being read twice. The partial update is mostly relevant if you do e.g .`count = count +1` (which is not susceptible to race conditions if you do it in the query)

packages/openid4vc/tests/openid4vc-wallet-key-attestation.e2e.test.ts
TimoGlastra: It's an old todo, so yeah will probably do it later

PR #2376, opened 2025-08-21
fix(openid4vc): update state after deferring credential issuance

hacdias:
This fixes an issue where multiple refresh tokens could not be requested, since we did not correctly update the state after deferring the credential issuance.


TimoGlastra:
Do we update the state to partially issued / completed even if the credential is deferred?

hacdias:
> Do we update the state to partially issued / completed even if the credential is deferred?

It only updates if they've actually been issued. So it will correctly reflect whether the request is complete (no more deferred credentials or credentials to be issues) or partially issued (some credentials issued, some not issued/deferred).

---

As a side note: I don't think the current state system works very well with the deferred credentials. For example: you can have a partially issued session. Then you go get a new access token with the refresh token, and suddenly the state is "access token requested" instead of "partially issued" / "credential request received". That's why I made this update, to ensure that the state updates after each request to the (deferred) credential endpoints.

However, I don't think revamping the whole state system is a priority right now, since it should work well like this. It would require way too many combinations of the current states to work well. What we instead could do is not update the session status to `AccessTokenCreated`/`AccessTokenRequested` **if** we are using a refresh token.

There's also no state to indicate "it is deferred". Because then you'd need all combinations of deferred, deferred+partially issued, and what is prioritized? Maybe renaming "partially issued" to "ongoing" or "not yet finished" would make it more clear.

TimoGlastra:
> What we instead could do is not update the session status to AccessTokenCreated/AccessTokenRequested if we are using a refresh token.

Yes i think that makes sense, then the state only moves forward, not backwards

TimoGlastra:
> Maybe renaming "partially issued" to "ongoing" or "not yet finished" would make it more clear.

That also sounds logical to me. We can track whether it's partially issued / deferred based on the other fields on the record

TimoGlastra (approved):
Approving, but would be good to at least make the change to not return to access token state after using a refresh token (either in this or follow up pr)

TimoGlastra:
The other state change is larger and thus better suited for a separate PR (and not really needed atm imo)

hacdias:
@TimoGlastra please see changes:

- Retains state when refreshing token
- Correctly asserts the possible states on the deferred credential endpoint

PR #2381, opened 2025-08-21
fix: Set pickup delivery to prefer websocket transport

jamshale:
This is to fix a couple problems we've been having with live mode and both http and ws sessions. I don't know much about the outbound context session id but I changed the preference to choose the session by connection id first, and then use the websocket session if it is in the session table.

Then for the removing of the pickup session when in live mode delivery, I changed this to only happen when websocket sessions close. All other delivery modes close pickup sessions on every session closed event.

TODO: Automated testing

### DIFF
--- a/packages/core/src/modules/message-pickup/services/MessagePickupSessionService.ts
+++ b/packages/core/src/modules/message-pickup/services/MessagePickupSessionService.ts
@@ -35,7 +35,7 @@ export class MessagePickupSessionService {
     eventEmitter
       .observable<TransportSessionRemovedEvent>(TransportEventTypes.TransportSessionRemoved)
       .pipe(
-        filter((e) => e.payload.session.type === 'Websocket'),
+        filter((e) => e.payload.session.type === 'WebSocket'),
         takeUntil(stop$)
       )
       .subscribe({
--- a/packages/core/src/modules/message-pickup/services/MessagePickupSessionService.ts
+++ b/packages/core/src/modules/message-pickup/services/MessagePickupSessionService.ts
@@ -3,7 +3,7 @@ import type { TransportSessionRemovedEvent } from '../../../transport'
 import type { MessagePickupLiveSessionRemovedEvent, MessagePickupLiveSessionSavedEvent } from '../MessagePickupEvents'
 import type { MessagePickupSession, MessagePickupSessionRole } from '../MessagePickupSession'
 
-import { takeUntil, type Subject } from 'rxjs'
+import { filter, takeUntil, type Subject } from 'rxjs'
 
 import { EventEmitter } from '../../../agent/EventEmitter'
 import { InjectionSymbols } from '../../../constants'
@@ -34,7 +34,10 @@ export class MessagePickupSessionService {
 
     eventEmitter
       .observable<TransportSessionRemovedEvent>(TransportEventTypes.TransportSessionRemoved)
-      .pipe(takeUntil(stop$))
+      .pipe(
+        filter((e) => e.payload.session.type === 'Websocket'),
+        takeUntil(stop$)
+      )
       .subscribe({
         next: (e) => {
           const connectionId = e.payload.session.connectionId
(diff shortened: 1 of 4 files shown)


packages/core/src/modules/message-pickup/services/MessagePickupSessionService.ts
```diff
@@ -30,15 +32,27 @@ export class MessagePickupSessionService {
   public start(agentContext: AgentContext) {
     const stop$ = agentContext.dependencyManager.resolve<Subject<boolean>>(InjectionSymbols.Stop$)
     const eventEmitter = agentContext.dependencyManager.resolve(EventEmitter)
+    const messageForwardingStrategy =
+      agentContext.dependencyManager.resolve(MediatorModuleConfig).messageForwardingStrategy
     this.sessions = []
 
     eventEmitter
       .observable<TransportSessionRemovedEvent>(TransportEventTypes.TransportSessionRemoved)
       .pipe(takeUntil(stop$))
```
genaris:
```suggestion
      .pipe(
        filter((e) => e.payload.session.type === 'WebSocket'),
        takeUntil(stop$)) 
```

I think this single modification should be enough, since we should only look for WebSocket transport sessions (as they are the only valid sessions for Live mode). And this statement is valid, regardless the MessageForwardingStrategy we are following.

genaris (changes requested):
Thanks @jamshale , I think now I understand better the problems you are experiencing and hopefully it is quite easy to fix them.

I think the changes you've introduced to `MessageSender` and `TransportService` are not necessary, since MessageSender's sendMessage method already includes an optional parameter where you can set it to give priority (and also restrict) to a particular transport. In `MessagePickupApi` we are using it accordingly in some cases, but by mistake we forgot to do it in `deliverMessages` and `deliverMessagesFromQueue`: this might be the reason why your agent is attempting to send messages through the existing HTTP session instead of the WS one. So the fix in that case would be to replace:

```ts
      await this.messageSender.sendMessage(
        new OutboundMessageContext(deliverMessagesReturn.message, {
          agentContext: this.agentContext,
          connection: connectionRecord,
        })
```

By:

```ts
      await this.messageSender.sendMessage(
        new OutboundMessageContext(deliverMessagesReturn.message, {
          agentContext: this.agentContext,
          connection: connectionRecord,
        }),
        { transportPriority: { schemes: ['ws', 'wss'], restrictive: true }}
      )
```

In both methods.

Regarding the changes you've introduced in `MessagePickupSessionService`, I think we can make them a lot simpler by just adding a filter by transport type in the event listener (see my comment).

I hope these changes will work for you and look forward to your feedback and updates to this PR to merge these fixes in!

jamshale:
Forced pushed the changes you suggested. Tested QueueAndLiveDelivery and QueueOnly.

I still want to add a couple automated tests.

packages/core/src/modules/message-pickup/services/MessagePickupSessionService.ts
```diff
@@ -34,7 +34,10 @@ export class MessagePickupSessionService {
 
     eventEmitter
       .observable<TransportSessionRemovedEvent>(TransportEventTypes.TransportSessionRemoved)
-      .pipe(takeUntil(stop$))
+      .pipe(
+        filter((e) => e.payload.session.type === 'Websocket'),
```
jamshale: This was actually the key to making it work. The other changes do fix the message delivery not using the websocket but the agent is still sending the status message via http and it's not breaking the pickupRepository.

I noticed in QueueOnly mode it doesn't actually create pickup sessions, so it continues to work correctly, the problem was it wasn't filtering them here so in live mode the http session were removing the pickup session for the websocket. This was what my patch was doing, but in a way more complicated way.

I think the status message should also go through `ws` transport but I haven't figured out where that was happening and it's no longer causing the problem.

genaris (approved):
Thanks!

PR #2384, opened 2025-08-26
chore: update the `replaceError` to better handle react native serialization

TimoGlastra:
We keep seeing issues in React Native with error serialization, making it hard to debug without adding a lot of console.logs in node_modules. I have verified this in our wallet, and it improves the error serialization, for errors that don't implement `toString`.


genaris (approved):
Nice!

PR #2385, opened 2025-08-26
feat: pickupV2 backwards compatibility with implicit pickup

jamshale:
This is my initial attempt at getting backwards compatibility of a v2 live mode enabled mediator when an agent is using implicit pickup.

1. When handling a trust ping. Check if the agent is configured for QueueAndLiveMode.
2. Check if there is already a live session or add a new live session.
3. Get messages from the repository queue and send all messages (up to 10) implicitly to the recipient.

Note: I need to wait until the real transport session is created. I wanted to keep this operation contained in the trust ping handler so I created a delay before add the pickup live session.


jamshale:
@genaris Can you have a look at this?

TimoGlastra:
Hey @jamshale, with the middleware API we provide, it would be possible to implement this on top of Credo. In this case I think i'd lean to taking that approach, since this seems very specific to your setup.

jamshale:
Sounds good. I can also create custom code in the mediator to listen to the TrustPingEvent. This is how I originally did it, but it would be nice to not have a custom mediator if possible. I'll wait for Ariel's review and then decide what to do.

It is specific to our use case, but it would be useful for any deployments that launched mobile agent with implicit pickup that wants to move to the pickupv2 protocols.

jamshale:
Thanks for the feedback. My credo knowledge is so limited that this was basically just an attempt at getting our mobile agents working with pickupv2 and the pickup repository when configured for Implicit. 

I'll think about the correct thing to do. I'll close this for now as I agree it's not the best approach.

genaris:
> Thanks for the feedback. My credo knowledge is so limited that this was basically just an attempt at getting our mobile agents working with pickupv2 and the pickup repository when configured for Implicit.
> 
> I'll think about the correct thing to do. I'll close this for now as I agree it's not the best approach.

No worries! I know Credo architecture is quite complex and really appreciate you are contributing with us. I think this PR brought some good ideas about how we can better implement this. Now I have the ideas fresh I'll give a try to this "ImplicitMessagePickupProtocol" and see how it would look like on Credo core. Your feedback and testing will be really key for this.

PR #2387, opened 2025-08-28
feat: W3C VCDM 2.0 credential and compliance with DIIP 4.0

hacdias:
This is a very big work in progress, and I'm opening this as a draft to get some early feedback/eyes.

The goals here are:

- Implement the W3C VCDM 2.0 credential format: https://www.w3.org/TR/vc-data-model-2.0/
- Implement the securing of such format with JWT
- Implement the securing of such format with SD-JWTs: https://www.w3.org/TR/vc-jose-cose/#with-sd-jwt
- Integrate the SD-JWT securing with OID4VP/VCI
- Ensure a compliant implementation with DIIP v4: https://fidescommunity.github.io/DIIP/ (see also: https://github.com/FIDEScommunity/DIIP/issues/67)


packages/core/src/modules/vc/models/presentation/W3cV2Presentation.ts
```diff
@@ -0,0 +1,67 @@
...
+import { W3cV2JsonPresentation } from './W3cV2JsonPresentation'
+
+export interface W3cV2PresentationOptions {
+  id?: string
+  context?: Array<string | JsonObject>
+  type?: Array<string>
+
+  // TODO: should we support other types of verifiable credentials here?
+  // Like plain ones without enveloping?
+  verifiableCredential: SingleOrArray<W3cV2EnvelopedVerifiableCredential>
```
TimoGlastra: It should support other v2 creds, but since we don't support those yet just enveloped is fine for now

packages/core/src/modules/vc/sd-jwt-vc/W3cV2SdJwtVerifiableCredential.ts
```diff
@@ -0,0 +1,51 @@
...
+import { W3cV2Credential } from '../models/credential/W3cV2Credential'
+import { getCredentialFromSdJwtPayload, sdJwtVcHasher } from './credentialTransformer'
+
+export interface W3cV2SdJwtVerifiableCredentialOptions {
+  compact: string
+  sdJwt: DecodedSDJwt
+}
+
+export class W3cV2SdJwtVerifiableCredential {
+  public readonly sdJwt: DecodedSDJwt
```
TimoGlastra: With the sd jwt vc we hide most of the internal sd-jwt structures, maybe we want to do that here as well, so breaking changes in sd-jwt do not affect credo directly

packages/core/src/modules/vc/sd-jwt-vc/W3cV2SdJwtVerifiableCredential.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    const compact = uri.slice('data:application/vc+sd-jwt,'.length)
+    return W3cV2SdJwtVerifiableCredential.fromCompact(compact)
+  }
+
+  /**
+   * The {@link ClaimFormat} of the credential. For SD-JWT credentials this is always `vc+sd-jwt`.
+   */
+  public get claimFormat(): ClaimFormat.SdJwtVc {
+    return ClaimFormat.SdJwtVc
```
TimoGlastra: We need a new claim format, since within credo we shouldn't have clashes. 

We might want to switch the current sd jwt vc to dc+sd-jwt and call this one SdJwtW3cVc/W3cVcSdJwt with value vc+sd-jwt

packages/core/src/modules/vc/sd-jwt-vc/credentialTransformer.ts
```diff
@@ -0,0 +1,29 @@
...
+import { W3cV2Credential, W3cV2CredentialOptions } from '../models/credential/W3cV2Credential'
+
+export function sdJwtVcHasher(data: string | ArrayBufferLike, alg: string) {
+  return Hasher.hash(typeof data === 'string' ? data : new Uint8Array(data), alg)
+}
+
+export function getCredentialFromSdJwtPayload(sdJwt: DecodedSDJwt): W3cV2Credential {
+  const { header, payload } = sdJwt.jwt
+
+  if ('typ' in header && header.typ !== 'vc+sd-jwt') {
```
TimoGlastra: Is typ optional?

packages/core/src/modules/vc/sd-jwt-vc/credentialTransformer.ts
```diff
@@ -0,0 +1,29 @@
...
+    throw new CredoError(`The provided W3C VC SD-JWT does not have the correct 'typ' header.`)
+  }
+
+  if ('cyt' in header && header.cyt !== 'vc') {
```
TimoGlastra: Same here?

packages/core/src/modules/vc/sd-jwt-vc/credentialTransformer.ts
hacdias: > The typ header parameter SHOULD be vc+sd-jwt.

https://www.w3.org/TR/vc-jose-cose/#securing-with-sd-jwt

packages/core/src/modules/vc/sd-jwt-vc/credentialTransformer.ts
hacdias: > When present, the cty header parameter SHOULD be vc.

https://www.w3.org/TR/vc-jose-cose/#securing-with-sd-jwt

packages/core/src/modules/vc/sd-jwt-vc/W3cV2SdJwtVerifiableCredential.ts
hacdias: I've updated the following:

- Renamed `ClaimFormat.SdJwtVc` to `ClaimFormat.SdJwtDc` (and changed the value to `dc+sd-jwt`)
- Added a new `ClaimFormat.SdJwtW3cVc` with the value `vc+sd-jwt`

packages/core/src/modules/vc/models/presentation/W3cV2Presentation.ts
hacdias: I refactored this to make clear that support for other credential types is intended.

hacdias:
There's a lot of tests failing with the following error, and I'm not yet sure what triggered it, since I don't. see any relevant changes around those files:

```
Cannot inject the dependency "w3cCredentialService" at position #1 of "W3cCredentialsApi" constructor. Reason:
        Cannot inject the dependency "w3cCredentialRepository" at position #0 of "W3cCredentialService" constructor. Reason:
            TypeInfo not known for "Object"
```

.changeset/shy-clubs-jam.md
```diff
@@ -0,0 +1,7 @@
+---
+"@credo-ts/openid4vc": patch
+"@credo-ts/core": patch
```
TimoGlastra: Can you also add a minor change, indicating the format identifier for sd jwt vc has changed?

packages/core/src/modules/dcql/DcqlService.ts
```diff
@@ -120,6 +123,26 @@ export class DcqlService {
...
+        credentialQuery.format === 'vc+sd-jwt' && !!credentialQuery.meta && 'type_values' in credentialQuery.meta
+    )
+
+    if (w3cSdJwts.length > 0) {
+      const w3cV2CredentialRepository = agentContext.dependencyManager.resolve(W3cV2CredentialRepository)
+
+      const w3cV2Records = await w3cV2CredentialRepository.findByQuery(agentContext, {
+        claimFormat: ClaimFormat.SdJwtW3cVc,
+        $or: dcqlQuery.credentials
+          .flatMap((c) => (c.format === 'jwt_vc_json' ? c.meta.type_values : []))
```
TimoGlastra:
```suggestion
          .flatMap((c) => (c.format === 'vc+sd-jwt' ? c.meta.type_values : []))
```


Right?

packages/core/src/utils/validators.ts
```diff
@@ -72,6 +72,40 @@ export function IsInstanceOrArrayOfInstances(
...
+
+          return (
+            asArray(values)
+              // all values MUST be instance of one of the class types
+              .every((value) => classTypes.some((classType) => isInstance(value, classType)))
+          )
+        },
+        defaultMessage: buildMessage(
+          (eachPrefix) =>
+            `${eachPrefix}$property value must be a string, an instance of, or an array of instances containing ${classTypes
```
TimoGlastra, line 98: Is the `$` intentional?

```suggestion
            `${eachPrefix} property value must be a string, an instance of, or an array of instances containing ${classTypes
```

packages/core/src/modules/vc/validators.ts
```diff
@@ -28,3 +48,45 @@ export function IsCredentialJsonLdContext(validationOptions?: ValidationOptions)
...
+      name: 'IsVerifiableCredentialType',
+      validator: {
+        validate: (value): boolean => {
+          return Array.isArray(value)
+            ? value.includes(VERIFIABLE_CREDENTIAL_TYPE) && value.every((v) => typeof v === 'string')
+            : value === VERIFIABLE_CREDENTIAL_TYPE
+        },
+        defaultMessage: buildMessage(
+          (eachPrefix) =>
+            `${eachPrefix}$property must be "VerifiableCredential" or an array of strings which includes "VerifiableCredential"`,
```
TimoGlastra, line 64: Same here and in below transformer, guess it's intentional then?

packages/core/src/modules/vc/sd-jwt-vc/__tests__/W3cV2SdJwtCredentialService.test.ts
```diff
@@ -0,0 +1,341 @@
...
+      expect(result.validations.dataModel?.error?.message).toContain('JWT expired at 1698151532')
+    })
+  })
+
+  describe('signPresentation', () => {
+    test('signs an ES256 JWT vp', async () => {
+      // Create a new instance of the credential from the serialized JWT
+      const parsedJwtVc = W3cV2SdJwtVerifiableCredential.fromCompact(CredoEs256DidJwkJwtVc)
+
+      const envelopedCredential = W3cV2EnvelopedVerifiableCredential.fromVerifiableCredential(parsedJwtVc)
```
TimoGlastra, line 322: How do we deal with selective disclosure? In SdJwtVc service we have a `.present` where you can provide a frame. Do we have something similar here? 

What about ensuring we create a valid presentation, still including all the required fields from the W3C VC spec, do we check that based on the class-validator class (and thus run the validation at least once?)

packages/core/src/modules/vc/sd-jwt-vc/__tests__/W3cV2SdJwtCredentialService.test.ts
```diff
@@ -0,0 +1,341 @@
...
+        },
+      })
+
+      expect(result.validations.dataModel?.error?.message).toContain('Failed to validate class')
+    })
+
+    test('returns invalid result when credential is expired', async () => {
+      const jwtVc = W3cV2SdJwtVerifiableCredential.fromCompact(CredoEs256DidJwkJwtVc)
+
+      jwtVc.sdJwt.payload.exp = new Date('2020-01-01').getTime() / 1000
```
TimoGlastra: might work, but good to prevent decimal values as part of the jwt exp

```suggestion
      jwtVc.sdJwt.payload.exp = Math.floor(new Date('2020-01-01').getTime() / 1000)
```

packages/core/src/modules/vc/sd-jwt-vc/W3cV2SdJwtVerifiablePresentation.ts
```diff
@@ -0,0 +1,54 @@
...
+export interface W3cV2SdJwtVerifiablePresentationOptions {
+  sdJwt: W3cV2SdJwt<ClaimFormat.SdJwtW3cVp>
+}
+
+/**
+ * Represents a Verifiable Presentation encoded as a SD-JWT.
+ *
+ * @see https://www.w3.org/TR/vc-jose-cose/#securing-vps-sd-jwt
+ */
+export class W3cV2SdJwtVerifiablePresentation {
```
TimoGlastra, line 16: Do we use an SD-JWT or JWT as presentation? I'm not actually sure what came out of the thread whether to do JWT or SD-JWT presentation

packages/core/src/modules/vc/sd-jwt-vc/W3cV2SdJwtCredentialService.ts
```diff
@@ -0,0 +1,544 @@
...
+    })
+
+    const compact = await sdjwt.issue<W3cV2JsonCredential>(
+      {
+        ...payload,
+        cnf: holderBinding?.cnf,
+        iss: parsedDid.did,
+        iat: nowInSeconds(),
+      },
+      options.disclosureFrame as DisclosureFrame<W3cV2JsonCredential>,
```
TimoGlastra: Do we ensure:
> When securing [verifiable credentials](https://www.w3.org/TR/vc-jose-cose/#dfn-verifiable-credentials) with [[SD-JWT](https://www.w3.org/TR/vc-jose-cose/#bib-sd-jwt)], implementers SHOULD ensure that properties necessary for the validation and verification of a credential are NOT selectively disclosable (i.e., such properties SHOULD be disclosed). These properties can include but are not limited to [@context](https://www.w3.org/TR/vc-data-model-2.0/#contexts), [type](https://www.w3.org/TR/vc-data-model-2.0/#types), [credentialStatus](https://www.w3.org/TR/vc-data-model-2.0/#status), [credentialSchema](https://www.w3.org/TR/vc-data-model-2.0/#data-schemas), and [relatedResource](https://www.w3.org/TR/vc-data-model-2.0/#integrity-of-related-resources).

packages/core/src/modules/vc/sd-jwt-vc/W3cV2SdJwtCredentialService.ts
```diff
@@ -0,0 +1,544 @@
...
+    const sdjwt = new SDJwtInstance({
+      ...this.getBaseSdJwtConfig(agentContext),
+      signer: getSdJwtSigner(agentContext, publicJwk),
+      hashAlg: options.hashingAlgorithm ?? 'sha-256',
+      signAlg: options.alg,
+    })
+
+    const compact = await sdjwt.issue<W3cV2JsonCredential>(
+      {
+        ...payload,
```
TimoGlastra: Do we verify:
> The JWT Claim Names vc and vp MUST NOT be present.

packages/core/src/modules/vc/sd-jwt-vc/W3cV2SdJwtCredentialService.ts
```diff
@@ -0,0 +1,544 @@
...
+      } catch (error) {
+        validationResults.validations.dataModel = {
+          isValid: false,
+          error,
+        }
+
+        return validationResults
+      }
+
+      const issuerVerificationMethod = await this.getVerificationMethodForJwtCredential(agentContext, {
```
TimoGlastra: How do we deal with this:
> Implementers SHOULD avoid setting JWT claims to values that conflict with the values of [verifiable credential](https://www.w3.org/TR/vc-jose-cose/#dfn-verifiable-credentials) properties when a claim and property pair refer to the same conceptual entity, especially with pairs such as iss and issuer, jti and id, and sub and credentialSubject.id. For example, JWK claim iss SHOULD NOT be set to a value which conflicts with the value of [verifiable credential](https://www.w3.org/TR/vc-jose-cose/#dfn-verifiable-credentials) property issuer.

It mentions SHOULD, what do we do if they don't match?

packages/core/src/modules/vc/sd-jwt-vc/W3cV2SdJwtCredentialService.ts
```diff
@@ -0,0 +1,544 @@
...
+    // The JWT payload is simply the presentation
+    const payload = JsonTransformer.toJSON(options.presentation) as W3cV2JsonPresentation
+
+    if (!isDid(options.verificationMethod)) {
+      throw new CredoError('Only did identifiers are supported as verification method')
+    }
+
+    const { publicJwk, parsedDid } = await this.resolveVerificationMethod(agentContext, options.verificationMethod, [
+      'assertionMethod',
+    ])
```
TimoGlastra: Do we check that the provided verificdtaion method matches one of the credentialSubject.ids?

packages/core/src/modules/vc/sd-jwt-vc/W3cV2SdJwtCredentialService.ts
TimoGlastra: Especially `cnf`/`sub` and `credentialSubject.id`. I'm also not sure how do `cnf` and `sub` relate to `credentialSubject.id`?

packages/core/src/modules/vc/sd-jwt-vc/W3cV2SdJwtCredentialService.ts
```diff
@@ -0,0 +1,544 @@
...
+      const credentials = asArray(presentation.resolvedPresentation.verifiableCredential)
+
+      // Verify all credentials in parallel, and await the result
+      validationResults.validations.credentials = await Promise.all(
+        credentials.map(async (credential) => {
+          if (!(credential.envelopedCredential instanceof W3cV2SdJwtVerifiableCredential)) {
+            return {
+              isValid: false,
+              error: new CredoError(
+                'Credential is of format SD-JWT. Presentations in JWT format can only contain credentials in JWT format.'
```
TimoGlastra: I think this error message is wrong? 
```suggestion
                'Credential is not of format SD-JWT. Presentations in SD-JWT format can only contain credentials in SD-JWT format.'
```

packages/core/src/modules/vc/sd-jwt-vc/W3cV2SdJwtCredentialService.ts
```diff
@@ -0,0 +1,544 @@
...
+          let credentialSubjectAuthentication: SingleValidationResult
+
+          // Check whether any of the credentialSubjectIds for each credential is the same as the controller of the verificationMethod
+          // This authenticates the presentation creator controls one of the credentialSubject ids.
+          // NOTE: this doesn't take into account the case where the credentialSubject is no the holder. In the
+          // future we can add support for other flows, but for now this is the most common use case.
+          // TODO: should this be handled on a higher level? I don't really see it being handled in the jsonld lib
+          // or in the did-jwt-vc lib (it seems they don't even verify the credentials itself), but we probably need some
+          // more experience on the use cases before we loosen the restrictions (as it means we need to handle it on a higher layer).
+          const credentialSubjectIds = credential.resolvedCredential.credentialSubjectIds
```
TimoGlastra, line 360: we should somehow check the relation between `cnf`/`sub`/`credentialSubject.id` here?

packages/core/src/modules/vc/sd-jwt-vc/W3cV2SdJwtCredentialService.ts
TimoGlastra: For `iss` and `issuer` they MUST match if both present. Do we check that?

packages/core/src/modules/vc/jwt-vc/W3cV2JwtCredentialService.ts
```diff
@@ -0,0 +1,531 @@
...
+          MessageValidator.validateSync(options.credential.resolvedCredential)
+        }
+
+        credential =
+          options.credential instanceof W3cV2JwtVerifiableCredential
+            ? options.credential
+            : W3cV2JwtVerifiableCredential.fromCompact(options.credential)
+
+        // Verify the JWT payload (verifies whether it's not expired, etc...)
+        credential.jwt.payload.validate()
```
TimoGlastra, line 109: similar notes as to the SdJwt service regarding VCDM specific validations, and  JWT <-> VCDM matching of values

packages/core/src/modules/vc/W3cV2CredentialsModuleConfig.ts
```diff
@@ -0,0 +1,46 @@
...
+   *       document: null
+   *     }
+   *   }
+   * }
+   * ```
+   *
+   *
+   * @default {@link defaultDocumentLoader}
+   */
+  documentLoader?: DocumentLoaderWithContext
```
TimoGlastra: This is not used right? Since we don't resolve the JSON-LD in v2 module yet?

packages/core/src/modules/vc/W3cV2CredentialsModule.ts
```diff
@@ -0,0 +1,56 @@
...
+        VERIFICATION_METHOD_TYPE_ED25519_VERIFICATION_KEY_2018,
+        VERIFICATION_METHOD_TYPE_ED25519_VERIFICATION_KEY_2020,
+      ],
+      supportedPublicJwkTypes: [Ed25519PublicJwk],
+    } satisfies SuiteInfo)
+    dependencyManager.registerInstance(SignatureSuiteToken, {
+      suiteClass: Ed25519Signature2020,
+      proofType: 'Ed25519Signature2020',
+      verificationMethodTypes: [VERIFICATION_METHOD_TYPE_ED25519_VERIFICATION_KEY_2020],
+      supportedPublicJwkTypes: [Ed25519PublicJwk],
```
TimoGlastra: these are not used for v2 right?

packages/core/src/utils/validators.ts
hacdias: This is correct.

packages/core/src/modules/vc/validators.ts
hacdias: This is correct.

packages/core/src/modules/vc/sd-jwt-vc/__tests__/W3cV2SdJwtCredentialService.test.ts
hacdias: My initial idea was that the user would do the following:

1. Having the compact SD-JWT, call `SdJwtService.present` with the disclosure frame they want.
2. Then parse the outputted SD JWT with `W3cV2SdJwtVerifiableCredential` and from there make an enveloped credential.

I can add a `present` method to the `W3cV2SdJwtCredentialService` that ensures that the required fields are kept.

packages/core/src/modules/vc/sd-jwt-vc/W3cV2SdJwtVerifiablePresentation.ts
hacdias: If you ask for `vc+sd-jwt`, SD-JWT. The [specification](https://www.w3.org/TR/vc-jose-cose/#securing-vps-sd-jwt) also states that for JWT presentations, the credentials must be JWTs and for SD-JWT presentations, the credentials must be SD-JWTs.

> **3.1.2 Securing JSON-LD Verifiable Presentations with JOSE**
> (...)
> Credentials in [verifiable presentations](https://www.w3.org/TR/vc-data-model-2.0/#dfn-verifiable-presentation) MUST be secured. In this case, these [credentials](https://www.w3.org/TR/vc-data-model-2.0/#dfn-credential) are secured using JWS.

> **3.2.2 Securing JSON-LD Verifiable Presentations with SD-JWT**
> (...) 
> Credentials in [verifiable presentations](https://www.w3.org/TR/vc-data-model-2.0/#dfn-verifiable-presentation) MUST be secured. These [credentials](https://www.w3.org/TR/vc-data-model-2.0/#dfn-credential) are secured using SD-JWT in this case.

My conclusion is that you can't mix and match.

packages/core/src/modules/vc/sd-jwt-vc/__tests__/W3cV2SdJwtCredentialService.test.ts
TimoGlastra: Hmm but there's a difference between an SdJwt and an SdJwtVc, and so the SdJwtVc service might do specific checks that are not relevant in case of SDJwt. We basically need a method to remove disclosures from a compact SD-JWT based on a disclosure frame.

But i think the integration with DCQL is fine for now (that automatically applied the discuslore frame right?) and we can extend it with more public APIs later

packages/core/src/modules/vc/sd-jwt-vc/W3cV2SdJwtCredentialService.ts
hacdias: We validate it when decoding the SD-JWT and throw an error if they don't match.

packages/core/src/modules/vc/sd-jwt-vc/W3cV2SdJwtVerifiablePresentation.ts
TimoGlastra: okay, makes sense 👍 . That diminishes the value of W3C creds even more, since you can't present two creds with different signature formats in a single presentation

hacdias: Yeah...

packages/core/src/modules/vc/sd-jwt-vc/W3cV2SdJwtCredentialService.ts
hacdias: This check has now been added

packages/core/src/modules/vc/sd-jwt-vc/W3cV2SdJwtCredentialService.ts
hacdias: This check has now been added

packages/core/src/modules/vc/jwt-vc/W3cV2JwtCredentialService.ts
hacdias: This check has now been added

packages/core/src/modules/vc/sd-jwt-vc/W3cV2SdJwtCredentialService.ts
hacdias: Largely based on what we do on V1. I did not add those additional verifications.

packages/core/src/modules/vc/sd-jwt-vc/__tests__/W3cV2SdJwtCredentialService.test.ts
hacdias: I've added a method

packages/core/src/modules/vc/sd-jwt-vc/W3cV2SdJwtCredentialService.ts
hacdias: I updated the validation. I was under the impression that the SD JWT package would do the checks between `cnf` and `sub`, or not? At least, we're not doing it in the regular SdJwtService.

---

I also added a separate check for `sub` and `credentialSubject.id`.

hacdias:
@TimoGlastra I addressed (or replied) to the feedback. I also noticed there were some things missing in the DCQL integration, which I've now fixed. So it's ready for a new review!

I still want to add one or two E2E tests, so let's not yet merge.

packages/core/src/modules/vc/W3cV2CredentialServiceOptions.ts
```diff
@@ -0,0 +1,229 @@
...
+   *
+   * @default 'sha-256'
+   */
+  hashingAlgorithm?: Exclude<HashName, 'sha-1'>
+
+  /**
+   * The holder to bind the SD-JWT VC to. If no holder is provided, the SD-JWT
+   * VC won't be bound to the holder.
+   */
+  holder?: SdJwtVcHolderBinding
```
TimoGlastra, line 90: I think we should not reuse the `SdJwtVcHolderBinding` here, since that also allows `jwk` binding. I'm not sure how that will work combined with `credntialSubject.id`? For W3C v1 we only support dids and i think it might make sense to also do that for v2? 

Currently i think you have to set the `credentialSubject.id` in the credential, and there's no `holder` property.

We do check that the `cnf` and `credentialSubject.id` match, so what would happen if we provide a jwk here?

hacdias: The usage of `cnf` seems to be recommended: https://www.w3.org/TR/vc-jose-cose/#cnf

We should indeed not allow a random JWK. We must support DIDs only for now.

We can either remove support for holder binding, or:

- Keep `holder` and make it `holder?: SdJwtVcHolderDidBinding`
- Match the given holder binding DID to one in `options.credential.credentialSubjectIds`. Otherwise error.
- Make the `cnf` using the given holder binding

?

packages/drizzle-storage/src/core/w3c-v2-credential-record/postgres.ts
```diff
@@ -0,0 +1,25 @@
+import { ClaimFormat, W3cVerifiableCredential } from '@credo-ts/core'
+import { jsonb, pgEnum, pgTable, text } from 'drizzle-orm/pg-core'
+import { getPostgresBaseRecordTable, postgresBaseRecordIndexes } from '../../postgres/baseRecord'
+
+export const w3cV2CredentialClaimFormat = pgEnum('W3cV2ClaimFormat', [ClaimFormat.SdJwtW3cVc, ClaimFormat.JwtW3cVc])
```
TimoGlastra: Small adjustment needed because of changes in #2411 
```suggestion
import type { ClaimFormat, W3cVerifiableCredential } from '@credo-ts/core'
import { jsonb, pgEnum, pgTable, text } from 'drizzle-orm/pg-core'
import { getPostgresBaseRecordTable, postgresBaseRecordIndexes } from '../../postgres/baseRecord'

export const w3cV2CredentialClaimFormat = pgEnum('W3cV2ClaimFormat', ['vc+sd-jwt' satisfies  `${ClaimFormat.SdJwtW3cVc}`, 'vc+jwt' satisfies`${ClaimFormat.JwtW3cVc}`])
```

packages/drizzle-storage/src/core/w3c-v2-credential-record/sqlite.ts
```diff
@@ -0,0 +1,23 @@
+import { ClaimFormat, W3cV2VerifiableCredential } from '@credo-ts/core'
```
TimoGlastra:
```suggestion
import type { ClaimFormat, W3cV2VerifiableCredential } from '@credo-ts/core'
```

packages/core/src/modules/dcql/DcqlService.ts
```diff
@@ -714,28 +714,18 @@ export class DcqlService {
...
 
-          // Determine a suitable verification method for the presentation
-          const verificationMethod = await this.getVerificationMethodForSubjectId(
-            agentContext,
-            presentationToCreate.subjectIds[0]
-          )
-
           const w3cV2CredentialService = agentContext.resolve(W3cV2CredentialService)
           const w3cV2Presentation = new W3cV2Presentation({
+            holder: presentationToCreate.credentialRecord.credential.resolvedCredential.credentialSchemaIds[0],
```
TimoGlastra: This seems incorrect. It should not be `credentialSchemaIds` right? Because that's the id from `credentialSchema.id` (not from `credentialSubject.id`)

packages/core/src/modules/vc/v2-jwt-utils.ts
```diff
@@ -0,0 +1,167 @@
...
+
+  const didResolver = agentContext.dependencyManager.resolve(DidResolverService)
+  let verificationMethod: VerificationMethod
+
+  // If the kid starts with # we assume it is a relative did url, and we resolve
+  //it based on the `iss` and the `kid`
+  if (kid?.startsWith('#')) {
+    if (!iss) {
+      throw new CredoError(`JWT 'kid' MUST be absolute when when no 'iss' is present in JWT payload`)
+    }
```
TimoGlastra: I think we should also check here that the `iss` is a did, and provide a clear error message

packages/core/src/modules/vc/sd-jwt-vc/__tests__/W3cV2SdJwtCredentialService.test.ts
```diff
@@ -327,11 +331,11 @@ describe('W3cV2SdJwtCredentialService', () => {
 
       const signedJwtVp = await w3cV2JwtCredentialService.signPresentation(agentContext, {
         presentation,
-        alg: KnownJwaSignatureAlgorithms.EdDSA,
+        // alg: KnownJwaSignatureAlgorithms.EdDSA,
```
TimoGlastra: can we remove these comments? or do they need to be uncommented?

packages/core/src/modules/vc/jwt-vc/__tests__/W3cV2JwtCredentialService.test.ts
```diff
@@ -340,11 +349,11 @@ describe('W3cV2JwtCredentialService', () => {
 
       const signedJwtVp = await w3cV2JwtCredentialService.signPresentation(agentContext, {
         presentation,
-        alg: KnownJwaSignatureAlgorithms.EdDSA,
+        // alg: KnownJwaSignatureAlgorithms.EdDSA,
```
TimoGlastra: same here

packages/core/src/modules/vc/jwt-vc/W3cV2JwtVerifiablePresentation.ts
```diff
@@ -76,5 +77,12 @@ export class W3cV2JwtVerifiablePresentation {
     if ('cyt' in header && header.cyt !== 'vp') {
       throw new CredoError(`The provided W3C VP JWT does not have the correct 'cyt' header.`)
     }
+
+    const iss = header.iss ?? payload.iss
```
TimoGlastra, line 81: Does header `iss` take precedence? I've never seen `iss` in header

hacdias: Me neither, but that's what they say: https://www.w3.org/TR/vc-jose-cose/#iss

> If iss is present in the [JOSE Header](https://www.rfc-editor.org/rfc/rfc7515#section-4.1), the [JWT Claims](https://www.rfc-editor.org/rfc/rfc7519#section-4.1.1), or the [COSE Header](https://www.rfc-editor.org/rfc/rfc9052#section-3), a [verifier](https://www.w3.org/TR/vc-data-model-2.0/#dfn-verifier) can use this parameter to obtain a [JSON Web Key](https://www.rfc-editor.org/rfc/rfc7517#section-4) to use in the [verification](https://www.w3.org/TR/vc-data-model-2.0/#dfn-verify) process.

kmk142789 (approved):

PR #2390, opened 2025-08-30
fix(vc): allow selecting proof purpose for VPs

genaris:
This is mainly to allow using `assertionMethod` proof purposes, which is needed in [Linked VPs](https://identity.foundation/linked-vp/) that are bound to a DID and not meant for authentication.

When we don't select any proof purpose, `@digitalcredentials/jsonld-signatures` library will take it as `authentication` and therefore require `challenge` and use `domain`. So this subtle change allows users to use any kind of proof purpose they want.

I'm not sure if this would be also useful for JWT credentials. Currently only implemented for JSON-LD because these are the ones we are creating for Linked VPs and did:webvh.


TimoGlastra (approved):
I think also makes sense to enable this for jwt creds!

PR #2396, opened 2025-09-02
feat: add registerSchema

lotharking:


packages/webvh/src/anoncreds/services/WebVhAnonCredsRegistry.ts
```diff
@@ -425,14 +427,37 @@ export class WebVhAnonCredsRegistry implements AnonCredsRegistry {
...
+      schemaState: { state: 'finished', schema: options.schema, schemaId },
+      registrationMetadata: {},
+      schemaMetadata: {},
+    }
   }
 
-  public async registerCredentialDefinition(agentContext: AgentContext): Promise<RegisterCredentialDefinitionReturn> {
-    agentContext.config.logger.warn('registerCredentialDefinition not implemented for WebVhAnonCredsRegistry')
-    throw new CredoError('Method not implemented.')
+  public async registerCredentialDefinition(
```
TimoGlastra, line 471: If we add these methods it would be great to also add the other registration methods.

Also, I'm thinking whether we need to return state `action` instead? Since you still need to publish it before it's finished, but in that case we don't store it yet, and you'd need to call `register` again, so that's not ideal. So i think this is fine.

TimoGlastra: The benefit of using action though, is that if anything fails with publishing, it won't get stored as a valid anoncreds object in local storage.

TimoGlastra: Maybe we should allow providing an `uploadAnoncredsResource` method in the constructor of the `WebVhAnonCredsRegistry` class? And that it uses that method to upload the resource? Because we will also have to sign the resource before hosting it right?


@brianorwhatever any thoughts?

swcurran: Also pinging @PatStLouis on this one.

brianorwhatever: Yeah we will need to sign the resource so an `uploadAnoncredsResource` method makes sense. Would that make the `action` state and a second `register` not required anymore? I'm also wondering how does the hosting of the resource work?

lotharking:
A new fork will be created targeting the `v0.5.x` branch https://github.com/openwallet-foundation/credo-ts/pull/2419. Once that work is completed, this pull request will be resumed.

packages/webvh/src/anoncreds/services/WebVhAnonCredsRegistry.ts
PatStLouis: @TimoGlastra have a look here at how I've implemented the registry in the acapy plugin
https://github.com/openwallet-foundation/acapy-plugins/blob/main/webvh/webvh/anoncreds/registry.py

At the bottom of the file, you will find the update/sign/upload functions. This plugin was designed around the didwebvh server, which credo will also be able to interact with/leverage if it follows the same pattern.

PatStLouis: We added some optional endorsement steps which you can ignore for the time being

packages/webvh/src/dids/WebVhDidRegistrar.ts
```diff
@@ -0,0 +1,218 @@
...
+      // Create DID
+      const { did, doc, log } = await createDID({
+        domain,
+        paths,
+        signer,
+        updateKeys: [publicKeyMultibase],
+        verificationMethods: [
+          {
+            controller: baseDid,
+            type: 'Ed25519VerificationKey2020',
```
PatStLouis:
```suggestion
            type: 'Multikey',
```

If this key is intended to sign the attested resources with eddsa-jcs-2022, I would recommend using the `Multikey` type to favor interop as per the specification:
https://www.w3.org/TR/vc-di-eddsa/#multikey

packages/webvh/src/anoncreds/services/WebVhAnonCredsRegistry.ts
```diff
@@ -460,4 +607,116 @@ export class WebVhAnonCredsRegistry implements AnonCredsRegistry {
...
+      metadata: Record<string, unknown>
+      issuerId: string
+      verificationMethod?: string
+    }
+  ) {
+    const verificationMethodId = await this.getVerificationMethodId(agentContext, issuerId, verificationMethod)
+
+    // Prepare the generic resource payload to be signed
+    const resourcePayload = {
+      '@context': ['https://opsecid.github.io/attested-resource/v1', 'https://w3id.org/security/data-integrity/v2'],
```
PatStLouis:
```suggestion
      '@context': ['https://identity.foundation/did-attested-resources/context/v0.1', 'https://w3id.org/security/data-integrity/v2'],
```
The spec is now published and the context is available at this endpoint

packages/webvh/src/anoncreds/services/__tests__/mock-resources.ts
```diff
@@ -37,6 +37,16 @@ export const mockResolvedDidDocument = {
...
+  role: 'created',
+  didDocument: mockResolvedDidDocument,
+  getTag: (key: string) => {
+    if (key === 'domain') return 'id.test-suite.app/credo/01'
+    return undefined
+  },
+}
+
 export const mockSchemaResource = {
   '@context': ['https://opsecid.github.io/attested-resource/v1', 'https://w3id.org/security/data-integrity/v2'],
```
PatStLouis, line 41:
```suggestion
  '@context': ['https://identity.foundation/did-attested-resources/context/v0.1', 'https://w3id.org/security/data-integrity/v2'],
```

PatStLouis (commented):
Might want to make these context URLs into variables

PatStLouis:
@lotharking would you mind trying to register a DID with this work on the https://id.test-suite.app/ test server?

just generate a did log with the following `did:webvh:{SCID}:id.test-suite.app:{namespace}:{identifier}` such as:
`did:webvh:{SCID}:id.test-suite.app:credo-ts:8f9366b5-a1ad-40b2-b524-a0754e562d14`

You can use the openapi interface to upload the log entry:
https://id.test-suite.app/docs#/Identifiers/new_log_entry__namespace___identifier__post

This endpoint has 2 properties:
- `logEntry`: the single json line of the new entry
- `witnessSignature`(optional): the single witness signature for the log entry

This will ensure interop of the solution. Feel free to write a test for this if deemed valuable.

You can also test updates by sending subsequent log entries to that same endpoint.

Once the DID is created, you can test uploading a resource (such as a schema) here:
https://id.test-suite.app/docs#/Attested%20Resources/upload_attested_resource__namespace___identifier__resources_post

You may ignore the `options` property on this second endpoint, and only use the `attestedResource` property.

packages/webvh/src/dids/WebVhDidRegistrar.ts
```diff
@@ -0,0 +1,227 @@
...
+      const { log: logResult, doc } = await updateDID({
+        log,
+        signer,
+        verifier,
+        domain,
+        updateKeys,
+        ...inputDidDocument,
+        verificationMethods: verificationMethods.map((vm) => ({
+          ...vm,
+          publicKeyMultibase: vm.publicKeyMultibase ?? '',
```
genaris: Don't forget here the same changes from #2440

packages/webvh/src/dids/WebVhDidRegistrar.ts
genaris, line 1: I've just noticed that in this module we are prefixing most classes with "Webvh", except WebVhAnonCredsRegistry. I think we'll need to be consistent, so we have two options:

- Rename everything to Webvh*: easiest path, since only WebVhAnonCredsRegistry is to be refactored (WebVhDidRegistrar is added in this PR... but of course we'll need to update it in 0.5.x branch before releasing the next stable version)
- Rename everything to WebVh*: some more effort, probably more correct (if we consider it an acronym of "Web with Verifiable History") but in the spec it is always referred to this method as did:webvh (all lowercase), so no conclusive way from my perspective

@TimoGlastra what do you think about this topic?

genaris: @lotharking as discussed in today WG call, I think we'll need to prefix all related classes and interfaces to WebVh, since it is more consistent with other modules (e.g. IndyVdr for Indy VDR). Can you do this refactor here and also for v0.5.x branch?

packages/webvh/src/dids/WebVhDidRegistrar.ts
lotharking: implemented

packages/webvh/src/dids/WebVhDidRegistrar.ts
lotharking: Implemented on https://github.com/openwallet-foundation/credo-ts/pull/2470

packages/webvh/README.md
```diff
@@ -0,0 +1,212 @@
+<p align="center">
+  <br />
+  <img
+    alt="Credo Logo"
+    src="https://github.com/openwallet-foundation/credo-ts/blob/c7886cb8377ceb8ee4efe8d264211e561a75072d/images/credo-logo.png"
+    height="250px"
+  />
+</p>
+<h1 align="center"><b>Credo WebVh Module</b></h1>
```
genaris:
```suggestion
<h1 align="center"><b>Credo did:webvh Module</b></h1>
```

PR #2399, opened 2025-09-03
feat: Add websocket priority and filtering to pickup live delivery

jamshale:
This adds the pickup live mode repairs https://github.com/openwallet-foundation/credo-ts/pull/2381 to target 0.6.x release.

### DIFF
--- a/packages/didcomm/src/modules/message-pickup/services/MessagePickupSessionService.ts
+++ b/packages/didcomm/src/modules/message-pickup/services/MessagePickupSessionService.ts
@@ -4,7 +4,7 @@ import type { MessagePickupLiveSessionRemovedEvent, MessagePickupLiveSessionSave
 import type { MessagePickupSession, MessagePickupSessionRole } from '../MessagePickupSession'
 
 import { EventEmitter, InjectionSymbols, injectable, utils } from '@credo-ts/core'
-import { filter, type Subject, takeUntil } from 'rxjs'
+import { type Subject, filter, takeUntil } from 'rxjs'
 
 import { TransportEventTypes } from '../../../transport'
 import { MessagePickupEventTypes } from '../MessagePickupEvents'
--- a/packages/didcomm/src/modules/message-pickup/__tests__/MessagePickupApi.test.ts
+++ b/packages/didcomm/src/modules/message-pickup/__tests__/MessagePickupApi.test.ts
@@ -4,17 +4,19 @@ import type { MessagePickupProtocol } from '../protocol/MessagePickupProtocol'
 
 import { Subject } from 'rxjs'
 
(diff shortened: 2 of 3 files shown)


packages/didcomm/src/modules/message-pickup/services/MessagePickupSessionService.ts
```diff
@@ -31,7 +31,10 @@ export class MessagePickupSessionService {
 
     eventEmitter
       .observable<TransportSessionRemovedEvent>(TransportEventTypes.TransportSessionRemoved)
-      .pipe(takeUntil(stop$))
+      .pipe(
+        filter((e) => e.payload.session.type === 'WebSocket'),
```
TimoGlastra, line 35: Why do we only remove these in case of websocket? This makes the session service only work with websickets? 

And it prevents from new duplex transports being supported dynamically.

Where there cases were wrong sessions were removed?

TimoGlastra: I see in the ither pr:

> Then for the removing of the pickup session when in live mode delivery, I changed this to only happen when websocket sessions close. All other delivery modes close pickup sessions on every session closed event.

But if this is the case, we don't have to filter based on websicket right, we can keep the code as it was?

genaris: > Why do we only remove these in case of websocket? This makes the session service only work with websickets?
> 
> And it prevents from new duplex transports being supported dynamically.
> 
> Where there cases were wrong sessions were removed?

The problem @jamshale was experiencing is that the same agent was creating both HTTP and WS sessions, so when the HTTP one was being closed, this event handler was clearing the Live mode session associated to it.

In practice, at the moment we only support WebSocket for Live mode sessions. Of course this hard-coding is not ideal since you might also add long-lived HTTP sessions with SSE (as you mentioned some time ago). I also don't like a lot the "['wss', 'ws']" we are hardcoding in transport preference when sending messages.

Probably an easy improvement we can do for 0.6.0 (without making big changes in the way Transport Sessions work) is to add some flags for InboundTransports to indicate if they are persistent or not, and put their type/schemes in these filters as required.

TimoGlastra: Okay, i understand it better now. I think we should store the transport session in the live session, so we can do a match? That way we only remove the live session based on if a matching transport session is closed, rather than harcodring WS

genaris: I think it makes more sense like that. Look at https://github.com/openwallet-foundation/credo-ts/pull/2408

PR #2405, opened 2025-09-11
feat: Add Hedera module

AlexanderShenshin:
This PR adds Hedera module, enabling support for [Hedera DID Method](https://github.com/hashgraph/did-method/blob/master/hedera-did-method-specification.md) and AnonCreds registry (see [Hedera/Hiero AnonCreds Method](https://dsrcorporation.github.io/hedera-anoncreds-method/)).

Includes automated tests (unit + integration) and integration with built-in Credo demo.

The implementation is based on [Hiero DID SDK JS](https://github.com/hiero-ledger/hiero-did-sdk-js).

### DIFF
--- a/.changeset/neat-starfishes-repeat.md
+++ b/.changeset/neat-starfishes-repeat.md
@@ -1,0 +1,5 @@
+---
+"@credo-ts/didcomm": patch
+---
+
+fix: allow did URLs in thread ids
--- a/packages/didcomm/src/decorators/thread/ThreadDecorator.test.ts
+++ b/packages/didcomm/src/decorators/thread/ThreadDecorator.test.ts
@@ -1,3 +1,5 @@
+import { validate } from 'class-validator'
+
 import { JsonTransformer } from '../../../../core'
 
 import { ThreadDecorator } from './ThreadDecorator'
@@ -44,5 +46,87 @@ describe('Decorators | ThreadDecorator', () => {
     }
 
     expect(json).toEqual(transformed)
+  })
+  describe('PthidRegExp validation', () => {
+    it('should accept valid message IDs for parentThreadId', async () => {
+      const decorator = new ThreadDecorator({
+        parentThreadId: 'ceffce22-6471-43e4-8945-b604091981c9',
+      })
+      const validationErrors = await validate(decorator)
+      expect(validationErrors).toHaveLength(0)
+    })
+
+    it('should accept valid DIDs for parentThreadId', async () => {
+      const validDids = [
+        'did:example:123456789abcdefghi',
+        'did:key:z6MkpTHR8VNsBxYAAWHut2Geadd9jSwuBV8xRoAnwWsdvktH',
+        'did:web:example.com',
+        'did:web:example%3Acom',
+        'did:sov:3ecf688c-cb3f-467b-8636-6b0c7f1d9022',
+        'did:indy:sovrin:staging:3ecf688c-cb3f-467b-8636-6b0c7f1d9022',
+        'did:peer:1zQmNbr8L7xqgCN7aQWG4TgH8xrp9Z3nR2vY6CqK8uDx',
+      ]
+
+      for (const did of validDids) {
+        const decorator = new ThreadDecorator({
+          parentThreadId: did,
+        })
+        const validationErrors = await validate(decorator)
+        expect(validationErrors).toHaveLength(0)
+      }
+    })
+
+    it('should accept valid DID URLs with query and fragment for parentThreadId', async () => {
+      const validDidUrls = [
+        'did:example:123456789abcdefghi?service=agent&relativeRef=my%20ref',
+        'did:example:123456789abcdefghi#keys-1',
+        'did:example:123456789abcdefghi?query=value#fragment',
+      ]
+
+      for (const didUrl of validDidUrls) {
+        const decorator = new ThreadDecorator({
+          parentThreadId: didUrl,
+        })
+        const validationErrors = await validate(decorator)
+        expect(validationErrors).toHaveLength(0)
+      }
+    })
+
+    it('should reject invalid formats for parentThreadId', async () => {
+      const invalidValues = [
+        'short', // too short
+        'invalid-did:format', // invalid DID format
+        'did:', // incomplete DID
+        'did:method', // incomplete DID
+        '12345678901234567890123456789012345678901234567890123456789012345', // too long
+        'invalid!@#$%^&*()format', // invalid characters
+        'did:example:incomplete%ZZencoding', // invalid percent-encoding
+      ]
+
+      for (const invalidValue of invalidValues) {
+        const decorator = new ThreadDecorator({
+          parentThreadId: invalidValue,
+        })
+        const validationErrors = await validate(decorator)
+        expect(validationErrors.length).toBeGreaterThan(0) // expect to fail on each invalid value
+        expect(validationErrors[0].property).toBe('parentThreadId')
+      }
+    })
+
+    it('should accept valid message IDs for threadId', async () => {
+      const decorator = new ThreadDecorator({
+        threadId: 'ceffce22-6471-43e4-8945-b604091981c9',
+      })
+      const validationErrors = await validate(decorator)
+      expect(validationErrors).toHaveLength(0)
+    })
+
+    it('should accept valid DIDs for threadId', async () => {
+      const decorator = new ThreadDecorator({
+        threadId: 'did:example:123456789abcdefghi',
+      })
+      const validationErrors = await validate(decorator)
+      expect(validationErrors).toHaveLength(0)
+    })
   })
 })
--- a/packages/didcomm/src/decorators/thread/ThreadDecorator.ts
+++ b/packages/didcomm/src/decorators/thread/ThreadDecorator.ts
@@ -1,12 +1,20 @@
 import { Expose } from 'class-transformer'
 import { IsInt, IsOptional, Matches } from 'class-validator'
 
-import { MessageIdRegExp } from '../../BaseDidCommMessage'
+// ^did:[a-z0-9]+:(?:%[0-9A-Fa-f]{2}|[A-Za-z0-9.-])+  # domain (no ':', allows %XX)
+// (?::(?:%[0-9A-Fa-f]{2}|[A-Za-z0-9._~!$&'()*+,;=:@/?-])*)*  # extra colon-separated segments
+// (?:\?(?:%[0-9A-Fa-f]{2}|[A-Za-z0-9._~!$&'()*+,;=:@/?-])*)?  # optional query
+// (?:#(?:%[0-9A-Fa-f]{2}|[A-Za-z0-9._~!$&'()*+,;=:@/?-])*)?   # optional fragment
+// $
+
+const PthidRegExp =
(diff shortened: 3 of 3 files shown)


packages/hedera/src/ledger/publisher/KmsPublisher.ts
```diff
@@ -0,0 +1,55 @@
+import { AgentContext, Kms, TypedArrayEncoder } from '@credo-ts/core'
+import { KeyManagementApi, KmsJwkPublicOkp } from '@credo-ts/core/src/modules/kms'
```
TimoGlastra: Should use `Kms.XX` instead of importing from `src`

AlexanderShenshin:
@TimoGlastra @genaris I've added Signer interface option to our SDK and updated Hedera module to use it instead of retrieving issuer key directly.
Please take a look when you have time.

I also see that E2E tests have failed again with error non-specific to Hedera and units started to fails as well with snapshot-related errors...
I guess it might be due to other recent changes on main, but please let me know if you see any related action items for this PR.

packages/hedera/src/ledger/publisher/KmsPublisher.ts
AlexanderShenshin: Done.

packages/hedera/CHANGELOG.md
```diff
@@ -0,0 +1,8 @@
+# Changelog
+
+## 0.5.13
+
+### Patch Changes
+
+- Added Hedera module 
+
```
TimoGlastra: Did you manually create this? The changelog will automatically be created when we make a new release, and be for 0.6.x. I think we can remove the file for now

packages/hedera/src/HederaModuleConfig.ts
```diff
@@ -0,0 +1,14 @@
+import { Cache } from '@hiero-did-sdk/core'
+import { HederaHcsServiceConfiguration } from '@hiero-did-sdk/hcs'
+
+export interface HederaModuleConfigOptions extends HederaHcsServiceConfiguration {
+  cache?: Cache
```
TimoGlastra, line 5: Do we want to allow integration with the Credo cache interface? (can be added later as well)

packages/hedera/src/dids/HederaDidRegistrar.ts
```diff
@@ -0,0 +1,182 @@
...
+    try {
+      const didRepository = agentContext.dependencyManager.resolve(DidRepository)
+      const ledgerService = agentContext.dependencyManager.resolve(HederaLedgerService)
+
+      const { did, didDocument, rootKey } = await ledgerService.createDid(agentContext, options)
+
+      const credoDidDocument = new DidDocument({
+        ...didDocument,
+        service: didDocument.service?.map((s) => new DidDocumentService(s)),
+      })
```
TimoGlastra: Usually we use DidDocument.fromJSON() if we transform from a json document, would that work here?

packages/hedera/src/dids/HederaDidRegistrar.ts
```diff
@@ -0,0 +1,182 @@
...
+          didDocumentMetadata: {},
+          didRegistrationMetadata: {},
+          didState: {
+            state: 'failed',
+            reason: 'Did not found',
+          },
+        }
+      }
+
+      const keys = this.concatKeys(didRecord.keys, options.secret?.keys)
```
TimoGlastra, line 86: Do we handle removal of keys? Do we keep them linked in the did record still?

packages/hedera/src/ledger/HederaLedgerService.ts
```diff
@@ -0,0 +1,534 @@
...
+  options?: {
+    network?: HederaNetwork | string
+  }
+  secret?: {
+    rootKeyId?: string
+    keys?: DidDocumentKey[]
+  }
+}
+
+export interface HederaCreateDIDResult extends CreateDIDResult {
```
TimoGlastra: For interfaces in Credo we usually use `Did` rather than `DID`

```suggestion
export interface HederaCreateDidResult extends CreateDIDResult {
```

packages/hedera/src/ledger/utils/index.ts
```diff
@@ -0,0 +1,40 @@
+import { Kms, TypedArrayEncoder } from '@credo-ts/core'
+
+export const getMultibasePublicKey = (publicJwk: Kms.KmsJwkPublicOkp & { crv: 'Ed25519' }): string => {
+  return `z${TypedArrayEncoder.toBase58(Uint8Array.from(TypedArrayEncoder.fromBase64(publicJwk.x)))}`
+}
+
+export const createOrGetKey = async (
```
TimoGlastra: Might be useful to use the PublicJwk instance here.

So you can do `Kms.PublickJwk.fromPublicJwk` and then you can also do `publicJwk.publicKey` and you get the public key as bytes (instead of having to base64 deocde the `x` value).

It should infer the type correctly if you pass `Ed25519` to createKey

```
    const { publicJwk } =await  kms.createKey({
      type: {
        crv: 'Ed25519',
        kty: 'OKP'
      }
    })

    const ed25519PublicJwk = Kms.PublicJwk.fromPublicJwk(publicJwk)

    // Public key bytes
    ed25519PublicJwk.publicKey.publicKey
```


You can then do `Kms.PublicJwk<>`

packages/hedera/src/ledger/publisher/KmsPublisher.ts
```diff
@@ -0,0 +1,53 @@
...
+    ).toPublicKey()
+  }
+
+  async setKeyId(keyId: string) {
+    const { publicJwk } = await createOrGetKey(this.kms, keyId)
+
+    this.keyId = keyId
+    this.submitPublicKey = KeysUtility.fromBytes(
+      Uint8Array.from(TypedArrayEncoder.fromBase64(publicJwk.x))
+    ).toPublicKey()
```
TimoGlastra: See my other comment in `createOrGetKey`

packages/hedera/src/ledger/signer/KmsSigner.ts
```diff
@@ -0,0 +1,51 @@
...
+import { KeysUtility, Signer } from '@hiero-did-sdk/core'
+import { createOrGetKey } from '../utils'
+
+export class KmsSigner extends Signer {
+  private keyId: string
+  private _publicKey: PublicKey
+
+  constructor(
+    private readonly kms: Kms.KeyManagementApi,
+    key: { keyId: string; publicJwk: Kms.KmsJwkPublicOkp & { crv: 'Ed25519' } }
```
TimoGlastra: You can also call the setter `.keyId = ` for the publicJwk, so you don't need the separate keyId

```suggestion
    publicJwk: Kms.PublicJwk<Kms.Ed25519PublicJwk>
```

packages/hedera/src/ledger/signer/KmsSigner.ts
```diff
@@ -0,0 +1,51 @@
...
+  ) {
+    super()
+
+    const { keyId, publicJwk } = key
+
+    this.keyId = keyId
+    this._publicKey = KeysUtility.fromBytes(Uint8Array.from(TypedArrayEncoder.fromBase64(publicJwk.x))).toPublicKey()
```
TimoGlastra: I would make an util `hederePublicKeyFromPublicJwk` and use that everwhere

pnpm-workspace.yaml
```diff
@@ -44,3 +44,5 @@ onlyBuiltDependencies:
   - "@hyperledger/anoncreds-nodejs"
   - "@hyperledger/indy-vdr-nodejs"
   - "@openwallet-foundation/askar-nodejs"
+  - "zstd-napi"
+  - "cbor-extract"
```
TimoGlastra, line 48: what are these libraries used for?

packages/hedera/CHANGELOG.md
AlexanderShenshin: I see, I didn't create it manually, but we may wrongly called changeset script in the fork.
Thanks for catching this, removed.

packages/hedera/src/HederaModuleConfig.ts
AlexanderShenshin: Yes, definitely. I noticed that it makes sense to do some adjustments around objects serialization in DID SDK before that, but I think we'll add this in near future.

packages/hedera/src/dids/HederaDidRegistrar.ts
AlexanderShenshin: Good catch, fixed.

packages/hedera/src/dids/HederaDidRegistrar.ts
AlexanderShenshin: Good point as well, I added removal of non-relevant keys after DID update operation.

packages/hedera/src/ledger/HederaLedgerService.ts
AlexanderShenshin: Done.

packages/hedera/src/ledger/utils/index.ts
AlexanderShenshin: Thanks for this, I have updated utils and related logic to use PublicJwk instance this way.

packages/hedera/src/ledger/publisher/KmsPublisher.ts
AlexanderShenshin: Done

packages/hedera/src/ledger/signer/KmsSigner.ts
AlexanderShenshin: Done

packages/hedera/src/ledger/signer/KmsSigner.ts
AlexanderShenshin: Done

pnpm-workspace.yaml
AlexanderShenshin: On Hedera, AnonCreds resources are stored according to [HCS-1 standard](https://hashgraphonline.com/docs/standards/hcs-1) that in our case makes use of zstd compression.
`zstd-napi` is a dependency for Node JS environment and also a dev dependency of Hiero DID SDK.

Just in case, for React Native environment we're using `react-native-zstd` package instead.
The SDK resolves either `zstd-napi` or `react-native-zstd` based on environment (using both as optional/dynamic dependencies).

Regarding `cbor-extract`, it's a transitive dependency of Hiero DID SDK JS. The SDK supports CBOR encoding for DID resolver result ("application/did+cbor") and uses [cbor-x](https://github.com/kriszyp/cbor-x) for encoding/decoding.

AlexanderShenshin:
@TimoGlastra Thanks for review and great suggestions!
I updated the PR to address the comments + added answers for a few questions that we have in threads here.

packages/hedera/src/dids/HederaDidRegistrar.ts
```diff
@@ -0,0 +1,188 @@
...
+  DidRepository,
+  DidUpdateResult,
+  JsonTransformer,
+} from '@credo-ts/core'
+import {
+  HederaDidCreateOptions,
+  HederaDidDeactivateOptions,
+  HederaDidUpdateOptions,
+  HederaLedgerService,
+} from '../ledger/HederaLedgerService'
```
genaris: I think it could be useful to create a ../ledger/index.ts where you expose all these types and then import them from here from '../ledger'). 

And also expose it to upper levels by exporting * from ./ledger in ../../index.ts. This will make it easier/cleaner to use from apps like in the example you have added to the docs (see https://github.com/openwallet-foundation/credo-ts-docs/blob/1833dbff7064f5776bcd2006ab77021548a2dc3d/snippets/current/src/set-up-hedera.ts#L19).

genaris (approved):
LGTM! Demo works great.

I've left just a minor suggestion to exports.

packages/hedera/src/dids/HederaDidRegistrar.ts
AlexanderShenshin: Good point, thanks. I updated exports as suggested + updated documentation PR as well.

AlexanderShenshin:
@genaris Thanks for review!
I've pushed exports changes and merged recent commits from 'main'.

One of tests jobs failed due to seemingly unrelated reason though + validation job was cancelled because of that.

PR #2408, opened 2025-09-12
refactor(didcomm): filter by transport session id in message pickup

genaris:
Some improvements after feedback on #2399


jamshale (approved):
That makes sense after I see it👍

PR #2409, opened 2025-09-17
chore: update askar library to 0.4.0, enabling 16KB page size support for Android

TimoGlastra:


swcurran:
@genaris -- Am I right that this solves the "16kb" issue for 0.6.x, but not for 0.5.x.  Does the 0.3.3 release that @andrewwhitehead did yesterday enable getting it supported in 0.5.x?

PR #2413, opened 2025-09-19
feat: support new OWF askar (0.5.x)

TimoGlastra:
Supports the new OWF Askar release in addition to the hyperledger relase for the 0.5.x release stream. This allows wallets on 0.5.x to still update to use 16kb page size.

New releases will also have to be created for Anoncreds and Indy but these should be simpler to update since the package name hasn't changed and there haven't been API changes i think? 

cc @swcurran @genaris

Can someone else pick up the updates for AnonCreds/Indy if they are needed?


TimoGlastra:
After this is released, wallets that want to use the new OWF askar should:
- add `@openwallet-foundation/askar-react-native` to their project
- provide `askar` imported from `@openwallet-foundation/askar-react-native` as value for `ariesAskar` in askar module

genaris (approved):
Great trick!

PR #2406, opened 2025-09-11
refactor: merge oid4vc module

TimoGlastra:

PR #2415, opened 2025-09-19
feat: merge default didcomm modules 

TimoGlastra:
Similar to #2406 but for didcomm. It adds all default didcomm modules to the didcomm mdoule, and enables them by default. You can disable these modules by providing false in the config. The agent type is dynamically inferred based on it. You can now also access these modules on `agent.didcomm`. 

I have not added support for registering extension modules under the didcomm module, and them also being available under agent.didcomm. But after thinking about i tmore, i think it makes sense to just register these on the normal agent modules.

Since these are custom extension modules. What do you think? it will also keep the code simpler, since we won't have to create basically another dynamic module registration layer (like we already have). But we do have the benefit that all default didcomm modules are now registered by default (easier to set up, harder to make mistakes).

Also moved message handler registartion to the module level instead of services/api


TimoGlastra:
@genaris @berendsliedrecht this is now ready for review!

PR #2417, opened 2025-09-22
fix(openid4vc): credential response credential structure

hacdias:


packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
```diff
@@ -1070,7 +1070,9 @@ export class OpenId4VciHolderService {
     this.logger.debug('Credential response', credentialResponse)
 
     const credentials =
-      credentialResponse.credentials ??
+      credentialResponse.credentials?.map((c) =>
+        typeof c === 'object' && Object.keys(c).length === 1 && 'credential' in c ? c.credential : c
```
TimoGlastra: I think the reason for making it an object is so that extension can be added for each credential. Checking for a single key break that.

How about checking that every item in the list is an object and has the `credential` key, and then we use `.credential`, otherwise just `c`

hacdias: Done

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerService.ts
```diff
@@ -434,7 +434,12 @@ export class OpenId4VcIssuerService {
 
       credentialResponse = vcIssuer.createCredentialResponse({
         credential: credentialRequest.proof ? credentials.credentials[0] : undefined,
-        credentials: credentialRequest.proofs ? credentials.credentials : undefined,
+        credentials: credentialRequest.proofs
```
TimoGlastra, line 438: issuerMetadata is retrieved from `getIssuer`, which only takes the issuer and the context, so it won't really have the context of the specific issuance session i think?

for verification we store the version on the record. Which we might need to do here as well if we can't infer it based on the offer or request

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerServiceOptions.ts
```diff
@@ -100,12 +100,9 @@ interface OpenId4VciCreateCredentialOfferOptionsBase {
   baseUri?: string
 
   /**
-   * @default v1.draft11-15
-   *
-   * NOTE: `v1.draft15` credential is compatible with draft 13 credential offer as well. Only the issuer metadata
-   * is different, so ensure you configure the issuer metadata in a compatible way based on the provided draft version.
+   * @default v1.draft11-14
```
TimoGlastra: I think we should make the default draft 15 (latest)

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
hacdias: Done!

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerService.ts
hacdias: Or we can just decide to use the latest format, since we don't support deferred issuance for Draft < 15

TimoGlastra: what about non-deferred?

TimoGlastra: this is the code flow for non-deferred issuance right?

hacdias: Done

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerService.ts
```diff
@@ -514,7 +519,7 @@ export class OpenId4VcIssuerService {
       })
 
       deferredCredentialResponse = vcIssuer.createDeferredCredentialResponse({
-        credentials: credentials.credentials,
+        credentials: credentials.credentials.map((c) => ({ credential: c })),
```
hacdias, line 522: ℹ️ For deferred issuance, we only supported Draft 15, so we just use the latest format, as this is a fix.

TimoGlastra (approved):

PR #2418, opened 2025-09-23
feat(core): improve the W3cJsonCredential and W3cV2JsonCredential types

hacdias:
Makes the aforementioned types more specific.

### DIFF
--- a/packages/core/src/modules/vc/jwt-vc/credentialTransformer.ts
+++ b/packages/core/src/modules/vc/jwt-vc/credentialTransformer.ts
@@ -1,5 +1,4 @@
 import type { JwtPayloadOptions } from '../../../crypto/jose/jwt'
-import type { W3cJsonCredential } from '../models/credential/W3cJsonCredential'
 
 import { isObject } from 'class-validator'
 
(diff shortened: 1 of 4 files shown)


packages/core/src/modules/vc/models/credential/W3cJsonCredential.ts
```diff
@@ -1,12 +1,22 @@
 import type { JsonObject, SingleOrArray } from '../../../../types'
 
+interface W3cJsonIssuer {
+  id: string
```
hacdias, line 4: ℹ️ This was initially marked as optional. However, the `id` is required for the issuer.

packages/core/src/modules/vc/jwt-vc/credentialTransformer.ts
```diff
@@ -11,7 +10,7 @@ import { W3cCredential } from '../models/credential/W3cCredential'
 import { w3cDate } from '../util'
 
 export function getJwtPayloadFromCredential(credential: W3cCredential) {
-  const vc = JsonTransformer.toJSON(credential) as Partial<W3cJsonCredential>
+  const vc = JsonTransformer.toJSON(credential)
```
hacdias, line 13: ℹ️ Removed the type to make sure that the `delete` wasn't complaining about deleting required types in nested properties. If we had a `PartialDeep` we could use it instead.

PR #2419, opened 2025-09-24
feat(webvh): implement did registrar and crypto signer; enhance anonCreds register methods

lotharking:


swcurran:
@lotharking -- FYI -- we have a did:webvh DIF Work Item meeting Thursday Sept. 25 at 9:00 Pacific / 18:00 Central Europe if you want to join and talk about this work.  We'd love to hear from you.  Agenda and zoom link are here: https://hackmd.io/k4cIK9vQSlaeg2pdHE51IQ?view#Meeting---25-Sept-2025

packages/webvh/src/dids/__tests__/WebVhDidRegistrar.integration.test.ts
```diff
@@ -0,0 +1,111 @@
...
+describe('WebVhDidRegistrar Integration Tests', () => {
+  let registar: WebVhDidRegistrar
+  let agentContext: ReturnType<typeof getAgentContext>
+
+  beforeEach(async () => {
+    // Reset only the storage (DidRepository) mocks before each test
+    jest.clearAllMocks()
+
+    // Create a real resolver instance
+    registar = new WebVhDidRegistrar()
```
genaris:
```suggestion
    registrar = new WebVhDidRegistrar()
```

packages/webvh/src/dids/__tests__/WebVhDidRegistrar.test.ts
genaris, line 1: I don't see this as an 'integration' test. Usually we name these kind of tests simply after the class we are testing, i.e. `WebVdhDidRegistrar.test.ts`.

packages/webvh/src/dids/__tests__/WebVhDidRegistrar.test.ts
```diff
@@ -0,0 +1,111 @@
...
+  describe('DID WebVH update', () => {
+    const did = mockResolvedDidDocument.id
+    const didDocument = new DidDocument(mockResolvedDidDocument as unknown as DidDocument)
+
+    it('should fail if DID record already exists', async () => {
+      mockDidsRepository.getSingleByQuery.mockResolvedValue(undefined)
+
+      const result = await registar.update(agentContext, { did, didDocument })
+
+      expect(result).toEqual({
```
genaris, line 162: Only a failure use case here?

What about checking something in the log when a succesful update is done?

packages/webvh/src/cryptosuites/types.ts
```diff
@@ -17,3 +17,7 @@ export interface SecuredDocument {
   [key: string]: string | object
   proof: Proof
 }
+
+export interface unsecuredDocument {
```
genaris:
```suggestion
export interface UnsecuredDocument {
```

Actually I don't see `SecuredDocument` used anywhere. Maybe we should clean up some things here... (or at least make SecuredDocument extend UnsecuredDocument, just adding the proof)

packages/webvh/src/dids/WebVhDidRegistrar.ts
```diff
@@ -0,0 +1,205 @@
...
+          {
+            controller: baseDid,
+            type: 'Ed25519VerificationKey2020',
+            publicKeyMultibase,
+          },
+        ],
+        verifier,
+      })
+
+      // Save didRegistry
```
genaris: did record (or just remove the comment, as the code is self-explanatory I think)

packages/webvh/src/dids/WebVhDidRegistrar.ts
```diff
@@ -0,0 +1,205 @@
...
+        verificationMethod: inputVerificationMethod,
+      } = inputDidDocument
+      const verificationMethods =
+        inputVerificationMethod ?? (log[log.length - 1].state.verificationMethod as VerificationMethod[])
+      const { updateKeys } = log[log.length - 1].parameters
+      if (!verificationMethods?.length || !verificationMethods[0].publicKeyMultibase)
+        return this.handleError('At least one verification method must be provided.')
+
+      // Get signer/verifier
+      const signer = new WebvhDidCryptoSigner(agentContext, verificationMethods[0].publicKeyMultibase)
```
genaris: I'm a bit confused about the verificationmethod used to sign the new log entry: here we are always using the first verification method of either the provided input DID Document or the previous state. Is that correct? What if I provide a DID Document that does not contain the initial signer verification method or it is not the first one in the array?

packages/webvh/README.md
```diff
@@ -0,0 +1,126 @@
...
+    }),
+  }
+})
+
+await agent.initialize()
+```
+
+#### Creating and Updating a `did:webvh` DID
+
+You can create and update a `did:webvh` DID using the agent's DID API:
```
genaris: I think in this section you can describe where users can locate and how are we supposed to publish the artifacts related to the DID, namely the did.jsonl. It is not evident that they will need to get it from the metadata with the key called 'log'.

Also I think it can be useful to expose in Credo (and mention in the docs) some utility methods we have in VS Agent for retrieving the did:web version of the did:webvh.

packages/webvh/README.md
```diff
@@ -0,0 +1,126 @@
...
+      version: options.version,
+      issuerId,
+    },
+    options: {},
+  })
+```
+
+#### Resolving Resources
+
+To resolve a resource associated with a `did:webvh` DID:
```
genaris: The problem with this example is that it is too specific to the implementation we have in VS Agent: it's a NestJS endpoint and the attested resource is taken from a generic record, which is something we don't want to enforce to every one.

I think we must explain somewhere that we are supposed to serve the resulting attestedResource in a `/resources/:resourceId` GET endpoint and that this example is storing them in generic records just for the sake of simplicity.

But we must also add an example of registration where we take the output (registrationMetadata) and store it in generic records to be consistent and make the reader fully understand the idea.

packages/webvh/README.md
```diff
@@ -0,0 +1,126 @@
...
+
+const result = await agent.dids.update({ did: publicDid, didDocument })
+```
+
+#### Registering AnonCreds Objects
+
+Register a schema using the AnonCreds module:
+
+```typescript
+const { schemaState, registrationMetadata: schemaMetadata } =
```
genaris, line 138: It's very important to make it explicit that the `schemaMetadata` will contain the attested resource we are supposed to host under `/resources/resourceId`.

Also we should clarify that, at the moment, we put everything related to AnonCreds under the subpath `resources`, which is on the examples but not strictly defined to that on the did:web anoncreds method specification.

packages/webvh/src/dids/WebVhDidRegistrar.ts
```diff
@@ -0,0 +1,205 @@
...
+import type { DIDLog } from 'didwebvh-ts'
+
+import { DidRepository, KeyType, DidDocument, DidRecord, DidDocumentRole, JsonTransformer } from '@credo-ts/core'
+import { createDID, multibaseEncode, MultibaseEncoding, updateDID } from 'didwebvh-ts'
+
+import { WebvhDidCrypto } from './WebvhDidCrypto'
+import { WebvhDidCryptoSigner } from './WebvhDidCryptoSigner'
+
+interface WebVhDidCreateOptions extends DidCreateOptions {
+  domain: string
```
genaris, line 20: Something that @TimoGlastra pointed out are paths on did:webvh creation. For instance, it is possible to have a did:webvh:SCID:domain.com:path1 whose DID Log will be resolvable at https://domain.com/path/did.jsonl.

At a first sight, I think it is possible to achieve this by specifying `domain: 'domain.com:path'` on DID create method. But maybe it would be a more robust solution if we add the path as an optional parameter that can be expressed as `path/subpath`. Supporting specific **ports** would be also interesting.

In any case I think it would be good to consider this use case in the tests, as it is commonly used.

packages/webvh/src/cryptosuites/types.ts
lotharking: SecuredDocument is used. That’s why I prefer to use inheritance, so it simply extends UnsecuredDocument by adding the proof field

packages/webvh/src/dids/WebVhDidRegistrar.ts
lotharking: For a better approach, we updated the implementation so that it no longer assumes the first verification method in the array. Instead, it now searches for the first verification method that has a publicKeyMultibase available. This ensures that if the initial signer verification method is not the first entry in the DID Document (or if some methods don’t provide a usable key), the correct verification method will still be selected.

packages/webvh/README.md
lotharking: Applied the recommendations: clarified how artifacts (did.jsonl) are persisted in record metadata under the log key, added details about updating metadata on DID updates, and included an example endpoint for retrieving the log.

packages/webvh/README.md
lotharking: Updated the doc to state that serving resources via /resources/:resourceId is required, and added an example for persisting registrationMetadata in generic records for consistency.

packages/webvh/README.md
lotharking: Implemented

packages/webvh/src/dids/WebVhDidRegistrar.ts
lotharking: You're right, treating the path as a separate optional parameter makes the API clearer and avoids overloading the domain field. We’ve added support for path/subpath and included tests to cover this use case. 
About the ports, I’m not entirely sure how they should behave in practice. Conceptually, passing them as part of the domain (e.g. domain.com:8080) should work without issues.

packages/webvh/README.md
```diff
@@ -0,0 +1,212 @@
...
+The endpoint retrieves the artifact from the DID record metadata and serves it to the client:
+```typescript
+@Get('/.well-known/did.jsonl')
+async getDidLog(@Res() res: Response) {
+  const agent = await this.agentService.getAgent()
+  const [didRecord] = await agent.dids.getCreatedDids({ did: agent.did })
+  const didLog = didRecord.metadata.get('log') as DIDLog[] | null
+
+  if (didLog) {
+    res.setHeader('Content-Type', 'application/jsonl; charset=utf-8')
```
genaris: Note that did:webvh specification states that the content type should be text/jsonl

packages/webvh/README.md
```diff
@@ -0,0 +1,212 @@
...
+      version: options.version,
+      issuerId,
+    },
+    options: {},
+  })
+```
+
+#### Returned Metadata
+
+The `registerSchema` method returns three key components:
```
genaris: You can generalize this section (i.e. "Object registration methods returns three key components" and use object in the naming e.g. "objectState", "objectMetadata").

packages/webvh/src/dids/WebVhDidRegistrar.ts
```diff
@@ -0,0 +1,207 @@
...
+        domain,
+        method: 'webvh',
+      })
+      if (record) return this.handleError(`A record with domain "${domain}" already exists.`)
+
+      // Create crypto instance
+      const publicKeyMultibase = await this.generatePublicKey(agentContext)
+      const signer = new WebvhDidCryptoSigner(agentContext, publicKeyMultibase)
+      const verifier = new WebvhDidCrypto(agentContext)
+      const baseDid = `did:webvh:{SCID}:${domain}`
```
genaris, line 64: Shouldn't this baseDid include the paths, if any?

packages/webvh/src/dids/WebVhDidRegistrar.ts
```diff
@@ -0,0 +1,207 @@
...
+      })
+
+      const didDocument = JsonTransformer.fromJSON(doc, DidDocument)
+      const didRecord = new DidRecord({
+        did,
+        didDocument,
+        role: DidDocumentRole.Created,
+      })
+      didRecord.metadata.set('log', log)
+      didRecord.setTags({ domain })
```
genaris: Shall paths also added to the tags, so we can retrieve the DID unambigously?

packages/webvh/src/dids/WebVhDidRegistrar.ts
```diff
@@ -0,0 +1,207 @@
...
+      const didRepository = agentContext.dependencyManager.resolve(DidRepository)
+      const didRecord = await didRepository.findSingleByQuery(agentContext, {
+        role: DidDocumentRole.Created,
+        did,
+        method: 'webvh',
+      })
+      if (!didRecord) return this.handleError('DID not found')
+
+      const log = didRecord.metadata.get('log') as DIDLog
+      const domain = didRecord.getTag('domain') as string
```
genaris, line 125: Make sure we support paths in update case as well.

genaris (changes requested):
In general LGTM! Just some minor comments regarding docs and the paths feature.

Also I´d like to leverage this PR to update the line in this repo README that says:

>- 🛂 **Extendable [DID](https://www.w3.org/TR/did-core/) resolver and registrar** - out of the box support for `did:web`, `did:key`, `did:jwk`, `did:peer`, `did:sov`, `did:indy` and `did:cheqd`.

To add also did:webvh, since now we fully support it.

packages/webvh/src/dids/WebVhDidRegistrar.ts
```diff
@@ -0,0 +1,207 @@
...
+
+import { DidRepository, KeyType, DidDocument, DidRecord, DidDocumentRole, JsonTransformer } from '@credo-ts/core'
+import { type DIDLog, createDID, multibaseEncode, MultibaseEncoding, updateDID } from 'didwebvh-ts'
+
+import { WebvhDidCrypto } from './WebvhDidCrypto'
+import { WebvhDidCryptoSigner } from './WebvhDidCryptoSigner'
+
+interface WebVhDidCreateOptions extends DidCreateOptions {
+  domain: string
+  paths?: string[]
```
genaris: I know using an array is handy because didwebvh-ts uses it like that, but don't you think it would be better for us to consider the path as a single string that contains as many as subpaths as needed (e.g. path: 'users/alice' ? This way there will be a single path tag to add to the record as well

packages/webvh/src/dids/WebVhDidRegistrar.ts
lotharking: No. The createDID method is responsible for handling this part and appending the correct paths if any are required.

packages/webvh/src/dids/WebVhDidRegistrar.ts
lotharking: I’m not sure about that. If we start checking the path, then we might allow multiple creations with different subpaths, and I’m not certain that this is the intended design.

packages/webvh/src/dids/WebVhDidRegistrar.ts
lotharking: By default, the update method does not handle any path updates.

packages/webvh/src/dids/WebVhDidRegistrar.ts
genaris: Why an agent would not be allowed to create did:webvh:domain:users:alice and did:webvh:domain:users:bob?

packages/webvh/src/dids/WebVhDidRegistrar.ts
genaris: How would you distinguish did:webvh:domain:alice from did:webvh:domain:bob if you only get domain as an input? 

Edit: Oh I see: the did parameter here includes the whole string and the library method doesn't use path at all (as per your comment above). Nevermind!

packages/webvh/src/dids/WebVhDidRegistrar.ts
lotharking: You are right, it has been implemented.

lotharking: You are right, this has been implemented. On the Creation method, we build the domain key with the path, and in the update method, we use the retrieved domain complete

PatStLouis:
if you have a way to create a did using the `did:webvh:scid:domain:users:alice` pattern (2 sub paths), you will be able to use the did webvh server to upload/publish your did/witness file and have it publicly resolvable. There's a sandbox test instance deployed at bc gov which we use for integration testing in the acapy plugin. Let me know if you want to try it out!

PR #2432, opened 2025-10-02
Feat/update pthid regex

Ellenn-A:
As discussed in [issue](https://github.com/openwallet-foundation/credo-ts/issues/2431)

This PR introduces a new `PthidRegexExp` which is now used against `threadId` and `parentThreadId` in ThreadDecorator. Unlike `MessageIdRegexExp` (from [RFC 0008](https://github.com/decentralized-identity/aries-rfcs/blob/main/concepts/0008-message-id-and-threading/README.md#message-id-requirements)), which enforces strict message ID formatting, `PthidRegexExp` is more permissive. It allows both traditional message IDs and valid DID URLs (per [RFC 0023](https://github.com/decentralized-identity/aries-rfcs/blob/main/features/0023-did-comm/README.md)), ensuring compatibility with DIDComm v1 flows where pthid/thid aren’t explicitly bound to the same format as message IDs.


genaris (approved):
Thanks @Ellenn-A . It would be great if you could also create a PR targetting main branch, so this fix is also present in 0.6.x and future versions.

packages/core/src/decorators/thread/ThreadDecorator.test.ts
```diff
@@ -45,4 +47,88 @@ describe('Decorators | ThreadDecorator', () => {
...
+        '12345678901234567890123456789012345678901234567890123456789012345', // too long
+        'invalid!@#$%^&*()format', // invalid characters
+        'did:example:incomplete%ZZencoding', // invalid percent-encoding
+      ]
+
+      for (const invalidValue of invalidValues) {
+        const decorator = new ThreadDecorator({
+          parentThreadId: invalidValue,
+        })
+        console.log('Testing invalid value:', invalidValue)
```
genaris:
```suggestion
```

We try to avoid using console logging to make tests less verbose

PR #2437, opened 2025-10-06
feat: Add Hedera module (backport of #2405)

AlexanderShenshin:
This PR is a 0.5.x backport for #2405.
Implementation differences are based on switching from new KMS API to 0.5.x Wallet API.

This PR adds Hedera module, enabling support for [Hedera DID Method](https://github.com/hashgraph/did-method/blob/master/hedera-did-method-specification.md) and AnonCreds registry (see [Hedera/Hiero AnonCreds Method](https://dsrcorporation.github.io/hedera-anoncreds-method/)).


AlexanderShenshin:
@TimoGlastra, @genaris Please take a look when you have time (looks like there is no explicit review request since the target is `0.5.x` branch).

PR #2440, opened 2025-10-06
fix(webvh): Allow updating VerificationMethods without requiring publicKeyMultibase

lotharking:
This PR modifies the handling of VerificationMethod objects in the webvh-ts library to allow updates even when publicKeyMultibase is not present.

By default, the library enforces the existence of a publicKeyMultibase, which limits support to multikey-only formats. In Credo, we use a broader format that may include different types of keys, not only multikey.

Update:
The didwebvh-ts library has been updated to version 2.5.5, and this implementation has been adjusted accordingly to align with the new release.


packages/webvh/src/dids/WebVhDidRegistrar.ts
```diff
@@ -153,7 +153,7 @@ export class WebVhDidRegistrar implements DidRegistrar {
         ...inputDidDocument,
         verificationMethods: verificationMethods.map((vm) => ({
           ...vm,
-          publicKeyMultibase: vm.publicKeyMultibase ?? '',
+          publicKeyMultibase: vm.publicKeyMultibase as string,
```
genaris: I think it is OK as a workaround but the actual fix for this would be to update didwebvh-ts to accept both `publicKeyJwk` and `publicKeyMultibase`, as stated in https://www.w3.org/TR/did-1.0/#verification-method-properties.

@brianorwhatever what do you think?

swcurran: @PatStLouis -- perhaps you can jump in here as well.

PatStLouis: The verification method is tied to the cryptosuite used on the proof. Currently, `eddsa-jcs-2022` is a normative requirement for webvh. The eddsa spec refers to the cid `retrieve-verification-method` algorithm, which supports any valid key representation. However, Multikey is recommended. I think its reasonable to allow `publicKeyJwk` when resolving, while being aware it might have an interoperability impact if used as a controller. Controllers should use Multikey representation when adding a eddsa-jcs-2022 data integrity proof to a document.

https://www.w3.org/TR/vc-di-eddsa/#multikey
https://www.w3.org/TR/cid-1.0/#retrieve-verification-method

swcurran: I believe you mean to say "`eddsa-jcs-2022` is a normative requirement for webvh **AnonCreds Resources**". It is not a requirement for a `verificationMethod` in an arbitrary did:webvh DIDDoc per se. However, we have said (for good or bad) that the proof in a did:webvh AnonCreds attested resource **MUST** be a `eddsa-jcs-2022`, hence the requirement on the `verificationMethod` that you are referencing.

I just want it to be clear that did:webvh does not put that requirement in place.  I'm not sure that helps with the question at hand, but I did want to clarify that.

genaris: Yeah, here we are not talking here about AnonCreds or even did:webvh's DID Log signing, but about updating verificationMethods of the  DID Document (this can include keys used for DIDComm or other arbitrary purposes). In that sense I think the library should accept any valid data as long as it is conformant to DID Core.

swcurran: If that is the context, then I agree. A key feature of did:webvh is that it does not constrain the DID Controller -- it can put anything it wants in the DIDDoc.

brianorwhatever: yes agreed, the VerificationMethod `publicKeyMultibase` should become optional here https://github.com/decentralized-identity/didwebvh-ts/blob/4db589b98a84563326b7a83320af3036f10a1c59/src/interfaces.ts#L66C1-L67C1

genaris: @brianorwhatever I've just create a PR for this: https://github.com/decentralized-identity/didwebvh-ts/pull/75

lotharking:
Hey @genaris,
Thanks for updating the library! The change has been applied in the PR.

PR #2441, opened 2025-10-07
feat: ESM

TimoGlastra:


TimoGlastra:
I think I may found some of the reasons why we have the Jest process crashing, it also happens with vitest, and it seems to happen with OpenID4VC tests, which use `nock`, which in turn uses `msw`, and this may have a memory leak: 
https://github.com/mswjs/msw/issues/2537

If we keep seeing the unexpected crash we can try downgrade to nock 14.0.1 that is using an older version of msw.

Another relevant issue: https://github.com/vitest-dev/vitest/issues/8201

TimoGlastra:
Superseded by #2444

PR #2447, opened 2025-10-10
fix: base64-encode `x5c` JWS header

Magnus-Kuhn:


TimoGlastra:
Can you add a changeset with a fix? I can't do that on org forks (see the link in the changeset bot message)

Magnus-Kuhn:
good - and this won't break something important that expects base64url?

TimoGlastra:
> good - and this won't break something important that expects base64url?

I hope not, but in that case that other implementation will also be wrong for expecting base64url. The credo base64 decoder supports both base64 and base64url in the same method, so it shouldn't cause issues.

PR #2453, opened 2025-10-13
fix: check supported key types dpop

TimoGlastra:


TimoGlastra:
ran into an issue where the issuer supports RSA, and the new KMS API supports RSA, but askar does not. This way we ensure we can actually use the key type

PR #2456, opened 2025-10-15
feat: Add support for JWT VC Issuer Metadata

GianfrancoMS:
Add support for JWT VC Issuer Metadata
https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-03.html#name-jwt-vc-issuer-metadata

### DIFF
--- a/packages/core/src/modules/kms/KeyManagementApi.ts
+++ b/packages/core/src/modules/kms/KeyManagementApi.ts
@@ -149,9 +149,9 @@ export class KeyManagementApi {
 
     const operation = { operation: 'verify', algorithm: options.algorithm } as const
     const kms =
-      backend || typeof options.key !== 'string'
+      backend || typeof options.key.keyId !== 'string'
         ? this.getKms(this.agentContext, backend, operation)
-        : (await this.getKmsForOperationAndKeyId(this.agentContext, options.key, operation)).kms
+        : (await this.getKmsForOperationAndKeyId(this.agentContext, options.key.keyId, operation)).kms
 
     return await kms.verify(this.agentContext, kmsOptions)
   }
@@ -172,9 +172,9 @@ export class KeyManagementApi {
       keyAgreement: options.key.keyAgreement,
     } as const
     const kms =
-      backend || typeof options.key !== 'string'
+      backend || typeof options.key.keyId !== 'string'
         ? this.getKms(this.agentContext, backend, operation)
-        : (await this.getKmsForOperationAndKeyId(this.agentContext, options.key, operation)).kms
+        : (await this.getKmsForOperationAndKeyId(this.agentContext, options.key.keyId, operation)).kms
 
     return await kms.encrypt(this.agentContext, kmsOptions)
   }
@@ -195,14 +195,14 @@ export class KeyManagementApi {
       keyAgreement: options.key.keyAgreement,
     } as const
     const kms =
-      backend || typeof options.key !== 'string'
+      backend || typeof options.key.keyId !== 'string'
         ? this.getKms(
             this.agentContext,
 
             backend,
             operation
           )
-        : (await this.getKmsForOperationAndKeyId(this.agentContext, options.key, operation)).kms
+        : (await this.getKmsForOperationAndKeyId(this.agentContext, options.key.keyId, operation)).kms
 
     return await kms.decrypt(this.agentContext, kmsOptions)
   }
--- a/packages/core/src/modules/sd-jwt-vc/SdJwtVcOptions.ts
+++ b/packages/core/src/modules/sd-jwt-vc/SdJwtVcOptions.ts
@@ -53,12 +53,25 @@ export interface SdJwtVcIssuerX5c {
   issuer: string
 }
 
+export interface SdJwtVcIssuerMetadata {
+  method: 'jwt-vc-issuer-metadata'
+
+  /**
+   * The JWK that should be present at GET /.well-known/jwt-vc-issuer or at the issuer's jwks_uri
+   */
+  jwk: PublicJwk
+
+  /**
+   * The issuer of the JWT. It should be a HTTPS URI.
+   * Holders and verifiers should be able to call GET /.well-known/jwt-vc-issuer from this issuer
+   */
+  issuer: string
+}
+
 // We support jwk and did based binding for the holder at the moment
 export type SdJwtVcHolderBinding = SdJwtVcHolderDidBinding | SdJwtVcHolderJwkBinding
 
-// We only support did based issuance currently, but we might want to add support
-// for x509 or issuer metadata (as defined in SD-JWT VC) in the future
-export type SdJwtVcIssuer = SdJwtVcIssuerDid | SdJwtVcIssuerX5c
+export type SdJwtVcIssuer = SdJwtVcIssuerDid | SdJwtVcIssuerX5c | SdJwtVcIssuerMetadata
 
 export interface SdJwtVcSignOptions<Payload extends SdJwtVcPayload = SdJwtVcPayload> {
   payload: Payload
--- a/packages/core/src/modules/sd-jwt-vc/SdJwtVcService.ts
+++ b/packages/core/src/modules/sd-jwt-vc/SdJwtVcService.ts
@@ -37,6 +37,7 @@ import {
   parseHolderBindingFromCredential,
   resolveDidUrl,
   resolveSigningPublicJwkFromDidUrl,
+  resolveSigningPublicJwkFromJwtVcIssuerMetadata,
 } from './utils'
 
 type SdJwtVcConfig = SDJwtVcInstance['userConfig']
@@ -231,6 +232,15 @@ export class SdJwtVcService {
       throw new SdJwtVcError(
         `The 'iss' claim in the payload does not match a 'SAN-URI' name and the domain extracted from the HTTPS URI does not match a 'SAN-DNS' name in the x5c certificate.`
       )
+    }
+  }
+
+  private assertValidJwtVcIssuer(agentContext: AgentContext, iss: string, publicJwk: PublicJwk) {
+    if (!iss.startsWith('https://') && !(iss.startsWith('http://') && agentContext.config.allowInsecureHttpUrls)) {
+      throw new SdJwtVcError('The issuer must be a HTTPS URI.')
+    }
+    if (!publicJwk.hasKeyId) {
+      throw new SdJwtVcError(`No kid found for key ${publicJwk.jwkTypehumanDescription}`)
     }
   }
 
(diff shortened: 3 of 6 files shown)


packages/core/src/modules/sd-jwt-vc/SdJwtVcService.ts
```diff
@@ -580,7 +600,32 @@ export class SdJwtVcService {
...
+      if (result.issuer !== iss) {
+        throw new SdJwtVcError(
+          `Failed to resolve signing public JWK from issuer metadata. Expected issuer: ${iss}, but got: ${result.issuer}`
+        )
+      }
+      return {
+        method: 'jwk',
+        jwk: result.jwk,
+        issuer: result.issuer,
+      }
```
TimoGlastra, line 627: I'm thinking whether we want to call the method something else than `jwk`. Since with JWK we really only verify the binding that the credential is issued by a JWK, but with the jwt vc issuer metadata, you also verify the domain of the issuer where the issuer metadata is hosted.

Should we thus call this maybe `well-known-issuer-metadata` or `well-known-issuer`? I'm open to other names, if we can come up with something simple that does capture the origin of the key (like with x5c and did)

packages/core/src/modules/sd-jwt-vc/SdJwtVcOptions.ts
```diff
@@ -53,12 +53,27 @@ export interface SdJwtVcIssuerX5c {
   issuer: string
 }
 
+export interface SdJwtVcIssuerJwk {
+  method: 'jwk'
+
+  /**
+   * The JWK that should be presente at GET /.well-known/jwt-vc-issuer or at the issuer's JWKs URI
```
TimoGlastra: did you mean present or presented?

```suggestion
   * The JWK that should be present at GET /.well-known/jwt-vc-issuer or at the issuer's JWKs URI
```

packages/core/src/modules/sd-jwt-vc/SdJwtVcOptions.ts
```diff
@@ -53,12 +53,27 @@ export interface SdJwtVcIssuerX5c {
...
+   */
+  jwk: PublicJwk
+
+  /**
+   * The issuer of the JWT. It should be a HTTPS URI.
+   * Holde rand verifiers should be able to call GET /.well-known/jwt-vc-issuer from this issuer
```
TimoGlastra:
```suggestion
   * Holder and verifiers should be able to call GET /.well-known/jwt-vc-issuer from this issuer
```

packages/core/src/modules/kms/KeyManagementApi.ts
```diff
@@ -149,9 +149,9 @@ export class KeyManagementApi {
 
     const operation = { operation: 'verify', algorithm: options.algorithm } as const
     const kms =
-      backend || typeof options.key !== 'string'
+      backend || typeof options.key.keyId !== 'string'
```
TimoGlastra, line 152: Ah thanks for cathing this 👍

packages/core/src/modules/sd-jwt-vc/typeMetadata.ts
```diff
@@ -57,3 +60,23 @@ export interface SdJwtVcTypeMetadata {
...
+    issuer: z.string().url(),
+    jwks_uri: z.string().url(),
+    jwks: z.never().optional(),
+  }),
+  z.object({
+    issuer: z.string().url(),
+    jwks_uri: z.never().optional(),
+    jwks: JwksSchema,
+  }),
+])
```
TimoGlastra, line 79: It probably makes sense to also move to zod gradually for other modules, but are these validation schemas also used somewhere? Or are they just used for types? I think it would be good to use this in `resolveSigningPublicJwkFromJwtVcIssuerMetadata`. 

You can use the `zParseWithErrorHandling` util method on the returned JSON payload.

packages/core/src/modules/sd-jwt-vc/utils.ts
```diff
@@ -23,6 +24,44 @@ export async function resolveDidUrl(agentContext: AgentContext, didUrl: string)
...
+  if (metadata?.jwks_uri) {
+    const jwks = await fetch<Jwks>(agentContext, metadata.jwks_uri)
+    if (jwks) {
+      const jwk = jwks.keys.find((key) => key.kid === kid)
+      if (jwk) {
+        return { jwk: PublicJwk.fromUnknown(jwk), issuer: metadata.issuer }
+      }
+    }
+  }
+  return null
```
TimoGlastra: Since either `jwks` or `jwks_uri` MUST be defined (Which we can check with the zod schema), I think we can simplify this a bit. I always try to return early when possible to avoid a lot of nesting. It's probably down to personal prefence, as it doe add more `return null` lines

```suggestion
  const metadata = await fetch<SdJwtVcIssuerMetadata>(agentContext, formattedUrl)
  if (!metadata) return null
  
  if (metadata.jwks) {
      const jwk = metadata.jwks.keys.find((key) => key.kid === kid)
      if (!jwk) return null
      
      return { jwk: PublicJwk.fromUnknown(jwk), issuer: metadata.issuer }
  }
  
  // jwks_uri is always defined if jwks was undefined
  const jwks = await fetch<Jwks>(agentContext, metadata.jwks_uri)
  if (!jwks) return null
  
  const jwk = jwks.keys.find((key) => key.kid === kid)
  if (!jwk) return null
  
  return { jwk: PublicJwk.fromUnknown(jwk), issuer: metadata.issuer }
```

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerModule.ts
```diff
@@ -115,6 +116,7 @@ export class OpenId4VcIssuerModule {
     // Configure endpoints
     configureIssuerMetadataEndpoint(endpointRouter)
     configureJwksEndpoint(endpointRouter, this.config)
+    configureJwtVcIssuerMetadataEndpoint(endpointRouter)
```
TimoGlastra: While i understand to add it to the openid4vc router, it does tightly couple the credential format with the issuance protocol. We could add this to the openid4vc module, so it could just register some additional routes, but we would need a new record type for it in that case, and now we can just use the openid4vc issuer record.

I think for this is fine, and we can always make it more flexible later on 👍

TimoGlastra: Hmm after thinking about this a bit more, I'm not so sure, since it would mean the jwt-vc-issuer would use the same url as the openid4vc-issuer. I'd expect you'd more often use an url like `example.com` over `example.com/oid4vci/<uuid>` as the value for `iss`.

Wouldn't it make more sense to just host the issuer metadata separately for now? We have the same for did:web, where you can create the did document in Credo, but you need to host it yourself on your webserver on the well konwn path

packages/openid4vc/src/openid4vc-issuer/router/jwtIssuerVcMetadataEndpoint.ts
```diff
@@ -0,0 +1,31 @@
+import type { SdJwtVcIssuerMetadata } from '@credo-ts/core'
+import type { Response, Router } from 'express'
+import { getRequestContext, sendJsonResponse, sendUnknownServerErrorResponse } from '../../shared/router'
+import { OpenId4VcIssuerService } from '../OpenId4VcIssuerService'
+import type { OpenId4VcIssuanceRequest } from './requestContext'
+
+export function configureJwtVcIssuerMetadataEndpoint(router: Router) {
+  router.get('/.well-known/jwt-vc-issuer', async (_request: OpenId4VcIssuanceRequest, response: Response, next) => {
```
TimoGlastra: This won't be hosted at the correct path looking at the spec. It will be hosted at `<url>/oid4vci/<uuid>/.well-known/jwt-vc-issuer`, while it should be hosted at `<url>/.well-known/jwt-vc-issuer/oid4vci/<uuid>`

We have this issue for other documents as well: https://github.com/openwallet-foundation/credo-ts/issues/2221

packages/core/src/modules/sd-jwt-vc/SdJwtVcService.ts
GianfrancoMS: I decided to go with `jwt-vc-issuer-metadata`

packages/core/src/modules/sd-jwt-vc/SdJwtVcOptions.ts
GianfrancoMS: Present. I will fix this.

packages/core/src/modules/sd-jwt-vc/SdJwtVcOptions.ts
GianfrancoMS: i will fix this.

TimoGlastra (commented):
Hi @GianfrancoMS, thanks a lot for this PR! It's a great addition.

I'm thinking whether we should get the support into sd-jwt-vc merged, and hold off on the oid4vci integration and just require it to be hosted separately for now. It's quite easy to add this, and I'm not convinced yet that we should bind this to the oid4vci record yet. Currently the issuance method (did/x5c) is separate from the oid4vci protocol.

packages/core/src/modules/sd-jwt-vc/typeMetadata.ts
GianfrancoMS: They are used at `resolveSigningPublicJwkFromJwtVcIssuerMetadata`

packages/core/src/modules/sd-jwt-vc/utils.ts
GianfrancoMS: Good suggestion. I will fix this.

packages/core/src/modules/sd-jwt-vc/typeMetadata.ts
TimoGlastra: They just use the types as a generic to the fetch method, but they are not actually used for validation. So we don't get runtime validation, only build time validation, and we cas the `.json()`  to the provided type.

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerModule.ts
GianfrancoMS: I think we can support both use cases where we can host our own endpoint and have issuers host on their own servers.

GianfrancoMS: Isn't this the same thing for the other well-known endpoints?
We host oauth and openid well-known endpoints.
It's the same logic here.

packages/openid4vc/src/openid4vc-issuer/router/jwtIssuerVcMetadataEndpoint.ts
GianfrancoMS: Yes, it's the same issue for other documents, but I think we should leave it here until we figure out how to host all the documents correctly.
Plus, I've added support for both URLs.

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerModule.ts
rmlearney-digicatapult: what if someone wants to issue a jwt vc over didcomm?

TimoGlastra: > We host oauth and openid well-known endpoints.

oauth and openid well-known endpoints are inherently tied to openid4vci. You can issue a credenetial over oid4vci without these metadata files. For jwt-vc-issuer metadata it's different. There's no inherent relation between the two, except that you can choose to issue an SD-JWT VC  credential over openid4vci (but as @rmlearney-digicatapult, it could theoretically also be done over DIDComm for example)

GianfrancoMS:
@TimoGlastra Let me know if you need anything else from my side to get this merged.

PR #2458, opened 2025-10-16
fix: authorization session expiry

Sshovon:
The `authorizationRequestExpiresInSeconds` getter was returning the default value (300) even when `authorizationRequestExpirationInSeconds` was passed in the verifier config during agent initialization.

The [config](https://github.com/openwallet-foundation/credo-ts/blob/c5e2a218c10b64bd5288c5f54b454259d588c162/packages/openid4vc/src/openid4vc-verifier/OpenId4VpVerifierService.ts#L101) object inside the `OpenId4VpVerifierService` class contains another `OpenId4VcVerifierModuleConfig` instance instead of the raw option attributes, causing the getter in `OpenId4VcVerifierModuleConfig` to always fall back to its default.

This PR fixes the issue by ensuring the getter resolves the correct configuration value for authorization session expiry.  I'm not entirely sure whether this nested config behavior is intentional or a construction issue.

```
OpenId4VcVerifierModuleConfig {
  options: OpenId4VcVerifierModuleConfig {
    options: {
      baseUrl: 'https://ad3d603bfbeb.ngrok-free.app/oid4vp',
      router: [Function],
      authorizationRequestExpirationInSeconds: 3600
    },
    router: [Function: router] {
      caseSensitive: undefined,
      mergeParams: undefined,
      params: [Object],
      strict: undefined,
      stack: [Array]
    }
  },
  router: [Function: router] {
    caseSensitive: undefined,
    mergeParams: undefined,
    params: { verifierId: [Array] },
    strict: undefined,
    stack: [ [Layer], [Layer], [Layer], [Layer], [Layer] ]
  }
}
```

### DIFF
--- a/packages/openid4vc/src/openid4vc-verifier/OpenId4VcVerifierModuleConfig.ts
+++ b/packages/openid4vc/src/openid4vc-verifier/OpenId4VcVerifierModuleConfig.ts
@@ -73,7 +73,7 @@ export class OpenId4VcVerifierModuleConfig {
    *
    * @default 300
    */
-  public get authorizationRequestExpiresInSeconds() {
+  public get authorizationRequestExpirationInSeconds() {
     return this.options.authorizationRequestExpirationInSeconds ?? 300
   }
 }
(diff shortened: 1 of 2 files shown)


TimoGlastra:
Hm the nested conflict behavior seems like an error. Could you also fix that in this PR?

TimoGlastra:
I think the issue should be resolved by #2464

Sshovon:
Great.. Closing the PR.

PR #2471, opened 2025-10-23
feat(didcomm): out of band proof proposals

genaris:
Supporting Out of Band proof proposals for V1 and V2 Proofs protocols. Supersedes #1370


TimoGlastra (approved):
Maybe we can add a nagative test that covers that it's not possible to reply to a connection-full proposal in a connection-less way.

So e.g.
- a sends proposal to connection id
- b modifies exchange record to make it look like a connection less one
- b send a request to a
- a should throw an error as the request should have been sent over a specific connection

packages/didcomm/src/modules/proofs/protocol/v2/__tests__/v2-indy-connectionless-proofs.e2e.test.ts
```diff
@@ -644,4 +647,98 @@ describe('V2 Connectionless Proofs - Indy', () => {
       threadId: requestMessage.threadId,
     })
   })
+
+  test('Alice start with oob proof proposal for Faber with aut-accept enabled', async () => {
```
rmlearney-digicatapult: auto

PR #2474, opened 2025-10-28
feat: add support for chained authorization code flow

hacdias:
This supersedes #2467. I'm opening as new PR since I don't have permissions to update the branch of the original PR.

### Tasks

- [x] Add new issuance record fields in Drizzle and generate necessary migrations
- [x] Update the oid4vc-ts library with the required exports / type parsings
- [x] See if it's feasible to add an E2E test
- [x] Changelog

### Follow-up Issues

- https://github.com/openwallet-foundation/credo-ts/issues/2401
- https://github.com/openwallet-foundation/credo-ts/issues/2479
- https://github.com/openwallet-foundation/credo-ts/issues/2480


demo-openid/src/Issuer.ts
```diff
@@ -317,12 +317,35 @@ export class Issuer extends BaseAgent<{
...
+        // So for now we comment this out to allow easier testing with Credo as an
+        // authorization server. How to fix it?
+        // {
+        //   type: 'direct',
+        //   issuer: PROVIDER_HOST,
+        //   clientAuthentication: {
+        //     clientId: 'issuer-server',
+        //     clientSecret: 'issuer-server',
+        //   },
+        // },
```
hacdias: @TimoGlastra this is something we have to think about. When we have "direct" authorization server configurations, they will be available in the eyes of the `oid4vc-ts` packages. If you define one of those, and then don't use it for an authorization code flow, you'll get an error: https://github.com/openwallet-foundation-labs/oid4vc-ts/blob/eaad24fd064a99450f39ce2e27485286551ab052/packages/openid4vci/src/credential-offer/credential-offer.ts#L143-L176

Are we removing those checks from the `oid4vc-ts` package, making creating a credential offer more flexible, or am I missing something?

demo-openid/src/Issuer.ts
```diff
@@ -317,12 +317,35 @@ export class Issuer extends BaseAgent<{
...
           },
+          supportedCredentials: [
+            {
+              credentialConfigurationId: 'UniversityDegreeCredential-sdjwt',
+              scopes: [
+                'https://www.googleapis.com/auth/userinfo.profile',
+                'https://www.googleapis.com/auth/userinfo.email',
+              ],
+            },
+          ],
```
hacdias: @TimoGlastra we need to think if we want a static configuration like this, or a callback. The callback would allow for more dynamic updates, but it'll move some complexity to the user of Credo. What do you think? But then I wonder: we could also just abstract the _whole_ chained identity configuration to a callback.

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
```diff
@@ -169,6 +169,15 @@ export class OpenId4VciHolderService {
...
       dpop,
+      additionalRequestPayload: {
+        // TODO: this ensures that the redirect_uri is sent to the authorization
+        // challenge endpoint, and in the end used in the auth code request bc the
+        // authorization challenge endpoint ends up working as PAR.
+        //
+        // Should we somehow in oid4vc-ts lib be "smarter" and if there is a redirect_uri
+        // then use the PAR endpoint if available instead of posting to the authorization challenge endpoint?
+        redirect_uri: redirectUri,
+      },
```
hacdias: @TimoGlastra at the end of the chained authorization code flow, Credo's authorization server `/redirect` endpoint needs to redirect somewhere else. The client (holder) provided a redirect URI to `initiateAuthorization`.

However, because we hit the authorization challenge endpoint first, and this ends up acting as a PAR request, we never get to see this redirect URI. This is because the authorization challenge endpoint should not be taking this parameter. That's why I added it as an additional payload to force it trickling down.

My take here would be to **not** make the authorization challenge endpoint act as a PAR request, but instead just throw an error as before, and then the oauth2 client will correctly make a request to the PAR endpoint with the proper `redirect_uri`. The issue there is that there will be two HTTP  requests instead of one.

**Or**: send the redirect_uri in any case here: https://github.com/openwallet-foundation-labs/oid4vc-ts/blob/eaad24fd064a99450f39ce2e27485286551ab052/packages/oauth2/src/Oauth2Client.ts#L101-L108, ensuring it'd still be used by the authorization challenge endpoint when acting as a PAR request, and Credo would throw an error when the redirect_uri is not present.

There's also this comment you had written:

https://github.com/openwallet-foundation/credo-ts/blob/1c80e445282ffd9eb72e093c02eecedc4146f467/packages/openid4vc/src/openid4vc-issuer/router/pushedAuthorizationRequestEndpoint.ts#L150-L155

That makes me a bit confused, since it sounds like a completely new redirect endpoint, and not the one the client (holder) had initially given.

What's your thoughts?

packages/openid4vc/src/openid4vc-issuer/repository/OpenId4VcIssuanceSessionRecord.ts
```diff
@@ -81,6 +81,61 @@ export interface OpenId4VcIssuanceSessionPkce {
...
+
+  /**
+   * The metadata of the external identity provider's authorization server.
+   */
+  chainedIdentityServerMetadata?: AuthorizationServerMetadata
+
+  /**
+   * The access token response received from the external identity provider.
+   */
+  accessTokenResponse?: AccessTokenResponse
```
hacdias: My idea was to just store this in the issuance session, and then Credo's user can use this information to produce the content of the credentials in the `credentialRequestToCredentialMapper` callback.

This may, or not, have an `id_token` from OpenID Connect. That all depends on the used scopes. The user can use it the way they want, they can also just use the actual access token and fetch something from a resource server. This way we would abstract those things from Credo.

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerServiceOptions.ts
```diff
@@ -85,6 +85,10 @@ export interface OpenId4VciAuthorizationCodeFlowConfig {
    * @default false
    */
   requirePresentationDuringIssuance?: boolean
+
+  // TODO: correct place? mutually exclusive with requirePresentationDuringIssuance
+  // TODO: needs correct server
+  requireChainedIdentity?: boolean
```
hacdias: @TimoGlastra Would this be the correct place to add this? I think we should also be indicating which auth server to use here, and then store it in the `chainedIdentity` field of the issuance session. Maybe replace this with:

```
chainedIdentityAuthorizationServerUrl?: string
```

Wdyt?

packages/openid4vc/src/openid4vc-issuer/repository/OpenId4VcIssuanceSessionRecord.ts
hacdias: Example: https://github.com/openwallet-foundation/credo-ts/blob/bc8bc955c8605277c78c59c25f39696aa3f24039/demo-openid/src/Issuer.ts#L128-L141

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
TimoGlastra: I think passing redirect_uri to the challenge endpoint makes sense. Since an explicit feature of the spec is that you can fallback to it being a PAR request.

I think this is also the case with the new interactive authorization endpoint in oid4vci.

TimoGlastra: Regarding the comment on redirect uri:

Since we are used chained oauth, there's multiple redirect uris involved. What i mostly meant here is that we need to provide our redirect uri to the external idp (e.g. google, where we also need to pre-register the redirect URI used by credo). After the authorizarion succeeded they will redirect to credeo's redirect uri. Credo's backend will process the authorization code from the redirect uri and issue it's own authorization code, and redirect to the redirect URI of the wallet.



We need to be careful with accepting any redirect uri from the wallet. Since for public clients it's dangerous to accept any redirect URI. With a wallet attestation it becomes less dangerous, but we need some way to verify the client on top of credo (so maybe a callback so the credo user can dynamically determine which wallets to accept).

packages/openid4vc/src/openid4vc-issuer/repository/OpenId4VcIssuanceSessionRecord.ts
TimoGlastra: Makes sense. Maybe we can rename the param to be more clear on it being a response from an external server.

TimoGlastra: Or make it part of the authorization for example

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerServiceOptions.ts
TimoGlastra: Yeah i guess so. I would like to think how we can also support dynamic issuance better. We support it already for external authorization server. So we create an issuance session when someones initiates an authorization request. That would allow us to create a static offer that can be used as long as you authorize.

But in this case we'd have to store in the issuer metadata which chained server (if any) should be used for each of the credential configurations. Or it should maybe be a callback as you mentioned.

demo-openid/src/Issuer.ts
TimoGlastra: With moving the whole chained auth to a callback, what would be the benefit over using an external auth server? The goal is to make it possible to leverage credo's oid4vc features for the authorization flow. Ideally as easy as possible.

Im open to a callback for the configuration, it would make syncing easier, but it also adds complexity so we'd have to think about how to combine the approaches. Since the issuer metadaya is kinda a "static" well known file, i think it's good to have them somewhat static. But the linker chained identity servers don't have to be static at all

demo-openid/src/Issuer.ts
TimoGlastra: Not sure i fully understand the question, but in the case of direct auth flow we add the external auth server to the issuer metadata. In case of chained auth flow we add credo as the authorization server. So i think these checks can stay?

packages/openid4vc/src/openid4vc-issuer/router/credentialEndpoint.ts
```diff
@@ -274,7 +274,7 @@ export function configureCredentialEndpoint(router: Router, config: OpenId4VcIss
         },
         openId4VciVersion:
           issuerMetadata.originalDraftVersion === Openid4vciDraftVersion.Draft15 ||
-          issuerMetadata.originalDraftVersion === Openid4vciDraftVersion.Draft16
+          issuerMetadata.originalDraftVersion === Openid4vciDraftVersion.V1
             ? 'v1.draft15'
```
TimoGlastra: shouldn't we update the openid4vci version here to v1? it's odd that v1 results in v1.draft15 being stored on the record?

demo-openid/src/Issuer.ts
```diff
@@ -117,15 +118,33 @@ function getCredentialRequestToCredentialMapper({
...
+    // Example of how to use the the access token information from the chained identity server.
+    let authorizedUser = authorization.accessToken.payload.sub
+    if (
+      issuanceSession.chainedIdentity?.externalAccessTokenResponse?.id_token &&
+      typeof issuanceSession.chainedIdentity?.externalAccessTokenResponse?.id_token === 'string'
+    ) {
+      const claims = decodeJwt(issuanceSession.chainedIdentity.externalAccessTokenResponse.id_token)
+      if (typeof claims.email === 'string') {
+        authorizedUser = claims.email
+      }
```
TimoGlastra, line 150: Can we verify this ID Token? To make sure the demo doesn't skip this important security step.

Thinking of it i think we need a callback to handle (optionally) the authorization response from the external authorization server.  In case there's just the normal `authorization_code` code flow, we can exchange it for an access token, but in the case of an id token, we either need to verify that id token, or we need to let the credo user verify it in a callback

packages/openid4vc/src/openid4vc-issuer/router/redirectEndpoint.ts
```diff
@@ -0,0 +1,190 @@
...
+  router.get(
+    config.redirectEndpoint,
+    async (request: OpenId4VcIssuanceRequest, response: Response, next: NextFunction) => {
+      const requestContext = getRequestContext(request)
+      const { agentContext, issuer } = requestContext
+      const openId4VcIssuerService = agentContext.dependencyManager.resolve(OpenId4VcIssuerService)
+
+      let issuanceSession: OpenId4VcIssuanceSessionRecord | null = null
+      try {
+        // TODO: oid4vc-ts needs method to parse authorization response / redirect url.
```
TimoGlastra: Will you pick this up as part of this PR?

packages/openid4vc/src/openid4vc-issuer/router/redirectEndpoint.ts
```diff
@@ -0,0 +1,190 @@
...
+            },
+            {
+              internalMessage: `Issuance session '${issuanceSession.id}' does not have identity chaining configured, so it's not compatible with the redirect endpoint.`,
+            }
+          )
+        }
+
+        if (chainedIdentityAuthorizationCode && typeof chainedIdentityAuthorizationCode === 'string') {
+          // TODO: is storing the whole metadata a good idea? Or should we refetch? What if it has changed?
+          const authorizationServerMetadata = issuanceSession.chainedIdentity.externalAuthorizationServerMetadata
```
TimoGlastra: I don't think we should store the whole metadata, maybe if we need one piece we can store it, but otherwise i'd refetch and cache it (which can be somethign we set up generically later). If it changes in an incompatible way during the exchange (which is highly unlikely) there's a good chance it will result in errors as well.

packages/openid4vc/src/openid4vc-issuer/router/redirectEndpoint.ts
```diff
@@ -0,0 +1,190 @@
...
+                internalMessage: `Missing chained identity server metadata on issuance session '${issuanceSession.id}'`,
+              }
+            )
+          }
+
+          const oauth2Client = openId4VcIssuerService.getOauth2Client(agentContext, issuer)
+
+          // Retrieve / verify access token.
+          // TODO: add support for DPoP
+          const { accessTokenResponse } = await oauth2Client
```
TimoGlastra: See earlier comment about needing callback, or also supporting the specific response type as used by e.g. google.

If the `openid` scope is included, there will be an id token in the access token response as well. We need to verify this token. I think verifying an ID Token is quite simple, so we might want to just do that. But open to other solutions as well.

Or we should explicitly clear that the id token is not verified (by e.g. placing it in a separate `unverifiedAccessTokenResponse`).

packages/openid4vc/src/openid4vc-issuer/router/redirectEndpoint.ts
```diff
@@ -0,0 +1,190 @@
...
+
+        if (error && typeof error === 'string') {
+          issuanceSession.errorMessage = `Error occurred during identity chaining. ${error}: ${request.query.error_description}`
+          await openId4VcIssuerService.updateState(agentContext, issuanceSession, OpenId4VcIssuanceSessionState.Error)
+
+          const redirectUri = new URL(issuanceSession.chainedIdentity.redirectUri)
+          redirectUri.searchParams.set('error', Oauth2ErrorCodes.ServerError)
+
+          // TODO: correctly map the error codes from the underlying oauth2 flow
+          // redirectUri.searchParams.set('error_description')
```
TimoGlastra: I think it would be good to add support for the most common params. e.g `access_denied` and `invalid_grant`

packages/openid4vc/src/shared/models/OpenId4VciAuthorizationServerConfig.ts
```diff
@@ -1,12 +1,77 @@
...
+ * Perform authentication based on a client secret. It will dynamically use
+ * `client_secret_post` or `client_secret_basic` based on the method supported
+ * by the authorization server.
+ */
+export interface OpenId4VciAuthorizationServerClientAuthenticationClientSecret {
+  /**
+   * @note if no type is defined, the default is `clientSecret` due to older versions
+   * of Credo not having a type
+   */
+  type: 'clientSecret'
```
TimoGlastra, line 32: storage of older versions won't have a type, so i made it optional in the type, have you handled this on another layer (e.g. by setting the default)?

packages/openid4vc/src/shared/models/OpenId4VciAuthorizationServerConfig.ts
```diff
@@ -1,12 +1,77 @@
...
+   *
+   * This will be used for exchanging the authorization code for an access token
+   */
+  clientAuthentication: OpenId4VciAuthorizationServerClientAuthenticationClientSecret
+
+  /**
+   * Mapping between credential scopes and authorization server scopes.
+   *
+   * This is mandatory. If a scope is missing, an error will be thrown when making
+   * a credential offer.
```
TimoGlastra: Are there cases where it's part of the default scope? So if i authorize i already have enough data to issue the cred, so no additional scopes are needd. Should an empty string be used in that case? Empty array?

demo-openid/src/IssuerInquirer.ts
```diff
@@ -56,17 +56,19 @@ export class IssuerInquirer extends BaseInquirer {
     }
 
     const authorizationMethod = await this.pickOne(
-      ['Transaction Code', 'Browser', 'Presentation', 'None'],
+      ['Transaction Code', 'Browser', 'Browser (Google Account)', 'Presentation', 'None'],
```
TimoGlastra: google account option should only be availalbe if env variables are provided i think?

packages/drizzle-storage/migrations/openid4vc/postgres/meta/_journal.json
```diff
@@ -8,6 +8,13 @@
       "when": 1758141133236,
       "tag": "0000_initial",
       "breakpoints": true
+    },
+    {
+      "idx": 1,
+      "version": "7",
+      "when": 1761744029446,
+      "tag": "0001_loud_reptil",
```
TimoGlastra: when generating these you can pass (i think) `--name`, which could help with identifying the changes over time

packages/drizzle-storage/src/openid4vc/openid4vc-issuance-session-record/DrizzleOpenId4VcIssuanceSessionRecordAdapter.ts
```diff
@@ -71,6 +77,12 @@ export class DrizzleOpenId4VcIssuanceSessionRecordAdapter extends BaseDrizzleRec
             codeExpiresAt: record.authorization.codeExpiresAt?.toISOString(),
           }
         : undefined,
+      chainedIdentity: record.chainedIdentity
+        ? {
+            ...record.chainedIdentity,
+            requestUriExpiresAt: record.chainedIdentity.requestUriExpiresAt?.toISOString(),
```
TimoGlastra, line 83: if we transform here, shouldn't we transform back in the `toRecord` method?

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerService.ts
```diff
@@ -969,9 +975,9 @@ export class OpenId4VcIssuerService {
         : []
 
     const authorizationServers =
-      issuerRecord.authorizationServerConfigs && issuerRecord.authorizationServerConfigs.length > 0
+      directAuthorizationServerConfigs && directAuthorizationServerConfigs.length > 0
         ? [
-            ...issuerRecord.authorizationServerConfigs.map((authorizationServer) => authorizationServer.issuer),
+            ...directAuthorizationServerConfigs.map((authorizationServer) => authorizationServer.issuer),
             // Our issuer is also a valid authorization server (only for pre-auth)
```
TimoGlastra:
```suggestion
            // Our issuer is also a valid authorization server (for pre-auth and chained auth)
```

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerService.ts
```diff
@@ -998,17 +1004,19 @@ export class OpenId4VcIssuerService {
...
-      // require_pushed_authorization_requests: true
+      pushed_authorization_request_endpoint: joinUriParts(issuerUrl, [config.pushedAuthorizationRequestEndpoint]),
+      require_pushed_authorization_requests: true,
 
       code_challenge_methods_supported: [PkceCodeChallengeMethod.S256],
       dpop_signing_alg_values_supported: issuerRecord.dpopSigningAlgValuesSupported,
     } satisfies AuthorizationServerMetadata
 
     return {
+      originalDraftVersion: Openid4vciDraftVersion.V1,
```
TimoGlastra, line 1019: Thinking, how do we handle the multi-draft issuer metadata, because they are incompatible right? Especially the credential configurations.

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerService.ts
```diff
@@ -1348,6 +1362,13 @@ export class OpenId4VcIssuerService {
         authorizationServerUrl = issuerMetadata.credentialIssuer.credential_issuer
       }
 
+      if (
+        issuer.authorizationServerConfigs?.find((server) => server.issuer === authorizationServerUrl)?.type ===
+        'chained'
+      ) {
+        authorizationServerUrl = issuerMetadata.credentialIssuer.credential_issuer
+      }
+
```
TimoGlastra:
```suggestion
      const authorizationServerConfig = issuer.authorizationServerConfigs?.find((server) => server.issuer === authorizationServerUrl)
      if (authorizationServerConfig?.type === 'chained') {
        authorizationServerUrl = issuerMetadata.credentialIssuer.credential_issuer
      }

```

packages/openid4vc/src/openid4vc-issuer/repository/OpenId4VcIssuanceSessionRecord.ts
```diff
@@ -153,6 +215,14 @@ export interface OpenId4VcIssuanceSessionRecordProps {
   // Transaction data for deferred credential issuances
   transactions?: OpenId4VcIssuanceSessionRecordTransaction[]
 
+  /**
+   * Identity chaining enables doing another OAuth2 authentication flow as part
+   * of the OpenID4VCI authorization flow. This allows leveraging the advances OAuth2
```
```suggestion
   * of the OpenID4VCI authorization flow. This allows leveraging the advanced OAuth2
```

packages/openid4vc/src/openid4vc-issuer/repository/OpenId4VcIssuanceSessionRecord.ts
```diff
@@ -266,6 +336,45 @@ export class OpenId4VcIssuanceSessionRecord extends BaseRecord<DefaultOpenId4VcI
...
+   */
+  @Transform(({ type, value }) => {
+    if (
+      type === TransformationType.PLAIN_TO_CLASS &&
+      isJsonObject(value) &&
+      typeof value.requestUriExpiresAt === 'string'
+    ) {
+      return {
+        ...value,
+        requestUriExpiresAt: new Date(value.requestUriExpiresAt),
```
TimoGlastra, line 353: ignore my previous comment!

packages/openid4vc/src/openid4vc-issuer/repository/OpenId4VcIssuerRecord.ts
```diff
@@ -102,6 +102,14 @@ export class OpenId4VcIssuerRecord extends BaseRecord<DefaultOpenId4VcIssuerReco
   public dpopSigningAlgValuesSupported?: [Kms.KnownJwaSignatureAlgorithm, ...Kms.KnownJwaSignatureAlgorithm[]]
   public batchCredentialIssuance?: OpenId4VciBatchCredentialIssuanceOptions
 
+  public get directAuthorizationServerConfigs() {
+    return this.authorizationServerConfigs?.filter((config) => config.type === 'direct')
```
TimoGlastra: type value may be undefined for agents using older storage version

packages/openid4vc/src/openid4vc-issuer/router/authorizationChallengeEndpoint.ts
```diff
@@ -148,6 +142,70 @@ async function handleAuthorizationChallengeNoAuthSession(options: {
...
+          response_type: 'code',
+          client_id: clientId,
+        },
+      },
+      request,
+    })
+
+    throw new Oauth2ServerErrorResponseError({
+      error: Oauth2ErrorCodes.RedirectToWeb,
+      error_description: 'Chained identity required before issuance',
```
TimoGlastra: To the wallet it's just important that an authorization is required before issuance.

```suggestion
      error_description: 'Authorization required before issuance',
```

packages/openid4vc/src/openid4vc-issuer/router/pushedAuthorizationRequestEndpoint.ts
```diff
@@ -0,0 +1,311 @@
...
+  }
+) {
+  const openId4VcIssuerService = agentContext.dependencyManager.resolve(OpenId4VcIssuerService)
+  const config = agentContext.dependencyManager.resolve(OpenId4VcIssuerModuleConfig)
+
+  const { issuanceSession, parsedAuthorizationRequest, issuer, request } = options
+  const issuerMetadata = await openId4VcIssuerService.getIssuerMetadata(agentContext, issuer)
+  const authorizationServer = openId4VcIssuerService.getOauth2AuthorizationServer(agentContext, {
+    issuanceSessionId: issuanceSession.id,
+  })
```
TimoGlastra: probably good to move this further down to after the initial validations, so we don't have to fetch additional data

packages/openid4vc/src/openid4vc-issuer/router/pushedAuthorizationRequestEndpoint.ts
```diff
@@ -0,0 +1,311 @@
...
+    issuanceSession.dpop = {
+      // If dpop is provided at the start, it's required from now on.
+      required: true,
+      dpopJkt: dpop.jwkThumbprint,
+    }
+  }
+
+  if (clientAttestation) {
+    issuanceSession.walletAttestation = {
+      // If dpop is provided at the start, it's required from now on.
```
TimoGlastra:
```suggestion
      // If client attestation is provided at the start, it's required from now on.
```

packages/openid4vc/src/openid4vc-issuer/router/pushedAuthorizationRequestEndpoint.ts
```diff
@@ -0,0 +1,311 @@
...
+  const authorizationServerUrl = issuanceSession.chainedIdentity.externalAuthorizationServerUrl
+
+  const authorizationServerConfig = issuer.chainedAuthorizationServerConfigs?.find(
+    (config) => config.issuer === authorizationServerUrl
+  )
+  if (!authorizationServerConfig) {
+    throw new Oauth2ServerErrorResponseError(
+      {
+        error: Oauth2ErrorCodes.ServerError,
+        error_description: `The issuer is not configured to use an external authorization server for identity chaining.`,
```
TimoGlastra: this feels more like an internal error message.

packages/openid4vc/src/openid4vc-issuer/router/pushedAuthorizationRequestEndpoint.ts
```diff
@@ -0,0 +1,311 @@
...
+
+  const scopes = requestedScopes.flatMap((scope) => {
+    if (scope in authorizationServerConfig.scopesMapping) {
+      return authorizationServerConfig.scopesMapping[scope]
+    }
+
+    throw new Oauth2ServerErrorResponseError(
+      {
+        error: Oauth2ErrorCodes.ServerError,
+        error_description: `The issuer is not configured to map the requested scope '${scope}' for the external authorization server.`,
```
TimoGlastra: same here, a wallet can not really action on this, it's an error on the credo configuration side

packages/openid4vc/src/openid4vc-issuer/router/pushedAuthorizationRequestEndpoint.ts
```diff
@@ -0,0 +1,311 @@
...
+      },
+      {
+        internalMessage: `Issuer '${issuer.issuerId}' does not have a scope mapping for scope '${scope}' for external authorization server '${authorizationServerConfig.issuer}'`,
+      }
+    )
+  })
+
+  // TODO: add support for DPoP here
```
TimoGlastra: will you add this? we can also add it in a next iteration

packages/openid4vc/src/openid4vc-issuer/router/pushedAuthorizationRequestEndpoint.ts
```diff
@@ -0,0 +1,311 @@
...
+    requestUriExpiresAt: addSecondsToDate(new Date(), config.requestUriExpiresInSeconds),
+    pkceCodeVerifier: pkce?.codeVerifier,
+    externalState: chainedIdentityState,
+    state: parsedAuthorizationRequest.authorizationRequest.state,
+    redirectUri: parsedAuthorizationRequest.authorizationRequest.redirect_uri,
+    externalAuthorizationServerMetadata: authorizationServerMetadata,
+  }
+
+  const { pushedAuthorizationResponse } = authorizationServer.createPushedAuthorizationResponse({
+    expiresInSeconds: 60,
```
TimoGlastra: this should use the value from the config i think?

packages/openid4vc/src/openid4vc-issuer/router/pushedAuthorizationRequestEndpoint.ts
```diff
@@ -0,0 +1,311 @@
...
+          url: fullRequestUrl,
+        } as const
+
+        const parseResult = authorizationServer.parsePushedAuthorizationRequest({
+          authorizationRequest: request.body,
+          request: requestLike,
+        })
+
+        // TODO: we could allow dynamic issuance sessions here. Maybe based on a callback?
+        // Not sure how to decide which credentials are allowed to be requested dynamically
```
TimoGlastra, line 258: i think for our case we'll need to support dynamic issuance sessions. But we could handle it by creating an offer on demand when the credential offer url is opened for now (so we can bypass this limitation and add that later)

TimoGlastra (approved):
Really nice! I did a thorough review, and mostly looks good. Very excited about this feature :-)

packages/openid4vc/src/openid4vc-issuer/router/redirectEndpoint.ts
hacdias: Why don't we just pass forward the full error? Is there a reason not to do it? If we only want some errors, how do we decide where to put the border between what we want to forward and what we don't?

TimoGlastra: Yeah maybe that'd be good? However there might be some errors the wallet can't action on, in which case i don't see the value in sending that to the wallet.

E.g. if we in our backend integrate with stripe and the stripe api key is invalid, then i wouldn't send that in. A detailed error message. I'd send it as a server error, as there's a misconfiguratiom on the server. Since credo is also a client in this case, there may be a misconfiguration on the client and thus the wallet wouldn't need to know the detailz of that.

packages/openid4vc/src/openid4vc-issuer/router/redirectEndpoint.ts
hacdias: I'm leaning towards a strategy as follows:

- Automatically verify the `id_token` if the `openid` scope is included by using a similar strategy to https://github.com/openwallet-foundation-labs/oid4vc-ts/blob/main/packages/oauth2/src/access-token/verify-access-token.ts, by checking the jwks_uri.
- Make clear to the user via a comment that we verify the id_token, but that is all. The user is then responsible for checking it later, when issuing the credential.

What we could do differently is:

- Add a `verifyAccessToken` callback in the issuer configuration. Call it for the access token.
- Do not directly verify the id_token, but export a method to do it, perhaps in the `oauth2` package, since that's where it makes the most sense. the user is then free to use that method to validate the id_token if they were expecting it, but also other things

Wdyt?

packages/openid4vc/src/shared/models/OpenId4VciAuthorizationServerConfig.ts
hacdias: Yes for the empty array, I will clarify the comment.

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerService.ts
hacdias: Good question. This option wasn't mandatory until I updated the `oid4vc-ts` packages. I'm not sure why it became mandatory, but I would assume this won't affect anything, since this is the just the indicator of the version that we, ourselves, generate.

packages/openid4vc/src/openid4vc-issuer/router/redirectEndpoint.ts
TimoGlastra: I agree with verifying the id token and nothing else

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerService.ts
TimoGlastra: I think we should add an option when calling createIssuer/uodateIssuer to specify the version? That's how i'd handle this in Pararym i think.

Can be a separate PR

packages/openid4vc/src/openid4vc-issuer/router/pushedAuthorizationRequestEndpoint.ts
hacdias: DPoP support is pushed to the next iteration I'd say.

packages/openid4vc/src/shared/models/OpenId4VciAuthorizationServerConfig.ts
hacdias: Since we don't really use the `type` anywhere yet, I did not really handle the defaults.

demo-openid/src/Issuer.ts
hacdias: It's now verified, I also added comments to the example explaining.

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerService.ts
hacdias: Yes, let's pick this up later.

packages/openid4vc/src/openid4vc-issuer/router/pushedAuthorizationRequestEndpoint.ts
hacdias: I think we should look into this in a different PR, once we have a clear idea of how we are actually going to create the request. I will add that to a list with follow-up points and after merging this we can open an issue.

hacdias:
@TimoGlastra this PR is ready for final review. I added a few [follow up issues](https://github.com/openwallet-foundation/credo-ts/pull/2474#issue-3562432257), which I will open once this is merged. Please take a look to see if there's anything else you think would be important.

packages/openid4vc/src/shared/models/OpenId4VciAuthorizationServerConfig.ts
TimoGlastra: I was confused with the other type introduced: `direct`. That is marked as required. What I'm worried about is that in the future we might rely on the type, and we forget that it's not defined for older storage versions, as the type does not match the possible types. So marking it as optional forces us to  handle it correctly (for bot type: 'direct' and type: 'clientSecret')

TimoGlastra:
Merging. We can add the type thing in a follow up

packages/openid4vc/src/shared/models/OpenId4VciAuthorizationServerConfig.ts
hacdias: See https://github.com/openwallet-foundation/credo-ts/pull/2481

PR #2475, opened 2025-10-30
feat: add sd-jwt credential status validation

Sshovon:
Added credential status validation for SD-JWT VCs when a status attribute is present in the credential payload.
Now the flow checks both:
- The status provder’s signature on the status list JWT (via JWKS endpoint)
- And the actual status value (only valid when status = 0 according to ietf status list draft 13)

Not 100% sure if I’ve covered all edge cases, so any suggestions or improvements are super welcome!

### DIFF
--- a/packages/core/src/modules/sd-jwt-vc/SdJwtVcService.ts
+++ b/packages/core/src/modules/sd-jwt-vc/SdJwtVcService.ts
@@ -610,8 +610,13 @@ export class SdJwtVcService {
           } when fetching status list from ${uri}. ${await response.text()}`
         )
       }
-
-      return await response.text()
+      const responseText = await response.text()
+      const statusListJwt = Jwt.fromSerializedJwt(responseText)
+      // according to draft-ietf-oauth-status-list-13 status provider jwt 'sub' must match 'status_list.uri'
+      if (statusListJwt.payload.sub !== uri) {
+        throw new CredoError(`Status list provider JWT sub does not match the status_list.uri.`)
+      }
+      return responseText
     }
   }
   private getStatusVerifier(agentContext: AgentContext) {


Sshovon:
You are absolutely correct..I missed some implementation in SD-JWT VC repo.. all these is being handled in the SD-JWT VC library.. Thanks for pointing out. 
But when I am trying to use a separate status provider, it throws me following error
```
{"serialized":"SDJWTException: Verify Error: Invalid JWT Signature","message":"Verify Error: Invalid JWT Signature","name":"SDJWTException","stack":"SDJWTException: Verify Error: Invalid JWT Signature
```
So basically the during jwt verification it fetches the status list from the uri. But I am not entirely sure how it is retrieving the public key of the status provider to validate the signature?
Am I missing something on this?

Sshovon:
I am closing the PR..but any suggestion from you will be  much appreciated.

Sshovon:
From the sd-jwt vc library doc,
```
You can pass a dedicated `statusVerifier` function in the configuration to verify the signature of the payload of the JWT of the statuslist. If no function is provided, it will fallback to the verifier that is also used for the sd-jwt-vc.
```
So I think if custom statusVerifier is not passed then it's using the default verifier which uses the issuer pub key (most probably). 

Don't you think we should fetch the signing keys from  the status provider `./well-known/jwks.json` endpoint . When a status provider can issuer or any other party.

TimoGlastra:
> So I think if custom statusVerifier is not passed then it's using the default verifier which uses the issuer pub key (most probably).

Yes that is correct. 

> Don't you think we should fetch the signing keys from the status provider ./well-known/jwks.json endpoint . When a status provider can issuer or any other party.

This discussion is coming up quite frequently lately. From our perspective it's important that there's a link between the credential isuser and the status list issuer. So for now we require the key to be the same. If the keys aren't the same, we would need some way to link them i think?

Sshovon:
> This discussion is coming up quite frequently lately. From our perspective it's important that there's a link between the credential isuser and the status list issuer. So for now we require the key to be the same. If the keys aren't the same, we would need some way to link them i think?

Yeah, that makes sense. I get why keeping the same key helps maintain the trust link between the credential issuer and the status list issuer. But I think it’d be great if the `SdJwtVcModule` could also take a custom `statusVerifier` when initializing. That way, we could handle both setups the default one where the same key issuer is used, and custom cases where status verification is handled by another trusted service. It’d make the module more flexible without breaking the standard flow.

TimoGlastra:
Yes i think that's a good interim solution. The idea is that credo will support common ways of linking them, but until there's some standard ways to do it a custom callback makes sense 👍

PR #2481, opened 2025-11-04
fix(openid4vc): migrate old auth server configs on import

hacdias:
This migrates the old authorization server configurations on import, by adding the `type` field if not set. This is similar to the migration we have for the display field.

I don't think we need a changelog since 0.6.0 isn't released yet.

### DIFF
--- a/docker-compose.arm.yml
+++ b/docker-compose.arm.yml
@@ -31,7 +31,7 @@ services:
       "
 
   cheqd-ledger:
-    image: ghcr.io/cheqd/cheqd-node:production-latest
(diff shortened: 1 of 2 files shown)


hacdias:
I'm investigating why the tests are failing.

hacdias:
[Ah, was returning the wrong value by mistake](https://github.com/openwallet-foundation/credo-ts/compare/f25319c5ba562c1dde237e974f0ca51c75ca4396..49f4e336ee5b91614745142f570bb167820e066c). Will merge once it's green.

hacdias:
There are some tests that keep failing for reasons unrelated to this PR. Could you take a look at it, please?

PR #2483, opened 2025-11-06
fix: handle string serviceEndpoint for DidCommV1Service.type

sownak:
- Upgraded cheqd SDK and ts-proto versions
- Added condition in didCheqdUtil.ts to account for DidCommV1 service type 
- Added test cases

### DIFF
--- a/docker-compose.arm.yml
+++ b/docker-compose.arm.yml
@@ -26,7 +26,7 @@ services:
       "
 
   cheqd-ledger:
-    image: ghcr.io/cheqd/cheqd-node:production-latest
+    image: ghcr.io/cheqd/cheqd-node:4.1.5
     platform: linux/amd64
     ports:
       - '26657:26657'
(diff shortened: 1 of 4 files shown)


.changeset/all-files-rush.md
```diff
@@ -0,0 +1,5 @@
+---
+'@credo-ts/cheqd': major
```
TimoGlastra: i think this file should be removed?

TimoGlastra (approved):
Do we need these changes in the `main` branch as well?

.changeset/all-files-rush.md
sownak: done.

genaris (approved):

TimoGlastra:
Hey @sownak, it seems the tests are failing. Were these passing locally for you?

sownak:
Hi @TimoGlastra, yes the tests pass locally for me, I will re-run them and update here.

sownak:
@TimoGlastra The E2E tests pass now, Unit test failed because of same artifact name. Maybe you can delete the artifact and rerun?

PR #2484, opened 2025-11-06
fix(cheqd): handle string serviceEndpoint for DidCommV1Service.type

sownak:
- Updated cheqd SDK version
- Added condition to check for DidCommV1Service.type when formatting service in Cheqd
- Added test case


sownak:
@TimoGlastra this one for main

PR #2489, opened 2025-11-11
feat: update to zod 4

hacdias:
Updates to Zod 4.

### DIFF
--- a/packages/core/src/modules/vc/data-integrity/W3cJsonLdCredentialService.ts
+++ b/packages/core/src/modules/vc/data-integrity/W3cJsonLdCredentialService.ts
@@ -360,7 +360,7 @@ export class W3cJsonLdCredentialService {
     }
 
     return proofs.map((proof) => {
-      const SuiteClass = this.signatureSuiteRegistry.getByProofType(proof.type)?.suiteClass
+      const SuiteClass = this.signatureSuiteRegistry.getByProofType(proof.type).suiteClass
       return new SuiteClass({
         LDKeyClass: WalletKeyPair,
         proof: {


packages/core/src/utils/zod-error.ts
```diff
@@ -16,7 +15,7 @@ function escapeQuotes(str: string): string {
   return str.replace(/"/g, '\\"')
 }
 
-function joinPath(path: Array<string | number>): string {
```
TimoGlastra, line 19: i think most of this can be removed since zod v4 has better human errors out of the box: https://zod.dev/error-formatting?id=zprettifyerror

pnpm-workspace.yaml
```diff
@@ -15,9 +15,7 @@ catalog:
   reflect-metadata: 0.2.2
   typescript: ~5.9.3
   "@stablelib/ed25519": ^2.0.2
-  # NOTE: keeping at zod 3 untwil we have updated oid4vc-ts to zod4 first
-  # https://github.com/openwallet-foundation-labs/oid4vc-ts/issues/102
-  zod: ^3.25.74
+  zod: ^4.1.12
```
TimoGlastra, line 18: i think i started this a while ago, and then ran into issues because credo/oid4vc-ts was not on zod  v4 yet. Since this doesn't update the other packages, i'm now wondering what the issue was. Did you notice any errors?

hacdias: Nope

packages/core/src/utils/zod-error.ts
hacdias: Ah, that's nice. I will update this PR tomorrow then.

pnpm-workspace.yaml
TimoGlastra: ah but we are already on zod v4 now with oid4vc-ts.

hacdias:
@TimoGlastra this has been updated to address the feedback.

PR #2491, opened 2025-11-12
build(webvh): update didwebvh-ts to latest version

lotharking:
- This new version implements support for numeric URLs (e.g. `2060.io`)
- In version 2.5.5, attempting to convert such URLs to ASCII caused a break in URL handling. This update fixes that issue and ensures proper usage of numeric URLs

### DIFF
--- a/docker-compose.arm.yml
+++ b/docker-compose.arm.yml
@@ -26,7 +26,7 @@ services:
       "
 
   cheqd-ledger:
-    image: ghcr.io/cheqd/cheqd-node:production-latest
+    image: ghcr.io/cheqd/cheqd-node:4.1.5
     platform: linux/amd64
     ports:
       - '26657:26657'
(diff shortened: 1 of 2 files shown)


packages/webvh/package.json
```diff
@@ -26,7 +26,7 @@
     "@credo-ts/core": "workspace:*",
     "class-transformer": "^0.5.1",
     "class-validator": "0.14.1",
-    "didwebvh-ts": "2.5.5",
+    "didwebvh-ts": "2.5.6",
```
TimoGlastra: Shouldn't we set the dependency to `^2.5.6` (see `^`), so we automatically get fixes and security updates

TimoGlastra:
We have the same issue with cheqd that was resovled in https://github.com/openwallet-foundation/credo-ts/pull/2481, could you apply the cheqd docker compose changes to this PR as well? (so just pinning the version)

lotharking:
Hey @TimoGlastra, just pinning the Docker version? That’s ok on my side.

Issue #2262, opened 2025-04-24
Move revocation data to the credential record rather than the DIDComm exchange record


TimoGlastra:

TimoGlastra:
- We should store metadata about the protocol on the credential record
  - connection id for didcomm
    - also other tags to allow querying
  - for openid4vc we need to store
    - issuer id
    - notification id
- it should work even when the exchange record is deleted
- probably needs a migration script. We could store it on both for now, to not introduce breakages.

TimoGlastra:
@genaris added an option to send a revocation notification by didcomm connection id (#2525), we will move the revocation data in a next update.

PR #2525, opened 2025-11-17
feat(didcomm): send revocation notification without credex record

genaris:
This is a partial work towards #2262, since we don't want the revocation notification to be dependent on the Credential Exchange record (especially on issuer side, since we don't want to permanetly store personal data).

It is a breaking change on the API but quite straightforward to upgrade, since Credential Exchange records already store the associated `connectionId`.

### DIFF
--- a/.changeset/rude-swans-breathe.md
+++ b/.changeset/rude-swans-breathe.md
@@ -1,0 +1,6 @@
+---
+"@credo-ts/anoncreds": patch
+"@credo-ts/didcomm": patch
+---
+
+feat(didcomm): allow sending revocation notifications without the need of keeping the related Credential Exchange record


packages/didcomm/src/modules/credentials/DidCommCredentialsApi.ts
```diff
@@ -513,43 +513,30 @@ export class DidCommCredentialsApi<CPs extends DidCommCredentialProtocol[]> impl
...
+   *
+   * @param connectionId The id of the connection record to send revocation notification
+   * @param revocationFormat Format of the credential to be revoked (anoncreds, indy-anoncreds)
+   * @param revocationId Credential revocation id
    *
-   * @param credentialExchangeRecordId The id of the credential record for which to send revocation notification
    */
   public async sendRevocationNotification(options: SendRevocationNotificationOptions): Promise<void> {
-    const { credentialExchangeRecordId, revocationId, revocationFormat, comment, requestAck } = options
+    const { connectionId, revocationId, revocationFormat, comment, requestAck } = options
```
TimoGlastra: since we required the `revocationId` and `revocationFormat` already to be passed using `connectionId` makes sense. However we do lose the functionality here to send a revocation notification to a connectionless exchange.

Maybe we can allow either the `connectionId` or `credentialExchangeRecordId`? And then we should probably make it `credentialExchangeId` to be consistent?

genaris: We weren't actually supporting connection-less flows for this particular message, but since I've already doing that for the work on #2262 I've added it here as well. 

Regarding connectionId / credentialExchangeRecordId naming, note that we are using those in all other interfaces defined in `DidCommCredentialsApiOptions` and `DidCommProofsApiOptions`. We are of course on time to refactor them to improve consistency all over the DIDComm module. It seems to me that credentialExchangeId and proofExchangeId would be a good fit.

PR #2530, opened 2025-11-20
fix(askar): create profile on provisioning

hacdias:


packages/askar/src/AskarModule.ts
```diff
@@ -56,8 +56,23 @@ export class AskarModule implements Module {
...
+    if (this.config.multiWalletDatabaseScheme === AskarMultiWalletDatabaseScheme.ProfilePerWallet) {
+      const storeManager = agentContext.dependencyManager.resolve(AskarStoreManager)
+      const { store, profile } = await storeManager.getInitializedStoreWithProfile(agentContext)
+      if (!profile) return
+
+      const profiles = await store.listProfiles()
+      if (profiles.includes(profile)) return
+
+      // Create profile for this context
+      await store.createProfile(profile)
```
TimoGlastra: I think we should just call `createProfile` here. `listProfiles` can become inefficient when you have a LOT of wallets.

TimoGlastra: I think Askar throws a duplicate error if the profile already exists, so we can also catch that and swallow the error, as it also means we got to the desired end state

hacdias: I updated the logic and removed the failing test, which is now outdated because we ensure the existence of the profile on a different level.

TimoGlastra (approved):

PR #2532, opened 2025-11-21
chore: update dependencies again

TimoGlastra:
- ~~Add support for Node 24~~ (see note below, not yet sadly)
- Update to openid4vc 0.3.x stable
- Update a few more dependencies

FYI @genaris @bryce-mcmath, i was doing some more testing with node 24 today and it seemed to work now? Not sure why, but maybe we don't need to rewrite the FFI layer just yet.


TimoGlastra:
> FYI @genaris @bryce-mcmath, i was doing some more testing with node 24 today and it seemed to work now? Not sure why, but maybe we don't need to rewrite the FFI layer just yet.

Ah it seems to work on macOS, not on Linux. So ignore this

PR #2539, opened 2025-11-27
chore: only output ESM

TimoGlastra:
This updates the build pipeline to only emit ESM. We have experienced issues integrating into Node.JS with some libraries requiring the CJS output, and other the ESM output. Since Credo is only meant to be installed a single time for the dependency injection to work correctly, this resulted in unexpected behavior.

As is explained in the NOTE, we can/should be able to safely update to ESM only.

### DIFF
--- a/packages/action-menu/package.json
+++ b/packages/action-menu/package.json
@@ -14,11 +14,7 @@
     "types": "./build/index.d.mts",
     "module": "./build/index.mjs",
     "exports": {
-      ".": {
-        "import": "./build/index.mjs",
-        "require": "./build/index.mjs",
-        "types": "./build/index.d.mts"
-      },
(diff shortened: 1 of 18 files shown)


hacdias (approved):

packages/action-menu/package.json
```diff
@@ -11,14 +11,12 @@
...
   "publishConfig": {
     "access": "public",
-    "main": "./build/index.js",
-    "types": "./build/index.d.ts",
+    "types": "./build/index.d.mts",
     "module": "./build/index.mjs",
     "exports": {
       ".": {
         "import": "./build/index.mjs",
-        "require": "./build/index.js",
```
hacdias, line 20: We probably still need the require here, but pointing at the `.mts` file.

PR #2550, opened 2025-12-02
feat: support caching for anoncreds resolving

TimoGlastra:
Basically what we have for DIDs but then for AnonCreds

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -46,6 +46,6 @@ jobs:
       # Upload the results to GitHub's code scanning dashboard (optional).
       # Commenting out will disable upload of results to your repo's Code Scanning dashboard
       - name: 'Upload to code-scanning'
-        uses: github/codeql-action/upload-sarif@fdbfb4d2750291e159f0156def62b853c2798ca2 # v3.29.5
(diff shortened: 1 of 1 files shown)


hacdias (approved):

swcurran:
@TimoGlastra -- can you point me to "what we have for DIDs"?  I'm interested in how caching is handled in the "real world" and wondering what Credo is doing for DIDs and AnonCreds objects.  Thanks.

TimoGlastra:
The default is to cache DIDs for 5 minutes, but only if the DID method allows it (so e.g. did:key and did:peer are not cached). We also serve DID documents from local records over resolving them (always), this means that to issue a credential, you don't need to resolve the DID Document at all.

We could increase the timeout, but that would result in issues as mentioned in the ACA-Py issue. However it is configurable when you resolve a did document whether to cache and for how long (overwriting the default of 5 minutes).

The main reason why we introduced it (and why we also now added it for AnonCreds) is to cover:
- one issuance/verification process sometimes resolves a did document multiple times. The cache makes this a lot faster and for this 5 minutes is plenty
- A ledger may not always be able to resolve a DID/anoncreds object right after it was written (due to how ledgers work). This results in errors sometimes if you immediately register a cred def after a schema. The local record 'cache' prevents this if you're doing it one the same agent.

Other than that it's up to the Credo user to decide how to configure the cache.

PR #2562, opened 2025-12-04
feat: support sd-jwt-vc without iss value

TimoGlastra:
And some other tweaks based on latest spec


TimoGlastra:
Fixes https://github.com/openwallet-foundation/credo-ts/issues/2407

PR #2563, opened 2025-12-04
fix(openid4vc): throw proper validation error on dcql

hacdias:
Throw the correct invalid request error when failing to validate a DCQL presentation. This aligns with the PEX behavior.


hacdias:
However, we could argue that this is indeed a server error, and that the PEX handler should not be resulting in an `invalid_request`.

PR #2571, opened 2025-12-08
chore: update to new askar

TimoGlastra:
Testing to see if all tests pass with the new Koffi-based askar

### DIFF
--- a/.changeset/tall-experts-pump.md
+++ b/.changeset/tall-experts-pump.md
@@ -1,0 +1,5 @@
+---
+"@credo-ts/askar": patch
+---
+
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -71,7 +71,18 @@
   "pnpm": {
     "overrides": {
       "vite": "npm:rolldown-vite@latest"
-    }
+    },
+    "onlyBuiltDependencies": [
```
TimoGlastra: This should be in the workspace file

TimoGlastra:
Can you add a changeswt?

PR #2580, opened 2025-12-08
fix(didcomm): send problem report when in proposal-sent

genaris:
This is to support the case where the proposal is dismissed due to factors such as a timeout or user manually cancelling the flow.


TimoGlastra:
proposal should be under lastSentMessage right?

PR #2581, opened 2025-12-09
fix: also update authorization server configs on update

TimoGlastra:
Also take into account the authorization server configs when updating


TimoGlastra:
I've also added a changeset for #2578 as i forgot to add one there

PR #2582, opened 2025-12-09
fix(askar): wallet error type

genaris:
Recently, by mistake I did not properly create the Askar storage configuration object and I got a weird runtime error saying that WalletError was not found. Indeed, this type does not exist anymore. 

The // @ts-expect-error makes me think that somehow this is left on purpose, but just in case I create this PR :smile:

### DIFF
--- a/.changeset/khaki-ends-love.md
+++ b/.changeset/khaki-ends-love.md
@@ -1,0 +1,5 @@
+---
+"@credo-ts/didcomm": patch
+---
+
+fix: allow sending Problem Report when a DIDComm proofs flow is in proposal-sent state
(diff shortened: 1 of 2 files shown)


packages/askar/src/utils/askarStoreConfig.ts
```diff
@@ -48,8 +48,7 @@ export const uriFromStoreConfig = (
       database.credentials.password
     )}@${database.config.host}/${encodeURIComponent(storeConfig.id)}`
   } else {
-    // @ts-expect-error
-    throw new WalletError(`Storage type not supported: ${database.type}`)
+    throw new AskarError(`Storage type not supported: ${database}`)
```
TimoGlastra: The ts expect error is there on purpose as the `database.type` does not exist (since we check all possible values). 

Not sure if we want to print the `database` instance as the error?

Maybe just `Storage type not supported` would be enough? There's not really a case in the near future i think where we will extend the storage type

TimoGlastra:
```suggestion
    throw new AskarError('Storage type not supported')
```

genaris: Yeah I agree. WIth that string is enough

PR #2591, opened 2025-12-11
fix: actually store kms key id for JWK-bound sd-jwt credentials

TimoGlastra:
Only mdoc actually stores them. Added a test to ensure it's stored


hacdias (approved):

hacdias:
@TimoGlastra could you check the tests please?

PR #2599, opened 2025-12-17
fix(core): add customContexts to default context

lotharking:
Previously, adding a custom document loader required implementing a completely new class. This often caused confusion and introduced unnecessary complexity, especially when only a single additional requirement was needed. By adding a complementary custom document loader, users can add new contexts for specific scenarios.


TimoGlastra (commented):
What if you provide a custom document loader? Then the customContexts does not really have effect right?

Should we make this a parameter to the default document loader, e.g.

documenLoader: defaultDocumentLoader(additionalContexts)

So it's more clear that this is an extension to the default document loader?

Or we should change the behavior maybe, that if a custom context is defined it will ALWAYS use that first, otherwise it will fallback to the document loader. That way the custom contexts are always used

lotharking:
With this implementation, we can reuse the default document loader and create a customized by extending it with additional context.
For example, a helper factory can wrap `defaultDocumentLoader` and inject custom context while preserving the default behavior.
```
export const createCustomDocumentLoader =
  (customContexts: Record<string, unknown> = {}) =>
  (agentContext: AgentContext): DocumentLoader => {
    const coreLoader = defaultDocumentLoader(agentContext, customContexts)
    return coreLoader
  }
export const CRED_EXAM_V2 = JSON.parse('{"@context":{"@vocab":"https://www.w3.org/ns/credentials/examples#"}}')
export const customContext = {
  'https://www.w3.org/ns/credentials/examples/v2': CRED_EXAM_V2,
}

const agentOptions = getAgentOptions(
  'W3cCredentialsApi',
  {},
  {},
  {
    w3cCredentials: new W3cCredentialsModule({
      documentLoader: createCustomDocumentLoader(customContext),
    }),
  }
)
```

packages/core/src/modules/vc/data-integrity/libraries/documentLoader.ts
```diff
@@ -9,25 +9,29 @@ import { getNativeDocumentLoader } from './nativeDocumentLoader'
 
 export type DocumentLoaderWithContext = (agentContext: AgentContext) => DocumentLoader
 
-export function defaultDocumentLoader(agentContext: AgentContext): DocumentLoader {
+export function defaultDocumentLoader(
+  agentContext: AgentContext,
+  customContexts: Record<string, unknown> = {}
```
genaris:
```suggestion
  additionalContexts: Record<string, unknown> = {}
```

As per @TimoGlastra suggestion, to make it clear that these are additional contexts to be added to the default ones.

packages/core/src/modules/vc/data-integrity/libraries/documentLoader.ts
```diff
@@ -9,25 +9,29 @@ import { getNativeDocumentLoader } from './nativeDocumentLoader'
...
+): DocumentLoader {
   const didResolver = agentContext.dependencyManager.resolve(DidResolverService)
+  const CONTEXTS = { ...DEFAULT_CONTEXTS, ...customContexts }
```
genaris: It seems to me more natural to use lowercase for `contexts`, since it is a variable we are just updating within a function.

genaris (approved):

genaris (approved):

PR #2610, opened 2026-01-07
feat: add method to create certificate signing request

TimoGlastra:
Adds support for creation and parsing of certificate signing requests.

### DIFF
--- a/packages/core/src/crypto/webcrypto/CredoWebCryptoKey.ts
+++ b/packages/core/src/crypto/webcrypto/CredoWebCryptoKey.ts
@@ -1,11 +1,11 @@
 import * as core from 'webcrypto-core'
 import { PublicJwk } from '../../modules/kms'
-import type { KeyImportParams, KeyType, KeyUsage } from './types'
+import type { KeyGenAlgorithm, KeyType, KeyUsage } from './types'
 
 export class CredoWebCryptoKey extends core.CryptoKey {
   public constructor(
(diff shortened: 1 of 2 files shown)


TimoGlastra:
NOTE: this PR also contains the setup for trusted publishing. As the main branch is currently failing to release

packages/core/src/crypto/webcrypto/CredoWebCryptoKey.ts
```diff
@@ -1,11 +1,11 @@
 import * as core from 'webcrypto-core'
 import { PublicJwk } from '../../modules/kms'
-import type { KeyGenAlgorithm, KeyType, KeyUsage } from './types'
+import type { KeyImportParams, KeyType, KeyUsage } from './types'
 
 export class CredoWebCryptoKey extends core.CryptoKey {
   public constructor(
     public publicJwk: PublicJwk,
-    public override algorithm: KeyGenAlgorithm,
+    public override algorithm: KeyImportParams,
```
berendsliedrecht: This seems strange. Either the type should be renamed or the property?

berendsliedrecht (approved):

packages/core/src/crypto/webcrypto/CredoWebCryptoKey.ts
TimoGlastra: I will revert. I started adding some RSA stuff, but it requires more work.

PR #2618, opened 2026-01-13
chore(cheqd): Bump @cheqd/sdk

DaevMithran:


packages/cheqd/package.json
```diff
@@ -26,8 +26,8 @@
     "test": "jest"
   },
   "dependencies": {
-    "@cheqd/sdk": "^5.3.6",
-    "@cheqd/ts-proto": "~2.5.0",
+    "@cheqd/sdk": "~5.4.2",
```
TimoGlastra: Is this also now using ESM? I don't think we can update to ESM in 0.5.x

DaevMithran: @TimoGlastra the ts-proto v2.6.0 is cjs. But it's getting overriden by 4.2.0 which is an esm package. We are debugging the issue

tweeddalex:
Hey @TimoGlastra if you are able to review/merge this and the ESM PR today, that would be much appreciated since we are upgrading our testnet to a new version. We therefore want to minimise disruption for Credo/Paradym users

TimoGlastra:
Same here. Looks good but tests are failing

PR #2619, opened 2026-01-13
feat: add cursor based pagination for drizzle based record querying

GHkrishna:
## What
- Drizzle by default supports cursor based pagination
- This PR adds cursor based support for querying record with drizzle storage

### DIFF
--- a/.changeset/icy-camels-turn.md
+++ b/.changeset/icy-camels-turn.md
@@ -1,0 +1,7 @@
+---
+"@credo-ts/drizzle-storage": patch
+"@credo-ts/askar": patch
+"@credo-ts/core": patch
+---
+
+Introduced cursor-based pagination for Drizzle-backed storage with support for before and after cursors. This ensures stable ordering using (createdAt, id) and enables efficient bidirectional pagination for large number of records.


packages/drizzle-storage/src/adapter/BaseDrizzleRecordAdapter.ts
```diff
@@ -113,52 +113,121 @@ export abstract class BaseDrizzleRecordAdapter<
...
+        // https://orm.drizzle.team/docs/guides/cursor-based-pagination 
+        // So we skip offset in case we have cursor passed
+        if (!cursor && queryOptions?.offset !== undefined) {
           queryResult = queryResult.offset(queryOptions.offset ?? 0) as typeof queryResult
         }
 
         const result = await queryResult
         return result.map(({ contextCorrelationId, ...item }) =>
-          this._toRecord(item as DrizzleAdapterRecordValues<SQLiteTable>)
+          this._toRecord(item as DrizzleAdapterRecordValues<PostgresTable>)
```
GHkrishna, line 188: This change has nothing to do with the cursor based pagination. This seems to have been added by mistake, so we can change this. (Unless it was purposefully added)

packages/drizzle-storage/src/adapter/BaseDrizzleRecordAdapter.ts
```diff
@@ -113,52 +113,93 @@ export abstract class BaseDrizzleRecordAdapter<
...
           .from(this.table.sqlite as SQLiteTable)
-          .where(
-            and(
-              // Always filter based on context correlation id
-              eq(this.table.sqlite.contextCorrelationId, agentContext.contextCorrelationId),
-              query ? queryToDrizzleSqlite(query, this.table.sqlite, this.tagKeyMapping) : undefined
-            )
-          )
+          .where(and(...whereConditions))
+          .orderBy(asc(this.table.sqlite.updatedAt), asc(this.table.sqlite.id))
```
TimoGlastra: This changes the default ordering correct? Could have some side effects, but i guess the order makes sense

packages/drizzle-storage/src/adapter/BaseDrizzleRecordAdapter.ts
```diff
@@ -113,52 +113,93 @@ export abstract class BaseDrizzleRecordAdapter<
...
+          // Existing query filters
+          query ? queryToDrizzleSqlite(query, this.table.sqlite, this.tagKeyMapping) : undefined,
+
+          // Cursor condition (keyset pagination)
+          cursor
+            ? or(
+                cursor.updatedAt ? gt(this.table.sqlite.updatedAt, cursor.updatedAt) : undefined,
+                and(
+                  cursor.updatedAt ? eq(this.table.sqlite.updatedAt, cursor.updatedAt) : undefined,
+                  gt(this.table.sqlite.id, cursor.id)
```
TimoGlastra: Why do we use both id and updatedAt? Is just id not enough?

packages/drizzle-storage/src/adapter/BaseDrizzleRecordAdapter.ts
TimoGlastra: Which chante are you referring to?

packages/drizzle-storage/src/adapter/BaseDrizzleRecordAdapter.ts
GHkrishna: Yes you are right

packages/drizzle-storage/src/adapter/BaseDrizzleRecordAdapter.ts
GHkrishna: Actually as per the drizzle docs, since the id is not incremental(UUID4). 
> If you are using primary key which is not sequential (e.g. UUIDv4), you should add sequential column (e.g. created_at column) and use multiple cursor.

We default to `updatedAt` as our first filter and then we filter by `id`. This way we can fetch correct records even in case more than one records share same time stamp.

Link to drizzle docs: https://orm.drizzle.team/docs/guides/cursor-based-pagination

packages/drizzle-storage/src/adapter/BaseDrizzleRecordAdapter.ts
GHkrishna: The transformation was initially made for `SQLiteTable` even though the db was `postgres`

GHkrishna:
I was also wondering if the cursor should be `updatedAt` or `createdAt`?

The `updatedAt` field might change as the state changes or in case the record is updated. I'm unsure if using `createdAt` makes more sense. What are your thoughts on this?

GHkrishna:
Adding test cases.

> Also maybe we can add a supportsCursorPagination on the storage service so you can detect it. E.g. credo internally could start using cursors when possible and fallback to offset otherwise.

Okay so you are suggesting to only add a flag to the storage service interface right? (And ofc update its implementations as true or false)

I don't get the fallback part for credo, because for now, either ways only drizzle storage would be able to handle the cursor parameter. And credo might not be able to do much with cursor based pagination unless a cursor is explicitly passed through. 

Please correct me if I'm misunderstanding anything. 
PS: I'll add the variable to storage service by then.

packages/drizzle-storage/src/adapter/BaseDrizzleRecordAdapter.ts
TimoGlastra: okay it makes sense. But in that case we should not make `updatedAt` optional in the cursor options.

Would it make sense to encode it as a single cursor value that can be provided in the next query?

e.g. first time your provide something like:
```
{
  limit: 100,
}
```

You get back:

```
{
  cursor: '<createdAt encoded as number>__<id>'
}
```

Next call you provide:

```
{
  limit: 100,
  // either or both can be provided
  cursor: {
    after: '<cursor>'
    before: '<cursor>'
  }
}
```

packages/drizzle-storage/src/adapter/BaseDrizzleRecordAdapter.ts
```diff
@@ -113,52 +113,93 @@ export abstract class BaseDrizzleRecordAdapter<
...
           .from(this.table.postgres as PgTable)
-          .where(
-            and(
-              // Always filter based on context correlation id
-              eq(this.table.postgres.contextCorrelationId, agentContext.contextCorrelationId),
-              query ? queryToDrizzlePostgres(query, this.table.postgres, this.tagKeyMapping) : undefined
-            )
-          )
+          .where(and(...whereConditions))
+          .orderBy(asc(this.table.postgres.updatedAt), asc(this.table.postgres.id))
```
TimoGlastra: It's not needed to use the same `orderBy` as the cursor right? So we could allow the user to provide a custom orderBy I think?

e.g. orderBy = updatedAt descending, but we still use createdAt and id for the cursor.

Not for now, but might be good to make this configurable in a follow up PR

packages/drizzle-storage/src/adapter/BaseDrizzleRecordAdapter.ts
GHkrishna: > we should not make updatedAt optional

Yes you are right. I'm about to push that change

> Would it make sense to encode it as a single cursor value that can be provided in the next query?

This sounds cool. 
- But, would we need to return this as an additional field in every record object? (So that we can have the cursor handy in the next query).

If yes, then it would be by forming the cursor dynamically after returning each query result. (I think this becomes a overhead that we are forming a dynamic variable by combining two properties. Which gets returned to the user and then the user to make further queries passes it again while querying and we internally separate two variables when making the query to DB)

- Or maybe the user forms the cursor dynamically with the returned records, they would still have an overhead of forming the cursor themselves(date to number and the string concat) and we would again restructure it to form two diff variables when making query to DB

GHkrishna: In addition to that, if we want a single cursor we can further add index to the table? maybe that is more efficient?

> Make sure to create indices for the columns that you use for cursor to make query efficient.

As per the docs. Maybe we can add this in further PRs

packages/drizzle-storage/src/adapter/BaseDrizzleRecordAdapter.ts
GHkrishna: Actually I was thinking of the something similar. But actually in cursor the `orderBy` needs to be exactly similar or else the result would be something random.

And in case of the direction of the order(`asc` and `desc`) ig we can make it configurable. But it would be a bit confusing. Consider a case where the user change the direction of order(maybe they add `desc` instead of the default `asc`) for a specific query, so for all subsequent queries they would need to make sure to add the same changed direction or else the query result would get changed, if they fail to pass the custom order direction in any of the query.

Ig a way to tackle this would be to have a persistent default ordering value in the storage service(a class variable) that the user sets initially. This `orderDirection` variable is applied by default until it is changed again?

GHkrishna:
Also, I think it makes more sense to have the direction of `orderBy` for `createdAt` to be set in a `desc` order as a default setting for now. So that way we are sure to get the latest records first.

packages/drizzle-storage/src/adapter/BaseDrizzleRecordAdapter.ts
TimoGlastra: I think an index makes sense. 

The single cursor value is just to hide the cursor workings a bit, and just allow a string to be used for pagination. This is a common pattern, see e.g. https://jsonapi.org/profiles/ethanresnick/cursor-pagination/

packages/drizzle-storage/src/adapter/BaseDrizzleRecordAdapter.ts
TimoGlastra: Hmm good points. I think we could encode it into the cursor. But maybe we shouldn't hammer it down too much yet. 

Maybe the approach you took is fine for now (assuming the order and hardcoding based on which fields to paginate)

packages/drizzle-storage/src/adapter/BaseDrizzleRecordAdapter.ts
GHkrishna: Okay I think I understand what you are trying to say. I'll make the changes and update it.

GHkrishna: Hey @TimoGlastra 
I've updated as per the cursor specs. Mind having a look at them. 

N ig you were right. This seems much better. Not only it abstracts the cursor logic but also helps go bidirectionally and set a range

packages/core/src/storage/StorageService.ts
```diff
@@ -23,6 +24,42 @@ interface AdvancedQuery<T extends BaseRecord<any, any, any>> {
...
+  createdAt: Date
+  id: string
+}
+
+export type recordToCursorBody = {
+  createdAt: string
+  id: string
+} & Record<string, unknown>
+
+export function recordToCursor(record: recordToCursorBody): string {
```
TimoGlastra:
```suggestion
export type RecordToCursorBody = {
  createdAt: string
  id: string
} & Record<string, unknown>

export function recordToCursor(record: RecordToCursorBody): string {
```

packages/drizzle-storage/src/adapter/BaseDrizzleRecordAdapter.ts
```diff
@@ -113,52 +115,138 @@ export abstract class BaseDrizzleRecordAdapter<
...
-              eq(this.table.postgres.contextCorrelationId, agentContext.contextCorrelationId),
-              query ? queryToDrizzlePostgres(query, this.table.postgres, this.tagKeyMapping) : undefined
-            )
-          )
+          .where(and(...filteredWhere))
+          .orderBy(...orderBy)
 
+        // TODO: I think since we are adding the cursor based pagination it makes more sense to add default limit even if no limit is passed
+        // This way we take the limit passed or the default limit. Maybe '100'
+        // Or we can add it as a default limit in the storageService itself
```
TimoGlastra, line 167: This would be a breaking change. I think a default makes sense, but we should do this in 0.7 then. Should also allow passing either `0` or `-1` or `null` to indicate no limit

packages/drizzle-storage/src/adapter/BaseDrizzleRecordAdapter.ts
```diff
@@ -113,52 +115,138 @@ export abstract class BaseDrizzleRecordAdapter<
...
+        // So we skip offset in case we have cursor passed
+        if (!cursor && queryOptions?.offset !== undefined) {
+          queryResult = queryResult.offset(queryOptions.offset ?? 0) as typeof queryResult
+        }
+
+        let result = await queryResult
+
+        // Restore canonical order for backward pagination
+        if (isBackward) {
+          result = result.reverse()
```
TimoGlastra, line 261: Since we're already changing the orderBy, do we still need to reverse? It seems we're doing a double reverse now?

TimoGlastra (approved):
This looks really good and ready to merge code-wise, except that we need at least some basic tests

packages/drizzle-storage/src/adapter/BaseDrizzleRecordAdapter.ts
GHkrishna: Makes sense.

packages/drizzle-storage/src/adapter/BaseDrizzleRecordAdapter.ts
GHkrishna: Yes, we need the .reverse().

When paginating backwards, we invert the SQL orderBy to efficiently fetch the records closest to the cursor (the "bottom" of the previous page). This causes the database to return the records in reverse order (e.g., Oldest -> Newest).

We must then use .reverse() to flip them back to the canonical display order (Newest -> Oldest) so the returned page looks consistent with the rest of the application.

(PS: I also realized we should also update the SQL query to invert the id sort as well, otherwise the tie-breaker logic will end up reversed in the final result.)

GHkrishna:
Hey @TimoGlastra ,
I've added the test cases.

Also, the failing test cases seems to be something unrelated with our changes. Can you please confirm that and if so, maybe we can merge the PR?

packages/drizzle-storage/src/util.ts
```diff
@@ -1,9 +1,17 @@
+// TODO: Maybe fix this for environment apart from node
+import { join } from 'node:path'
+import { fileURLToPath } from 'node:url'
```
TimoGlastra: We should not import from node dependencies, that's why we relied on import.meta. What is not working for you? We can't merge it like this as it will break browser and React Native integration

GHkrishna: Apologies, this was something not meant to be pushed. Removed it.

GHkrishna:
I think all tests are passing now. Maybe we can merge the PR?

TimoGlastra:
This looks good @GHkrishna, could you add a changeset to the PR? I can't do that if you're creating a PR from an org account

GHkrishna:
Hey @TimoGlastra 
I've added the changeset

PR #2622, opened 2026-01-13
fix: make readme commands more elaborative and fix the migrate order

GHkrishna:
## What
- Update migrate command with more bundles
- Fix migrate command order to run successfully

### DIFF
--- a/packages/drizzle-storage/README.md
+++ b/packages/drizzle-storage/README.md
@@ -113,7 +113,7 @@ The following bundles are provided out of the box by the Drizzle storage module:
 The following defines the bundles for an agent using the `didcomm` and `anoncreds` extension modules, but doesn't use the `action-menu` extension module:
 
 ```ts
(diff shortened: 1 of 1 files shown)


packages/drizzle-storage/README.md
```diff
@@ -113,7 +113,7 @@ The following bundles are provided out of the box by the Drizzle storage module:
 The following defines the bundles for an agent using the `didcomm` and `anoncreds` extension modules, but doesn't use the `action-menu` extension module:
 
 ```ts
-const bundles = [coreBundle, didcommBundle, anonCredsBundle] as const;
+const bundles = [coreBundle, didcommBundle, actionMenuBundle, anoncredsBundle, questionAnswerBundle] as const;
```
TimoGlastra, line 116: Any specific reason for adding these?

GHkrishna: I wanted to add the migrations for them so added to bundles as well

PR #2626, opened 2026-01-16
feat: support sd-jwt-vc extends

TimoGlastra:
Based on implementation in SD-JWT-JS.

Note that there are some breaking changes with the current code, but none with 0.6.1. I added the SD-JWT resolving only after the 0.6.1 release.

Also adds the drizzle storage migration for it


TimoGlastra:
This PR needs a new release of SD-JWT JS lib and also needs: https://github.com/openwallet-foundation/sd-jwt-js/pull/359

packages/core/src/modules/sd-jwt-vc/SdJwtVcService.ts
```diff
@@ -1,7 +1,8 @@
+import type { NonEmptyArray } from '@animo-id/pex'
 import type { SDJwt } from '@sd-jwt/core'
 import { decodeSdJwtSync } from '@sd-jwt/decode'
 import { selectDisclosures } from '@sd-jwt/present'
-import { SDJwtVcInstance } from '@sd-jwt/sd-jwt-vc'
+import { SDJwtVcInstance, type VcTFetcher } from '@sd-jwt/sd-jwt-vc'
```
berendsliedrecht, line 5: Not related to your changes, but why is it `VcTFetcher` Would `VctFetcher` or `VCTFetcher` not be better?

packages/drizzle-storage/src/core/sd-jwt-vc-record/DrizzleSdJwtVcRecordAdapter.ts
```diff
@@ -19,28 +19,27 @@ export class DrizzleSdJwtVcRecordAdapter extends BaseDrizzleRecordAdapter<
...
       multiInstanceState: record.multiInstanceState,
+      typeMetadata: record.typeMetadata,
+      typeMetadataChain: record.typeMetadataChain,
       customTags,
     }
   }
 
   public toRecord(values: DrizzleSdJwtVcAdapterValues): SdJwtVcRecord {
-    const { sdAlg, alg, vct, customTags, ...remainingValues } = values
+    // biome-ignore lint/correctness/noUnusedVariables: no explanation
```
berendsliedrecht: Would be good to add an explanation here.

packages/core/src/modules/sd-jwt-vc/SdJwtVcService.ts
TimoGlastra: Maybe from VC Type Fetcher? Dunno, you can open an issue in sd-jwt?

packages/drizzle-storage/src/core/sd-jwt-vc-record/DrizzleSdJwtVcRecordAdapter.ts
TimoGlastra:
```suggestion
    // biome-ignore lint/correctness/noUnusedVariables: tags are computed
```

PR #2636, opened 2026-01-26
fix(openid4vc): add `alg` to `client_metadata` jwk for encryption

TimoGlastra:


packages/openid4vc/src/openid4vc-verifier/OpenId4VpVerifierService.ts
```diff
@@ -941,12 +941,12 @@ export class OpenId4VpVerifierService {
     ]
     const supportedProofTypes = signatureSuiteRegistry.supportedProofTypes
 
-    type JarmEncryptionJwk = Kms.Jwk & { kid: string; use: 'enc' }
+    type JarmEncryptionJwk = Kms.Jwk & { kid: string; use: 'enc'; alg: 'ECDH-ES' }
```
berendsliedrecht, line 944: Does JARM only support ECDH-ES or should this be dynamic in the future?

TimoGlastra: You could use any alg, but Credo only supports ECDH-ES with a few encryption algorithms. Since the verifier creates the request it's ok for now to hardcode this (we were basically already implicitly doing it, this just makes it a bit more explicit to the wallet)

PR #2641, opened 2026-01-26
fix: add wait for wma calculation step in e2e job

Zzocker:
This is a quick fix, ideally wait for `wma` should be moved to `packages/cheqd` test setup file as typescript function (invokes `cheqd/sdk` method).


TimoGlastra:
Thanks for this @Zzocker , is this to fix the test failures with cheqd?

Zzocker:
> Thanks for this @Zzocker , is this to fix the test failures with cheqd?

The plan is to resolve the issue; however, this fix does not appear to solve it

PR #2652, opened 2026-02-05
feat: dynamic callback for chained authorization request

hacdias:
Introduces a new callback in the issuer configuration, which can be used to dynamically provide the request scopes and additional payload to the chained authorization server that is being used with the request.

This means that the `scopesMapping` configuration is now optional. Either `scopesMapping` ot the new callback must be defined in order to fullfil a chained authorization request.

### DIFF
--- a/packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerModuleConfig.ts
+++ b/packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerModuleConfig.ts
@@ -2,7 +2,8 @@ import type { Express } from 'express'
 import type {
   OpenId4VciCredentialRequestToCredentialMapper,
   OpenId4VciDeferredCredentialRequestToCredentialMapper,
-  OpenId4VciGetVerificationSessionForIssuanceSessionAuthorization,
+  OpenId4VciGetChainedAuthorizationRequestPayload,
+  OpenId4VciGetVerificationSession,
 } from './OpenId4VcIssuerServiceOptions'
 
 const DEFAULT_C_NONCE_EXPIRES_IN = 1 * 60 // 1 minute
@@ -116,13 +117,27 @@ export interface InternalOpenId4VcIssuerModuleConfigOptions {
   deferredCredentialRequestToCredentialMapper?: OpenId4VciDeferredCredentialRequestToCredentialMapper
 
   /**
+   * @deprecated use `getVerificationSession` instead.
+   */
+  getVerificationSessionForIssuanceSessionAuthorization?: OpenId4VciGetVerificationSession
+
+  /**
    * Callback to get a verification session that needs to be fulfilled for the authorization of
    * of a credential issuance session. Once the verification session has been completed the user can
    * retrieve an authorization code and access token and retrieve the credential(s).
    *
    * Required if presentation during issuance flow is used
    */
-  getVerificationSessionForIssuanceSessionAuthorization?: OpenId4VciGetVerificationSessionForIssuanceSessionAuthorization
+  getVerificationSession?: OpenId4VciGetVerificationSession
+
+  /**
+   * Callback to get additional details for the chained authorization server flow.
+   * This will be called when a credential offer request is configured to use a chained
+   * authorization server, but the scopesMapping configuration is not defined.
+   *
+   * Required if chained authorization server flow is used without a static scopes mapping configuration.
+   */
+  getChainedAuthorizationRequestPayload?: OpenId4VciGetChainedAuthorizationRequestPayload
 
   /**
    * Custom the paths used for endpoints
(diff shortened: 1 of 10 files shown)


packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerModuleConfig.ts
```diff
@@ -124,6 +125,15 @@ export interface InternalOpenId4VcIssuerModuleConfigOptions {
...
   getVerificationSessionForIssuanceSessionAuthorization?: OpenId4VciGetVerificationSessionForIssuanceSessionAuthorization
 
+  /**
+   * Callback to get additional details for the chained authorization server flow.
+   * This will be called when a credential offer request is configured to use a chained
+   * authorization server, but the scopesMapping configuration is not defined.
+   *
+   * Required if chained authorization server flow is used without a static scopes mapping configuration.
+   */
+  getChainedAuthorizationOptionsForIssuanceSessionAuthorization?: OpenId4VciGetChainedAuthorizationOptionsForIssuanceSessionAuthorization
```
hacdias: Is this a good name? Can we make it more explicit/shorter?

packages/openid4vc/src/openid4vc-issuer/router/pushedAuthorizationRequestEndpoint.ts
```diff
@@ -184,20 +184,46 @@ export async function handlePushedAuthorizationRequest(
   const redirectUri = joinUriParts(config.baseUrl, [issuer.issuerId, 'redirect'])
   const chainedIdentityState = utils.uuid()
 
-  const scopes = requestedScopes.flatMap((scope) => {
-    if (scope in authorizationServerConfig.scopesMapping) {
-      return authorizationServerConfig.scopesMapping[scope]
+  let scopes: string[] = []
+  let additionalRequestPayload: Record<string, unknown> | undefined
+
+  if (authorizationServerConfig.scopesMapping) {
```
TimoGlastra: My assumption would be that the if the callback is registered it should always be called. So if there's a scopesMapping it will just be passed as param, and it can be returned as is (if no extra fields needed).

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerModuleConfig.ts
TimoGlastra: Since we're already in the issuance module, maybe just `getChainedAuthorizationPayload` or `getChainedAuthorizationPayload`?

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerServiceOptions.ts
```diff
@@ -234,6 +237,34 @@ export type OpenId4VciGetVerificationSessionForIssuanceSessionAuthorization = (o
...
+   * The scopes to request to the chained authorization server. If no scopes are required,
+   * an empty array should be returned.
+   */
+  scopes: string[]
+
+  /**
+   * Additional properties that will be sent as payload in the authorization request to
+   * the chained authorization server.
+   */
+  additionalPayload?: Record<string, string>
```
TimoGlastra, line 271: Should be clear that `additionalPayload` will override ANY EXISTING properties.

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerServiceOptions.ts
```diff
@@ -234,6 +237,34 @@ export type OpenId4VciGetVerificationSessionForIssuanceSessionAuthorization = (o
   }
 >
 
+export type OpenId4VciGetChainedAuthorizationOptionsForIssuanceSessionAuthorization = (options: {
```
TimoGlastra: do we already have access to the authorization request payload, so you can change it based on the current contents? It's not really important, but it may be hard to know what the payload is before extension?

packages/openid4vc/tests/openid4vci-chained.e2e.test.ts
```diff
@@ -385,4 +387,273 @@ describe('OpenId4Vc (Chained Authorization)', () => {
...
+          type: 'chained',
+          issuer: 'http://localhost:4747',
+          clientAuthentication: {
+            type: 'clientSecret',
+            clientId: idpClientId,
+            clientSecret: idpClientSecret,
+          },
+          // scopesMapping: {
+          //   UniversityDegreeCredential: ['MappedUniversityDegreeCredential', 'openid'],
+          // },
```
TimoGlastra: Should we either remove the commented code, or add e.g. a comment `// test case where scopesMapping is not provided`

hacdias: Lol, I thought I removed them.

packages/openid4vc/src/openid4vc-issuer/router/pushedAuthorizationRequestEndpoint.ts
hacdias: Indeed. I will turn it around.

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerModuleConfig.ts
hacdias: Good point, I just wanted to make it consistent with the other one, which is also very long and a bit repetitive: `getVerificationSessionForIssuanceSessionAuthorization`.

What do you think about `getChainedAuthorizationRequestPayload`?

Should we also consider renaming `getVerificationSessionForIssuanceSessionAuthorization` to just `getVerificationSession`? It's a bit hard to change that without breaking change though, unless we keep both names around.

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerServiceOptions.ts
hacdias: I don't think so. We do have more information we could provide. Anything in these lines can be provided if we deem relevant:

https://github.com/openwallet-foundation/credo-ts/blob/fbe43b15bb78b851b326fe41b1109da803ebdd9c/packages/openid4vc/src/openid4vc-issuer/router/pushedAuthorizationRequestEndpoint.ts#L38-L185

TimoGlastra: let's keep it like this right now

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerModuleConfig.ts
TimoGlastra: Ah yeah consistency is always good. We should probably rename the other one. We can add the new name and deprecate the old one and just prefer the new one over the old one

`getChainedAuthorizationRequestPayload` sounds good

PR #2665, opened 2026-02-12
feat(webvh): allow adding extra metadata to AnonCreds objects

genaris:
We are currently using this Attested Resource metadata to bind these AnonCreds objects with an object from a ledger. I don't see anything wrong for doing this in neither Attested Resources spec or did:webvh DID AnonCreds method (as long as the basic metadata fields are there), so I guess it is fine to make this open in our API.


genaris:
@TimoGlastra can you check this?

PR #2666, opened 2026-02-12
fix(openid4vc): update session state to Error if failed to create response

hacdias:
Updates the session state to `Error` if an error occurred while creating a credential response. The session is, at this point, irrecoverable and can't be continued.


packages/openid4vc/src/openid4vc-issuer/router/credentialEndpoint.ts
```diff
@@ -320,6 +320,8 @@ export function configureCredentialEndpoint(router: Router, config: OpenId4VcIss
         credentialResponse.transaction_id ? 202 : 200
       )
     } catch (error) {
+      await openId4VcIssuerService.updateState(agentContext, issuanceSession, OpenId4VcIssuanceSessionState.Error)
```
TimoGlastra, line 324: should also set the error message on the issuance session?

PR #2667, opened 2026-02-13
fix(askar): only auto-provision at initialization

genaris:
In yesterday's WG call we discussed about an issue I had, a bit difficult to debug: when a record was being queried (e.g. because the app called to `agent.didcomm.mediationRecipient.findDefaultMediator()` and the wallet wasn't yet created (agent not initialized yet or just shut down), Credo 0.6 was attempting to auto-provision the wallet, which led to an unexpected race between the code that was actually creating the wallet (i.e. re-initializing the agent) and the one that was wrongly trying to access a record.

I was thinking that a not so-elegant but probably efficient idea to work around this (and make Credo behave like it was at 0.5.x), by simply adding an optional flag for auto-proivisioning, that is only set up at initialization . So this is how it looks like. As far as I know, the only use case where we really are supposed to provision the wallet is when we initialize either the root agent or the tenants.

@TimoGlastra WDYT about this approach?


packages/askar/src/AskarStoreManager.ts
```diff
@@ -532,7 +532,7 @@ export class AskarStoreManager {
     }
   }
 
-  public async getInitializedStoreWithProfile(agentContext: AgentContext) {
+  public async getInitializedStoreWithProfile(agentContext: AgentContext, options: { provisionIfNotExists?: boolean } = {}) {
```
TimoGlastra: I like the approach, it's simple. But shouldn't we make it more generic, so that it will also only initialize in the agent initialization.

So we don't just throw an error if the store doesn't exist, but also when you're e.g. querying and the store exists, but you're not calling the initialization method

GHkrishna:
Hey guys,
Actually when testing credo with a combination of drizzle and askar, we've come across a similar issue.

The agent would throw the following error:
```
Failed to send message: Message is undeliverable to connection 2f0b851f-19d4-475d-84a9-30d3557dc679 (undefined). 
Reasons: - Key with key id '072d21a1-8cc1-481e-b01d-532b63edc1d1' not found in backend 'askar'
```

Do you think this might be a related issue, as the wallet is freshly provisioned each time?

_Environment: A containerized @credo-ts/rest initialization (credo-ts-ext/rest) with all credo packages updated to 0.6.1_


Edit: This was an unrelated issue caused because of passing incorrect wallet configuration

TimoGlastra (approved):
I like the change, but shouldn't we update the base agent isInitialized to use this isInitialized value now? I think if we keep track of it a layer lower, no need to separately track it on the root agent class.

Furthermore, it feels weird that we can still open the store if we're querying when the agent is shutdown/not initialized. Should we somehow guard this already earlier by checking if the agentContext is initialized/initializing? And then isInitialized may not be the best term but rather it should indicate whether the AgentContext is allowed to be used (which it's not if the agent is shutdown)


Accepting as my answer complicates things and we maybe just want to quickly fix the issue

PR #2704, opened 2026-03-18
feat: initial DIDComm V2 support

genaris:
New branch created from the current work in #2698. Added here to allow releasing pr-versions and test behaviour on different platforms before the initial merge to main.

Former `feat/didcomm-v2` has been renamed to `feat/didcomm-v2-old` for reference and in case we need to take some things from there.


packages/askar/src/kms/crypto/deriveKey.ts
```diff
@@ -17,6 +18,78 @@ type AskarSupportedKeyAgreementDecryptOptions = Kms.KmsKeyAgreementDecryptOption
...
+    return {
+      contentEncryptionKey,
+      encryptedContentEncryptionKey,
+    }
+  } finally {
+    ephemeralKey.handle.free()
+    derivedKey.handle.free()
+    if (contentEncryptionKey) {
+      // contentEncryptionKey is returned to caller; they free it
+    }
```
genaris: I didn't get the purpose of this empty if statement

packages/didcomm/src/__tests__/DidCommMessageReceiver.test.ts
```diff
@@ -0,0 +1,42 @@
...
+describe('DidCommMessageReceiver', () => {
+  describe('v2 message handling', () => {
+    it('throws when receiving v2 encrypted message and v2 is not in didcommVersions', async () => {
+      const agent = new Agent(
+        getAgentOptions('ReceiverTest', { didcommVersions: ['v1'] }, {}, undefined, { requireDidcomm: true })
+      )
+      setupSubjectTransports([agent])
+      await agent.initialize()
+
+      const v2Protected = JsonEncoder.toBase64URL({
```
genaris: Note the type error here because of the new toBase64Url method on TypedArrayEncoder

packages/didcomm/src/modules/basic-messages/repository/DidCommBasicMessageRecord.ts
```diff
@@ -22,6 +22,8 @@ export interface DidCommBasicMessageStorageProps {
   parentThreadId?: string
   content: string
   sentTime: string
+  /** Protocol version (1.0 or 2.0). Optional for backwards compatibility. */
+  protocolVersion?: '1.0' | '2.0'
```
genaris: In other protocols we use 'v1' and 'v2, since we expect minor versions to be compatible with the major version they belong. I think that for consistency we can keep the same approach here.

packages/didcomm/src/modules/basic-messages/protocol/v2/handlers/DidCommBasicMessageV2Handler.ts
genaris, line 1: In order to start following the convention used in other protocol modules, maybe we can reorganize the directories to separate v1 and v2 (even maintaining the same class and interface names, to not introduce a breaking change)?

packages/didcomm/src/modules/message-pickup/protocol/v3/messages/DidCommDeliveryRequestV3Message.ts
genaris, line 1: All these message classes (and their filenames) do not include the DidComm prefix, like in v1 and v2.

packages/didcomm/src/modules/message-pickup/protocol/v3/handlers/DidCommDeliveryRequestV3Handler.ts
genaris, line 1: All V3 classes and filenames do not include the DidComm prefix, like in V1 and V2-

packages/didcomm/src/modules/routing/handlers/v2/index.ts
genaris, line 1: I know it is hard to reorganize this module (since Coordinate Mediation and the base Routing protocol are mixed up), but don't forget to add the appropriate prefixes (DidComm) and suffixes (V2) as needed.

packages/didcomm/src/modules/routing/messages/v2/ForwardMessage.ts
genaris, line 1: I think this file is misleading, because it doesn't include a message class, but rather a builder function. Shouldn't this be part of ForwardMessageV2 (which should be renamed to DidCommForwardV2Message?

packages/didcomm/src/modules/routing/protocol/v1/messages/DidCommForwardMessage.ts
genaris, line 1: I think it would be good to at least move these V1 messages and handler to their corresponding `v1` directory, so they are at the same level as the `v2` ones.

packages/didcomm/src/modules/routing/messages/DidCommForwardMessage.ts
```diff
@@ -16,6 +16,7 @@ export interface DidCommForwardMessageOptions {
  */
 export class DidCommForwardMessage extends DidCommMessage {
   public readonly allowDidSovPrefix = true
+  public readonly supportedDidCommVersions: ('v1' | 'v2')[] = ['v1']
```
genaris: What about creating an enum type for valid DidComm versions to hold this 'v1' | 'v2' currently replicated in different places?

packages/didcomm/src/modules/routing/DidCommMediatorApi.ts
```diff
@@ -47,4 +49,27 @@ export class DidCommMediatorApi {
...
+    const outboundMessageContext = new DidCommOutboundMessageContext(message, {
+      agentContext: this.agentContext,
+      connection: connectionRecord,
+      associatedRecord: mediationRecord,
+    })
+
+    await this.messageSender.sendMessage(outboundMessageContext)
+
+    return mediationRecord
+  }
```
genaris: Do we really need to add a dedicated method for this? I mean, isn't it possible for the mediator to automatically tell if it should send a v1 or v2 grant message by looking at the mediation record?

packages/didcomm/src/modules/routing/DidCommMediatorModuleConfig.ts
```diff
@@ -12,6 +12,20 @@ export interface DidCommMediatorModuleConfigOptions {
...
+   * Required for v2 mediate-grant. Typically a did:peer with service_endpoint.
+   */
+  mediatorRoutingDid?: string
+
+  /**
+   * Mediation protocol versions to support. When '2.0' is included, Coordinate Mediation 2.0 handlers are registered.
+   *
+   * @default ['1.0']
+   */
+  mediationProtocolVersions?: ('1.0' | '2.0')[]
```
genaris: Here we can also use 'v1' and 'v2' like in other protocols

packages/didcomm/src/modules/routing/DidCommMediationRecipientApi.ts
```diff
@@ -372,11 +472,165 @@ export class DidCommMediationRecipientApi {
...
+      this.agentContext,
+      connection
+    )
+    const outboundMessage = new DidCommOutboundMessageContext(message, {
+      agentContext: this.agentContext,
+      connection,
+    })
+    await this.sendMessage(outboundMessage)
+    return mediationRecord
+  }
```
genaris, line 465: I wouldn't expect a DIDComm V1 connection to use Mediation Coordination 2.0 or a DIDComm V2 connection to use Mediation Coordination 1.0, so I think it would be simpler if we just use `requestMediation` method and determine which one to use based on the connection record.

packages/didcomm/src/modules/routing/DidCommMediationRecipientApi.ts
```diff
@@ -372,11 +472,165 @@ export class DidCommMediationRecipientApi {
...
+
+  public async requestAndAwaitGrantV2(
```
genaris: Same here. I think we can automatically determine if we should use V1 or V2 based on the connection record.

packages/didcomm/src/modules/routing/DidCommMediationRecipientApi.ts
```diff
@@ -372,11 +472,165 @@ export class DidCommMediationRecipientApi {
...
+        )
+      }
+    } else {
+      this.logger.debug(`Mediator invitation has already been ${mediation.isReady ? 'granted' : 'requested'}`)
+    }
+
+    return mediation
+  }
+
+  public async notifyKeylistUpdateV2(
```
genaris: Like in other comments, I think it should be possible to determine if we should use V1 or V2 based on the mediation record.

packages/didcomm/src/modules/routing/DidCommMediationRecipientApi.ts
```diff
@@ -372,11 +472,165 @@ export class DidCommMediationRecipientApi {
...
+        // Resolution can fail for exotic stacks; primary did still sent
+      }
+    }
+    return [...variants]
+  }
+
+  /**
+   * Full v2 provision: mediate-request → grant → keylist-update (post-grant, adds recipient DID).
+   */
+  public async provisionV2(connection: DidCommConnectionRecord): Promise<DidCommMediationRecord> {
```
genaris: Is the logic of provisionV2 so different from V1 that we need to duplicate most of the code? Maybe with some assumptions like the one mentioned (always use V1 for DIDComm V1, and V2 for DIDComm V2) the code of this method will be reusable.

genaris (commented):
@tarunvaddeSoul I've added several comments regarding class/interface naming and API usage that can simplify a lot the changes needed to use V2 in existing agents (since most of the logic to choose between V1 or V2 comes directly from the Connection Record).

I know there are some remaining fixes to do on the key management in order to support both V1 and V2 at the same time in an Agent, so I didn't do a deep analysis of that yet.

tarunvaddeSoul:
Hi @genaris, I have made changes as per your comments here #2752. Please review and do let me know for any more changes.

genaris:
> Hi @genaris, I have made changes as per your comments here #2752. Please review and do let me know for any more changes.

Hi @tarunvaddeSoul, can you fix the validation issues? You can check for them by running these commands:
- pnpm style:check
- pnpm types:check

In some cases it is acceptable to override linter rules for some lines or even files, but usually there is a better way to format the code.

Please also make sure all the current test suite is running. Thanks!!

packages/didcomm/src/modules/routing/DidCommMediationRecipientApi.ts
```diff
@@ -248,9 +271,25 @@ export class DidCommMediationRecipientApi {
...
+        mediatorPickupStrategy === DidCommMediatorPickupStrategy.PickUpV1 ||
+        mediatorPickupStrategy === DidCommMediatorPickupStrategy.PickUpV2
+      ) {
+        mediatorPickupStrategy = DidCommMediatorPickupStrategy.PickUpV3
+      } else if (mediatorPickupStrategy === DidCommMediatorPickupStrategy.PickUpV2LiveMode) {
+        mediatorPickupStrategy = DidCommMediatorPickupStrategy.PickUpV3LiveMode
+      }
+    } else {
+      assertDidCommV1Connection(mediatorConnection, 'Mediation')
+    }
```
genaris: We have to check this logic because sometimes I'm having troubles when running initiateMessagePickup(mediatorRecord). It attempts to run it with v3 (I didn't debug it yet deeply),

tarunvadde: Hi @genaris - thanks for flagging this, you were right to be suspicious. I traced
the issue and the root cause is asymmetric forcing in `getPickupStrategyForMediator`:

- For **CM 2.0 mediators**, v1/v2 pickup is auto-upgraded to v3 (correct - pickup
  v1/v2 can't run over a v2 connection).
- But for **CM 1.0 mediators**, a globally-configured `PickUpV3*` strategy was
  not downgraded. So if an app sets a default like `mediatorPickupStrategy:
  PickUpV3LiveMode` and provisions a CM 1.0 mediator, the code happily attempts
  v3 pickup over a v1 connection - exactly the "it attempts to run it with v3"
  you observed.

I've raised a small follow-up PR (#2766) that adds the symmetric downgrade
(v3 => v2 for CM 1.0) in both `getPickupStrategyForMediator` and the explicit-
strategy path in `initiateMessagePickup`, with a `logger.warn` so the coercion
isn't silent.

Let me know if you'd prefer it to throw instead of downgrading - I went with
silent-force-plus-warn to mirror the existing v2 upgrade behaviour, but
strict mode is a one-line change.

packages/didcomm/src/models/DidCommRouting.ts
genaris, line 1: Comments from [#2807](https://github.com/openwallet-foundation/credo-ts/pull/2807#discussion_r3303913858):

> @[TimoGlastra](https://github.com/TimoGlastra)
> As I understand it, routingKeys should/could be keyAgreementKeys. Why not extend the type of routingKeys, and if it's a X25519 key we don't transform it anymore?
> 
> @[genaris](https://github.com/genaris)
> I think this interface became confusing, because it is used for both V1 and V2 cases: in DIDComm V2, routingDid is used, so routingKeys are actually left to V1 and there we'll still have the Ed25519 as it was in legacy cases. @tarunvadde can you confirm this?
> 
> Maybe a better approach would be to use different interfaces for each, and make a strong separation between V1 and V2.
> 
> @[tarunvadde](https://github.com/tarunvadde)
> Yeah you're right. routingKeys only gets used in V1 forward-message construction, and routingDid is V2-only. With keyAgreementKey added, only endpoints is really shared.
> 
> What about discriminated union like (DidCommV1Routing | DidCommV2Routing)? over fully separate interfaces.
> 
> Separate interfaces is cleaner long-term but need to make changes into OutOfBandApi, ConnectionsApi, DidExchangeProtocol, getDidCommOutboundMessageContext.

demo/src/BaseAgent.ts
```diff
@@ -39,7 +39,17 @@ import { greenText } from './OutputClass'
...
+{"reqSignature":{"type":"ED25519","values":[{"from":"6WdiPynfLWAp67M9TtX5hR","value":"4BNUD13NrDx2ADAV13kyePTvbC1iJq5YWwAPDg42m1JSGQkJgacAHtdunD22sJKiiRuijhnL5HaYXD5qKgLZ2pfE"},{"from":"MjjBAYk1ef7PGdXAx7zR5J","value":"4hLK4N4esjaMRRgViAfjQ5FWijM5hkLdurmFh6qPTrTvHQhkgjAQw8qzkYHD5QshRKgXtLut3icJtXQ9C7PNrmEk"}]},"txn":{"data":{"data":{"alias":"Node4","services":[]},"dest":"4PS3EDQ3dW1tci1Bp6543CfuuebjFrg36kLAUcskGfaA"},"metadata":{"digest":"9c8e078bab9179d53d236906f9bc500c89dfaba9db628ddebc688e7d8a6b3914","from":"6WdiPynfLWAp67M9TtX5hR","payloadDigest":"c39040f5ca85d04d21419ac895c0eafaa0168095302ecd780d35da46bf9af92b","reqId":1758282424273905271},"protocolVersion":2,"type":"0"},"txnMetadata":{"seqNo":14,"txnTime":1758286183},"ver":"1"}`
```
genaris: Are these changes actually needed?

demo/src/BaseAgent.ts
```diff
@@ -48,16 +58,20 @@ export const indyNetworkConfig = {
   connectOnStartup: true,
 } satisfies IndyVdrPoolConfig
 
+const useDidCommV2 = process.env.DIDCOMM_V2 === 'true'
+
```
genaris: Why don't we just call this "didcommVersion" (environment variable DIDCOMM_VERSION) and make the demo to work exclusively on a particular version of DIDComm? This way we can pass either "v1" or "v2" to didcommVersions array on DidCommModuleConfig

demo/src/BaseAgent.ts
```diff
@@ -70,30 +84,39 @@ export class BaseAgent {
...
   return {
     didcomm: new DidCommModule({
       ...didcommConfig,
+      ...(useDidCommV2 && {
+        didcommVersions: ['v1', 'v2'],
+      }),
+      basicMessages: { protocols: ['v1', 'v2'] },
       connections: {
         autoAcceptConnections: true,
+        ...(useDidCommV2 && { autoCreateConnectionOnFirstMessage: true }),
```
genaris: Is there any problem to enable this flag even if we support v1 only?

packages/core/src/crypto/JwsService.ts
```diff
@@ -265,9 +265,9 @@ export class JwsService {
     }
   ) {
     const { jwsSigner } = options
-    const x509ModuleConfig = agentContext.dependencyManager.resolve(X509ModuleConfig)
 
     if (jwsSigner.method === 'x5c') {
+      const x509ModuleConfig = agentContext.dependencyManager.resolve(X509ModuleConfig)
```
genaris, line 270: Is this an intended change? Seems unrelated to this work

packages/core/src/modules/dids/methods/peer/PeerDidRegistrar.ts
```diff
@@ -129,16 +129,18 @@ export class PeerDidRegistrar implements DidRegistrar {
...
       const didRecord = new DidRecord({
         did,
         role: DidDocumentRole.Created,
         didDocument: isPeerDidNumAlgo1CreateOptions(options) ? didDocument : undefined,
         keys,
         tags: {
-          // We need to save the recipientKeys, so we can find the associated did
-          // of a key when we receive a message from another connection.
-          recipientKeyFingerprints: didDocument.recipientKeys.map((key) => key.fingerprint),
+          recipientKeyFingerprints,
```
genaris: What's the change introduced here? We are doing the same but just defining the constant before, right? (maybe that's a not fully reverted change?)

packages/didcomm/src/modules/basic-messages/repository/DidCommBasicMessageRecord.ts
```diff
@@ -22,6 +23,8 @@ export interface DidCommBasicMessageStorageProps {
   parentThreadId?: string
   content: string
   sentTime: string
+  /** Protocol version ('v1' or 'v2). Optional for backwards compatibility. */
+  protocolVersion?: DidCommVersion
```
genaris: I think DidCommVersion is not the appropriate type, as it is not strictly related to DIDComm version but Basic Messages protocol version (which currently can be v1 or v2. Probably the simplest approach would be to just set it to 'string' like we do in Proofs and Credential Exchanges

packages/didcomm/src/modules/basic-messages/services/DidCommBasicMessageService.ts
genaris, line 1: Why not split this Service class in two, so everything related to V1 remains at `DidCommBasicMessageService` and the different methods for V2 are at a `DidCommBasicMessageV2Service`? The few shared logic between both can be in a BaseDidCommMessageService class, much in the same way we do in other modules. 

Note that I'd still call them "Service" instead of "Protocol" just to maintain naming compatibility with the current codebase.

packages/didcomm/src/modules/basic-messages/DidCommBasicMessagesModuleConfig.ts
```diff
@@ -0,0 +1,37 @@
...
+ */
+export interface DidCommBasicMessagesModuleConfigOptions {
+  /**
+   * Basic message protocol versions to enable.
+   * - 1.0: Original BasicMessage (https://didcomm.org/basicmessage/1.0)
+   * - 2.0: BasicMessage 2.0 (https://didcomm.org/basicmessage/2.0), works over both DIDComm v1 and v2
+   *
+   * @default ['v1']
+   */
+  protocols?: DidCommVersion[]
```
genaris: Same as for the record: use either string or restrict it to 'v1' | 'v2' (unless you want to use dynamic typing like we do for other modules, e.g. Message Pickup, which is certainly more robust)

packages/didcomm/src/modules/basic-messages/DidCommBasicMessagesModule.ts
```diff
@@ -28,13 +34,24 @@ export class DidCommBasicMessagesModule implements Module {
     const messageHandlerRegistry = agentContext.resolve(DidCommMessageHandlerRegistry)
     const basicMessageService = agentContext.resolve(DidCommBasicMessageService)
 
-    messageHandlerRegistry.registerMessageHandler(new DidCommBasicMessageHandler(basicMessageService))
+    if (this.config.supportsV1) {
```
genaris: `supportsV1` and `supportsV2` flags on config seems to me unnecessary for what they are being actually used (e.g. just read protocols and that's it). But that being said, I think we should register Basic Message 2.0 protocol only in case DIDCommModuleConfig.didCommVersions includes v2 (and the same for Basic Message 1.0 / v1).

Also we might look at the patterns followed in other modules, where we register the handlers directly on the protocol/service classes.

packages/didcomm/src/modules/connections/handlers/DidCommDidExchangeResponseHandler.ts
```diff
@@ -58,8 +59,10 @@ export class DidCommDidExchangeResponseHandler implements DidCommMessageHandler
     }
 
     // Validate if recipient key is included in recipient keys of the did document resolved by
-    // connection record did
-    if (!ourDidDocument.recipientKeys.find((key) => key.fingerprint === recipientKey.fingerprint)) {
+    // connection record did. The compare bridges Ed25519 <-> X25519 (legacy v1) and matches
+    // same-curve keys directly for v2 X25519 / P-256.
```
genaris: Is this used for anything under DIDComm v2? I guess we won't reach it, since DID Exchange protocol only works in v1.

packages/didcomm/src/modules/routing/repository/DidCommMediationRecord.ts
```diff
@@ -14,6 +15,9 @@ export interface DidCommMediationRecordProps {
   endpoint?: string
   recipientKeys?: string[]
   routingKeys?: string[]
+  routingDid?: string
+  recipientDids?: string[]
+  mediationProtocolVersion?: DidCommVersion
```
genaris:
```suggestion
  protocolVersion?: string
```

MediationRecord is about Coordinate Mediation, so the mediation prefix seems redundant to me. And also it's independent from didcomm version, as in a future we could have a coordinate mediation 3.0 that might work for DIDComm V2.

I think it is a bit unfortunate to have different keys for each, but probably it's a trade-off if we don't want to introduce storage breaking changes now. Maybe for next major version we should think on refactoring the whole mediation module and make it easier to maintain.

packages/didcomm/src/modules/message-pickup/protocol/v3/handlers/DidCommDeliveryRequestV3Handler.ts
genaris, line 1: Regarding the usage of Message Pickup V3, I'm wondering why are we stuck on it instead of going directly to V4. We discussed about it some time ago, and V3 was implemented for some compatibility with existing mediators... but it turns out that such existing mediators are not compatible because of some did:peer:2 limitations (they are using an oudated version of the spec that produces different ids for services in DID Documents). @tarunvadde can you confirm that was the reason?

samples/mediator.ts
```diff
@@ -10,6 +10,9 @@
  * 'mediatorConnectionsInvite' parameter in the agent config to the
  * url that is returned by the '/invitation/ endpoint. This will connect
  * to the mediator, request mediation and set the mediator as default.
+ *
+ * Mediation 2.0 + Pickup 3.0: Set MEDIATION_V2=true to enable. For v2,
```
genaris: Why don't we just use "DIDCOMM_VERSION" like in other demos/samples/tests/etc and directly choose a single DIDComm version based on it?

genaris (commented):
@tarunvadde I left some additional comments, mostly regarding storage and configuration, and maybe some refactoring to make the code a bit cleaner.

I still need to analyze a bit more the hardest part (key/mediation routing management) but I think we are pretty close to have a reviewable/mergeable PR.

packages/didcomm/src/modules/message-pickup/protocol/v3/handlers/DidCommDeliveryRequestV3Handler.ts
tarunvadde: Yes, we started with implementing Message Pickup V3 to be compatible with others (like RootsID) but later realised that the others still rely on older didcomm libraries and did:peer:2 limitation (key naming issue multibase vs key-N at credo). I honestly didn't think of V4 but yes we could definitely do that.

packages/core/src/crypto/JwsService.ts
tarunvadde: Yes, intended, from #2815: resolve(X509ModuleConfig) outside the x5c branch throws TypeInfo not known in minimal agent contexts, which the v2 signed-message tests hit.

rmlearney-digicatapult:
Oh no. I don't know what happened there. Was using interactive editor to fix merge conflicts :(

rmlearney-digicatapult:
**Critical DIDComm v2.1 interoperability and trust-boundary findings**
This review identifies three connected defects in the DIDComm v2 implementation that should be addressed together before merge.

**1. Internal KMS identifiers are sent on the DIDComm wire**
Outbound `authcrypt` envelopes emit local Askar KMS UUIDs as protected-header `skid` and recipient-header `kid`, rather than DID URLs identifying `keyAgreement` verification methods.

* `skid` is a local KMS UUID, not a fragment-bearing verification-method DID URL.
* Recipient `kid` is likewise a local KMS UUID.
* The emitted identifiers contain no `#fragment`.
* The wire values equal the local sender/recipient KMS key IDs.

This violates DIDComm v2.1’s identifier requirements and security boundaries. Existing single-agent round-trip tests do not expose this because the same KMS both emits and resolves the identifier.

`apu` and `apv` are correctly derived from the emitted header values, but that does not make the emitted UUIDs valid protocol identifiers.

_Required direction:_ preserve KMS key IDs strictly for local crypto operations and carry wire `kid/skid` as distinct, fragment-bearing DID URLs. Prefer the authoritative verification-method ID from DID-document metadata; use a `did:key#fragment` fallback only when no DID metadata exists.

**2. Receive-side resolution accepts non-protocol identifiers**
The receive path contains a fallback that can resolve a remote `kid` through local KMS state. A KMS identifier received from another agent has no cross-agent meaning and must never be treated as a DIDComm key identifier.

This combines with the outbound UUID behavior to make implementations appear functional only in same-agent tests, while failing when keys are owned by independent agents or another implementation.

_Required direction:_ require wire identifiers to be resolvable DID URLs for keyAgreement verification methods. Remove the fallback that accepts a non-DID `kid` as a local key reference, and reject fragment-less `skid` values instead of selecting a default `keyAgreement` entry.

**3. Plaintext addressing is not bound to authenticated envelope identities**
After decryption, plaintext from/to values can reach connection lookup and message processing without being checked against the authenticated envelope identities.

This is a trust-boundary problem:

* For authcrypt, plaintext from must match the DID represented by the authenticated sender `skid`; mismatch must fail.
* The authenticated sender key must be authorized by that DID for the appropriate purpose.
* For a nested signed plaintext, the inner signer must not assert a different sender from the outer authcrypt envelope.
* For anoncrypt, plaintext from is untrusted metadata and must not be used for connection lookup, authorization, auto-creation, or `from_prior` processing.
* A plaintext `t`o mismatch should produce a warning rather than rejection, consistent with the non-fatal DIDComm requirement.
* Additional conformance gap: `apu` fallback
* Authcrypt envelopes lacking `skid` are currently rejected. DIDComm v2.1 requires recovering the sender identifier from `apu`, which is the `base64url` encoding of `skid`.

Do not rely on existing single-agent round-trip tests as proof of interoperability. Add regression coverage with two independent agents/KMS instances for P-256, P-384, and X25519, plus focused coverage for:

* fragment-bearing DID URL `kid/skid` emission;
* rejection of KMS UUIDs and fragment-less identifiers on receive;
* `apu` fallback when `skid` is absent;
* authcrypt from binding and anoncrypt sender exclusion;
* successful cross-agent pack/unpack using DID-document verification-method identifiers.

rmlearney-digicatapult:
**Node KMS ECDH-1PU implementation cannot work across the supported Node range**

**File:** `packages/node/src/kms/crypto/deriveKey.ts`

- `createECDH('x25519')` (L113, L120, L233) is the wrong Node API for X25519 and throws `ERR_CRYPTO_INVALID_CURVE`. This affects the branch's supported Node range (`node >=20.19`), including Node 20, 22, and 24. X25519 must use `KeyObject`s with `crypto.diffieHellman`.
- `apu` and `apv` are ignored; PartyUInfo and PartyVInfo are zero-length (L129–130).
- The key-encryption key is derived before content encryption and without the tag. This does not follow draft-madden-04 §2.3 or the DIDComm requirement to use the tag.
- `ephemeralKeyId` is ignored and a new ephemeral key is generated. The `epk` actually used therefore differs from the `epk` the envelope service writes into the protected header.
- The content encryption key is 32 bytes (L137). A256CBC-HS512, the mandatory authcrypt algorithm, needs 64 bytes.
- Only X25519 is handled; P-256 and P-384 are not.

`packages/didcomm/src/v2/__tests__/DidCommV2EnvelopeService.test.ts` hides the problem by falling back to a pass-through mock KMS when `createECDH('x25519')` throws. This is therefore a Node backend implementation defect, not a version-specific Node regression. The supported replacement is Node's `KeyObject`-based X25519 support through `crypto.diffieHellman`.

rmlearney-digicatapult:
**Forward chain is reversed in rotate-to-nothing**

**File:** `packages/didcomm/src/modules/connections/services/DidCommDidRotateV2Service.ts` L303

`next = i === reversed.length - 1 ? recipientNext : new DidKey(reversed[i + 1]).did` reverses the hop order whenever there are two or more routing keys. `packages/didcomm/src/envelope/DidCommV2Envelope.ts` L173 (`wrapInForward`) gets this right, and its comment notes that single-routing-key tests cannot detect an inverted chain.

rmlearney-digicatapult:
**HTTP transport uses the v1 media type**

**Spec:** §Transports > HTTPS. The media type of the POST request MUST be the corresponding media type, e.g. `application/didcomm-encrypted+json`.

- **Outbound:** `DidCommHttpOutboundTransport` sends v2 envelopes with `Content-Type: application/didcomm-envelope-enc`. The default comes from `packages/didcomm/src/DidCommModuleConfig.ts` L246.
- **Inbound:** `packages/node/src/transport/DidCommHttpInboundTransport.ts` L16 accepts only the V0 and V1 media types. A spec-conformant v2 sender gets HTTP 415.

rmlearney-digicatapult:
**Plaintext conversion merges headers and body**

- **Receive:** `packages/didcomm/src/v2/normalize.ts` (`normalizeV2PlaintextToV1`) builds `{ '@type', '@id', ...body, ...rest }`.
  - A body key `@type` or `~thread` overrides the header-derived value.
  - Body fields called `from` or `to` are silently overwritten by the headers.
  - Unknown headers share a namespace with body fields.
  - `lang` is moved to `~l10n`, so `DidCommBasicMessageV2.lang` is lost on receive.
- **Send:** `packages/didcomm/src/v2/plaintextBuilder.ts` (`buildV2PlaintextFromMessage`) puts leftover v1 decorators (`~please_ack`, `~timing`, `~service`) into `body`. The spec says acknowledgement requests belong in the `please_ack` header.

rmlearney-digicatapult:
**Encryption**

- **Single recipient key.** Encryption goes to `recipientKeys[0]` only (`DidCommV2Envelope.ts` L122), and `apv` is built from that one kid (`DidCommV2EnvelopeService.ts` L91). §Key IDs and §DID Document Keys say the default recipients SHOULD be all of the recipient's `keyAgreement` keys, with a separate envelope per key type.
- **`skid` fallback from `apu`.** A missing `skid` is rejected (`DidCommV2EnvelopeService.ts` L331–334; `DidCommV2Envelope.ts` L236). §ECDH-1PU key wrapping says authcrypt implementations MUST be able to resolve the sender kid from `apu` when `skid` is not set.
- **`typ` handling.** Envelope detection requires exactly `typ === 'application/didcomm-encrypted+json'` or `'application/didcomm-signed+json'` (`packages/didcomm/src/util/didcommVersion.ts` L37, L63). `typ` is only SHOULD, so a missing value should not cause rejection. A media type without `/` MUST be treated as having the `application/` prefix.
- **`kid` and `skid` form.** Bare `did:key` values without a fragment, and raw KMS key IDs, are emitted and accepted (`DidCommV2Envelope.ts` L125, L175; `resolveV2Keys.ts` L113–133, L324). §Key IDs says the `kid` MUST be a DID URL that points to a verification method in the `keyAgreement` relationship.
- **Ed25519 as a sender key.** The Ed25519 sender key is converted to X25519 at unpack (`DidCommV2EnvelopeService.ts` L353–355), and the v2 service resolver accepts Ed25519 keys listed under `keyAgreement` (`DidCommDocumentService.ts` L310).
- **Content encryption with authcrypt.** Unpack accepts `ECDH-1PU+A256KW` combined with `A256GCM` (`DidCommV2EnvelopeService.ts` L305, L327). The spec's algorithm table lists A256GCM for anoncrypt only.
- **Nested anoncrypt(authcrypt).** Receiving anoncrypt(authcrypt(plaintext)) is not handled; only a nested JWS is detected after decryption. This combination is listed in the spec for §Protecting the Sender Identity.
- **NIST curve point validation.** No explicit on-curve check of `epk`; the implementation relies on the KMS key import (Askar validates on import). Adding an explicit check would be prudent.
- **Ephemeral keys stored in the KMS.** Ephemeral keys are persisted with `kms.createKey` and then deleted for every message. This is not a conformance issue, but it costs performance and can leave keys behind if the process crashes between the two calls.

rmlearney-digicatapult:
**Plaintext headers**

- **Connectionless authcrypt has no `from`.** `from` and `to` are filled in only when a connection exists (`DidCommV2Envelope.ts` L148–152). §Message Headers says `from` is REQUIRED for authcrypt.
- **`to` on received messages.** The recipient never checks that its own identifier is in `to` (SHOULD warn).

rmlearney-digicatapult:
**Signed messages**

- **`to` not enforced.** Signing does not require `to`. §DIDComm Signed Messages says it SHOULD be present for standalone signed messages, and MUST be present when the message is also encrypted.
- **Flattened JWS not supported.** Only General serialization is parsed (`types.ts`; `didcommVersion.ts` L51–69). §Construction says recipients MUST be able to process both forms.
- **No `from`, no check.** `verifySignedMessage` skips the signer-to-`from` check when `from` is absent (`DidCommV2EnvelopeService.ts` L495). When `from` is present, the specification requires it to match the signer's `kid`; absence of `from` is not by itself a universal MUST violation for a standalone signed message, but signed-and-encrypted messages still require `to`.
- **`kid` not integrity-protected.** The `kid` is emitted only in the unprotected per-signature header (L235–247) for SICPA interop. The spec example places it in the protected header.
- **No send path.** Sending signed messages is not wired into `DidCommMessageSender`; the envelope service methods exist but nothing calls them.

rmlearney-digicatapult:
**DID rotation**

- **Invalid `from_prior`.** A bad rotation JWT is logged and the message is processed anyway (`DidCommDidRotateV2Service.ts` L176–180).
- **Signing key assumed.** `createFromPrior` hard-codes `alg: EdDSA` and `crv: Ed25519` (L452–455), which fails for authentication keys on other curves.
- **No `iat` check.** `iat` is required but not compared with previously accepted rotations, so an old rotation JWT can be replayed.

`from_prior` issuer is not bound to the connection

**Spec:** §DID Rotation. The rotating party includes `from_prior` until it receives a message addressed to the new DID. During that transition, the rotation JWT is used to extract the prior DID (`iss`), and the recipient associates the message with the known sender. After the new DID is acknowledged, further `from_prior` processing is not necessary and the header may be ignored.

**Implementation:** The main transmission and acknowledgement lifecycle is already implemented:

- `rotateOurDid` stores pending rotation metadata and outbound v2 messages attach the pending `from_prior`;
- `processFromPrior` verifies the JWT and updates the peer DID;
- `clearPendingRotationIfAcknowledged` clears the pending JWT after receiving a message addressed to the new DID;
- tests cover normal rotation, peer acknowledgement, and multi-use OOB rotation.

The remaining security gap is that `processFromPrior` verifies the JWT against whichever DID appears in `iss`, but does not require `iss` to equal the connection's current or accepted prior `theirDid`. Invalid JWTs are logged and ignored while the message continues through normal processing. The receiver also continues to verify supplied headers after acknowledgement instead of explicitly treating stale `from_prior` as ignorable.

**Impact:** an attacker can send a message with `from = Bob` and a self-signed `from_prior` whose `iss` is an attacker-controlled DID. A no-`sub` payload can then trigger `processRotateToNothing` and terminate Alice's connection with Bob.

rmlearney-digicatapult:
**Out-of-band 2.0**

- **`pthid` must be preserved on every OOB response path.** The normal OOB outbound path sets the parent thread ID to the invitation ID (`getDidCommOutboundMessageContext.ts`, `addExchangeDataToMessage`), and the receive path validates and sets the `pthid` (`DidCommOutOfBandApi.ts`, `ensureParentThreadId`). The specification states that the invitation `id` MUST be used as the response `pthid`; implicit and connectionless paths must be verified separately to ensure they do not bypass this requirement.
- **`accept` dropped.** `accept` is not carried into the unified invitation record (`packages/didcomm/src/modules/oob/converters.ts`, `convertV2InvitationToOutOfBandInvitation`).
- **Internal attachment conversion is acceptable, but v2 wire behavior should be completed.** Mapping v2 invitation attachments into Credo's existing internal v1-shaped attachment representation is acceptable for current internal processing. The implementation should nevertheless preserve the v2 attachment shape when sending or re-emitting v2 invitations and attached v2 protocol messages, including v2 fields such as `id`, `media_type`, and `data`.
- **`_oobid` short URLs.** Not handled specifically; only the generic short-URL fetch applies.
- **`goal` as label.** `goal` is stored as `theirLabel` (`DidCommConnectionsApi.ts` L131).
- **Default version.** v2 is chosen by default only when there are no messages, attachments, handshake protocols or routing (`DidCommOutOfBandApi.ts`, `createInvitation`).

Auto-creating a connection trusts an unauthenticated `from`

**Implementation:** `packages/didcomm/src/DidCommMessageReceiver.ts` L341–390. When `autoCreateConnectionOnFirstMessage` is enabled (v2 OOB needs it), a `Completed` connection is created with `theirDid = from`. There is no proof that the sender controls that DID.

**Impact:**

- The implicit branch (L373) fires for any DID the agent has created, not only invitation DIDs. Each message also triggers a DID rotation and mediator keylist registration (`rotateInviterDidForV2OOB`), so unauthenticated traffic can exhaust resources.
- `packages/didcomm/src/modules/oob/DidCommOutOfBandService.ts` L255–266 (`findCreatedByRecipientDid`) does not check the OOB record's state, so a single-use invitation can be accepted more than once.

PR #2729, opened 2026-04-01
feat(core): X509 CRL verification

TimoGlastra:
Adds support for X509 CRL verification. Also improves the extension validation a bit more, and takes into account critical extensions. It also updates to align a bit more with base certificate requirements on which extensions should be critical.

I'm not 100% sure yet on the reason for revocation checking. If no reason is defined it will just match any distribution point, but you may want to check for a specific reason in some cases?

Still need to write tests, and ensure everything is working as intended


hacdias:
Superseded by #2816

PR #2749, opened 2026-04-10
fix(didcomm): v1<=>v2 interop with CM 2.0 mediators

tarunvaddeSoul:
## Summary

Fixes cross-version interop when a Coordinate Mediation 2.0 mediator is in the path. Builds on #2718.

## Changes

- Split v1/v2 peer DID creation: v1 gets plain URL + Ed25519 routing keys, v2 keeps the routing DID as service URI
- Register both Ed25519 and X25519 did:key forms of the recipient key in keylist-update so v1 and v2 senders both match
- Sender always uses X25519 did:key as Forward `next` (removed peer DID fallback)
- Dedup routing keys by X25519 fingerprint to prevent v2 Forward double wrapping
- Path 3 fallback in DidCommV2KeyResolver to look up did:key kids in DidCommMediatorRoutingRepository
- Include Ed25519 auth VM in auto-generated mediator routing did:peer:2
- Wrap MessagesReceivedV3Message in outbound context so MP 3.0 delivery acks get sent (parity with v2 handler)
- Select BM 1.0/2.0 based on connection.didcommVersion
- Drop recipient_did filter from MP 3.0 status-request
- TypedArrayEncoder.fromBase64 accepts base64url variants

## Test plan

Verified end to end against the 2060.io didcomm-mediator and Hologram:

- [x] v1 Faber demo => v2 mediated Hologram
- [x] v2 Faber demo => v2 mediated Hologram
- [x] Hologram <=> 2060.io mediator (CM 2.0 provisioning, keylist-update, MP 3.0 live delivery, ack roundtrip)
- [x] Hologram <=> Hologram welcome service (v2 <=> v2 mediated)
- [x] Existing routing module tests pass (37/37)


packages/didcomm/src/modules/routing/DidCommMediationRecipientApi.ts
```diff
@@ -290,10 +290,17 @@ export class DidCommMediationRecipientApi {
...
+    //
+    // Per CM 2.0 / MP 3.0, `recipient_did` in status-request is optional — when omitted
+    // the mediator returns messages for the entire connection. We rely on that because
+    // the mediation record may contain many registered DIDs (the holder's peer DID from
+    // provisioning, plus a did:key per downstream connection registered via keylist-update).
+    // Filtering by any single one (e.g. `recipientDids[0]`) would hide messages queued
+    // under a different recipient DID (e.g. a forward whose `next` is a per-connection
+    // did:key). Fetching all pending messages for the holder's mediator connection is
+    // the correct semantic.
+    const recipientDidV3 = undefined
```
genaris: How is this different from our current Pickup V2 implementation, that needs this long explanation?

I think you can just not define any `recipientDidV3` constant at all, and call pickupMessages without it.

packages/didcomm/src/modules/basic-messages/DidCommBasicMessagesApi.ts
```diff
@@ -41,10 +41,15 @@ export class DidCommBasicMessagesApi {
...
 
-    const { message: basicMessage, record: basicMessageRecord } =
-      this.config.supportsV2
-        ? await this.basicMessageService.createMessageV2(this.agentContext, message, connection, parentThreadId)
-        : await this.basicMessageService.createMessage(this.agentContext, message, connection, parentThreadId)
+    // Select BM protocol version based on the connection's DIDComm envelope version:
+    //   - v2 connection => BM 2.0 (only if config supports it)
+    //   - v1 connection => BM 1.0 (legacy interop, even if config includes v2)
+    // This ensures we don't send BM 2.0 to a peer that only speaks v1.
+    const useBmV2 = this.config.supportsV2 && (connection.didcommVersion ?? 'v1') === 'v2'
```
genaris: Wouldn't it be simpler ?:


```suggestion
    const useBmV2 = this.config.supportsV2 && connection.didcommVersion === 'v2'
```

genaris (commented):
Just a few comments for the moment. It will need some more deep review but maybe we can continue it on the main PR (where there are already some outstanding comments).

packages/didcomm/src/modules/routing/DidCommMediationRecipientApi.ts
tarunvaddeSoul: Done

genaris (approved):
Continuing the review on #2704

PR #2754, opened 2026-04-14
ci: run tests on node 22 and 24 (drop 20)

genaris:


hacdias (approved):
LGTM. Someone needs to change the required tests on the Settings of the repository to match the new versions. I don't have permissions for that.

berendsliedrecht:
> LGTM. Someone needs to change the required tests on the Settings of the repository to match the new versions. I don't have permissions for that.

Oh same... Maybe @ryjones could change this? We have some required tests which we removed, all the node 20 tests. And now we should require all node 24 tests.

ryjones:
OK, I've disabled node 20

ryjones:
24 is now required

PR #2760, opened 2026-04-22
fix(openid4vc): preserve optional issuer metadata fields on update

Sshovon:
**updateIssuerMetadata** previously overwrote all optional issuer metadata properties even when they were omitted from the input, which could unintentionally erase existing values in the issuer record. 

This change makes the method safer and more intuitive by treating optional properties as patch-style updates: **display**, **dpopSigningAlgValuesSupported**, **batchCredentialIssuance**, and **authorizationServerConfigs** are now updated only when explicitly provided, and otherwise remain unchanged.


packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerApi.ts
```diff
@@ -61,10 +61,18 @@ export class OpenId4VcIssuerApi {
     const issuer = await this.openId4VcIssuerService.getIssuerByIssuerId(this.agentContext, issuerId)
 
     issuer.credentialConfigurationsSupported = credentialConfigurationsSupported
-    issuer.display = display
-    issuer.dpopSigningAlgValuesSupported = dpopSigningAlgValuesSupported
-    issuer.batchCredentialIssuance = batchCredentialIssuance
-    issuer.authorizationServerConfigs = authorizationServerConfigs
+    if (display !== undefined) {
+      issuer.display = display
+    }
```
hacdias, line 77: I'm under the impression this was something we wanted, so that's nice!

---

I think implementing it like this won't allow you to "remove" a field anymore. What if I want to remove the `display`? Maybe we should check if the key "display" exists and then set it to that value?

Same applies for the other options.

I'd also add a changelog entry since this changes the behavior. Maybe also add a comment to the function `updateIssuerMetadata` to explain what the behavior is.

Sshovon: Thanks for the feedback, @hacdias ,  great point about removing existing attributes. I’ve pushed updates so optional fields are now handled based on key presence in the options object (so omitted keys are preserved, and explicitly provided undefined clears the value). Could you please take another look?

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerApi.ts
```diff
@@ -61,10 +68,18 @@ export class OpenId4VcIssuerApi {
     const issuer = await this.openId4VcIssuerService.getIssuerByIssuerId(this.agentContext, issuerId)
 
     issuer.credentialConfigurationsSupported = credentialConfigurationsSupported
-    issuer.display = display
-    issuer.dpopSigningAlgValuesSupported = dpopSigningAlgValuesSupported
-    issuer.batchCredentialIssuance = batchCredentialIssuance
-    issuer.authorizationServerConfigs = authorizationServerConfigs
+    if ('display' in options) {
```
berendsliedrecht: We can make this assignment a bit shorter :)

```typescript
issuer.display ??= display
```

etc. This only assigns a new value if the value is undefined.

Sshovon: @berendsliedrecht   `??=` only assigns when the existing property is **nullish**. In this update function we want to assign when the option is explicitly provided, even if the issuer already has a value. So  I think `??=` would change the behavior here.

Sshovon: With current logic: 
```
const issuer = {
  display: ['Old Display'],
}
const options = {
  display: ['New Display'],
}
// Output: ['New Display']
```
With `??=` operator
```
const issuer = {
  display: ['Old Display'],
}

const options = {
  display: ['New Display'],
}
// Output: ['Old Display']
```

TimoGlastra:
I think this PR makes sense, but we don't take this approach elsewhere i think, and I'm not sure treating the following different is safer and more intuitive:

```
const options = {
 display: undefined
}

const options2 = {

}
```

In both cases `options.display` resolves to `undefined` and it's very uncommon to have different logic based on whether the key exists or not (AFAIK). I think if you pass undefined keeping the same value makes sense.

Would it be an option to handle this with `null` if you explicitly want to unset a value? 

In Prisma it also works that with an update:
- undefined means no change
- null means set it to null/undefined
- value means set it to the value

Sshovon:
Messed up something here I think...will close this one and create a clean one.

PR #2761, opened 2026-04-22
fix(didcomm): tolerate base64url in signed attachment payloads

genaris:
`data.base64` in signed attachments is spec'd as standard base64, but peers routinely emit base64url (padded or not). Earlier Credo versions, AFJ 0.5.x, ACA-Py. Credo 0.7's `TypedArrayEncoder.fromBase64` uses `@scure/base` which is strict, so DID Exchange responses (and any other signed-attachment flow) fail with:

```
Could not decode data from base64 string
  cause: padding: invalid, string should have whole number of bytes
```

So here we are creating some new helpers and updates in `DidCommAttachment` that do strict base64 first, fall back to base64url (stripping `=` to cover both padded and unpadded). DidExchangeProtocol makes use of this logic.

### DIFF
--- a/packages/didcomm/src/decorators/attachment/DidCommAttachment.ts
+++ b/packages/didcomm/src/decorators/attachment/DidCommAttachment.ts
@@ -138,27 +138,25 @@ export class DidCommAttachment {
   /*
    * Helper function returning the raw bytes of a base64 attachment payload.
    *
-   * Although the DIDComm specification mandates standard base64 for `data.base64`,
-   * many peers (including earlier Credo versions) emit the payload using the
-   * base64url alphabet, with or without padding. To preserve interoperability we
-   * accept both alphabets here — this is the only place such leniency lives, so
-   * the underlying `TypedArrayEncoder.fromBase64` can remain strictly spec-compliant
-   * for every other caller.
+   * The DIDComm/Aries RFC 0017 spec mandates base64url for `data.base64`, so we try
+   * decoding it as base64url first. Credo itself (and other implementations) has
+   * historically emitted standard base64 with `+`/`/` and `=` padding, so we fall
+   * back to that to preserve interop with older agents.
    */
   public getDataAsUint8Array(): Uint8Array {
     if (typeof this.data.base64 !== 'string') {
       throw new CredoError('No base64 attachment data found.')
     }
     try {
-      return TypedArrayEncoder.fromBase64(this.data.base64)
-    } catch (strictError) {
+      // `fromBase64Url` expects the no-padding variant, strip trailing `=` so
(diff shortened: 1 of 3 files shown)


berendsliedrecht:
Wait isn't it supposed to be base64url for data.base64 (i know confusing...)

- https://github.com/decentralized-identity/aries-rfcs/blob/main/concepts/0017-attachments/README.md#base64url
- https://identity.foundation/didcomm-messaging/spec/#attachments

I want to change it to base64url, but @TimoGlastra preferred not to, to keep better interoperability.

berendsliedrecht (approved):

genaris:
> Wait isn't it supposed to be base64url for data.base64 (i know confusing...)
> 
> * https://github.com/decentralized-identity/aries-rfcs/blob/main/concepts/0017-attachments/README.md#base64url
> * https://identity.foundation/didcomm-messaging/spec/#attachments
> 
> I want to change it to base64url, but @TimoGlastra preferred not to, to keep better interoperability.

The problem is that with the current code as it is right now on main, I have problems when trying to make it interact with a mediator using credo 0.5.17, so it already has an interop issue.

But you have a good point. Maybe we should first check for base64url and use base64 as a fallback? I'll explore this problem a bit more in detail to see what is actually going on here.

genaris:
@berendsliedrecht initially I thought to take the same approach as we did in feat/didcomm-v2 branch, by modifying the [TypedArrayEncoder](https://github.com/openwallet-foundation/credo-ts/blob/c12e7df22223dc86238712db5831c2971bec5e06/packages/core/src/utils/TypedArrayEncoder.ts#L27) to accept both, which seems to be actually the same behaviour as we had prior to upgrading to `@scure/base` (as Node's `Buffer.from(x, 'base64')` seems to accept both the standard and URL-safe alphabets, tolerating these missing padding/ignoring whitespaces/newlines. 

So maybe it would be better to treat it as a regression of TypedArrayEncoder (and fix it to match the previous behaviour) rather than a specific case for DIDComm attachments. What do you think?

berendsliedrecht:
I'd prefer to keep `fromBase64` and `fromBase64Url` separate in the TypedArrayEncoder. I think its good to first try the spec-defined encoding and fallback if we know it deviates. I would rather be more explicit then having a function that does more.

genaris:
> I'd prefer to keep `fromBase64` and `fromBase64Url` separate in the TypedArrayEncoder. I think its good to first try the spec-defined encoding and fallback if we know it deviates. I would rather be more explicit then having a function that does more.

Yeah, I agree. Better to keep it focused in DIDComm attachments and eventually remove the fallback once most agents are properly encoding them.

PR #2766, opened 2026-04-25
fix(didcomm): downgrade v3 pickup strategy on CM 1.0 mediators

tarunvadde:
## Summary

Pickup v1/v2 was already auto-upgraded to v3 for CM 2.0 mediators, but the
reverse wasn't enforced - a globally-configured `PickUpV3*` strategy bled
into CM 1.0 mediators, causing failed pickups over v1 connections.

This PR adds the symmetric downgrade (v3 => v2 for CM 1.0) with a `logger.warn`,
in both `getPickupStrategyForMediator` (config path) and `initiateMessagePickup`
(explicit-strategy path).


packages/didcomm/src/modules/routing/DidCommMediationRecipientApi.ts
```diff
@@ -288,6 +288,20 @@ export class DidCommMediationRecipientApi {
...
+        this.logger.warn(
+          `Pickup strategy 'PickUpV3' is not compatible with Coordinate Mediation 1.0. Downgrading to 'PickUpV2'.`
+        )
+        mediatorPickupStrategy = DidCommMediatorPickupStrategy.PickUpV2
+      } else if (mediatorPickupStrategy === DidCommMediatorPickupStrategy.PickUpV3LiveMode) {
+        this.logger.warn(
+          `Pickup strategy 'PickUpV3LiveMode' is not compatible with Coordinate Mediation 1.0. Downgrading to 'PickUpV2LiveMode'.`
+        )
+        mediatorPickupStrategy = DidCommMediatorPickupStrategy.PickUpV2LiveMode
+      }
```
genaris: I'm confused about all this logic. Why isn´t everything within `getPickupStrategyForMediator()`?

I'm sure there is a more simple way to handle all different cases and auto-correct if needed.

genaris (changes requested):

genaris (approved):
I still think this is too verbose (typical of AI-assisted code), but let's keep the discussion on the main PR.

PR #2768, opened 2026-04-27
fix(PEX): Update PEX to branch on claimFormat with runtime guards and fail-fast

rmlearney-digicatapult:
Current `presentationsToCreate.ts` mixes branching selection between `claimFormat` and runtime class checks (`credentialRecord instanceof`)

This PR unifies branching logic to use `credential.claimFormat` in a `switch` statement to disambiguate fall-through.

Runtime guarding by asserting record types in each case with fail-fast `CredoError` which reports the input descriptor identifier for easier tracing/debugging

Added unit tests for failure modes (incompatible `claimFormat` + record types)

Invalid combinations should now fail immediately & predictably. Selector is deterministic and failures surfaced early rather than propagating through to later VP creation logic.

This also allows easier extension to more supported claim types through new `case` statements + record type guard for that case.

Motivation: introducing `vc+sd-jwt` support over DIDComm


rmlearney-digicatapult:
Withdrawing this PR in favour of a more comprehensive PEX `2.1.1` spec alignment PR

PR #2769, opened 2026-04-28
fix(CI): Local DID document to avoid Indy ledger resolution in CI

rmlearney-digicatapult:
E2E (24) but not E2E (22) failed on recent PR due to setup race condition

Test harness seeds during container startup, E2E begins immediately with no check gate, E2E (24) races

`indy-vdr-anoncreds-registry` suite runs early, `helpers.ts` imports did:indy without `didDocument` forcing resolution through `DidsApi.ts` which fails in beforeAll - race condition

Later test `indy-vdr-did-register` used same seed/DID and passed, meaning DID was resolvable at this later stage

Fix changes `helpers.ts` to pass an in-memory didDocument


berendsliedrecht (approved):

rmlearney-digicatapult:
Haha that's fun - now failing due to Cheqd network resolution

PR #2770, opened 2026-04-29
chore(webvh): bump didwebvh-ts to 2.8.0

genaris:
This is an important one: solves a serious bug that made all WebVH DIDs created by previous versions of Credo non-conforman with the spec, making resolvers to fail. See https://github.com/decentralized-identity/didwebvh-ts/issues/93 for more details.

@berendsliedrecht since this is a forced breaking change, I think it is a good opportunity to merge before releasing 0.7.0.


berendsliedrecht:
@genaris I completely missed this! Is didwebvh-ts not semver compliant? They seem to have updated it as a patch?

TimoGlastra:
I think a fix can be pushed as patch (if the fix outweighs the breaking change)

TimoGlastra (approved):

TimoGlastra:
Should we just merge this, even though it will invalidate all previously created DIDs? Should we add some backwards compat so that we do allow to resolve these previously invalid ones?

brianorwhatever:
We have more changes that might possibly break existing DIDs. We also have some library breaking changes to align with DID resolution and the other implementations more. All for the sake of spec alignment and interoperability!

So we are considering releasing a 2.8.0 with the former which would continue to drop in here (other than potentially breaking some DIDs)

And then a 3.0.0 with library breaking changes which will take a tiny bit of work here.

So I would say hold off until 2.8.0 at least. Which should be in the next couple days

brianorwhatever:
Version 2.8.0 has been released and it is recommended to update this PR to that version

genaris:
> Version 2.8.0 has been released and it is recommended to update this PR to that version

Thanks @brianorwhatever , I've updated this PR to use this new version. BTW @TimoGlastra I've noticed that, after https://github.com/openwallet-foundation/credo-ts/pull/2848 , main is now targetting 0.8.0-alphas. Was this intended?

TimoGlastra:
Ooops, no. I will fix it tomorrow. There's no breaking changes.

TimoGlastra:
Don't we have any tests that now break with the updating hash calucation? Might be good to add some test vectors

PR #2774, opened 2026-05-01
feat(did): extend Dids API with findAllByQuery for tag search

rmlearney-digicatapult:
Without tags, querying DID records is currently O(n) requiring full retrieval and filtering

Added `findAllByQuery` to Dids API to reduce consumer tag search to O(1) + unit test

### DIFF
--- a/packages/core/src/modules/dids/DidsApi.ts
+++ b/packages/core/src/modules/dids/DidsApi.ts
@@ -2,12 +2,13 @@ import { parseDid } from '@sphereon/ssi-types'
 import { AgentContext } from '../../agent'
 import { CredoError, RecordNotFoundError } from '../../error'
 import { injectable } from '../../plugins'
+import type { Query, QueryOptions } from '../../storage/StorageService'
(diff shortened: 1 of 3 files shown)


genaris (commented):
Thanks @rmlearney-digicatapult !

I'm not against this improvement, but I'm not sure if adding tags to `getCreatedDids` is the cleanest way to achieve this, since I think this method is intentionally trying to 'hide' the complexities of tags by asking just a couple of them (method, did) and hard-coding others (role) in the query. 

Your change certainly works, as it will take the correct tags even if you pass different values in the `tags` parameter. But do you think that adding a more generic method to `DidsApi`, such as `findAllByQuery()` (which calls `repository.findByQuery`) could satisfy your needs? The good thing about it is  that it would be more aligned with the API of other modules (most of them already have the `findAllByQuery`).

rmlearney-digicatapult:
Ah yes that's a better idea. Have amended the PR.

genaris (approved):

PR #2776, opened 2026-05-04
fix(webvh): decouple webvh from anoncreds

rmlearney-digicatapult:
`webvh` currently couples generic `did:webvh` capabilities with the AnonCreds adapter (resource model, parsing, and resource URL handling). This makes core `webvh` reuse less clear to application logic.

Changes:

This PR decouples the `webvh` subsystem while preserving the delegation of method-specific semantics to `didwebvh-ts` to improve the user/dev experience.

The shared `AttestedResource` contracts are refactored into `webvh` core resources:
* `WebVhAttestedResource`
* `WebVhAttestedResourceProof`
* `WebVhAttestedResourceLink`
* `WebVhResourceIdParts`

alongside helpers:
* `isWebVhAttestedResource`
* `getResourceType` (permissive, returns `string` or `undefined`)
* `requireResourceType` (strictly returns `string`, throws on missing/invalid `metadata.resourceType`)
* `parseResourceId`

The refactor also exposes core log contract aliases from `didwebvh-ts` for downstream consumers:
* `WebVhDidLogEntry`
* `WebVhDidLog`

All of these are exported through explicit surfaces for `core`, `anoncreds`, and `resources`.

The AnonCreds registry/transform path has been updated to consume the shared core resource contracts/helpers instead of duplicating local shape logic.

Added boundary contract tests to verify compatibility with `didwebvh-ts` for log shape/state transitions, upstream round-trip resolution, and resource URL parsing alignment.

Result:

* `did:webvh` types and methods can be consumed independent of AnonCreds.
* AnonCreds remains a consumer adapter over generic `webvh` core behavior.

### DIFF
--- a/packages/webvh/package.json
+++ b/packages/webvh/package.json
@@ -1,8 +1,12 @@
 {
   "name": "@credo-ts/webvh",
   "exports": {
-    "cli": "./build/index.mjs",
-    "import": "./src/index.ts"
+    ".": {
+      "cli": "./build/index.mjs",
+      "import": "./src/index.ts"
+    },
+    "./core": "./src/core.ts",
+    "./anoncreds": "./src/anoncreds/index.ts"
   },
   "version": "0.7.0",
   "description": "Credo WebVH DID method implementation",
@@ -21,7 +25,17 @@
     "module": "./build/index.mjs",
     "exports": {
       ".": "./build/index.mjs",
-      "./package.json": "./package.json"
+      "./package.json": "./package.json",
+      "./core": {
+        "import": "./build/core.mjs",
+        "require": "./build/core.mjs",
+        "types": "./build/core.d.mts"
+      },
+      "./anoncreds": {
+        "import": "./build/anoncreds/index.mjs",
+        "require": "./build/anoncreds/index.mjs",
+        "types": "./build/anoncreds/index.d.mts"
+      }
     }
   },
   "scripts": {
--- a/packages/webvh/package.json
+++ b/packages/webvh/package.json
@@ -6,7 +6,8 @@
       "import": "./src/index.ts"
     },
     "./core": "./src/core.ts",
-    "./anoncreds": "./src/anoncreds/index.ts"
+    "./anoncreds": "./src/anoncreds/index.ts",
+    "./resources": "./src/resources/index.ts"
   },
   "version": "0.7.0",
   "description": "Credo WebVH DID method implementation",
@@ -35,6 +36,11 @@
         "import": "./build/anoncreds/index.mjs",
         "require": "./build/anoncreds/index.mjs",
         "types": "./build/anoncreds/index.d.mts"
+      },
+      "./resources": {
+        "import": "./build/resources/index.mjs",
+        "require": "./build/resources/index.mjs",
+        "types": "./build/resources/index.d.mts"
       }
     }
   },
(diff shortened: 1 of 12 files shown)


rmlearney-digicatapult:
CI is failing because `indy-pool*.dockerfile` use `snel/von-image:node-1.12-4-arm64` or `bcgovimages/von-image:node-1.12-6` which are built on the EOL Ubuntu Bionic 18.04 and `apt-get update` is failing to fetch `bionic/bionic-security` indexes

swcurran:
FYI @PatStLouis

packages/webvh/package.json
```diff
@@ -21,7 +26,22 @@
...
+      "./anoncreds": {
+        "import": "./build/anoncreds/index.mjs",
+        "require": "./build/anoncreds/index.mjs",
+        "types": "./build/anoncreds/index.d.mts"
+      },
+      "./resources": {
+        "import": "./build/resources/index.mjs",
+        "require": "./build/resources/index.mjs",
+        "types": "./build/resources/index.d.mts"
+      }
```
genaris, line 44: Leaving the exports like this means that we'll still be able to get all classes from `@credo-ts/webvh` and, at the same time, get the scoped ones under `@credo-ts/webvh/core`, `@credo-ts/webvh/anoncreds` and `@credo-ts/webvh/resources`, right?

It might be a bit awkward now, but it makes sense if we don't want to introduce breaking changes here.

packages/webvh/src/anoncreds/services/__tests__/WebVhTransform.test.ts
```diff
@@ -1,14 +1,14 @@
 import { JsonTransformer } from '@credo-ts/core'
 
-import { WebVhResource } from '../../utils/transform'
+import { WebVhAnonCredsResource } from '../../utils/transform'
```
genaris, line 3: From my understanding, this refactoring does not introduce a breaking change, because it was not exposed before. The only public method from an exported class was `verifyProof` from `WebVhAnonCredsRegistry` (which IMO is only public to make it easier to test), and even though, the shape of the resulting object seems to be compatible, so it should not represent any issue to merge before next major release.

genaris (approved):
I'd say this is more a refactor than a fix. I think it is a good step towards moving to scoped exports. As long as everything keeps working like before, I'm good to merge

PatStLouis:
@rmlearney-digicatapult I agree with the direction here. Let me provide you with context on where `AttestedResource` come from. One of the challenges we had is to maintain immutability in some posted resource, while allowing to edit some components of it. The best example is, once we publish a revocation registry, we may want to be able to add links to future published entries of this registry.

This requirement evolved from a specific anoncreds implementation with webvh to this `AttestedResource` type. The immutable portion gets placed in a specific `data` property, and the hash of this content is used to build the resource id. The rest of the document is protected with a `DataIntegrity` proof.

My only concern is how this will impact downstream application when updating credo (such as Bifold). Ideally there should be no impact.

packages/webvh/package.json
rmlearney-digicatapult: @genaris that's the intent to allow migration without breaking changes. You made me re-examine my code and I'd missed exporting the new scoped `resources` under global, now updated.

rmlearney-digicatapult:
@PatStLouis I don't think there should be any drift because the surface for AnonCreds has not migrated, all existing public methods are still exposed. The existing test harness was untouched and passed with all green :) 

One change is the renamed proof input type on `verifyProof` which now takes `WebVhAttestedResource` instead of `WebVhResources` but this is functionally equivalent for consumers using the registry normally.

However, there doesn't seem to be a specific `data` property - the content was always stored under `.content` on main and this branch and this contract remains unchanged.

packages/webvh/src/anoncreds/services/__tests__/WebVhTransform.test.ts
rmlearney-digicatapult: Thanks @genaris that was the goal - no breaking changes, only refactoring

PR #2777, opened 2026-05-04
feat(didcomm): support attachments in v2 OOB invitations

tarunvadde:
Adds `attachments` to `DidCommOutOfBandInvitationV2` per the DIDComm v2.1 OOB spec, and wires `messages` / `appendedAttachments` through `createV2Invitation` so one v2 OOB can carry a credential offer or proof request alongside the connection.

On receive, v2 attachments surface as v1 `requests~attach` via a shared `convertV2InvitationToOutOfBandInvitation` helper in `converters.ts`, so the existing dispatch path handles both wire formats uniformly. v1 paths untouched.


packages/core/src/utils/TypedArrayEncoder.ts
```diff
@@ -164,4 +164,4 @@ export class TypedArrayEncoder {
       throw new CredoError(`Could not check if array is typed array`, { cause: error })
     }
   }
-}
\ No newline at end of file
+}
```
genaris, line 167: Is this a whitespace change? Revert if if you can, since it could trigger style errors.

genaris (approved):

packages/core/src/utils/TypedArrayEncoder.ts
tarunvadde: This was actually added to fix the style error

PR #2779, opened 2026-05-06
fix(webvh): domain URI encoding

rmlearney-digicatapult:
Critical error which only appeared when testing the latest changes against implicit invitation flows for `did:webvh`

Credo builds `baseDid` from the caller input, while `didwebvh-ts` builds DID identifiers from `encodeURIComponent(domain)`. Calls from `credo-ts` to `didwebvh-ts` must be normalised at this boundary:

```ts
// existing credo-ts WebVhDidRegistrar.ts
const baseDid = `did:webvh:{SCID}:${domain}`

const { did, doc, log } = await createDID({
   domain,
   verificationMethods: [
      {
         controller: baseDid,
         type: 'Multikey',
         publicKeyMultibase,
      },
   ],
   // ...
})
```

```ts
// didwebvh-ts
const encodedDomain = encodeURIComponent(options.domain)
const computedController = `did:webvh:{SCID}:${encodedDomain}`

const didDocumentInput = {
   ...options,
   controller: computedController,
   verificationMethods: normalizedVerificationMethods,
}
```

*Credo:* `domain` is `alice:8443` -> `baseDid` `did:webvh:{SCID}:alice:8443` in `controller:`
*didWebVh-ts:* encodes and constructs `did:webvh:{SCID}:alice%3A8443`

This creates a mismatch in the verifier which expects both of these DIDs to be percent encoded and point to the same canonical DID

If `domain` is passed as `alice%3A8443`, credo keeps the percent-encoded `baseDid` but this is then double-encoded by `didwebvh-ts` into `alice%253A8443`

Fix:

Build `baseDid` with `encodeURIComponent(domain)` at the boundary and pass the original raw `domain` string into `createDID` to ensure exact match between `credo` and `didwebvh-ts` encodings

Added unit test to ensure correct encoding

### DIFF
--- a/packages/webvh/src/dids/WebVhDidRegistrar.ts
+++ b/packages/webvh/src/dids/WebVhDidRegistrar.ts
@@ -63,7 +63,7 @@ export class WebVhDidRegistrar implements DidRegistrar {
       const { publicKeyMultibase, keyId } = await this.generatePublicKey(agentContext)
       const signer = new WebVhDidCryptoSigner(agentContext, publicKeyMultibase, keyId)
       const verifier = new WebVhDidCrypto(agentContext)
-      const baseDid = `did:webvh:{SCID}:${domain}`
+      const baseDid = `did:webvh:{SCID}:${encodeURIComponent(domain)}`
 
       // Create DID
       const { did, doc, log } = await createDID({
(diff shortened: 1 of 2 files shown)


rmlearney-digicatapult:
Would welcome comments on this in case the preference is a fix in `didwebvh-ts` instead? Or if you expect people to pass in pre-encoded `domain`?

berendsliedrecht (approved):
IMO I am good with the changes, depending on what the preferred fix is of the didwebvh-ts maintainers.

genaris (approved):
Thanks, I also think it makes sense to add this here. I suggest to also propose this fix (or to add a new parameter such.`port` or similar) in the didwebvh-ts library as well.

PR #2784, opened 2026-05-07
fix(openid4vc): include auth-code scope in credential offer

Sshovon:
We are trying to align our OID4VCI issuer flow with [HAIP 1.0-05 section 4.2](https://openid.net/specs/openid4vc-high-assurance-interoperability-profile-1_0-05.html#section-4.2), which states that for **authorization_code** the issuer **MUST** include a **scope** value so the wallet can identify the credential type, and the wallet MUST use it in the authorization request.
#### Interoperability question:

- **Scope validation gap**: In this draft we currently do not validate that the **scope** used in **authorization_code**. Should we enforce this mapping for [HAIP 4.3 compliance](https://openid.net/specs/openid4vc-high-assurance-interoperability-profile-1_0-05.html#section-4.3) _(scope MUST map to specific Credential Type)_, and if multiple credentials are offered, then what should be the validation approach?

But as section 4.3 also mentions _scope value may be pre-agreed_ , I think we can give flexibility to the users to set their scope.

### DIFF
--- a/packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerService.ts
+++ b/packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerService.ts
@@ -1480,7 +1480,7 @@ export class OpenId4VcIssuerService {
         authorization_server: config.issuerMetadata.credentialIssuer.authorization_servers
           ? authorizationServerUrl
           : undefined,
-        scope: scope,
+        scope,
       }
     }
 
--- a/packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerService.ts
+++ b/packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerService.ts
@@ -1453,6 +1453,7 @@ export class OpenId4VcIssuerService {
     if (authorizationCodeFlowConfig) {
       const { requirePresentationDuringIssuance } = authorizationCodeFlowConfig
       let authorizationServerUrl = authorizationCodeFlowConfig.authorizationServerUrl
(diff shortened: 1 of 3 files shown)


Sshovon:
@hacdias need your opinion on this..

hacdias:
I think it's better to give the flexibility to check it or not, allowing you to comply with HAIP, or not.

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerServiceOptions.ts
```diff
@@ -96,6 +96,14 @@ export interface OpenId4VciAuthorizationCodeFlowConfig {
    * @default false
    */
   requirePresentationDuringIssuance?: boolean
+  /**
```
hacdias, line 100:
```suggestion
  requirePresentationDuringIssuance?: boolean

  /**
```

hacdias:
@Sshovon could you add a changelog entry?

Sshovon:
@hacdias  change log added.

hacdias (approved):

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerService.ts
```diff
@@ -1479,6 +1480,7 @@ export class OpenId4VcIssuerService {
         authorization_server: config.issuerMetadata.credentialIssuer.authorization_servers
           ? authorizationServerUrl
           : undefined,
+        scope: scope,
```
berendsliedrecht:
```suggestion
        scope,
```

PR #2788, opened 2026-05-08
fix(webvh): revocation status list shape validation

genaris:
After I've noticed an error when resolving a revocation status list, and it was because it was expecting the timestamp on the shape of it. So a simple way here is to inject the given timestamp so the validation is properly done (and the returned object is a complete revocation status list object).

cc @rmlearney-digicatapult


rmlearney-digicatapult:
Interesting :) Which of the guards/throws were you hitting?

berendsliedrecht (approved):

genaris:
> Interesting :) Which of the guards/throws were you hitting?

When it tried to resolve a revocation status list, threw because the object didn't have a `timestamp` and it was expected on the object shape.

PR #2793, opened 2026-05-12
feat(didcomm): implement DIDComm V2 DID rotation and relationship termination

tarunvadde:
Implements the V2 DID rotation and "Ending a Relationship" mechanisms in credo-ts.

## What's new

- `DidCommEmptyMessage` (`empty/1.0/empty`) with `from_prior` header, used as the carrier when a rotation or termination signal has no other message to ride on.
- `DidCommDidRotateV2Service` unifies all V2 rotation work: rotate our DID, sign and verify `from_prior` JWTs, apply incoming rotations, terminate the connection. Replaces the earlier `DidCommFromPriorService`.
- `from_prior` is now processed in the inbound pipeline before handler dispatch, so any V2 message can announce a rotation or termination. The empty-message handler becomes a no-op stub because the work moved upstream.
- Outbound V2 messages automatically attach `from_prior` from connection metadata until the peer acknowledges by sending a message to the new DID. Sender re-includes it on every outbound message until cleared.
- For multi-use V2 OOB invitations and implicit invitations, the inviter rotates to a fresh per-pair DID on the first inbound message from each accepter. Two accepters on the same invitation get distinct pairwise DIDs and cannot be correlated through the invitation DID.
- `DidRotateV2` connection metadata key tracks the pending rotation state (`fromPriorJwt`, `priorDid`, `newDid`) until acknowledged.
- `DidCommConnectionsApi.hangup` works for both V1 (`hangup` message) and V2 (rotate-to-nothing).
- E2E tests for V2 rotation.

## Some fixes

- `DidCommConnectionsApi.acceptOutOfBandInvitation` now sets `mediatorId` on v2 peer-DID connections.
- `DidCommConnectionsApi.removeRouting` falls back to `previousDids.at(-1)` so the post-hangup `deleteById` flow still emits the keylist update.

## Out of scope

Sending a `report-problem/2.0/problem-report` back to the rotating party when `from_prior` fails to verify. The V2.1 spec marks this as MAY, not MUST, and the V2 report-problem message isn't wired up in this codebase yet. We reject invalid `from_prior` in `processFromPrior` (spec-compliant SHOULD-reject). Notification can be a follow-up.


packages/didcomm/src/modules/connections/services/DidCommDidRotateService.ts
```diff
@@ -144,6 +144,21 @@ export class DidCommDidRotateService {
     })
   }
 
+  /**
+   * Mark a connection as terminated by the peer per DIDComm V2 rotate-to-nothing.
+   *
+   * @see https://identity.foundation/didcomm-messaging/spec/v2.1/#ending-a-relationship
+   */
+  public async processRotateToNothing(agentContext: AgentContext, connection: DidCommConnectionRecord) {
```
genaris: DidCommDidRotateService is currently thought to address [DID Rotate](https://github.com/decentralized-identity/aries-rfcs/tree/main/features/0794-did-rotate) protocol messages. I think we can do this logic in a different service, alongside other DID rotation operations for DIDComm V2 (e.g. including the logic we have now in DidCommFromPriorService).

packages/didcomm/src/modules/connections/DidCommConnectionsApi.ts
```diff
@@ -516,6 +536,105 @@ export class DidCommConnectionsApi {
     }
   }
 
+  /**
+   * Emit a DIDComm V2 relationship-termination signal to `connection.theirDid`.
+   */
+  private async sendV2TerminationSignal(connection: DidCommConnectionRecord): Promise<void> {
```
genaris: Since this is quite complex and can be shared with the logic needed for general DID rotation over DIDComm V2, it could be moved to the aforementioned "DidCommDidRotateV2Service" (or find a better name for it).

genaris (changes requested):
Wow I didn't expect this to be so complex! But in any case I would like to see it combined to a more general approach for DID rotation over DIDComm V2. The challenge here is that not only an empty message can trigger a DID Rotation (or DIDComm connection termination), but actually ANY message can do it.

Can you tell me how it works right now for Out of Band invitations? Because if we want to create a multi use invitation, we'll need for sure to rotate the DID as soon as we send the first message to the other party. I don't see any logic for that in current code. It might be acceptable (for some time) to not support multi use invitations on DIDComm V2, but we'll still need to support it because that's what we need when connecting to agents using implicit invitations (otherwise, we are going to use always the same DID to sign messages to all our connections).

tarunvadde:
> Wow I didn't expect this to be so complex! But in any case I would like to see it combined to a more general approach for DID rotation over DIDComm V2. The challenge here is that not only an empty message can trigger a DID Rotation (or DIDComm connection termination), but actually ANY message can do it.
> 
> Can you tell me how it works right now for Out of Band invitations? Because if we want to create a multi use invitation, we'll need for sure to rotate the DID as soon as we send the first message to the other party. I don't see any logic for that in current code. It might be acceptable (for some time) to not support multi use invitations on DIDComm V2, but we'll still need to support it because that's what we need when connecting to agents using implicit invitations (otherwise, we are going to use always the same DID to sign messages to all our connections).

Actually, to scope the PR to only 'Ending the relationship' for DIDComm V2, so the only `from_prior` trigger I did is on the empty message, and it's only the rotate-to-nothing case. General `from_prior` handling on arbitrary v2 messages is not here.

Currently, there is no DID rotation for OOB V2 invitations. 

`createInvitation({ didCommVersion: 'v2' })` generates one peer DID per invitation and reuses it as the invitation's `from`. On the inviter side, when the first inbound message from an accepter arrives, the auto-create path in `DidCommMessageReceiver` just sets `connection.did = recipient`, which is the invitation DID. So for both multi-use and implicit invitations, every accepter's inviter-side connection ends up sharing that one DID. Accepter side is fine, since `createPeerDidForV2OOB` already generates a fresh peer DID per acceptance, so the asymmetry is only on the inviter end.

Do you want these to be implemented in the same PR or a follow up?

genaris:
> Do you want these to be implemented in the same PR or a follow up?

Yes, do it in the same PR since it will help to better think the whole code. Sharing DID among different connections is a possible but not ideal case in terms of security.

tarunvadde:
> > Do you want these to be implemented in the same PR or a follow up?
> 
> Yes, do it in the same PR since it will help to better think the whole code. Sharing DID among different connections is a possible but not ideal case in terms of security.

Will do that.

genaris (approved):

PR #2795, opened 2026-05-12
fix: bug fix -  React Native secure-environment signing and biometric key configuration

MosCD3:
Adds SecureEnvironmentKeyManagementService options so consumers can disable biometric-backed keys with `new SecureEnvironmentKeyManagementService({ biometricsBacked: false })`, while keeping the default behavior as true.

Also converts KMS signing input to a plain Uint8Array before calling `@animo-id/expo-secure-environment`, fixing Buffer inputs from Credo’s JWS path that can fail at the iOS native bridge.


packages/react-native/src/kms/SecureEnvironmentKeyManagementService.ts
```diff
@@ -117,7 +126,11 @@ export class SecureEnvironmentKeyManagementService implements Kms.KeyManagementS
       // Kms.assertKeyAllowsSign(publicJwk)
 
       // Perform the signing operation
-      const signature = await secureEnvironment.sign(options.keyId, options.data)
+      const signature = await secureEnvironment.sign(
```
genaris, line 136: Have you analyzed what happens in case a keypair was created with `biometricsBacked` on (current default) and we want to sign data with it and `biometricsBacked` set to false?

If I understand correctly, under iOS it will still work because the biometrics prompt is enforced at OS level. But in Android it would throw an error. In the opposite case (key created without biometrics, sign with biometrics) it would work in both scenarios, probably showing a biometrics prompt that is not actually needed.

I don't see any big problem with this if it's like I described, but maybe it's worth to add some notes about this possible mismatch to avoid confusions.

TimoGlastra: I think (but we need input from @berendsliedrecht) biometricsBacked is not needed to be passed for the sign operation, just the generate call? But maybe my understanding is wrong

MosCD3: I checked `@animo-id/expo-secure-environment` and `biometricsBacked` is still used by the `sign()` wrapper on Android, while iOS ignores the sign-time value and relies on the key access-control settings. So I think we should keep passing it for now. The mismatch you described is possible on Android if a key is created with one setting and signed with another, so I added a note to the exported options type explaining that consumers should use the same biometricsBacked setting for signing as was used when creating the key

PR #2797, opened 2026-05-18
feat(core): W3C VC Data Integrity module

rmlearney-digicatapult:
**Summary**

This PR adds a new Data Integrity module to core, including proof types, validation, proof processing utilities, cryptosuite registration, proof creation/verification services, and agent/module wiring.

The module provides a standards-aligned foundation for Data Integrity proofs and makes Data Integrity available by default in agent module configuration.

**Why**
We need a first-class Data Integrity implementation in core so internal (e.g. VC, webvh) and external consumers can:

- Create and verify Data Integrity proofs consistently
- Reuse shared proof validation and processing logic
- Register and resolve cryptosuites through a single registry
- Access Data Integrity APIs without duplicating proof orchestration logic

**What’s Included**

1. Proof model and validation foundation

- Added Data Integrity proof/document types and proof option builders
- Added validation/assertion helpers for:
  - Single-proof document flows
  - Proof-set document flows
  - Required proof member checks
- Added structured Data Integrity error/result model for creation and verification outcomes

2. Proof-processing utilities
- Added utilities for:
  - Proof chain handling
  - Parsing/normalization
  - Key resolution helpers
  - Strict `dateTimeStamp` checks including leap year, time zone as required by spec
- Centralises reusable spec-oriented processing steps used by services and cryptosuites

3. Cryptosuite integration
- Added cryptosuite interfaces and registry
- Added `EdDSA-JCS-2022` cryptosuite as first implementation
- Wired cryptosuite registration into module setup
- Comprehensive test suite

4. Service/API and module surface
- Added proof service and API layer for:
  - Proof creation
  - Single-proof verification (permissive and fail-fast)
  - Proof-set/chain verification orchestration (permissive and fail-fast)
- Added module public/internal barrels
- Added module export on core index
- Registered Data Integrity as a default agent module

This module was built to be conformant with the `https://www.w3.org/TR/vc-data-integrity/` specification while preserving separation of responsibilities between different potential callers. For example, @context validation does not occur within the module as this is the responsibility of the VC service during the post-proof flow.

The `eddsa-jcs-2022` cryptosuite is fully conformant with the `https://www.w3.org/TR/vc-di-eddsa/#eddsa-jcs-2022` standard and the code is fully enumerated with each specification paragraph and line item implemented and a full test suite.

**Opinionated elements**
I have implemented Javascript normalisation to drop fields containing `undefined` - this is not within the spec but I do not believe that we should be signing over `undefined` fields

While power users could access the cryptosuite directly, most services will call via the `W3cDataIntegrityProofService`. For this reason some of the strict shape-checks live in the calling service, not the cryptosuite.

I have implemented 2 barrels for export - `index.ts` which will be hooked through `@credo-ts/core` and `internal.ts` which will be used by internal `credo-ts` consumers by direct import. These contain different exports for protecting internal functionality and avoiding complex ecosystem migrations in future.

I have installed `json-canonicalize` for strict JSON canonicalisation as required by `eddsa-jcs-2022`

**NB**
At the moment this module is not wired externally, only instantiated on the base agent. I have local branches where `webvh` is migrated to use this core data integrity module and `W3CV2` uses this module for `DiVp` and `DiVc` support. I built these to help shape the implementation and API surface and will push these after merge.


packages/core/src/modules/w3c-di/proof-processing/iso8601Datetime.ts
genaris, line 1: `proof-processing/` contains kebab-case (`proof-validation.ts`, `iso8601-datetime.ts`, `proof-set-validation.ts`, `proof-purpose-validation`) alongside camelCase (`keyUtils.ts`). Although there are some exceptions, Core's convention for non-class files is camelCase or single lowercase words (`decodeSdJwtVc.ts`, `disclosureFrame.ts`, `mdocUtil.ts`, `validators.ts`), so I suggest to rename the 4 kebab-case files to camelCase.

packages/core/src/modules/w3c-di/proof-processing/normalization.ts
genaris, line 1: Identifiers and filenames across `core/src` are consistently American (`normalize` ×11, `normalized` ×9, `initialize` ×56, zero `-ise`/`-isation`). This module introduces British forms into exported identifiers and a filename (`normalisedMediaType`, `normalisation.ts`). Main does contain `behaviour`/`licence`/`recognise`, but only in comments and one test string, never in code. Worth aligning the identifiers; and note the same pattern appears in #2827's `W3cV2DataIntegrityContextValidator.ts`, so it's worth fixing in both.

packages/core/src/modules/w3c-di/internal.ts
genaris, line 1: No problem on adding it here, but it looks like it will only be consumed once #2827 lands.

genaris (changes requested):
After so much time being opened, informally reviewed and tested in some other following PRs, I think it's worth to merge this to main ASAP.

I've left some minor comments, most of them cosmetic. 

Can you please add a changeset for this?

packages/core/src/modules/w3c-di/proof-processing/normalization.ts
rmlearney-digicatapult: Sorry force of habit, I'll amend

packages/core/src/modules/w3c-di/internal.ts
rmlearney-digicatapult: The other early consumer should be `didwebvh` as well :)

genaris (approved):

genaris:
OIDF Conformance tests are failing for external reasons (see #2888), so merging this to allowing progressing with related PRs meanwhile the upstream suite is fixed.

PR #2799, opened 2026-05-20
feat(didcomm): make DidCommOutOfBandInvitationV2 sendable through a connection

tarunvadde:
`DidCommOutOfBandInvitationV2` was a plain class, so v2 OOB invitations couldn't be passed to `messageSender.sendMessage()`. To send a v2 OOB through an existing v2 connection consumers had to wrap it via `convertV2InvitationToOutOfBandInvitation` and send the v1 wrapper, producing `out-of-band/1.1/invitation` on the wire even for v2 to v2 flows.

This change makes it a `DidCommMessage` subclass mirroring `DidCommBasicMessageV2`.

Some additional types + style fixes are present too

### DIFF
--- a/packages/didcomm/src/modules/oob/messages/DidCommOutOfBandInvitationV2.ts
+++ b/packages/didcomm/src/modules/oob/messages/DidCommOutOfBandInvitationV2.ts
@@ -1,12 +1,15 @@
-import { CredoError, JsonEncoder } from '@credo-ts/core'
+import { CredoError, JsonEncoder, JsonTransformer } from '@credo-ts/core'
 import { Expose } from 'class-transformer'
 import { IsOptional, IsString } from 'class-validator'
 import queryString from 'query-string'
 
 import { DidCommMessage } from '../../../DidCommMessage'
+import { DidCommAttachment } from '../../../decorators/attachment/DidCommAttachment'
 import type { DidCommPlaintextMessage } from '../../../types'
 import type { DidCommVersion } from '../../../util/didcommVersion'
 import { IsValidMessageType, parseMessageType } from '../../../util/messageType'
+import { normalizeV2PlaintextToV1 } from '../../../v2/normalize'
+import { mapV1AttachmentToV2, mapV2AttachmentToV1 } from '../../../v2/plaintextBuilder'
 import type { DidCommV2Attachment, DidCommV2PlaintextMessage } from '../../../v2/types'
 
 const LINK_PARAM = '_oob'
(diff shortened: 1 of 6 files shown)


packages/didcomm/src/modules/oob/messages/DidCommOutOfBandInvitationV2.ts
```diff
@@ -19,51 +26,97 @@ export interface DidCommOutOfBandInvitationV2Options {
...
 
   public static fromJson(json: Record<string, unknown>): DidCommOutOfBandInvitationV2 {
     const type = json.type as string
-    if (type !== DidCommOutOfBandInvitationV2.type) {
-      throw new CredoError(`Invalid v2 OOB invitation type: expected ${DidCommOutOfBandInvitationV2.type}, got ${type}`)
+    if (type !== DidCommOutOfBandInvitationV2.type.messageTypeUri) {
+      throw new CredoError(
+        `Invalid v2 OOB invitation type: expected ${DidCommOutOfBandInvitationV2.type.messageTypeUri}, got ${type}`
+      )
     }
```
genaris: In DidCommOutOfBandInvitation we simply have:

```
 public static fromJson(json: Record<string, unknown>) {
    return JsonTransformer.fromJSON(json, DidCommOutOfBandInvitation)
  }
```

Why this doesn't work here?

tarunvadde: Tried this and it actually drops most of the body fields, here's why:

The wire JSON `fromJson` receives looks like:
`{ "type": "...", "id": "...", "from": "...", "body": { "goal_code": "...", "goal": "...", "accept": [...] }, "attachments": [...] }`

But the `@Expose` decorators on the class expect:

- `@type` / `@id` (not `type` / `id`)
- top-level `goal` / `goal_code` / `accept` (not nested under `body`)

That's because they're aligned with the post-`normalizeV2PlaintextToV1` shape, since that's what the inbound dispatcher feeds the class when a v2 OOB arrives in-band through a v2 envelope.

So `JsonTransformer.fromJSON(json, DidCommOutOfBandInvitationV2) `on the raw wire would leave `id, goal, goalCode, accept` all `undefined`.

tarunvadde: Updated. fromJson is now a one-liner via JsonTransformer.fromJSON (with normalizeV2PlaintextToV1 first to reconcile the wire shape).

Also moved attachments to the base class ~attach mixin so both inbound paths (URL/QR and in-band v2) populate the same field. attachments stays exposed as a v2-shape getter for convertV2InvitationToOutOfBandInvitation.

genaris (approved):

PR #2800, opened 2026-05-20
feat(webvh): wire webvh to core data integrity module

rmlearney-digicatapult:
Built on top of #2797, pushed for visibility

Wire webvh `attested resource` handling to use core data integrity service + cryptosuite and remove cryptosuite from webvh

Bidirectional interoperability testing prior to removing existing cryptosuite from webvh confirming `@context` handling does not break signing/verification


packages/webvh/src/anoncreds/services/WebVhAnonCredsRegistry.ts
```diff
@@ -23,6 +23,9 @@ import type {
 import {
   type AgentContext,
   CredoError,
+  DataIntegrityApi,
```
TimoGlastra: Should we call this the `W3cDataIntegrityApi` (as well as the other classes)? Data Integrity is quite generic as a name.

packages/core/src/modules/data-integrity/DataIntegrityError.ts
```diff
@@ -0,0 +1,152 @@
...
+function formatIssueSummary(error: DataIntegrityProcessingIssue) {
+  return `[${error.type}] ${error.title}`
+}
+
+function formatIssueDetail(error: DataIntegrityProcessingIssue) {
+  return error.detail ? `${formatIssueSummary(error)}: ${error.detail}` : formatIssueSummary(error)
+}
+
+export function formatDataIntegrityIssueSummary(errors: DataIntegrityProcessingIssue[]): string {
+  return errors.map(formatIssueSummary).join('; ')
```
TimoGlastra: Will this all be printend on a single line? Do we want to split these onto new lines?

TimoGlastra:
Ah my comments should have been made on #2797 instead

packages/webvh/src/anoncreds/services/WebVhAnonCredsRegistry.ts
rmlearney-digicatapult: Good recommendation, the spec is centred around W3C VCs anyway

rmlearney-digicatapult:
@TimoGlastra I've made changes to #2797 :)

Prefixed classes & types (except for those required by spec) with w3c* and renamed the module folder to `/w3c-di`

packages/core/src/modules/data-integrity/DataIntegrityError.ts
rmlearney-digicatapult: Introduced newlines

rmlearney-digicatapult:
Going to provide a comprehensive `did:webvh` wrapper replacement after release of `3.0.0`

PR #2803, opened 2026-05-22
feat: mdoc revocation for issuance and verification

berendsliedrecht:
- Add revocation check for mdoc
- Added token-status-list module to create/update/fetch a token-status-list.

Right now there is no, yet, great synergy between an mdoc/sd-jwt and status list.

We could add an API like `setCredential(token: Uint8Array | string, credential: SdJwtRecord | MdocRecord): Promise<Uint8Array | string>` which adds the credential to the list and adds the metadata to the credential. @TimoGlastra what do you think?


.changeset/cold-gifts-brush.md
```diff
@@ -0,0 +1,5 @@
+---
+"@credo-ts/core": patch
+---
+
+Mdoc revocation for issuance and verification
```
TimoGlastra, line 5: + new status list module

packages/core/package.json
```diff
@@ -53,7 +53,9 @@
     "@multiformats/base-x": "^4.0.1",
     "@noble/curves": "^2.0.1",
     "@noble/hashes": "^2.0.1",
-    "@owf/mdoc": "^0.6.0",
+    "@owf/cose": "0.3.0-alpha-20260522120032",
+    "@owf/mdoc": "0.6.1-alpha-20260522131848",
+    "@owf/token-status-list": "0.3.0-alpha-20260522120032",
```
TimoGlastra: We should probably have new stable releases for all of these?

Also, i think the mdoc should be labeleled as breaking / 0.7 due to the addition of `fetch` (breaking new requirement)

packages/core/src/modules/token-status-list/TokenStatusListApi.ts
```diff
@@ -0,0 +1,45 @@
...
+export class TokenStatusListApi {
+  private agentContext: AgentContext
+  private tokenStatusListService: TokenStatusListService
+
+  public constructor(agentContext: AgentContext, tokenStatusListService: TokenStatusListService) {
+    this.agentContext = agentContext
+    this.tokenStatusListService = tokenStatusListService
+  }
+
+  public async createTokenStatusList(options: CreateTokenStatusListOptions): Promise<Uint8Array | string> {
```
TimoGlastra: just for future extensibility, I'd make this an object with a single property (for all of these). Biggest reason for breaking changes we've had

berendsliedrecht: huh? What do you mean? an object with an object?

packages/core/package.json
berendsliedrecht: Yes! When this is approved and all good we can do a full release of all subdeps.

packages/core/src/modules/token-status-list/TokenStatusListOptions.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  /**
+   * Will determine whether it will be signed or authenticated
+   */
+  keyId: string
+}
+
+export type BatchUpdateTokenStatusListOptions<TSL extends Uint8Array | string> = {
+  token: TSL
+  indexAndValue: Array<[number, StatusType]>
```
TimoGlastra: I'd make this an array of objects. With `index` and `status` fields

And we can just call this `values` maybe. Operation = update, property = values, each value needs index and status

packages/core/src/modules/token-status-list/TokenStatusListOptions.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  /**
+   * Will determine whether it will be signed or authenticated
+   */
+  keyId: string
+}
+
+export type FetchTokenStatusListOptions = {
+  uri: string
```
TimoGlastra, line 35: should we add `acceptedFormats` array here, which by default is cwt and jwt, can be limited to only one.

TimoGlastra: will influence the return type of the method

packages/core/src/modules/token-status-list/TokenStatusListOptions.ts
```diff
@@ -0,0 +1,39 @@
...
+  hostingUri: string
+  aggregationUri?: string
+
+  /**
+   * Will determine whether it will be signed or authenticated
+   */
+  keyId: string
+}
+
+export type UpdateTokenStatusListOptions<TSL extends Uint8Array | string> = {
```
TimoGlastra, line 24: I'd just merge these two, there's no difference between batch and single (single = batch with only one entry)

packages/core/src/modules/token-status-list/TokenStatusListOptions.ts
```diff
@@ -0,0 +1,39 @@
+import { type BitsPerStatus, StatusType } from '@owf/token-status-list'
+
+export type CreateTokenStatusListOptions = {
+  format: 'cwt' | 'jwt'
+  statusListLength: number
+  bitsPerStatus: BitsPerStatus
+  hostingUri: string
```
TimoGlastra:
```suggestion
  statusListUri: string
```

packages/core/src/modules/token-status-list/TokenStatusListApi.ts
```diff
@@ -0,0 +1,45 @@
...
+  public async batchUpdateTokenStatusList<TSL extends Uint8Array | string>(
+    options: BatchUpdateTokenStatusListOptions<TSL>
+  ): Promise<TSL> {
+    return this.tokenStatusListService.batchUpdateTokenStatusList<TSL>(this.agentContext, options)
+  }
+
+  public async fetchTokenStatusList<TSL extends Uint8Array | string = Uint8Array | string>(
+    options: FetchTokenStatusListOptions
+  ): Promise<TSL> {
+    return this.tokenStatusListService.fetchTokenStatusList<TSL>(this.agentContext, options)
```
TimoGlastra: Here I'd also add the parsed status list as part of the return value (object with raw and parsed). Do we need a way to verify it as well?

packages/core/src/modules/token-status-list/TokenStatusListApi.ts
TimoGlastra: The return value i mean. It's now `Uint8array | string`. I'd make it an object. So it's easier in the future to e.g. add the encoded but also decoded format.

packages/core/src/modules/token-status-list/TokenStatusListOptions.ts
berendsliedrecht: Need to fix this in token-status-list, will do.

packages/core/src/modules/mdoc/MdocContext.ts
```diff
@@ -117,12 +120,16 @@ export const getMdocContext = (agentContext: AgentContext, { now }: { now?: Date
     x509: {
       getIssuerNameField: (input) => {
         const { certificate, field } = input
-        const x509Certificate = X509Certificate.fromRawCertificate(certificate)
+        const x509Certificate = X509Certificate.fromRawCertificate(
+          certificate instanceof Uint8Array ? certificate : certificate[0]
+        )
         return x509Certificate.getIssuerNameField(field)
```
TimoGlastra: why can certificate be an array? it's a bit weird the the `getIssuerNameField` or `getPublicKey` can be arrays, since it's meant to be a method on a specific certificate. I think we need to change this in the mdoc library?

packages/core/src/modules/mdoc/MdocOptions.ts
```diff
@@ -113,4 +113,6 @@ export type MdocSignOptions = {
    */
   issuerCertificate: X509Certificate | X509Certificate[]
   holderKey: PublicJwk
+
+  statusInfo?: { idx: number; uri: string; certificate?: X509Certificate }
```
TimoGlastra: We only support the index based method. Should we somehow incorporate that into this api? (there's also an uuid based list i think?), and should we use `index` over `idx` (that seems to be more our style for APIs?)

Also should we add this same property to the SD-JWT api?

packages/core/src/modules/mdoc/Mdoc.ts
```diff
@@ -182,6 +191,7 @@ export class Mdoc {
           ),
           issuerSigned: this.issuerSigned,
           disableCertificateChainValidation: false,
+          disableStatusValidation: false,
```
TimoGlastra, line 199: in sd-jwt it is called "verifyCredentialStatus". I think consistency with the existing mdoc properties makes sense, but i would also like to converge these APIs. 

https://github.com/openwallet-foundation/credo-ts/blob/main/packages/core/src/modules/sd-jwt-vc/SdJwtVcOptions.ts#L165

Maybe we can add this to sd-jwt as well, and we deprecate the current `verifyCredentialStatus`.

packages/core/src/modules/token-status-list/TokenStatusListService.ts
```diff
@@ -0,0 +1,158 @@
...
+ * @internal
+ *
+ * @todo with what value should we initialize the array? Valid or Invalid?
+ */
+@injectable()
+export class TokenStatusListService {
+  public async createTokenStatusList(
+    agentContext: AgentContext,
+    options: CreateTokenStatusListOptions
+  ): Promise<Uint8Array | string> {
```
TimoGlastra: I think a return type inferred from the input `format` can be helpful. And if it's an object make it 

```
{
 format: 'cwt'
 statusList: Uint8Array
} 
|
{
 format: 'jwt'
 statusList: string
} 
```

makes working with the return value easier

packages/core/src/modules/token-status-list/context/sign1Context.ts
```diff
@@ -0,0 +1,58 @@
...
+        )
+        return x509Certificate.getIssuerNameField(field)
+      },
+      getPublicKey: async (input) => {
+        const certificate = X509Certificate.fromRawCertificate(
+          input.certificate instanceof Uint8Array ? input.certificate : input.certificate[0]
+        )
+        return CoseKey.fromJwk(certificate.publicJwk.toJson()).publicKey
+      },
+    },
```
TimoGlastra: this seems duplicated, can't we reuse the logic between the different contexts?

TimoGlastra (commented):
The overall API looks great, but I have some nitpicks

packages/core/src/modules/token-status-list/context/sign1Context.ts
berendsliedrecht: Yes, I can add it to mdoc. good catch

packages/core/src/modules/mdoc/Mdoc.ts
berendsliedrecht: It seems to not be used in the service 😅 

```typescript
  public async verify<Header extends SdJwtVcHeader = SdJwtVcHeader, Payload extends SdJwtVcPayload = SdJwtVcPayload>(
    agentContext: AgentContext,
    { compactSdJwtVc, keyBinding, requiredClaimKeys, fetchTypeMetadata, trustedCertificates, now }: SdJwtVcVerifyOptions // <-- Not being extracted or used :0
  ): Promise<
    | {
        isValid: true
        sdJwtVc: SdJwtVc<Header, Payload>
        /**
         * The full type metadata chain, can be used for further processing the type metadata.
         *
         * Only populated if `fetchTypeMetadata` is set to true, and the resolver returned a type
         * metadata document.
         */
        typeMetadataChain?: NonEmptyArray<SdJwtVcTypeMetadata>
      }
    | { isValid: false; sdJwtVc?: SdJwtVc<Header, Payload>; error: Error }
  > {
```

I also noticed we use that boolean in more places, like openid4vp and w3c, what do we want to do with that? Maybe a different PR for these changes?

packages/core/src/modules/token-status-list/TokenStatusListApi.ts
berendsliedrecht: Yes, verification is actually quite important. I think I missed it previously, adding it now.

packages/core/src/modules/mdoc/Mdoc.ts
TimoGlastra: different PR is fine, but if we use it in OID4VC API, we now also need to pass this down to mdoc api

packages/core/src/modules/mdoc/MdocOptions.ts
berendsliedrecht: Changed to index. The UUID would be a feature to be added to the token-status-list rpeo itself. I think it would be a good addition for sure.

packages/core/src/crypto/contexts/mdocContext.ts
```diff
@@ -0,0 +1,95 @@
...
+        const { publicKey, privateKey, salt, info, digestAlgorithm } = input
+        const ikm = p256.getSharedSecret(privateKey, publicKey, true).slice(1)
+        const hashedSalt = Hasher.hash(salt, digestAlgorithm ?? 'sha-256')
+        return hkdf(sha256, ikm, hashedSalt, info, 32)
+      },
+    },
+
+    cose: {
+      mac0: {
+        sign: getMac0Context(agentContext).mac,
```
TimoGlastra: why is it called mac in the mac0 context, but sign here?  Shouldn't `Mac0Context` follow the same interface as `mac0` in the MdocContext?

berendsliedrecht: I kept it initially to stay backwards comptabile, but we have breaking changes anyways so I can change it.

TimoGlastra (approved):

PR #2807, opened 2026-05-25
feat(didcomm): use independent keys for verification and agreement

genaris:
Removing the old practice of having a single key for verification and agreement (with one-way Ed -> X conversion). In V2 we encourage using two independent keys.

cc @tarunvadde


packages/didcomm/src/models/DidCommRouting.ts
```diff
@@ -3,6 +3,14 @@ import type { Kms } from '@credo-ts/core'
...
   endpoints: string[]
   recipientKey: Kms.PublicJwk<Kms.Ed25519PublicJwk>
+  /**
+   * Separate X25519 key for DIDComm V2 key agreement. Stored independently in the KMS
+   * (not derived from recipientKey at runtime). Used as the `keyAgreement` verification
+   * method in the DID document and for ECDH-ES / ECDH-1PU encryption.
+   *
+   * When not set, the X25519 key will be derived from `recipientKey` at runtime (legacy behavior).
+   */
+  keyAgreementKey?: Kms.PublicJwk<Kms.X25519PublicJwk>
```
TimoGlastra, line 13: As I understand it, `routingKeys` should/could be `keyAgreementKeys`. Why not extend the type of `routingKeys`, and if it's a `X25519` key we don't transform it anymore?

genaris: I think this interface became confusing, because it is used for both V1 and V2 cases: in DIDComm V2, `routingDid` is used, so `routingKeys` are actually left to V1 and there we'll still have the Ed25519 as it was in legacy cases. @tarunvadde can you confirm this?

Maybe a better approach would be to use different interfaces for each, and make a strong separation between V1 and V2.

tarunvadde: Yeah you're right. `routingKeys` only gets used in V1 forward-message construction, and `routingDid` is V2-only. With `keyAgreementKey` added, only `endpoints` is really shared.

What about discriminated union like (`DidCommV1Routing | DidCommV2Routing`)? over fully separate interfaces.

Separate interfaces is cleaner long-term but need to make changes into `OutOfBandApi`, `ConnectionsApi`, `DidExchangeProtocol`, `getDidCommOutboundMessageContext`.

genaris: Ok, I'm merging and leaving the comments on main PR for DIDComm V2 support.

PR #2808, opened 2026-05-26
feat(didcomm-v2): add A256CBC-HS512 content encryption and apu/apv headers

tarunvadde:
Brings `DidCommV2EnvelopeService` up to spec for ECDH-1PU + A256CBC-HS512 authcrypt: top-level `recipients`, shared `epk` in protected, `apu`/`apv`, JWE AAD, and the `ccTag`. The missing `ccTag` was the root cause of `Encryption error` at askar's `keyUnwrapKey` against SICPA-encrypted envelopes, also seen with RootsID Mediator.

X25519 only. P-256, signed messages, and XC20P anoncrypt are follow-ups.

### DIFF
--- a/packages/didcomm/src/v2/DidCommV2EnvelopeService.ts
+++ b/packages/didcomm/src/v2/DidCommV2EnvelopeService.ts
@@ -9,18 +9,23 @@ import {
   type Logger,
   TypedArrayEncoder,
 } from '@credo-ts/core'
-import type { DidCommV2EncryptedMessage, DidCommV2PlaintextMessage } from './types'
+import { computeApu, computeApv } from './apuApv'
+import type { DidCommV2ContentEncryptionAlgorithm, DidCommV2EncryptedMessage, DidCommV2PlaintextMessage } from './types'
 
 export interface DidCommV2EnvelopeKeys {
   recipientKey: Kms.PublicJwk<Kms.X25519PublicJwk>
   senderKey: Kms.PublicJwk<Kms.X25519PublicJwk>
   /** DID URL of the sender key; used as skid in JWE so recipient can resolve it. Falls back to senderKey.keyId if absent. */
   senderKeySkid?: string
+  /** Content encryption algorithm. Defaults to A256CBC-HS512 (mandatory authcrypt enc per DIDComm v2.1). */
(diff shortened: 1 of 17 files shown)


packages/didcomm/src/v2/DidCommV2EnvelopeService.ts
```diff
@@ -9,18 +9,23 @@ import {
...
   senderKeySkid?: string
+  /** Content encryption algorithm. Defaults to A256CBC-HS512 (mandatory authcrypt enc per DIDComm v2.1). */
+  contentEncryptionAlgorithm?: DidCommV2ContentEncryptionAlgorithm
 }
 
 /** Keys for anoncrypt: only recipient key; no sender (anonymous). */
 export interface DidCommV2AnoncryptKeys {
   recipientKey: Kms.PublicJwk<Kms.X25519PublicJwk>
+  /** Content encryption algorithm. Defaults to A256GCM (legacy default; A256CBC-HS512 and XC20P are also valid). */
+  contentEncryptionAlgorithm?: DidCommV2ContentEncryptionAlgorithm
```
genaris: Can you explain why A256GCM is considered a 'legacy default' for AnonCrypt?

Also in this comment it says "XC20P" is also valid but we don't actually support it.

genaris (commented):
I'm OK with the changes as long as they are interoperable with most DIDComm V2 implementations, but I'm a bit concerned about the defaults used, since A256CBC-HS512 looks clear for Authcrypt but I'm not sure A256GCM for Anoncrypt is the right choice, considering that its support is not mandatory on the spec. And also I don't see any way for the user to change this default, so it can be a bit risky if we are not sures about it.

packages/didcomm/src/v2/DidCommV2EnvelopeService.ts
tarunvadde: I treated credo's prior behavior (hardcoded A256GCM) as "the default to preserve" and labeled it "legacy". 

Switched the anoncrypt default from A256GCM to A256CBC-HS512. 

I also have two more follow-up branches: one adds `XC20P` anoncrypt support (SICPA's default), and one adds two `DidCommModuleConfig` options to override the defaults. Kept them separate to keep PRs small for review.

genaris (approved):

PR #2811, opened 2026-05-27
feat(didcomm-v2): support P-256 keyAgreement

tarunvadde:
Adds P-256 keyAgreement support for DIDComm v2. Opt in via `DidCommModuleConfig.v2KeyAgreementCurve: 'X25519' | 'P-256'` (default X25519, throws at construction if 'P-256' is set without 'v2' in didcommVersions).

KMS layer accepts P-256 in ECDH-1PU+A256KW with a same-curve sender/recipient invariant. Envelope service generates P-256 ephemerals, emits EC/P-256 EPK with `x` and `y`, dispatches KDF per curve. OOB peer-DID builder emits the keyAgreement VM as `JsonWebKey2020` for P-256, preserving the existing `X25519KeyAgreementKey2019` path for X25519.

`resolveV2Keys` and the v2 connection equality checks widen to `X25519 | P-256`. New `keyAgreementsEqual(a, b)` helper replaces `toX25519(a).equals(toX25519(b))` and only bridges Ed25519 to X25519 birationally as per RFC 7748 (P-256 is direct-compare only). Core's `DidDocument.recipientKeys` widens to include P-256 (additive; v1 paths skip P-256 via existing curve guards).

### DIFF
--- a/packages/core/src/modules/dids/domain/DidDocument.ts
+++ b/packages/core/src/modules/dids/domain/DidDocument.ts
@@ -246,7 +246,8 @@ export class DidDocument {
   }): Array<{
     verificationMethod: VerificationMethod
     publicJwk: PublicJwk<
-      MapX25519ToEd25519 extends true ? Ed25519PublicJwk : Ed25519PublicJwk | X25519PublicJwk | P256PublicJwk
+      | (MapX25519ToEd25519 extends true ? Ed25519PublicJwk : Ed25519PublicJwk | X25519PublicJwk)
+      | P256PublicJwk
     >
   }> {
     const recipientKeys: Array<{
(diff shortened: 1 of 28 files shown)


packages/core/src/modules/dids/domain/DidDocument.ts
```diff
@@ -245,11 +245,13 @@ export class DidDocument {
     mapX25519ToEd25519: MapX25519ToEd25519
   }): Array<{
     verificationMethod: VerificationMethod
-    publicJwk: PublicJwk<MapX25519ToEd25519 extends true ? Ed25519PublicJwk : Ed25519PublicJwk | X25519PublicJwk>
+    publicJwk: PublicJwk<
+      MapX25519ToEd25519 extends true ? Ed25519PublicJwk : Ed25519PublicJwk | X25519PublicJwk | P256PublicJwk
+    >
```
TimoGlastra, line 251: Is this correctly taking into account all variants? Can `MapX25519ToEd25519` be true and also have `P256PublicJwk` as return type?

tarunvadde: No. When `MapX25519ToEd25519` is `true` the conditional resolves to `Ed25519PublicJwk` only, and the runtime at line 343 drops P-256 entries (no Ed25519 birational sibling), so the returned array is Ed25519-only.

`if (publicJwk.is(P256PublicJwk)) return []`

TimoGlastra: Why though? It is not clear to me why `MapX25519ToEd25519` means "drop P-256 keys"? A did document can have multiple services, which mix ed25519/p-256 right?

tarunvadde: I was thinking of :true as "v1-compatible output" so figured P-256 had no place. Agree the flag name doesn't say that. P-256 passes through both branches now.

tarunvadde:
Hey @TimoGlastra, could you please review and add comments if any?

genaris (approved):
Let's merge to move forward with #2815 and move any outstanding comment on #2704

PR #2812, opened 2026-05-27
feat(didcomm-v2): signed messages and sign-then-encrypt

tarunvadde:
Adds DIDComm v2.1 signed messages (typ application/didcomm-signed+json) plus sign-then-encrypt and sign-then-anoncrypt convenience methods on the envelope service. Supports the three algorithms the v2.1 spec mandates for verification: EdDSA, ES256, and ES256K.

Two notable decisions worth flagging:

1. First, kid lives in the unprotected per-signature header on emit. The spec example shows it in the protected header, but SICPA's didcomm-python and didcomm-rust both reject anything else in their validate_jws path. Verify accepts kid in either location, so we interop with both styles. 

**Update on 1st pointer**: 
After testing interop with affinidi, there's divergence. Affinidi reads kid only from the protected header. So interop with Affinidi fails here. Any suggestions here? @TimoGlastra @genaris [**FIXED** from Affinidi]

2. Second, single-signer only. The spec says "only the message sender's signature is valid", multi-sig has no defined semantics anywhere in v2.1, and no reference impl produces multi-sig envelopes. Multi-sig is rejected on verify with an explicit error so the limit is loud rather than hidden.

Round-trip tested on Askar for all three algorithms. Three SICPA fixtures from didcomm-rust are committed as known-good inputs. Cross-impl interop verified locally against didcomm-python in both directions.

### DIFF
--- a/packages/core/src/crypto/index.ts
+++ b/packages/core/src/crypto/index.ts
@@ -2,8 +2,11 @@ export * from './hashes'
 export { JwsService } from './JwsService'
 export type { JwsSigner, JwsSignerDid, JwsSignerJwk, JwsSignerWithJwk, JwsSignerX5c } from './JwsSigner'
 export type {
+  Jws,
+  JwsCompactFormat,
   JwsDetachedFormat,
   JwsFlattenedDetachedFormat,
+  JwsFlattenedFormat,
   JwsGeneralFormat,
   JwsProtectedHeaderOptions,
 } from './JwsTypes'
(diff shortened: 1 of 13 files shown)


stormer78:
Hi Tarun,

Great work on interop testing and identifying several issues in the Affinidi Library.

Your issues have all been addressed on our side
https://github.com/affinidi/affinidi-tdk-rs/pull/326

Let us know if any other issues remain.

tarunvadde:
Hi @stormer78,

Thank you for taking the time in addressing these issues. I will continue my interop testing between Credo and Affinidi and let you know if I find any issues.

genaris (approved):

PR #2814, opened 2026-06-03
feat: last changes for mdoc revocation

berendsliedrecht:


packages/core/src/modules/token-status-list/__tests__/TokenStatusListService.test.ts
```diff
@@ -77,13 +102,18 @@ describe('TokenStatusListService', () => {
...
+          bitsPerStatus: 1,
+        },
         statusListUri: 'https://example.com/status/1',
-        keyId: key.keyId,
-      }
+        signer: {
+          method: 'x5c',
+          x5c: [certificate],
+        },
+      } as unknown as CreateTokenStatusListOptions
```
TimoGlastra, line 127: why `as unknown`? that seems like there's something wrong with the types?

packages/core/src/modules/token-status-list/TokenStatusListOptions.ts
```diff
@@ -1,36 +1,175 @@
...
   /**
-   * Will determine whether it will be signed or authenticated
+   * Additional claims to include in the token payload.
    */
-  keyId: string
+  additionalPayload?: Record<string, unknown>
 }
 
+export type CreateTokenStatusListOptions = CreateJwtTokenStatusListOptions | CreateCwtTokenStatusListOptions
+
```
TimoGlastra: Why two interfaces if they're mostly the same? I think a base type will prevent the types from diverging in the future. I think only `format` and `additionalPayload` are different right?

packages/core/src/modules/token-status-list/__tests__/TokenStatusListService.test.ts
berendsliedrecht: That's the test, the type does not match so I want to test if we handle that correctly.

packages/core/src/modules/token-status-list/TokenStatusListSigner.ts
```diff
@@ -0,0 +1,36 @@
...
+   */
+  x5c: X509Certificate[]
+}
+
+export interface TokenStatusListSignerJwk {
+  method: 'jwk'
+  jwk: Kms.PublicJwk
+}
+
+export interface OpenId4VcJwtIssuerJwkEncoded {
```
TimoGlastra: why is this added here? it seems oid4vc specific

packages/core/src/modules/token-status-list/TokenStatusListService.ts
```diff
@@ -31,118 +38,222 @@ import type {
...
+          ...(jwt.payload.sub !== undefined && { sub: jwt.payload.sub }),
+          ...(jwt.payload.aud !== undefined && { aud: jwt.payload.aud }),
+          ...(jwt.payload.exp !== undefined && { exp: jwt.payload.exp }),
+          ...(jwt.payload.nbf !== undefined && { nbf: jwt.payload.nbf }),
+          ...(jwt.payload.jti !== undefined && { jti: jwt.payload.jti }),
+        }
+        const updatedPayload: Record<string, unknown> = {
+          ...existingClaims,
+          ...options.additionalPayload,
+          iat: Math.floor(issuedAt.getTime() / 1000),
```
TimoGlastra: we have dateToSeconds  util

packages/core/src/modules/token-status-list/TokenStatusListService.ts
```diff
@@ -31,118 +38,222 @@ import type {
...
+        // so we pull sub from additionalClaims explicitly to preserve it.
+        const existingClaims = {
+          ...jwt.payload.additionalClaims,
+          ...(jwt.payload.iss !== undefined && { iss: jwt.payload.iss }),
+          ...(jwt.payload.sub !== undefined && { sub: jwt.payload.sub }),
+          ...(jwt.payload.aud !== undefined && { aud: jwt.payload.aud }),
+          ...(jwt.payload.exp !== undefined && { exp: jwt.payload.exp }),
+          ...(jwt.payload.nbf !== undefined && { nbf: jwt.payload.nbf }),
+          ...(jwt.payload.jti !== undefined && { jti: jwt.payload.jti }),
+        }
```
TimoGlastra: interesting, we should probably fix this in `toJson`? This is abit odd logic (it was added by Claude, i didn't write this 😅 )

packages/core/src/modules/token-status-list/TokenStatusListService.ts
```diff
@@ -31,118 +38,222 @@ import type {
...
+  }
+
+  public async updateTokenStatusList<Format extends TokenStatusListFormat>(
+    agentContext: AgentContext,
+    options: UpdateTokenStatusListOptions
+  ): Promise<Extract<TokenStatusListResult, { format: Format }>> {
+    // TODO: jwk could also be supported
+    if (options.signer.method !== 'x5c') {
+      throw new Error(
+        `signer method '${options.signer.method}' is not supported for creating a token status list. Only x5c is`
```
TimoGlastra, line 163: we need support for did (in Paradym). What i meant is that we can skip did support in CWT for now (as it's only used with mdoc and x509). But for the JWT variant we need it, otherwise we can't use this API

packages/core/src/modules/token-status-list/TokenStatusListService.ts
```diff
@@ -31,118 +38,222 @@ import type {
...
-      new Array(options.statusListLength).fill(0),
-      options.bitsPerStatus,
-      options.aggregationUri
-    )
-    const kms = agentContext.dependencyManager.resolve(KeyManagementApi)
-    const jwk = await kms.getPublicKey({ keyId: options.keyId })
-    if (!jwk.alg) {
-      throw new CredoError(`Found JWK for key id '${options.keyId}', but did not find a required algorithm`)
+    // TODO: jwk could also be supported
+    if (options.signer.method !== 'x5c') {
```
TimoGlastra: same here

TimoGlastra (approved):
Most LGTM, only real issue is support for DIDs for JWT status list.

One other minor suggestion, but not really important right now, is that there's quite some duplication in the create/update logic. Basically a shared util that takes a status list (which is created in `create` method, and updated based on existing one in `update` method beforehand) would remove all the duplication

packages/core/src/modules/token-status-list/TokenStatusListOptions.ts
```diff
@@ -47,126 +46,57 @@ type CreateCwtTokenStatusListOptions = {
...
-}
+} & CreateBaseTokenStatusListOptions
 
 export type CreateTokenStatusListOptions = CreateJwtTokenStatusListOptions | CreateCwtTokenStatusListOptions
 
 export type TokenStatusListResult =
   | { format: 'cwt'; statusList: Uint8Array; parsed: StatusListCwt }
   | { format: 'jwt'; statusList: string; parsed: Jwt }
 
+type UpdateBaseTokenStatusListOptions = Omit<CreateBaseTokenStatusListOptions, 'statusList' | 'statusListUri'> & {
```
berendsliedrecht, line 77: Update cannot change the statuslisturi, does that make sense? We basically enforce that when you host it in another place, you have to create a new one.

packages/core/src/modules/token-status-list/TokenStatusListService.ts
```diff
@@ -41,13 +42,26 @@ export class TokenStatusListService {
...
         `signer method '${options.signer.method}' is not supported for creating a token status list. Only x5c is`
       )
     }
 
-    const [{ keyId }] = options.signer.x5c
+    let keyId: string | undefined
+    if (options.signer.method === 'x5c') {
+      keyId = options.signer.x5c[0].keyId
+    } else if (options.signer.method === 'did') {
+      const dids = agentContext.dependencyManager.resolve(DidsApi)
```
berendsliedrecht, line 55: Taken from sd-jwt-vc. Might be good to double check if this is the de-facto way to resolve a did here.

packages/core/src/modules/token-status-list/TokenStatusListOptions.ts
```diff
@@ -1,35 +1,104 @@
...
+  timeToLive?: number
 }
 
+type CreateCwtTokenStatusListOptions = {
+  format: 'cwt'
+
+  /**
+   * Additional claims to include in the token payload.
+   */
+  additionalPayload?: Map<number, unknown>
```
TimoGlastra: CWT does not require number keys right? You can also have string keys AFAIK

packages/core/src/modules/token-status-list/TokenStatusListOptions.ts
```diff
@@ -1,35 +1,104 @@
...
+  format: 'cwt'
+  token: Uint8Array
 
   /**
-   * Will determine whether it will be signed or authenticated
+   * Additional claims to include in the token payload.
    */
-  keyId: string
-}
+  additionalPayload?: Map<number, unknown>
```
TimoGlastra: same

packages/core/src/modules/token-status-list/TokenStatusListOptions.ts
TimoGlastra: yeah makes sense

packages/core/src/modules/token-status-list/TokenStatusListService.ts
```diff
@@ -31,118 +39,235 @@ import type {
...
-      options.aggregationUri
-    )
-    const kms = agentContext.dependencyManager.resolve(KeyManagementApi)
-    const jwk = await kms.getPublicKey({ keyId: options.keyId })
-    if (!jwk.alg) {
-      throw new CredoError(`Found JWK for key id '${options.keyId}', but did not find a required algorithm`)
+    // TODO: jwk could also be supported
+    if (options.signer.method !== 'x5c' && options.signer.method !== 'did') {
+      throw new Error(
+        `signer method '${options.signer.method}' is not supported for creating a token status list. Only x5c is`
```
TimoGlastra:
```suggestion
        `signer method '${options.signer.method}' is not supported for creating a token status list. Only x5c and did are`
```

packages/core/src/modules/token-status-list/TokenStatusListService.ts
```diff
@@ -31,118 +39,235 @@ import type {
...
+        const { header, payload } = createHeaderAndPayload(statusList, basePayload as { sub: string; iat: number }, {
+          alg: options.alg,
+          typ: 'statuslist+jwt',
+        })
+        const headerWithKeyReference = {
+          ...header,
+          ...(options.signer.method === 'x5c'
+            ? { x5c: options.signer.x5c.map((cert) => cert.toString('base64')) }
+            : {}),
+          ...(options.signer.method === 'did' ? { kid: options.signer.didUrl } : {}),
```
TimoGlastra: I think the common pattern is to use a relative didUrl and set the `iss` parameter. That is what we use in SD-JWT-VC, also what we use in Paradym for status lists (`iss` for did, but not fot x5c)

PR #2816, opened 2026-06-04
feat(core): X509 CRL verification

hacdias:
Re-opening this superseding #2729, so that @TimoGlastra can properly review and approve the PR.

- Adds support for verifying X509 CRLs.
- Improves extension validation, taking into account critical extensions.


packages/core/src/modules/x509/X509CertificateRevocationList.ts
```diff
@@ -0,0 +1,196 @@
...
+        return {
+          serialNumber: entry.serialNumber,
+          revocationDate: entry.revocationDate,
+          reason: entry.reason,
+        }
+      }
+    }
+
+    return null
+  }
```
hacdias, line 196: See https://github.com/PeculiarVentures/x509/pull/137.

packages/core/src/modules/x509/X509ValidationResult.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  /**
+   * Individual validation checks performed
+   */
+  validations: Partial<X509CertificateValidations>
+
+  /**
+   * Error that occurred during validation
+   */
+  error?: Error
```
TimoGlastra, line 61: How does this error relate to the individual errors, is it just the latest?

packages/core/src/modules/x509/X509ValidationOptions.ts
```diff
@@ -0,0 +1,80 @@
...
+   * @default {@link X509RevocationCheckMode.SoftFail}
+   *
+   * Note: Currently defaults to 'SoftFail' to provide enhanced security
+   * without breaking existing functionality in network failure scenarios
+   */
+  mode?: X509RevocationCheckMode
+
+  /**
+   * Whether to check revocation for all certificates in chain (except root)
+   * @default false (only check end-entity certificate)
```
TimoGlastra: is this a common default?

packages/core/src/modules/x509/X509ValidationOptions.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  /**
+   * Specific revocation reasons that must be covered by CRLs
+   * If a certificate has no reasons specified in its CRL distribution points (a "full" DP),
+   * it covers all reasons automatically.
+   *
+   * If a certificate only has distribution points with specific reasons, this array determines
+   * which reasons must be covered for the check to be considered complete.
+   *
+   * @default All revocation reasons must be covered
```
TimoGlastra: One thing I still wanted to change (or check) is that if you don't provide `requiredReasons` and the CRL e.g. only include one reason, based on the current wording that would be invalid, since the default is ` All revocation reasons must be covered`. But if the Certificate does not include the reason, why would we require it to be handled by default?

TimoGlastra: Ah i see this comment in the code:
>    // Default: require all reasons that are actually present in the certificate's DPs
>    // (no point requiring reasons the issuer never published DPs for)

Which makes sense. Is this actually the case in implementation, and then we need to update this comment?

packages/core/src/modules/x509/X509RevocationService.ts
```diff
@@ -0,0 +1,454 @@
...
+          const unsupportedReasons = X509RevocationService.reasonMaskToNames(unsupported.reasons & missingReasons)
+          errorMsg += ` Unsupported CRL feature ${unsupported.feature} covers missing reasons: [${unsupportedReasons.join(', ')}].`
+        }
+      }
+
+      if (fetchErrors.length > 0) {
+        errorMsg += ` Fetch errors: ${fetchErrors.join('; ')}`
+      }
+
+      throw new X509Error(errorMsg)
```
TimoGlastra: We throw an error here if the reasons are not covered. But in `handleRevocationError` we also set `isValid` to true if the mode is soft fail. The softfail mode should ONLY succeed in case of a network error. Currently it succeeds in other cases as well it seems?

packages/core/src/modules/x509/__tests__/X509RevocationService.test.ts
```diff
@@ -0,0 +1,338 @@
...
+
+      const result = await X509RevocationService.checkRevocation(agentContext, leaf, issuerCertificate, {
+        mode: X509RevocationCheckMode.Require,
+        requiredReasons: [X509RevocationReason.KeyCompromise],
+      })
+      expect(result.isValid).toBe(false)
+      expect(result.isRevoked).toBe(true)
+    })
+
+    it('fails Require but passes SoftFail when a required partitioned reason is unreachable', async () => {
```
TimoGlastra: Do we have some tests that cover SoftFail failing if another error than network error occurs?

TimoGlastra (approved):
Do we need to add some new service/API methods for working with CRLs?

packages/core/src/modules/x509/X509ValidationOptions.ts
hacdias: From what I found it's a common (historically) default in browsers, but RFC 5280 asks for all non-root certs to be checked. So I'll update the default.

packages/core/src/modules/x509/X509ValidationResult.ts
hacdias: It's a fail-fast validation, so this is the first error that occurred. If you think it's better to include them all, we can make it non fail-fast, but  from a performance POV, I wouldn't do it.

I'll update the comment in the meanwhile to make it clear that it's a fail-fast.

packages/core/src/modules/x509/X509ValidationOptions.ts
hacdias: Yes, this is a comment issue. Will update :D

packages/core/src/modules/x509/__tests__/X509RevocationService.test.ts
hacdias: Will add

packages/core/src/modules/x509/X509RevocationService.ts
hacdias: Will fix

hacdias:
@TimoGlastra I added new API calls, and also cleaned up the code a bit to make a better separation of the X509Service and the X509RevocationService. Also made some existent function parameters more consistent with nomenclature we already had, and fixed a few small issues.

Could you take another look?

packages/core/src/modules/x509/__tests__/x509CrlTestUtils.ts
```diff
@@ -136,3 +137,76 @@ export async function generateCrl(
...
+    subjectCommonName: string
+    serialNumber: string
+    notBefore: Date
+    notAfter: Date
+    /** Each entry becomes a partitioned distribution point covering the single given reason (a ReasonFlags bit index). */
+    distributionPoints: Array<{ url: string; reason: number }>
+  }
+): Promise<Uint8Array> {
+  const webCrypto = new CredoWebCrypto(agentContext)
+  x509.cryptoProvider.set(webCrypto)
```
TimoGlastra: This will override the global context. We have avoided this every where else and always provide the context via params since you can have multiple agentContext instances active at the same time. May cause weird issues in the future?

TimoGlastra (approved):

packages/core/src/modules/x509/__tests__/x509CrlTestUtils.ts
hacdias: Oh, will remove, iut's not needed. I had updated the calls to pass the webcrypto. These are just test utilities btw.

Issue #2196, opened 2025-02-18
Support Verifiable Credentials Data Model v2.0


jrhender:
As the VC v2.0 Data Model becomes more mature, I think it would be valuable to add support to Credo. For example, Credo isn't able to pass many of the tests on [canivc.com](https://canivc.com) without support for VC v2.0.

Perhaps we could start with MVP support of the mandatory properties and add support for optional properties gradually. We could start by:
1. Not [requiring the v1 context](https://github.com/openwallet-foundation/credo-ts/blob/c6369e4740c0d6dc430e9d354dc7fa1f39c7da5b/packages/core/src/modules/vc/validators.ts#L18)
2. Note requiring `issuanceDate` if it is a v2.0 VC

TimoGlastra:
Hey! Would be great to support this. Although i'm not sure if it's that simple?

The context is different for v2 meaning different are/are not defined. Also i think for v2 mostly DataIntegrityProof is used, and the rules around jwt vcs changed. 

I would rather create a v2 service and handle it separately and avoid creating an implementation that is not fully compliant to avoid interoperability issues.

What do you think?

jrhender:
Thanks for your thoughts @TimoGlastra 

> I would rather create a v2 service and handle it separately and avoid creating an implementation that is not fully compliant to avoid interoperability issues.

That seems like a reasonable, low risk approach to me 👍 

I'd like to have a go at the above at some point, but I don't have a timeline for doing so.

PR #2827, opened 2026-06-11
feat(vc2): implement VC 2.0 support

rmlearney-digicatapult:
Apologies I couldn't break this into smaller units for review but it wouldn't have been functional.

Large core element of fully spec-compliant VC Data Model 2.0.

This PR introduces:

**W3cV2 Model Changes for VC 2.0**

* Refined JSON model shapes
* Updated verify result envelope
* New shape detectors and guards for enveloped credentials and presentations for routing
* Fixed type error for `enveloped_verifiable_presentation_type`
* New codec file to extract conversion functions out of `W3cV2EnvelopedVerifiablePresentation.ts` and break a circular imports with the `Jwt` and `SdJwt` models

**W3cV2 JWT & SD-JWT alignment to VC-JOSE-COSE**

* Deterministic header/media-type routing, credential subject authentication
* Compact JWT envelope types have cty/type checks
* SD-JWT VC2 handlers with enforced challenge/domain binding
* Removed old `jwt-vc-presentation-profile.ts`
* VC 2.0 Verifiable Presentation Handling

**Full VC 2.0 + VC-JOSE-COSE multi-format nested VC/VP compliance**

* Outer envelope decoded and dispatched to format-specific service based on typed-matched presentation class (`jwt`, `sd-jwt`, `di`), not string matches
* Recursive JSON-LD tree walk extracting, dispatching and verifying all internal combinations of VC, Enveloped VC (`jwt`/`sd-jwt` outer) and Enveloped VP (`jwt`/`sd-jwt` outer) to comply with VC 2.0 and VC-JOSE-COSE
* Fully tested with 18-combination matrix in `matrixVpFixture.ts`

**VC 2.0 Routing & API**

* `W3cV2CredentialService` as full-featured top-level router capable of dispatching `jwt`, `sd-jwt`, `di` credential types
* Extended service options for multi-format support and DI sign/verify
* Wired DI VC support into API
* Extracted shared JWT utility helpers from service layer into `v2-jwt-utils.ts`
* Created equivalent for DI helpers named `v2-di-utils.ts`
* Centralised VC 2.0 @context and credential validity period validators into `validators.ts`

**JSON-LD Contexts for VC 2.0**

* New bundled context files & hashes per VC 2.0 and DI 1.0 specifications

**Data Integrity v1.0 Handling**

* New data-integrity-v1 handler using `w3c-di` proof module for VC 2.0 semantics
* @context compliance checks, proof purpose & challenge validation in VC subsystem
* DI credential models
* DI error code mapping
* DI credential signing/verification
* Full test coverage with production-like credential fixture in `fixtures/credo-di-vc.ts`

**Credential Record**

* Updated `W3cV2CredentialRecord.ts` to handle DI + cryptosuite tagging
* Schema extensions and database migrations

**DCQL, PEX, OIDC**

* Guards and hardening for `di_*` paths because not yet supported


rmlearney-digicatapult:
I implemented tree-walking support for VP-within-VP (per `VC-JOSE-COSE`) but it appears this is not directly supported by `VC 2.1` at this time (https://github.com/w3c/vc-data-model/issues/1633) 

I will remove this functionality for `VC 2.0` spec compliance.

jrhender:
This PR would close #2196

rmlearney-digicatapult:
Tests all passing locally with Docker :(

Looks like the E2E and Drizzle errors are arising from `indy-pool` building on `Bionic` which is EOL so package server potentially down

`Failed to fetch http://security.ubuntu.com/ubuntu/dists/bionic-security/InRelease  Connection failed [IP: 91.189.92.23 80]`

rmlearney-digicatapult:
Design question: This is currently a strict spec-compliant implementation. Do we want to extend options to allow user to inject custom `documentLoader` into the module at init?

The implementation also currently uses direct native `fetch` in `W3cV2DataIntegrityContextValidator.ts` but this is also the case for existing `vc/jsonld/nativeDocumentLoader.ts` and `nativeDocumentLoader.native.ts`

I could modify all of these sites to use injected fetch plus wrapper.

TimoGlastra:
Ideally we use the custom fetch from whenever possible. We didn't use it because the document loader didn't support it.

packages/core/src/modules/vc/data-integrity/W3cV2DataIntegrityContextValidator.ts
```diff
@@ -0,0 +1,311 @@
...
+      'Context hash verification failed (§2.4)',
+      `Context '${contextUri}' hash '${actualHash}' did not match expected '${expectedHash}'`
+    )
+  }
+
+  return undefined
+}
+
+async function getContextBytes(contextUrl: string): Promise<Uint8Array> {
+  const response = await fetch(contextUrl, {
```
genaris: Beware on using bare fetch (use the one from agent dependencies, either directly or through the fetchWithTimeout utility method).

packages/core/src/modules/vc/data-integrity/W3cV2DataIntegrityCredentialService.ts
```diff
@@ -0,0 +1,245 @@
...
+
+    const isValid = credentialStatus.isValid
+
+    return {
+      isValid,
+      validations: {
+        dataModel: { isValid: true },
+        signature: { isValid: true },
+        credentialStatus,
+        issuerIsSigner: { isValid: true },
```
genaris: I couldn't find the place where the binding of the proof's `verificationMethod` to the credential's `issuer`, and the same with `holderIsSigner` in `verifyPresentation`.

packages/core/src/modules/vc/W3cV2CredentialService.ts
```diff
@@ -101,34 +151,281 @@ export class W3cV2CredentialService {
...
+      if (!presentationResult.presentation.isValid) {
+        validationResults.isValid = false
+        return validationResults
+      }
+
+      signerId = presentation.resolvedPresentation.holderId
+      if (!signerId) {
+        validationResults.isValid = false
+        return validationResults
+      }
```
genaris, line 297: If holder is not present (optional in VCDM 2.0), wouldn't this always fail?

packages/core/src/modules/vc/data-integrity/W3cV2DataIntegrityContextValidator.ts
```diff
@@ -0,0 +1,311 @@
...
+function hasBundledContext(contextUrl: string): boolean {
+  if (contextUrl in DEFAULT_CONTEXTS) return true
+  const withoutFragment = contextUrl.split('#')[0]
+  return withoutFragment in DEFAULT_CONTEXTS
+}
+
+let cachedContextValidationDocumentLoader: DocumentLoader | undefined
+
+async function getContextValidationDocumentLoader(): Promise<DocumentLoader> {
+  if (cachedContextValidationDocumentLoader) return cachedContextValidationDocumentLoader
```
genaris: Is it intended to have this scoped document loader (cached on module-owned variable)? I mean, is it fine to have a separate loader configured for W3CCredentialsModule? There we can define a custom loader that wouldn't be used here.

packages/core/src/modules/vc/W3cV2CredentialsModule.ts
genaris, line 1: `W3cV2CredentialsModule.ts` declares `W3cV2CredentialsModuleConfigOptions` inline and stores this.options. Its direct 1.1 sibling does it conventionally:

```ts
export class W3cCredentialsModule implements Module {
  public readonly config: W3cCredentialsModuleConfig
  public constructor(config?: W3cCredentialsModuleConfigOptions) {
    this.config = new W3cCredentialsModuleConfig(config)
  }
```

Related: `W3cV2DataIntegrityContextValidator.configure()` is a mutable setter on an `@injectable()` class, called once at registration. A config class passed to the constructor would match the codebase and remove the mutable-singleton shape.

packages/core/src/modules/vc/data-integrity/index.ts
genaris, line 1: The directory's `v1` means Data Integrity 1.0, but adding that suffix does not make it "future proof", since the classes within it do only include the "V2" that corresponds to the VCDM 2.0. I think we can simply leave the directory as `data-integrity` and, in case a future major revision of DI is created, we can rename the classes to take into account the DI version (if actually needed).

genaris (approved):
Thanks @rmlearney-digicatapult ! I left some comments and minor suggestions but overall I think it's a great addition. I'd like to see it soon in main and that's why I'm submitting my approval!

packages/core/src/modules/vc/data-integrity/W3cV2DataIntegrityContextValidator.ts
rmlearney-digicatapult: have used method from agent dependencies

packages/core/src/modules/vc/W3cV2CredentialsModule.ts
rmlearney-digicatapult: have modified to match 1.1

packages/core/src/modules/vc/data-integrity/index.ts
rmlearney-digicatapult: have renamed

packages/core/src/modules/vc/data-integrity/W3cV2DataIntegrityContextValidator.ts
rmlearney-digicatapult: Have now changed to used the shared `JsonLdModule` for context validation

packages/core/src/modules/vc/W3cV2CredentialService.ts
rmlearney-digicatapult: Logic now supports missing holder

rmlearney-digicatapult:
Have addressed all comments from @genaris in sequential commits for ease of review

genaris (changes requested):
The creation of a shared `JsonLdModuleConfig` seems a good step to me. However I'm wondering about two things in how it's wired:

- `W3cV2CredentialsModule.register()` resolves `JsonLdModuleConfig` eagerly and bakes the instance into the validator. Since custom modules register after the default ones, a user overriding the loader via a standalone `JsonLdModule` would be picked up by `W3cJsonLdCredentialService` (lazy injection) but not here. Is there a reason not to make the validator a regular `registerSingleton` with constructor-injected `JsonLdModuleConfig` + `W3cV2CredentialsModuleConfig`, like the other services?

- `getContextValidationDocumentLoader()` short-circuits all of DEFAULT_CONTEXTS before consulting the configured loader, but `defaultDocumentLoader` already serves bundled contexts first, so this only has an effect when a custom loader is set, where it prevents that loader from overriding bundled contexts (the 1.1 path passes the configured loader wholesale). Couldn't we just use `jsonLdModuleConfig.documentLoader(agentContext)` directly so both VCDM versions behave the same? The §2.4 bundled-only trust check wouldn't be affected, as it doesn't go through the loader.

In addition to this, note that there is an error on Validate workflow due to a missing export related to `WebVhAnonCredsRegistry`.

rmlearney-digicatapult:
@genaris think I've managed to address your comments

packages/core/src/modules/vc/W3cV2CredentialsModule.ts
```diff
@@ -1,22 +1,46 @@
...
 import { W3cV2SdJwtCredentialService } from './sd-jwt-vc'
 import { W3cV2CredentialService } from './W3cV2CredentialService'
 import { W3cV2CredentialsApi } from './W3cV2CredentialsApi'
+import type { W3cV2CredentialsModuleConfigOptions } from './W3cV2CredentialsModuleConfig'
+import { W3cV2CredentialsModuleConfig } from './W3cV2CredentialsModuleConfig'
 
 /**
  * @public
  */
+@injectable()
```
genaris:
```suggestion
```

genaris (approved):
LGTM

packages/core/src/modules/vc/W3cV2CredentialsModule.ts
```diff
@@ -1,22 +1,45 @@
 import type { DependencyManager, Module } from '../../plugins'
+import { injectable } from '../../plugins'
```
genaris:
```suggestion
```

PR #2831, opened 2026-06-12
fix(mdoc): compare X.509 chains between the mdoc, status list and identifier list

berendsliedrecht:


packages/core/src/modules/mdoc/Mdoc.ts
```diff
@@ -192,19 +192,49 @@ export class Mdoc {
...
+
+      if (
+        trustedIdentifierListChain &&
+        !x509ChainsAreEqual(
+          trustedIdentifierListChain.map((c) => X509Certificate.fromRawCertificate(c)),
+          issuanceChain
+        )
+      ) {
+        throw new MdocError('Trusted identifier list chain does not match the trusted issuance chain')
+      }
```
TimoGlastra: We only want this checks if there's NO status certificates right and we default to the issuance certificates? 

Because with LoTE the certificates can be different. I don't see a conditional check here, but maybe I'm missing something?

TimoGlastra: - issuance and status certificates -> use issuance for issuance, status for status, chains/keys do not have to match
- issuance, but no status certificates -> use issuance for issuance, require status to match chain of issuance

packages/core/src/modules/token-status-list/TokenStatusListService.ts
```diff
@@ -158,13 +158,26 @@ export class TokenStatusListService {
     options: UpdateTokenStatusListOptions
   ): Promise<Extract<TokenStatusListResult, { format: Format }>> {
     // TODO: jwk could also be supported
-    if (options.signer.method !== 'x5c') {
+    if (options.signer.method !== 'x5c' && options.signer.method !== 'did') {
       throw new Error(
-        `signer method '${options.signer.method}' is not supported for creating a token status list. Only x5c is`
+        `signer method '${options.signer.method}' is not supported for updating a token status list. Only x5c and did`
```
TimoGlastra:
```suggestion
        `signer method '${options.signer.method}' is not supported for updating a token status list. Only x5c and did are`
```

TimoGlastra (changes requested):

packages/core/src/modules/mdoc/Mdoc.ts
TimoGlastra: Now the status certificates don't add anything. If the chains always have to match.

packages/core/src/modules/mdoc/Mdoc.ts
```diff
@@ -216,24 +216,36 @@ export class Mdoc {
...
+
+      if (!hasDedicatedStatusCertificates) {
+        if (
+          trustedIdentifierListChain &&
+          !x509ChainsAreEqual(
+            trustedIdentifierListChain.map((c) => X509Certificate.fromRawCertificate(c)),
+            issuanceChain
+          )
+        ) {
+          throw new MdocError('Trusted identifier list chain does not match the trusted issuance chain')
```
TimoGlastra: Should we mention in the error, "and no trusted status certificates were provided for the trusted issuance certificate"?

```
          throw new MdocError('Trusted identifier list chain does not match the trusted issuance chain, and no trusted status certificates were provided for the trusted issuance certificate')
```

Same for status list error below

packages/core/src/modules/mdoc/Mdoc.ts
```diff
@@ -192,19 +192,61 @@ export class Mdoc {
...
+        )
+
+      const x509ChainsAreEqual = (a: X509Certificate[], b: X509Certificate[]) =>
+        a.length === b.length && a.every((cert, i) => cert.equal(b[i]))
+
+      const issuanceChain = trustedIssuanceChain.map((c) => X509Certificate.fromRawCertificate(c))
+
+      if (!x509ChainsAreEqual(certificateChain, issuanceChain)) {
+        throw new MdocError('Certificate chain does not match the trusted issuance chain')
+      }
```
TimoGlastra, line 234: This is not correct i think? The mDoc certificate chain should NEVER contain the ROOT certificate. So if we create the chain with the root certificate it will always fail

packages/core/src/modules/mdoc/Mdoc.ts
```diff
@@ -192,19 +192,61 @@ export class Mdoc {
...
-          disableStatusValidation: false,
-          now: options?.now,
-        },
-        mdocContext
+      const { trustedIssuanceChain, trustedStatusListChain, trustedIdentifierListChain } =
+        await Holder.verifyIssuerSigned(
+          {
+            trustedCertificates: convertedTrustedCertificates.map(({ issuance, status }) => ({
+              issuance: issuance.map((cert) => X509Certificate.fromEncodedCertificate(cert).rawCertificate),
+              status: status?.map((cert) => X509Certificate.fromEncodedCertificate(cert).rawCertificate),
```
TimoGlastra: Not sure this will have the desired behavior. If the user provides no `status` certificates. It means the mdoc library will throw right? Because it REQUIRES status:

https://github.com/openwallet-foundation-labs/mdoc-ts/blob/main/src/mdoc/models/issuer-auth.ts#L123-L127

So if the user does not provide `status` (so undefined, empty array we leave as is), we should set the value to the same as the `issuance`. Then later we do a check to ensure they use the same chain

packages/core/src/modules/mdoc/__tests__/mdocServer.test.ts
```diff
@@ -346,6 +346,117 @@ describe('mdoc service test', () => {
...
+      statusInfo: { index: 1, uri: statusListUri },
+    })
+
+    const result = await mdoc.verify(agentContext, {
+      trustedCertificates: [
+        { issuance: [certificate.toString('base64'), certificate2.toString('base64')] },
+      ],
+    })
+
+    expect(result.isValid).toBe(false)
```
TimoGlastra: Can we check the actual error, because I'm not sure this is because of the same key, i think the mdoc library is already erroring before, even if the certificate would be the same (i also see no test where we check it succeeds when the same key/chain is used but no status is provided)

packages/core/src/modules/mdoc/__tests__/mdocServer.test.ts
```diff
@@ -346,6 +346,117 @@ describe('mdoc service test', () => {
...
+      namespaces: { hello: { world: 'world' } },
+      issuerCertificate: certificate,
+      validityInfo: { validUntil: nextDay },
+      statusInfo: { index: 1, uri: statusListUri },
+    })
+
+    const result = await mdoc.verify(agentContext, {
+      trustedCertificates: [{ issuance: [certificate.toString('base64')], status: [certificate.toString('base64')] }],
+    })
+
```
TimoGlastra, line 376: can we add tests where the root certiifcate is used as the trusted certiifcate? I think that's why your tests are passing on the check certificateChain == issuanceChain, but that's not correct

packages/core/src/crypto/contexts/mdocContext.ts
```diff
@@ -58,14 +58,16 @@ export const getMdocContext = (agentContext: AgentContext, { now }: { now?: Date
...
         ) as [string, ...string[]]
 
+        const validatedChain = await X509Service.validateCertificateChain(agentContext, {
+          certificateChain,
+          trustedCertificates,
+          verificationDate: input.now ?? now,
+        })
+
+        // X509Service.validateCertificateChain returns the chain root-first. The mdoc context contract
+        // requires the chain to be leaf-first with the trust anchor (root) as the last entry, so reverse here.
```
berendsliedrecht, line 68: @TimoGlastra would you suggest to swap the ordering for the mdoc lib, or keep it as-is?

TimoGlastra: Hmm, not sure what is common? I'm used to leaf first from jose. Is it common to use root first with mdoc/x509?

berendsliedrecht: I think leaf first makes the most sense according to: https://claude.ai/share/81554a87-8b85-456c-ac95-e289664d2501 which gives quite some examples for leaf-first, like openssl.

I do think it would be breaking though.

TimoGlastra (approved):

PR #2839, opened 2026-06-18
feat(didcomm): migrate message pickup to 4.0

tarunvadde:
Migrates message pickup 3.0 -> 4.0 (https://didcomm.org/message-pickup/4.0).

- Renames PIURIs to message-pickup/4.0, `limit` -> `message_count_limit`, drops `longest_waited_seconds`, status timing as epoch ms.
- Empty queue answers with an empty `delivery` instead of a status; an empty delivery signals completion.
- messages-received is fire-and-forget (no return_route, no status reply); the recipient drains via delivery-requests.
- Adds the message-pickup/4.0/problem-report for live-mode-not-supported.
- Bumps the pickup strategy enum to PickUpV4 (+ drizzle migration).
- Surfaces v2 return_route as a top-level header instead of nesting it in the body.


packages/didcomm/src/modules/routing/DidCommMediationRecipientApi.ts
```diff
@@ -304,18 +304,18 @@ export class DidCommMediationRecipientApi {
           })
         return subscription
       }
-      case DidCommMediatorPickupStrategy.PickUpV3: {
+      case DidCommMediatorPickupStrategy.PickUpV4: {
```
genaris, line 307: I think there could be an issue in the non-live PickUp V4 mode: it calls `pickupMessages` without the await completion flag (which is fine), but it will not complete the pickup if the amount of queued messages is higher than the batch size (config.maximumMessagePickup).

Suppose we have 11 messages in the queue, and the batch size is 10: `pickupMessages()` will create a `status-request` message, wich will be replied by a `status` message with messageCount = 11. Then, client will send a `delivery-request` with limit 10, and mediator will send back a `delivery` with the first 10 messages. Upon the reception, the client will send a `messages-received` with the IDs of all those 10 messages... and that's it.

Eventually, the pickup process will be completed, since a new pickup loop will be executed after `mediatorPollingInterval` seconds, and that remaining message will be delivered.

In Message Pickup V1, we followed a conservative approach, that will always create a new pickup message unless the received batch is empty. I'm not sure if it would be a good fit here, since this might work like that **only** if we are NOT on Live mode (otherwise, we will be forced to exchange extra messages every time a single message is received). If there is an elegant way to tell if we are using a live mode or not, we can handle this case like that. Otherwise I'd just leave this limitation as it is but comment it somewhere (and probably create an issue to address it better in a future version).

genaris (approved):
I think it is fine, but I have a concern about a particular case (PickUp V4 in non-live mode). Probably the best would be to just accept it like it is and just document the lmitation within the comments + raise an issue.

packages/didcomm/src/modules/routing/DidCommMediationRecipientApi.ts
tarunvadde: Thanks Ariel, good thing to flag. I traced it and it actually does drain the whole queue in a single pickup, so it doesn't wait for the next poll.

The v4 loop isn't status-driven anymore. `messages-received` is fire-and-forget (no `return_route`, mediator replies with nothing), and after each non-empty delivery the delivery handler sends the ack and then a fresh `delivery-request`. So it keeps requesting until it gets an empty delivery, which is the completion signal. Basically the conservative V1 behavior you described, just expressed through delivery-requests instead of status.

I validated your exact case: queued 11 messages with the default batch of 10 and ran one `pickupMessages()`. The client sends 3 delivery-requests:

```
status-request -> status(11)
-> delivery-request -> delivery(10) -> messages-received
-> delivery-request -> delivery(1)  -> messages-received
-> delivery-request -> (empty delivery -> completed)
```

and the queue ends empty. I confirmed it both by the drain (count goes to 0) and by counting the actual messages on the wire.

For live mode I used the delivery `thid` to tell the two apart: pushed messages have no `thid`, so they get acked but don't trigger a follow-up request (no extra round-trips, which was your concern there). Polled deliveries carry a `thid` and continue the drain.

genaris: Oh, you are right! I haven't properly checked `DidCommMessagDeliveryV4Handler` code.

PR #2842, opened 2026-06-23
test: setup oid4vc conformance testing

TimoGlastra:
Sets up OIDC conformance testing for OID4VP HAIP in CI for Credo.

Two tests are failing (allowed to fail for now), which can be resolved once https://github.com/openwallet-foundation/sd-jwt-js/pull/378 is addressed and we add iat too far in past check in Credo.


berendsliedrecht (approved):
I think this is a very good addition! 

I do wonder about the performance impact and I hope it will not be flaky for some reason as we have had issues with that in the past. 

I see a timeout for 1800000ms which is 30minutes, if that is close to the test time that would be unfortunate. Maybe we can do it before a patch release only? Or only on oid4vc package changes? Would love to have the same for the cheqd e2e tests..

TimoGlastra:
Yeah let's enable it, and see how it goes. It's not marked as required, so it won't impact mergability of PRs. That way we can see if we enable it for a while, how flaky it is.

hacdias:
> I see a timeout for 1800000ms which is 30minutes, if that is close to the test time that would be unfortunate.

Could we see how long they take? On this PR it seems to have been just a minute. Maybe we can decrease the timeout if we don't expect it to take so long. Otherwise, I think @berendsliedrecht's idea of running it when there's changes in the oid4vc package (and maybe core also) is a good one.

.github/workflows/conformance.yml
```diff
@@ -0,0 +1,78 @@
...
+# OIDF conformance suite. Pulls the conformance suite images (MongoDB + Java
+# server + nginx) via testcontainers and drives the credo-ts verifier against
+# them. Only runs on changes to the core and openid4vc packages since it is
+# heavy and only exercises those modules.
+on:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - 'packages/core/**'
```
berendsliedrecht, line 12: Wait can we just do this for cheqd as well? Only run e2e tests with cheqd when that package is updated? Or would that be not enough as we might update something in core that could break the  cheqd package?

Issue #2607, opened 2026-01-07
Runtime issue related to using credo from ESM project


Patrik-Stas:
Creating isolated issue for issue  discovered as part of https://github.com/openwallet-foundation/credo-ts/issues/2579

When project is using ESM modules and credo, it leads to runtime issues. 

For reproduction checkout commit `13df6845bc4dc8c1671bf00698deed1658d10473` from repo https://github.com/Patrik-Stas/tsx-credo-06-reproduction/tree/main

If you run `pnpm run test` you should see in output 
```
[ERROR] DID not created, didCreateResult: {
  "didDocumentMetadata": {},
  "didRegistrationMetadata": {},
  "didState": {
    "state": "failed",
    "reason": "unknownError: Error creating key"
  }
}
[ERROR] ❌ Script failed with error
```

but if you remove `type: module`  it will succeed

```
[INFO] ✅ DID key created successfully
[INFO]   DID: did:key:z6MkfnXzhM8iivsgd9774ZHCqKcXDkBEVnY9jwb2BQUFWqfH
[INFO]   KID: did:key:z6MkfnXzhM8iivsgd9774ZHCqKcXDkBEVnY9jwb2BQUFWqfH#z6MkfnXzhM8iivsgd9774ZHCqKcXDkBEVnY9jwb2BQUFWqfH
[INFO] ✅ Script completed successfully
```

The repo is using simplest possible agent
```
    const agent = new Agent({
      config: {
        logger: new SimpleLogger(logLevel),
      },
      dependencies: agentDependencies,
      modules: {
        askar: new AskarModule({
          askar: askar,
          store: {
            id: walletId,
            key: walletKey,
          },
          enableKms: true,
          enableStorage: true, // Use Askar's default SQLite storage
        }),
      },
    });
```

TimoGlastra:
Hi @Patrik-Stas,

I think this issue is caused by #https://github.com/openwallet-foundation/credo-ts/issues/2597.

Somehow ESM makes the registration of Askar work differently. For now I can make your repo work by adding the import of Askar nodejs to the top of the file. 

Generally we throw error and the full stack trace would show this, but we don't log the error in the key did registrar. We should change that.

We are about to release Askar wrapper 0.5.x, which has a fix for this. But there's still some issues with the new FFI library we're using.

genaris:
> We are about to release Askar wrapper 0.5.x, which has a fix for this. But there's still some issues with the new FFI library we're using.

[...]

Issue #2597, opened 2025-12-16
Possible Askar Key Management issue after migrating to v0.6.1


Sshovon:
I am trying to migrate my agents to Credo v0.6.1, but I am encountering issues related to Askar key management. On this version, the agent is unable to create new keys or access existing keys to any ops. When attempting to create a new key, the process fails with a `KeyManagementError: Error creating key, caused by a TypeError: Cannot read properties of undefined (reading 'keyGetJwkSecret')` originating from `AskarKeyManagementService.privateJwkFromKey`.

```
KeyManagementError: Error creating key
    at AskarKeyManagementService.createKey (file:///home/frcs/my-zone/temp-credo-agent/node_modules/@credo-ts/askar/src/kms/AskarKeyManagementService.ts:250:13)
    at KeyManagementApi$1.createKey (file:///home/frcs/my-zone/temp-credo-agent/node_modules/@credo-ts/core/src/modules/kms/KeyManagementApi.ts:81:27)
    at Function.build (/home/frcs/my-zone/temp-credo-agent/holder.ts:42:48)
    at main (/home/frcs/my-zone/temp-credo-agent/holder.ts:89:5) {
  [cause]: TypeError: Cannot read properties of undefined (reading 'keyGetJwkSecret')
      at AskarKeyManagementService.privateJwkFromKey (file:///home/frcs/my-zone/temp-credo-agent/node_modules/@credo-ts/askar/src/kms/AskarKeyManagementService.ts:721:13)
      at AskarKeyManagementService.publicJwkFromKey (file:///home/frcs/my-zone/temp-credo-agent/node_modules/@credo-ts/askar/src/kms/AskarKeyManagementService.ts:710:45)
      at AskarKeyManagementService.createKey (file:///home/frcs/my-zone/temp-credo-agent/node_modules/@credo-ts/askar/src/kms/AskarKeyManagementService.ts:235:30)
      at KeyManagementApi$1.createKey (file:///home/frcs/my-zone/temp-credo-agent/node_modules/@credo-ts/core/src/modules/kms/KeyManagementApi.ts:81:27)
      at Function.build (/home/frcs/my-zone/temp-credo-agent/holder.ts:42:48)
      at main (/home/frcs/my-zone/temp-credo-agent/holder.ts:89:5)
}

```
[...]

PR #2843, opened 2026-06-23
fix(askar): use NativeAskar.instance instead of the deprecated askar export

tarunvadde:
The KMS reads the askar binding through the deprecated `askar` export from `askar-shared`. That binding is only set as a side effect of importing `askar-nodejs`, so if anything loads the KMS first (ESM, bundlers, test runners) it's still `undefined` and key ops throw `Cannot read properties of undefined (reading 'keyGetJwkSecret')`.

This switches `AskarKeyManagementService` and the ECDH derive helpers (plus the two tests that used it) to `NativeAskar.instance`, the getter added in askar 0.6.0 ([openwallet-foundation/askar-wrapper-javascript#77](https://github.com/openwallet-foundation/askar-wrapper-javascript/issues/77) / [#78](https://github.com/openwallet-foundation/askar-wrapper-javascript/pull/78)), which resolves the binding on each call so import order stops mattering.

`NativeAskar` only exists from 0.6.0, so the `askar-shared` peer is narrowed to `^0.6.0` and the changeset is `minor`. Flagging that this drops pre-0.6.0 support, in case anyone still pins an older askar.

Relates to #2597, #2607.

### DIFF
--- a/packages/askar/src/kms/crypto/deriveKey.ts
+++ b/packages/askar/src/kms/crypto/deriveKey.ts
@@ -1,5 +1,5 @@
 import { Kms, TypedArrayEncoder } from '@credo-ts/core'
-import { askar, Key } from '@openwallet-foundation/askar-shared'
+import { Key, NativeAskar } from '@openwallet-foundation/askar-shared'
 import { jwkEncToAskarAlg } from '../../utils'
 
 export const askarSupportedKeyAgreementAlgorithms = [
@@ -47,7 +47,7 @@ export function deriveEncryptionKey(options: {
       : undefined
 
   const derivedKey = new Key(
-    askar.keyDeriveEcdhEs({
+    NativeAskar.instance.keyDeriveEcdhEs({
       algId: TypedArrayEncoder.fromUtf8String(
         keyAgreement.algorithm === 'ECDH-ES' ? encryption.algorithm : keyAgreement.algorithm
       ),
@@ -130,7 +130,7 @@ export function deriveDecryptionKey(options: {
       : undefined
 
   const derivedKey = new Key(
-    askar.keyDeriveEcdhEs({
+    NativeAskar.instance.keyDeriveEcdhEs({
       algId: TypedArrayEncoder.fromUtf8String(
         keyAgreement.algorithm === 'ECDH-ES' ? decryption.algorithm : keyAgreement.algorithm
       ),
--- a/.changeset/askar-native-instance.md
+++ b/.changeset/askar-native-instance.md
@@ -1,0 +1,7 @@
+---
+'@credo-ts/askar': minor
+---
+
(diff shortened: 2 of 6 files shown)


packages/askar/src/kms/crypto/deriveKey.ts
```diff
@@ -1,5 +1,5 @@
 import { Kms, TypedArrayEncoder } from '@credo-ts/core'
-import { askar, Key } from '@openwallet-foundation/askar-shared'
+import { Key, NativeAskar } from '@openwallet-foundation/askar-shared'
```
TimoGlastra, line 2: I think we should try to read Askar from the module config: https://github.com/openwallet-foundation/credo-ts/blob/main/packages/askar/src/AskarModuleConfig.ts#L130

- inject Moduleconfig on agent context
- extract askar instance from it
- We should probably update the `askar` getter in the module config, to see if Askar is passed or the `NativeAskar`?

tarunvadde: Makes sense, reading the injected instance from `AskarModuleConfig` is cleaner than the global getter,.

Two things I want to confirm before I push it:

1. **Scope.** `deriveKey`, `fetchAskarKey`, and encrypt/decrypt all have `agentContext`, so those are straightforward. But `keyFromJwk`, `privateJwkFromKey`, and `publicJwkFromKey` don't receive `agentContext`, and they're called from ~13 spots across `getPublicKey`/`createKey`/`sign`/`verify`/`encrypt`/`decrypt`. Doing this consistently means threading the resolved instance through those helpers too (doing it only in `deriveKey` would leave the JWK paths on the global, which defeats the purpose). Good with that?

2. **The getter.** `askar` is currently a required option, so `config.askar` always resolves and a fallback isn't strictly needed. Do you want me to make it optional and fall back to `NativeAskar.instance` when it isn't passed? Heads up that this decides the peer range: with the fallback we keep referencing `NativeAskar`, so `askar-shared` stays at `^0.6.0` (changeset `minor`); if `askar` stays required, nothing references `NativeAskar` anymore and the peer can go back to the broad range (`patch`).

genaris (commented):
@TimoGlastra I think if we are going to merge the drop node 20 soon and move to 0.8.0, it would be great if we can look into this (dropping older versions of askar as well).

TimoGlastra:
Yes let's do it

TimoGlastra:
So let's first release 0.7.1 and then we move towards 0.8.

TimoGlastra:
I have updated the PR to be backwards compatible, and also made the same change for Anoncreds. This can still land in 0.7.x (i will backport once merged)

genaris (approved):
OK let's merge it now, just before the first breaking change commit, so it will be already present in v0.7.2.

PR #2844, opened 2026-06-24
feat: add trusted entities for verification callback

TimoGlastra:
This adds a new callback to retrieve trusted entities for a verification callback. It's an continuation of the get trusted certificates for verification, but applied more broadly to also support DIDs, and possibly other mechanisms to trust an entity in the future.

The X509 callbacks are kept in place, to not cause a breaking change. the new trusted entitites callback takes precendence. In 0.8 we can remove the x509 callback, it should be very straigtforward to migrate from the x509 callback to the general callback

There's some work in merging with main, as PRs were merged for mdoc status certificates, and we need to integrate this new structure into the general callbacks as well (issuance +status certificates). For DIDs I'm not sure if we want to allow issuance dids + status dids?

PR is still very much WIP, needs testing, and needs to be integrated in the other parts of the codebase as well (where we don't have X509 but do have did support, like W3C credentials).


packages/core/src/agent/TrustedIssuerContext.ts
```diff
@@ -0,0 +1,97 @@
...
+  VerificationSigner,
+} from './TrustedIssuersForVerification'
+
+type TrustedIssuerForSigner<SignerMethod extends VerificationSigner['method']> = Extract<
+  TrustedIssuer,
+  { method: SignerMethod }
+>
+
+// biome-ignore lint/complexity/noStaticOnlyClass: no explanation
+export class TrustedIssuerContext {
```
hacdias, line 20: I'm a bit unsure about the name of this class.

hacdias:
@TimoGlastra I further worked the PR with the following changes:

- Rebased
- W3C V1 and V2 JWT/SD-JWT credential verification now enforce trust using the new method.
- Added tests for SD-JWTs, mdoc, W3C JWT/SD-JWT
- Updated callbacks to use the new methods, we should bne able to simplify the code a lot once the old x509 function is gone 

Not sure if there are places where it might be missing (DIDComm stuff?)?

.changeset/trusted-issuers-for-verification.md
```diff
@@ -0,0 +1,6 @@
+---
+"@credo-ts/core": patch
+"@credo-ts/openid4vc": patch
+---
+
+Add a global `getTrustedIssuersForVerification` agent callback for resolving trusted issuers during verification. Unlike the now-deprecated X.509-module `getTrustedCertificatesForVerification` callback (which it takes precedence over), it supports both X.509 certificate chains and DIDs and is extensible to other trust mechanisms. It is wired into SD-JWT VC, mdoc, W3C JWT and OpenID4VP verification.
```
TimoGlastra: not JSON-LD? That also uses DIDs

packages/core/src/agent/TrustedIssuerContext.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  public static async ensureTrustedSigner<
+    Signer extends VerificationSigner,
+    AdditionalVerificationTypes extends { type: string } = never,
+  >(
+    agentContext: AgentContext,
+    context: TrustedIssuersForVerificationContext<Signer, AdditionalVerificationTypes>,
+    /**
+     * The trusted issuer to verify the signer against. If not provided
+     * the signer will be dynamically fetched.
```
TimoGlastra:
```suggestion
     * The trusted issuers to verify the signer against. If not provided
     * the issuers will be dynamically fetched.
```

Should be issuers no?

packages/core/src/agent/TrustedIssuersForVerification.ts
```diff
@@ -0,0 +1,88 @@
...
+  /** A DID that is trusted as an issuer */
+  did: string
+}
+
+export interface TrustedIssuerX509 {
+  method: 'x509'
+
+  /** Base64 DER or PEM encoded trusted root certificate */
+  certificate: string
+}
```
TimoGlastra, line 47: We should update this based on the new structure also used for the X509 verification callback:

```
{
  "issuance": [...certs...],
  "status": [...certs...]
}
```

packages/core/src/agent/TrustedIssuersForVerification.ts
```diff
@@ -0,0 +1,88 @@
...
+  certificateChain: X509Certificate[]
+}
+
+export type VerificationSigner = VerificationSignerDid | VerificationSignerX509
+
+export interface TrustedIssuerDid {
+  method: 'did'
+
+  /** A DID that is trusted as an issuer */
+  did: string
```
TimoGlastra: not sure if we need to differentiate between issuance and status dids? or if that's more a certificate thingy

packages/core/src/modules/sd-jwt-vc/decodeSdJwtVc.ts
```diff
@@ -18,10 +18,21 @@ export function decodeSdJwtVc<
   const { jwt, disclosures, kbJwt } = decodeSdJwtSync(compactSdJwtVc, sdJwtVcHasher)
   const prettyClaims = getClaimsSync(jwt.payload, disclosures, sdJwtVcHasher)
 
+  // Decoding is structural and best-effort: a credential whose issuer cannot be resolved to a
+  // supported signing method (did/x5c) is still decoded, with `issuer` left undefined. Verification
+  // re-parses the issuer and enforces a supported method.
```
TimoGlastra: I guess it makes sense, can otherwise be error prone, but did you run into issues? we don't do that for other things that fail parsing.

packages/core/src/modules/x509/utils/convertLegacyTrustedCertificates.ts
```diff
@@ -7,3 +8,11 @@ export function convertLegacyTrustedCertificates(
     ? trustedCertificates.map((tc) => ({ issuance: [tc] }))
     : (trustedCertificates as X509VerificationTrustedCertificates[])
 }
+
+export function legacyTrustedCertificatesToTrustedIssuers(
+  trustedCertificates: string[] | X509VerificationTrustedCertificates[]
+): TrustedIssuerX509[] {
+  return convertLegacyTrustedCertificates(trustedCertificates)
+    .flatMap(({ issuance }) => issuance)
```
TimoGlastra: loses all status certificates

packages/openid4vc/src/shared/OpenId4VcTrustedIssuersVerificationTypes.ts
```diff
@@ -0,0 +1,66 @@
...
+  }
+}
+
+/**
+ * Credential verification bound to an OpenID4VP verification session.
+ * Use this instead of the core `VerificationTypeCredential` when the credential
+ * is verified as part of an OpenID4VP presentation flow.
+ */
+export type OpenId4VcVerificationTypeOid4VpCredential = {
+  type: 'oid4VpCredential'
```
TimoGlastra: inconsistent with the openid4vci types.

packages/core/src/agent/TrustedIssuersForVerification.ts
hacdias: Ah, yes. I missed that on the rebase.

packages/core/src/modules/vc/linked-data-proofs/W3cJsonLdCredentialService.ts
```diff
@@ -106,6 +107,14 @@ export class W3cJsonLdCredentialService {
     try {
       const verifyCredentialStatus = options.verifyCredentialStatus ?? true
 
+      // Ensure the issuer is trusted according to the (optional) `getTrustedIssuersForVerification`
+      // callback. For did-based issuers this is a no-op when no trusted issuers are configured,
+      // preserving the previous "trust any valid signature" behavior. Throws when the issuer is not trusted.
+      await TrustedIssuerContext.ensureTrustedSigner(agentContext, {
+        signer: { method: 'did', didUrl: options.credential.issuerId },
```
hacdias: Just to confirm: is the issuer ID reliably the DID in JSON-LD credentials?

packages/core/src/modules/sd-jwt-vc/decodeSdJwtVc.ts
hacdias: We have many tests that would otherwise fail because they do not set an `iss` ([see here](https://github.com/openwallet-foundation/credo-ts/actions/runs/28223938388/job/83611331439?pr=2844)). At the moment, when determining `issuer`, we fail if it's neither X509 or DID.

An alternative approach is to return `undefined` when `iss` is not set, and only fail if `iss` is set but unsupported. This way we can assume that that `iss` not being defined means there's no issuer. However, that's not always the case since some signing methods might not force you to set it (e.g.: with X509 you don't need to set `iss`).

Another option is to update all existent tests and break behavior (major release).

packages/core/src/agent/TrustedIssuersForVerification.ts
hacdias: Good question. I searched a bit and consulted Claude, but I couldn't find a clear answer. Using a different key from the same DID is of course fine. But a completely different DID doesn't seem to be supported. It seems _technically_ possible, but there's no formal way to establish trust between a DID credential signer and a DID status list signer:

> A different DID entirely is where it gets thin. For X.509 the spec gives you a concrete way to bind a separate status signer to the issuer: a dedicated id-kp-oauthStatusSigning EKU plus the recommendation that the Issuer's certificate and the Status Issuer's certificate be issued by the same Certificate Authority and the Status Issuer's certificate utilize that extended key usage. There is no DID equivalent in the TSL spec — no native "this DID is authorized to publish status on behalf of that issuer DID" assertion. So a different DID is mechanically valid (the token verifies against whatever DID its kid resolves to), but the authorization binding — proving that DID is allowed to speak for the credential issuer — isn't something the spec gives you. That has to come from the profile or ecosystem: a trust list, a controller relationship, alsoKnownAs/linked-VM constructs, or a DID that itself carries multiple keys. [IETF](https://datatracker.ietf.org/doc/draft-ietf-oauth-status-list/10/)


So I think we can leave it as it is now.

packages/core/src/modules/sd-jwt-vc/decodeSdJwtVc.ts
TimoGlastra: But `iss` is not required when using X509 certificates, only with DID. So are the SD-JWT VCs without `iss` and without `x5c`?

But this is fine, we can make it required later on 👍

packages/core/src/agent/TrustedIssuersForVerification.ts
TimoGlastra: Yeah it's more forward thinking, like with the introduction of LoTE we now need separate ones because they modeled it like this. In the EduWallets pilot the EduID issuer signed with a different DID, in the end that was reverted because there was no way to know which did to trust for status, but the theoretical concept is definitely feasible and desired in some cases. Maybe we can set it up in a way that it's a non-breaking extension in the future if we want to add status?

packages/core/src/modules/vc/linked-data-proofs/W3cJsonLdCredentialService.ts
TimoGlastra: To be honest, I'm not 100% sure during verification, we only allow signing with DID, but i think the resolver does support HTTPS controller document resolving, but I'm really not sure if that will work with our logic in Credo. (so I'm fine with the DID restriction as that is how it was implemented)

packages/core/src/agent/TrustedIssuerContext.ts
```diff
@@ -0,0 +1,96 @@
...
+      const x509Issuers = trustedIssuers.filter((t) => t.method === 'x509')
+      await X509Service.validateCertificateChain(agentContext, {
+        // TODO: we should allow x509 cert instance here
+        certificateChain: signer.certificateChain.map((certificate) => certificate.toString('base64')),
+        trustedCertificates: x509Issuers.flatMap((t) => t.issuance),
+      })
+    } else if (signer.method === 'did') {
+      // For DIDs we do allow verification to continue if there's no trusted issuers returned. To mimic original behavior.
+      // This can easily be disabled by returning an empty array
+      if (!trustedIssuers) return
```
hacdias: I don't like this inconsistency with the X509 behavior. However, I'm not 100% sure how we can retain the previous behavior without this.

TimoGlastra: Hmm yeah the thing with DIDs is that the `iss` value is clearly indicating who the issuer is, so it's less important than with X509 to have a trusted issuer (where we need the root for verification).


But i also agree the inconsistency is not great. Maybe we can change it in 0.8

packages/core/src/agent/TrustedIssuersForVerification.ts
hacdias: We could call the field `issuance: string` to be more in line with the x509. In the future, we can always add a `status` and even make it an array if necessary.

packages/core/src/modules/sd-jwt-vc/decodeSdJwtVc.ts
hacdias: I think it's just weird fixtures. I can also fix them and make this mandatory.

packages/core/src/agent/TrustedIssuersForVerification.ts
hacdias: I updated the name

packages/core/src/agent/TrustedIssuerContext.ts
TimoGlastra: I think this is fine 👍

packages/core/src/modules/mdoc/__tests__/mdocDeviceResponse.test.ts
```diff
@@ -166,4 +166,73 @@ QucCIHCvouHEm/unjBXMCeUZ7QR/ympjGyHITw25/B9H9QsC
...
+    })
+
+    test('returning an empty trusted issuers array hard-rejects the mdoc', async () => {
+      const certificate = await createIssuerCertificate()
+      const mdoc = await signMdoc(certificate)
+
+      agent.context.config.setTrustedIssuersForVerification(async () => ({ trustedIssuers: [] }))
+
+      const result = await mdoc.verify(agent.context)
+      expect(result.isValid).toBe(false)
```
TimoGlastra: can we check the error message, just so we know it's failing on the actual error.

packages/core/src/modules/sd-jwt-vc/SdJwtVcOptions.ts
```diff
@@ -158,23 +159,38 @@ export type SdJwtVcVerifyOptions = {
 
   /**
    * Whether to verify the status of the credential. If set to false and the credential
-   * has a status, it will not be fetched and verified.
+   * has a status, it will not be checked. Due to a limitation of the library we use,
+   * the status list will still be fetched and checked, but the result will not influence
+   * the verification result.
```
TimoGlastra: Hmm we should add an option to the sd-jwt library ideally. We use this option mainly so we don't have to fetch 10+ status lists when a batch is issued. But this kinda negates that.

packages/core/src/modules/sd-jwt-vc/SdJwtVcService.ts
```diff
@@ -675,6 +659,87 @@ export class SdJwtVcService {
...
+      try {
+        header = JsonEncoder.fromBase64Url(data.split('.')[0])
+      } catch {
+        return false
+      }
+
+      if (matchedTrustedIssuer.method === 'x509') {
+        // Mirror mdoc: fall back to the issuance certificates when no dedicated status certificates are
+        // configured. An explicit empty array means "trust no status list signer".
+        const statusCertificates = matchedTrustedIssuer.status ?? matchedTrustedIssuer.issuance
```
TimoGlastra: Hmm not sure if we should fall back to `issuance` certificates if no `status` certificates were found? There's `issuance` and `status` certificates for a reason, and it would allow some weird setups. 

I think we had:
- `issuance` but no `status` -> MUST be same chain/key. 
- `issuance` and `status` -> MUST be a `status` certificate

But maybe my understanding is wrong?

packages/core/src/modules/sd-jwt-vc/SdJwtVcService.ts
```diff
@@ -675,6 +659,87 @@ export class SdJwtVcService {
...
+        if (!header.x5c?.length || statusCertificates.length === 0) return false
```
TimoGlastra: If we throw an error, i think the verification result will include it and allows to give more context (As it's not signature invalid, but it's not able to extract the signer)

TimoGlastra: Since there's also no logging, it will be really hard to debug this otherwise

packages/core/src/modules/vc/linked-data-proofs/W3cJsonLdCredentialService.ts
TimoGlastra: Since this is called early on in the verification, we may want to add a check that it is indeed a did. so you can't get weird situations where` didUrl` is actually an https url.

Also the `didUrl` should point to the actual key within the did document. `issuerId` is just `did:example:123`, but we want `did:example:123#456` here. The `verificationMethod` in the `proof` should give the exact signer did url (but we would need to check also if they `issuerId` and `verificationMethod` point to the same did.

packages/core/src/modules/vc/sd-jwt-vc/W3cV2SdJwtCredentialService.ts
```diff
@@ -151,6 +154,14 @@ export class W3cV2SdJwtCredentialService {
       const issuerVerificationMethod = await getVerificationMethodForJwt(agentContext, credential, ['assertionMethod'])
       const issuerPublicKey = getPublicJwkFromVerificationMethod(issuerVerificationMethod)
 
+      // Ensure the issuer is trusted according to the (optional) `getTrustedIssuersForVerification`
+      // callback. For did-based issuers this is a no-op when no trusted issuers are configured,
+      // preserving the previous "trust any valid signature" behavior. Throws when the issuer is not trusted.
+      await TrustedIssuerContext.ensureTrustedSigner(agentContext, {
+        signer: { method: 'did', didUrl: issuerVerificationMethod.id },
```
TimoGlastra: i think here we extract it correctly, and first do checks

packages/openid4vc/src/shared/callbacks.ts
```diff
@@ -76,36 +93,68 @@ export function getOid4vcJwtVerifyCallback(
...
-        },
-      })
+      const jwtPayload = JwtPayload.fromJson(payload)
+
+      if (certificateChain) {
+        // Only fetch the issuance session record when the generic callback is configured.
+        if (agentContext.config.getTrustedIssuersForVerification) {
+          const issuanceSessionRecord = await agentContext.dependencyManager
+            .resolve(OpenId4VcIssuanceSessionRepository)
+            .getById(agentContext, options.issuanceSessionId)
```
TimoGlastra: I think it would be better/more efficient if we pass the session to the callback at the time the callbacks are instantiated. That way we prevent the re-fetch. We already pass the `issuanceSessionId`, so should be quite easy to pass the `issuanceSession` instead

packages/openid4vc/src/shared/OpenId4VcTrustedIssuersVerificationTypes.ts
```diff
@@ -0,0 +1,66 @@
...
+    payload: JwtPayload
+  }
+}
+
+export type OpenId4VcVerificationTypeOpenId4VciKeyAttestation = {
+  type: 'openId4VciKeyAttestation'
+  openId4VcIssuanceSessionRecord: OpenId4VcIssuanceSessionRecord
+  keyAttestation: {
+    jwt: string
+    payload: JwtPayload
```
TimoGlastra, line 26: Don't we have the parsed JWT payloads from the oid4vc-ts library here?

packages/openid4vc/src/shared/OpenId4VcTrustedIssuersVerificationTypes.ts
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * Credential verification bound to an OpenID4VP verification session.
+ * Use this instead of the core `VerificationTypeCredential` when the credential
+ * is verified as part of an OpenID4VP presentation flow.
+ */
+export type OpenId4VcVerificationTypeOid4VpCredential = {
+  type: 'openId4VpCredential'
+  openId4VcVerificationSessionRecord: OpenId4VcVerificationSessionRecord
+  credential: SdJwtVc | Mdoc
```
TimoGlastra: What about the other formats supported by OID4VP? With the did-based ones

packages/core/src/modules/sd-jwt-vc/SdJwtVcOptions.ts
hacdias: Yeah, we should. I can take a look.

It's still better than before this PR: the option exists now but it's a no-op 🤦

hacdias: See. https://github.com/openwallet-foundation/sd-jwt-js/pull/381

packages/openid4vc/src/shared/OpenId4VcTrustedIssuersVerificationTypes.ts
hacdias: Yes. I should add them, we just need to make sure we don't double call the verification logic

hacdias: I've added this in the latest commit.

packages/openid4vc/src/shared/callbacks.ts
hacdias: Updated!

packages/core/src/modules/sd-jwt-vc/SdJwtVcService.ts
hacdias: Updated to throw errors.

packages/core/src/modules/sd-jwt-vc/SdJwtVcService.ts
hacdias: Updated to match that behavior. It sounds more sensical.

packages/openid4vc/src/shared/OpenId4VcTrustedIssuersVerificationTypes.ts
hacdias: They are parsed, but we don't have the types. This callback receives a generic payload.

If we want to use the types, we either need to actually export them from the library (which we don't) and then either force cast, or make the input some union where we can actually get narrowed types from TypeScript.

packages/core/src/modules/sd-jwt-vc/SdJwtVcOptions.ts
hacdias: Lib updated with new option!

packages/openid4vc/src/shared/OpenId4VcTrustedIssuersVerificationTypes.ts
TimoGlastra: Okay, let's leaf it for now

hacdias (approved):

PR #2845, opened 2026-06-24
fix(didcomm): support DID rotation on v2 connections

tarunvadde:
`connections.rotate()` was v1 only. On a v2 connection it returned a new DID but never updated `connection.did` or sent a `from_prior`, so it did nothing.

This adds a v2 branch that mints a new peer DID, writes the `from_prior`, updates `connection.did`, and sends an empty message so the peer gets the new DID right away. `toDid` is not supported on v2. The receiver side needs no change.


packages/didcomm/src/modules/connections/DidCommConnectionsApi.ts
```diff
@@ -331,13 +333,54 @@ export class DidCommConnectionsApi {
...
+  public async rotate(options: {
+    connectionId: string
+    toDid?: string
+    routing?: DidCommRouting
+  }): Promise<{ newDid: string }> {
     const { connectionId, toDid } = options
     const connection = await this.connectionService.getById(this.agentContext, connectionId)
 
+    if (connection.didcommVersion === 'v2') {
+      if (toDid) {
```
genaris: I don't get why we are limiting this. In V1, this method calls `DidCommDidRotateService.createRotate()`, which accepts both `toDid` and `routing`. Can't we follow the same pattern here?

packages/didcomm/src/modules/connections/DidCommConnectionsApi.ts
```diff
@@ -331,13 +333,54 @@ export class DidCommConnectionsApi {
...
+      const pendingRotation = connection.metadata.get(DidCommConnectionMetadataKeys.DidRotateV2)
+      if (pendingRotation) {
+        return { newDid: pendingRotation.newDid }
+      }
+
+      const didRotateV2Service = this.agentContext.dependencyManager.resolve(DidCommDidRotateV2Service)
+      const routing =
+        options.routing ??
+        (await this.routingService.getRouting(this.agentContext, { mediatorId: connection.mediatorId }))
+
```
genaris: This can also be handled directly by `DidCommV2DidRotateService.rotateOurDid()`

And, to be consistent with V1, throw in case there is an pending DID Rotation flow instead of just sending again the newDid.

genaris (changes requested):
Behaviour looks OK to me but I think we can make it more consistent with the patterns followed in DIDComm V1 rotation (unless there is a limitation I'm not considering).

packages/didcomm/src/modules/connections/DidCommConnectionsApi.ts
tarunvadde: Done. `rotateOurDid` now takes `{ toDid, routing }` and follows `createRotate`: if `toDid` is set it uses that wallet-owned DID (and picks up its mediator), otherwise it mints a peer DID from routing.

packages/didcomm/src/modules/connections/DidCommConnectionsApi.ts
tarunvadde: Done. Moved the check into rotateOurDid, and it now throws the same as v1.

tarunvadde:
> Behaviour looks OK to me but I think we can make it more consistent with the patterns followed in DIDComm V1 rotation (unless there is a limitation I'm not considering).

One difference from v1: v2 has no rotation protocol message, so after rotating I send a plain empty message to notify the peer. The rotation is already saved and `from_prior` rides the next message anyway, so that send is best-effort and does not fail `rotate()`. Can make it throw like v1 if you prefer. Please re-review.

genaris (approved):

PR #2846, opened 2026-06-24
feat(didcomm): Support RFC 881 - vc+sd-jwt over issue-credential-v2

gmulhearn:
https://github.com/decentralized-identity/aries-rfcs/pull/881

This PR adds a format service handler for RFC 881 (vc+sd-jwt over issue-credential-v2) - the code is largely derived from the DI format service


gmulhearn:
@openwallet-foundation/credo-maintainers ready to review again

gmulhearn:
hey @genaris or @TimoGlastra , i fear this one might've flown under the review radar because i opened it awhile ago (whoops). Could i please request your review when possible? 🙏

genaris:
> hey @genaris or @TimoGlastra , i fear this one might've flown under the review radar because i opened it awhile ago (whoops). Could i please request your review when possible? 🙏

Sure! We've mentioned it in today's Credo Contributor call.

.changeset/bumpy-doors-knock.md
```diff
@@ -0,0 +1,5 @@
+---
+"@credo-ts/didcomm": minor
```
genaris:
```suggestion
"@credo-ts/didcomm": patch
```

Since it does not introduce a breaking change.

gmulhearn:
@genaris fixed that test failure after the https://github.com/openwallet-foundation/credo-ts/pull/2827/changes merge, working again

packages/anoncreds/src/formats/DataIntegrityDidCommCredentialFormatService.ts
genaris, line 1: Now that this has been moved to didcomm package, the cross-package import disappears, so the helpers no longer need to be part of the public `@credo-ts/didcomm` surface. They should not be re-exported from `formats/index.ts` unless third-party format services are an explicit goal.

packages/didcomm/tests/w3cV2SdJwt-flow.test.ts
genaris, line 1: `packages/didcomm/tests/` does not exist on main. I think a didcomm-only format belongs in `formats/w3cV2SdJwt/__tests__/`, like `formats/jsonld/__tests__/JsonLdCredentialFormatService.test.ts`. The test also imports the service by deep path instead of from `../src`.

packages/didcomm/src/modules/credentials/formats/w3cV2SdJwt/DidCommW3cV2SdJwtCredentialFormatService.ts
```diff
@@ -0,0 +1,530 @@
...
+  }
+
+  public async processRequest(
+    _agentContext: AgentContext,
+    _options: DidCommCredentialFormatProcessOptions
+  ): Promise<void> {
+    // Validation is done in acceptRequest on the issuer side
+  }
+
+  public async acceptRequest(
```
genaris, line 243: `acceptRequest` re-implements logic the DI service in the same package already has, in a form that is partly out of date:

- **Verification method lookup** (lines 291-297) uses the hand-rolled `assertionMethod ?? authentication ?? verificationMethod` pattern that #2934 replaced with `DidDocument.findVerificationMethodsByPurpose`, and uses a caller-supplied method without checking that it belongs to the issuer and carries an allowed purpose, which DI's `getVerificationMethod` does through `dereferenceKey(id, purposes)`.
- **Algorithm selection** (lines 299-304) takes the key's first algorithm without intersecting with what the KMS can sign, and does not validate a caller-supplied `alg` against the key, even though `getSupportedJwaSignatureAlgorithms` sits in the same class.
- **`assertAndSetCredentialSubjectId`** (line 457) and **`getFormatData`** (line 510) are copies of DI members; DI's subject-id helper is already generic over both credential classes.

We can probably extract these from the DI service into `credentials/util/` as shared functions, with the purpose order for the verification method lookup as a parameter so this service can prefer `assertionMethod` while DI keeps its current order.

packages/didcomm/src/modules/credentials/formats/w3cV2SdJwt/w3cV2SdJwtExchange.ts
genaris, line 1: Some naming/parameters that we could update, just for consistency with others:

- `W3cV2SdJwtBindingMethod` is the container of binding methods. DI calls its container `DataIntegrityBindingMethods` (plural)
- `W3cV2SdJwtCredentialIssue` for the credential attachment. DI uses `DataIntegrityCredential`. `W3cV2SdJwtCredential` would match.
- Accept-request option `verificationMethod`: DI's sibling option is `issuerVerificationMethod` (JSON-LD uses `verificationMethod`, so there is mixed precedent, but DI is the format this one is derived from).
- Accept-request option `holderBinding`: the core signing option it feeds is called `holder` (`W3cV2SdJwtSignCredentialOptions.holder`). `holder` avoids a rename at the call site.
- DI's accept-request format also exposes `credentialSubjectId`. This PR derives the subject id from the binding proof instead, which is arguably better, but an issuer of a bearer credential has no way to set it. Consider offering the same option.
- `W3cV2SdJwtCredentialOffer.credential` is typed `JsonObject` and the `W3cV2Credential` instance is converted in the service. DI's offer class accepts the instance and converts in its constructor. Same file was the template, so keep the same shape.
- The DI exchange classes carry per-field RFC comments. The new ones have none.

packages/didcomm/src/modules/credentials/formats/w3cV2SdJwt/DidCommW3cV2SdJwtCredentialFormatService.ts
```diff
@@ -0,0 +1,530 @@
...
+  }
+
+  public async acceptProposal(
+    _agentContext: AgentContext,
+    _options: DidCommCredentialFormatAcceptProposalOptions<DidCommW3cV2SdJwtCredentialFormat>
+  ): Promise<DidCommCredentialFormatCreateOfferReturn> {
+    throw new CredoError('Proposal is not supported for W3C VCDM 2.0 SD-JWT credential format')
+  }
+
+  public async createOffer(
```
genaris, line 91: `W3cV2SdJwtCredentialService.validateDisclosureFrame` rejects frames that disclose `NON_DISCLOSEABLE_FIELDS`. An offer whose `selectively_disclosable_claims` names one of them passes `createOffer` and only fails inside core during `acceptRequest`. Would it be more convenient to share the validation logic here, exporting the constant from core?

packages/didcomm/src/modules/credentials/formats/w3cV2SdJwt/DidCommW3cV2SdJwtCredentialFormatService.ts
```diff
@@ -0,0 +1,530 @@
...
+    const credentialJson = credentialOffer.credential
+    const credentialToValidate = {
+      ...credentialJson,
+      issuer: credentialJson.issuer ?? 'https://placeholder.com',
+      validFrom: credentialJson.validFrom ?? new Date().toISOString(),
+    }
+    JsonTransformer.fromJSON(credentialToValidate, W3cV2Credential)
+
+    if (credentialOffer.bindingRequired && !credentialOffer.bindingMethod?.didcommSignedAttachment) {
+      throw new CredoError('Invalid credential offer. Missing binding method when binding_required is true.')
```
genaris: This throws `CredoError` where DI throws `DidCommProblemReportError` with `IssuanceAbandoned`.

packages/didcomm/src/modules/credentials/formats/shared/didCommSignedAttachment.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  let resolvedKid: string | undefined
+
+  const jwsService = agentContext.dependencyManager.resolve(JwsService)
+  const { isValid } = await jwsService.verifyJws(agentContext, {
+    jws: {
+      header: jws.header,
+      protected: jws.protected,
+      signature: jws.signature,
+      payload: TypedArrayEncoder.toBase64Url(signedAttachment.getDataAsUint8Array()),
```
genaris, line 103: Note that DI code signed payload with `TypedArrayEncoder.fromBase64(signedAttachment.data.base64)`. The new shared helper uses `signedAttachment.getDataAsUint8Array()`, which tries base64url first and falls back to base64, and `getDataAsJson()` for the payload. That widens what DI accepts. It is likely an improvement, but it would be nice to mention in the changeset.

packages/didcomm/src/modules/credentials/formats/shared/didCommSignedAttachment.ts
```diff
@@ -0,0 +1,135 @@
...
+): Promise<{ nonce: string; kid: string }> {
+  const jws = signedAttachment.data.jws as JwsDetachedFormat
+  if (!jws) throw new CredoError('Missing jws in signed attachment')
+  if (!jws.protected) throw new CredoError('Missing protected header in signed attachment')
+  if (!signedAttachment.data.base64) throw new CredoError('Missing payload in signed attachment')
+
+  let resolvedKid: string | undefined
+
+  const jwsService = agentContext.dependencyManager.resolve(JwsService)
+  const { isValid } = await jwsService.verifyJws(agentContext, {
```
genaris: `JwsService.verifyJws` already returns `jwsSigners`, and for a `method: 'did'` signer each entry carries `didUrl`, which is the `kid` that was resolved. Could that work in this case, instead of the assignment through a callback?

packages/didcomm/src/modules/credentials/formats/w3cV2SdJwt/DidCommW3cV2SdJwtCredentialFormatService.ts
```diff
@@ -0,0 +1,530 @@
...
+      record: W3cV2CredentialRecord.fromCredential(verifiableCredential),
+    })
+
+    credentialExchangeRecord.credentials.push({
+      credentialRecordType: this.credentialRecordType,
+      credentialRecordId: record.id,
+    })
+  }
+
+  private verifyCredentialMatchesOffer(
```
genaris, line 384: Would it be possible here to do some more advanced checks, like we do in DI?

Also we can check `selectively_disclosable_claims` (absents from `sdJwt.payload` but present in `sdJwt.prettyClaims`), `claimPathsFromDisclosureFrame` helper has some useful logic for that.

packages/didcomm/src/modules/credentials/formats/w3cV2SdJwt/DidCommW3cV2SdJwtCredentialFormatService.ts
```diff
@@ -0,0 +1,530 @@
...
+    const w3cV2SdJwtFormat = credentialFormats.w3cV2SdJwt
+    if (!w3cV2SdJwtFormat) throw new CredoError('Missing w3cV2SdJwt credential format data')
+
+    const { credential, bindingRequired, disclosureFrame, didCommSignedAttachmentBinding } = w3cV2SdJwtFormat
+
+    const credentialJson = credential instanceof W3cV2Credential ? JsonTransformer.toJSON(credential) : credential
+
+    if ('proof' in credentialJson) throw new CredoError('The offered credential MUST NOT contain any proofs.')
+
+    // Validate credential as VCDM 2.0 (with offer exceptions: issuer/validFrom can be missing)
```
genaris: `acceptRequest` (line 275) parses the same offered credential with `JsonTransformer.fromJSON(..., W3cV2Credential)`, which fails validation without an issuer, and then reads `credential.issuer` to resolve the signing key. `DidCommW3cV2SdJwtAcceptRequestFormat` has no `issuer` option, so there is no way to supply it. The offer side and the issue side disagree about what a valid offer is.

packages/didcomm/src/modules/credentials/formats/w3cV2SdJwt/DidCommW3cV2SdJwtCredentialFormatService.ts
```diff
@@ -0,0 +1,530 @@
...
+    const credentialToValidate = {
+      ...credentialJson,
+      issuer: credentialJson.issuer ?? 'https://placeholder.com',
+      validFrom: credentialJson.validFrom ?? new Date().toISOString(),
+    }
+    JsonTransformer.fromJSON(credentialToValidate, W3cV2Credential)
+
+    // Build binding method if required
+    let bindingMethod: W3cV2SdJwtBindingMethod | undefined
+    if (bindingRequired) {
```
genaris: Here we only build the method when `bindingRequired` is true and silently drop the binding options otherwise, and `acceptOffer` (line 173) only creates a proof when `bindingRequired` is true and silently ignores holder options otherwise. `acceptRequest` already handles "proof present although not required", so only the two earlier steps are in the way.

Could we mirror here what we do in DI?

- In `createOffer`, build the binding method whenever `didCommSignedAttachmentBinding` is given, and keep the existing check that throws when `bindingRequired` is true and no binding options were given.
- In `acceptOffer`, create the signed attachment whenever the holder passes `didCommSignedAttachment` options, throw "binding method not offered" when the offer has no `didcomm_signed_attachment`, and throw "missing required binding proof" only when `bindingRequired` is true and no proof was produced.
- `shouldAutoRespondToOffer` can then stay as it is (`!bindingRequired`), since an optional binding still needs a holder decision.

genaris (changes requested):
Thanks @gmulhearn and sorry for the time it took me to review it. I have some comments, nothing so blocking, but we'll need to solve the merge conflicts now that DataIntegrityDidCommCredentialFormatService resides directly on DIDComm package.

packages/didcomm/tests/w3cV2SdJwt-flow.test.ts
gmulhearn: have updated the `../src` import, but some considerations;

The goal of this file was to be a more system-level test, rather than a unit test like `JsonLdCredentialFormatService.test.ts`, `<package-root>/tests/` seemed to be the convention for that, rather than nested in `__tests__`.

note as well that the file naming, positioning and structure was also based on test files like `packages/anoncreds/tests/data-integrity-flow.test.ts`. I see as well that there is a new `packages/anoncreds/tests/data-integrity-flow-vcdm2.test.ts` test file too, which doesn't have any anoncreds dependency - i wonder if that would be better suited for `packages/didcomm/tests/*` too?

genaris: Yes, honestly I think we need to refactor and clean up a lot of test suites that have some outdated/duplicated tests. I'm OK on the file location. In a further PR we can try to harmonize the whole tests to be consistent and avoid redundant checks.

gmulhearn: ok got it, i'll leave as is then? ..src imports updated here: https://github.com/openwallet-foundation/credo-ts/pull/2846/changes/f4ace549fe016d14e45b3f147ec112ddc1a3668d

packages/didcomm/src/modules/credentials/formats/w3cV2SdJwt/DidCommW3cV2SdJwtCredentialFormatService.ts
gmulhearn: extracted those key parts and have DI & vc+sd-jwt service's utilizing it: https://github.com/openwallet-foundation/credo-ts/pull/2846/changes/0bfc55b58b444750c6ba3a84a1561688666eb997

packages/didcomm/src/modules/credentials/formats/w3cV2SdJwt/w3cV2SdJwtExchange.ts
gmulhearn: applies those in this commit; https://github.com/openwallet-foundation/credo-ts/pull/2846/changes/6871ea2af8d74d44ccd8e6e398306a9f342f00c2

packages/didcomm/src/modules/credentials/formats/w3cV2SdJwt/DidCommW3cV2SdJwtCredentialFormatService.ts
gmulhearn: helper function extracted & exported https://github.com/openwallet-foundation/credo-ts/pull/2846/changes/2ebfa555ada79746ce71f5575ea4b4b54ea58971

packages/didcomm/src/modules/credentials/formats/w3cV2SdJwt/DidCommW3cV2SdJwtCredentialFormatService.ts
gmulhearn: https://github.com/openwallet-foundation/credo-ts/pull/2846/changes/70637977dc5f29d6df26f34ba39b280ccf69269f

packages/didcomm/src/modules/credentials/formats/shared/didCommSignedAttachment.ts
gmulhearn: mentioned in changeset https://github.com/openwallet-foundation/credo-ts/pull/2846/changes/5028163210ceaf1d73dcdf73fb7a9b378c30b603

packages/didcomm/src/modules/credentials/formats/w3cV2SdJwt/DidCommW3cV2SdJwtCredentialFormatService.ts
gmulhearn: had a go at this here, + extracted some DI common logic; https://github.com/openwallet-foundation/credo-ts/pull/2846/changes/0f683586f11f8d37ca751ad6eb54ac6760b20a1e

packages/didcomm/src/modules/credentials/formats/w3cV2SdJwt/DidCommW3cV2SdJwtCredentialFormatService.ts
gmulhearn: doh.. i missed this path. added `issuer` option + checks https://github.com/openwallet-foundation/credo-ts/pull/2846/changes/6931f10a97c6e197bbe31d14d99332776a48d426

packages/didcomm/src/modules/credentials/formats/w3cV2SdJwt/DidCommW3cV2SdJwtCredentialFormatService.ts
gmulhearn: https://github.com/openwallet-foundation/credo-ts/pull/2846/changes/a3f9e1c0c3d79ccd7fe6024222a9b788956754b4

packages/didcomm/src/modules/credentials/formats/shared/didCommSignedAttachment.ts
gmulhearn: ah yep, that's less dodgy https://github.com/openwallet-foundation/credo-ts/commit/5028163210ceaf1d73dcdf73fb7a9b378c30b603

packages/didcomm/src/modules/credentials/formats/w3cV2SdJwt/disclosureFrame.ts
```diff
@@ -0,0 +1,104 @@
...
+      }
+      current = current[segment] as IDisclosureFrame
+    }
+
+    // Add the last segment to _sd. Numeric indices must be numbers for array item disclosure.
+    if (!current._sd) current._sd = []
+    const claim = segments[segments.length - 1]
+    const sdValue = /^\d+$/.test(claim) ? Number(claim) : claim
+    if (!current._sd.includes(sdValue as string)) {
+      current._sd.push(sdValue as string)
```
genaris: The cast is there because `IDisclosureFrame._sd` is typed `string[]` in core `packages/core/src/modules/sd-jwt-vc/SdJwtVcOptions.ts`, while the library accepts numbers for array elements, as this code relies on. Since the PR already touches core's SD-JWT module, widening it to `(string | number)[]` would make the cast unnecessary.

packages/didcomm/src/modules/credentials/formats/shared/didCommSignedAttachment.ts
genaris: Now that the other shared helpers moved to `credentials/util/`, this is the only file left under `formats/shared/`. Moving it to `credentials/util/didCommSignedAttachment.ts` would keep one home for cross-format helpers and let us drop the shared folder.

packages/didcomm/src/modules/credentials/formats/w3cV2SdJwt/DidCommW3cV2SdJwtCredentialFormatService.ts
```diff
@@ -0,0 +1,484 @@
...
+    }
+
+    // Claims the issuer said would be selectively disclosable are absent from the signed payload, where
+    // they are replaced by an `_sd` digest, but present once the disclosures are applied.
+    for (const claimPath of offer.selectivelyDisclosableClaims ?? []) {
+      if (!resolveClaimPath(credential.sdJwt.prettyClaims, claimPath).found) {
+        throw new CredoError(`Received credential is missing claim '${claimPath}' offered as selectively disclosable.`)
+      }
+
+      if (resolveClaimPath(credential.sdJwt.payload, claimPath).found) {
```
genaris: I might be wrong here, but does this check handle array elements? If I read `@sd-jwt/core` correctly, an object property that is selectively disclosable is removed from the payload, but an array element is replaced in place by a `{ "...": "<digest>" }` placeholder at the same index. If that's the case, a path like `$.credentialSubject.result[0]` would still be found in `sdJwt.payload` and the credential would be rejected as "not disclosable" even though it is. Since `claimPathsFromDisclosureFrame` produces array-index paths, that would affect any offer disclosing array elements.

Could you double check? A flow test with an array element in the disclosure frame (e.g.`{ credentialSubject: { result: { _sd: [0] } } }`) going through `acceptRequest` and `processCredential` would settle it either way. If it turns out to be the case, treating a resolved value that is an object with only a `...`key as disclosable should be enough.

genaris (commented):
Thanks @gmulhearn ! It's looking pretty good. I have a concern regarding selectively-disclosable check rejects array in `DidCommW3cV2SdJwtCredentialFormatService`, but otherwise I'm good to merge!

PR #2855, opened 2026-06-30
fix(sd-jwt): iat is not a mandatory field

berendsliedrecht:


packages/core/src/modules/sd-jwt-vc/SdJwtVcService.ts
```diff
@@ -144,14 +147,15 @@ export class SdJwtVcService {
...
       {
-        ...payload,
+        ...payloadWithoutIat,
         cnf: holderBinding?.cnf,
         iss: issuer.iss,
-        iat: nowInSeconds(),
+        iat: payloadIat,
         vct: payload.vct,
-      },
+      } as Payload,
```
TimoGlastra: Why not just like this?

```suggestion
const compact = await sdJwt.issue(
{
        ...payload
cnf: holderBinding?.cnf,
iss: issuer.iss,
vct: payload.vct,
      } as Payload,
```

TimoGlastra (commented):
Let's not do this, as it's a breaking change. It will cause unwanted side effects.

What if you can explicitly disable it for now by providing `null`?

packages/core/src/modules/sd-jwt-vc/SdJwtVcService.ts
berendsliedrecht: I wanted behaviour before differently, but changed it now!

berendsliedrecht:
> Let's not do this, as it's a breaking change. It will cause unwanted side effects.
> 
> What if you can explicitly disable it for now by providing `null`?

Would still require a change, but I can change it like that yes.

TimoGlastra (approved):

PR #2856, opened 2026-07-02
feat(openid4vc): support for wallet attestations draft 9

hacdias:


packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
```diff
@@ -1539,6 +1548,37 @@ export class OpenId4VciHolderService {
...
+   * Fetches a Client Attestation PoP challenge from the authorization server's `challenge_endpoint`
+   * (draft 09 of OAuth 2.0 Attestation-Based Client Authentication).
+   */
+  private async fetchClientAttestationChallenge(
+    agentContext: AgentContext,
+    challengeEndpoint: string
+  ): Promise<string> {
+    const { fetch } = agentContext.config.agentDependencies
+
+    const response = await fetch(challengeEndpoint, {
```
TimoGlastra: Is there no method in the oid4vc library to fetch the nonce? We have standard way to fetch with zod validation in that library

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerModuleConfig.ts
```diff
@@ -185,6 +198,11 @@ export interface InternalOpenId4VcIssuerModuleConfigOptions {
      */
     authorizationChallenge?: string
 
+    /**
+     * @default /client-attestation-challenge
+     */
+    clientAttestationChallenge?: string
```
TimoGlastra, line 204: Is this endpoint intended to be reused for other challenges? So you request /challenge, and it now returns "client_challenge" but maybe others in the future?

Or is the sole purpose attestation challenge? In which case the default endpoint makes sense

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerService.ts
```diff
@@ -1383,6 +1383,11 @@ export class OpenId4VcIssuerService {
       authorization_challenge_endpoint: joinUriParts(issuerUrl, [config.authorizationChallengeEndpointPath]),
       authorization_endpoint: joinUriParts(issuerUrl, [config.authorizationEndpoint]),
 
+      // Client Attestation PoP challenge (draft 09). The challenge endpoint is always exposed; the
+      // nonce is only required when explicitly enabled through config.
+      challenge_endpoint: joinUriParts(issuerUrl, [config.clientAttestationChallengeEndpointPath]),
+      client_attestation_pop_nonce_required: config.clientAttestationPopNonceRequired || undefined,
```
TimoGlastra: On purpose that it is never false?

TimoGlastra: And weird that the spec says:

> it MUST signal support for the challenge endpoint by including the metadata entry challenge_endpoint containing the URL of the endpoint as its value. If the Authorization Server offers a challenge endpoint, the Client MUST retrieve a challenge and MUST use this challenge in the Client Attestation PoP JWT or DPoP Proof as defined in Section 5.

Why then have the separate metadata param?

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerModuleConfig.ts
TimoGlastra: Ah i see in the spec:

> attestation_challenge: REQUIRED if the Authorization Server or Resource Server supports Client Attestations and server-provided challenges as described in this document. String containing a Challenge to be used in the Client Attestation PoP JWT or DPoP Proof as defined in Section 5. The intention of this element not being required in other circumstances is to preserve the ability for the challenge endpoint to be used in other applications unrelated to client attestations.


So i think we should call it /challenge

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerService.ts
TimoGlastra: Also where did you find this nonce required param? I don't see it in the spec.


And do we need to add client_attestation_signing_alg_values_supported and others?

TimoGlastra (approved):
Have you considered the combined dpop/attestation flow? Should we leave that for a follow up?

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerModuleConfig.ts
hacdias: Hmm. `authorizationChallengeEndpointPath` has `/challenge` as default. I can check if it's possible to easily distinguish the requests and reuse the same endpoint.

hacdias: I think this would become very cumbersome since this new attestation challenge endpoint is just a plain POST request, while the authorization challenge has body and everything. We can use that to distinguish them, but I'm afraid it'll be a fragile. They're also used for different specs.

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerService.ts
hacdias: Oh, oops. I'll remove the nonce thing. That's from Draft 5. The `oid4vc-ts` libraries kept it, but it seems like we never had it in Credo.

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerModuleConfig.ts
TimoGlastra: Ah no shouldn't reuse the same endpoint. I thought the other was called authorization-challenge. Let's keep it then.

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerService.ts
```diff
@@ -1367,6 +1370,20 @@ export class OpenId4VcIssuerService {
...
+    const clientAttestationAuthMethods = [
+      // Standard method: Client Attestation JWT + standalone Client Attestation PoP JWT.
+      ...(issuerRecord.clientAttestationSigningAlgValuesSupported &&
+      issuerRecord.clientAttestationPopSigningAlgValuesSupported
+        ? [SupportedClientAuthenticationMethod.ClientAttestationJwt]
+        : []),
+      // DPoP-bound method (§5.2): Client Attestation JWT + DPoP proof.
+      ...(issuerRecord.clientAttestationSigningAlgValuesSupported && issuerRecord.dpopSigningAlgValuesSupported
+        ? [SupportedClientAuthenticationMethod.ClientAttestationJwtDpop]
+        : []),
```
hacdias, line 1382: @TimoGlastra do we want to support the combined proof _if_ DPoP and client attestations are enabled? Or do we want to add a specific flag for it? I think it's fine this way.

hacdias:
@TimoGlastra I updated some details and fixed some things. There's also support for the combined DPoP and attestation proof. Could you re-review?

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
```diff
@@ -396,20 +396,50 @@ export class OpenId4VciHolderService {
...
-        })
-      : // We should be careful about this case. It could just be the user didn't correctly
-        // provide the DPoP from the auth response. In which case different DPoP will be used
-        // However it might be that they only use DPoP for the token request (esp in pre-auth case)
-        isDpopSupported.supported
+    // draft 09 §5.2: when the authorization server supports the DPoP-bound client attestation method, the
+    // client instance key is used as the DPoP key so a single DPoP proof serves as both the DPoP proof and
+    // the client attestation pop. Only applied when there is no pre-existing DPoP binding (i.e. the
+    // pre-authorized code flow); for the authorization code flow the DPoP key is already established at the
+    // pushed authorization request.
```
TimoGlastra: Just so i understand. If I already used DPoP-bound client attestation during PAR, it will again detect that we can do DPoP-bound client attestation here, and use that same key, as we can consistently extract it from the wallet attestatino?


How does it work with the refresh token (deferred or credential refresh)? Do we expect the same key for the full interaction? I think based on our previous DPoP implementation we do, and looking into the spec it's required to use the same key for the full interaction, even with refresh: https://www.ietf.org/archive/id/draft-ietf-oauth-attestation-based-client-auth-09.html#name-refresh-token-binding

TimoGlastra: If we don't do DPoP-bound Client Attestation, do we then also require the same Client Attestation key for the full interaction (PAR, Access Token, Refresh Token)

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
```diff
@@ -1493,13 +1534,23 @@ export class OpenId4VciHolderService {
...
-            authorizationServerMetadata.challenge_endpoint
-              ? await this.fetchClientAttestationChallenge(agentContext, authorizationServerMetadata.challenge_endpoint)
-              : undefined
+          // Draft 09: when the authorization server exposes a challenge endpoint, the client MUST include
+          // a fresh challenge in the client attestation pop jwt. Fetch one proactively (covers PAR, token
+          // and authorization challenge). On the token endpoint the library additionally auto-retries the
+          // `use_attestation_challenge` error with a server-provided challenge, which takes precedence.
+          const challenge = authorizationServerMetadata.challenge_endpoint
+            ? (await requestClientAttestationChallenge({ authorizationServerMetadata, callbacks })).challenge
+            : undefined
```
TimoGlastra, line 1554: I think the challenge is allowed to be reused for a certain time. Now we will request a new challenge for both PAR and access token request.

For DPoP we had a mechanism to pass around the nonce, for credential endpoint we call the nonce endpoint once and reuse that until we get an error that we need to use a new one. 

Since we only call this endpoint twice, i think it may be fine? Not sure how easy it would be to reuse

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerService.ts
TimoGlastra: I think this is fine yes. Less to configure

packages/openid4vc/src/openid4vc-issuer/router/accessTokenEndpoint.ts
```diff
@@ -235,10 +240,40 @@ export function handleTokenRequest(config: OpenId4VcIssuerModuleConfig) {
...
       }
 
-      await openId4VcIssuerService.verifyClientAttestationPopChallenge(agentContext, issuer, {
-        clientAttestation: verificationResult.clientAttestation,
-        required: config.clientAttestationPopNonceRequired,
-      })
+      // Client Attestation PoP challenge (draft 09). At the token endpoint we use the reactive
+      // `use_attestation_challenge` flow: when a required challenge is missing/invalid we mint a fresh
+      // one, return it in the `OAuth-Client-Attestation-Challenge` header, and the client retries.
+      if (config.clientAttestationPopChallengeRequired && verificationResult.clientAttestation) {
```
TimoGlastra: What if a challenge is not required, but is provided? Shouldn't we still verify it, or handle it in some way?

TimoGlastra (approved):
just some comments, but this looks good already

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
hacdias: Uh, just updated it to make sure we ensure that the same key is used indeed (it was just a flag since the library already supported it).

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
hacdias: Yes, it's not ideal. We can improve this in a follow up to be able to reuse it.

packages/openid4vc/src/openid4vc-issuer/router/accessTokenEndpoint.ts
hacdias: Yes, you're right, we should verify it. A client shouldn't be sending it unless advertised, so it should always be verified. Just updated.

PR #2863, opened 2026-07-08
feat(openid4vc): improved key attestation level hierarchy

hacdias:
Interpret key attestation ISO 18045 levels hierarchically on the issuer, so a stronger attested level (e.g. `iso_18045_high`) satisfies a weaker required level (e.g. `iso_18045_moderate`). Also adds the `OpenId4VciKeyAttestationLevel` enum and `keyAttestationLevelSatisfies` helper.

### DIFF
--- a/packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderServiceOptions.ts
+++ b/packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderServiceOptions.ts
@@ -13,6 +13,7 @@ import type {
   OpenId4VciAccessTokenResponse,
   OpenId4VciCredentialConfigurationSupportedWithFormats,
   OpenId4VciCredentialConfigurationsSupportedWithFormats,
+  OpenId4VciKeyAttestationLevel,
   OpenId4VciMetadata,
 } from '../shared'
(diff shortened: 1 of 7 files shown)


packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderServiceOptions.ts
```diff
@@ -483,8 +489,8 @@ export type OpenId4VciProofOfPressionProofTypes = Record<
        *  by the `keyStorage` and `userAuthentication` values.
        */
       keyAttestationsRequired?: {
-        keyStorage?: string[]
-        userAuthentication?: string[]
+        keyStorage?: (OpenId4VciKeyAttestationLevel | string)[]
+        userAuthentication?: (OpenId4VciKeyAttestationLevel | string)[]
```
TimoGlastra: You can do `OpenId4VciKeyAttestationLevel | (string & {})` to get auto-complete while also allowing string values

hacdias: Done

TimoGlastra (approved):

PR #2868, opened 2026-07-14
fix(sd-jwt-vc): use JWT header alg for SD-JWT sign and verify

Sshovon:
This PR fixes SD-JWT VC signature verification and signing when the algorithm in the JWT header does not match the default on the resolved public key. Keys from DID documents or x5c certificates contain key material only, so Credo falls back to the first supported signature algorithm on the JWK, which may not be the one used to sign that JWT. Verification and signing could therefore use the wrong algorithm and fail for otherwise valid credentials from external issuers. The change adds **resolveSignatureAlgorithm** and updates **getSdJwtSigner** and **getSdJwtVerifier** to prefer the JWT header alg when the public key supports it, with call sites updated to pass the relevant header value. The update is backward compatible: when the header alg is missing or unsupported, behavior falls back to _key.signatureAlgorithm_ as before, so existing Credo issuance and verification flows are unchanged.


packages/core/src/modules/sd-jwt-vc/SdJwtVcService.ts
```diff
@@ -135,7 +135,7 @@ export class SdJwtVcService {
 
     const sdJwt = new SDJwtVcInstance({
       ...this.getBaseSdJwtConfig(agentContext),
-      signer: getSdJwtSigner(agentContext, issuer.publicJwk),
+      signer: getSdJwtSigner(agentContext, issuer.publicJwk, { jwtHeaderAlg: issuer.alg }),
```
TimoGlastra: What if we add this to the `PublicJwk` class?

```
  /**
   * Get the first signature algorithm to use with this jwk. If the jwk has an `alg` field defined
   * it will use that alg, and otherwise fall back to the first supported signature algorithm.
   *
   * If no algorithm is supported it will throw an error
   */
  public set signatureAlgorithm(alg: KnownJwaSignatureAlgorithm) {
    const supportedSignatureAlgorithms: KnownJwaSignatureAlgorithm[] = this.jwk.supportedSignatureAlgorithms

    if (!supportedSignatureAlgorithms.includes(alg)) {
      throw new KeyManagementError(`${this.jwkTypeHumanDescription} does not support alg '${alg}'.`)
    }

    this.jwk.jwk.alg = alg
  }
```

And then we can do:
`issuer.publicJwk.alg = issuer.alg`.

Then the other code later on will work automatically (it will prefer `alg` if it's defined

packages/core/src/modules/sd-jwt-vc/SdJwtVcService.ts
```diff
@@ -674,23 +680,26 @@ export class SdJwtVcService {
...
     return async (data, signatureBase64Url) => {
+      let header: { alg?: string; x5c?: string[]; kid?: string }
+      try {
+        header = JsonEncoder.fromBase64Url(data.split('.')[0])
+      } catch (error) {
+        throw new SdJwtVcError('Unable to decode the status list JWT header to extract the status list signer.', {
+          cause: error,
+        })
+      }
+
```
TimoGlastra, line 697: I think if we already set the `alg` on the JWK earlier on, we don't need to parse the JWT here i think? we can do it when parsing the jwt in the first place.

Sshovon:
@TimoGlastra, I implemented **setAlgFromJwtHeader** helper on **PublicJwk** for verify paths, alongside the **signatureAlgorithm** setter you suggested for signing. On verify, alg comes from an external JWT header as an untyped string and may be missing or unsupported for the resolved public key, I validate it, set jwk.alg when valid, and otherwise leave the key unchanged so **signatureAlgorithm** falls back to the existing default. That keeps backward compatibility instead of throwing like the setter would, which is what I want on sign paths where the algorithm is already chosen and must be valid. In short: the **setter** for signing (issuer.alg, options.alg, holder.alg); **setAlgFromJwtHeader** for verifying (credential and status list JWT headers).

One question on naming: you suggested _issuer.publicJwk.alg = issuer.alg_ but I used _publicJwk.signatureAlgorithm = issuer.alg_ instead, since **PublicJwk** already exposes **signatureAlgorithm** as a **getter** that reads from jwk.alg. Same effect, but more explicit about intent. Happy to change if you prefer **alg** setter on the public API. Let me know.

TimoGlastra:
Hey @Sshovon, i have pushed some changes to the PR. It seems this is an issue broader than just SD-JWT so i'll look into to see if we need to fix this elsewhere as well

TimoGlastra (approved):

PR #2870, opened 2026-07-20
feat(core): improved x509 crl caching mechanism

hacdias:
Improves X509 CRL caching by caching the data of the parsed CRL instead of the raw bytes. The download is cheap, but the parsing isn't. We also cache information about the signer key in order to be able to fast check/verify on subsequent calls. Also fixes a bug where the globally configured CRL caching time wasn't being taken into account.

No changelog because this feature (CRL) hasn't been released yet.

### DIFF
--- a/packages/core/src/modules/x509/X509RevocationService.ts
+++ b/packages/core/src/modules/x509/X509RevocationService.ts
@@ -1,10 +1,16 @@
-import * as x509 from '@peculiar/x509'
 import type { AgentContext } from '../../agent'
 import { CredoWebCrypto } from '../../crypto/webcrypto'
 import { injectable } from '../../plugins'
-import { validateCriticalCrlExtensions } from './utils/criticalExtensions'
-import { fetchCrl, getCachedCrl, setCachedCrl } from './utils/crlFetcher'
-import { X509Certificate } from './X509Certificate'
+import { validateCriticalCrlExtensionIds } from './utils/criticalExtensions'
+import { fetchCrl, getCachedCrlSummary, setCachedCrlSummary } from './utils/crlFetcher'
+import {
+  buildCrlSummary,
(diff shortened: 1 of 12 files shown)


packages/core/src/modules/x509/X509RevocationService.ts
```diff
@@ -548,9 +553,61 @@ export class X509RevocationService {
...
+    if (useCache) {
+      const summary = await getCachedCrlSummary(agentContext, url)
+      const verificationTime = verificationDate.getTime()
+      const issuerKeyUsage = issuerCertificate.keyUsage
+      if (
+        summary &&
+        summary.issuerNameSha256 === computeIssuerNameSha256(issuerCertificate) &&
+        summary.issuerPublicJwkThumbprint === computeIssuerPublicJwkThumbprint(issuerCertificate) &&
+        // Mirror verify()'s cRLSign check: key usage, when present and non-empty, must include cRLSign
+        !(issuerKeyUsage && issuerKeyUsage.length > 0 && !issuerKeyUsage.includes(X509KeyUsage.CrlSign)) &&
```
TimoGlastra, line 593: Shouldn't we just store the certificate on the summary and compare the certificates? Is there a specific reason for going with these checks?

TimoGlastra (approved):

packages/core/src/modules/x509/X509RevocationService.ts
hacdias: Yes, the idea is to support the case when a certificate is renewed, but the same key is used (e.g. some CAs). In that case, the certificate itself won't match but the rest of the information should match. I'm unsure how common this case is, but my search says it's a possible case.

hacdias: I'll merge it now, we can massage this after if needed.

PR #2871, opened 2026-07-21
feat: allow for per-context and global caching

hacdias:
Allows per-context and global caching.

### DIFF
--- a/packages/core/src/modules/cache/Cache.ts
+++ b/packages/core/src/modules/cache/Cache.ts
@@ -1,7 +1,27 @@
 import type { AgentContext } from '../../agent/context'
 
+export interface CacheOptions {
+  /**
+   * The scope of the cache entry:
+   *
+   * - `'context'` (default): the entry is isolated to the `contextCorrelationId` of the provided
+   *   agent context. Implementations MUST default to this scope.
+   * - `'global'`: the entry is shared across all agent contexts. Only use this for public,
+   *   non-context-specific data (e.g. CRLs, DID documents of public DID methods, ledger objects).
+   *
+   * @default 'context'
+   */
+  scope?: 'context' | 'global'
+}
+
 export interface Cache {
-  get<CacheValue>(agentContext: AgentContext, key: string): Promise<CacheValue | null>
-  set<CacheValue>(agentContext: AgentContext, key: string, value: CacheValue, expiresInSeconds?: number): Promise<void>
-  remove(agentContext: AgentContext, key: string): Promise<void>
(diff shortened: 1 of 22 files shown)


packages/core/src/modules/cache/Cache.ts
```diff
@@ -1,7 +1,27 @@
 import type { AgentContext } from '../../agent/context'
 
+export interface CacheOptions {
```
hacdias, line 3: Was a bit undecided between `CacheOptions` and `CachingOptions`. Wdyt?

packages/core/src/modules/cache/Cache.ts
```diff
@@ -1,7 +1,27 @@
...
-  get<CacheValue>(agentContext: AgentContext, key: string): Promise<CacheValue | null>
-  set<CacheValue>(agentContext: AgentContext, key: string, value: CacheValue, expiresInSeconds?: number): Promise<void>
-  remove(agentContext: AgentContext, key: string): Promise<void>
+  get<CacheValue>(agentContext: AgentContext, key: string, options?: CacheOptions): Promise<CacheValue | null>
+  set<CacheValue>(
+    agentContext: AgentContext,
+    key: string,
+    value: CacheValue,
+    expiresInSeconds?: number,
+    options?: CacheOptions
```
hacdias, line 24: I'm not very happy with this, but I kept it like this for backwards compatibility.

I'd rather here have:


```suggestion
    options?: CacheOptions & { expiresInSeconds?: number }
```

packages/core/src/modules/dids/services/DidResolverService.ts
```diff
@@ -12,6 +12,13 @@ import { parseDid } from '../domain/parse'
 import { DidRepository } from '../repository'
 import type { DidResolutionOptions, DidResolutionResult, ParsedDid } from '../types'
 
+/**
+ * Did methods whose documents live in public, context-independent sources: resolving them yields
+ * the same result in any agent context, so their documents are cached in the global scope shared
+ * across all agent contexts. Documents of other did methods are cached per agent context.
+ */
+const PUBLIC_DID_METHODS = ['web', 'indy', 'sov', 'cheqd', 'hedera', 'webvh']
+
```
TimoGlastra: I think generally we'd solve this based on a property on the resolver class, so it's more extensible. Like we have allowsCaching (not 100% sure what its called)

TimoGlastra (approved):

packages/core/src/modules/dids/services/DidResolverService.ts
hacdias: I will make this an option with this default.

PR #2878, opened 2026-07-24
feat(didDoc): purpose aware key lookup

rmlearney-digicatapult:
Enhanced `findVerificationMethodByKeyType()` in `DidDocument.ts` to accept an optional `proofPurpose`, matching existing `resolveVerificationMethodFromCreatedDidRecord(vm, allowedPurposes?)` and `dereferenceKey(keyId, purposes?)`

Optimise key lookup in `LDP` DIDComm signing path by passing `proofPurpose` to support retrieving `assertionMethod` from DID Documents with multiple keys and not just pick the first key found

### DIFF
--- a/packages/didcomm/src/modules/credentials/formats/jsonld/DidCommJsonLdCredentialFormatService.ts
+++ b/packages/didcomm/src/modules/credentials/formats/jsonld/DidCommJsonLdCredentialFormatService.ts
@@ -301,7 +301,7 @@ export class DidCommJsonLdCredentialFormatService
       throw new CredoError(`No Key Type found for proofType ${proofType}`)
     }
 
-    // Support issuer DID documents with separate keys for different purposes (authentication vs assertionMethod).
+    // Support issuer DID documents with separate keys for different purposes
     const proofPurpose = credentialRequest.options.proofPurpose ?? 'assertionMethod'
 
     try {
--- a/packages/didcomm/src/modules/credentials/formats/jsonld/DidCommJsonLdCredentialFormatService.ts
+++ b/packages/didcomm/src/modules/credentials/formats/jsonld/DidCommJsonLdCredentialFormatService.ts
@@ -1,4 +1,4 @@
-import type { AgentContext } from '@credo-ts/core'
+import type { AgentContext, DidDocument } from '@credo-ts/core'
 import {
   ClaimFormat,
   CredoError,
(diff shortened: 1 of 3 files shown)


packages/didcomm/src/modules/credentials/formats/jsonld/DidCommJsonLdCredentialFormatService.ts
```diff
@@ -300,12 +301,40 @@ export class DidCommJsonLdCredentialFormatService
...
+      const { didDocument } = await didsApi.resolveCreatedDidDocumentWithKeys(parsedDid.did)
+
+      const verificationMethod = await findVerificationMethodByKeyType(keyType[0], didDocument, [
+        proofPurpose,
+        'verificationMethod',
+      ])
+
+      if (!verificationMethod) {
+        throw new CredoError(`Missing verification method for key type ${keyType}`)
+      }
```
TimoGlastra: The did resolver service already supports caching. I think rather than having to deal with local vs remote resolving everywhere, I think we should use the existing logic for this.


Did you actuall run into issues with it not using the local records?
https://github.com/openwallet-foundation/credo-ts/blob/961546786cd5dad4c758fe64d0f24553d3082155/packages/core/src/modules/dids/services/DidResolverService.ts#L94-L95

TimoGlastra (changes requested):

packages/didcomm/src/modules/credentials/formats/jsonld/DidCommJsonLdCredentialFormatService.ts
rmlearney-digicatapult: Ah! Is that todo still active or can the cache be used now?

rmlearney-digicatapult:
@TimoGlastra thanks for the note! Now solved with minimal diffs from main

packages/didcomm/src/modules/credentials/formats/jsonld/DidCommJsonLdCredentialFormatService.ts
TimoGlastra: I think the todo hints at the case where we don't store the did document on the local record. we should in that case update the record with the resolved document, so we can use it in the future.

TimoGlastra (approved):

PR #2885, opened 2026-07-28
feat(openid4vc): support updating the signed metadata signer on issuer update

IliasElbarnoussi:
Allow configuring, replacing, or removing the metadata signer when updating an existing issuer. When the signer is omitted, the previously configured signer is preserved and the signed metadata is re-signed to stay in sync with updated issuer metadata. Passing `null` explicitly removes the signer.

- Add `OpenId4VcUpdateIssuerOptions` with optional `metadataSigner`
- Thread the new options through `updateIssuerMetadata` and `updateIssuer`
- Re-sign metadata automatically using the stored signer on updates
- Add tests for the configure → keep-in-sync → remove lifecycle
- Add test for x5c signer round-trip (encode/decode with key id)

Signed-off-by: Ilias Elbarnoussi <ilias@animo.id>

### DIFF
--- a/.changeset/lucky-moons-repeat.md
+++ b/.changeset/lucky-moons-repeat.md
@@ -1,0 +1,5 @@
+---
+"@credo-ts/openid4vc": patch
+---
+
+Support configuring the signed issuer metadata signer when updating an issuer. `updateIssuerMetadata` and `updateIssuer` now accept a `metadataSigner`, which can be omitted to keep the current signer, set to a signer to enable or replace it, or set to `null` to stop signing the metadata. Previously a signer could only be configured when creating an issuer.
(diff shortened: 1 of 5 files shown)


hacdias:
Is using `null` to remove a field more consistent with the rest of the codebase than checking for the presence of the field (to distinguish between explicit and implicit `undefined`)?

TimoGlastra:
I think it's related to the discussion in #2760. We need a way to unset values, and i'm not a fan of distinguishing between key not set and undefined passed (it makes for clunky code on the caller side):
```
{
  ...(metadataSigner ? { metadataSigner } : {})
}
```

I prefer null, and it's a pattern we can start to adopt.

TimoGlastra (approved):

PR #2887, opened 2026-07-29
feat(vc): signature suite registration

rmlearney-digicatapult:
Address `TODO` in `SignatureSuiteRegistry.ts` by registering `SignatureSuitesRegistry` on `W3cCredentialsModule.ts` and constructor-style registration of both signature suites within the registry.

Backwards-compatible token registration with deprecation marker for removal in `v0.8`.

Necessary change for simplifying final fix for `json-ld` strict parsing requirements when using `ed25519-2020` suite (next PR after merge).

Required updates throughout multiple test suites, mostly basic changes.


packages/core/src/modules/vc/linked-data-proofs/SignatureSuiteRegistry.ts
```diff
@@ -1,12 +1,11 @@
 import { CredoError } from '../../../error'
-import { injectAll, injectable } from '../../../plugins'
+import { injectable } from '../../../plugins'
 import { PublicJwk, type SupportedPublicJwkClass } from '../../kms/jwk/PublicJwk'
 
 import { suites } from './adapters/jsonld-signatures-adapter'
 
 const LinkedDataSignature = suites.LinkedDataSignature
 
-export const SignatureSuiteToken = Symbol('SignatureSuiteToken')
```
TimoGlastra, line 9: If we want to make this a non-breaking addition, we should keep the token for now and still inject them (external registration outside of core repo). We can mark it as deprecated, and remove it in 0.8

packages/core/src/modules/vc/W3cCredentialsModule.ts
```diff
@@ -41,20 +37,23 @@ export class W3cCredentialsModule implements Module {
...
-      ],
-      supportedPublicJwkTypes: [Ed25519PublicJwk],
-    } satisfies SuiteInfo)
-    dependencyManager.registerInstance(SignatureSuiteToken, {
-      suiteClass: Ed25519Signature2020,
-      proofType: 'Ed25519Signature2020',
-      verificationMethodTypes: [VERIFICATION_METHOD_TYPE_ED25519_VERIFICATION_KEY_2020],
-      supportedPublicJwkTypes: [Ed25519PublicJwk],
-    } satisfies SuiteInfo)
+    const signatureSuiteRegistry = dependencyManager.resolve(SignatureSuiteRegistry)
```
TimoGlastra, line 43: I think we need to register this first as a singleton before we can register it. Since we always use agent scoped registraiotn ,we don't use `@singleton` from tsyringe, but then we first need to registerSingleton on the dependency manager.

So it would need to happen in the base agent then as we need to make sure it runs before any code calls resolve on the SignatureSuiteRegistry. This is the reason it has this kinda hacky approach.

rmlearney-digicatapult: I think we're safe to register inside `W3cCredentialsModule` and not on the base agent.

Registration order on the base agent is:
* EventEmitter
* JwsService
* StorageVersionRepository
* StorageUpdateService
* AgentConfig
* AgentDependencies
* Stop$
* Filesystem

Then default modules:
* dcql <- resolves `SignatureSuiteRegistry` lazily from request-handling method, not during `DcqlModule.register()`
* genericRecords
* dids
* w3cCredentials <- SignatureSuiteRegistry singleton registered here and Ed25519 suites populated

Unless a user passes an alternative `w3cCredentials` overriding the default and forgets to call `registerSingleton(SignatureSuiteRegistry)` we should be safe.

packages/core/src/modules/vc/linked-data-proofs/SignatureSuiteRegistry.ts
rmlearney-digicatapult: Done :)

rmlearney-digicatapult:
Conformance test issue #2888

packages/core/src/modules/vc/W3cCredentialsModule.ts
rmlearney-digicatapult: One alternative could be to start to migrate all crypto functionality into core registrable modules on the agent itself (JOSE/COSE, W3C-DI, etc.) so that these are available across all higher level callers?

rmlearney-digicatapult:
@TimoGlastra the goal with this is to allow signature suites to pass their compatible @context at registration, so that JSON-LD credential composition can easily match @context to pass the strict parsing requirement. I'm current facing issues with strict mode because context for Eddsa25519-2020 is not compatible with the DID Core context (which was finalised in 2018 and only contains the context for the Eddsa25519-2018 suite)

genaris (approved):
We can either merge it as is (non-breaking) or directly as breaking if we decide to release 0.8.0 (dropping node 20 support) right now.

TimoGlastra:
Yeah making the breaking change now is fine by me

TimoGlastra:
We should also do it in the document loader then, so it's moved to the new module completely

PR #2890, opened 2026-07-31
fix(openid4vc): allow a DPoP key that differs from the client attestation confirmation key

### LINKED ISSUE #2889: [openid4vc] PAR rejects separate DPoP and Client Attestation PoP keys allowed by draft 10
The OpenID4VC issuer appears to require the DPoP public key to match the confirmation key from the OAuth Client Attestation whenever both DPoP and client attestation authentication are used at the PAR endpoint.

This prevents interoperability with clients using the normal, non-combined mode described in [draft-ietf-oauth-attestation-based-client-auth-10](https://datatracker.ietf.org/doc/draft-ietf-oauth-attestation-based-client-auth/) (Section 5.2).

In normal mode, a client can send:

* `OAuth-Client-Attestation`
* `OAuth-Client-Attestation-PoP`
* `DPoP`

The `OAuth-Client-Attestation-PoP` proves possession of the Client Instance Key referenced by the attestation’s `cnf` claim.

The DPoP proof is validated independently according to RFC 9449 and its public key is not required to match the Client Instance Key.

Key equality is required only when using DPoP combined mode, where the dedicated `OAuth-Client-Attestation-PoP` header is omitted and the DPoP proof serves both purposes.

Credo currently passes the following option unconditionally when verifying the pushed authorization request:

```ts
ensureConfirmationKeyMatchesDpopKey: true
```

This check should only be enabled for DPoP combined mode, not when a dedicated `OAuth-Client-Attestation-PoP` header is present.

hacdias:
Fixes #2889.

The pushed authorization request and authorization challenge endpoints passed `ensureConfirmationKeyMatchesDpopKey: true`, so a client presenting `OAuth-Client-Attestation` + `OAuth-Client-Attestation-PoP` alongside a DPoP proof signed with a separate key was rejected with `invalid_request`. That breaks any wallet keeping its client instance key separate from its DPoP key.

### Spec

The issue cites draft-10 §5.2, but the decisive text is §7.2/§7.3, which is already present in **draft-09** — the version we target:

- **§7.2** (Client Attestation PoP JWT) lists nine MUST rules for the PoP-header method; none mention the DPoP key. It opens with *"When operating in DPoP combined mode as defined in Section 5.2, this section does not apply; instead, see Section 7.3."*
- **§7.3** (DPoP Combined Mode) is where the equality MUST lives, and it opens with the mirror statement: *"When the Client Attestation PoP JWT is used as the Proof of Possession instead, this section does not apply."*

Draft-10 then makes it explicit — §7 intro: *"If the request contains an OAuth-Client-Attestation-PoP HTTP request header field, the receiving server MUST apply the validation rules of Section 7.2 and if present, a DPoP proof present in the request is validated according to [RFC9449] **independently of this specification**."* And §5.2: *"its public key is **not required to match** the key in the cnf claim"*, plus *"When using dpop_jkt, the normal mode has to be used"* — which makes combined mode inapplicable at PAR by construction.

Binding the interaction to the client instance is still achieved the way §9.4 intends: client attestation at PAR and again at the token endpoint (`issuanceSession.clientId` ← attestation `sub`, `expectedClientId`).

### Notes

- Removal rather than a conditional: `verifyAuthorizationRequestClientAttestation` in `@openid4vc/oauth2` hard-requires both the attestation and PoP JWTs, so combined mode is unreachable at these endpoints and a gated flag would be dead code. The token endpoint's combined-mode path (`verifyAccessTokenRequestClientAttestationDpop`) enforces equality itself, independent of this option — unaffected by this change.
- Second commit is comment-only: the `we might want to enforce this? Not sure` note in `accessTokenEndpoint.ts` invited reintroducing the same bug for `attest_jwt_client_auth`. Drop it if you'd rather keep the diff to the three broken call sites.

### Tests

Two e2e regression tests in `openid4vc-wallet-key-attestation.e2e.test.ts`, both forcing the standard method with a DPoP key that is not the attestation `cnf.jwk`. Existing tests never hit this because our holder reuses the instance key as its DPoP key whenever the issuer advertises `attest_jwt_client_auth_dpop`.

With the flag restored, both fail with *"Expected the DPoP JWK thumbprint value to match…"*. Reverting only the PAR site makes the second pass while the first still fails, confirming the three call sites are independently covered.

### DIFF
--- a/packages/openid4vc/src/openid4vc-issuer/router/pushedAuthorizationRequestEndpoint.ts
+++ b/packages/openid4vc/src/openid4vc-issuer/router/pushedAuthorizationRequestEndpoint.ts
@@ -94,10 +94,8 @@ export async function handlePushedAuthorizationRequest(
       ...parsedAuthorizationRequest.clientAttestation,
       // First session config, fall back to global config
       required: issuanceSession.walletAttestation?.required ?? config.walletAttestationsRequired,
-      // draft 09 §5.2: bind the interaction to the client instance key. When a client attestation is
-      // presented alongside DPoP, the DPoP key must be the attestation's confirmation key, so the
-      // DPoP-bound method can be used consistently at the token endpoint.
-      ensureConfirmationKeyMatchesDpopKey: true,
+      // NOTE: `ensureConfirmationKeyMatchesDpopKey` is intentionally not set. Per draft §7.2/§7.3 the DPoP key
+      // only has to match the attestation `cnf.jwk` in DPoP combined mode, which this endpoint doesn't support.
     },
     dpop: {
       ...parsedAuthorizationRequest.dpop,
--- a/.changeset/client-attestation-separate-dpop-key.md
+++ b/.changeset/client-attestation-separate-dpop-key.md
@@ -1,0 +1,5 @@
+---
+"@credo-ts/openid4vc": patch
+---
+
+Don't require the DPoP proof to use the client attestation confirmation key at the pushed authorization request and authorization challenge endpoints. Per the attestation-based client authentication draft, the DPoP key and the client instance key only have to be the same for the `attest_jwt_client_auth_dpop` method (DPoP combined mode), where a single DPoP proof replaces the client attestation PoP JWT. When the dedicated `OAuth-Client-Attestation-PoP` header is present, the DPoP proof is validated according to RFC 9449 independently, and its public key is not required to match the `cnf` claim of the client attestation. The issuer wrongly enforced key equality for both methods, rejecting interoperable wallets that use a separate DPoP key. Key equality for the DPoP-bound method at the token endpoint is unaffected.
--- a/packages/openid4vc/src/openid4vc-issuer/router/accessTokenEndpoint.ts
+++ b/packages/openid4vc/src/openid4vc-issuer/router/accessTokenEndpoint.ts
@@ -138,8 +138,9 @@ export function handleTokenRequest(config: OpenId4VcIssuerModuleConfig) {
             // First session config, fall back to global config
             required: issuanceSession.walletAttestation?.required ?? config.walletAttestationsRequired,
 
-            // NOTE: we might want to enforce this? Not sure
-            // ensureConfirmationKeyMatchesDpopKey: true
+            // NOTE: `ensureConfirmationKeyMatchesDpopKey` is intentionally not enabled. Per draft §7.2 a request
+            // with an `OAuth-Client-Attestation-PoP` header has its DPoP proof verified per RFC 9449
+            // independently. The library enforces key equality itself for DPoP combined mode (§7.3).
           },
           dpop: {
             ...dpop,
@@ -182,8 +183,9 @@ export function handleTokenRequest(config: OpenId4VcIssuerModuleConfig) {
             // set required to true previously if client attestations were provided or required.
             required: issuanceSession.walletAttestation?.required,
 
-            // NOTE: we might want to enforce this? Not sure
-            // ensureConfirmationKeyMatchesDpopKey: true
+            // NOTE: `ensureConfirmationKeyMatchesDpopKey` is intentionally not enabled. Per draft §7.2 a request
+            // with an `OAuth-Client-Attestation-PoP` header has its DPoP proof verified per RFC 9449
+            // independently. The library enforces key equality itself for DPoP combined mode (§7.3).
           },
           dpop: {
             ...dpop,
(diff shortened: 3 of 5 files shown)


packages/openid4vc/src/openid4vc-issuer/router/pushedAuthorizationRequestEndpoint.ts
```diff
@@ -94,10 +94,8 @@ export async function handlePushedAuthorizationRequest(
       ...parsedAuthorizationRequest.clientAttestation,
       // First session config, fall back to global config
       required: issuanceSession.walletAttestation?.required ?? config.walletAttestationsRequired,
-      // draft 09 §5.2: bind the interaction to the client instance key. When a client attestation is
-      // presented alongside DPoP, the DPoP key must be the attestation's confirmation key, so the
-      // DPoP-bound method can be used consistently at the token endpoint.
-      ensureConfirmationKeyMatchesDpopKey: true,
+      // NOTE: `ensureConfirmationKeyMatchesDpopKey` is intentionally not set. Per draft §7.2/§7.3 the DPoP key
+      // only has to match the attestation `cnf.jwk` in DPoP combined mode, which this endpoint doesn't support.
```
TimoGlastra, line 98: Hmm but i think in EUDI context you might want to enable/require this no? Since the wallet attestation key is attested by the Wallet Provider, while the DPoP key by itself is not.

But maybe we should then enforce that by only allowing the combined mode? And if you accept the non-combined mode, they never have to match?

Do we support limiting these methods?

hacdias:
Btw, the conformance tests are failing now on the main branch too: https://github.com/openwallet-foundation/credo-ts/actions/runs/30476131094

Most likely something changed in the tests.

packages/openid4vc/src/openid4vc-issuer/router/pushedAuthorizationRequestEndpoint.ts
hacdias: > But maybe we should then enforce that by only allowing the combined mode? And if you accept the non-combined mode, they never have to match?

This is done by the PR. The oid4vc-ts library takes care of this.

> Do we support limiting these methods?

Not manually, it's inferred from the request.

---

Claude's more in-depth answer.

> Good instinct, and EUDI actually went down this exact path — then reversed it. [TS3 (Wallet Unit Attestation)](https://github.com/eu-digital-identity-wallet/eudi-doc-standards-and-technical-specifications/blob/main/docs/technical-specifications/ts3-wallet-unit-attestation.md#versioning) v1.5.1 (2026-05-08) added "the Wallet Unit SHALL use the WIA cnf key as the DPoP key when requesting an Access Token", and v1.5.2 (2026-05-26) "rolled back the WIA cnf/DPoP requirements added in 1.5.1". It's gone from [§2.2.1.1](https://github.com/eu-digital-identity-wallet/eudi-doc-standards-and-technical-specifications/blob/main/docs/technical-specifications/ts3-wallet-unit-attestation.md#2211-wallet-providers-responsibilities-for-transport-of-wias) today.
> 
> Two things worth noting even about the 1.5.1 version: it was a Wallet Unit obligation, not an issuer check — and [§2.2.1.2](https://github.com/eu-digital-identity-wallet/eudi-doc-standards-and-technical-specifications/blob/main/docs/technical-specifications/ts3-wallet-unit-attestation.md#2212-pid-providers-and-attestation-providers-responsibilities-for-transport-of-wias), which lists what a PID/Attestation Provider SHALL verify about a WIA, has only three checks (signature + trust chain, expiry, PoP under cnf). No DPoP equality, then or now.
> 
> If you do want the token bound to an attested key, the spec's lever is the method, not the check: that's what attest_jwt_client_auth_dpop is for ([§9.1](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-attestation-based-client-auth-10#section-9.1) — "authenticates (attests) the DPoP key used for sender-constraining tokens"). Forcing equality inside the PoP-header method gets the same property by violating [§7.2](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-attestation-based-client-auth-10#section-7.2), and undiscoverably — nothing in metadata says "I require equality", so a conformant wallet just gets a 401. That's #2889.
> 
> Can we require the DPoP-bound method instead? Only for pre-auth today. [verifyAuthorizationRequestClientAttestation](https://github.com/openwallet-foundation-labs/oid4vc-ts/blob/v0.5.4/packages/oauth2/src/authorization-request/verify-authorization-request.ts#L147) requires both JWTs, and PAR + authorization challenge both route into it — consistent with [draft-10 §5.2](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-attestation-based-client-auth-10#section-5.2) ("When using dpop_jkt, the normal mode has to be used"). And limiting methods is half-supported: [derived, not configurable](https://github.com/openwallet-foundation/credo-ts/blob/d527e34c06149b292f16dc099d5308c8307f57cf/packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerService.ts#L1393-L1405), only enforced for the DPoP-bound path, and [isClientAttestationSupported](https://github.com/openwallet-foundation-labs/oid4vc-ts/blob/v0.5.4/packages/oauth2/src/Oauth2Client.ts#L67) only looks for attest_jwt_client_auth, so our own holder would send no attestation at all.
> 
> Suggest keeping this as the conformance fix and tracking "require the DPoP-bound method" separately.

TimoGlastra: Okay, so we could add an option to the issuerRecord to limit the supported methods. And if we only supported the combined mode, we should not allow `dpop_jkt` to be used (just the header appraoch).

But maybe that's a separate issue we need to track

TimoGlastra (approved):

PR #2894, opened 2026-08-02
fix(openid4vc): preserve optional issuer metadata fields on update

Sshovon:


Sshovon:
Hey @TimoGlastra continuing our discussion from #2760. Switched to null for unsetting instead of checking key presence. So it works the same as metadataSigner in #2885 now: leave a field out to keep it, pass null to remove it, pass a value to replace it.

I left _credentialConfigurationsSupported_ required and always replaced, since an issuer isn't valid without it and null wouldn't mean anything there. Would like your opinion on it though: we could make it optional as well, so you don't have to resend the whole map just to change the display. Didn't want to widen the scope here.

I also added a small Nullable helper in shared/types.ts, figured it could be useful if we start using this convention in more places.

PR #2898, opened 2026-08-03
feat(didcomm): support VCDM 2.0 credentials over RFC0809

genaris:
Adds support for exchanging W3C VC Data Model 2.0 credentials over Aries RFC 0809 (Data Integrity credential attachment format), and moves that format out of `@credo-ts/anoncreds` into `@credo-ts/didcomm`, where the rest of the format already lives.

No breaking changes are introduced. `DataIntegrityDidCommCredentialFormatService` is still exported from `@credo-ts/anoncreds` as a deprecated alias to keep compatibility with the 0.7.x baseline.

### Format moved to `@credo-ts/didcomm`
- New `DidCommDataIntegrityCredentialFormatService` in `@credo-ts/didcomm`. `DataIntegrityDidCommCredentialFormatService` in `@credo-ts/anoncreds` remains as a deprecated alias.
- The `anoncreds_link_secret` binding method is provided through a new `DidCommDataIntegrityLinkSecretBindingProvider` interface. `AnonCredsModule` registers the anoncreds implementation, so existing agents need no changes. Using the binding method without `AnonCredsModule` fails with a clear error.
- Issuing a plain W3C credential over this format no longer requires `@credo-ts/anoncreds`.
- `deleteCredentialById` no longer routes every credential of this format through the anoncreds holder service. It looks the record up in the data model 1.1 store and then the 2.0 store; a 1.1 record is deleted through the link secret binding provider only when it carries the anoncreds metadata, otherwise it is removed as the plain `W3cCredentialRecord` or `W3cV2CredentialRecord` it is.

### Data Model 2.0 support
- `data_model_versions_supported` is derived from the base JSON-LD context of the offered credential (the first `@context` entry, as both data model versions require), and the holder echoes its choice as `data_model_version` on the request, as RFC 0809 describes.
- New `cryptosuite` format option (e.g. `eddsa-jcs-2022`) for 2.0 credentials, ignored for 1.1. When omitted, the first registered cryptosuite supporting the key type of the issuer verification method is used. `anoncreds-2023` is rejected explicitly, as it is only produced through the link secret binding method.
- `W3cDataIntegrityApi.getSupportedCryptosuites` accepts an optional public JWK type filter, which the fallback above relies on.
- Received 2.0 credentials are verified and stored as `W3cV2CredentialRecord`. The `credentialRecordType` on the exchange record stays `'w3c'` for both versions, since it is what the credential protocol uses to route deletion back to this format service.
- The anoncreds link secret binding is rejected for 2.0 credentials, as it is defined for data model 1.1 and `anoncredsvc-2023` only.

### Fixes
- The comparison of the offered and received credential subject rejected any claim with a falsy value (`0`, `false`, `''`). Claims are now compared by presence.

Enveloped 2.0 credentials (JWT, SD-JWT) are out of scope: RFC 0809 only carries embedded Data Integrity proofs. SD-JWT over DIDComm is covered by RFC 881 in #2846.


genaris:
@TimoGlastra @rmlearney-digicatapult I'm about to test this in our stack to see that everything works as expected, but I think it's ready to review, so I'd appreciate if you can take a look at it.

packages/didcomm/src/modules/credentials/formats/dataIntegrity/DidCommDataIntegrityCredentialFormatService.ts
```diff
@@ -0,0 +1,1050 @@
...
+      await linkSecretBindingProvider?.deleteCredentialById(agentContext, credentialRecordId)
+      return
+    }
+
+    const w3cCredentialService = agentContext.dependencyManager.resolve(W3cCredentialService)
+    const w3cV2CredentialService = agentContext.dependencyManager.resolve(W3cV2CredentialService)
+
+    try {
+      await w3cCredentialService.getCredentialRecordById(agentContext, credentialRecordId)
+    } catch {
```
rmlearney-digicatapult, line 826: This discriminates `W3c` from `W3cV2` records by catching a lookup failure in the 1.1 store and inferring 'must be 2.0'. Since the two tables are independent, this only works if IDs never collide.

The bare `catch` also swallows failures in the storage layer and routes into the 2.0 delete which could then fail itself.

We should prefer an explicit lookup-then-fail on each credential repository because the presence in the store is the version signal.

packages/didcomm/src/modules/credentials/formats/dataIntegrity/DidCommDataIntegrityCredentialFormatService.ts
```diff
@@ -0,0 +1,1050 @@
...
+      if (result && !result.isValid) {
+        throw new CredoError(`Failed to validate credential, error = ${result.error}`)
+      }
+
+      const w3cV2CredentialRecord = await w3cV2CredentialService.storeCredential(agentContext, {
+        record: W3cV2CredentialRecord.fromCredential(credential),
+      })
+
+      credentialExchangeRecord.credentials.push({
+        credentialRecordType: 'w3c-v2',
```
rmlearney-digicatapult, line 753: Line 84 defines the `credentialRecordType` as `w3c`. Using `w3c-v2` will lead to the format service throwing in delete paths leaving the `W3cV2CredentialRecord` orphaned with no exchange record left to try from.

The test in `data-integrity-flow-vcdm2.test.ts` line 156 asserts the `w3c-v2` binding but does not exercise deletion to identify this.

Because `deleteCredentialById` already covers both `W3c` and `W3cV2` (see my note there) we can make this `credentialRecordId: this.credentialRecordType` and narrow line 60 in `DidCommDataIntegrityCredentialFormat.ts` back to `w3c` which represents the format, not the storage model.

packages/anoncreds/src/formats/AnonCredsLinkSecretBindingProvider.ts
```diff
@@ -0,0 +1,408 @@
...
+
+    return w3cCredentialRecord.id
+  }
+
+  public async ownsCredentialRecord(agentContext: AgentContext, credentialRecordId: string): Promise<boolean> {
+    const w3cCredentialService = agentContext.dependencyManager.resolve(W3cCredentialService)
+
+    try {
+      const w3cCredentialRecord = await w3cCredentialService.getCredentialRecordById(agentContext, credentialRecordId)
+      return getAnonCredsTagsFromRecord(w3cCredentialRecord) !== undefined
```
rmlearney-digicatapult, line 359: `getAnonCredsTagsFromRecord` will return `undefined` if any of the seven independent tags are missing and will lead to the W3C record being deleted.

We could test presence of `W3cAnonCredsCredentialMetadataKey` instead for a single check (`w3cCredentialRecord.metadata.get(W3cAnonCredsCredentialMetadataKey) !== undefined`)

packages/didcomm/src/modules/credentials/formats/dataIntegrity/DidCommDataIntegrityCredentialFormatService.ts
```diff
@@ -0,0 +1,1050 @@
...
+    }
+
+    return provider
+  }
+
+  private getCredentialVersion(credentialJson: JsonObject): W3C_VC_DATA_MODEL_VERSION {
+    const context = credentialJson['@context']
+    if (!context || !Array.isArray(context)) throw new CredoError('Invalid @context in credential offer')
+
+    const isV1Credential = context.find((c) => c === 'https://www.w3.org/2018/credentials/v1')
```
rmlearney-digicatapult, line 182: we should compare the first entry only `context[0]` by definition, in case a malformed credential carries both

packages/didcomm/src/modules/credentials/formats/dataIntegrity/DidCommDataIntegrityCredentialFormat.ts
```diff
@@ -42,11 +42,22 @@ export interface DidCommDataIntegrityOfferCredentialFormat {
 export interface DidCommDataIntegrityAcceptRequestFormat {
   credentialSubjectId?: string
   issuerVerificationMethod?: string
+
+  /**
+   * The Data Integrity cryptosuite to secure a data model 2.0 credential with, for example
+   * `eddsa-jcs-2022`.
+   *
+   * RFC 0809 leaves the choice of cryptosuite to the issuer, so this is not negotiated with the
```
rmlearney-digicatapult, line 50: Automatic fallback logic not yet implemented in `signV2Credential` in `DidCommDataIntegrityCredentialFormatService.ts` which currently throws if `cryptosuite` is missing. 

The cryptosuite registry allows querying first-match registered cryptosuites by `W3cDataIntegrityCryptosuiteRegistry.getAllByPublicJwkType()` and `signV2Credential` already resolves the VM immediately after the throw.

packages/didcomm/src/modules/credentials/formats/dataIntegrity/DidCommDataIntegrityCredentialFormatService.ts
```diff
@@ -0,0 +1,1050 @@
...
+   *
+   * RFC 0809 leaves the choice of cryptosuite to the issuer, so it is not negotiated with the holder
+   * and has to be provided by the issuer.
+   */
+  private async signV2Credential(
+    agentContext: AgentContext,
+    credential: W3cV2Credential,
+    { cryptosuite, issuerVerificationMethod }: { cryptosuite?: string; issuerVerificationMethod?: string }
+  ) {
+    if (!cryptosuite) {
```
rmlearney-digicatapult, line 504: we should explicitly reject `anoncreds-2023` to prevent it falling through by accident. It does not resolve to a cryptosuite.

packages/didcomm/src/modules/credentials/formats/dataIntegrity/DidCommDataIntegrityCredentialFormatService.ts
```diff
@@ -0,0 +1,1050 @@
...
+      !anonCredsLinkSecretDataIntegrityBindingProof && !didCommSignedAttachmentBindingProof
+        ? undefined
+        : {
+            anoncreds_link_secret: anonCredsLinkSecretDataIntegrityBindingProof,
+            didcomm_signed_attachment: didCommSignedAttachmentBindingProof,
+          }
+
+    if (credentialOffer.bindingRequired && !bindingProof) throw new CredoError('Missing required binding proof')
+
+    const dataModelVersion = dataIntegrityFormat?.dataModelVersion ?? credentialOffer.dataModelVersionsSupported[0]
```
rmlearney-digicatapult, line 394: This and the subsequent check both resolve to the single value `const dataModelVersionsSupported: W3C_VC_DATA_MODEL_VERSION[] = [dataModelVersion]` from line 927. Not a big issue, just FYI.

rmlearney-digicatapult (commented):
👍 we previously discussed that VCDM 2.0 supports both enveloped and embedded credential proofs. This work enables the embedded path but we will have to do some thinking about supporting enveloped VCDM 2.0

packages/didcomm/src/modules/credentials/formats/dataIntegrity/DidCommDataIntegrityCredentialFormatService.ts
```diff
@@ -807,29 +833,38 @@ export class DidCommDataIntegrityCredentialFormatService
     return supportedAttachment
   }
 
+  /**
+   * Deletes a credential received over this format. The record type on the exchange record is the
+   * same for every credential of this format, so the store holding the record decides how it is
+   * deleted: a data model 1.1 credential is a `W3cCredentialRecord`, which is deleted through the
```
rmlearney-digicatapult: should we extend this note to include linked data proof vc 1.1 credentials in addition to anoncreds

rmlearney-digicatapult (approved):

PR #2899, opened 2026-08-03
fix(webvh): use a verification method URL in DID log proofs

AirKyzzZ:
`getVerificationMethodId()` returns the bare `did:key:z6Mk…`, so every log entry we sign carries a `proof.verificationMethod` that is a DID rather than a verification method URL. Data Integrity expects that property to point at the method itself, and strict verifiers enforce it.

The Swiss `didresolver` used by the swiyu wallet rejects the whole log over it, so no wallet on that stack can resolve a DID this library creates. The did:webvh examples and other implementations' fixtures all use the `did:key:<mk>#<mk>` form.

Signing is untouched: same key, same encoding, same proof value, only the reference changes. webvh unit suite passes (48), types:check and biome clean.


AirKyzzZ:
@genaris tagging you as requested, this is the upstream fix for the DID log proof verification method we discussed in verana-labs/vs-agent#553.

rmlearney-digicatapult:
Hi @AirKyzzZ we're soon to release a version `3.0.0` of `didwebvh-ts` from https://github.com/decentralized-identity/didwebvh-ts which will have:

* multi-version log resolution
* clean APIs for `controller` versus `witness` functionality

We're also thinking of implementing a `signer`/`verifier` callback so that wrappers can pass in their own proof-aware signers/verifiers (matching the interface of `Data Integrity v1.0`). I see that @genaris has approved #2797 (I'm addressing comments now) which will make this much easier to develop.

This PR could be a stop-gap measure, or I'd encourage you to take a look at the upstream `didwebvh-ts` work because the `credo-ts` wrapper will almost certainly completely change after `3.0.0` is released

genaris:
> This PR could be a stop-gap measure, or I'd encourage you to take a look at the upstream `didwebvh-ts` work because the `credo-ts` wrapper will almost certainly completely change after `3.0.0` is released

@rmlearney-digicatapult Do you think this small fix could represent any drawback towards the upgrade to 3.0.0?

AirKyzzZ:
> Hi @AirKyzzZ we're soon to release a version `3.0.0` of `didwebvh-ts` from https://github.com/decentralized-identity/didwebvh-ts which will have:
> 
> 
> 
> * multi-version log resolution
> 
> * clean APIs for `controller` versus `witness` functionality
> 
> 
> 
> We're also thinking of implementing a `signer`/`verifier` callback so that wrappers can pass in their own proof-aware signers/verifiers (matching the interface of `Data Integrity v1.0`). I see that @genaris has approved #2797 (I'm addressing comments now) which will make this much easier to develop.
> 
> 
> 
> This PR could be a stop-gap measure, or I'd encourage you to take a look at the upstream `didwebvh-ts` work because the `credo-ts` wrapper will almost certainly completely change after `3.0.0` is released

Alright thank you @rmlearney-digicatapult for your answer, im going to take a deeper look at the upstream didwebvh-ts now

rmlearney-digicatapult:
The updated key parser in `didwebvh-ts` enforces that if a `did:key` fragment exists, it must equal the body multibase. The parser explicitly accepts both forms and validates fragment consistency, so I don't foresee any negative impact from this change.

genaris:
> The updated key parser in `didwebvh-ts` enforces that if a `did:key` fragment exists, it must equal the body multibase. The parser explicitly accepts both forms and validates fragment consistency, so I don't foresee any negative impact from this change.

Ok great, let's merge this for now and then upgrade to the newer 3.0.0 once available!

genaris (approved):

TimoGlastra:
What is the impact on existing did:webvh dids? Any way to migrate this? Can we regenerate the log?

rmlearney-digicatapult:
The change actually improves `v2.8.0` because the parser rejects mismatched fragment forms, but there should be no impact on resolving. 

The signature bytes (`proofValue`) for new entries will differ however but it is not a backwards or forwards compatibility break.

Issue #2888, opened 2026-07-30
iso_mdl happy-flow module fails OID4VP verifier conformance test


rmlearney-digicatapult:
Authored by `copilot` after a failing test on PR #2813 

## Description

The `oid4vp-1final-verifier-happy-flow` module fails for the `iso_mdl` variant in the [OID4VP verifier conformance test](https://github.com/openwallet-foundation/credo-ts/blob/7ab90619758cce0ac7fb19c313d2d2caf484f022/packages/openid4vc/tests/conformance/oid4vp-verifier.conformance.test.ts). The `sd_jwt_vc` variant passes successfully.

**Failing CI job:** `90837319566` — [Conformance Tests workflow](https://github.com/openwallet-foundation/credo-ts/actions)

## Error

```
AssertionError: Non-conformant modules:
oid4vp-1final-verifier-happy-flow: result=FAILED status=FINISHED (expected one of PASSED, WARNING, SKIPPED)
— https://localhost:8443/log-detail.html?log=FRzeHDXcwm2bDnv
```

> **Location:** `packages/openid4vc/tests/conformance/oid4vp-verifier.conformance.test.ts:367`

## Root Cause

For the `iso_mdl` variant, the conformance suite uses its own built-in mdoc document signer (a self-signed `certification.openid.net` cert) regardless of the `signing_jwk` configured in the plan. The verifier is set up to trust the entire embedded certificate chain of any presented `Mdoc` credential:

```typescript
// packages/openid4vc/tests/conformance/oid4vp-verifier.conformance.test.ts (lines 278–285)
verifier.agent.x509.config.setTrustedCertificatesForVerification(
  (_agentContext, { verification, certificateChain }) => {
    if (verification.type === 'credential' && verification.credential instanceof Mdoc) {
      return certificateChain.map((certificate) => certificate.toString('pem'))
    }
    return undefined
  }
)
```

As noted in the code comment, the OIDF built-in issuer cert **rotates periodically**. If this cert has expired or rotated since the test was last confirmed passing, mdoc credential verification will fail during the happy-flow, causing the module result to be `FAILED`.

### Update the OIDF conformance suite Docker image

Check whether the conformance suite Docker image is stale and the built-in mdoc issuer cert has rotated. Pulling the latest image should resolve the issue if this is the cause.

## Additional Context

[...]

PR #2903, opened 2026-08-06
test: update OIDF conformance suite image

genaris:
As reported in #2888, OIDF conformance tests are failing. The suite embeds a hardcoded, self-signed mdoc document signer (`CN=certification.openid.net`) whose certificate expired on 2026-07-30, which fails every `iso_mdl` module with "used after it is allowed". 

The fix ([conformance-suite!2123](https://gitlab.com/openid/conformance-suite/-/merge_requests/2123), which regenerated the certificate with validity 2026-08-03 -> 2027-08-03) is merged to master but is not in any release tag yet. Only the mutable `latest` tag carries it, so we pin its digest to keep runs reproducible instead of tracking a moving tag.

Another change here is at `oid4vp-verifier.conformance.test.ts`: narrowed the kb-jwt-iat FIXME tolerance to iat-in-past only. iat-in-future now genuinely passes, so it's no longer masked.


TimoGlastra (approved):
Thanks for this 🙌

PR #2906, opened 2026-08-12
fix(webvh): improve error logging when DID Document is not found

genaris:
Something I've experienced recently when trying to resolve the DID Log of a service that was not yet ready and thus throwing a 503 status code. `didwebvh-ts` surfaces it as a thrown fetch error, but `WebVhDidResolver.resolveDidDoc()` ends up calling `DidDocument.fromJSON(undefined)`, throwing the unhelpful `ClassValidationError: Cannot validate instance of undefined`.

cc @rmlearney-digicatapult @brianorwhatever


rmlearney-digicatapult (approved):
👍

TimoGlastra (approved):

PR #2910, opened 2026-08-18
feat(pex): improve Presentation Exchange 2.1.1 conformance and credential selection

rmlearney-digicatapult:
Implement full spec conformance for `PEX 2.1.1`

* Unify branching logic in `presentationsToCreate.ts` to use `credential.claimFormat` in a switch statement to disambiguate fall-through with runtime guarding by asserting record types in each case with fail-fast CredoError which reports the input descriptor identifier for easier tracing/debugging

* Support recursive `from_nested` submission requirements including `pick` and `all` rules

* Validate presentation and credential format capabilities independently and select compatible authentication key for `LDP-VP` versus `LDP-VC`

* Expose predicate support metadata, but only support predicates on `AnonCreds W3C` paths which have the existing necessary logic

* Handle non-DID subject identifiers

* Preserve input-descriptor ordering in generated submissions

* Workaround for external multi-VP `$[n]` path

* Block PEX for OIDC4VP `v1.0` which now uses `DCQL`

* Split monolithic test suite into individual stage-focused suites

* Skip failing `draft-24` SD-JWT tests - see note below

**NB**

Existing `draft-24` test suite (`packages/openid4vc/tests/openid4vp-draft24.drizzle.e2e.test.ts`) now fails because we have introduced format validation. 

`main` was more permissive and ignored descriptor credential-format declarations during creation, silently accepting the mismatch between the fixture declared `vc+sd-jwt` and the internal Credo-ts representation of `dc+sd-jwt`.

To support `dc+sd-jwt` in `draft-24` paths I'd like to submit some small changes to `@animo-id/pex` and `@sphereon/pex-models` but given the move to DCQL in `v1.0` this is only a nice-to-have to round out support and we should probably modify the tests themselves to use `v1.0`


TimoGlastra:
Thanks for the PR @rmlearney-digicatapult. 

The PEX-OID4VP integration is in maintenance mode, and so i want to be careful with shipping changes there.

I do think dc+sd-jwt support would be good in PEX, but we're struggling with the PEX library not being maintained anymore. We made a fork to push necessary fixes, but we're only still using PEX for legacy openid4vp support and will stop using it after that. If we want to support dc+sd-jwt we also need to fork/release the underlying models library. I think someone should write a DIDComm presentation protocol using DCQL.

I have previously proposed we move PEX support to the DIDComm library once we remove PEX/OID4VP integration, as it pulls in heavy dependencies i would like to move out of core.

Just some thoughts on the PEX integration, i will review the PR more in depth, but we need to figure out the maintenance approach of PEX in Credo.

rmlearney-digicatapult:
@TimoGlastra completely agree with that direction. This PR blocks `v1.0` OIDC4VP from PEX and has workarounds for `draft-21` but `draft-24` support would require changes to the dependencies, which as you have said might not happen.

It would be a shame not to 'finish' PEX, and I also agree that a future `present-proof-v4` should probably specifically link `DIDComm v2` and `DCQL`

TimoGlastra:
Ideally we remove all format specific processing from the pex library and provide the parsed credentials to PEX, similar to DCQL, then the lib will be a lot lighter, but this will take significant engineering effort (basically a rewrite)

rmlearney-digicatapult:
Ah interesting. Yes agreed.

If we're planning to strip `draft` OIDC4VP support from Credo and only implement `v1.0` we could break the link completely and keep PEX focused on DIDComm paths (until DCQL-over-DIDComm arrives)

TimoGlastra:
> If we're planning to strip draft OIDC4VP support from Credo and only implement v1.0 we could break the link completely and keep PEX focused on DIDComm paths (until DCQL-over-DIDComm arrives)

Eventually, yes. I think we can push it in Credo 0.8, but good to keep it in for now.

PR #2911, opened 2026-08-18
feat(openid4vc): validate chained scope mappings before PAR

rmlearney-digicatapult:
Validate static chained authorization-server scope mappings when an Authorization-code credential offer is created. Use same validation during PAR processing.

* Configuration errors now detected early in flows
Code previously allowed auth-code offer to be created even when chained server had no mapping for one of the offered credential scopes. Failure was deferred until wallet submitted PAR request

* Improve error messages for PAR
PAR previously returned generic OAuth `server_error` for both missing mapping cases. Now reports missing internal scope and upstream auth server and PAR translates into specific OAuth error response

* No duplicated mapping logic
Both offer creation and PAR now use shared helper centralised in issuer service

* Separated PAR validation from request handling
Issuer config error previously only appeared after interaction had begun. Static mapping validation now happens at offer creation and PAR still performs same validation as defensive check.

### DIFF
--- a/packages/openid4vc/src/openid4vc-issuer/router/pushedAuthorizationRequestEndpoint.ts
+++ b/packages/openid4vc/src/openid4vc-issuer/router/pushedAuthorizationRequestEndpoint.ts
@@ -215,7 +215,6 @@ export async function handlePushedAuthorizationRequest(
       throw new Oauth2ServerErrorResponseError(
         {
           error: Oauth2ErrorCodes.ServerError,
-          error_description: 'Invalid chained authorization server scope mapping.',
         },
         {
           internalMessage: `Invalid scope mapping for chained authorization server '${authorizationServerConfig.issuer}'.`,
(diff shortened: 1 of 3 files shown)


packages/openid4vc/src/openid4vc-issuer/router/pushedAuthorizationRequestEndpoint.ts
```diff
@@ -206,30 +206,22 @@ export async function handlePushedAuthorizationRequest(
...
+    try {
+      scopes = openId4VcIssuerService.getStaticChainedAuthorizationScopes({
+        authorizationServerConfig,
+        requestedScopes,
+      })
+    } catch (error) {
+      throw new Oauth2ServerErrorResponseError(
+        {
+          error: Oauth2ErrorCodes.ServerError,
+          error_description: 'Invalid chained authorization server scope mapping.',
```
TimoGlastra: I think this should not be communicated to the client, it's actually a server error and not relevant to the client

TimoGlastra (approved):
One note, otherwose LGTM

PR #2912, opened 2026-08-18
feat(openid4vc): retry logic for chained authorization

rmlearney-digicatapult:
Discovered on testing chained authorization when upstream IdP fails during callback processing

Add `AuthorizationRetryable` state for existing issuance sessions & reset logic inside `OpenId4VcIssuanceSessionRecord`

Holder now able to retry authorization from same offer with stale authorization data cleared and fresh upstream authorization state and PKCE material generated.

Added test covering upstream failure & recovery


TimoGlastra:
There's quite some other states which i think would also be final.

Isn't it common in this case to create a new offer? Should we limit it to a slimmer set of errors? Do you have an example where you encountered this?

rmlearney-digicatapult:
Hi @TimoGlastra the use case for this was more narrow. 

Credo currently persists an issuance session with `AuthorizationInitiated`, but if the IdP (Keycloak in my case) goes down or there's a network glitch so that metadata retrieval, token exchange or ID-token verification fails, the catch block returns an OAuth error but does not update the issuance session. So Credo remains locked in `AuthorizationInitiated`. This means the holder cannot restart auth with the same offer.

This change is designed to unblock the case where the IdP goes down temporarily by moving the state to `AuthorizationRetryable`. The system generates fresh security tokens for the new attempt and completes the exchange successfully.

TimoGlastra:
okay, so it's twofold:
- correctly change the session state (it could also be changed to Error?)
- allow retry of the authorization in some cases

rmlearney-digicatapult:
We could change to error and drop/clear state and make the user go back to the start, also a viable option

rmlearney-digicatapult:
@TimoGlastra do you think it's safer to error and drop/clear state to make the user go back to the start? Trying to think of side-effects of erroring out if the IdP goes down temporarily

Issue #2915, opened 2026-08-21
Allow DIDComm HTTP inbound transport to be externally hosted, matching OpenID4VC


rmlearney-digicatapult:
When Credo-TS was only built around DIDComm, the `app` option assumed the transport could bind whatever app it was given. But with the release of OpenID4VC support users might want 3 different configurations:

* DIDComm-only listener
* OpenID4VC-only listener
* Combined listeners

The current shape of `@credo-ts/node` doesn't allow one ownership model across all 3. Enabling DIDComm forces Credo-owned while OpenID4VC is host-owned by design. The WS transport can accept an injected server, but still closes it on `stop()`.

`OpenId4VcModuleConfig` accepts a caller-owned `Express` app, registers its routes during `agent.initialize()`, and never binds so the host owns the `http.Server`. `DidCommHttpInboundTransport` also accepts an `app`, but `port` is required and `start()` ends with `this.app.listen(this.port)`, so the transport is both route registrar and listener owner.

A host running both DIDComm and OpenID4VC therefore cannot use one public listener without a workaround, and in an `OpenID4VC`-only configuration no public listener is created at all.

`DidCommWsInboundTransport` already solves the equivalent problem with an exclusive union for `{ server }` or `{ port }` so DIDComm HTTP is the outlier within its own package.

We could do the following union for DIDComm HTTP transport:

```ts
| { app: Express; port?: undefined; path?: string; processedMessageListenerTimeoutMs?: number }
| { app?: Express; port: number; path?: string; processedMessageListenerTimeoutMs?: number }
```

With this, existing `{ app, port }` behaviour is unchanged. Credo registers the route and returns, the host creates, binds, and closes the server. Route registration is extracted from `start()` so message receipt, sessions, return-route and timeout behaviour are untouched. The `text()` body parser moves from the constructor's `app.use` onto the `POST` route, so a shared `app` carries no side effects because it's MIME-filtered.

I'd also propose two additional fixes: `app.listen()` currently has no error or listening handling, so `EADDRINUSE` is an uncaught exception rather than rejecting `initialize()` and `WsInboundTransport.stop()` closes the server without terminating open sockets, which can stall shutdown. Should `stop()` stop closing a _supplied_ `WebSocketServer` (breaking change)?

Issue #2763, opened 2026-04-24
Loosen dependency on Express for openid4vc


berendsliedrecht:
Right now, if I understand it correctly, we have such a tight dependency on Express that we do not allow any other web application framework to be used for the required openid4vc endpoints. (if this is incorrect please close this issue :)).

My idea was to replace this strict `app: Express` type and just accept an object with some basic properties where credo could add some of the required endpoints. We would have to go through all the uses of the `.app` in the framework, but it does not seem like a lot. For instance, [this whole function](https://github.com/openwallet-foundation/credo-ts/blob/b1b1cfa4156764b33247bdba15884796ac1ecc8d/packages/openid4vc/src/openid4vc-verifier/OpenId4VcVerifierModule.ts#L49) would need to accept a generic web framework to set all the configuration on.

We should, in the node or openid4vc package maybe?, supply an express adapter/app which will make it work exactly as how it worked before.

Maybe an interface like:
```typescript
export type WebApplication = {
  setPostEndpoint(path: string, cb: (request: Request, response: Response) => unknown): void
  setGetEndpoint(...): ...
  // etc.
}
```

Discussion is welcome :).

https://github.com/openwallet-foundation/credo-ts/blob/b1b1cfa4156764b33247bdba15884796ac1ecc8d/packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerModuleConfig.ts#L27-L28

PR #2916, opened 2026-08-21
feat(node): support host-owned DIDComm HTTP listeners

rmlearney-digicatapult:
Addresses #2915 and sets the groundwork for #2763 for DIDComm as well

The optional `app?: Express` in the transport constructor allows the host to supply the app, but Credo binds it (`start()` ends with `this._server = this.app.listen(this.port)`) so the host cannot control the socket. These changes permit moving the lifecycle boundary outside Credo so the socket is host-owned and the listener exists whether or not DIDComm is enabled.

* Credo-owned HTTP listener now awaits `listening` and rejects bind failures
* Supports registering DIDComm HTTP routes on host-owned `Express` app without creating or closing its `http.server`
* Existing port-based DIDComm HTTP transport unchanged but ownership model now aligned with DIDComm WS in same package
* DIDComm body parsing moved to DIDComm route so unrelated routes on shared `Express` app unaffected
* Terminate active DIDComm WebSocket clients during shutdown so shutdown no longer hangs on live DIDComm connection
* Test suite


TimoGlastra (commented):
I like this change, not sure if the port should be the trigger between it hosting the server or not? It feels a bit like a big side effect for a single param.

rmlearney-digicatapult:
So I tried to stay as close to main branch behaviour as possible but giving callers a new option.

* **main** `port` is required and the transport always listens with no opt-out
* **here** `port` is optional but if provided transport can bind and listen as before, if omitted only registers the route on the caller's `app` and the caller listens

The discriminated union preserves the old behaviour along with adding the new

A more explicit/obvious listener ownership would involve a change to the constructor instead of overloading so we could have:

```ts
new DidCommHttpInboundTransport.standalone({ port })
new DidCommHttpInboundTransport.attachTo({ app, path: '/didcomm' })
```

This would be breaking and also involve test & demo changes

TimoGlastra:
Okay let's rework this in 0.8 then

TimoGlastra (approved):

PR #2920, opened 2026-08-25
feat: support KMS backend selection for issuer access token keys

### LINKED ISSUE #2919: feat: allow selecting a KMS backend for OpenID4VC issuer access-token keys
## Problem

`OpenId4VcIssuerService.createIssuer()` creates the access-token signing key through `KeyManagementApi.createKey()`, but `OpenId4VciCreateIssuerOptions` does not expose a KMS backend. As a result, applications that register multiple KMS backends cannot select a non-default backend for an issuer access-token signer without changing the global default backend.

`rotateAccessTokenSigningKey()` has the same limitation. Rotation also needs to remember the previous key's backend so the old key is deleted through the backend that owns it, especially when rotating between backends.

Changing the global default is unsafe for multi-tenant agents because unrelated and concurrent key operations can require different backends.

## Proposed API

- Add optional `accessTokenSignerBackend?: string` to `OpenId4VciCreateIssuerOptions`.
- Persist the selected backend on `OpenId4VcIssuerRecord` as an optional field.
- Allow `OpenId4VcIssuerApi.rotateAccessTokenSigningKey()` to accept the same optional key type/backend options.
- During rotation:
  - default the new key to the issuer record's current backend;
  - allow an explicit backend to move the signer to another registered backend;
  - delete the previous key through the previous backend;
  - update the record to the new backend.

The field remains optional. Existing calls and records therefore continue using the configured default KMS backend with no behavior change.

## Example

```ts
await agent.modules.openid4vc.issuer.createIssuer({
  credentialConfigurationsSupported,
  accessTokenSignerBackend: 'openbao',
})

await agent.modules.openid4vc.issuer.rotateAccessTokenSigningKey(issuerId, {
  accessTokenSignerBackend: 'openbao',
})
```

## Acceptance criteria

- Issuer creation forwards the optional backend to `KeyManagementApi.createKey()`.
- The selected backend is persisted on the issuer record.
- Rotation creates the new key on the selected/current backend and deletes the old key on its owning backend.
- Omitting the option preserves current default-backend behavior.
- Tests cover explicit backend selection, legacy/default behavior, and rotation between backends.

This is needed for server-side issuers that keep non-exportable signing keys in a remote KMS while retaining another backend as the default for unrelated key operations.

sign-mark:
## Summary

- add an optional `accessTokenSignerKmsBackend` to OpenID4VC issuer creation
- persist the backend that owns an issuer access-token signing key, including Drizzle SQLite and PostgreSQL storage
- allow key rotation to keep the current backend or explicitly move to another backend
- delete the previous signing key through its owning backend
- preserve the default KMS behavior for existing calls and legacy issuer records

Closes #2919.

## Rationale

Applications can register multiple Credo KMS backends, but issuer access-token keys were always created through the default backend. Changing the global default is not safe for multi-tenant agents with concurrent key purposes. Persisting the selected backend also makes rotation deterministic when moving a signer between backends.

## Backwards compatibility

`accessTokenSignerKmsBackend` is optional and the new Drizzle column is nullable. When omitted on a new issuer, or absent from an existing issuer record, key creation and rotation continue to use the configured default KMS backend. No legacy-data backfill is required.

## Verification

- 24 focused OpenID4VC unit tests passed
- full `pnpm types:check` passed
- `@credo-ts/openid4vc` and `@credo-ts/drizzle-storage` builds passed
- generated SQLite migration applied successfully from an empty database
- targeted Biome check and `git diff --check` passed

Tests cover explicit backend selection, backend persistence, rotation between backends, and deletion of the previous key through its original backend.


sign-mark:
@TimoGlastra @IliasElbarnoussi This is a small backwards-compatible OpenID4VC API addition for selecting the KMS backend of issuer access-token keys. It also persists the owning backend so rotation and old-key deletion remain deterministic across backends. Focused tests and the full TypeScript check pass. Feedback on the proposed API would be appreciated.

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerServiceOptions.ts
```diff
@@ -738,6 +738,11 @@ export type OpenId4VciCreateIssuerOptions = {
    */
   accessTokenSignerKeyType?: Kms.KmsCreateKeyTypeAsymmetric
 
+  /**
+   * KMS backend to use for the access token signing key. If not provided, the default KMS backend is used.
+   */
+  accessTokenSignerBackend?: string
```
TimoGlastra: Maybe we should call it accessTokenSignerKmsBackend, to indicate it's the kms backend?

packages/openid4vc/src/openid4vc-issuer/repository/OpenId4VcIssuerRecord.ts
```diff
@@ -84,6 +89,7 @@ export class OpenId4VcIssuerRecord extends BaseRecord<DefaultOpenId4VcIssuerReco
    */
   public accessTokenPublicKeyFingerprint?: string
   public accessTokenPublicJwk?: Kms.KmsJwkPublicAsymmetric
+  public accessTokenSignerBackend?: string
```
TimoGlastra: We also need to persist it in the drizzle record and add a migration

TimoGlastra (approved):
Two comments, but this is a great addition!

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerServiceOptions.ts
sign-mark: Renamed throughout to `accessTokenSignerKmsBackend` for clarity and consistency with Credo KMS terminology in a08e476.

packages/openid4vc/src/openid4vc-issuer/repository/OpenId4VcIssuerRecord.ts
sign-mark: Fixed in a08e476. The backend is now mapped by the Drizzle adapter and persisted in a nullable `access_token_signer_kms_backend` column for both SQLite and PostgreSQL. Migrations and snapshots were generated with the repository CLI; legacy NULL values continue to select the default KMS backend. I also applied the SQLite migration chain successfully from an empty database.

genaris (commented):
Hey @sign-mark can you fix the formatting issues?

sign-mark:
> Hey @sign-mark can you fix the formatting issues?

Fixed

PR #2921, opened 2026-08-28
feat(json-ld): abstract json-ld document loader into a module

rmlearney-digicatapult:
`documentLoader` is currently owned by `W3cCredentialsModule`. Incoming VC 2.0 support requires JSON-LD document loader functionality independent of this module

Abstract into `JsonLdModule` with backwards-compatible shim and mark `W3cCredentialsModuleConfigOptions` (which only configured the document loader) as `@deprecated` - new consumers should use `JsonLdModuleConfig`.

Constructor dependency of `W3cJsonLdCredentialService` changed from `W3sCredentialsModuleConfig` to `JsonLdModuleConfig`

All existing loader functionality unchanged, `new W3CCredentialsModule({ documentLoader })` unchanged

**NB**
This abstraction was the easiest way to address the comment by @genaris on #2827


packages/core/src/modules/vc/linked-data-proofs/W3cJsonLdCredentialService.ts
```diff
@@ -33,14 +33,11 @@ import { SignatureSuiteRegistry } from './SignatureSuiteRegistry'
 @injectable()
 export class W3cJsonLdCredentialService {
   private signatureSuiteRegistry: SignatureSuiteRegistry
-  private w3cCredentialsModuleConfig: W3cCredentialsModuleConfig
+  private jsonLdModuleConfig: JsonLdModuleConfig
```
TimoGlastra, line 36: what happens if you provide both jsonLdModule and W3c module? Due to the alias i suspect it's registered twice? Should we not just keep the W3cModuleConfig, and if resolving do something like w3cModuleLoader ?? jsonldmoduleloader? and in the future we migrate it?

rmlearney-digicatapult: I'll take a look. The W3cV2 PR is going to change to use the Credo document loader and I'd rather not access it through W3cCredentialModule

rmlearney-digicatapult:
@TimoGlastra have added guard logic and (temporary) tests to ensure the runtime resolution has no order-dependency to ensure `JsonLdModule` token resolves to `JsonLdModule` document loader when both modules configured

TimoGlastra (approved):

PR #2922, opened 2026-08-29
feat: add support for iso 18013-7 dc api

TimoGlastra:
Adds support for creating mdoc verification sessions for the org-iso-mdoc protocol over DC API. This has been tested on both iOS and Android and can be used with e.g. the expo-digital-credentials-api library (see https://github.com/animo/expo-digital-credentials-api/pull/8).

DC API requires HPKE, this is not supported by Askar or Node.JS natively, so we use the `hpke` library for this. We still use the raw ECDH key from askar, so we don't have to export the private key (only the derived key). We do require `@panva/hpke-noble` for the other features in the Askar KMS, so it works in both RN And Node.JS. The Node KMS depends on the Node crypto. (there's an issue open in Askar to support HPKE). 

HPKE will probably see more usages in the future (e.g. OpenID4VP/HAIP will also adopt it), and now we have a generalized implementation.

Since this adds a new protocol on the verifier side, it introduces new exchange records.

https://github.com/user-attachments/assets/11b133a3-73cb-4c36-aed6-b763f2b5efe7


packages/core/src/modules/mdoc/MdocDcApiService.ts
```diff
@@ -0,0 +1,527 @@
...
+    }
+  ) {
+    const x509ModuleConfig = agentContext.dependencyManager.resolve(X509ModuleConfig)
+
+    let trustedCertificates = options.trustedReaderCertificates?.map((certificate) =>
+      typeof certificate === 'string' ? certificate : certificate.toString('pem')
+    )
+
+    if (!trustedCertificates) {
+      const trustedIssuers = await TrustedIssuerContext.getTrustedIssuersForVerification(agentContext, {
```
TimoGlastra: Should fallback to the legacy callback

packages/core/src/modules/mdoc/MdocDcApiService.ts
```diff
@@ -0,0 +1,540 @@
...
+        type: 'sessionTranscriptBytes',
+        sessionTranscriptBytes: decrypted.sessionTranscript.encode(),
+      } as const
+
+      if (options.trustedCertificates) {
+        await deviceResponse.verify(agentContext, {
+          trustedCertificates: options.trustedCertificates,
+          sessionTranscriptOptions,
+          now: options.now,
+        })
```
hacdias, line 184: Should we not provide the `deviceRequest` in order to make sure it matches the `deviceResponse`?

packages/core/src/modules/mdoc/MdocService.ts
```diff
@@ -86,4 +98,44 @@ export class MdocService {
...
+    await this.mdocVerificationSessionRepository.delete(agentContext, verificationSession)
+
+    const deleteAssociatedKey = options?.deleteAssociatedKey ?? true
+    if (deleteAssociatedKey) {
+      const kms = agentContext.resolve(KeyManagementApi)
+
+      // Returns false if the key was already gone, which is the state we want it in
+      await kms.deleteKey({ keyId: verificationSession.sessionKeyId })
+    }
+  }
```
hacdias, line 140: What if `deleteKey` throws?

packages/core/src/modules/mdoc/MdocDcApiService.ts
TimoGlastra: Hmm yeah good catch. We need to verify this. 🤦

packages/core/src/modules/mdoc/MdocService.ts
TimoGlastra: what do you mean? It throws an error, is that not desired? Or do you mean we should first remove the key, so we still have the reference if it fails?

packages/core/src/modules/mdoc/MdocDcApiService.ts
TimoGlastra: I will add this to the mdoc library.

TimoGlastra: Ah it already lives in the device response verify. but it's a bit minial (just one error). i will make it more like the DCQL verification. So it provides a full overview of what is present/what is not. As mdoc is more lenient in missing credentials (multiple creds in a doc request usually means OR)

packages/core/src/modules/mdoc/MdocService.ts
hacdias: I mean: we delete the session, then we fail to delete the key and we have a dangling key. Is that the desired behavior?

packages/core/src/modules/mdoc/MdocDcApiService.ts
```diff
@@ -0,0 +1,540 @@
...
+        sessionKeyId: keyId,
+        expiresAt: new Date(Date.now() + (options.expiresInSeconds ?? defaultExpiresInSeconds) * 1000),
+      })
+
+      await this.mdocVerificationSessionRepository.save(agentContext, verificationSession)
+      this.emitStateChangedEvent(agentContext, verificationSession, null)
+
+      return { verificationSession, request }
+    } catch (error) {
+      await kms.deleteKey({ keyId })
```
hacdias, line 123: If this throws, the initial error will be swallowed.

packages/core/src/modules/mdoc/MdocService.ts
TimoGlastra: I will switch it around

packages/core/src/utils/credentialUse.ts
```diff
@@ -136,7 +136,7 @@ export async function useInstanceFromCredentialRecord<Record extends CredentialR
   credentialRecord,
   useMode,
   agentContext,
-  updateMode,
+  updateMode = CredentialMultiInstanceUseUpdateMode.RefetchAndUpdateWithLock,
```
hacdias: Shouldn't we add a small changelog line about this change?

.changeset/mdoc-dc-api-annex-c.md
```diff
@@ -0,0 +1,29 @@
+---
```
hacdias, line 1: Nit: the other changelog entries don't do line wrapping, this one does.

hacdias (approved):

PR #2924, opened 2026-08-31
refactor(didcomm-v2): make interfaces for DIDComm envelopes

genaris:
DIDComm v1 and v2 each had an envelope service, but with different APIs. Each caller had to know the version before it could make a call. This caused duplicated `if (didcommV2)` chains in `DidCommMessageSender` and `DidCommMessageReceiver`. The two copies in the sender had already drifted: only one of them checked `supportedDidCommVersions`. The envelope key types also said Ed25519 while they held v2 keys, so the code used unsafe casts.

In this PR, enveloping now follows the same pattern as message pickup, credentials and proofs: one interface, one implementation per version, and a registry that resolves the implementation.

- `DidCommEnvelopeProtocol`:  the interface. It declares `pack`, `unpack`, `supportsPacking`, `supportsUnpacking` and `buildReturnRouteKeys`.
- `DidCommV1EnvelopeProtocol`: delegates to `DidCommEnvelopeService`. No cryptography moved.
- `DidCommV2EnvelopeProtocol`: wraps `DidCommV2EnvelopeService`. It absorbs the v2 code that lived in the sender and the receiver: Forward wrapping, plaintext build, v1-shape normalization, signer resolution, nested-JWS unwrap, and return-route `skid` derivation.
- `DidCommEnvelopeProtocolRegistry`: resolves a protocol. The caller then calls `pack` or `unpack` on the protocol.
  - `getProtocolForOutbound({ message, connection, keys })`
  - `getProtocolForInbound(message)`
  - `getProtocolForDidCommVersion(version)`
  - `assertVersionEnabled(message)`

Outbound selection applies three rules, in order:

1. A message that declares `supportedDidCommVersions` restricts the choice.
2. The connection version decides. Connectionless exchanges default to v1.
3. The protocol must be enabled in `didcommVersions` and must accept the key material.

Related cleanups:

- `DidCommConnectionRequestMessage` and `DidCommConnectionResponseMessage` now declare `supportedDidCommVersions = ['v1']`. This replaces a string match on the message type.
- New `DidCommEnvelopeKey` union (Ed25519, X25519, P-256, P-384) removes every `as never` cast on envelope keys. The four readers that need Ed25519 now narrow with `.is(Kms.Ed25519PublicJwk)`.
- `DidCommModuleConfig`: the `acceptsV2` and `sendsV2` getters collapsed into `isSupported(didcommVersion)`.

Some Changes: 

- The `didcommVersions` config gate now also applies to outbound messages. Before, it applied only to inbound.
- The `supportedDidCommVersions` check now runs on both send paths. Before, the session path skipped it.
- The v2-pack fallback to v1 exists in one place: `DidCommMessageSender.encryptMessage`. It logs at warn level. Before, it existed in three places and logged at debug level.
- A v2 key-agreement key can no longer reach `findByKeys` typed as Ed25519.

Other than that, no changes on APIs on behaviour, so it should work like it was on feat/didcomm-v2 branch up to now.

cc @tarunvadde to take a look of this to find any incompatibility on the different use cases (v1 only, v2 only, v1+v2, mediated cases, etc.)


tarunvadde:
Took a look, focused on what actually changes behavior in this PR, and checked the tricky parts against the DIDComm v2.1 spec. The registry as a single selection point is a nice improvement, and wrapInForward wraps in the order the spec says. Four things I'd fix:

1. The new Ed25519 checks in findConnection (both spots) and in the mediator service kill the key-based connection fallback for v2. Those paths only run for v2 messages, and v2 envelope keys are always X25519 or P-256/P-384, never Ed25519 (the spec only uses Ed25519 for signing). The comment above the check isn't right either. DidRecord tags do index X25519 fingerprints, getTags uses mapX25519ToEd25519: false and PeerDidRegistrar stores exactly those, so findByKeys did match these keys before. In practice, connection recovery after a rotation, or when the peer sends the long form of a did:peer:4 while we stored the short form, now fails. On the mediator side, a mediate-request where from is missing or in a different form now throws "No connection associated with incoming message" instead of resolving through the sender key. I'd widen findByKeys and findByTheirDidOrSender to accept any PublicJwk instead of filtering the keys out before the call.

2. The isSupported('v2') check in wantsV2 is new. Before, we packed v2 whenever the connection said v2, without looking at config. Now a stored v2 connection on an agent running with the default ['v1'] silently picks v1. With X25519 recipient keys the v1 pack even succeeds and the peer just can't decrypt it. With P-256 you get an "Unsupported key conversion" error that says nothing about the version mismatch. I think this should throw with a clear error when the connection is v2 but v2 isn't enabled.

3. assertMessageSupportsConnection now also runs when we reply over a return-route session. That path had no version check before. So if a peer we marked as v2 falls back to v1 (which is what the spec recommends when profiles don't match) and sends a v1-packed pickup 2.0 status-request with return_route, our reply now throws, because the status message declares v1 only and the check compares against the stored connection version, not the envelope the request came in. Before this change the reply went out over the open session.

4. resolveReturnRouteSkid changes a default. The old receiver left the skid unset when the connection had no did yet, and the pack fell back to the sender key's own kid. Now it returns a did:key of our recipient key. A did:key skid can't match a did:peer from, and the spec is strict here: "The from attribute in the plaintext message MUST match the skid attribute in the encryption layer", a mismatch is an error. There's also a subtle case where ourDid comes from the incoming message's to field, which can still be the prior DID during a rotation, while the plaintext from is already the new DID.

Smaller stuff: supportsUnpacking on the v2 protocol says yes to signed messages but unpack then rejects them, which is why receiveSignedMessage has to go around the registry. The envelope protocol gets resolved twice per incoming message and the protected header ends up decoded five or six times along the way. isDidCommV2AuthcryptMessage and isDidCommV1EncryptedMessage have no callers left outside their tests now that the logic is re-inlined. And the comment in the sender tests says the registry covers version selection, but didcommVersions is never set there, so wantsV2 is always false and the v2 path is never exercised.

genaris:
Thanks @tarunvadde , merging so we can test it.

PR #2926, opened 2026-09-07
feat(x509): expose ASN.1 parse options and raise the default node limit

hacdias:
Updates `@peculiar/*` packages, which enforce the asn1js parse limits. The 10000 node default rejects any CRL over ~3300 entries while `fetchCrl` downloads up to 10 MB, and an unparseable CRL counts as reachable-but-invalid, so `SoftFail` does not tolerate it: a CA with a large CRL would hard-fail chain validation in every mode but `Disabled`.

- Exposes the limits as `X509ParseOptions` on the parsing factories, on the `X509Service` and revocation methods that parse, and as `parseOptions` on the module.
- Adds `defaultX509ParseOptions` with `maxNodes` at 1500000, sized to that 10 MB cap. On `X509ModuleConfig` for what resolves through the config, and as a default parameter on the six factories, which mdoc, sd-jwt, dcql and openid4vc call directly.
- `X509Service.parseCertificateSigningRequest` now takes an `agentContext` first, like every other method on it. `X509Api` is unchanged.
- Builds CRL distribution points from their DER: `@peculiar/asn1-x509` 2.9.4 added an `exports` map and resolves to ESM, `@peculiar/x509` declares none and resolves to CommonJS, so the two get separate schema registries and the lookup fails.

Creating large CRLs is still capped, the generators re-parse their own output under the default limits. That is the trusted path, fixed upstream in https://github.com/PeculiarVentures/x509/pull/150 and https://github.com/PeculiarVentures/asn1-schema/pull/156.


hacdias:
@TimoGlastra I'm wondering whether or not we should wait for the linked PRs. The last version addresses the case of parsing certificates, CRLs, etc. However, the x509 package also does some back and forth parsing inside the constructors and generators, and doesn't take any parsing options (I would argue that parsing options don't belong on a creator), so the PRs I opened there bypass the restrictions when creating / generating.

This means that with this PR it won't be possible to create CRLs/certificates with more than 10 000 ASN.1 nodes (parser default). I'd rather wait, but I have no idea how long it'll take. The last PRs took longer than 3 months and if we want to wait we need to pin all versions. Another option is to patch the dependency in order to bypass the restrictions for now during the creation/generation code paths, which, in my view, are trusted paths.

packages/core/src/modules/x509/utils/extensions.ts
```diff
@@ -173,7 +174,17 @@ export const createCrlDistributionPointsExtension = (
...
-  return new CRLDistributionPointsExtension([distributionPoint], options.markAsCritical)
+  // The extension is built from its DER rather than through `@peculiar/x509`'s
+  // `CRLDistributionPointsExtension`, because that class resolves the ASN.1 schema of the values it
+  // is given through its own copy of `@peculiar/asn1-x509`. `@peculiar/asn1-x509` declares an
+  // `exports` map that resolves to its ESM build, while `@peculiar/x509` declares none and so
+  // resolves to its CommonJS `main` even when imported from ESM, pulling in the CommonJS build of
+  // `@peculiar/asn1-x509` instead. Each build keeps its own schema registry, so a value created
+  // here has no schema over there. Serializing keeps the values and their schema within a single
+  // instance, as the other extensions below do.
+  const value = AsnConvert.serialize(new CRLDistributionPoints([distributionPoint]))
```
TimoGlastra, line 185: Should we link to https://github.com/PeculiarVentures/x509/issues/145?

packages/core/src/modules/x509/__tests__/x509CrlTestUtils.ts
```diff
@@ -149,7 +157,15 @@ export async function generateLeafWithPartitionedDistributionPoints(
       notBefore: options.notBefore,
       notAfter: options.notAfter,
       serialNumber: options.serialNumber,
-      extensions: [new x509.CRLDistributionPointsExtension(distributionPoints)],
+      // Built from DER rather than `x509.CRLDistributionPointsExtension`, see
+      // `createCrlDistributionPointsExtension`.
```
TimoGlastra, line 161: This is quite unfortunate and brittle

TimoGlastra (approved):

TimoGlastra:
We could wait a bit and see how quickly the PRs get merged. Patches don't really work for published packages, as all dependants om Credo wouldn't have them.

hacdias:
Let's wait a bit then.

PR #2928, opened 2026-09-08
fix(webvh): sign attested resources with an assertionMethod key

hacdias:
Attested resource proofs use the `assertionMethod` proof purpose, but the signing key was selected as `verificationMethod[0]` without checking that it is referenced from `assertionMethod`. Resources were written successfully and then failed to resolve with `Resolved resource proof is invalid.`

The mismatch dates to #2396, but was masked by the vendored `eddsa-jcs-2022` cryptosuite, which dereferenced the key without validating the verification relationship. #2925 routed verification through the core `W3cDataIntegrityApi`, which does validate it, so this surfaces there and has not been in a stable release.

- Resource registration selects a verification method authorized for `assertionMethod`, preferring one the agent holds a key for, and validates an explicitly passed `options.verificationMethod` against the purpose.
- `createDID` now references the created key from `assertionMethod`; previously didwebvh-ts placed it in `authentication` only, so no did the registrar created could sign a verifiable resource. Updating a did no longer drops the reference, and drops stale ones for removed keys.
- `verifyProof` asserts the expected proof purpose and logs the verifier's issues.

Existing dids need an `agent.dids.update` adding the reference, and the updated log republished, before their resources verify.

### DIFF
--- a/.changeset/webvh-assertion-method-proofs.md
+++ b/.changeset/webvh-assertion-method-proofs.md
@@ -1,0 +1,18 @@
+---
+"@credo-ts/webvh": patch
+---
+
+Sign did:webvh attested resources with a verification method authorized for the proof purpose
+
+Attested resource proofs use the `assertionMethod` proof purpose, but the signing key was selected as
+`verificationMethod[0]` without checking that it is referenced from `assertionMethod`. Resources were
+written successfully and then failed to resolve with `Resolved resource proof is invalid.`
+
+- Resource registration now selects a verification method authorized for `assertionMethod`, and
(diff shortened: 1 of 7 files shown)


rmlearney-digicatapult:
@hacdias would be great if you could help out over at `didwebvh-ts` where we're hoping to get 3.0 released soon :)

hacdias:
@rmlearney-digicatapult thanks for the pointing. I opened two issues over there: https://github.com/decentralized-identity/didwebvh-ts/issues/180, https://github.com/decentralized-identity/didwebvh-ts/issues/179

genaris (approved):

PR #2934, opened 2026-09-11
feat(DidDocument): keyType and Purpose-aware Verification Method helpers

rmlearney-digicatapult:
Multiple points in the codebase use the same logic to resolve keys within a DID document

Created 2 self-describing helpers inside `didDocument.ts` for:

* `findVerificationMethodsByPurpose()`
* `findVerificationMethodsByTypeAndPurpose()`

Wired these helpers into call sites throughout the codebase to simplify reasoning:
* DQCL
* DidComm
* PEX
* Cheqd
* Webvh
* W3cJwt

Helpers necessary for fixing currently-broken `linked-data-proofs` when using `Ed25519-2020` keys in subsequent PR

### DIFF
--- a/packages/anoncreds/src/formats/DataIntegrityDidCommCredentialFormatService.ts
+++ b/packages/anoncreds/src/formats/DataIntegrityDidCommCredentialFormatService.ts
@@ -557,18 +557,18 @@ export class DataIntegrityDidCommCredentialFormatService
 
     let verificationMethod: VerificationMethod
     if (issuerVerificationMethod) {
-      verificationMethod = didDocument.dereferenceKey(issuerVerificationMethod, ['authentication', 'assertionMethod'])
+      verificationMethod = didDocument.dereferenceKey(issuerVerificationMethod, [
+        'assertionMethod',
+        'verificationMethod',
+      ])
     } else {
-      const vms = didDocument.authentication ?? didDocument.assertionMethod ?? didDocument.verificationMethod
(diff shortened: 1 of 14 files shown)


packages/anoncreds/src/formats/DataIntegrityDidCommCredentialFormatService.ts
```diff
@@ -557,18 +557,18 @@ export class DataIntegrityDidCommCredentialFormatService
 
     let verificationMethod: VerificationMethod
     if (issuerVerificationMethod) {
-      verificationMethod = didDocument.dereferenceKey(issuerVerificationMethod, ['authentication', 'assertionMethod'])
+      verificationMethod = didDocument.dereferenceKey(issuerVerificationMethod, [
+        'assertionMethod',
+        'verificationMethod',
```
TimoGlastra, line 565: Why do we replace authentication with verification method in both places? That seems like a change in behavior?

TimoGlastra (approved):
A small note but otherwise LGTM!

packages/anoncreds/src/formats/DataIntegrityDidCommCredentialFormatService.ts
rmlearney-digicatapult: Oh let me look at that

rmlearney-digicatapult: Restored to original purposes

TimoGlastra (approved):

PR #2935, opened 2026-09-11
fix(indy): re-point docker image repository

rmlearney-digicatapult:
Dockerhub bcgov image marked deprecated, recommends using `gchr` image instead


genaris (approved):
OK, although https://github.com/bcgov/von-image/pkgs/container/von-image is archived as well.

rmlearney-digicatapult:
Ugh. Not great.

PR #2936, opened 2026-09-14
fix(openid4vc): issue bearer credentials for proof-free credential requests

### LINKED ISSUE #2927: OpenID4VCI issuer cannot issue proof-free bearer SD-JWT VCs
The core SD-JWT API supports issuing a VC without `holder`, but the OpenID4VCI issuer cannot complete a proof-free credential request.

For a `dc+sd-jwt` credential configuration that omits both `cryptographic_binding_methods_supported` and `proof_types_supported`, a Credential Request without `proof`/`proofs` is rejected with:
```
invalid_proof: Missing required proof(s) in credential request
```
If that check is bypassed, the issuer later requires the mapper to return exactly one credential per verified holder-binding key. With no proof, there are zero keys, so a bearer credential cannot be returned.
OpenID4VCI 1.0 permits bearer credentials ([§14.2](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-binding-of-the-credential-w)). Could the issuer support this case by allowing a proof-free request and returning one credential without holder binding?

Tested with `@credo-ts/openid4vc@0.7.1-alpha-20260904141201`

AirKyzzZ:
Closes #2927

A credential request with no `proof`/`proofs` was always rejected with `invalid_proof`, even for a configuration that declares neither `proof_types_supported` nor `cryptographic_binding_methods_supported`. Proofs are only required when `proof_types_supported` is present (§8.2), and §14.2 allows credentials that aren't bound to a holder key, so that request should give back a bearer credential. Core already signs SD-JWTs without a `holder`, it was only the OID4VCI layer in the way. Three things had to move: the proof check, the one-credential-per-holder-binding-key check, and the response shape, which has no request shape to mirror here and now follows the negotiated version instead.

One judgement call worth a look. I still require a proof when a configuration declares only `cryptographic_binding_methods_supported`, which isn't literally §8.2, but `allowedProofTypes` has always defaulted to jwt when `proof_types_supported` is missing, so following the spec to the letter would turn those issuers into bearer issuers on upgrade. Say if you'd rather have an explicit opt-in on the issuer config instead. `holderBinding` on the mapper options is optional now, which is a type break, so also say if the changeset should be more than `patch`.

The holder still refuses these offers, I left that out of scope and can do it separately. types:check, biome and the openid4vc unit and e2e suites are clean.


TimoGlastra:
Thanks for the PR @AirKyzzZ. It would be great if you can also at least address the holder side. So it's a complete feature. We can scope presentation separately.

I think we have the default for proof types as older drafts were not as explicit, and we still support those draft. I think with the behvaiour you introduced we should remove the default (it would simplify the other code as well). 

So maybe we should only set the default if you're using non-1.0 version (or better we figure out which draft it changed)? i think that would be more correct anyways?

genaris (approved):
I think @TimoGlastra feedback is addressed, but I leave him the last word before merging.

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerServiceOptions.ts
```diff
@@ -618,8 +618,11 @@ export interface OpenId4VciCredentialRequestToCredentialMapperOptions {
    *
    * All keys and dids have a verified proof, or in the case a key attestation is provided
    * are attested by a key attestation. Ensure the issuer of the key attestation is trusted.
+   *
+   * Undefined if the credential request contained no proofs, in which case a single credential
+   * without holder binding should be returned.
    */
-  holderBinding: VerifiedOpenId4VcCredentialHolderBinding
+  holderBinding?: VerifiedOpenId4VcCredentialHolderBinding
```
TimoGlastra: Making holderBinding optional is a breaking change. So we should mark this as a minor change and include it in 0.8.

 if your code depends on the value being defined (which most flows do i think) pushing this change will break their code.

.changeset/openid4vci-bearer-credentials.md
```diff
@@ -0,0 +1,26 @@
+---
+"@credo-ts/openid4vc": patch
```
TimoGlastra:
```suggestion
"@credo-ts/openid4vc": minor
```

"On the issuer, `holderBinding` in `OpenId4VciCredentialRequestToCredentialMapperOptions` is now
optional and is undefined for these requests" 

This makes it a breaking change, even though the behvaiour at runtime does not change if you correcly provide proof types, the TypeScript build will now suddenly fail if you don't handle the undefiend case

packages/openid4vc/src/openid4vc-issuer/OpenId4VcIssuerService.ts
```diff
@@ -742,18 +742,18 @@ export class OpenId4VcIssuerService {
       const credentials = await this.getSignedCredentials(agentContext, signOptionsOrDeferral, {
```
TimoGlastra: Do we correctly check that if no proof is required, the credentials are not issued bound to a key? or is that not in the scope of Credo? Are there cases where you don't require proof, but do issue to a key (it's weird because then the wallet won't be able to present the credential)

packages/openid4vc/tests/openid4vci.e2e.test.ts
```diff
@@ -536,4 +553,75 @@ pUGCFdfNLQIgHGSa5u5ZqUtCrnMiaEageO71rjzBlov0YUH4+6ELioY=
...
+    expect(sdJwtVc.holder).toBeUndefined()
+    expect(sdJwtVc.kmsKeyId).toBeUndefined()
+    expect(sdJwtVc.payload).not.toHaveProperty('cnf')
+    expect(sdJwtVc.payload.vct).toEqual('UniversityDegreeCredential')
+
+    await holderTenant.endSession()
+    await issuerTenant.endSession()
+  })
+
+  it('throws when a configuration requires a proof but no credentialBindingResolver is provided', async () => {
```
TimoGlastra: can we also add a test where we cover the issuer side rejecting it because proof must be provided?

packages/openid4vc/src/openid4vc-holder/OpenId4VciHolderService.ts
```diff
@@ -1482,15 +1494,17 @@ export class OpenId4VciHolderService {
...
-            )
-          }
+          if (options.jwkThumbprintKmsKeyIdMapping) {
+            const jwkThumbprint = TypedArrayEncoder.toBase64(mdoc.deviceKey.getJwkThumbprint())
+            const kmsKeyId = options.jwkThumbprintKmsKeyIdMapping[jwkThumbprint]
+            if (!kmsKeyId) {
+              throw new CredoError(
+                `Missing kmsKeyId for jwk with thumbprint ${jwkThumbprint}. A credential was issued for a key that was not in the credential request.`
+              )
+            }
```
TimoGlastra: This can result in maybe some cases where we skip the check if `jwkThumbprintKmsKeyIdMapping` is not defined. I think if the cred has a key, it should be checked against `jwkThumbprintKmsKeyIdMapping always (so inverse the check so we don't skip it)

Also some questions this raises
- Do we check that if no proof is required the cred is also not bound to a key (i think we should)
- That automatically means Mdoc is not supported without proof types (as it requires a key, it's always key-bound)
- We should then also limit creating credenital configurations without proof_types_supported for mdoc / bearer mdoc. So it doesn't fail very late in the process suddenly


A wallet should not check the metadata (and replicate the isProofRequired check) before requesting a credential, as the callback will not be called and otherwise it will silently issue a credential not bound to a key. This is another change in behaviour that really makes this breaking i think

TimoGlastra:
Sorry i have a few more comments/concerns. Although the code makes sense and is quite straightforwrad, the impact of this is quite big, and so i want to make sure it's properly released.

Especially since presentation does not support this yet, this will silently allow issuance of bearer credentials into the wallet which then cannot be presented.

PR #2937, opened 2026-09-14
test: pin conformance suite to release-v5.2.2

AirKyzzZ:
The conformance job has been failing on every PR for a while, the pinned digests are gone from the registry:

```
manifest for registry.gitlab.com/openid/conformance-suite@sha256:0c86e0ae... not found: manifest unknown
```

#2903 pinned those digests on purpose, because the regenerated mdoc document-signer certificate from [conformance-suite!2123](https://gitlab.com/openid/conformance-suite/-/merge_requests/2123) was on master but not in any release tag yet. That stopped being true a month ago, the fix is in `release-v5.2.2` onwards, so this does what the TODO there asked and goes back to a release tag. Digests off `latest` get garbage collected once the tag moves, which is what bit us, so a release tag also avoids a repeat.

To be clear about scope: this gets the images pulling again, it does not get the job green on its own. `release-v5.2.2` also added a manual `ExpectVerifierSuccessfulVerificationPage` review step that the old master digest did not have, so the happy-flow modules now run the whole protocol successfully and then park in `WAITING` waiting for a screenshot upload that never comes in CI. That needs a harness change rather than a different tag, and there is no release that has the cert fix but not the review step, so I am keeping this PR to the pin and doing the harness part separately. Evidence in the comment below.


TimoGlastra (approved):

AirKyzzZ:
CI green fix posted in a new pr : https://github.com/openwallet-foundation/credo-ts/pull/2938

genaris (commented):
Yes, I tried locally a few weeks ago and unfortunately reached the same issue as you. It would be great to know what's missing in Credo's implementation (or test suite execution) to make it succeed.

genaris:
Closing in favour of #2938

PR #2938, opened 2026-09-14
test: answer conformance suite manual review steps

AirKyzzZ:
Two things, the second is why the job still hung after fixing the first.

The pinned digests are gone from the registry, so every run died on `manifest unknown`. #2903 pinned them deliberately, because the regenerated mdoc document-signer certificate from [conformance-suite!2123](https://gitlab.com/openid/conformance-suite/-/merge_requests/2123) was on master but not in any release tag yet. It is now, from `release-v5.2.2` onwards, so this goes back to a release tag as the TODO there asked. Digests off `latest` get garbage collected once the tag moves, a release tag doesn't.

That alone still hangs. `release-v5.2.2` also added an `ExpectVerifierSuccessfulVerificationPage` step that asks a human to upload a screenshot of the verifier showing the credential. Credo passes every automated assertion before it, then parks in `WAITING`, the no-progress watchdog gives up, and the next module reusing the same alias interrupts it. Those log entries carry an upload id, so the client now posts a 1x1 placeholder PNG to `/api/log/{id}/images/{uploadId}`, the same endpoint the suite's own upload form uses. The module then reaches `FINISHED` with result `REVIEW`, which is accepted alongside PASSED/WARNING/SKIPPED. There is nothing real to screenshot, credo is a library with no UI, and the upload only fires after the watchdog would otherwise have failed so it can't mask a genuine hang.

Also made the httpd host port overridable with `OIDF_HTTPD_HOST_PORT`, 8443 was taken on my machine and there was no way to move it.

Green locally on both plans with all modules, and Conformance Tests passes on CI here.


genaris (approved):
This is the right way to fix it. There is no need to stack over the other one, since there are very little file changes. I'll close #2937 in favour of this one.

packages/openid4vc/tests/conformance/oid4vp-verifier.conformance.test.ts
```diff
@@ -134,15 +134,19 @@ describe('OIDF - oid4vp-1final-verifier-haip-test-plan', () => {
...
 
     // FIXME: will be addressed once merged and updated:
     // https://github.com/openwallet-foundation/sd-jwt-js/pull/378
-    if (signal.includes('oid4vp-1final-verifier-kb-jwt-iat-in-past')) return ['PASSED', 'WARNING', 'SKIPPED', 'FAILED']
+    if (signal.includes('oid4vp-1final-verifier-kb-jwt-iat-in-past')) {
+      return ['PASSED', 'WARNING', 'SKIPPED', 'REVIEW', 'FAILED']
+    }
 
-    return ['PASSED', 'WARNING', 'SKIPPED']
+    return ['PASSED', 'WARNING', 'SKIPPED', 'REVIEW']
```
TimoGlastra, line 149: Soo does this mean that `REVIEW` is seen a valid end-state? It seems we now do upload the image, won't it reach PASSED / WARNING then after that?

AirKyzzZ: No, it stays `REVIEW`. Uploading only unblocks the module so it can finish instead of hanging, the suite won't call it passed without a human actually looking at the screenshot.

PR #2942, opened 2026-09-15
fix(changeset): use patch version in did document helper changeset

rmlearney-digicatapult:
use patch version in did document helper changeset


TimoGlastra (approved):
Thanks 🙌

PR #2944, opened 2026-09-17
fix(didcommv2): fix sender binding

rmlearney-digicatapult:
Addresses critical spec conformance issue.

`§Message Layer Addressing Consistency` The `from` attribute MUST match `skid`, and `to` MUST contain the `kid`; any failure MUST be an error. `§Message Headers` also says the sender key's authorisation for the `from` DID MUST be verified by the recipient.

Implementation resolves `skid` for decryption and exposes resolved sender key, but does not validate or preserve the authenticated binding between envelope identities and plaintext addressing. Receiver then selects connections from plaintext `from` and `to` without the v2 envelope layer first enforcing those relationships.

* Update `DidCommV2Envelope.unpack` to ensure `from` matches `skid` and `to` contains `kid`
 
```text
_authcrypt:_
  require plaintext.from
  require DID(plaintext.from) == DID(skid)
  require plaintext.to contains the authenticated recipient identity

_anoncrypt:_
  allow absent from
  if from is present, preserve it as untrusted metadata only
  do not create authenticated sender identity
```

Tests for conformity including nested signed plaintext matching


rmlearney-digicatapult:
Additional issues discovered requiring different fix

PR #2945, opened 2026-09-17
feat: support array paths and several dcql fixes

TimoGlastra:
Fixes numerous edge cases with DCQL and SD-JWT mainly focused on array and nested disclosures. Was causing some bugs in our wallet where we showed different content on what would be shared vs what would actually be shared


packages/openid4vc/src/openid4vc-holder/OpenId4vpHolderService.ts
```diff
@@ -295,7 +296,7 @@ export class OpenId4VpHolderService {
             transaction_data_hashes_alg: transactionDataHahsesAlg,
           },
         }
-      })
+      }) as unknown as T[string]
```
hacdias: Nit: Isn't there a better way to set the type without the `as unknown`?

packages/core/src/modules/sd-jwt-vc/disclosureFrame.ts
```diff
@@ -50,3 +64,268 @@ export function applyDisclosuresForPayload(compactSdJwt: string, requestedPayloa
...
+ * another claims query of the claim set. Its path is then selected by that claims query anyway.
+ */
+export function getClaimPathsForDcqlClaimSet(
+  claimsQueryPaths: Array<Array<string | number | null>>,
+  output: unknown
+): ClaimPath[] {
+  const getPaths = (queryPath: Array<string | number | null>, value: unknown, path: ClaimPath): ClaimPath[] => {
+    // FIXME: logic should be updated once this PR is released:
+    // https://github.com/openwallet-foundation-labs/identity-common-ts/pull/261
+    // DCQL leaves `null`, `undefined` or nothing where no claim is selected
```
hacdias, line 96: It's already merged. Can't this already be updated?

hacdias (approved):

packages/core/src/modules/sd-jwt-vc/disclosureFrame.ts
TimoGlastra: It's a breaking change in the output, so unfortunately not. It would have to wait until credo 0.8

packages/openid4vc/src/openid4vc-holder/OpenId4vpHolderService.ts
TimoGlastra: i removed this cast. There's now a cast at the end, but it's just `as T` not an `as unknown as`

